PR #4416, opened 2024-06-19
Standalone ASM priority sampler and tag propagation

iunanua:
### What does this PR do?
Continuation of https://github.com/DataDog/dd-trace-js/pull/4291

Defines two new channels in `opentracing/propagation/text_map.js` in order to alter trace propagation:
- `dd-trace:span:inject` which publishes the trace span context and the carrier that is going to be injected in a downstream req 
- `dd-trace:span:extract` which publishes the trace span context and the carrier when a span is going to be extracted from an upstream req.

When standalone ASM is enabled it:
- replaces default `PrioritySampler` with `StandaloneAsmPrioritySampler` via `DatadogTracer` constructor. 
  - `StandaloneAsmPrioritySampler` searches for `_dd.p.appsec` in the tags in order to keep a trace, sets a 1 req/min rate limiter and disables user rules.
- when injecting a trace and if the trace doesn't contain `_dd.p.appsec` tag, removes `x-datadog- ` and datadog info from `tracestate`
- when extracting a trace if the trace doesn't contain `_dd.p.appsec` tag, resets the sampling priority.
- disables span stats and includes `'Datadog-Client-Computed-Stats': 'yes'` header when exporting traces to the agent to pretend stats computation is made by the tracer

ST [PR](https://github.com/DataDog/system-tests/pull/2657) to enable Standalone ASM tests

### Motivation
Support Standalone ASM billing

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/dd-trace/src/priority_sampler.js
```diff
@@ -66,11 +66,13 @@ class PrioritySampler {
...
 
-    const tag = this._getPriorityFromTags(context._tags)
+    const tag = this._getPriorityFromTags(context._tags, context)
 
     if (this.validate(tag)) {
       context._sampling.priority = tag
-      context._sampling.mechanism = SAMPLING_MECHANISM_MANUAL
+      if (!context._sampling.mechanism) {
+        context._sampling.mechanism = SAMPLING_MECHANISM_MANUAL
+      }
```
uurien, line 75:
```suggestion
      context._sampling.mechanism ??= SAMPLING_MECHANISM_MANUAL
```

packages/dd-trace/src/priority_sampler.js
```diff
@@ -206,4 +208,4 @@ function hasOwn (object, prop) {
   return Object.prototype.hasOwnProperty.call(object, prop)
 }
 
-module.exports = PrioritySampler
+module.exports = { PrioritySampler, hasOwn }
```
uurien, line 211: I think that `hasOwn` is not related with `priority_sampler.js` file, it is just a utils method to call to `hasOwnProperty`, you could use `hasOwnProperty` where you need it, or extract `hasOwn` function to utils file, but I think that it is not enough related with this file to export it here.

packages/dd-trace/src/opentracing/propagation/text_map.js
```diff
@@ -54,6 +58,10 @@ class TextMapPropagator {
     this._injectB3SingleHeader(spanContext, carrier)
     this._injectTraceparent(spanContext, carrier)
 
+    if (injectCh.hasSubscribers) {
```
uurien, line 61: Why are we applying changes in `TextMapPropagator` and not in `HttpPropagator`? Shouldn't be enough with `HttpPropagator`?

packages/dd-trace/src/exporters/agent/index.js
```diff
@@ -7,15 +7,15 @@ const Writer = require('./writer')
...
+    const { url, hostname, port, lookup, protocolVersion, stats = {}, appsec } = config
     this._url = url || new URL(format({
       protocol: 'http:',
       hostname: hostname || 'localhost',
       port
     }))
 
     const headers = {}
-    if (stats.enabled) {
+    if (stats.enabled || appsec?.standalone?.enabled) {
```
uurien, line 18: In `packages/dd-trace/src/span_stats.js`, to enable stats, you are checking `enabled && !appsec?.standalone?.enabled` but here, you are doing different check, `||` instead of `&& ! `.

Are both ok? It seems like both should be the same check.

packages/dd-trace/src/appsec/standalone.js
```diff
@@ -17,27 +65,72 @@ function onSpanStart ({ span, fields }) {
...
+    spanContext._sampling.priority = USER_KEEP
+  }
+}
+
+function sample (span) {
+  const spanContext = span.context?.()
+  if (enabled && spanContext._trace?.tags) {
+    spanContext._trace.tags[APPSEC_PROPAGATION_KEY] = '1'
+
+    // TODO: ask. can we reset here sampling like this?
```
uurien, line 98: pending todo here, maybe to ask before opening the PR?

packages/dd-trace/src/appsec/standalone.js
```diff
@@ -17,27 +65,72 @@ function onSpanStart ({ span, fields }) {
...
   if (enabled) {
     startCh.subscribe(onSpanStart)
+    injectCh.subscribe(onSpanInject)
+    extractCh.subscribe(onSpanExtract)
+
+    prioritySampler = new StandAloneAsmPrioritySampler(config.env)
   } else {
     startCh.unsubscribe(onSpanStart)
+    injectCh.unsubscribe(onSpanInject)
+    extractCh.unsubscribe(onSpanExtract)
```
uurien, line 126:
```suggestion
    if (startCh.hasSubscribers) startCh.unsubscribe(onSpanInject)
    if (injectCh.hasSubscribers) injectCh.unsubscribe(onSpanInject)
    if (extractCh.hasSubscribers) extractCh.unsubscribe(onSpanExtract)
```

packages/dd-trace/src/opentracing/propagation/text_map.js
iunanua: They are the same:

`class HttpPropagator extends TextMapPropagator {}`

packages/dd-trace/src/exporters/agent/index.js
iunanua: stats must be disabled if `appsec?.standalone?.enabled`

and `'Datadog-Client-Computed-Stats': 'yes'` must be sent if `appsec?.standalone?.enabled`

packages/dd-trace/src/priority_sampler.js
iunanua: I had doubts about it too, was tempted to implement `_hasOwn` in `PrioritySampler` to avoid the export but I could move it to `utils.js` file...

packages/dd-trace/src/opentracing/propagation/text_map.js
uurien: `HttpPropagator` extends `TextMapPropagator`,  so they are not the same. What I want to say is that in the tracer there is defined `HttpPropagator` as `HTTP_HEADERS` and `TextMapPropagator` as `TEXT_MAP`:

```javascript
    this._propagators = {
      [formats.TEXT_MAP]: new TextMapPropagator(config),
      [formats.HTTP_HEADERS]: new HttpPropagator(config),
      [formats.BINARY]: new BinaryPropagator(config),
      [formats.LOG]: new LogPropagator(config)
    }
```

Do we want to apply to both, or only to `HTTP_HEADERS`?

packages/dd-trace/src/appsec/standalone.js
```diff
@@ -17,27 +65,72 @@ function onSpanStart ({ span, fields }) {
...
+      } else if (lKey === 'tracestate') {
+        const tracestate = TraceState.fromString(carrier[key])
+        tracestate.forVendor('dd', state => state.clear())
+        carrier[key] = tracestate.toString()
+      }
+    }
+  }
+}
+
+function onSpanExtract ({ spanContext, carrier }) {
```
CarlesDD, line 84: I cannot find a reference for `carrier` in this function. Can we remove it?

packages/dd-trace/src/exporters/agent/index.js
uurien: ok, understood. we want to send 'Datadog-Client-Computed-Stats': 'yes'` even when stats are disabled if we are in standalone mode. 
Understood 👍

packages/dd-trace/src/priority_sampler.js
uurien: it might makes sense to have it in `packages/dd-trace/src/util.js`

packages/dd-trace/src/opentracing/propagation/text_map.js
iunanua: both, there are plugins like aws sdk, google cloud pub sub, kafka, etc using `TEXT_MAP` to inject or extract trace data from attributes, metadata or headers.
I guess, the propagation in those cases should behave the same as in an http service

integration-tests/standalone-asm.spec.js
```diff
@@ -0,0 +1,302 @@
...
+          assert.isArray(payload)
+
+          const innerReq = payload.find(p => p[0].resource === 'GET /sdk')
+          assert.notStrictEqual(innerReq, undefined)
+          assertKeep(innerReq[0])
+        }, undefined, undefined, true)
+      })
+
+      // proc/propagation-with-event triggers an appsec ev and calls downstream proc2/down with no event
+      it('should keep if parent trace is (prio:2, _dd.p.appsec:1) but there is no ev in the local trace', async () => {
```
uurien, line 203: you are not doing a warmup of `proc2`, doesn't it means that in `proc2` all the requests are the first request and should be kept always?

integration-tests/standalone-asm.spec.js
```diff
@@ -0,0 +1,301 @@
...
+  FakeAgent,
+  spawnProc,
+  curlAndAssertMessage,
+  curl
+} = require('./helpers')
+
+describe('Standalone ASM', () => {
+  let sandbox, cwd, startupTestFile, agent, proc, env
+  before(async () => {
+    sandbox = await createSandbox(['express'], true)
```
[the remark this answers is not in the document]
iunanua: removed

integration-tests/standalone-asm.spec.js
```diff
@@ -0,0 +1,301 @@
...
+    cwd = sandbox.folder
+    startupTestFile = path.join(cwd, 'standalone-asm/index.js')
+  })
+
+  after(async () => {
+    await sandbox.remove()
+  })
+
+  context('enabled', () => {
```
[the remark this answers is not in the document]
iunanua: copied from another test.

integration-tests/standalone-asm/index.js
```diff
@@ -0,0 +1,111 @@
...
+
+if (process.env.AGENT_URL) {
+  options.url = process.env.AGENT_URL
+}
+
+const standalone = process.env.DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED
+if (standalone) {
+  options.experimental.appsec.standalone.enabled = standalone === 'true' || standalone === '1'
+}
+
```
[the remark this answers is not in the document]
iunanua: it seems, if they are not passed as options the tracer is unable to get the correct value.
Could it be related to the way in which the test is spawning the process?

packages/dd-trace/test/appsec/standalone.spec.js
```diff
@@ -106,4 +154,245 @@ describe('Appsec Standalone', () => {
...
+    })
+
+    describe('_getPriorityFromTags', () => {
+      it('should keep the trace if manual.keep and _dd.p.appsec are present', () => {
+        context._trace.tags[APPSEC_PROPAGATION_KEY] = 1
+        assert.strictEqual(prioritySampler._getPriorityFromTags(tags, context), USER_KEEP)
+        assert.strictEqual(context._sampling.mechanism, SAMPLING_MECHANISM_APPSEC)
+      })
+
+      it('should return undefined if manual.keep or _dd.p.appsec are not present', () => {
```
uurien, line 358: Why is this test working? in `tags` manual.keep is defined: `{ 'manual.keep': 'true' }`

packages/dd-trace/test/exporters/agent/exporter.spec.js
```diff
@@ -43,6 +43,17 @@ describe('Exporter', () => {
     })
   })
 
+  it('should pass computed stats header through to writer if standalone configured', () => {
+    const stats = { enabled: false }
+    const appsec = { standalone: { enabled: true } }
+    exporter = new Exporter({ url, flushInterval, stats, appsec }, prioritySampler)
+    expect(Writer).to.have.been.calledWithMatch({
```
uurien, line 50:
```suggestion

    expect(Writer).to.have.been.calledWithMatch({
```

packages/dd-trace/test/opentracing/tracer.spec.js
```diff
@@ -103,6 +103,15 @@ describe('Tracer', () => {
     expect(SpanProcessor).to.have.been.calledWith(agentExporter, prioritySampler, config)
   })
 
+  it('should allow to configure an alternative prioritySampler', () => {
+    const sampler = {}
+    tracer = new Tracer(config, sampler)
+
+    expect(AgentExporter).to.have.been.called
```
uurien, line 110: This check looks redundant

packages/dd-trace/test/span_stats.spec.js
```diff
@@ -255,6 +255,13 @@ describe('SpanStatsProcessor', () => {
     expect(processor.tags).to.deep.equal(config.tags)
   })
 
+  it('should construct a disabled instance if appsec standalone is enabled', () => {
+    const standaloneConfig = { appsec: { standalone: { enabled: true } }, ...config }
+    const processor = new SpanStatsProcessor(standaloneConfig)
+    expect(processor.enabled).to.be.false
```
uurien, line 261:
```suggestion

    expect(processor.enabled).to.be.false
```

packages/dd-trace/test/appsec/standalone.spec.js
iunanua: but not `_dd.p.appsec`, both tags must be present in order to get `USER_KEEP` priority

integration-tests/standalone-asm.spec.js
iunanua: oops, fixed. now there is a warmup for both processes.

integration-tests/standalone-asm/index.js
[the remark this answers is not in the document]
iunanua: arg, it was my fault

packages/dd-trace/test/opentracing/tracer.spec.js
iunanua: removed

packages/dd-trace/src/proxy.js
```diff
@@ -179,9 +179,9 @@ class Tracer extends NoopProxy {
         this._modules.appsec.enable(config)
       }
       if (!this._tracingInitialized) {
-        this._tracer = new DatadogTracer(config)
+        const prioritySampler = appsecStandalone.configure(config)
+        this._tracer = new DatadogTracer(config, prioritySampler)
```
simon-id, line 183: wdyt about this ? it removes the need for changes in the constructors and it shows the intention better i feel like, we're injecting our own sampler only if it exists.
```suggestion
        this._tracer = new DatadogTracer(config)
        const prioritySampler = appsecStandalone.configure(config)
        if (prioritySampler) {
            this._tracer._prioritySampler = priortySampler
        }
        
```

iunanua: `DatadogTracer` is creating an instance of an `Exporter` and `SpanProcessor` [here](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/opentracing/tracer.js#L32) which in turn depend on the prioritySampler instance.

We should update those references too when doing `this._tracer._prioritySampler = prioritySampler`

simon-id: Ah, alright i see, then it's not as nice, let's keep it like that

integration-tests/standalone-asm/index.js
```diff
@@ -0,0 +1,101 @@
...
+  const span = tracer.scope().active()
+  span.context()._trace.tags['_dd.p.other'] = '1'
+
+  const resFetch = await fetch(url)
+  await resFetch.text()
+
+  res.status(200).send('propagation-without-event')
+})
+
+app.get('/down', async (req, res) => {
```
simon-id, line 80:
```suggestion
app.get('/down', (req, res) => {
```

integration-tests/standalone-asm/index.js
```diff
@@ -0,0 +1,101 @@
...
+  tracer.appsec.trackCustomEvent('custom-event')
+  res.status(200).send('sdk')
+})
+
+app.get('/vulnerableHash', (req, res) => {
+  const result = crypto.createHash('sha1').update(valueToHash).digest('hex')
+  res.status(200).send(result)
+})
+
+app.get('/propagation-with-event', async (req, res) => {
```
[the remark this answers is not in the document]
iunanua: I think it does...

integration-tests/standalone-asm.spec.js
```diff
@@ -0,0 +1,305 @@
...
+          assert.propertyVal(headers, 'datadog-client-computed-stats', 'yes')
+          assert.isArray(payload)
+
+          const innerReq = payload.find(p => p[0].resource === 'GET /down')
+          assert.notStrictEqual(innerReq, undefined)
+          assertKeep(innerReq[0], false)
+        }, undefined, undefined, true)
+      })
+
+      it('should remove parent trace data if there is no ev in the local trace', async () => {
```
simon-id, line 219: what's `ev` ?

simon-id: event ? not sure it's super clear

packages/dd-trace/test/exporters/agent/exporter.spec.js
```diff
@@ -43,6 +43,18 @@ describe('Exporter', () => {
     })
   })
 
+  it('should pass computed stats header through to writer if standalone configured', () => {
```
simon-id, line 46:
```suggestion
  it('should pass computed stats header through to writer if standalone appsec is enabled', () => {
```

integration-tests/standalone-asm/index.js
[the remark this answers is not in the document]
uurien: not support for it, it just ignores the result of the function, so the promise response is ignored. But if the only goal is use `async` to prevent using `.them()`..., I think that it is ok. (promise rejection is not going to be managed as Error throw, but for these tests... i think that it is ok)

packages/dd-trace/src/appsec/standalone.js
simon-id: So can we ?

packages/dd-trace/src/appsec/standalone.js
```diff
@@ -17,27 +62,71 @@ function onSpanStart ({ span, fields }) {
...
+  }
+}
+
+function sample (span) {
+  const spanContext = span.context?.()
+  if (enabled && spanContext?._trace?.tags) {
+    spanContext._trace.tags[APPSEC_PROPAGATION_KEY] = '1'
+
+    // TODO: ask. can we reset here sampling like this?
+    if (spanContext._sampling?.priority < AUTO_KEEP) {
```
simon-id, line 100: bit scared of the `<` here, you're sure there are no edge case ? with negative numbers or something

packages/dd-trace/src/appsec/standalone.js
```diff
@@ -1,12 +1,57 @@
...
+const { MANUAL_KEEP } = require('../../../../ext/tags')
+const PrioritySampler = require('../priority_sampler')
+const RateLimiter = require('../rate_limiter')
+const TraceState = require('../opentracing/propagation/tracestate')
+const { hasOwn } = require('../util')
+
+const {
+  APM_TRACING_ENABLED_KEY,
+  APPSEC_PROPAGATION_KEY
+} = require('../constants')
```
simon-id, line 14: not sure why those constants are multiline or separate from the rest of the requires ? since we also have other constants like priority and tags ?

packages/dd-trace/src/appsec/standalone.js
iunanua: these are the values
```javascript
declare const priority: {
  USER_REJECT: -1
  AUTO_REJECT: 0
  AUTO_KEEP: 1
  USER_KEEP: 2
}
```

packages/dd-trace/test/appsec/standalone.spec.js
```diff
@@ -106,4 +196,284 @@ describe('Appsec Standalone', () => {
...
     })
   })
+
+  describe('onSpanExtract', () => {
+    it('should not fail with invalid spanContext, trace, etc', () => {
+      standalone.configure(config)
+
+      assert.doesNotThrow(() => {
+        extractCh.publish({ })
+      })
```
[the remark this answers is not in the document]
iunanua: try it

```javascript
const { channel } = require('dc-polyfill')
const extractCh = channel('dd-trace:span:extract')

describe('dc channel', () => {
  it('should not throw err', () => {
    extractCh.subscribe(() => {
      throw new Error('crash')
    })

    assert.doesNotThrow(() => {
      extractCh.publish({})
    })
  })
})
```

packages/dd-trace/src/appsec/standalone.js
simon-id: i'm also a bit scared of potential future values of that enum ? ultimately, up to you, just pointing out the potential problems here

packages/dd-trace/test/appsec/standalone.spec.js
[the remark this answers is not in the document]
iunanua: ummm, what is going on here?
image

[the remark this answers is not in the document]
iunanua: is the assertion fulfilled but the test ends with an exception?

packages/dd-trace/test/appsec/standalone.spec.js
```diff
@@ -106,4 +196,284 @@ describe('Appsec Standalone', () => {
...
+    })
+
+    describe('_getPriorityFromAuto', () => {
+      it('should keep one trace per 1 min', () => {
+        const span = {
+          _trace: {}
+        }
+
+        const clock = sinon.useFakeTimers(new Date())
```
simon-id, line 451: doesn't that work ?
```suggestion
        const clock = sinon.useFakeTimers()
```

packages/dd-trace/test/appsec/standalone.spec.js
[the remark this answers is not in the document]
uurien: uncaught error !== "doesNotThrow". Test is failing because is crashing.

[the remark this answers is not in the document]
iunanua: yes, that made me think that `publish` wasn't catching the error

simon-id (approved):

Qard (approved):

PR #3608, opened 2023-09-01
Shell execution integration

hoolioh:
### What does this PR do?
This PR adds support for sending possible cmd execution threats. It is achieved by creating a new plugin which derives from TracingPlugin which, in turn, creates a new span every time a new process is spawn using `child_process` with all the cmdi information.

### Additional information
APM files modified:

- integration-tests/test-api-manual.spec.js
- packages/datadog-instrumentations/src/child_process.js
- packages/datadog-instrumentations/src/helpers/hooks.js
- packages/datadog-instrumentations/test/child_process.spec.js
- packages/datadog-plugin-child_process/src/index.js
- packages/datadog-plugin-child_process/src/scrub-cmd-params.js
- packages/datadog-plugin-child_process/test/index.spec.js
- packages/datadog-plugin-child_process/test/scrub-cmd-params.spec.js
- packages/dd-trace/src/plugins/index.js
- packages/dd-trace/src/plugins/util/exec.js

CI Visibility files modified:

- ci/init.js

[APPSEC-6584]

[APPSEC-6584]: https://datadoghq.atlassian.net/browse/APPSEC-6584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4417, opened 2024-06-20
fix: child_process: default to tracer service name

tlhunter:
### What does this PR do?
- defaults the service name of child process calls to the tracer's name

### Motivation
- see #3608
- a bunch of users are seeing `unnamed-service` entries for child processes


crysmags (approved):
LGTM

PR #4427, opened 2024-06-22
move parametric tests to a 16 cores runner

rochdev:
### What does this PR do?

Move parametric tests to a 16 cores runner.

### Motivation

These tests are slow, but they run in parallel so adding cores should mitigate their poor performance.

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -26,7 +26,7 @@ jobs:
 
 
   system-tests:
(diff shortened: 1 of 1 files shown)


rochdev:
For some reason, parallelizing on multiple cores doesn't actually improve the wall time needed to execute the tests. I ended up adding the ability to split on separate machines instead in #4436.

cc @cbeauchesne to look into why parallelizing parametric tests doesn't make things faster.

PR #4428, opened 2024-06-22
remove most get-port usage in appsec tests

rochdev:
### What does this PR do?

Remove most get-port usage in AppSec tests.

### Motivation

Follow-up of #4426

### Additional Notes

The only usage left is in Next plugin test because it sends the port to an external process. If that test could be refactored to avoid doing this, then AppSec tests could be completely free of get-port.

### DIFF
--- a/packages/dd-trace/test/appsec/graphql.apollo-server-express.plugin.spec.js
+++ b/packages/dd-trace/test/appsec/graphql.apollo-server-express.plugin.spec.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const getPort = require('get-port')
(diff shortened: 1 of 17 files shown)


uurien (approved):
> The only usage left is in Next plugin test because it sends the port to an external process

I'll take a look at this today

---
_edit:_
in a quick check, It seems that it is not possible for standalone nextjs tests, in the compiled files I can see this:
```javascript
const currentPort = parseInt(process.env.PORT, 10) || 3000
```
If we don't define the port externally, the app will use 3000, if it is not free, it will fail.

PR #4429, opened 2024-06-22
remove most usage of get-port for plugin tests

rochdev:
### What does this PR do?

Remove most get-port usage in plugin tests.

### Motivation

Follow-up of #4426


tlhunter (approved):
Why did we ever use get-port?

rochdev:
> Why did we ever use get-port?

In some cases it was (and still is) needed, and also originally it was used to get a port before initializing the tracer. It ended up being way overused though, which is problematic as it can result in flaky tests.

Qard (approved):

PR #4366, opened 2024-05-31
feat(openai): support token metrics for completions and chat completions when not provided

sabrenner:
### What does this PR do?
Adds support for tagging token metrics and submitting them as metrics when they are not provided. This only happens for `completions` and `chat.completions` in the case of streamed responses. 

This is a bit of a gray area, as although the Node.js OpenAI SDK does not provide these usage counts at all for streamed responses, the OpenAI API itself does, through the following option:

```
stream_options: {
  include_usage: true
}
```

Which, in the Node.js SDK, looks like:

```node
const stream = await openai.chat.completions.create({
  model,
  messages:,
  stream: true,
  stream_options: {
    include_usage: true,
  },
});
```

While `stream_options` is not yet a supported option in the Node.js SDK, it is passed through as a part of the request body to the OpenAI API, and is reflected in the returned response for streamed cases, where the last chunk includes the usage.

However, in the case users do not want to specify this option all throughout their code, or are uneasy about using an unsupported option for the SDK, this change also introduces computing these token metrics client-side. This is done either by:

1. Leveraging `tiktoken` if it is installed, a popular library for token estimation for a variety if models. This is done by trying to require it in a `try/catch`, something I don't think is common for the Node.js tracer, but is done in [the Python tracer](https://github.com/DataDog/dd-trace-py/blob/46a39848236de566132a29d043468eb84a7bf570/ddtrace/contrib/openai/utils.py#L13-L18), and is implemented here as well for consistency in experience.
2. In the case `tiktoken` isn't installed, or if there's an issue with using `tiktoken`, some basic estimations are used. This, again, follows the [standard set in the Python tracer](https://github.com/DataDog/dd-trace-py/blob/46a39848236de566132a29d043468eb84a7bf570/ddtrace/contrib/openai/utils.py#L154-L173), and does not seem to have a noticeable performance impact (at least from tests, but will try to add some benchmarks).

### Motivation
Expand support for OpenAI integration.

### Plugin Checklist

- [x] Unit tests.
- [ ] TypeScript [definitions][1].
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CircleCI [jobs/workflows][4].
- [ ] Plugin is [exported][5].

[1]: https://github.com/DataDog/dd-trace-js/blob/master/index.d.ts
[2]: https://github.com/DataDog/dd-trace-js/blob/master/docs/test.ts
[...]

PR #4433, opened 2024-06-24
fix(openai): safe require `tiktoken` for webpack bundlers

### LINKED ISSUE #4424: Module not found: Can't resolve 'tiktoken'
I just upgraded to v5.17.0 from v5.10.0

since i'm getting this error logs in my console : 
```
⚠ ../../node_modules/dd-trace/packages/datadog-plugin-openai/src/index.js
Module not found: Can't resolve 'tiktoken' in 'node_modules/dd-trace/packages/datadog-plugin-openai/src'
```
I don't use any OpenAPI  or LLM stuff,

Next.js application deployed on GKE.
Node version : v20.12.1
Next.JS version : 14.2.2

Can you help on this please ?

sabrenner:
### What does this PR do?
Add-on to #4366, where we added support for leveraging `tiktoken` if it is installed to capture token metrics for the `openai` integration. However, this does not play nicely with `webpack`, which popular framework Next.js uses under the hood. The workaround involves setting `resolve.fallback.tiktoken = false` in `webpack.config.js` or `next.config.js`. However, to make using the tracer as seamless as possible, this fix wraps the require in a function to bypass this behavior (by not directly requiring it).

Additionally, cleans up some test logic and utilizes `tiktoken` as a dev dependency for regression purposes.

### Motivation
Fixes #4424 

The following webpack error would come up when bundling:
```
⚠ ../../node_modules/dd-trace/packages/datadog-plugin-openai/src/index.js
Module not found: Can't resolve 'tiktoken' in 'node_modules/dd-trace/packages/datadog-plugin-openai/src'
```

#### Unit
Added `tiktoken` as a dev dependency for OpenAI tests. All tests now run against this, and any changes to that library or how we use it (ie typos) are reflected in failing tests. To test the in-house estimations, I moved the `estimateTokens` function into its own file, and added a small set of regression tests to assert basic behavior. This can be cleaned up more in a refactor of the OpenAI integration & tests in a future PR.

#### Local
Ran webpack on a test script importing `dd-trace`, before and after this fix. I was getting the error before, and not after. I was wondering why our Next.js tests weren't picking up on this, but I believe it's because of how we set up those tests, and how we use a `server.js` file, which I'm not sure gets compiled with webpack.


tlhunter (changes requested):
Sam and I spoke OOB. Will need a test as the `encodingForModel` typo alluded.

packages/datadog-plugin-openai/test/streamed-responses/completions.simple.txt
sabrenner, line 1: had to change the model here to be compatible with `tiktoken`

packages/datadog-plugin-openai/test/index.spec.js
```diff
@@ -3088,13 +3088,11 @@ describe('Plugin', () => {
                 expect(span.meta).to.have.property('openai.response.choices.0.message.content',
                   'Hello! How can I assist you today?')
 
-                // token metrics - these should be estimated counts
                 expect(span.metrics).to.have.property('openai.response.usage.prompt_tokens')
-                expect(span.metrics).to.have.property('openai.response.usage.prompt_tokens_estimated', 1)
+                expect(span.metrics).to.not.have.property('openai.response.usage.prompt_tokens_estimated')
```
sabrenner, line 3092: this kind of change ensures we were able to use `tiktoken` correctly

tlhunter (approved):

PR #4434, opened 2024-06-24
tracing(aws): add configuration option to inject all messages with trace context during SQS, SNS, and Kinesis batch send operations

wconti27:
### What does this PR do?
Adds an `aws-sdk` plugin config option to inject all AWS messages during batch send operations with trace context (SQS, SNS, Kinesis).
The config option can be enabled for all three services using the following:
`tracer.use('aws-sdk', { kinesis: { batchPropagationEnabled: true }, sns: { batchPropagationEnabled: true }, sqs: { batchPropagationEnabled: true } })`

### Motivation
Customer request

### Plugin Checklist

- [x] Unit tests.
- [x] TypeScript [tests][2].
- [x] API [documentation][3].

### DIFF
--- a/packages/datadog-plugin-aws-sdk/src/services/kinesis.js
+++ b/packages/datadog-plugin-aws-sdk/src/services/kinesis.js
@@ -52,7 +52,7 @@ class Kinesis extends BaseAwsSdkPlugin {
 
         // extract DSM context after as we might not have a parent-child but may have a DSM context
         this.responseExtractDSMContext(
-          request.operation, response, span || null, streamName
+          request.operation, response, span || null, { streamName }
         )
       }
     })
@@ -100,7 +100,8 @@ class Kinesis extends BaseAwsSdkPlugin {
     }
   }
 
-  responseExtractDSMContext (operation, response, span, streamName) {
+  responseExtractDSMContext (operation, params, response, span, kwargs = {}) {
+    const { streamName } = kwargs
     if (!this.config.dsmEnabled) return
     if (operation !== 'getRecords') return
     if (!response || !response.Records || !response.Records[0]) return
--- a/packages/datadog-plugin-aws-sdk/src/services/kinesis.js
+++ b/packages/datadog-plugin-aws-sdk/src/services/kinesis.js
@@ -52,7 +52,7 @@ class Kinesis extends BaseAwsSdkPlugin {
 
         // extract DSM context after as we might not have a parent-child but may have a DSM context
         this.responseExtractDSMContext(
-          request.operation, response, span || null, { streamName }
+          request.operation, request.params, response, span || null, { streamName }
         )
       }
     })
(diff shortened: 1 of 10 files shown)


tabgok (dismissed):
Approving from a functionality standpoint

Qard (dismissed):

packages/datadog-plugin-aws-sdk/src/services/kinesis.js
```diff
@@ -152,7 +152,12 @@ class Kinesis extends BaseAwsSdkPlugin {
       case 'putRecords':
         stream = params.StreamArn ? params.StreamArn : (params.StreamName ? params.StreamName : '')
         for (let i = 0; i < params.Records.length; i++) {
-          this.injectToMessage(span, params.Records[i], stream, i === 0)
+          this.injectToMessage(
+            span,
+            params.Records[i],
+            stream,
+            i === 0 || (this.config.kinesis && this.config.kinesis.batchPropagationEnabled)
```
tlhunter, line 159: Can you hoist this just before the for loop?

```js
const enableKinesisBatch = this.config.kinesis && this.config.kinesis.batchPropagationEnabled
// ...
i === 0 || enableKinesisBatch
```

packages/datadog-plugin-aws-sdk/src/services/sns.js
```diff
@@ -59,7 +59,12 @@ class Sns extends BaseAwsSdkPlugin {
         break
       case 'publishBatch':
         for (let i = 0; i < params.PublishBatchRequestEntries.length; i++) {
-          this.injectToMessage(span, params.PublishBatchRequestEntries[i], params.TopicArn, i === 0)
+          this.injectToMessage(
+            span,
+            params.PublishBatchRequestEntries[i],
+            params.TopicArn,
+            i === 0 || (this.config.sns && this.config.sns.batchPropagationEnabled)
```
tlhunter, line 66: Same hoisting here

tlhunter:
What are the drawbacks for having this always enabled? If there are none, shouldn't we always have it enabled? If there is a drawback, shouldn't the API docs for the configuration call this out?

Are batches only used when user code manually calls a batch method? Or does the AWS library decide to batch some things?
I'm curious if two requests to a Node server come in at the same time, one with request ID 1, one with request ID 2, and both making an outbound AWS request. Does the AWS SDK internally every combine those into a batch? And if so, would the request IDs be properly associated?

packages/datadog-plugin-aws-sdk/src/services/kinesis.js
wconti27: done

wconti27:
> What are the drawbacks for having this always enabled? If there are none, shouldn't we always have it enabled? If there is a drawback, shouldn't the API docs for the configuration call this out?
> 
> Are batches only used when user code manually calls a batch method? Or does the AWS library decide to batch some things? I'm curious if two requests to a Node server come in at the same time, one with request ID 1, one with request ID 2, and both making an outbound AWS request. Does the AWS SDK internally every combine those into a batch? And if so, would the request IDs be properly associated?

The drawbacks are mainly performance related. APM already adds a good amount of overhead, and injecting every message (batches are limited to 10 messages) per API call, especially if they are doing multiple/many batch API calls could push overhead higher than some customers want. There's also size limitations/concerns, and generally messaging services like SQS will charge per set amount of data in a message (64 kb = 1 "SQS request"), so increasing message size will increase customer cost. 

After typing all that out, I think it's best we keep this configurable.

For your second question, batch operations are specifically called by the user. The library does not try to batch requests together. So two separate `SQS.sendMessage` operations sending 1 message each will result in 2 API calls.

Qard (approved):

penchef:
🙌

PR #4435, opened 2024-06-24
deps: upgrade lru-cache@^7.14.0 -> ^10.2.2

tlhunter:
### What does this PR do?
- upgrades lru-cache from ^7.14.0 to latest ^10.2.2

### Motivation
- newer is cooler


rochdev:
@tlhunter This version doesn't support Node 16, so we should either add a label to not merge this on 4.x, or revert this PR to keep the version used consistent.

PR #4436, opened 2024-06-25
split parametric tests on multiple machines

rochdev:
### What does this PR do?

Split parametric tests on multiple machines.

### Motivation

Using xdist to split across cores doesn't improve performance all that much.

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -84,6 +84,8 @@ jobs:
           path: artifact.tar.gz
 
   parametric:
+    needs:
+      - build-artifacts
     uses: DataDog/system-tests/.github/workflows/run-parametric.yml@main
     secrets: inherit
     with:
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -84,7 +84,7 @@ jobs:
           path: artifact.tar.gz
 
   parametric:
-    uses: DataDog/system-tests/.github/workflows/run-parametric.yml
+    uses: DataDog/system-tests/.github/workflows/run-parametric.yml@main
     secrets: inherit
     with:
       library: nodejs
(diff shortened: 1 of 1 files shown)


.github/workflows/system-tests.yml
```diff
@@ -84,11 +84,10 @@ jobs:
           path: artifact.tar.gz
 
   parametric:
-    needs:
-      - build-artifacts
```
simon-id, line 88: why doesn't it need build anymore ?

.github/workflows/system-tests.yml
```diff
@@ -84,11 +84,10 @@ jobs:
...
-    uses:  DataDog/system-tests/.github/workflows/system-tests.yml@main
+    uses: DataDog/system-tests/.github/workflows/run-parametric.yml@main
     secrets: inherit
     with:
-      scenarios: PARAMETRIC
```
simon-id, line 92: why not this var anymore ?

.github/workflows/system-tests.yml
```diff
@@ -84,11 +84,10 @@ jobs:
...
       library: nodejs
       binaries_artifact: system_tests_binaries
+      _experimental_job_count: 8
+      _experimental_job_matrix: '[1,2,3,4,5,6,7,8]'
```
simon-id, line 93: are those pytest arguments ? or where are they from ? where are htey documented ?

.github/workflows/system-tests.yml
rochdev: It probably does, this was removed by mistake.

.github/workflows/system-tests.yml
rochdev: They are converted to pytest arguments. Not documented anywhere, it's part of system-tests repo.

simon-id (approved):

PR #4440, opened 2024-06-25
Disable crashing pugins

bengl:
### What does this PR do?

When one of the plugins subscriptions fails, catch it and disable the whole plugin.

### Motivation

If we run into an unforeseen incompatibility on the Diagnostics Channel subscriber side, we don't want to crash user apps.

### DIFF
--- a/packages/dd-trace/test/plugins/plugin.spec.js
+++ b/packages/dd-trace/test/plugins/plugin.spec.js
@@ -1,0 +1,69 @@
+'use strict'
+
+require('../setup/tap')
+
+const Plugin = require('../../src/plugins/plugin')
+const plugins = require('../../src/plugins')
+const { channel } = require('dc-polyfill')
+
+describe('Plugin', () => {
+  class BadPlugin extends Plugin {
+    static get id () { return 'badPlugin' }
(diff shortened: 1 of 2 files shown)


rochdev (dismissed):

Qard (dismissed):

ida613 (dismissed):

packages/dd-trace/test/plugins/plugin.spec.js
```diff
@@ -0,0 +1,69 @@
...
+      throw new Error('this is one bad plugin')
+    }
+  }
+
+  class GoodPlugin extends Plugin {
+    static get id () { return 'goodPlugin' }
+
+    constructor () {
+      super()
+      this.addSub('apm:badPlugin:start', this.start)
```
ida613, line 28: should this be `apm:goodPlugin:start`?

bengl: It should! Fixing now.

Qard (dismissed):
LGTM other than some minor nits from the linter.

Qard (approved):

PR #4442, opened 2024-06-25
decrease all green poll interval and timeout

rochdev:
### What does this PR do?

Decrease "All Green" poll interval to every 18 seconds and timeout to 6 minutes.

### Motivation

The tests were just made faster over the last few days, and there should be a check in place to make sure they don't become slow again.


BridgeAR:
Discussed with @rochdev. Closing

PR #4446, opened 2024-06-27
fix(lambda): gate timeout spans and add missing clear

duncanista:
### What does this PR do?
Gates AWS Lambda tracer instrumentation with the key `lambda`.
Adds a `clearTimeout` to non-promisified functions.

### Motivation
Support ticket where customers was being timed out because they were using promises and returning the function as non-promisified, making it to successfully work with AWS Lambda, but not with our tracer instrumentation.

### Plugin Checklist

- [x] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### Additional Notes

Disabled by `DD_TRACE_DISABLED_INSTRUMENTATIONS=lambda,other,another` by following the current workaround for instrumentations.


packages/dd-trace/src/lambda/index.js
```diff
@@ -2,4 +2,11 @@
 
 const { registerLambdaHook } = require('./runtime/ritm')
 
-registerLambdaHook()
+const _DD_TRACE_DISABLED_INSTRUMENTATIONS = process.env.DD_TRACE_DISABLED_INSTRUMENTATIONS || ''
```
duncanista, line 5: with underscores since these are already defined by another file as global, but during uni testing, this file is not run there.

tlhunter: Is it OK that this only applies to users who disable integrations via ENV var and not to users who disable integrations via the init config object?

tlhunter: Oh, do Lambda users never touch that object and instead configure everything via ENV vars?

tlhunter: If that's the case it's probably worth adding a comment.

duncanista: > Oh, do Lambda users never touch that object and instead configure everything via ENV vars?

Yeah, so this is enabled by default already, and there's no way to enable/disable this through the init config object or any other manner.

duncanista: Added a comment around it~

tlhunter (approved):

PR #3649, opened 2023-09-21
[ci-visibility] Add support for unskippable suites for jest

juan-fernandez:
### What does this PR do?

* Add `jest-docblock` dependency. The dependency is tiny: https://github.com/jestjs/jest/blob/main/packages/jest-docblock/package.json and allows to parse a test file for docblocks at the top. See size: https://arve0.github.io/npm-download-size/#jest-docblock
* Add the option to mark suites as unskippable. This is done via the following docblock at the top of the suite:
```
/**
 * @datadog {"unskippable": true}
 */
```

### Motivation

* Intelligent test runner does not always work (see [ITR limitations](https://docs.datadoghq.com/continuous_integration/intelligent_test_runner/#limitations)) and we want to let users avoid ITR skipping certain suites. 

Extra context: these comment blocks are used by `jest` in a similar way: https://jestjs.io/docs/configuration#testenvironmentoptions-object

### Plugin Checklist
- [x] Unit tests.

PR #4448, opened 2024-06-28
Move testing libraries to `devDependencies`

mleonhard-notion:
### What does this PR do?
This PR move testing libraries `jest-docblock` and `istanbul-lib-coverage` from `dependencies` to `devDependencies`.

### Motivation
These libraries are used only during testing and should not be included when using dd-trace in production.  We want to eliminate these test libraries from a NodeJS server prod build.  The dependencies were added to dd-trace's `dependencies` in https://github.com/DataDog/dd-trace-js/pull/2271 and https://github.com/DataDog/dd-trace-js/pull/3649.


simon-id (changes requested):
Hello,
I think this PR is mistaken as the CI-visibility product is actually using those two dependencies in production code. If you look at the PRs and code you linked, or if you search the repo for those dependencies, you can clearly see it.
- #2271 description says `Additionally, a new dependency istanbul-lib-coverage is added for parsing the code coverage`
- #3649 description says `Add jest-docblock dependency [...] and allows to parse a test file for docblocks at the top.`

As for the code:
- `jest-docblock`: https://github.com/DataDog/dd-trace-js/blob/5daf54de526f2c0477e1eaf27146fb620b88dd03/packages/datadog-plugin-jest/src/util.js#L2
- `istanbul-lib-coverage`:
https://github.com/DataDog/dd-trace-js/blob/5daf54de526f2c0477e1eaf27146fb620b88dd03/packages/datadog-instrumentations/src/cucumber.js#L2
https://github.com/DataDog/dd-trace-js/blob/5daf54de526f2c0477e1eaf27146fb620b88dd03/packages/datadog-instrumentations/src/mocha/main.js#L3
https://github.com/DataDog/dd-trace-js/blob/5daf54de526f2c0477e1eaf27146fb620b88dd03/packages/dd-trace/src/plugins/util/test.js#L6
@juan-fernandez just to double check

juan-fernandez:
hey @mleonhard-notion ! Thanks for the PR. @simon-id is right in his response. You can think of these dependencies as the dependencies that a test framework or a mocking library has in its package. For example:

* https://github.com/cucumber/cucumber-js/blob/4c67f49687e8039967c217e129454362e4c536ca/package.json#L227
* https://github.com/sinonjs/sinon/blob/da4230a00c929f56738d272da469a5ecb9d40da7/package.json#L80

Even though you're likely using `cucumber` or `sinon` only in testing (and you'd have them as a dev dependency), these libraries need to include their dependencies in their package. 

We're open to other suggestions to reduce the package size though 😄

mleonhard-notion:
@simon-id @juan-fernandez Thanks for responding quickly.  Would you please explain more?  Our Prod build depends on `dd-trace` which depends directly on `istanbul-lib-coverage` and `jest-docblock`.  Why does `dd-trace` need to parse test coverage data or docblocks?  It is prod code for reporting trace info to Datadog.  We don't want to bring in these extra dependencies.

Is it because the testing plugins are distributed in the same package as the prod code?

A related issue is that `dd-trace` depends on `istanbul-lib-coverage` version `3.2.0` instead of `^3.2.0`, missing bugfixes in 3.2.1 and 3.2.2.

juan-fernandez:
> Is it because the testing plugins are distributed in the same package as the prod code?

yeah, it's because the plugins for [Test Visibility](https://docs.datadoghq.com/tests/setup/javascript/?tab=cloudciprovideragentless) are in the same bundle. It could be argued that they could be separated, but that's how it is today. 

In any case, is there any specific worry about the packages being there? They're quite small:

* https://bundlephobia.com/package/istanbul-lib-coverage@3.2.0 1.9KB gzipped 
* https://bundlephobia.com/package/jest-docblock@29.7.0 1KB gzipped 

Other dependencies like `@opentelemetry/api`, `@opentelemetry/core` or `semver` are significantly bigger.

PR #4449, opened 2024-07-01
[DEBUG-2972] Add initial support for Code Origin for Spans (Fastify entry spans only)

watson:
### What does this PR do?

**Note: This PR relates to an upcoming Datadog feature that's not generally available yet.**

This PR does two things:
- It lays the groundwork for a new feature called "Code Origin for Spans" (previously called "Span Origin").
- To showcase this feature, it adds limited support for just Fastify entry-spans.

This feature will, in its most basic form, allow users to see where in their code base a span was started/created.

Code origin is only set for spans whos callstack contains user land stack frames.

For reference, here's a PR that implements Code Origin in the Java tracer: https://github.com/DataDog/dd-trace-java/pull/7001

### Motivation

This makes it easier to understand what each span shown in the flame graph represents.

### Performance considerations

Keeping high performance in mind, we'll only compute the code origin for certain pre-defined span types. Specifically what we call entry and exit spans.

- Entry spans are defined as either the root span in a trace, or the "local" root span within a given service (it has a parent span, but that belongs to a different service). 
- Exit spans are defined as spans "calling" out to other services, e.g. database spans, outgoing HTTP request spans etc.

### Scope

Adding support for all types of entry and exist spans in a single PR will make it harder to review, so **this PR will only concern itself with adding support for Fastify entry spans**; That is spans representing incoming HTTP requests to a Fastify app.

Middlewares have to be handled in a special way, as a middleware might respond early to an incoming HTTP request in which case the route handler is never called. In that instance, we want to the code origin to point to where the middleware function was registered. **Doing that for Fastify however, is out of scope of this PR** and support for this will be added in an upcoming PR.

### Note to reviewers

Since this feature has not yet been rolled out to staging, you need a hash-link to test it. If you need access, please contact me.

### Tasks

- [ ] Benchmarks?


watson:
* **#4449** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4449?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/dd-trace/src/plugins/util/stacktrace.js
```diff
@@ -0,0 +1,101 @@
...
+
+// TODO: This should be somewhere else specifically related to Code Origin for Spans
+function getCodeOriginForSpansTags (callsite) {
+  if (!callsite) return
+  const file = callsite.getFileName()
+  const line = String(callsite.getLineNumber())
+  const method = callsite.getFunctionName()
+  return method
+    ? {
+        '_dd.entry_location.file': file,
```
shatzi, line 93: this have changed to:
```
_dd.code_origin.type = "entry" | "exit"
_dd.code_origin.frames.0.file
_dd.code_origin.frames.0.line
_dd.code_origin.frames.0.type
_dd.code_origin.frames.0.method

_dd.code_origin.frames.1...
```

shatzi (commented):
very cool

packages/dd-trace/src/plugins/util/stacktrace.js
```diff
@@ -0,0 +1,101 @@
...
+        '_dd.entry_location.line': line,
+        '_dd.entry_location.method': method
```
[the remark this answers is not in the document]
watson: Good point. Actually we can do that. I'll update it

packages/dd-trace/src/plugins/util/stacktrace.js
watson: Thank you. Yes, this code is from before we changed it. I have a TODO to update it to the new style :)

packages/dd-trace/src/plugins/util/stacktrace.js
```diff
@@ -0,0 +1,101 @@
...
+    // because I have the source code in a parent folder called `node_modules`. So the code below thinks that
+    // it's not in user-land
+    const relativePath = relative(cwd, containsFileProtocol ? filename.substring(7) : filename)
+    if (relativePath.startsWith('node_modules' + sep) || relativePath.includes(sep + 'node_modules' + sep)) {
+      continue
+    }
+
+    return returnTopUserLandFrameOnly
+      ? callsite
+      : (i === 0 ? callsites : callsites.slice(i))
```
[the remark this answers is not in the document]
watson: The algorithm is as follows:

We start with an array of callsites. The top n callsites might not be user-land callsites, so we want to remove those from the array. In case `i === 0` it means that the top callsite is actually a user-land frame, so there's no need to slice the array (which will shallow-clone it). Additionally, we want to limit the number of frames to `limit`. If there's less callsites in the array then the max `limit` and `i === 0`, again, there's no need to call `slice`.

packages/dd-trace/src/plugins/util/stacktrace.js
watson: Updated 👍

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -27,6 +31,11 @@ function wrapFastify (fastify, hasParsingEvents) {
       app.addHook('preHandler', preValidation)
     }
 
+    // Hack to check if the Code Origin for Spans feature is enabled
+    if (codeOriginEnabledChannel.hasSubscribers) {
```
juan-fernandez, line 35: we try to avoid importing `dd-trace` code in instrumentations unless it's absolutely necessary. I'm referring to `const { entryTag } = require('../../dd-trace/src/code_origin')`. 

You might be able to achieve the same with something like
```suggestion
   app.addHook('onRoute', function (routeOptions) {
      if (codeOriginEnabledChannel.hasSubscribers) {
         codeOriginEnabledChannel.publish(routeOptions)
      }
   })
```

and change the code in `packages/datadog-plugin-fastify/src/index.js` to:

```javascript
this.addSub('datadog:code-origin-enabled', routeOptions => {
  if (this._tracerConfig.codeOriginForSpansEnabled) {
    if (!routeOptions.config) routeOptions.config = {}
    routeOptions.config[kCodeOriginForSpansTagsSym] = entryTag(onRoute)
  }
})
```

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -142,6 +152,10 @@ function getRes (reply) {
   return reply && (reply.raw || reply.res || reply)
 }
 
+function getCodeOriginForSpansTags (request) {
+  return request?.routeOptions?.config?.[kCodeOriginForSpansTagsSym]
```
juan-fernandez, line 156: if you do my other suggested changes, you can do this in the `apm:fastify:request:handle` handler, and have the `Symbol` defined in `packages/datadog-plugin-fastify/src/index.js`

juan-fernandez (commented):
LGTM. Only some minor comments

packages/datadog-instrumentations/src/fastify.js
watson: Fixed 👍

juan-fernandez (dismissed):
LGTM, but it might be worth waiting for some reviewer with more experience with the fastify plugin 😄

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -27,6 +28,11 @@ function wrapFastify (fastify, hasParsingEvents) {
       app.addHook('preHandler', preValidation)
     }
 
+    // No need to add the onRoute hook unless Code Origin for Spans is enabled
+    if (codeOriginForSpansChannel.hasSubscribers) {
```
rochdev, line 32: This should be in the hook instead because subscribers can be enabled/disabled at runtime.

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -5,6 +5,7 @@ const { addHook, channel, AsyncResource } = require('./helpers/instrument')
 
 const errorChannel = channel('apm:fastify:middleware:error')
 const handleChannel = channel('apm:fastify:request:handle')
+const codeOriginForSpansChannel = channel('datadog:code-origin-for-spans')
```
rochdev, line 8: There shouldn't be explicit references to tracing in instrumentation.

rochdev: Also, channels are easier to reason about if they match the event that they represent. For example, different subscribers may be interested in different aspects of routing, so it would be better to have something like a routeChannel instead which would add the route when any subscriber is interested in that.

packages/datadog-plugin-fastify/src/index.js
```diff
@@ -10,9 +13,18 @@ class FastifyPlugin extends RouterPlugin {
   constructor (...args) {
     super(...args)
 
-    this.addSub('apm:fastify:request:handle', ({ req }) => {
+    this.addSub('apm:fastify:request:handle', ({ req, routeConfig }) => {
       this.setFramework(req, 'fastify', this.config)
+      const tags = routeConfig?.[kCodeOriginForSpansTagsSym]
+      if (tags) this.setSpanTags(req, tags)
```
rochdev, line 19: Could this be done in a separate plugin instead? The reasoning is that adding product-specific code directly to plugin increases size and complexity even when those products are not used, whereas we could disable product-specific plugins entirely and not even import them when a product is disabled for example.

packages/datadog-plugin-fastify/test/index.spec.js
```diff
@@ -542,7 +542,193 @@ describe('Plugin', () => {
             })
           }
         })
+
+        describe('with tracer config codeOriginForSpansEnabled', () => {
```
rochdev, line 546: Is this name Node-specific or has it been standardized? Mostly asking because it exposes implementation details in the name.

packages/datadog-plugin-fastify/src/index.js
rochdev: That would actually also remove the need to modify `web` since you could then have something like `setOrigin` on your di-specific base class.

rochdev (changes requested):

packages/datadog-plugin-fastify/test/index.spec.js
watson: It's being implemented across the board. It's already implemented in the Java and Python tracer.

watson: The environment variable is standardized: `DD_CODE_ORIGIN_FOR_SPANS_ENABLED`

However the concept of initializing the tracer programmatically, and therefore the need to have equivalent names for the config options, I think is something that's unique to the Node.js tracer?

So no, the name `codeOriginForSpansEnabled` is not standardized.

> Mostly asking because it exposes implementation details in the name.

How is `codeOriginForSpansEnabled` exposing an implementation detail?

packages/datadog-instrumentations/src/fastify.js
watson: The hook is only fired at boot when the routes are registered (not sure if there's a use-case for registering routes later though 🤔). So I put it here to not be too noisy and add hooks that were not needed. But since it's not a hot code-path (as it normally only happens at boot), I can move it 👍 

In what scenarios will this channel have its subscriber added or removed at runtime? I know that it technically can, but in this case we control when it happens, so shouldn't it be ok?

watson: Actually, scratch that. I think you idea of making this a generic "route has been added" channel is really great and it removes the coupling with the specific feature, which I wasn't happy about to begin with. I'll update the code 👍

packages/datadog-plugin-fastify/src/index.js
watson: > Could this be done in a separate plugin instead?

Do you mean a generic "Code Origin for Spans" plugin that would be used for all the different frameworks? The implementation details for each framework would be slightly different as the way we hook into the router is slightly different. But I guess I can just add a lot of `this.addSub()` calls for the different frameworks we end up supporting in that one generic plugin 🤔

watson: > That would actually also remove the need to modify `web` since you could then have something like `setOrigin` on your di-specific base class.

I can of course move the logic for getting the current span based on the current `req` object to a file just related to the plugin, but I thought the whole reason for having the `web` file was to concentrate all the logic in one central place?

Say for example I do create a custom plugin called `datadog-plugin-code-origin`, I could do something like the following code snippet - is that what you had in mind?

```js
const web = require('../../dd-trace/src/plugins/util/web')

// ...

this.addSub('apm:fastify:request:handle', ({ req, routeConfig }) => {
  const tags = routeConfig?.[kCodeOriginForSpansTagsSym]
  if (!tags) return
  context.span?.addTags(tags)
})
```

watson: After trying to implement this with a custom plugin I discovered that the plugin system doesn't seem well suited for plugins that are not actually related to modules being instrumented.

I've just pushed a few commits addressing your comments and I've made a "hack" the current plugin system to make it work. But in the long rung, using plugins for stuff like this, seem to require us to refactor the whole plugin system if we want to do it right - unless of course that I've misunderstood something.

Let's have Zoom call and go over this in person when you have time.

packages/datadog-instrumentations/src/fastify.js
watson: I've pushed a commit with a fix. Let me know if that's what you had in mind

packages/datadog-instrumentations/src/fastify.js
watson: I've pushed a commit with a fix. Let me know if that's what you had in mind

packages/datadog-plugin-fastify/src/index.js
watson: Also, just realized that my approach of detecting if the new plugin should be enabled or not doesn't work if you try to enable it programmatically in the call to `tracer.init`  🙄(hence the failing tests now)

watson: There's something to be said for having the Fastify related code in `datadog-plugin-fastify` etc. That way it's only executed if Fastify is actually used in the application being instrumented.

That's a point against the idea of having a central `datadog-plugin-code-origin` plugin

watson: Based on our Zoom call yesterday, I've now pushed yet another commit that uses the composite plugin approach we discussed. Please let me know what you think.

juan-fernandez (dismissed):

tlhunter (approved):

packages/datadog-plugin-code-origin/src/fastify.js
```diff
@@ -0,0 +1,29 @@
+'use strict'
```
rochdev, line 1: Not a blocker but more of a thought, I'm not sure about this structure of having a `code-origin` plugin. It makes sense from an ownership perspective, but makes it a bit difficult to navigate the Fastify plugin since it jumps to other places. With a centralized plugin manager this wouldn't be an issue, but from an existing plugin it can be surprising. This is especially true since right now every plugin folder is for a specific module, so from a reader it seems like there is a `code-origin` module that we support.

If you prefer to keep the code structure that is currently in this PR, I'd probably use a `datadog-code-origin` folder instead to hold the base plugin and also any other helper related to code origin.

packages/datadog-plugin-code-origin/src/lib/tags.js
```diff
@@ -0,0 +1,38 @@
+'use strict'
+
+const { getUserLandFrames } = require('../../../dd-trace/src/plugins/util/stacktrace')
+
+const limit = Number(process.env._DD_CODE_ORIGIN_MAX_USER_FRAMES) || 8
```
rochdev, line 5: Why the underscore prefix?

packages/datadog-plugin-fastify/src/index.js
```diff
@@ -1,18 +1,16 @@
...
-    super(...args)
-
-    this.addSub('apm:fastify:request:handle', ({ req }) => {
-      this.setFramework(req, 'fastify', this.config)
-    })
+class FastifyPlugin extends CompositePlugin {
+  static get id () { return 'fastify' }
+  static get plugins () {
+    return {
+      instrumentation: FastifyInstrumentationPlugin,
```
rochdev, line 11: Instrumentation is what publishes to these plugins. That plugin is specific to tracing.

packages/datadog-plugin-fastify/test/index.spec.js
```diff
@@ -26,14 +28,12 @@ describe('Plugin', () => {
...
         describe('without configuration', () => {
-          if (NODE_MAJOR <= 18 && semver.satisfies(specificVersion, '>=5')) return
-
           before(() => {
             return agent.load(['fastify', 'find-my-way', 'http'], [{}, {}, { client: false }])
           })
 
           after(() => {
-            return agent.close({ ritmReset: false })
+            return agent.close({ ritmReset: false, wipe: true })
```
rochdev, line 36: This should no longer be needed now that the tracer URL can be reconfigured and plugins can be enabled/disabled.

rochdev (changes requested):

packages/datadog-plugin-code-origin/src/lib/tags.js
watson: It's from the RFC. Seems odd, I agree. Apparently it's been done before, though not something we've used in the Node.js tracer. It's to signify it's not to be used by end users - as in it's private and undocumented - use at your own risk. I'm working on changing it, but for now it's like this if I'm to follow the RFC.

packages/datadog-plugin-fastify/test/index.spec.js
watson: If I remove it, the tests fail because it's only running the plugin constructers before the first test. I need to re-run it because the tests change the tracer config

packages/datadog-plugin-code-origin/src/fastify.js
watson: So are you saying you prefer this to be in the `datadog-plugin-fastify` folder again and not make it a composite plugin after-all, or have I misunderstood you?

packages/datadog-plugin-fastify/test/index.spec.js
rochdev: It should be enough to reload the agent. Did you try doing `agent.load` in a `beforeEach` instead of `before`?

packages/datadog-plugin-code-origin/src/fastify.js
rochdev: For now at least until this is improved with proper support for a different structure.

packages/datadog-plugin-fastify/test/index.spec.js
watson: Yes, didn't fix it

watson: Oh, actually after refactoring the code as you suggested and letting the composite plugin take care of enabling/disabling the subscriptions based on the config option, it all works now without `wipe: true`. I'll remove it

packages/datadog-plugin-fastify/src/index.js
watson: Fixed 👍

packages/datadog-plugin-code-origin/src/fastify.js
watson: Fixed, based on our Zoom call 👍

packages/datadog-plugin-code-origin/src/index.js
```diff
@@ -0,0 +1,23 @@
...
+
+class CodeOriginForSpansPlugin extends Plugin {
+  static get id () {
+    return 'code-origin-for-spans'
+  }
+
+  constructor (...args) {
+    super(...args)
+
+    if (this._tracerConfig.codeOriginForSpansEnabled) {
```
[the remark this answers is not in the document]
watson: Wow, that's nice. Works like a charm 👍 I've pushed a new commit

packages/datadog-plugin-fastify/test/index.spec.js
```diff
@@ -542,7 +542,198 @@ describe('Plugin', () => {
             })
           }
         })
+
+        describe('with tracer config codeOriginForSpans.enabled: true', () => {
```
rochdev, line 546: Everything LGTM at this point, but I'm wondering if this test could also be split to its own file. If it's too complicated we can keep it here for now but if it's easy I'd say let's do it.

watson: Not a problem. I've pushed a new commit with the tests split into two files

rochdev (approved):

PR #4459, opened 2024-07-03
ci: exclude aerospike on arm64 archs when running 'yarn services'

watson:
The `aerospike` package didn't get arm support until v5.2.0.

If running `yarn services` on an arm64 based system, you'd get an error when trying to install older versions of `aerospike`. This PR fixes that by not installing `aerospike` on arm64 architectures.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,7 +95,7 @@ $ docker-compose up -d -V --remove-orphans --force-recreate
 $ yarn services
 ```
 
-**Note**
+> **Note**
 > The `aerospike`, `couchbase`, `grpc` and `oracledb` instrumentations rely on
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -95,10 +95,10 @@ $ docker-compose up -d -V --remove-orphans --force-recreate
 $ yarn services
 ```
 
-> **Note**
-> The `couchbase`, `grpc` and `oracledb` instrumentations rely on native modules
-> that do not compile on ARM64 devices (for example M1/M2 Mac) - their tests
-> cannot be run locally on these devices.
+**Note**
```
watson, line 98:
```suggestion
> **Note**
```

juan-fernandez (approved):

PR #4470, opened 2024-07-05
split cypress integration tests by module type

rochdev:
### What does this PR do?

Split Cypress integration tests by module type.

### Motivation

Cypress tests are slow.


juan-fernandez (approved):
thanks!

PR #4476, opened 2024-07-05
pull agent image instead of rebuilding

rochdev:
### What does this PR do?

Pull agent image instead of rebuilding.

### Motivation

The agent can only change in the system tests repo, so there is no reason to keep rebuilding it on every commit.

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -68,9 +68,11 @@ jobs:
         run: ./build.sh -i weblog
       - name: Build runner
         uses: ./.github/actions/install_runner
-      - name: Build agent
+      - name: Pull agent
         id: build-agent
-        run: ./build.sh -i agent
+        run: |
+          docker pull ghcr.io/datadog/system-tests/docker/agent:latest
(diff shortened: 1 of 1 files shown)


BridgeAR:
@rochdev is this something you would like to still look into? I do think it makes sense to not have to built the agent on our own, but I didn't check how long that currently takes either.

rochdev:
> is this something you would like to still look into? I do think it makes sense to not have to built the agent on our own, but I didn't check how long that currently takes either.

Yes this is still something I think we should do. I think this might even be possible in system tests themselves at this point because adding `ca-certificates` is no longer needed.

@cbeauchesne Thoughts on this?

cbeauchesne:
As we do not install anymore ca-certificates, this changes should make no difference, the current build step pull the base image (which you also do), and run a basic set of docker command that does not rely on any external call, so it's less than 1s.

Furthermore, `ghcr.io/datadog/system-tests/docker/agent:latest` is produced by the non-official docker mode, so we do not closely check it works fine, so I do not recommend using it.

PR #4483, opened 2024-07-08
release dev version only when a label is used

rochdev:
### What does this PR do?

Release dev version only when a label is used.

### Motivation

Otherwise, every time a PR is merged, a deployment is created which fails if no one approves it, which is in most cases. This makes it explicit so that a label must be added to a PR to release instead of having every PR fail 24h later because nobody approved the deployment before the timeout.

### DIFF
--- a/.github/workflows/release-dev.yml
+++ b/.github/workflows/release-dev.yml
@@ -1,12 +1,14 @@
 name: Release dev release line
 
 on:
-  push:
+  pull_request_target:
+    types: [labeled]
     branches:
       - master
 
 jobs:
   dev_release:
+    if: ${{ github.event.label.name == 'release-dev' }}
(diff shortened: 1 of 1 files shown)


Qard (approved):

Qard:
I'm guessing it's not possible, but do you know if it's possible for adding the label to auto-approve the publish? Otherwise this means there's manual involvement both before _and_ after the CI run. 🤔

rochdev:
> I'm guessing it's not possible, but do you know if it's possible for adding the label to auto-approve the publish? Otherwise this means there's manual involvement both before and after the CI run. 🤔

I don't think it is as the point of the deployment is to enforce a manual interaction. The good news is that both steps can be completed one right after the other since adding the label will instantly create the deployment.

PR #4488, opened 2024-07-09
Code injection vulnerability detection

uurien:
### What does this PR do?

This PR adds support to code injection vulnerability detection. To be able to instrument `eval` code, which has different behaviour depending of how it is called ([link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#direct_and_indirect_eval)), we should instrument it using rewriter instead of the usual monkey patching instrumentation.

### Motivation

Support `CODE_INJECTION` vulnerability.

### Plugin Checklist

- [x] Unit tests.
- [x] System tests https://github.com/DataDog/system-tests/pull/2768.

### Additional Notes

[APPSEC-47275]

[APPSEC-47275]: https://datadoghq.atlassian.net/browse/APPSEC-47275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Qard (approved):

packages/dd-trace/test/appsec/iast/vulnerability-formatter/index.spec.js
```diff
@@ -16,6 +22,14 @@ function doTest (testCase, parameters) {
...
   }
 
+  if (expected.vulnerabilities?.length && excludedVulnerabilityTypes.includes(expected.vulnerabilities[0].type)) {
+    return
+  }
+
+  if (excludedTests.includes(description)) {
+    return
+  }
+
```
iunanua, line 32: should these ifs be in a filtering function? something like 

```javascript
function filterByTypeAndDescription (testCase){
  if (testCase.type !== 'VULNERABILITIES' ||
    excludedVulnerabilityTypes.includes(testCase.input[0]?.type) ||
    excludedTests.include(testCase.description)) {
    return false
  } else {
    return true
  }
}

suite.filter(filterByTypeAndDescription)
```

uurien: not easily. Keep in mind that the final description is filled few lines above. They are not filled when we are filtering suite in line 80.

iunanua (approved):

PR #4489, opened 2024-07-09
Make iast config available out of experimental config properties

uurien:
### What does this PR do?

Adds `iast` property to config options, to be able to configure it without `experimental` namespace.

### Motivation

We should have implemented it long time ago.

### Plugin Checklist

- [x] Unit tests.
- [x] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -555,7 +555,12 @@ declare namespace tracer {
         /**
          * Specifies a regex that will redact sensitive source values in vulnerability reports.
          */
-        redactionValuePattern?: string
+        redactionValuePattern?: string,
+
+        /**
+         * Specifies the verbosity of the send telemetry. Default 'INFORMATION'
+         */
+        telemetryVerbosity?: string
       }
 
       appsec?: {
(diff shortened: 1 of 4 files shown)


index.d.ts
```diff
@@ -555,7 +555,12 @@ declare namespace tracer {
         /**
          * Specifies a regex that will redact sensitive source values in vulnerability reports.
          */
-        redactionValuePattern?: string
+        redactionValuePattern?: string,
+
+        /**
+         * Specifies the verbosity of the send telemetry. Default 'INFORMATION'
+         */
+        telemetryVerbosity?: string
```
uurien, line 563: Not directly related with this changes, but it wasn't added in definitions and test.ts when it was implemented.

simon-id: maybe worth having the iast options in a variable so you don't need to copy the definition twice ?

simon-id (dismissed):

simon-id (approved):

Qard (approved):

PR #4494, opened 2024-07-10
add action to retry yarn install when it fails

rochdev:
### What does this PR do?

Add action to retry `yarn install` when it fails.

### Motivation

The registry often returns 500 errors. By retrying the install, it should reduce how much this affects our CI

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -1,0 +1,7 @@
+name: Install dependencies
+runs:
+  using: composite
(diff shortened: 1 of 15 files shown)


juan-fernandez (approved):

rochdev:
It looks like the most important commit of actually using the new action got reverted when I initially opened the PR. I've redone and pushed it so it should be good now.

juan-fernandez (approved):

PR #4506, opened 2024-07-11
move plugin tests to node built-in test runner

rochdev:
### What does this PR do?

Move plugin tests to Node built-in test runner.

### Motivation

Mocha has been problematic as it doesn't support parallelism, it doesn't make it easy to run test files directly, and it has many dependencies that increase the size of the install and can also sometimes be problematic and not upgradable because of our Node version support policy.


BridgeAR:
@rochdev is this still worked upon? I would love to see the Node.js test runner being used. Would this still require a lot more work?
Retrying a test in Node.js test runner does not exist but we could just write/change the test accordingly.

If it's not planned to fllouw up, I guess we should close this?

rochdev:
@BridgeAR I'd keep it open as it's definitely something we'd like to do. The main issue was older versions of Node, but with Node 16 support having just been dropped, it might become possible to get this to the finish line.

BridgeAR:
This has a lot of conflicts and to rebase this will likely need quite some work.
I know @watson also wrote a helper to rewrite tests to another testing framework by checking the AST.
I suggest we close this due to that.

PR #4515, opened 2024-07-16
Remove tight coupling between logger and config

watson:
Removes the need to create a dummy `Config` object just in order to initialize the logger. This was previously required because the logging config wasn't initialized before calling the `Config` constructor. With this PR, the logger is now responsible for initializing itself and the `Config` constructor simply fetches the already initialized logging config from the logger.

## Improvement ideas

Though reduced, there still exist a coupling between the `Config` class and the logger as is evident in the tests. To further reduce this, one option is to simply remove the following properties from the config object:

- `debug`: This is currently accessed in a few places, but it should be possible to change those to get it directly from the logger
- `logger`: This is currently accessed in a few places, but it should be possible to change those to get it directly from the logger
- `logLevel`: This is currently never accessed outside tests

Furthermore, we could remove the ability to configure the logger when calling the `Config` constructor. This is also currently only used by tests, but keeping this functionality only makes our tests more complicated, so I don't think it provides any benefits.

## Note to reviewers

Before this PR, the `Config` class was initialized twice during boot when the `./init.js` file was required (via for example `node -r dd-trace/init server.js`). This had the side-effect of calling `updateConfig` in `./packages/dd-trace/src/telemetry/index.js` twice as well. But since this function [behaves differently](https://github.com/DataDog/dd-trace-js/blob/4624d06be34dc18e3bdf97edde89d0c51157718f/packages/dd-trace/src/telemetry/index.js#L346-L354) the first time it's called, the telemetry that was collected when using `./init.js` vs not, differed.

This PR changes this by only ever only initializing `Config` once, which means that only subsequent calls to `config.configure` will trigger more calls to `updateConfig`. Please verify that this change in behavior is ok.

If it's ok, it must mean that we collected too much telemetry in the past when ever someone used the `./init.js` file.

### DIFF
--- a/init.js
+++ b/init.js
@@ -3,12 +3,9 @@
 const path = require('path')
 const Module = require('module')
 const semver = require('semver')
+const log = require('./packages/dd-trace/src/log')
 const { isTrue } = require('./packages/dd-trace/src/util')
 const telemetry = require('./packages/dd-trace/src/telemetry/init-telemetry')
-
-const log = semver.satisfies(process.versions.node, '>=16')
-  ? require('./packages/dd-trace/src/log')
-  : { info: isTrue(process.env.DD_TRACE_DEBUG) ? console.log : () => {} } // eslint-disable-line no-console
 
 let initBailout = false
 let clobberBailout = false
--- a/init.js
+++ b/init.js
@@ -2,22 +2,13 @@
 
 const path = require('path')
 const Module = require('module')
-const telemetry = require('./packages/dd-trace/src/telemetry/init-telemetry')
 const semver = require('semver')
+const { isTrue } = require('./packages/dd-trace/src/util')
+const telemetry = require('./packages/dd-trace/src/telemetry/init-telemetry')
 
-function isTrue (envVar) {
-  return ['1', 'true', 'True'].includes(envVar)
-}
-
-// eslint-disable-next-line no-console
-let log = { info: isTrue(process.env.DD_TRACE_DEBUG) ? console.log : () => {} }
-if (semver.satisfies(process.versions.node, '>=16')) {
-  const Config = require('./packages/dd-trace/src/config')
-  log = require('./packages/dd-trace/src/log')
-
-  // eslint-disable-next-line no-new
-  new Config() // we need this to initialize the logger
-}
+const log = semver.satisfies(process.versions.node, '>=16')
+  ? require('./packages/dd-trace/src/log')
+  : { info: isTrue(process.env.DD_TRACE_DEBUG) ? console.log : () => {} } // eslint-disable-line no-console
 
 let initBailout = false
 let clobberBailout = false
--- a/packages/dd-trace/src/config.js
+++ b/packages/dd-trace/src/config.js
@@ -211,33 +211,23 @@ function propagationStyle (key, option, defaultValue) {
 }
 
 class Config {
-  constructor (options) {
-    options = options || {}
+  constructor (options = {}) {
     options = this.options = {
       ...options,
       appsec: options.appsec != null ? options.appsec : options.experimental?.appsec,
       iast: options.iast != null ? options.iast : options.experimental?.iast
     }
 
-    checkIfBothOtelAndDdEnvVarSet()
-
     // Configure the logger first so it can be used to warn about other configs
-    this.debug = isTrue(coalesce(
-      process.env.DD_TRACE_DEBUG,
-      process.env.OTEL_LOG_LEVEL && process.env.OTEL_LOG_LEVEL === 'debug',
-      false
-    ))
-    this.logger = options.logger
-
-    this.logLevel = coalesce(
-      options.logLevel,
-      process.env.DD_TRACE_LOG_LEVEL,
-      process.env.OTEL_LOG_LEVEL,
-      'debug'
-    )
+    const logConfig = log.getConfig()
+    this.debug = logConfig.enabled
+    this.logger = coalesce(options.logger, logConfig.logger)
+    this.logLevel = coalesce(options.logLevel, logConfig.logLevel)
 
     log.use(this.logger)
-    log.toggle(this.debug, this.logLevel, this)
+    log.toggle(this.debug, this.logLevel)
+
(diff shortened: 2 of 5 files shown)


init.js
```diff
@@ -2,22 +2,13 @@
...
-if (semver.satisfies(process.versions.node, '>=16')) {
-  const Config = require('./packages/dd-trace/src/config')
-  log = require('./packages/dd-trace/src/log')
-
-  // eslint-disable-next-line no-new
-  new Config() // we need this to initialize the logger
-}
+const log = semver.satisfies(process.versions.node, '>=16')
+  ? require('./packages/dd-trace/src/log')
+  : { info: isTrue(process.env.DD_TRACE_DEBUG) ? console.log : () => {} } // eslint-disable-line no-console
```
watson, line 11: Is the Node.js version check still required now that we no longer support Node.js <16? If not, we could just always require `./packages/dd-trace/src/log`.

packages/dd-trace/src/config.js
```diff
@@ -211,33 +211,27 @@ function propagationStyle (key, option, defaultValue) {
...
-  constructor (options) {
-    options = options || {}
+  constructor (options = {}) {
     options = this.options = {
       ...options,
       appsec: options.appsec != null ? options.appsec : options.experimental?.appsec,
       iast: options.iast != null ? options.iast : options.experimental?.iast
     }
 
-    checkIfBothOtelAndDdEnvVarSet()
```
watson, line 222: I moved this down below the update to the logger to be true to the "Configure the logger first so it can be used to warn about other configs" comment below since the call to `checkIfBothOtelAndDdEnvVarSet` might try to log some warnings.

init.js
```diff
@@ -2,22 +2,13 @@
...
 const path = require('path')
 const Module = require('module')
-const telemetry = require('./packages/dd-trace/src/telemetry/init-telemetry')
 const semver = require('semver')
+const { isTrue } = require('./packages/dd-trace/src/util')
+const telemetry = require('./packages/dd-trace/src/telemetry/init-telemetry')
 
-function isTrue (envVar) {
-  return ['1', 'true', 'True'].includes(envVar)
-}
```
watson, line 10: I don't see why we can't just reuse the existing `isTrue` from `./packages/dd-trace/src/util`?

watson:
@DataDog/telemetry-and-analytics I've requested a review from you, because there's a change in the collected telemetry. I _think_ we previously collected too much data, which should now be fixed by this PR. So I wanted to loop you in in case you had any comments. For details see the "Note to reviewers" section in the PR description.

init.js
simon-id: sounds reasonable, i've also removed a >=16 check somewhere else in the code so i'd say we can remove this one too ? unless someone knows about an edge case idk

packages/dd-trace/src/log/index.js
```diff
@@ -20,13 +22,29 @@ function processMsg (msg) {
...
+  logger: undefined,
+  logLevel: 'debug'
+}
+
 const log = {
+  /**
+   * @returns Read-only version of logging config. To modify config, call `log.use` and `log.toggle`
+   */
+  getConfig () {
+    return Object.freeze({ ...config })
```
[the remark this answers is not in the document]
watson: Yes it will throw. If we're concerned about crashing, we could skip the `freeze` part.

My thought was that I'd rather fail as early as possible than risk a bug being introduced because it was possible to modify the returned config (as someone might mistakingly think they could use it to modify the config).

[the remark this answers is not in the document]
simon-id: Yeah i totally understand your intention here, and i'd agree if it was under a different code constraints. In my opinion, i'd rather risk a config not being applied, than a customer's app crashing, it's a very important constraint we try to follow.
Now it could be interesting to test this on our own apps for a bit, see if any crash happen. But doing this in customer's prod would be basically letting them find bugs for us at a great cost for them.
I'll let you and the others be the judge of the risk reward ratio, i'm just adding some context here :+1:

simon-id:
Otherwise soft LGTM, because I'm not well placed to approve this part of the code

yshapiro-57:
@paullegranddc , do you know the history of why we (Telemetry and Analytics) were made the code owners for a part of the Node.js tracer? This code does not look familiar to me either, there is probably some background story that I am missing.

Also CC @ida613 as the author of the telemetry in the Node.js tracer on the off chance that you know.

watson:
I've talked with @yshapiro-57 on Slack to better understand the telemetry teams thoughts on this type of telemetry. I'll try to summarize our chat here.

First, I think I should explain the bug a little better for anyone reading this:

The Node.js tracer is supposed to send telemetry data with `request_type: 'app-client-configuration-change'` every time the tracer configuration is changed. The way the code is wired up, the `Config` class will call a function to send the telemetry even when the config is first initialized (i.e. on process boot). However, this telemetry function contains a condition to just skip sending the telemetry the first time it's called. I assume(?) this was done intentionally to _not_ send the telemetry on boot, but only on later changes (e.g. if it's updated via Remote Config).

The bug happens when we execute the `Config` constructor twice during boot (which happens if you execute the `./init.js` file), in which case the 2nd execution will trigger the sending of telemetry.

I assume once again, that this isn't what we intended.

After talking with @yshapiro-57, I now understand that it's up to the Node.js tracer "team" if we want this boot-up telemetry or not. If we do however, there's still a bug, in that it currently doesn't differentiate between config options that are actively configured by the user vs config options that will just have their default values initialized.

We probably just want to send telemetry for the config options specifically set by the user at boot, right?

packages/dd-trace/src/log/index.js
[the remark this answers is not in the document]
watson: Fixed in #a99a2f3 👍

init.js
watson: I've removed in b376911 👍 

@bengl are you ok with this as well?

Qard (approved):

Qard (approved):

init.js
bengl: Fine with removing >=v16 check 👍

PR #4517, opened 2024-07-16
Add missing null guards in typeof object checks

watson:


simon-id (dismissed):

Qard (dismissed):

packages/datadog-plugin-openai/src/index.js
```diff
@@ -118,7 +118,7 @@ class OpenApiPlugin extends TracingPlugin {
     }
 
     // createChatCompletion, createCompletion
-    if (typeof payload.logit_bias === 'object' && payload.logit_bias) {
+    if (payload.logit_bias && typeof payload.logit_bias === 'object') {
```
simon-id, line 121: why not check for null directly ? like everywhere else ?

packages/dd-trace/src/appsec/iast/vulnerabilities-formatter/index.js
```diff
@@ -43,6 +43,8 @@ class VulnerabilityFormatter {
     const valueParts = []
     let fromIndex = 0
 
+    if (evidence.value == null) return
```
simon-id, line 46: @uurien little eyes por favor (careful it's a double equal)

watson: A little background on the double equal usage: I found that earlier in the call chain, there's a check for `evidence.value === undefined`, in which case it aborts, so in reality it can never be `undefined` at this location as the code current looks. However, since the _intention_ is that an `undefined` value should never reach this location, I used `==` in case the call chain ever changes in the future.

packages/datadog-plugin-openai/src/index.js
watson: Good catch. Fixed in 824d157

watson: Actually, I had left these truthy checks alone in this PR. But you comment made me revisit them, so I just pushed yet another PR addressing all of them in a similar manner: 4d5cfba

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
```diff
@@ -45,7 +45,7 @@ class TaintTrackingPlugin extends SourceIastPlugin {
     this.addSub(
       { channelName: 'apm:express:middleware:next', tag: HTTP_REQUEST_BODY },
       ({ req }) => {
-        if (req && req.body && typeof req.body === 'object') {
+        if (req && req.body !== null && typeof req.body === 'object') {
```
simon-id, line 48: optional chaining at this point

packages/dd-trace/src/appsec/index.js
```diff
@@ -121,16 +121,16 @@ function incomingHttpEndTranslator ({ req, res }) {
   }
 
   // TODO: temporary express instrumentation, will use express plugin later
-  if (req.params && typeof req.params === 'object') {
+  if (req.params !== null && typeof req.params === 'object') {
```
simon-id, line 124: actually now that i think of it, why change it to explicit null check ?
most of the time the value will either be undefined, so that would require the execution of typeof, or it would be an object, basically almost never null. At last with the truthy check we skip the typeof check for undefined yknow ?

packages/dd-trace/src/appsec/iast/vulnerabilities-formatter/index.js
uurien: This method returns `{ valueParts }`, even when the array is empty. I think that the right solution is:
```suggestion
    if (evidence.value == null) return { valueParts }
```

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
watson: Since I was making so many edits, I wanted to keep the rewriting to a minimum, otherwise it quickly spirals out of control 😅

packages/dd-trace/src/appsec/index.js
watson: That's also why I left these untouched in the first version of the PR. However, I do also like consistency, and while I'm here fixing the actual issues as well, I might as well make it all constant. But I'm happy to revert that change if you think it's better to not touch this

packages/dd-trace/src/appsec/iast/vulnerabilities-formatter/index.js
```diff
@@ -43,6 +43,8 @@ class VulnerabilityFormatter {
     const valueParts = []
     let fromIndex = 0
 
+    if (evidence.value == null) return { valueParts }
```
tlhunter, line 46: `===`?

watson: See previous discussion here: https://github.com/DataDog/dd-trace-js/pull/4517#discussion_r1680714394

tlhunter:
Thanks, Brendan.

tlhunter (approved):

packages/dd-trace/src/appsec/index.js
simon-id: eh whatever it's microoptimization anyway :+1:

simon-id (approved):

PR #4525, opened 2024-07-17
aws-sdk: add batchPropagationEnabled env variable configuration

wconti27:
### What does this PR do?
This PR adds the following env variables:
`DD_TRACE_AWS_SDK_BATCH_PROPAGATION_ENABLED` and `DD_TRACE_AWS_SDK_<SERVICE>_BATCH_PROPAGATION_ENABLED` . These env variables enable Datadog APM trace context injection for each message in batch send operations, as opposed to only injecting the first message in batches. The configuration is available for the following supported AWS SDK messaging services:
- SQS
- SNS
- Kinesis

### Motivation
Customer request

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/datadog-plugin-aws-sdk/src/base.js
+++ b/packages/datadog-plugin-aws-sdk/src/base.js
@@ -163,9 +163,15 @@ function normalizeConfig (config, serviceIdentifier) {
       break
   }
 
+  // check if AWS batch propagation or AWS_[SERVICE] batch propagation is enabled via env variable
+  const serviceId = serviceIdentifier.toUpperCase()
+  const serviceBatchPropagationValue = process.env.DD_TRACE_AWS_SDK_BATCH_PROPAGATION_ENABLED ??
+    process.env[`DD_TRACE_AWS_SDK_${serviceId}_BATCH_PROPAGATION_ENABLED`]
+  const serviceBatchPropagationEnabled = serviceBatchPropagationValue ? isTrue(serviceBatchPropagationValue) : false
+
   return Object.assign({}, config, specificConfig, {
     splitByAwsService: config.splitByAwsService !== false,
-    batchPropagationEnabled: config.batchPropagationEnabled !== false,
+    batchPropagationEnabled: config.batchPropagationEnabled !== false || serviceBatchPropagationEnabled !== false,
     hooks
   })
 }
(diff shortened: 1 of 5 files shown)


tlhunter (changes requested):
This will need a test, plus it appears to be breaking existing tests?

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -163,11 +163,34 @@ function normalizeConfig (config, serviceIdentifier) {
       break
   }
 
-  return Object.assign({}, config, specificConfig, {
+  const baseAWSBatchPropagationValue = process.env.DD_TRACE_AWS_SDK_BATCH_PROPAGATION_ENABLED
```
tlhunter, line 166: I think these should have correlating config values for when a user calls `tracer.init({})` manually.

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -163,11 +163,34 @@ function normalizeConfig (config, serviceIdentifier) {
...
+  const baseAWSBatchPropagationEnabled = baseAWSBatchPropagationValue ? isTrue(baseAWSBatchPropagationValue) : false
+
+  // check if AWS batch propagation or AWS_[SERVICE] batch propagation is enabled via env variable
+  const serviceId = serviceIdentifier.toUpperCase()
```
tlhunter, line 170: Do the service identifiers consist entirely of a-z characters? Are there ever hyphens?

Qard (dismissed):

packages/datadog-plugin-aws-sdk/src/base.js
wconti27: I'm not 100% sure, but we already use this pattern within line 109 of this file and it hasn't caused any problems

wconti27:
> This will need a test, plus it appears to be breaking existing tests?

I've simplified the configuration and added a test case.

packages/datadog-plugin-aws-sdk/src/base.js
wconti27: I simplified this, they are now one and the same.

tlhunter (approved):

PR #4526, opened 2024-07-17
hardcode the latest version of node as 22.4

rochdev:
### What does this PR do?

Hardcode the latest version of node as 22.4

### Motivation

Node 22.5 seems to have a pretty nasty [bug](https://github.com/nodejs/node/issues/53902) that fails a lot of our tests.


Qard (approved):

watson:
Do we have a plan to remember to revert this once the bug is fixed in node core?

PR #4537, opened 2024-07-18
feat: create DSM public api

wconti27:
### What does this PR do?
add DSM public API with consume and produce methods for manual checkpointing

### Motivation
this feature is present in all other tracers

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


Qard (commented):
Seems like this needs tests?

Qard (approved):

piochelepiotr (commented):
I Java, we have the concept of tags that are "hashable", meaning they are used to compute the pathway hash.
manual_checkpoint is not used for the computation of the hash, that way, if the producer is instrumented with automatic instrumentation & manual instrumentation, and the consumer with only automatic, the parent of the consumer might be incorrectly set to the automatic instrumentation (but we prioritize keeping the manual instrumentation).

piochelepiotr (dismissed):

.github/workflows/system-tests.yml
```diff
@@ -33,6 +33,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: 'DataDog/system-tests'
+          ref: conti/dsm-manual-checkpoint-tests
```
Qard: Should this still be pointing at this branch?

packages/dd-trace/src/opentracing/tracer.js
```diff
@@ -38,7 +39,8 @@ class DatadogTracer {
       [formats.TEXT_MAP]: new TextMapPropagator(config),
       [formats.HTTP_HEADERS]: new HttpPropagator(config),
       [formats.BINARY]: new BinaryPropagator(config),
-      [formats.LOG]: new LogPropagator(config)
+      [formats.LOG]: new LogPropagator(config),
+      text_map_dsm: new DSMTextMapPropagator(config)
```
Qard, line 43: Should this not have its own formats key?

packages/dd-trace/test/datastreams/data_streams_checkpointer.spec.js
```diff
@@ -0,0 +1,71 @@
...
+    agent.expectPipelineStats(dsmStats => {
+      let statsPointsReceived = 0
+      // we should have 1 dsm stats points
+      dsmStats.forEach((timeStatsBucket) => {
+        if (timeStatsBucket && timeStatsBucket.Stats) {
+          timeStatsBucket.Stats.forEach((statsBuckets) => {
+            statsPointsReceived += statsBuckets.Stats.length
+          })
+        }
+      })
```
Qard, line 34: It's best to avoid `forEach(...)` as it's generally considered legacy JS at this point and is not well supported by the optimizer. You could do something like:

```suggestion
      let statsPointsReceived = 0
      // we should have 1 dsm stats points
      for (const timeStatsBucket of dsmStats) {
        if (timeStatsBucket && timeStatsBucket.Stats) {
          for (const statsBucket of timeStatsBucket) {
            statsPointsReceived += statsBuckets.Stats.length
          })
        }
      })
```

If you want to take a more functional approach you could do something more like:

```suggestion
      // we should have 1 dsm stats points
      const statsPointsReceived = dsmStats
        .filter((bucket) => bucket && bucket.Stats)
        .flatMap((bucket) => bucket.Stats)
        .map((stats) => stats.Stats.length)
        .reduce((m, v) => m + v, 0)
```

packages/dd-trace/test/datastreams/data_streams_checkpointer.spec.js
```diff
@@ -0,0 +1,71 @@
...
+    agent.expectPipelineStats(dsmStats => {
+      let statsPointsReceived = 0
+      // we should have 2 dsm stats points because of the earlier produce
+      dsmStats.forEach((timeStatsBucket) => {
+        if (timeStatsBucket && timeStatsBucket.Stats) {
+          timeStatsBucket.Stats.forEach((statsBuckets) => {
+            statsPointsReceived += statsBuckets.Stats.length
+          })
+        }
+      })
```
Qard, line 58: Same as above. I would try to avoid using `forEach`.

.github/workflows/system-tests.yml
wconti27: removed this

packages/dd-trace/src/opentracing/tracer.js
wconti27: yes it should, fixing

packages/dd-trace/test/datastreams/data_streams_checkpointer.spec.js
wconti27: Removed all uses of `.forEach`

piochelepiotr (approved):

Qard (approved):

PR #4539, opened 2024-07-19
Add new ESLint plugin: @datadog/eslint-plugin

watson:
Add a new ESLint plugin: [@datadog/eslint-plugin](https://github.com/DataDog/eslint-plugin)

This plugin currently only contains a single rule, `@datadog/safer-typeof-object`, which will ensure that any `typeof x === 'object'` is preceded by a not-null check:

**Bad:**

```js
typeof foo === 'object'
```

**Good:**

```js
foo !== null && typeof foo === 'object'
```

This is because `typeof null === 'object'`, so any operation on the variable that doesn't expect a `null` value will at best lead to unexpected behaviour, and at works to an uncaught exception crashing the process.

The source code previously contained a lot of instances violating this code, but this was recently fixed in #4517. This PR is an attempt to ensure no new violations sneak in.

## Note to reviewers

I'm working on publishing `@datadog/eslint-plugin` to npm (but need access first). But since it's not really an issue to have a dev-dependency that points to a GitHub repo, I think it's safe to merge this as-is. Once the package is published to npm, I'll make a follow up PR to update the dependency to use that version.

### DIFF
--- a/packages/datadog-plugin-child_process/src/scrub-cmd-params.js
+++ b/packages/datadog-plugin-child_process/src/scrub-cmd-params.js
@@ -63,7 +63,7 @@ function scrubChildProcessCmd (expression) {
 
       if (token === null) {
         continue
-      } else if (typeof token === 'object') {
+      } else if (typeof token === 'object') { // eslint-disable-line @datadog/safe-typeof-object
         if (token.pattern) {
           result.push(token.pattern)
         } else if (token.op) {
--- a/packages/dd-trace/src/appsec/iast/taint-tracking/operations-taint-object.js
+++ b/packages/dd-trace/src/appsec/iast/taint-tracking/operations-taint-object.js
@@ -25,6 +25,7 @@ function taintObject (iastContext, object, type) {
           } else {
             parent[key] = tainted
           }
+          // eslint-disable-next-line @datadog/safe-typeof-object
         } else if (typeof value === 'object' && !visited.has(value)) {
           visited.add(value)
 
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,7 +5,8 @@
   "extends": [
     "eslint:recommended",
     "standard",
-    "plugin:mocha/recommended"
+    "plugin:mocha/recommended",
+    "plugin:@datadog/recommended"
   ],
   "plugins": [
     "mocha",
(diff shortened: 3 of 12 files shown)


packages/datadog-plugin-child_process/src/scrub-cmd-params.js
```diff
@@ -63,7 +63,7 @@ function scrubChildProcessCmd (expression) {
 
       if (token === null) {
         continue
-      } else if (typeof token === 'object') {
+      } else if (typeof token === 'object') { // eslint-disable-line @datadog/safe-typeof-object
```
rochdev, line 66: Seems like a limitation of the plugin to me, as there is a check for `null`. If that check is removed for any reason, because of this comment the rule won't catch the now invalid code.

packages/dd-trace/src/appsec/iast/taint-tracking/operations-taint-object.js
```diff
@@ -25,6 +25,7 @@ function taintObject (iastContext, object, type) {
           } else {
             parent[key] = tainted
           }
+          // eslint-disable-next-line @datadog/safe-typeof-object
```
rochdev, line 28: Why not just add the check for `null` instead of disabling the rule? Same question for pretty much all other occurrences.

yoannmoinet:
(Adding my slack comment here to avoid the retention clean)

In web-ui we also use eslint 8, and we have [local custom rules](https://github.com/DataDog/web-ui/tree/preprod/packages-to-migrate/eslint-plugin-datadog) without issues.
As long as you have a yarn workspace and the name of your package as `eslint-plugin-<whatevs>` it will be available as a plugin to eslint as `<whatevs>`.

Example in `web-ui`:
1. [package's name as `eslint-plugin-datadog`](https://github.com/DataDog/web-ui/blob/c5e6ce05b7d546621ec7bb67c89787b6d24fb71d/packages-to-migrate/eslint-plugin-datadog/package.json#L2)
2. [depend on it at root](https://github.com/DataDog/web-ui/blob/9baee5140874a377688e9640d555e5d0e15e33d3/package.json#L345)
3. [use it in `.eslintrc.js`](https://github.com/DataDog/web-ui/blob/7d18ace6f4c47306c663b958bef305a4c164e377/.eslintrc.js#L88)
4. BONUS: [configuration](https://github.com/DataDog/web-ui/blob/7d18ace6f4c47306c663b958bef305a4c164e377/.eslintrc.js#L186-L237)

There's also stuffs like https://github.com/taskworld/eslint-plugin-local  or [`--rulesdir`](https://eslint.org/docs/latest/use/command-line-interface#--rulesdir) that could potentially work, but I think the yarn workspace solution is the most adapted to your needs.

I don't think we should publish a public package just for this, especially with the current effort in https://datadoghq.atlassian.net/wiki/spaces/FRON/pages/3017179854/2023-07-26+-+Common+eslint+and+prettier+config+for+JS+TS+projects (but you probably were not aware of it)

tlhunter (approved):

BridgeAR:
I do believe it would be good to have a datadog scoped eslint package, while it likely won't contain many rules for now.

Is it something we should continue to look into sometime soon?

watson:
Replaced by #5993

PR #4554, opened 2024-07-29
Revise OTel Context and API Implementation To Enable OTel Auto Instrumentation Support

khanayan123:
### What does this PR do?
This PR reworks the existing OpenTelemetry context and API implementation to align more closely with the dd-trace-py implementation. This change aims to fix the bug affecting the integration between OTel's Express auto-instrumentation and dd-trace-js and thus add support for OTel auto instrumentations through dd-trace-js.

### Motivation
Upon testing the OTel Express auto-instrumentation through dd-trace-js, broken traces were observed. These issues stemmed from a lack of proper handling of distributed trace context propagation in the dd-trace-js implementation of the OpenTelemetry context and core API.

### DIFF
--- a/packages/dd-trace/src/opentelemetry/context_manager.js
+++ b/packages/dd-trace/src/opentelemetry/context_manager.js
@@ -14,13 +14,21 @@ class ContextManager {
 
   active () {
     const activeSpan = tracer.scope().active()
+    const store = this._store.getStore()
+    const context = (activeSpan && activeSpan.context()) || store || ROOT_CONTEXT
 
-    const context = (activeSpan && activeSpan.context()) || this._store.getStore() || ROOT_CONTEXT
-    if (context instanceof DataDogSpanContext) {
-      const newSpanContext = new SpanContext(context)
-      return trace.setSpanContext(ROOT_CONTEXT, newSpanContext)
+    if (!(context instanceof DataDogSpanContext)) {
+      return context
     }
-    return context
+
+    if (!context._otelSpanContext) {
+      const newSpanContext = new SpanContext(context)
+      context._otelSpanContext = newSpanContext
+    }
+    if (store && trace.getSpanContext(store) === context._otelSpanContext) {
+      return store
+    }
+    return trace.setSpanContext(store || ROOT_CONTEXT, context._otelSpanContext)
   }
 
   with (context, fn, thisArg, ...args) {
--- a/packages/dd-trace/src/opentelemetry/context_manager.js
+++ b/packages/dd-trace/src/opentelemetry/context_manager.js
@@ -26,15 +26,14 @@ class ContextManager {
   with (context, fn, thisArg, ...args) {
     const span = trace.getSpan(context)
     const ddScope = tracer.scope()
-    if (span && span._ddSpan) {
-      return ddScope.activate(span._ddSpan, () => {
-        const cb = thisArg == null ? fn : fn.bind(thisArg)
-        return this._store.run(context, cb, ...args)
-      })
+    const run = () => {
+      const cb = thisArg == null ? fn : fn.bind(thisArg)
+      return this._store.run(context, cb, ...args)
     }
-
-    const cb = thisArg == null ? fn : fn.bind(thisArg)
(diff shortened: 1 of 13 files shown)


packages/dd-trace/src/opentelemetry/context_manager.js
```diff
@@ -2,61 +2,39 @@
...
-    })
+    if (span && span._ddSpan) {
+      return ddScope.activate(span._ddSpan, () => {
+        const cb = thisArg == null ? fn : fn.bind(thisArg)
+        return this._store.run(context, cb, ...args)
+      })
+    }
+
+    const cb = thisArg == null ? fn : fn.bind(thisArg)
+    return this._store.run(context, cb, ...args)
```
Qard, line 37: You could reduce some duplication with:

```suggestion
    const run = () => {
        const cb = thisArg == null ? fn : fn.bind(thisArg)
        return this._store.run(context, cb, ...args)
    }

    if (span && span._ddSpan) {
      return ddScope.activate(span._ddSpan, run)
    }

    return run()
```

packages/dd-trace/src/opentracing/propagation/text_map.js
```diff
@@ -618,6 +619,104 @@ class TextMapPropagator {
...
+  static _convertOtelContextToDatadog (traceId, spanId, traceFlag, ts, meta = {}) {
+    const origin = null
+    let samplingPriority = traceFlag
+
+    ts = ts?.traceparent || null
+
+    if (ts) {
+      const tsList = ts.split(',').map(member => member.trim())
+      ts = tsList.join(',')
+      if (/[^\x20-\x7E]+/.test(ts)) {
```
Qard, line 632: We already have [a tracestate parser](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/opentracing/propagation/tracestate.js). Is there some reason not to just use that?

packages/dd-trace/src/opentelemetry/context_manager.js
```diff
@@ -2,61 +2,39 @@
...
-    return new Proxy(active, {
-      get (target, key) {
-        return key === 'getValue' ? wrappedGetValue(target) : target[key]
-      }
-    })
+    const context = (activeSpan && activeSpan.context()) || this._store.getStore() || ROOT_CONTEXT
+    if (context instanceof DataDogSpanContext) {
+      const newSpanContext = new SpanContext(context)
+      return trace.setSpanContext(ROOT_CONTEXT, newSpanContext)
+    }
```
Qard, line 22: Wouldn't this mean it'd _never_ get the context from the OTel context manager as long as _our own_ context manager has a span of its own? When would it ever _not_ enter this branch? 🤔

khanayan123: there's quite a few test cases where it wouldn't get a context from our context manager:

`active () {
    const activeSpan = tracer.scope().active()
    const context = (activeSpan && activeSpan.context()) || this._store.getStore() || ROOT_CONTEXT
    const contextSource = activeSpan ? 'activeSpan' : this._store.getStore() ? '_store' : 'ROOT_CONTEXT'
    console.log('Context is coming from:', contextSource)
    if (context instanceof DataDogSpanContext) {
      console.log('inside if statement', 'does it have ._ddContext?', context._ddContext)
      const newSpanContext = new SpanContext(context)
      return trace.setSpanContext(ROOT_CONTEXT, newSpanContext)
    }
    return context
  }`
  
  if you run the unit and integration tests with this it better illustrates it, but it's definitely possible since the with code:
  
  `with (context, fn, thisArg, ...args) {
    const span = trace.getSpan(context)
    const ddScope = tracer.scope()
    const run = () => {
      const cb = thisArg == null ? fn : fn.bind(thisArg)
      return this._store.run(context, cb, ...args)
    }
    if (span && span._ddSpan) {
      return ddScope.activate(span._ddSpan, run)
    }
    return run()
  }`
  can run a context outside the ddScope

Qard (approved):

Qard (approved):

PR #4560, opened 2024-07-29
[test-visibility] Update test code owners extraction logic to use test source file

juan-fernandez:
### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/plugins/ci_plugin.js
+++ b/packages/dd-trace/src/plugins/ci_plugin.js
@@ -16,7 +16,8 @@ const {
   getTestSuiteCommonTags,
   TEST_STATUS,
   TEST_SKIPPED_BY_ITR,
-  ITR_CORRELATION_ID
+  ITR_CORRELATION_ID,
+  TEST_SOURCE_FILE
 } = require('./util/test')
 const Plugin = require('./plugin')
 const { COMPONENT } = require('../constants')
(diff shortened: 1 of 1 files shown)


integration-tests/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+# CODEOWNERS for testing purposes
+ci-visibility/subproject @datadog-dd-trace-js
```
juan-fernandez, line 2: errors are not a problem here since it's just for testing purposes

integration-tests/vitest/vitest.spec.js
```diff
@@ -203,5 +204,34 @@ versions.forEach((version) => {
...
       })
     })
+
+    it('correctly calculates test code owners when working directory is not repository root', (done) => {
+      const eventsPromise = receiver
+        .gatherPayloadsMaxTimeout(({ url }) => url.endsWith('/api/v2/citestcycle'), (payloads) => {
+          const events = payloads.flatMap(({ payload }) => payload.events)
+
+          const test = events.find(event => event.type === 'test').content
+          assert.equal(test.meta[TEST_CODE_OWNERS], JSON.stringify(['@datadog-dd-trace-js']))
```
juan-fernandez, line 214: ℹ️ vitest's `test.suite` is the same as `test.source.file`, so we don't check whether they're different. The reason why this is not the case for other frameworks is backwards compatibility, which does not apply for vitest since its support is recent

romainkomorndatadog (approved):
LGTM

PR #4563, opened 2024-07-31
Baggage support

ida613:
### What does this PR do?
Add support for [baggage](https://opentelemetry.io/docs/concepts/signals/baggage/) in datadog public api

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/dd-trace/src/config.js
```diff
@@ -566,6 +566,9 @@ class Config {
     this._setValue(defaults, 'url', undefined)
     this._setValue(defaults, 'version', pkg.version)
     this._setValue(defaults, 'instrumentation_config_id', undefined)
+    this._setValue(defaults, 'baggagePropagation', true)
+    this._setValue(defaults, 'baggageInject', true)
+    this._setValue(defaults, 'baggageExtract', true)
```
lucaspimentel, line 571: The idea was that the separate `inject`/`extract` settings take precedence if they are explicitly set by the user. I'm not sure how to handle this in Node.js, but you may want to set the default for those two as `undefined`.

This is similar to what we do with
```
DD_TRACE_PROPAGATION_STYLE (use for both extract and inject if they are not set)
DD_TRACE_PROPAGATION_STYLE_EXTRACT (overrides DD_TRACE_PROPAGATION_STYLE for extraction only)
DD_TRACE_PROPAGATION_STYLE_INJECT (overrides DD_TRACE_PROPAGATION_STYLE for injection only)
```

packages/dd-trace/src/opentracing/span.js
```diff
@@ -145,6 +145,18 @@ class DatadogSpan {
     return this._spanContext._baggageItems[key]
   }
 
+  getAllBaggageItems () {
+    return this._spanContext._baggageItems
```
lucaspimentel, line 149: These 3 functions LGTM. My only question is, should we return `_baggageItems` directly here, or instead return an immutable copy, or something like an iterator? The OpenTelemetry spec says we should, but I don't think we necessarily need to follow that in the this API. Are there any concerns with exposing the internal `_baggageItems` to users? This may vary between languages.

ida613: good idea, I'll return an immutable copy to be safe

packages/dd-trace/src/config.js
```diff
@@ -545,8 +549,8 @@ class Config {
     this._setValue(defaults, 'traceId128BitGenerationEnabled', true)
     this._setValue(defaults, 'traceId128BitLoggingEnabled', false)
     this._setValue(defaults, 'tracePropagationExtractFirst', false)
-    this._setValue(defaults, 'tracePropagationStyle.inject', ['datadog', 'tracecontext'])
-    this._setValue(defaults, 'tracePropagationStyle.extract', ['datadog', 'tracecontext'])
+    this._setValue(defaults, 'tracePropagationStyle.inject', ['datadog', 'tracecontext', 'baggage'])
```
rochdev, line 552: A baggage is not a propagation style. If anything, the 2 baggage formats actually map 1:1 with the existing propagation styles.

packages/dd-trace/src/config.js
```diff
@@ -462,6 +462,9 @@ class Config {
     this._setValue(defaults, 'appsec.stackTrace.maxDepth', 32)
     this._setValue(defaults, 'appsec.stackTrace.maxStackTraces', 2)
     this._setValue(defaults, 'appsec.wafTimeout', 5e3) // µs
+    this._setValue(defaults, 'baggageMaxBytes', 8192)
```
rochdev, line 465: There wasn't a max before so this is a breaking change.

packages/dd-trace/src/config.js
```diff
@@ -501,6 +504,7 @@ class Config {
     this._setValue(defaults, 'isManualApiEnabled', false)
     this._setValue(defaults, 'ciVisibilityTestSessionName', '')
     this._setValue(defaults, 'ciVisAgentlessLogSubmissionEnabled', false)
+    this._setValue(defaults, 'legacyBaggageEnabled', true)
```
rochdev, line 507: This being configured in a different way than the new baggage is confusing (one having a toggle and the other one being in the list of propagators)

packages/dd-trace/src/opentracing/span_context.js
```diff
@@ -36,13 +36,15 @@ class DatadogSpanContext {
         ? this._trace.tags[TRACE_ID_128] + this._traceId.toString(16).padStart(16, '0')
         : this._traceId.toString(16).padStart(32, '0')
     }
+    // if (!this._traceId) return ''
```
rochdev, line 39: Dead code should be removed and not just commented out.

rochdev (changes requested):
What happens when there is a conflict between a Datadog baggage and an OpenTelemetry baggage?

packages/dd-trace/src/config.js
bm1549: It's worth mentioning that OTEL considers baggage as a [propagator](https://opentelemetry.io/docs/specs/otel/context/api-propagators/)

rochdev: That design is flawed though. For example, they mention in the same doc that they will add a binary propagator in the future in addition to TextMap. But then they also refer to the data that is added to the carrier within a propagator as also a propagator. So how would you differentiate between baggage propagated by TextMap versus by Binary at that point?

packages/dd-trace/src/config.js
ida613: this doesn't apply to the legacy opentracing baggage

packages/dd-trace/src/opentracing/span_context.js
ida613: will do

packages/dd-trace/src/config.js
lucaspimentel: We are phasing out OpenTracing baggage so we don't want to add it as a supported value to `DD_TRACE_PROPAGATION_STYLE`.

Also, if we add a value for the OpenTracing baggage propagator (`ot-baggage`?) to `DD_TRACE_PROPAGATION_STYLE`, it would have to be included in the default list for Node.js, since OpenTracing is already enabled by default. But this would still be a breaking change for all users that _already_ set `DD_TRACE_PROPAGATION_STYLE`, since _none_ of them will have `ot-baggage` in their config.

By having a separate setting, we avoid a breaking change for users since we can keep it enabled by default for everyone (for now).

packages/dd-trace/src/config.js
rochdev: That could be confusing for a user then, but at least it's non-breaking 👍

packages/dd-trace/src/config.js
rochdev: My assumption was that the other config wouldn't be added to the propagators (since that's semantically incorrect) but if we are just following OpenTelemetry then I guess the discussion should happen over there.

ida613:
> What happens when there is a conflict between a Datadog baggage and an OpenTelemetry baggage?

when the new baggage propagation style this PR proposes comes in conflict with [existing baggage propagation](https://github.com/DataDog/dd-trace-js/blob/fcecd865bf9dec0b79a965d6bf20d7279a7f2ac0/packages/dd-trace/src/opentracing/propagation/text_map.js#L22), the existing style takes precedence. Is this what you are asking?

packages/dd-trace/test/opentracing/propagation/text_map.spec.js
```diff
@@ -67,25 +68,60 @@ describe('TextMapPropagator', () => {
...
+      }
+      const spanContext = createContext({ baggageItems })
 
       propagator.inject(spanContext, carrier)
 
       expect(carrier['ot-baggage-number']).to.equal('1.23')
       expect(carrier['ot-baggage-bool']).to.equal('true')
       expect(carrier['ot-baggage-array']).to.equal('foo,bar')
       expect(carrier['ot-baggage-object']).to.equal('[object Object]')
+      expect(carrier.baggage).to.be.equal('number=1.23,bool=true,array=foo%2Cbar,object=%5Bobject%20Object%5D')
```
lucaspimentel, line 90: `array=foo%2Cbar` here decodes to
```
key:array
value:foo,bar
```

Which we should be able to handle correctly, depite the comma in `foo,bar`. I need to add a test with commas to make sure we handle this correctly .NET, thanks!

rochdev (approved):

packages/dd-trace/src/opentracing/propagation/text_map.js
```diff
@@ -109,10 +109,42 @@ class TextMapPropagator {
...
+      for (const [key, value] of Object.entries(spanContext._baggageItems)) {
+        baggage += `${this._encodeOtelBaggageKey(String(key).trim())}=${encodeURIComponent(String(value).trim())},`
+        if (counter === this._config.baggageMaxItems || counter > this._config.baggageMaxItems) break
+        counter += 1
+      }
+      baggage = baggage.slice(0, baggage.length - 1)
+      let buf = Buffer.from(baggage)
+      if (buf.length > this._config.baggageMaxBytes) {
+        const originalBaggages = baggage.split(',')
+        buf = buf.subarray(0, this._config.baggageMaxBytes)
```
lucaspimentel, line 138: I have a few notes here. While this code looks mostly correct, there's an edge case in `buf.subarray(0, this._config.baggageMaxBytes)` where we could theoretically split a multi-byte character and end up with invalid UTF-8 bytes. I imagine if this happens the `buf.toString('utf8')` below would fail.

BUT, I don't think this will actually ever happen because we should never get multi-byte characters here, since we should have encoded all of them already with `encodeURIComponent(key)` above.

So now I'm thinking, if all characters in the string are guaranteed to be 1-byte ASCII characters after encoding, do we need to `Bugger.from(baggage)` at all, or can use use split the string directly without converting to UTF-8 bytes and back?

lucaspimentel: I think you can keep a count of bytes like the way you already keep a count of items, and follow similar logic. Something like

```js
// consider this pseudo-code 😅
let itemCounter = 0
let byteCounter = 0
let item =`${...},`

itemCounter += 1
byteCounter += item.length // all characters in item are 1 byte after uri-encoding

// stop if we reach either limit
if (itemCounter > maxItems || byteCounter > maxBytes) break

// ...

File: packages/dd-trace/test/opentracing/propagation/text_map.spec.js [hunk 9]
```diff
@@ -77,25 +78,61 @@ describe('TextMapPropagator', () => {
...
+    })
+
+    it('should handle special characters in baggage', () => {
+      const carrier = {}
+      const baggageItems = {
+        '",;\\()/:<=>?@[]{}': '",;\\'
```
lucaspimentel, line 106: Can we add non-ASCII characters like `é` or `我` or `🐶` to ensure those are encoded correctly, too?

(I thought I added this comment before, but I don't see it. Sorry if it's a duplicate.)

lucaspimentel (approved):
I think we can simplify the code that handles the header's byte limit and add a few test cases, but aside from that LGTM!

packages/dd-trace/test/opentracing/propagation/text_map.spec.js
ida613: oops merged before I saw your comments, will create a separate PR to address your comments

PR #4566, opened 2024-08-01
Exploit prevention SQLi in pg

uurien:
### What does this PR do?

Implements detection and blocking of SQL injection attacks in applications using `pg` library in the points where the threat could be exploited.

### Plugin Checklist

- [x] Unit tests.

### Additional Notes

System tests: https://github.com/DataDog/system-tests/pull/2862

[APPSEC-53981]

[APPSEC-53981]: https://datadoghq.atlassian.net/browse/APPSEC-53981?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/datadog-instrumentations/src/pg.js
```diff
@@ -53,21 +53,56 @@ function wrapQuery (query) {
...
       const finish = asyncResource.bind(function (error) {
         if (error) {
           errorCh.publish(error)
         }
         finishCh.publish()
       })
 
+      if (abortController.signal.aborted) {
+        const error = abortController.signal.reason || new Error('Aborted')
+        const reusingQuery = typeof pgQuery.submit === 'function'
```
uurien, line 74: Based on:
https://github.com/brianc/node-postgres/blob/54eb0fa216aaccd727765641e7d1cf5da2bc483d/packages/pg/lib/client.js#L510

packages/dd-trace/src/appsec/rasp/sql_injection.js
```diff
@@ -0,0 +1,64 @@
...
+    [addresses.DB_SYSTEM]: 'postgresql' // TODO: Extract to constant
+  }
+
+  const result = waf.run({ persistent }, req, RULE_TYPES.SQL_INJECTION)
+
+  const res = store?.res
+  handleResult(result, req, res, ctx.abortController, config)
+}
+
+function analyzePgSqlInjectionInPool (ctx) {
```
uurien, line 47: pg.Pool internally calls to pg.Query, we could detect threats only listening to pg.Query, but the stacktrace that we have in pg.Client when we are using pg.Pool is bad, it is executed in other Stack and the information that we have is not important. The stacktrace when pool.Query is executed is more important that we have when pg.Query is executed under pool

packages/dd-trace/test/appsec/rasp/index.spec.js
```diff
@@ -0,0 +1,34 @@
+'use strict'
```
uurien, line 1: Test moved from `rasp.spec.js`

packages/dd-trace/test/appsec/rasp/utils.spec.js
```diff
@@ -0,0 +1,79 @@
+'use strict'
```
uurien, line 1: These test were in `rasp.spec.js` before.

packages/dd-trace/src/appsec/rasp/sql_injection.js
```diff
@@ -0,0 +1,64 @@
...
+  const parentStore = storage.getStore()
+  if (!parentStore) return
+
+  analyzePgSqlInjection(ctx)
+
+  storage.enterWith({ ...parentStore, raspSqlAnalyzed: true, raspSqlParentStore: parentStore })
```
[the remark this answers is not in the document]
uurien: If finish is not executed, it is a instrumentation bug, and it will be a problem also for the tracer spans, because pg span is not going to be closed.

> what if it's not called ?

I guess that worst case, we are not going to analyze more sql calls in the same request, because we are not going t orestore `raspSqlAnalyzed` property. 

> Are you trying to make sure the query is not analyzed twice ?
Yes, don't want to analyze twice the same query, in pg.Pool and pg.Query, because pg-pool calls internally to pg query.

> in that case instead of passing a weird store around you could have a weakSet ?

I'll try to do something like this... but I don't think that I could have a key object that would be unique and same instance in pool and query. I'll try to find something.

Maybe having `req` as key of the weakMap and `Set<String>` with queries as value...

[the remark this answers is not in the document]
simon-id: not sure the query is unique enough.
Another thing to take into consideration, enterWith() is not part of the tc39 async context spec, so we should probably avoid to rely on it in new code, as it probably won't be here forever

packages/dd-trace/src/appsec/rasp/utils.js
```diff
@@ -0,0 +1,67 @@
...
+    this.name = 'DatadogRaspAbortError'
+    this.req = req
+    this.res = res
+    this.blockingAction = blockingAction
+  }
+}
+
+function getGenerateStackTraceAction (actions) {
+  return actions?.generate_stack
+}
```
simon-id, line 26: do we really need a function for that ?

packages/dd-trace/src/appsec/rasp/sql_injection.js
```diff
@@ -0,0 +1,90 @@
...
+    [addresses.DB_STATEMENT]: query,
+    [addresses.DB_SYSTEM]: DB_SYSTEM_POSTGRES
+  }
+
+  const result = waf.run({ persistent }, req, RULE_TYPES.SQL_INJECTION)
+
+  handleResult(result, req, res, ctx.abortController, config)
+}
+
+function hasInputAddress (payload) {
```
simon-id, line 60: this function is quite hot, and i'm not sure the whole concat thing is efficient, you can also just iterate over both list of keys independenlty, no need to create intermediary variables just to be more DRY, in this case a little optimisation would be useful i think

packages/datadog-instrumentations/test/pg.spec.js
```diff
@@ -0,0 +1,265 @@
...
+              it('Should abort query', (done) => {
+                queryClientStartChannel.subscribe(abortQuery)
+
+                client.query('SELECT 1', (err) => {
+                  if (err && err.message === 'Test') {
+                    done()
+                    return
+                  }
+
+                  done(new Error('Query was not aborted'))
```
iunanua, line 81: Could it be something like that?
```suggestion
                  assert.propertyVal(err, 'message', 'Test')
                  done()
```

packages/dd-trace/src/appsec/rasp/sql_injection.js
```diff
@@ -0,0 +1,90 @@
...
+  let appliedAddresses = []
+  if (payload.ephemeral) {
+    appliedAddresses = appliedAddresses.concat(Object.keys(payload.ephemeral))
+  }
+  if (payload.persistent) {
+    appliedAddresses = appliedAddresses.concat(Object.keys(payload.persistent))
+  }
```
iunanua: Less optimized than iterating but more compact and fancy :P
```suggestion
function hasInputAddress ({ ephemeral = [], persistent= [] }) {
  const appliedAddresses = [...Object.keys(ephemeral), ...Object.keys(persistent)]
```

uurien: This was already modified when you send the message, now the code is different.

packages/datadog-instrumentations/test/pg.spec.js
```diff
@@ -0,0 +1,260 @@
...
+
+                  const query = new Query('SELECT 1')
+
+                  client.query(query, err => {
+                    if (err && err.message === 'Test') {
+                      done()
+                      return
+                    }
+
+                    done(err || new Error('Query was not aborted'))
```
iunanua, line 187:
```suggestion
                    assert.propertyVal(err, 'message', 'Test')
```

packages/datadog-instrumentations/test/pg.spec.js
```diff
@@ -0,0 +1,260 @@
...
+                  queryClientStartChannel.subscribe(abortQuery)
+
+                  const query = new Query('SELECT 1')
+                  query.callback = err => {
+                    if (err && err.message === 'Test') {
+                      done()
+                      return
+                    }
+
+                    done(err || new Error('Query was not aborted'))
```
iunanua, line 160:
```suggestion
                    assert.propertyVal(err, 'message', 'Test')
```

packages/datadog-instrumentations/test/pg.spec.js
```diff
@@ -0,0 +1,260 @@
...
+                  const query = new Query('SELECT 1')
+
+                  client.query(query)
+
+                  query.on('error', err => {
+                    if (err && err.message === 'Test') {
+                      done()
+                      return
+                    }
+                    done(err || new Error('Query was not aborted'))
```
iunanua, line 131:
```suggestion
                    assert.propertyVal(err, 'message', 'Test')
```

packages/dd-trace/src/appsec/rasp/sql_injection.js
```diff
@@ -0,0 +1,86 @@
...
+}
+
+function analyzePgSqlInjection (ctx) {
+  const query = ctx.query?.text
+  if (!query) return
+
+  const store = storage.getStore()
+  if (!store) return
+
+  const { raspSqlAnalyzed, req, res } = store
```
iunanua, line 35: Is `raspSqlAnalyzed` needed now?

packages/dd-trace/src/appsec/rasp/sql_injection.js
```diff
@@ -0,0 +1,86 @@
+'use strict'
+
+const { pgQueryStart, pgPoolQueryStart, wafRunFinished } = require('../channels')
+const { storage } = require('../../../../datadog-core')
+const addresses = require('../addresses')
+const waf = require('../waf')
+const { RULE_TYPES, handleResult } = require('./utils')
+
+const DB_SYSTEM_POSTGRES = 'postgresql'
+const reqQueryMap = new WeakMap() // WeakMap<Request, Set<querystring>>
```
iunanua, line 10: would `querytext` more appropriate? 
`querystring` can be confused with the http request query string

packages/dd-trace/src/appsec/rasp/sql_injection.js
uurien: not, it's not needed. deleting

packages/dd-trace/src/appsec/rasp/sql_injection.js
```diff
@@ -0,0 +1,90 @@
...
+  let appliedAddresses = []
+  if (payload.ephemeral) {
+    appliedAddresses = appliedAddresses.concat(Object.keys(payload.ephemeral))
+  }
+  if (payload.persistent) {
+    appliedAddresses = appliedAddresses.concat(Object.keys(payload.persistent))
+  }
+
+  return !!appliedAddresses
+    .find(address => address.startsWith('server.request') || address.startsWith('graphql.server'))
```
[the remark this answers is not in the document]
uurien: Not reasons, I just forgot that it exists. Replacing.

packages/dd-trace/src/appsec/rasp/sql_injection.js
uurien: makes sense, changing it.

iunanua (dismissed):

packages/dd-trace/test/appsec/waf/waf_context_wrapper.spec.js
```diff
@@ -52,6 +53,29 @@ describe('WAFContextWrapper', () => {
     }, 1000)
   })
 
+  it('should publish the payload in the dc channel', () => {
+    const ddwafContext = {
+      run: sinon.stub()
+    }
+    ddwafContext.run.returns([])
```
simon-id, line 60:
```suggestion
      run: sinon.stub().returns([])
    }
```

packages/dd-trace/test/appsec/rasp/sql_injection.spec.js
```diff
@@ -0,0 +1,116 @@
...
+        }
+      }
+      datadogCore.storage.getStore.returns({})
+
+      pgQueryStart.publish(ctx)
+
+      sinon.assert.notCalled(waf.run)
+    })
+
+    it('should not analyze sql injection if no url', () => {
```
simon-id, line 105: copy paste from ssrf im guessing xD
```suggestion
    it('should not analyze sql injection if no query', () => {
```

simon-id (dismissed):
last nit otherwise LGTM

simon-id (approved):

iunanua (approved):

PR #4568, opened 2024-08-02
[test-visibility] Add compatibility with nyc's `all` input

juan-fernandez:
### What does this PR do?

When the `all` option is passed to [nyc](https://github.com/istanbuljs/nyc?tab=readme-ov-file#common-configuration-options), `nyc` calculates the "empty coverage" of all the files that are included in the code coverage calculation. This happens _before_ the test process even begins, so the test process does not have this info in memory. 

Luckily, `nyc` leaves this code coverage info in disc, so it can be read after the test process has finished.

What we do in this PR is to access this info in disc as the test process begins, so we can correctly report total code coverage when `all` is passed. Since we don't want to read the files directly ourselves, we grab the `nyc` instance that runs _in the test process_. How we grab the `nyc` instance is explained as follows:

* The test process passed to `nyc` is added certain files to preload via `node-preload`: [code](https://github.com/istanbuljs/nyc/blob/ae657b66264fe7c19a710271510e41b94a152cfa/bin/nyc.js#L62), one of which is the `wrap` file: [code](https://github.com/istanbuljs/nyc/blob/ae657b66264fe7c19a710271510e41b94a152cfa/lib/wrap.js).
* In this `wrap` file, which already runs in the test process, the `NYC` class is instantiated to call `nyc#wrap`: [code](https://github.com/istanbuljs/nyc/blob/ae657b66264fe7c19a710271510e41b94a152cfa/lib/wrap.js#L28)
* In this `wrap` class method some handles are registered: [code](https://github.com/istanbuljs/nyc/blob/ae657b66264fe7c19a710271510e41b94a152cfa/index.js#L374-L386).

This means that `nyc` is imported from the test process, so we can hook into that require and we can grab an instance by wrapping `wrap`. 

### Motivation

Correctly report total code coverage whenever `all` option is passed. 

Affected test frameworks:
* mocha
* cucumber

Jest and vitest do not rely on `nyc` and playwright and cypress do not report this tag.

### Plugin Checklist

- [x] Unit tests.


packages/datadog-instrumentations/src/mocha/main.js
```diff
@@ -153,6 +159,83 @@ function getOnEndHandler (isParallel) {
...
+        onDone: mochaRunAsyncResource.bind(onReceivedSkippableSuites)
+      })
+    } else {
+      onFinishRequest()
+    }
+  }
+
+  libraryConfigurationCh.publish({
+    onDone: mochaRunAsyncResource.bind(onReceivedConfiguration)
+  })
```
juan-fernandez, line 236: ℹ️ `getExecutionConfiguration` is just a refactor

PR #4575, opened 2024-08-06
[test-visibility] Update telemetry for test visibility

juan-fernandez:
### What does this PR do?
* Improve status_code tag to show the actual 4xx error
* Add CI provider tag
* Add `is_new` tag for new tests
* Add vitest telemetry
* Add telemetry metrics for get known tests request
* Add `is_rum` and browser driver tag
* Add `itr_skipped` metric to cypress skippable tests

Missing (will be done in a follow up PRs):
* Check if git executable is missing - it needs a bit of a refactor in the git.js file

### Motivation
Improve telemetry experience

### DIFF
--- a/packages/datadog-plugin-cypress/src/cypress-plugin.js
+++ b/packages/datadog-plugin-cypress/src/cypress-plugin.js
@@ -44,7 +44,8 @@ const {
   TELEMETRY_ITR_UNSKIPPABLE,
   TELEMETRY_CODE_COVERAGE_NUM_FILES,
   incrementCountMetric,
-  distributionMetric
+  distributionMetric,
+  TELEMETRY_ITR_SKIPPED
 } = require('../../dd-trace/src/ci-visibility/telemetry')
 
 const {
@@ -364,6 +365,7 @@ class CypressPlugin {
         const { skippableTests, correlationId } = skippableTestsResponse
         this.testsToSkip = skippableTests || []
         this.itrCorrelationId = correlationId
+        incrementCountMetric(TELEMETRY_ITR_SKIPPED, { testLevel: 'test' }, this.testsToSkip.length)
       }
     }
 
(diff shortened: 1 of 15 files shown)


packages/datadog-plugin-cypress/src/cypress-plugin.js
```diff
@@ -321,7 +322,8 @@ class CypressPlugin {
     incrementCountMetric(name, {
       testLevel,
       testFramework: 'cypress',
-      isUnsupportedCIProvider: this.isUnsupportedCIProvider,
+      isUnsupportedCIProvider: !this.ciProviderName,
+      provider: this.ciProviderName,
```
anmarchenko, line 326: According to the RFC provider metric is not added on the `event_created` metric, but rather on a new metric `test_session`: 

> A nice to have feature is being able to collect the CI name of the supported CI list of a given test session in order to check the behavior of a given CI provider. Currently we only have the "is_unsupported_ci" boolean tag in the event_create metric, but we are not able to distinguish between supported providers. 

> At the same time the event_create metric along with the event_finish metric are the ones with the largest number of tags which makes them the metrics with the biggest cardinality.

> That's why the proposal is to include a new "test_session" metric with....

juan-fernandez: great catch, thanks! That's the problem with diagonal reading 🤦

packages/dd-trace/src/ci-visibility/exporters/agentless/writer.js
```diff
@@ -57,10 +56,9 @@ class Writer extends BaseWriter {
         Date.now() - startRequestTime
       )
       if (err) {
-        const errorType = getErrorTypeFromStatusCode(statusCode)
         incrementCountMetric(
           TELEMETRY_ENDPOINT_PAYLOAD_REQUESTS_ERRORS,
-          { endpoint: 'test_cycle', errorType }
+          { endpoint: 'test_cycle', statusCode }
```
anmarchenko, line 61: From what I understood, error_type is still valid tag, we just add statusCode as an additional tag: but I might be wrong here

PR #4576, opened 2024-08-06
chore(w3c): avoid setting _dd.parent_id to 16 zeros

mabdinur:
### What does this PR do?
Currently when w3c tracestate headers are extracted by Datadog W3C Propagators:
1) The p value in the tracestate is stored in SpanContext.reparentID.
2) If the p value is not found in the tracestate then SpanContext.reparentID (which will be used to set Span.meta["_dd.parent_id"]) is set to 0000000000000000.

The second condition is no longer required. Datadog internal services no
longer handle the case where `Span.meta["_dd.parent_id"] == 0000000000000000`. `0000000000000000` is treated as invalid value and ignored. We should avoid setting and handling this value in client libraries.

### Motivation
- Simplify the extraction logic for w3c tracecontext headers.
- Avoids unnecessary tags to the Datadog agent.

### Plugin Checklist

- [x] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/opentracing/propagation/text_map.js
+++ b/packages/dd-trace/src/opentracing/propagation/text_map.js
@@ -236,8 +236,7 @@ class TextMapPropagator {
   }
 
   _hasParentIdInTags (spanContext) {
-    return tags.DD_PARENT_ID in spanContext._trace.tags &&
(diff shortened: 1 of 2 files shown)


khanayan123 (approved):
Just need to pass tests but otherwise LGTM

PR #4578, opened 2024-08-07
remove optional chaining operator in agent exporter

khanayan123:
### What does this PR do?
remove optional chaining operator in agent exporter inorder to mitigate any issues with dependency versions that don't support optional chaining

### Motivation
Customer is running into issues with appsec?.standalone?.enabled with the error:

`if (stats.enabled || appsec?.standalone?.enabled) {
                                ^

SyntaxError: Invalid or unexpected token
`

### DIFF
--- a/packages/dd-trace/src/exporters/agent/index.js
+++ b/packages/dd-trace/src/exporters/agent/index.js
@@ -15,7 +15,7 @@ class AgentExporter {
     }))
 
     const headers = {}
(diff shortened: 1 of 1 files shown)


juan-fernandez:
how can this happen? Doesn't our lowest supported node.js version support this syntax?

simon-id:
@khanayan123 yes can we have more info about this ? what exactly is the parser here ? our all our supported node versions should be able to parse optional chaining. So I'm assuming it's a different parser ?

PR #4583, opened 2024-08-07
fix(openai): tracing occurs in correct async context

sabrenner:
### What does this PR do?
Ensures tracing and parentage is tracked appropriately across different async contexts by migrating the OpenAI plugin to use tracing channel/appropriate internal APIs.

### Motivation
Previously, something like:
```javascript
await tracer.trace('outer', async () => {
  await openai.chat.completions.create(...)
  tracer.trace('child', () => {})
})
```
Would have `child` as a child span of the `openai.request` span, and not the `outer` span. By migrating to tracing channel and using the internal APIs for later versions, this parentage/context is maintained. This is validated with a regression test, for both streamed and non-streamed cases.

### Plugin Checklist

- [x] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/datadog-instrumentations/src/openai.js
+++ b/packages/datadog-instrumentations/src/openai.js
@@ -1,15 +1,7 @@
 'use strict'
 
-const {
-  channel,
-  addHook,
-  AsyncResource
-} = require('./helpers/instrument')
+const { addHook } = require('./helpers/instrument')
 const shimmer = require('../../datadog-shimmer')
-
-const startCh = channel('apm:openai:request:start')
-const finishCh = channel('apm:openai:request:finish')
-const errorCh = channel('apm:openai:request:error')
 
 const tracingChannel = require('dc-polyfill').tracingChannel
 const ch = tracingChannel('apm:openai:request')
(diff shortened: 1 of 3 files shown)


rochdev:
> By properly wrapping the traced functions in a shared async resource, this parentage/context is maintained.

Using `AsyncResource` alters the async context for all code in the process. Could we instead switch the `openai` integration over to `TracingChannel` to avoid unintended side-effects of altering the async stack?

sabrenner:
> Could we instead switch the openai integration over to TracingChannel to avoid unintended side-effects of altering the async stack?

For sure! I think this should be pretty doable for the patching of older versions of OpenAI (`<=3`), but I'm wondering if it would be a bit harder for the newer patching because of how we currently trace streamed responses (based on my understanding of TracingChannel). Happy to chat offline with more details, but if we have an instance of a more complex migration of an integration to TracingChannel, that might help me out.

rochdev:
> if we have an instance of a more complex migration of an integration to TracingChannel, that might help me out.

There are `grpc` and `http` that both use the lower-level API since they are also stream-based.

sabrenner:
@rochdev thanks! I think I've implemented this correctly, the test I added is still passing for both streamed and non-streamed cases, and i think it's following established patterns for other `TracingChannel` migrations for other plugins.

Qard (approved):

PR #4585, opened 2024-08-08
Safer shimming

bengl:
### What does this PR do?
Adds a layer of try/catch around instrumentation code, and makes a best-effort attempt to attribute errors to instrumentation (i.e. rather than the instrumented code). In those cases, it handles the errors transparently. This is all only enabled in SSI.

### Additional Notes

Caveats about code that isn't covered here are mentioned in the code comments.

### DIFF
--- a/packages/datadog-shimmer/test/shimmer.spec.js
+++ b/packages/datadog-shimmer/test/shimmer.spec.js
@@ -434,6 +434,7 @@ describe('shimmer', () => {
     it('should not work with a wrap()', () => {
       expect(() => shimmer.wrap(() => {}, () => {})).to.throw()
     })
+
     it('should wrap the function', () => {
       const count = inc => inc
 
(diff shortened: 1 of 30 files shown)


simon-id:
I don't have time to review today but what kind of error would it be able to handle ?

bengl:
@simon-id this is specifically covering errors thrown in instrumentation code on the publisher side of the diagnostics channel.

packages/datadog-shimmer/test/shimmer.spec.js
```diff
@@ -227,6 +227,207 @@ describe('shimmer', () => {
...
+                throw new Error('wrapper error')
+              }
+              return returnValue
+            }
+          })
+
+          expect(await obj.count()).to.equal(3)
+        })
+      })
+      // describe('callback', () => {
```
Qard, line 397: Maybe add a comment to explain why all this is here but commented out.

Qard (approved):
LGTM, just one suggestion.

bengl:
@Qard I converted this back to draft because I'm concerned that `shimmer.wrap()` may have been broken by this approach, when SSI is enabled. Until I verify that's all good, I'd prefer to leave this in draft mode.

Qard (dismissed):

Qard (dismissed):

Qard (approved):

PR #4588, opened 2024-08-09
[test visibility] Add check to determine whether `git` is available

juan-fernandez:
### What does this PR do?

* Add `isGitAvailable` that executes `which git` or `where git` depending on os platform
* Call `isGitAvailable` at the beginning of `sendGitMetadata` (where we upload git metatada)

### Motivation

* Give a better error message whenever git is missing
* Report telemetry when git is missing

### Plugin Checklist

- [x] Unit tests.

### DIFF
--- a/packages/dd-trace/src/plugins/util/git.js
+++ b/packages/dd-trace/src/plugins/util/git.js
@@ -77,6 +77,18 @@ function isDirectory (path) {
   }
 }
 
+function isGitAvailable () {
+  const isWindows = os.platform() === 'win32'
+  const command = isWindows ? 'where' : 'which'
+  try {
+    cp.execFileSync(command, ['git'], { stdio: 'pipe' })
+    return true
+  } catch (e) {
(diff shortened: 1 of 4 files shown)


packages/dd-trace/src/plugins/util/git.js
```diff
@@ -77,6 +77,18 @@ function isDirectory (path) {
...
 }
 
+function isGitAvailable () {
+  const isWindows = os.platform() === 'win32'
+  const command = isWindows ? 'where' : 'which'
+  try {
+    cp.execFileSync(command, ['git'], { stdio: 'pipe' })
+    return true
+  } catch (e) {
+    incrementCountMetric(TELEMETRY_GIT_COMMAND_ERRORS, { command: 'check_git', exitCode: 'missing' })
```
juan-fernandez, line 87: ⚠️ `check_git` is not standard but TBH it seems weird to use any of the other tags. In any case, when checking for git errors, we group by `command`, so this value should also appear in our dashboards.

Let me know if I'm missing something

anmarchenko (approved):
LGTM, should work correctly as far as I understand

PR #4592, opened 2024-08-09
PROF-10320: Separate SSI telemetry and heuristics activation

szegedi:
### What does this PR do?
Currently `DD_PROFILING_ENABLED=auto` will trigger SSI telemetry emission, just as if `DD_INJECTION_ENABLED=profiler` was specified. However, the two are not identical in requirements for emitting telemetry.
* Telemetry should be emitted if and only if `DD_INJECTION_ENABLED` is defined with any value, and `DD_PROFILING_ENABLED=false` is not explicitly specified.
* Heuristics should activate if and only if `DD_INJECTION_ENABLED=profiler` or `DD_PROFILING_ENABLED=auto` are specified.

We need to separate the two.
* This PR starts off by simplifying the env var configuration around SSI. Instead of `_applyEnvironment()` deriving values for `profiling.ssi` and `profiling.heuristicsEnabled`, it now just mirrors the values for `DD_PROFILING_ENABLED` and `DD_INJECTION_ENABLED`. To this end, `profiling.enabled` is no longer a boolean, as it now needs to support four values (true, false, auto, undefined) instead of three ones (true, false, undefined.) For less confusion around boolean-ey value names, we now use a string with values "enabled", "disabled", and "auto" (while keeping undefined.)
* We completely removed `profiling.ssi` and `profiling.heuristicsEnabled` in favor of the now 4-state `profiling.enabled` and the newly introduced `injectionEnabled` that is an array of parsed values from `DD_INJECTION_ENABLED`. 
* We touched up `telemetry.logCollection` as it, too, used to be derived in `_applyEnvironment()`. Now it got moved to its rightful place in `_applyCalculated()`.
* Finally, we use these new configuration values to revamp `ssi-heuristics.js` and its uses. We managed to eliminate the frankly unnecessary profiling `enabled` config too – if profiling is not enabled, we short circuit the path to creating any profiling related objects now in `proxy.js`.

### Motivation
SSI telemetry and heuristics are actually orthogonal concerns. We also need the separation to later cleanly implement the emission of SSI information in profiles' system info.


packages/dd-trace/src/config.js
```diff
@@ -1019,6 +1013,13 @@ class Config {
       calc['tracePropagationStyle.inject'] = calc['tracePropagationStyle.inject'] || defaultPropagationStyle
       calc['tracePropagationStyle.extract'] = calc['tracePropagationStyle.extract'] || defaultPropagationStyle
     }
+
+    const iastEnabled = coalesce(this._options['iast.enabled'], this._env['iast.enabled'])
+    const profilingEnabled = coalesce(this._options['profiling.enabled'], this._env['profiling.enabled'])
+    const injectionIncludesProfiler = (this._env.injectionEnabled || []).includes('profiler')
+    if (iastEnabled || ['auto', 'true'].includes(profilingEnabled) || injectionIncludesProfiler) {
```
uurien, line 1020: If I understanding this correctly... previous approach: `DD_TELEMETRY_LOG_COLLECTION_ENABLED=false` (manually set to false) with iast enabled => telemetry.logCollection == disabled

with current approach
`DD_TELEMETRY_LOG_COLLECTION_ENABLED=false` (manually set to false) with iast enabled => telemetry.logCollection == enabled

Am I right? Or am I missunderstanding something related with `_setBoolean(calc...`

szegedi: Calculated values have lower priority than environment variables  (see the order of elements in the `containers` array in `_merge()` function), so an explicit `false` in env vars should still take precedence.

Qard (approved):

PR #4603, opened 2024-08-13
Prepare prerelease branch after each push

uurien:
### What does this PR do?

Each time that something is pushed in `master` branch, it will try to create a `v5.x-proposal` (and `v4x-proposal`) branch from v5.x branch to cherry pick all the changes.
This new branch will be pushed to remote, and each time that something is merged in `master` the proposal branch will be updated. 
If conflict is detected, the job will fail and we are going to have to fix it manually, cherry-picking and resolving the error in local and pushing it. 

These prerelease proposal branches are going to be used to create the new `v5.x.y-proposal` branch and the PR, with all the conflicts already resolved by the people who created it.

> **TODO** in a not too distant future, we could run tests nightly to check if next release will start to fail.

### Motivation

Try to make the release process easier, without need to understand and fix other people code when a conflict happens.

### DIFF
--- a/.github/actions/prepare-prerelease/action.yml
+++ b/.github/actions/prepare-prerelease/action.yml
@@ -1,0 +1,67 @@
+name: Prepare prerelease branch
+
+inputs:
+  base-branch:
+    description: 'The base branch to create the prerelease branch from'
+    required: true
+  github-token:
+    description: 'Github token to use in branch-diff'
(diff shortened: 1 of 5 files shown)


Qard (approved):

uurien:
With the new release process, this is a bit outdated.

PR #4610, opened 2024-08-16
remove shimmer.unwrap and shimmer.massUnwrap

bengl:
We weren't using them outside a few use cases:

1) In appsec, where it was used as enabled/disabled state, and that's
   now been replaced with a variable.
2) In OpenAI, where it was added out of fear of the map in shimmer
   holding on to objects. That map gets removed by this commit anyway.

Meanwhile, unwrap added an extra unnecessary stack frame to every function call when using `wrapFunction`.

### DIFF
--- a/packages/datadog-instrumentations/src/openai.js
+++ b/packages/datadog-instrumentations/src/openai.js
@@ -331,11 +331,6 @@ for (const shim of V4_PACKAGE_SHIMS) {
 
                 throw error
               })
-              .finally(() => {
-              // maybe we don't want to unwrap here in case the promise is re-used?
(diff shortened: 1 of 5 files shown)


Qard (approved):

iunanua (approved):

rochdev (approved):

BridgeAR (commented):
RSLGTM (I did look at the code and it seemed fine, I just do not yet know if and under what circumstances this is needed).

Since it was approved before, is anything keeping this from being merged (after another round of updating and giving another approval)?

BridgeAR:
Closing due to already being removed

PR #4621, opened 2024-08-23
[test visibility] Add test session name

juan-fernandez:
### What does this PR do?

* Read `DD_SESSION_NAME` env var
* Set it to `test_session.name` tag if it's there

Additionally, a bug in `vitest` was fixed:
* Now `test.command` is properly set in test suites and tests

### Motivation

We want to increase the likelihood that the user has a stable session fingerprint.

### Plugin Checklist

- [x] Unit tests.

### DIFF
--- a/packages/dd-trace/src/plugin_manager.js
+++ b/packages/dd-trace/src/plugin_manager.js
@@ -136,10 +136,19 @@ module.exports = class PluginManager {
       dbmPropagationMode,
       dsmEnabled,
       clientIpEnabled,
-      memcachedCommandEnabled
+      memcachedCommandEnabled,
+      ciVisibilitySessionName
     } = this._tracerConfig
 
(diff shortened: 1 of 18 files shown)


packages/dd-trace/src/plugin_manager.js
```diff
@@ -161,10 +166,6 @@ module.exports = class PluginManager {
       sharedConfig.clientIpEnabled = clientIpEnabled
     }
 
-    sharedConfig.site = site
-    sharedConfig.url = url
-    sharedConfig.headers = headerTags || []
-
```
juan-fernandez, line 167: In addition to adding `ciVisibilitySessionName`, I refactored of the `sharedConfig.*` assignments, which could be done at once

Qard (approved):

anmarchenko (approved):

PR #4624, opened 2024-08-23
test: fix the chrome selenium identifier

tlhunter:
### What does this PR do?
- fixes a test currently failing in master

### Motivation
- all green policy
- I suspect a dependent package updated the identifier
- https://pptr.dev/guides/headless-modes


szegedi (approved):

juan-fernandez:
thanks for taking care of this. Weird that it happened for older versions of selenium too. I'll have a look

PR #4626, opened 2024-08-26
Test: Allow FakeAgent to respond to remote config requests

watson:
Allow integration tests to mock the remote config abilities of the agent.

This is done by first setting up a new remote config "file" using the new `FakeAgent#addRemoteConfig` method before running the main part of the integration test. This primes the fake agent to respond with this config for http requests to the `/v0.7/config` endpoint.

## Note to reviewers

This ability is needed by the integration tests in an upcoming Dynamic Instrumentation PR. I've separated it into its own PR to make the review easier.

I have not updated any of the existing tests to make uses of this ability. So this new ability of the `FakeAgent` class isn't actually used by any of our existing code in this PR. Let me know if there's an obvious place where we could make use of this so that I can update the PR to do so.

To make the review process easier, I've split this PR into two commits:

1. 73cf5d765948f368857a7a50448555bfb160b614: This commit contains the actual code changes and is a good starting point for easier diffing.
2. bc05f98d46c8634ae4c42cc16771533114cdd505: This commit contains only some refactoring to move code around to easier maintenance.

## Usage example

Here's an example of how we'd use the new `FakeAgent#addRemoteConfig` method to register a Remote Config file for use in Dynamic Instrumentation:

```js
agent.addRemoteConfig({
  product: 'LIVE_DEBUGGING',
  id: 'foobar',
  config: {
    id: 'foobar,
    version: 0,
    type: 'LOG_PROBE',
    language: 'javascript',
    where: { sourceFile: 'index.js', lines: ['15'] },
    tags: [],
    template: 'Hello World!',
    segments: [{ str: 'Hello World!' }],
    captureSnapshot: false,
    capture: { maxReferenceDepth: 3 },
    sampling: { snapshotsPerSecond: 5000 },
    evaluateAt: 'EXIT'
  }
})
```

### DIFF
--- a/integration-tests/helpers.js
+++ b/integration-tests/helpers.js
@@ -1,6 +1,8 @@
 'use strict'
 
 const { promisify } = require('util')
+const { createHash } = require('crypto')
+const uuid = require('crypto-randomuuid')
 const express = require('express')
 const bodyParser = require('body-parser')
 const msgpack = require('msgpack-lite')
@@ -24,6 +26,7 @@ class FakeAgent extends EventEmitter {
   constructor (port = 0) {
     super()
     this.port = port
+    this._rc_files = []
   }
 
   async start () {
@@ -36,6 +39,78 @@ class FakeAgent extends EventEmitter {
       this.emit('message', {
         headers: req.headers,
         payload: msgpack.decode(req.body, { codec })
+      })
+    })
+    app.post('/v0.7/config', (req, res) => {
+      const {
+        client: { products },
+        cached_target_files: cachedTargetFiles
+      } = req.body
+
+      const expires = (new Date(Date.now() + 1000 * 60 * 60 * 24)).toISOString() // in 24 hours
+      const clientID = uuid() // TODO: What is this? It isn't the runtime-id
+
+      // Currently, only `opaque_backend_state` and `targets` are used by dd-trace-js in the object below
+      const targets = {
+        signatures: [],
+        signed: {
+          _type: 'targets',
+          custom: {
+            agent_refresh_interval: 5,
+            opaque_backend_state: ''
+          },
+          expires,
+          spec_version: '1.0.0',
+          targets: {},
+          version: 12345
+        }
+      }
+      const opaqueBackendState = {
+        version: 2,
+        state: { file_hashes: { key: [] } }
+      }
+      const targetFiles = []
+      const clientConfigs = []
+
+      const files = this._rc_files.filter(({ product }) => products.includes(product))
+
+      for (const { orgId, product, id, name, config } of files) {
+        const path = `datadog/${orgId}/${product}/${id}/${name}`
+        const fileDigest = createHash('sha256').update(config).digest()
+        const fileDigestHex = fileDigest.toString('hex')
+
+        if (cachedTargetFiles.some((cached) =>
+          path === cached.path &&
+          fileDigestHex === cached.hashes.find((e) => e.algorithm === 'sha256').hash
+        )) {
+          continue // skip files already cached by the client so we don't send them more than once
+        }
+
+        opaqueBackendState.state.file_hashes.key.push(fileDigest.toString('base64'))
+
+        targets.signed.targets[path] = {
+          custom: {
+            c: [clientID],
+            'tracer-predicates': { tracer_predicates_v1: [{ clientID }] },
+            v: 20
+          },
+          hashes: { sha256: fileDigestHex },
+          length: config.length
+        }
+
+        targetFiles.push({ path, raw: base64(config) })
+        clientConfigs.push(path)
+      }
+
+      targets.signed.custom.opaque_backend_state = base64(opaqueBackendState)
+
+      // TODO: What does the real agent do if there's nothing to return? Does it just return empty arrays and objects
+      // like we do here, or do we need to change the algorithm to align?
+      res.json({
+        roots: [], // Not used by dd-trace-js currently, so left empty
+        targets: base64(targets),
+        target_files: targetFiles,
+        client_configs: clientConfigs
       })
     })
     app.post('/profiling/v1/input', upload.any(), (req, res) => {
@@ -73,6 +148,31 @@ class FakeAgent extends EventEmitter {
       this.server.on('close', resolve)
       this.server.close()
     })
+  }
+
+  /**
+   * Remove any existing config added by calls to FakeAgent#addRemoteConfig.
+   */
+  resetRemoteConfig () {
+    this._rc_files = []
+  }
+
+  /**
+   * Add a config object to be returned by the fake Remote Config endpoint.
+   * @param {Object} config - Object containing the Remote Config "file" and metadata
+   * @param {number} [config.orgId=2] - The Datadog organization ID
+   * @param {string} config.product - The Remote Config product name
+   * @param {string} config.id - The Remote Config config ID
+   * @param {string} [config.name] - The Remote Config "name". Defaults to the sha256 hash of `config.id`
+   * @param {Object} config.config - The Remote Config "file" object
+   */
+  addRemoteConfig (config) {
+    config = { ...config }
+    config.orgId = config.orgId || 2
+    config.name = config.name || createHash('sha256').update(config.id).digest('hex')
+    config.config = JSON.stringify(config.config)
+
+    this._rc_files.push(config)
   }
 
   // **resolveAtFirstSuccess** - specific use case for Next.js (or any other future libraries)
@@ -484,6 +584,11 @@ function useSandbox (...args) {
 }
 function sandboxCwd () {
   return sandbox.folder
+}
+
+function base64 (strOrObj) {
+  const str = typeof strOrObj === 'string' ? strOrObj : JSON.stringify(strOrObj)
+  return Buffer.from(str).toString('base64')
 }
 
 module.exports = {
--- a/integration-tests/helpers/fake-agent.js
+++ b/integration-tests/helpers/fake-agent.js
@@ -13,7 +13,8 @@ module.exports = class FakeAgent extends EventEmitter {
   constructor (port = 0) {
     super()
     this.port = port
-    this._rc_files = {}
+    this._rcFiles = {}
+    this._rcTargetsVersion = 0
   }
 
   async start () {
@@ -60,7 +61,8 @@ module.exports = class FakeAgent extends EventEmitter {
       length: config.config.length
     }
 
-    this._rc_files[config.id] = config
+    this._rcFiles[config.id] = config
+    this._rcTargetsVersion++
   }
 
   /**
@@ -71,7 +73,7 @@ module.exports = class FakeAgent extends EventEmitter {
   updateRemoteConfig (id, config) {
     config = JSON.stringify(config)
     config = Object.assign(
-      this._rc_files[id],
+      this._rcFiles[id],
       {
         config,
         fileHash: createHash('sha256').update(config).digest('hex')
@@ -80,6 +82,7 @@ module.exports = class FakeAgent extends EventEmitter {
     config.meta.custom.v++
     config.meta.hashes.sha256 = config.fileHash
     config.meta.length = config.config.length
+    this._rcTargetsVersion++
   }
 
   /**
@@ -87,14 +90,16 @@ module.exports = class FakeAgent extends EventEmitter {
    * @param {string} id - The ID of the config object that should be removed
    */
   removeRemoteConfig (id) {
-    delete this._rc_files[id]
+    delete this._rcFiles[id]
+    this._rcTargetsVersion++
   }
 
   /**
    * Remove any existing config added by calls to FakeAgent#addRemoteConfig.
    */
   resetRemoteConfig () {
-    this._rc_files = {}
+    this._rcFiles = {}
+    this._rcTargetsVersion++
   }
 
   // **resolveAtFirstSuccess** - specific use case for Next.js (or any other future libraries)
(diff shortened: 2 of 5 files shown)


watson:
* **#4549** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4549?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4492** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4492?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4626** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4626?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

simon-id:
I see there is no mention of RC capabilities. The `client.capabilities` field is a bitflag number encoded in base64 where each bit tells the backend that this client supports a specific feature. It's not technically required for the RC client to work, but we assume that the backend will send only what we support. I guess we can just assume the tests won't ever send data that we don't support :+1:

watson:
> I guess we can just assume the tests won't ever send data that we don't support 👍

That was my idea, which is why I didn't bother validating the `capabilities` bitfield 🤷

integration-tests/helpers.js
```diff
@@ -38,6 +41,69 @@ class FakeAgent extends EventEmitter {
...
+        const opaqueBackendState = {
+          version: 2,
+          state: { file_hashes: { key: [] } }
+        }
+        const targetFiles = []
+        const clientConfigs = []
+
+        const files = this._rc_files.filter(({ product }) => products.includes(product))
+
+        for (const { orgId, product, id, name, config } of files) {
```
simon-id: when a config file is already received and applied by the library, it's specified in the `cached_target_files` field, and the agent is supposed to not send them again to optimize it. I think sending them should not cause any problem as we have a check that make sure to ignore config files that have the same hashes as already applied ones, but I'm mentioning it just in case.

integration-tests/helpers/fake_agent.js
```diff
@@ -0,0 +1,257 @@
+'use strict'
```
bengl, line 1: nearly every file in our repo uses a hyphen `-` as a separator in the name rather than underscores `_`. Lets keep that consistency.

bengl (changes requested):
Just the file name change, that's all.

integration-tests/helpers/fake_agent.js
simon-id: folders ? yes. Files ? eeehhhh ? this is the dd-trace/src/ folder:
image

simon-id: image

simon-id: now it does look like integration-tests files have a tendency to use hyphen instead of underscore :+1:

integration-tests/helpers.js
```diff
@@ -38,6 +41,69 @@ class FakeAgent extends EventEmitter {
...
+    app.post('/v0.7/config', (req, res) => {
+      const buffers = []
+      req.on('data', buffers.push.bind(buffers))
+      req.on('end', () => {
+        const { products } = JSON.parse(Buffer.concat(buffers).toString()).client
+
+        const expires = (new Date(Date.now() + 1000 * 60 * 60 * 24)).toISOString() // in 24 hours
+        const clientID = uuid() // TODO: What is this? It isn't the runtime-id
+
+        // Currently, only `opaque_backend_state` and `targets` are used by dd-trace-js in the object below
```
simon-id: we use `opaque_backend_state` ??? this field should be as it's called, opaque, it's just a value that we need to keep in memory so the backend doesn't have to. Not implementing it shouldn't cause any problem ?

integration-tests/helpers/fake_agent.js
watson: I guess that's why I got confused and used the underscore. In any case, I've now reamed the file to `fake-agent.js` (in Kibana we had a lint rule for our filenames 😉)

integration-tests/helpers.js
watson: Fixed 👍

integration-tests/helpers.js
```diff
@@ -38,6 +41,69 @@ class FakeAgent extends EventEmitter {
         payload: msgpack.decode(req.body, { codec })
       })
     })
+    app.post('/v0.7/config', (req, res) => {
+      const buffers = []
+      req.on('data', buffers.push.bind(buffers))
+      req.on('end', () => {
+        const { products } = JSON.parse(Buffer.concat(buffers).toString()).client
```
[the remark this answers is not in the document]
watson: Now that I've added support for `cached_target_files` as you suggested in https://github.com/DataDog/dd-trace-js/pull/4626#discussion_r1732966936, does that also satisfy this comment, or did you mean something completely different?

integration-tests/helpers.js
watson: Yes, it's referenced here:

https://github.com/DataDog/dd-trace-js/blob/2ded9362c2eac679791c55b6c7b5c14574362f07/packages/dd-trace/src/appsec/remote_config/manager.js#L218

watson: besides that, it's not being used for anything (e.g. it's not being parsed), so we could technically send down any random value or nothing at all and the code would continue to function.

Do you want me to make any changes?

integration-tests/helpers/fake-agent.js
```diff
@@ -0,0 +1,266 @@
...
+   * @param {Object} config - Object containing the Remote Config "file" and metadata
+   * @param {number} [config.orgId=2] - The Datadog organization ID
+   * @param {string} config.product - The Remote Config product name
+   * @param {string} config.id - The Remote Config config ID
+   * @param {string} [config.name] - The Remote Config "name". Defaults to the sha256 hash of `config.id`
+   * @param {Object} config.config - The Remote Config "file" object
+   */
+  addRemoteConfig (config) {
+    config = { ...config }
+    config.orgId = config.orgId ?? 2
```
bengl, line 60: The `??` syntax isn't supported on Node.js 12, which we're testing for `init.spec.js`.

There are multiple ways around this:

* Use legacy syntax in this file.
* Load this file (and indeed the fake agent) only when needed. (e.g. a lazy-loading getter)
* Other refactoring 🤷 

Unfortunately, `init.spec.js` needs other things in the helper file, so it can't just not use this entirely. The good news is that file does _not_ need the fake agent.

bengl (changes requested):

integration-tests/helpers/fake-agent.js
watson: Thanks. I've just updated to use `||` instead of `??`. Easier than implementing lazy loading of this module 😅

watson:
FYI, you can see this new capability of the `FakeAgent` in use in this upcoming PR: https://github.com/DataDog/dd-trace-js/pull/4492

integration-tests/helpers/fake-agent.js
```diff
@@ -0,0 +1,266 @@
...
+   * Add a config object to be returned by the fake Remote Config endpoint.
+   * @param {Object} config - Object containing the Remote Config "file" and metadata
+   * @param {number} [config.orgId=2] - The Datadog organization ID
+   * @param {string} config.product - The Remote Config product name
+   * @param {string} config.id - The Remote Config config ID
+   * @param {string} [config.name] - The Remote Config "name". Defaults to the sha256 hash of `config.id`
+   * @param {Object} config.config - The Remote Config "file" object
+   */
+  addRemoteConfig (config) {
+    config = { ...config }
```
simon-id, line 59: nit: isn't Object.assign() faster ? :p

integration-tests/helpers/fake-agent.js
```diff
@@ -0,0 +1,266 @@
...
+  })
+
+  app.post('/v0.7/config', (req, res) => {
+    const {
+      client: { products },
+      cached_target_files: cachedTargetFiles
+    } = req.body
+
+    const expires = (new Date(Date.now() + 1000 * 60 * 60 * 24)).toISOString() // in 24 hours
+    const clientID = uuid() // TODO: What is this? It isn't the runtime-id
```
simon-id, line 177: I'd say it's the value sent in `client.id` which isn't the library runtimeId, it's a RC specific ID that the backend use to target configs

simon-id: But tbf you can remove expires, clientID, opaqueState, and all, and it would still work just fine

simon-id: (especially if you don't check that expired config are ignored, and that the opaque backend state is sent back correctly)

integration-tests/helpers/fake-agent.js
```diff
@@ -0,0 +1,266 @@
...
+    }
+    const targetFiles = []
+    const clientConfigs = []
+
+    const files = agent._rc_files.filter(({ product }) => products.includes(product))
+
+    for (const { orgId, product, id, name, config } of files) {
+      const path = `datadog/${orgId}/${product}/${id}/${name}`
+      const fileDigest = createHash('sha256').update(config).digest()
+      const fileDigestHex = fileDigest.toString('hex')
```
simon-id, line 206: those 3 could be precomputed when you add them to `agent._rc_files` no ? instead of recomputing them everytime ?

integration-tests/helpers/fake-agent.js
```diff
@@ -0,0 +1,266 @@
...
+
+      if (cachedTargetFiles.some((cached) =>
+        path === cached.path &&
+        fileDigestHex === cached.hashes.find((e) => e.algorithm === 'sha256').hash
+      )) {
+        continue // skip files already cached by the client so we don't send them more than once
+      }
+
+      opaqueBackendState.state.file_hashes.key.push(fileDigest.toString('base64'))
```
simon-id, line 215: difficult to validate what this is doing as we don't use it anyway :/

integration-tests/helpers/fake-agent.js
```diff
@@ -0,0 +1,266 @@
...
+      headers: req.headers,
+      payload: msgpack.decode(req.body, { codec })
+    })
+  })
+
+  app.post('/v0.7/config', (req, res) => {
+    const {
+      client: { products },
+      cached_target_files: cachedTargetFiles
+    } = req.body
```
simon-id, line 174: if client.state.error exists you might want to print it ? it means the RC client couldn't parse stuff correctly

integration-tests/helpers/fake-agent.js
```diff
@@ -0,0 +1,266 @@
...
+        length: config.length
+      }
+
+      targetFiles.push({ path, raw: base64(config) })
+      clientConfigs.push(path)
+    }
+
+    targets.signed.custom.opaque_backend_state = base64(opaqueBackendState)
+
+    // TODO: What does the real agent do if there's nothing to return? Does it just return empty arrays and objects
```
simon-id, line 233: it returns `{}`

integration-tests/helpers/fake-agent.js
```diff
@@ -0,0 +1,266 @@
...
+      )) {
+        continue // skip files already cached by the client so we don't send them more than once
+      }
+
+      opaqueBackendState.state.file_hashes.key.push(fileDigest.toString('base64'))
+
+      targets.signed.targets[path] = {
+        custom: {
+          c: [clientID],
+          'tracer-predicates': { tracer_predicates_v1: [{ clientID }] },
```
simon-id, line 220: we don't use those do we ?

integration-tests/helpers/fake-agent.js
simon-id: Same thing for `client.state.config_states[].apply_error`, if it's not empty it means the config could not be properly applied. Not really a mocking thing but could be useful to print for debugging tests ?

integration-tests/helpers/fake-agent.js
```diff
@@ -0,0 +1,266 @@
...
+          v: 20
+        },
+        hashes: { sha256: fileDigestHex },
+        length: config.length
+      }
+
+      targetFiles.push({ path, raw: base64(config) })
+      clientConfigs.push(path)
```
simon-id, line 228: pretty sure this needs to be included even when the target_file is cached. So add it before the check. Otherwise if the path is not in client configs, it will consider it needs to delete that config.

simon-id (changes requested):
I would say let's remove all the superfluous fields. RC is complicated enough as it is haha. Let's keep the futurproofing for the futur, and focus on what we need to work first.
Lots of coments here sorry!

integration-tests/helpers/fake-agent.js
watson: I've seen benchmarks suggesting the opposite, so I don't think so. But in any case:

1. If there's a performance difference, it's only noticeable for very large objects, so not worth optimizing for in this case.
2. This is a test and not a hot code-path, so we should always opt for the more modern syntax if possible.

But in general, unless there's a huge difference (which is usually because of a bug in the interpreter), it's always better to not try and optimize your code against a specific version of V8. V8 changes over time and these things will improve. If you optimize for a quirk in V8, you end up loosing as V8 will later fix the performance degradation and your code is now slower than it could have been if you didn't try to optimize it. Just my 5 cents 😅

integration-tests/helpers/fake-agent.js
watson: I'll remote it

integration-tests/helpers.js
watson: I'll remove it in an upcoming commit 👍

integration-tests/helpers/fake-agent.js
watson: I'll remove it in an upcoming commit 👍

integration-tests/helpers/fake-agent.js
watson: No we don't. I'll remote it in an upcoming commit 👍

integration-tests/helpers/fake-agent.js
watson: Fixed 👍

integration-tests/helpers/fake-agent.js
watson: Fixed 👍

integration-tests/helpers/fake-agent.js
watson: I see now that I misunderstood your original question. I've fixed it now 👍

integration-tests/helpers/fake-agent.js
watson: Fixed 👍

integration-tests/helpers.js
watson: Fixed 👍

watson:
FYI, I've updated the https://github.com/DataDog/dd-trace-js/pull/4492 PR with usage of `FakeAgent#updateRemoveConfig` and `FakeAgent#removeRemoveConfig` (look for the file integration-tests/debugger/index.spec.js)

integration-tests/helpers/fake-agent.js
```diff
@@ -0,0 +1,298 @@
...
+
+    for (const { apply_error: error } of state.config_states) {
+      if (error) {
+        // Print the error sent by the client in case it's useful in debugging tests
+        console.error(error) // eslint-disable-line no-console
+      }
+    }
+
+    if (agent._rcTargetsVersion === state.targets_version) {
+      // If the state hasn't changed since the last time the client asked, hyst return an empty result
```
simon-id, line 227:
```suggestion
      // If the state hasn't changed since the last time the client asked, just return an empty result
```

simon-id (approved):

bengl (approved):

PR #4633, opened 2024-08-27
speed up shimmer by about 50x

bengl:
The `copyProperties` function was doing an `Object.setPrototypeOf`, which is rather costly. We don't actually need this because all it was doing was acting as a stopgap in case we don't get all the properties from the original function. We're using `Reflect.ownKeys()` to get those properties, so this can't happen, therefore there's no need to set the prototype.

On a benchmark I whipped up separately for another project, I had noticed that shimmer + our instrumentations adds a significant amount of overhead. On a run with 100 iterations, I found that overhead to be about 10000%. When I ran the same benchmark after this change, I saw that overhead to be about 200%. To be fair, this benchmark is not a realistic application loading, and is explicitly clearing the cache and re-loading everything repeatedly, so I'm not sure how impactful this change will be. That said, we'll see what the benchmarks say.

### DIFF
--- a/packages/datadog-shimmer/src/shimmer.js
+++ b/packages/datadog-shimmer/src/shimmer.js
@@ -6,8 +6,6 @@ const log = require('../../dd-trace/src/log')
 const unwrappers = new WeakMap()
 
 function copyProperties (original, wrapped) {
-  Object.setPrototypeOf(wrapped, original)
-
   const props = Object.getOwnPropertyDescriptors(original)
   const keys = Reflect.ownKeys(props)
 


tlhunter (approved):

Qard (approved):

rochdev:
I think this was mostly added as a safety net, but I don't know the scenarios where the prototype would not be accessible by `getOwnPropertyDescriptors`. However, I cannot reproduce this performance gain at all, and running quick benchmark locally result in the exact same timing. My assumption is that there is something specific in your setup where maybe the wrapping stays in place and gets reapplied on top of the previous wrapping or something similar, which would explain the exponential impact of running the benchmark longer. I think it would be worth it to find exactly why this is happening just in case there is some other bug that this either this is fixing or that would be present in the benchmark. In any case, I don't think it's needed, so it should be safe to remove.

rochdev (approved):

bengl:
FWIW my benchmark was effectively:

```js
const origRequire = Module.prototype.require

for (let i = 0; i < SOME_BIG_ENOUGH_NUMBER; i++) {
  require('dd-trace/packages/datadog-instrumentations')
  startTimer()
  require('undici')
  stopTime()
  clearCache()
  Module.prototype.require = origRequire
}
```

When I ran it with 0x, I saw that nearly all the time was being spent in `setPrototypeOf`.

> I don't know the scenarios where the prototype would not be accessible by `getOwnPropertyDescriptors`.

We don't even need the prototype since we're copying all the `ownKeys()` and these are all functions. On the off chance that we're wrapping a function whose prototype is _not_ `Function.prototype`, then we'd have a problem. If you want, I can add a check for that and special-case it. It should be pretty quick.

rochdev:
> If you want, I can add a check for that and special-case it.

If you don't think it's a valid real-world case I don't really mind. I was more interested to know why rerunning shimmer many times would cause an exponential delay in the benchmark, whereas running the function directly on a new object every time shows no improvement.

PR #4634, opened 2024-08-27
wrap all callbacks in instrumentations with wrapFunction

bengl:
These were the remaining areas with non-trivial code that might not be caught by try/catches when our code throws an exception. Adding `wrapFunction` adds that try/catch appropriately, but only when SSI is enabled. Otherwise these are all passthroughs. In the future, we may enable those try/catches at all times.

I considered the following to be out of scope:

* Many (but not all) completely trivial wrappers.
* Exceptions that might originate in `async_hooks`.

I may have missed some. It's an extremely difficult task to verify that all opportunities for throwing an exception are covered.

### DIFF
--- a/packages/datadog-instrumentations/src/grpc/client.js
+++ b/packages/datadog-instrumentations/src/grpc/client.js
@@ -88,10 +88,10 @@ function wrapMethod (method, path, type, hasPeer) {
     return method
   }
 
-  const wrapped = function () {
+  const wrapped = shimmer.wrapFunction(method, method => function () {
     const args = ensureMetadata(this, arguments, 1)
     return callMethod(this, method, args, path, args[1], type, hasPeer)
-  }
+  })
 
   Object.assign(wrapped, method)
 
@@ -101,7 +101,7 @@ function wrapMethod (method, path, type, hasPeer) {
 }
 
 function wrapCallback (ctx, callback = () => { }) {
-  return function (err) {
+  return shimmer.wrapFunction(callback, callback => function (err) {
     if (err) {
       ctx.error = err
       errorChannel.publish(ctx)
@@ -111,7 +111,7 @@ function wrapCallback (ctx, callback = () => { }) {
       return callback.apply(this, arguments)
       // No async end channel needed
     })
-  }
+  })
 }
 
 function createWrapEmit (ctx, hasPeer = false) {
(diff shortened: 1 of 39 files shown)


Qard (dismissed):

packages/datadog-instrumentations/src/winston.js
```diff
@@ -43,12 +43,12 @@ function wrapMethod (method, logCh) {
         if (patched.has(transport) || typeof transport.log !== 'function') continue
 
         const log = transport.log
-        transport.log = function wrappedLog (level, msg, meta, callback) {
+        transport.log = shimmer.wrapFunction(log, log => function wrappedLog (level, msg, meta, callback) {
```
simon-id, line 46: shouldn't this be 
```javascript
shimmer.wrap(transport, 'log', ...)
```
instead ?

packages/datadog-instrumentations/src/mongoose.js
```diff
@@ -128,22 +128,21 @@ addHook({
                     const resolve = arguments[0]
                     const reject = arguments[1]
 
-                    // not using shimmer here because resolve/reject could be empty
-                    arguments[0] = function wrappedResolve () {
```
simon-id, line 132: it says it doesn't use shimmer for a specific reason here, is that handled correctly by shimmer now ?

packages/datadog-instrumentations/src/winston.js
bengl: It probably always should have been `shimmer.wrap()` yeah. I can change this one.

bengl: Done.

packages/datadog-instrumentations/src/grpc/client.js
```diff
@@ -88,10 +88,10 @@ function wrapMethod (method, path, type, hasPeer) {
...
   }
 
-  const wrapped = function () {
+  const wrapped = shimmer.wrapFunction(method, method => function () {
     const args = ensureMetadata(this, arguments, 1)
     return callMethod(this, method, args, path, args[1], type, hasPeer)
-  }
+  })
 
   Object.assign(wrapped, method)
```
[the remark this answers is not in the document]
bengl: Fixed.

simon-id (approved):

bengl:
The two failing are all-green and next.js, so I'll just go ahead and merge.

PR #4636, opened 2024-08-28
next.js: cap test range to 14.2.6 for now

tlhunter:
### What does this PR do?
- caps tested next versions

### Motivation
- recent change broke CI

### DIFF
--- a/.github/workflows/appsec.yml
+++ b/.github/workflows/appsec.yml
@@ -209,7 +209,6 @@ jobs:
         version:
           - 18
           - 22.6
-        range: ['9.5.0', '11.1.4', '13.2.0', '*']
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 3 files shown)


RohitK89 (dismissed):

m-sanders:
I recently upgraded to 14.2.7+ and dd-trace is not working as expected in production. 

Are you sure this isn't a bug rather than an issue with the integration?

I have added logging to the dd-trace next plugin code and it appears that the 'apm:next:page:load'` event no longer fires, resulting in the `pageLoad` method not being triggered and the spans no longer being tagged with `resource.name` and `next.page`. 

I have also noticed an elevated amount of time being attributed to the Next.js-based service in the time graph but no discernable difference in latency or degradation of user experience. Profiling levels look the same. I have since reverted back to the previous version.

PR #4638, opened 2024-08-28
Allow Remote Config acknowledgements to be async

watson:
## Old API

Remote Config updates coming from the Agent needs to be acknowledged. Currently this is done synchronously in the Node.js tracer:

If a listener is attached to a given product, any config updates to that product is automatically acknowledged (if the listener doesn't throw, the state is set to `ACKNOWLEDGED` and if it throws, it's set to `ERROR`).

However, just because there's a listener and it doesn't throw, doesn't mean that the received config can be processed by the listener. It could be that the listener needs to do some async things before it can determine if the received config is valid.

## New API

~This PR changes the existing behavior by making the acknowledgement async: A 4th argument is emitted to listeners of `rc.on(product)` which is an `ack` callback which should be called once the listener wants to acknowledge the received config. It can optionally be called with an error object as its only argument to set the state to `ERROR` instead of `ACKNOWLEDGED`.~

**Update:**

This PR changes the existing behavior by optionally allowing the acknowledgement to be async. This is achieved by:

- Changing the way you register products from using the `EventEmitter` API to a custom API that only allows a single handler per product (event emitters can have more than one listener, which is problematic as we can't support more than one of the listeners to acknowledge the config). The new API is as follows:
  - `rc.setProductHandler(product, handler)` (prevously `rc.on(product, handler)`)
  - `rc.removeProductHandler(product)` (prevously `rc.off(product, handler)`)
- The new product handler is called similar to the old event listener, with the following exception: A new optional 4th argument is supplied called `ack`. This is a callback which can be called (sync or async) either without any arguments (to set the state to `ACKNOWLEDGED`) or with a single error argument (to set the state to `ERROR`).
  - If the handler function signature takes less than 4 arguments or doesn't use the rest operator (`...args`), and...
    - ...the handler doesn't return a `Promise`: The state is set to `ACKNOWLEDGED` right away without the handler having to do anything (existing behavior)
    - ...the handler returns a `Promise`, we wait until the promise is resolved or rejected and set the state accordingly (new behavior)
  - If the handler function signature takes 4 or more arguments or use the rest operator (`...args`): The state is left as `UNACKNOWLEDGED` until the `ack` callback is called (new behavior)
  - In any case, the state is still set to `ERROR` if the handler throws (existing behavior)

The `RemoteConfigManager` is still an `EventEmitter` however because the `kPreUpdate` symbol is still being emitted. That part could do with a complete refactor, but is out of scope for this PR.

### DIFF
--- a/packages/dd-trace/src/appsec/remote_config/index.js
+++ b/packages/dd-trace/src/appsec/remote_config/index.js
@@ -28,8 +28,7 @@ function enable (config, appsec) {
       rc.updateCapabilities(RemoteConfigCapabilities.ASM_API_SECURITY_SAMPLE_RATE, true)
     }
 
-    rc.on('ASM_FEATURES', (action, rcConfig, id, ack) => {
-      ack()
+    rc.setProductHandler('ASM_FEATURES', (action, rcConfig) => {
       if (!rcConfig) return
 
       if (activation === Activation.ONECLICK) {
@@ -77,9 +76,9 @@ function enableWafUpdate (appsecConfig) {
     rc.updateCapabilities(RemoteConfigCapabilities.ASM_CUSTOM_BLOCKING_RESPONSE, true)
     rc.updateCapabilities(RemoteConfigCapabilities.ASM_TRUSTED_IPS, true)
 
-    rc.on('ASM_DATA', ackNoop)
-    rc.on('ASM_DD', ackNoop)
-    rc.on('ASM', ackNoop)
+    rc.setProductHandler('ASM_DATA', noop)
+    rc.setProductHandler('ASM_DD', noop)
+    rc.setProductHandler('ASM', noop)
 
     rc.on(RemoteConfigManager.kPreUpdate, RuleManager.updateWafFromRC)
   }
@@ -99,15 +98,14 @@ function disableWafUpdate () {
     rc.updateCapabilities(RemoteConfigCapabilities.ASM_CUSTOM_BLOCKING_RESPONSE, false)
     rc.updateCapabilities(RemoteConfigCapabilities.ASM_TRUSTED_IPS, false)
 
-    rc.off('ASM_DATA', noop)
-    rc.off('ASM_DD', noop)
-    rc.off('ASM', noop)
+    rc.removeProductHandler('ASM_DATA')
+    rc.removeProductHandler('ASM_DD')
+    rc.removeProductHandler('ASM')
 
     rc.off(RemoteConfigManager.kPreUpdate, RuleManager.updateWafFromRC)
   }
 }
 
-function ackNoop (a, b, c, ack) { ack() }
 function noop () {}
 
 module.exports = {
--- a/packages/dd-trace/src/appsec/remote_config/index.js
+++ b/packages/dd-trace/src/appsec/remote_config/index.js
@@ -76,6 +76,7 @@ function enableWafUpdate (appsecConfig) {
     rc.updateCapabilities(RemoteConfigCapabilities.ASM_CUSTOM_BLOCKING_RESPONSE, true)
     rc.updateCapabilities(RemoteConfigCapabilities.ASM_TRUSTED_IPS, true)
 
+    // TODO: delete noop handlers and kPreUpdate and replace with batched handlers
     rc.setProductHandler('ASM_DATA', noop)
     rc.setProductHandler('ASM_DD', noop)
     rc.setProductHandler('ASM', noop)
--- a/packages/dd-trace/src/appsec/remote_config/index.js
+++ b/packages/dd-trace/src/appsec/remote_config/index.js
@@ -28,7 +28,8 @@ function enable (config, appsec) {
       rc.updateCapabilities(RemoteConfigCapabilities.ASM_API_SECURITY_SAMPLE_RATE, true)
     }
 
-    rc.on('ASM_FEATURES', (action, rcConfig) => {
+    rc.on('ASM_FEATURES', (action, rcConfig, id, ack) => {
+      ack()
       if (!rcConfig) return
 
       if (activation === Activation.ONECLICK) {
@@ -76,9 +77,9 @@ function enableWafUpdate (appsecConfig) {
     rc.updateCapabilities(RemoteConfigCapabilities.ASM_CUSTOM_BLOCKING_RESPONSE, true)
     rc.updateCapabilities(RemoteConfigCapabilities.ASM_TRUSTED_IPS, true)
 
-    rc.on('ASM_DATA', noop)
-    rc.on('ASM_DD', noop)
-    rc.on('ASM', noop)
+    rc.on('ASM_DATA', ackNoop)
+    rc.on('ASM_DD', ackNoop)
+    rc.on('ASM', ackNoop)
 
     rc.on(RemoteConfigManager.kPreUpdate, RuleManager.updateWafFromRC)
   }
@@ -106,6 +107,7 @@ function disableWafUpdate () {
   }
 }
 
+function ackNoop (a, b, c, ack) { ack() }
 function noop () {}
 
 module.exports = {
--- a/packages/dd-trace/src/appsec/remote_config/manager.js
+++ b/packages/dd-trace/src/appsec/remote_config/manager.js
@@ -37,10 +37,19 @@ class RemoteConfigManager extends EventEmitter {
 
     this.state = {
       client: {
-        state: { // updated by `parseConfig()`
+        state: { // updated by `parseConfig()` and `poll()`
           root_version: 1,
           targets_version: 0,
-          config_states: [],
+          // Use getter so `apply_*` can be updated async and still affect the content of `config_states`
+          get config_states () {
+            return Array.from(appliedConfigs.values()).map((conf) => ({
+              id: conf.id,
+              version: conf.version,
+              product: conf.product,
+              apply_state: conf.apply_state,
+              apply_error: conf.apply_error
+            }))
+          },
           has_error: false,
           error: '',
           backend_client_state: ''
@@ -63,7 +72,7 @@ class RemoteConfigManager extends EventEmitter {
     }
 
     this._handlers = new Map()
-    this.appliedConfigs = new Map()
+    const appliedConfigs = this.appliedConfigs = new Map()
   }
 
   updateCapabilities (mask, value) {
@@ -222,21 +231,9 @@ class RemoteConfigManager extends EventEmitter {
       this.dispatch(toApply, 'apply')
       this.dispatch(toModify, 'modify')
 
-      // TODO: Possible race condition: If this property is overwritten before a product handler has time to update the
-      // `apply_state` of the previous call to `parseConfig`, the future state update will be ignored.
-      this.state.client.state.config_states = []
       this.state.cached_target_files = []
 
       for (const conf of this.appliedConfigs.values()) {
-        this.state.client.state.config_states.push({
-          id: conf.id,
-          version: conf.version,
-          product: conf.product,
-          // Use getters so `conf.apply_*` can be updated async and still affect the new object pushed to config_states
-          get apply_state () { return conf.apply_state },
-          get apply_error () { return conf.apply_error }
-        })
-
         this.state.cached_target_files.push({
           path: conf.path,
           length: conf.length,
--- a/packages/dd-trace/src/appsec/remote_config/manager.js
+++ b/packages/dd-trace/src/appsec/remote_config/manager.js
@@ -222,6 +222,8 @@ class RemoteConfigManager extends EventEmitter {
       this.dispatch(toApply, 'apply')
       this.dispatch(toModify, 'modify')
 
+      // TODO: Possible race condition: If this property is overwritten before a product handler has time to update the
+      // `apply_state` of the previous call to `parseConfig`, the future state update will be ignored.
       this.state.client.state.config_states = []
       this.state.cached_target_files = []
 
--- a/packages/dd-trace/src/appsec/remote_config/manager.js
+++ b/packages/dd-trace/src/appsec/remote_config/manager.js
@@ -244,7 +244,7 @@ class RemoteConfigManager extends EventEmitter {
       // TODO: we need a way to tell if unapply configs were handled by kPreUpdate or not, because they're always
       // emitted unlike the apply and modify configs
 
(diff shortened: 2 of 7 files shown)


watson:
* **#4549** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4549?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4492** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4492?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4638** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4638?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

Qard (approved):

Qard (approved):

Qard (dismissed):

packages/dd-trace/src/appsec/remote_config/index.js
```diff
@@ -76,9 +76,9 @@ function enableWafUpdate (appsecConfig) {
     rc.updateCapabilities(RemoteConfigCapabilities.ASM_CUSTOM_BLOCKING_RESPONSE, true)
     rc.updateCapabilities(RemoteConfigCapabilities.ASM_TRUSTED_IPS, true)
 
-    rc.on('ASM_DATA', noop)
-    rc.on('ASM_DD', noop)
-    rc.on('ASM', noop)
+    rc.setProductHandler('ASM_DATA', noop)
```
simon-id, line 79:
```suggestion
    // TODO: delete noop handlers and kPreUpdate and replace with batched handlers
    rc.setProductHandler('ASM_DATA', noop)
```

packages/dd-trace/src/appsec/remote_config/manager.js
```diff
@@ -61,7 +71,8 @@ class RemoteConfigManager extends EventEmitter {
       cached_target_files: [] // updated by `parseConfig()`
     }
 
-    this.appliedConfigs = new Map()
+    this._handlers = new Map()
+    const appliedConfigs = this.appliedConfigs = new Map()
```
simon-id, line 75: wow, i know it gets hoisted but, use before declaration is weeeeeird, we can just move it above this.clients ?

packages/dd-trace/src/appsec/remote_config/manager.js
```diff
@@ -228,18 +231,9 @@ class RemoteConfigManager extends EventEmitter {
       this.dispatch(toApply, 'apply')
       this.dispatch(toModify, 'modify')
 
-      this.state.client.state.config_states = []
       this.state.cached_target_files = []
 
       for (const conf of this.appliedConfigs.values()) {
```
simon-id, line 236: this can just be a .map() now

packages/dd-trace/test/appsec/index.spec.js
```diff
@@ -1182,6 +1182,7 @@ describe('IP blocking', function () {
         }).then(() => {
           throw new Error('Not expected')
         }).catch((err) => {
+          expect(err.message).to.not.equal('Not expected')
```
simon-id, line 1185: unrelated change ? what made me you change that ?

packages/dd-trace/src/appsec/remote_config/manager.js
```diff
@@ -256,16 +250,42 @@ class RemoteConfigManager extends EventEmitter {
...
       if (item.apply_state === UNACKNOWLEDGED || action === 'unapply') {
-        try {
-          // TODO: do we want to pass old and new config ?
-          const hadListeners = this.emit(item.product, action, item.file, item.id)
-
-          if (hadListeners) {
-            item.apply_state = ACKNOWLEDGED
+        const handler = this._handlers.get(item.product)
+
+        if (handler) {
```
simon-id, line 255: wdyt about a continue to avoid an indentation level ?
```suggestion
        if (!handler) continue
```
or you could also have a `_handlers.has(item.product)` directly in the if case above

packages/dd-trace/src/appsec/remote_config/manager.js
```diff
@@ -256,16 +250,42 @@ class RemoteConfigManager extends EventEmitter {
...
+            if (supportsAckCallback(handler)) {
+              // If the handler accepts an `ack` callback, expect that to be called and set `apply_state` accordinly
+              // TODO: do we want to pass old and new config ?
+              handler(action, item.file, item.id, (err) => {
+                if (err) {
+                  item.apply_state = ERROR
+                  item.apply_error = err.toString()
+                } else if (item.apply_state !== ERROR) {
+                  item.apply_state = ACKNOWLEDGED
+                }
```
simon-id, line 266: could do with some code dedup here, we set handler errors for apply states in 3 different places (cb, promise, catch). A function like this:
```javascript
setApplyState (item, err) {
  if (item.apply_state !== UNACKNOWLEDGED) return

  if (err) {
    item.apply_state = ERROR
    item.apply_error = err.toString()
  } else {
    item.apply_state = ACKNOWLEDGED
  }
}
```

packages/dd-trace/src/appsec/remote_config/manager.js
```diff
@@ -299,4 +319,22 @@ function parseConfigPath (configPath) {
   }
 }
 
+function supportsAckCallback (handler) {
+  if (kSupportsAckCallback in handler) return handler[kSupportsAckCallback]
```
simon-id, line 323: you're caching the result/allowing overriding right ?

packages/dd-trace/src/appsec/remote_config/manager.js
```diff
@@ -299,4 +319,22 @@ function parseConfigPath (configPath) {
...
+
+  const numOfArgs = handler.length
+  let result = false
+
+  if (numOfArgs >= 4) {
+    result = true
+  } else if (numOfArgs !== 0) {
+    const source = handler.toString()
+    result = source.slice(0, source.indexOf(')')).includes('...')
+  }
```
simon-id, line 333: oh wow that's cursed, whould anyone even want to use spread operator anyway ? like do we really need that special case ?

packages/dd-trace/src/appsec/remote_config/manager.js
```diff
@@ -256,16 +250,42 @@ class RemoteConfigManager extends EventEmitter {
...
+            }
+          } catch (err) {
+            item.apply_state = ERROR
+            item.apply_error = err.toString()
           }
-        } catch (err) {
-          item.apply_state = ERROR
-          item.apply_error = err.toString()
         }
       }
```
simon-id, line 291: also (can't see the code but it's right below this line), configs will be deleted before the async handler returns (and no way to report if it was deleted correctly), is that a problem ?

packages/dd-trace/test/appsec/remote_config/manager.spec.js
```diff
@@ -149,28 +149,27 @@ describe('RemoteConfigManager', () => {
...
       expect(rc.state.client.products).to.deep.equal(['ASM_FEATURES'])
       expect(rc.scheduler.start).to.have.been.calledOnce
 
-      rc.on('ASM_DATA', noop)
-      rc.on('ASM_DD', noop)
+      rc.setProductHandler('ASM_DATA', noop)
+      rc.setProductHandler('ASM_DD', noop)
 
       expect(rc.state.client.products).to.deep.equal(['ASM_FEATURES', 'ASM_DATA', 'ASM_DD'])
-      expect(rc.scheduler.start).to.have.been.calledThrice
```
simon-id, line 163: why remove that line ? can update it to .calledOnce instead ?

Qard (dismissed):

packages/dd-trace/src/appsec/remote_config/manager.js
watson: Yes, checking for the rest-operator unfortunately requires us to read the function source code, so to now have to do that more than once, I cache the result.

Not sure what you mean by the last part of your question ("allowing overriding")?

packages/dd-trace/src/appsec/remote_config/manager.js
watson: Good idea. Btw, is there a specific reason you used `err.toString()` instead of `err.message`?

packages/dd-trace/src/appsec/remote_config/manager.js
watson: > wdyt about a continue to avoid an indentation level ?

I can't use `continue` as that would skip the last part of the the for loop, not only the nested if-statement.

> or you could also have a _handlers.has(item.product) directly in the if case above

I need the actual `handler` function so that I can call it inside the if-statement, so using `has()` would be an unnecessary function-call as I already have retrieved the `handler` a few lines above.

packages/dd-trace/test/appsec/index.spec.js
watson: Yes, it's unrelated. During the development of this PR I ran into an issue where these tests failed with a different error than expected. But since the `err.message` was never checked, it wasn't clear why it failed. By adding this check it's going to be much easier to debug if these tests ever fail in the future. At the time I also had some changes in this file which was actually related to the PR, but those were later dropped. But I kept this in, because it's an actual improvement to the developer experience

packages/dd-trace/src/appsec/remote_config/manager.js
watson: We don't currently use it or have a need for it, but if we don't have this check, the code is going to behave unexpected if someone at some point decides to use it. So writing it this way is safer

packages/dd-trace/src/appsec/remote_config/manager.js
simon-id: well it's not really meant for it, but if i get a hold of the symbol, i can force the RC manager to consider a handler supporting ack or not, might be useful one day. Resolving the thread...

packages/dd-trace/src/appsec/remote_config/manager.js
simon-id: No super specific reason, but toString() is supposed to be the standard interface that returns how the object knows it wants to serialize itself, by picking a specific field, you don't allow those errors to decide themselves. Any reason you want to use message instead ?

packages/dd-trace/test/appsec/remote_config/manager.spec.js
watson: Besides knowing that the timer has been started, this test shouldn't need to care about how many times it's called. That's an implementation detail.

All this test should care about is that a timer is running, which it knows because of the line `expect(rc.scheduler.start).to.have.been.calledOnce` above (which actually could just be `expect(rc.scheduler.start).to.have.been.called` since we don't care about the amount)

packages/dd-trace/test/appsec/index.spec.js
simon-id: just out of curiosity what was the problem that made this test throw ?

packages/dd-trace/src/appsec/remote_config/manager.js
simon-id: Right, but if you fail to remove it, then what? do you add it back in the config states ? or do you silently "forget" about it ?

packages/dd-trace/test/appsec/remote_config/manager.spec.js
simon-id: Ah yeah sorry didn't look above, there is another check :+1:

packages/dd-trace/src/appsec/remote_config/manager.js
watson: `err.message` is the standard way to get the error message if that's all we care about. If we want to allow the error to give us more than the error message, then we can of course use `toString()`. I just fear that some errors might return too much data with the `toString()` call and all we care about is just the error message itself. But I'm fine either way 🤷

packages/dd-trace/test/appsec/index.spec.js
watson: It somehow ended up in the `throw new Error('Not expected')` then-block, but I can't remember why right now. And it took me a while to figure out that this is what was wrong because it just failed with `Uncaught TypeError: Cannot read properties of undefined (reading 'status')`

watson: ... as in I got an unexpected exception in the test-code itself, not an assertion failure. This extra check means that the test fails properly due an assertion error, instead of the test-code failing

packages/dd-trace/src/appsec/remote_config/manager.js
watson: Hmm after implementing your suggested change, I'm not sure I like it as much as I thought I would. You can see the diff here:

image

I think it's harder to read and it's actually one line more than the original implementation. However, I still agree that this big nested if-block isn't nice. So I would rather want to move the entire block somewhere else and just replace it with a function call, eg:

image

packages/dd-trace/src/appsec/remote_config/manager.js
simon-id: Irrelevant as the new code doesn't have the problem anymore

packages/dd-trace/src/appsec/remote_config/manager.js
watson: Conclusion after discussing this over Zoom: Yes, we could technically keep the item in the local tracer-state instead of deleting it if an error occurred when un-applying it. But that wouldn't put us in any better position: We would not be able to resolve the error automatically and try again later - we would just be in an unknown state. So it's better to give up early, log the error, and just delete the item as we do now.

packages/dd-trace/src/appsec/remote_config/manager.js
```diff
@@ -254,20 +244,7 @@ class RemoteConfigManager extends EventEmitter {
...
-
-          if (hadListeners) {
-            item.apply_state = ACKNOWLEDGED
-          }
-        } catch (err) {
-          item.apply_state = ERROR
-          item.apply_error = err.toString()
-        }
-      }
+      callHandlerFor.call(this, action, item)
```
simon-id, line 247: i mean if you need `this` at this point just make it a method of the class no ?

watson: Fixed 👍

simon-id (approved):

PR #4644, opened 2024-08-30
fix(openai): include a workaround in `register.js` to resolve ESM + OpenAI errors

sabrenner:
### What does this PR do?
Introduces a workaround to not produce errors when using `openai` with ESM and the `dd-trace/register.js` import.

**Note:** This does not fix _tracing_. This is a temporary workaround to resolve import errors when using `dd-trace` + `import-in-the-middle` + `openai`. A real fix will need investigation and implementation in `import-in-the-middle`

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/register.js
+++ b/register.js
@@ -1,4 +1,4 @@
 const { register } = require('node:module')
 const { pathToFileURL } = require('node:url')
 
-register('./loader-hook.mjs', pathToFileURL(__filename))
+register('./loader-hook.mjs', pathToFileURL(__filename), { data: { include: ['openai'] } })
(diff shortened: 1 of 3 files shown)


register.js
```diff
@@ -1,4 +1,4 @@
 const { register } = require('node:module')
 const { pathToFileURL } = require('node:url')
 
-register('./loader-hook.mjs', pathToFileURL(__filename))
+register('./loader-hook.mjs', pathToFileURL(__filename), { data: { include: ['openai'] } })
```
sabrenner, line 4: it looks like this breaks other things, ie `vitest` integration tests are failing ... i'm not 100% on this approach, or if there's a different way I should be specifying this

sabrenner:
closing in favor of #5267

PR #4647, opened 2024-09-03
clearer profiling error

bengl:
Indicating that a backoff has expired isn't enough information to diagnose what's causing the backoff in the first place, so rather than creating a new error object, this change just passes the last error along.

### DIFF
--- a/packages/dd-trace/src/profiling/exporters/agent.js
+++ b/packages/dd-trace/src/profiling/exporters/agent.js
@@ -199,7 +199,7 @@ class AgentExporter {
             this._logger.error(`Error from the agent: ${err.message}`)
             return
           } else if (err) {
-            reject(new Error('Profiler agent export back-off period expired'))
+            reject(err)
             return
           }
 
--- a/packages/dd-trace/test/profiling/exporters/agent.spec.js
+++ b/packages/dd-trace/test/profiling/exporters/agent.spec.js
@@ -270,7 +270,7 @@ describe('exporters/agent', function () {
       try {
         await exporter.export({ profiles, start, end, tags })
       } catch (err) {
-        expect(err.message).to.match(/^Profiler agent export back-off period expired$/)
+        expect(err.message).to.match(/^HTTP Error 500$/)
         failed = true
       }
       expect(failed).to.be.true


szegedi (dismissed):

nsavoire (dismissed):

nsavoire:
Unit tests probably need to be updated

Qard (dismissed):

Qard (approved):

tlhunter:
clearer

packages/dd-trace/src/profiling/exporters/agent.js
```diff
@@ -199,7 +199,7 @@ class AgentExporter {
             this._logger.error(`Error from the agent: ${err.message}`)
             return
           } else if (err) {
-            reject(new Error('Profiler agent export back-off period expired'))
```
tlhunter, line 202: Isn't there like a convention to combine/chain errors? Something like this:

```js
            const error = new Error('Profiler agent export back-off period expired')
            error.reason = err
            reject(error)
```

tlhunter: Or even this:
```js
reject(new Error('Profiler agent export back-off period expired: ' + err.message))
// Profiler agent export back-off period expired: HTTP Error 500
```
Though that might ruin the cardinality or error messages and undermine the problem you're trying to solve...

Qard: There's [error.cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause), but I'm not sure it's supported across our full version range.

watson: If we don't use the custom syntax:

```js
new Error('parent', { cause: new Error('child') })
```

But instead build it manually:

```js
const parent = new Error('parent')
const child = new Error('child')
parent.cause = child
```

Then the Node.js version being used only matters if we care how `console.log` outputs the error - which I guess we don't. I assume we just need to make sure that the `log.error` function knows to also take into account any linked `cause` error.

This is something we should make sure regardless of this PR as other libraries might make use of the `cause` property and we need to properly capture that information.

Gotcha: There is a slight different, between how `console.log` outputs an error created with the two approaches listed above. If using the options object with the `cause` property that's only supported in Node.js 16.9+, then the stack trace outputted looks like this:

```
Error: parent
    at REPL53:1:5
    at ContextifyScript.runInThisContext (node:vm:136:12)
    ... 7 lines matching cause stack trace ...
    at [_line] [as _line] (node:internal/readline/interface:887:18) {
  [cause]: Error: child
      at REPL53:1:34
      at ContextifyScript.runInThisContext (node:vm:136:12)
      at REPLServer.defaultEval (node:repl:598:22)
      at bound (node:domain:432:15)
      at REPLServer.runBound [as eval] (node:domain:443:12)
      at REPLServer.onLine (node:repl:927:10)
      at REPLServer.emit (node:events:532:35)
      at REPLServer.emit (node:domain:488:12)
      at [_onLine] [as _onLine] (node:internal/readline/interface:416:12)
      at [_line] [as _line] (node:internal/readline/interface:887:18)
}
```

If we just naively add a `cause` property on an error object after creation-time, the `console.log` output looks like this:

```
Error: parent
    at REPL31:1:5
    at ContextifyScript.runInThisContext (node:vm:136:12)
    ... 7 lines matching cause stack trace ...
    at [_line] [as _line] (node:internal/readline/interface:887:18) {
  cause: Error: child
      at REPL32:1:5
      at ContextifyScript.runInThisContext (node:vm:136:12)
      at REPLServer.defaultEval (node:repl:598:22)
      at bound (node:domain:432:15)
      at REPLServer.runBound [as eval] (node:domain:443:12)
      at REPLServer.onLine (node:repl:927:10)
      at REPLServer.emit (node:events:532:35)
      at REPLServer.emit (node:domain:488:12)
      at [_onLine] [as _onLine] (node:internal/readline/interface:416:12)
      at [_line] [as _line] (node:internal/readline/interface:887:18)
}
```

Notice the tiny difference between `[cause]` and `cause`. I'm not sure if that's important, but it means that there's a slight difference in the implementation.

Qard: If we sub-class our errors we could have it version detect and do `super(message, options)` where supported and do `super(message); this.cause = options.cause` where not supported.

bengl: I think it's moot in this case because I don't think the wrapper/parent error adds any value here.

PR #2495, opened 2022-11-03
Add scheduling for serverless profiling

amymin00:
### What does this PR do?
Collects 65 one-second intervals before flushing rather than profiling over a 65s period.

Closes https://github.com/DataDog/datadog-lambda-js/issues/252

### Motivation
Serverless functions often have short runtimes and may not run for much of the 65s profiling period. We want to collect profiling data only when the function's running to avoid flushing mostly empty profiles.

PR #4652, opened 2024-09-05
[PROF-10484] refactor: use standard profiler in serverless contexts

wincent:
### What does this PR do?

This PR removes the implementation of the `ServerlessProfiler`, because its behavior was effectively identical to that of the standard `Profiler`. It keeps the old constant around, just in case anybody is importing and using it, but now it's just a reference to the base `Profiler`.

Lots of whitespace differences here in the tests because I removed an outer `describe()`, so [view with whitespace off](https://github.com/DataDog/dd-trace-js/pull/4652/files?w=1) to get a clearer picture of what actually changed.

### Motivation

See the original description below for my initial investigation, when I added a failing test to reveal the failure of the serverless scheduler to actually do anything different from the non-serverless one. I could have kept the scheduler around and fixed it, but it seems better to remove it. When a Lambda function is frozen, any timers will get frozen too, so the base `Profiler` class will work just fine (instead of waiting 65 ⨉ 1-second chunks, it can just wait a single 65-second chunk with the same effect).

<details>
 <summary>Original description</summary>

In a non-serverless environment, the profiler runs for 65 seconds, then we grab the profile data and submit it.

In a serverless context, we do something different, using a different scheduler. Serverless functions may run for a short interval before being suspended (or frozen, or paused, or whatever you want to call it) and then be resumed. The intent is for us to grab profile data in smaller chunks of 1 second each. After 65 chunks, we submit to the server. I'm not sure of exactly _why_ we want to do it like that, because in practice we should end up obtaining the same profile data (the `profiler.profile()` call takes as parameters the start and end date that we care about, along with a `restart` parameter if we want the profiler to continue sampling; the fact that our timers may pause and resume should have no impact on the timestamps that get passed into the profiler).

Comments in Slack suggest that the intent of the original PR:

- https://github.com/DataDog/dd-trace-js/pull/2495

was:

> if a function is called once and runs for 1s, then frozen for a minute until the next call, the profiler would flush a very small profile
>
> so instead the idea was to collect/merge the 65 profiles over the course of multiple calls

However:

1. What the current scheduler is actually doing is just setting 65 1-second timers in a row and then grabbing all the profile data and submitting it. This commit adds a failing test to show that this is the case (the existing test only asserted that we "ticked" 65 times before flushing, not that we collected 65 1-second chunks of profiling data before flushing).
2. It's not clear to me that the scheduler is needed at all. In its current form, it is behaving basically equivalent to the non-serverless scheduler. But I don't think a scheduler that did what that PR intended to do would stop us from submitting small profiles anyway: if a Lambda is frozen (along with its timers) and resumed, we'd still only submit after 65 seconds of actual profiling. And if the Lambda is stopped (gracefully as opposed to being unceremoniously terminated), then our `_stop()` logic will run which cleans up the timers and does not flush. (The only time we eagerly submit before the time is up is if our `_nearOOMExport()` method gets triggered via a callback.)
3. 
</details>

### Additional Notes

Nope.

### DIFF
--- a/packages/dd-trace/src/profiling/index.js
+++ b/packages/dd-trace/src/profiling/index.js
@@ -1,13 +1,13 @@
 'use strict'
 
-const { Profiler, ServerlessProfiler } = require('./profiler')
+const { Profiler } = require('./profiler')
 const WallProfiler = require('./profilers/wall')
 const SpaceProfiler = require('./profilers/space')
 const { AgentExporter } = require('./exporters/agent')
 const { FileExporter } = require('./exporters/file')
 const { ConsoleLogger } = require('./loggers/console')
 
-const profiler = process.env.AWS_LAMBDA_FUNCTION_NAME ? new ServerlessProfiler() : new Profiler()
+const profiler = new Profiler()
 
 module.exports = {
   profiler,
--- a/packages/dd-trace/src/profiling/profiler.js
+++ b/packages/dd-trace/src/profiling/profiler.js
@@ -198,29 +198,6 @@ class Profiler extends EventEmitter {
   }
 }
 
-class ServerlessProfiler extends Profiler {
-  constructor () {
-    super()
-    this._profiledIntervals = 0
-    this._interval = 1
-    this._flushAfterIntervals = undefined
-  }
-
-  _setInterval () {
-    this._timeoutInterval = this._interval * 1000
-    this._flushAfterIntervals = this._config.flushInterval / 1000
-  }
-
-  async _collect (snapshotKind, restart = true) {
-    if (this._profiledIntervals >= this._flushAfterIntervals || !restart) {
-      this._profiledIntervals = 0
-      await super._collect(snapshotKind, restart)
-    } else {
-      this._profiledIntervals += 1
-      this._capture(this._timeoutInterval, new Date())
-      // Don't submit profile until 65 (flushAfterIntervals) intervals have elapsed
-    }
-  }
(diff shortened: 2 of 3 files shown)


wincent:
Closing this after verifying that there _is_ a difference between the two profilers in the context of short-lived invocations interspersed with frozen pauses. In the following examples, let's assume invocations that take about 100ms, and the Lambda getting frozen immediately after each one:

1. **Normal profiler:** single timer with 65s deadline; if the Lambda freezes for > 65s, when it resumes, runtime will check timer, see deadline has passed, and invoke callback, flushing the profile data.
2. **Serverless profile:** 65 ⨉ 1-second deadlines; if the Lambda freezes for > 65s, when it resumes, only first timer's deadline will have passed, and next 1-second timer will be scheduled; no flushing happens until at least 65 timers have run.

The net effect is that the Normal profiler might flush a tiny profile if it spends a lot of time frozen, but the Serverless profiler will at least accumulate samples across multiple invocations (ie. they may still be small, but not _as_ small; the trade-off, however, is that there may be significant latency before flushing[^worst]).

[^worst]: Anecdotally, we see frozen Lambdas getting shut down after about 7 minutes of inactivity. So, the worst cases would be Lambdas which don't submit profiling data at all (because they get shutdown before we can flush), or Lambdas with big gaps between invocations (ie. imagine a Lambda that gets one invocation every 6 minutes; 390 minutes of wall-clock time could elapse before flushing).

PR #4619, opened 2024-08-21
don't analyze destroyed responses

simon-id:
### Motivation

 

TODO:
test: 
```
router.get('/wafcontext', (req, res) => {
  new Promise((resolve, reject) => {
    setTimeout(() => {
      reject('errrrrr')
    }, 5000)
  }).catch((err) => {
    res.status(500).json({ message: 'Error on request' + err })
  })
})
```
`curl -m 1 -v http://localhost:3000/ssrf/wafcontext`

PR #4653, opened 2024-09-05
Check context disposal before calling WAF run

CarlesDD:
### What does this PR do?
Checks WAF context before calling WAF run. If the context has been already disposed, log a warning.

### Motivation
Change the logging level of this situation to warning.

### DIFF
--- a/packages/dd-trace/src/appsec/waf/waf_context_wrapper.js
+++ b/packages/dd-trace/src/appsec/waf/waf_context_wrapper.js
@@ -51,6 +51,11 @@ class WAFContextWrapper {
 
     if (!payloadHasData) return
 
+    if (this.ddwafContext.disposed) {
+      log.warn('Calling run on a disposed context')
+      return
+    }
+
     try {
       const start = process.hrtime.bigint()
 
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/appsec/waf/waf_context_wrapper.js
```diff
@@ -51,6 +51,11 @@ class WAFContextWrapper {
 
     if (!payloadHasData) return
 
+    if (this.ddwafContext.disposed) {
+      log.warn('Calling run on a disposed context')
+      return
+    }
```
simon-id, line 57: worth putting this check at the start of the method so we don't waste time parsing data in case it's already disposed

CarlesDD: Moved the check to the start of the method.

packages/dd-trace/test/appsec/waf/waf_context_wrapper.spec.js
```diff
@@ -50,4 +51,72 @@ describe('WAFContextWrapper', () => {
...
+          [addresses.HTTP_INCOMING_QUERY]: { key: 'value' }
+        }
+      }
+
+      wafContextWrapper.run(payload)
+
+      sinon.assert.notCalled(ddwafContext.run)
+    })
+
+    it('Should log a warn when attempting to call run on a disposed context', () => {
```
simon-id, line 108: could merge this test with the previous one ? it's the same case

simon-id (approved):

CarlesDD:
Closes: #4619

PR #4656, opened 2024-09-06
Filter addresses before calling to waf to prevent running without knownAddresses

uurien:
### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/appsec/waf/waf_context_wrapper.js
+++ b/packages/dd-trace/src/appsec/waf/waf_context_wrapper.js
@@ -11,12 +11,13 @@ const preventDuplicateAddresses = new Set([
 ])
 
 class WAFContextWrapper {
-  constructor (ddwafContext, wafTimeout, wafVersion, rulesVersion) {
+  constructor (ddwafContext, wafTimeout, wafVersion, rulesVersion, knownAddresses) {
     this.ddwafContext = ddwafContext
     this.wafTimeout = wafTimeout
     this.wafVersion = wafVersion
     this.rulesVersion = rulesVersion
     this.addressesToSkip = new Set()
+    this.knownAddresses = knownAddresses
   }
 
   run ({ persistent, ephemeral }, raspRuleType) {
@@ -27,7 +28,8 @@ class WAFContextWrapper {
 
     const payload = {}
     let payloadHasData = false
-    const inputs = {}
+    const persistentInputs = {}
+    const ephemeralInputs = {}
     const newAddressesToSkip = new Set(this.addressesToSkip)
 
     if (persistent !== null && typeof persistent === 'object') {
@@ -35,8 +37,8 @@ class WAFContextWrapper {
       for (const key of Object.keys(persistent)) {
         // TODO: requiredAddresses is no longer used due to processor addresses are not included in the list. Check on
         // future versions when the actual addresses are included in the 'loaded' section inside diagnostics.
-        if (!this.addressesToSkip.has(key)) {
-          inputs[key] = persistent[key]
+        if (!this.addressesToSkip.has(key) && this.knownAddresses.has(key)) {
+          persistentInputs[key] = persistent[key]
           if (preventDuplicateAddresses.has(key)) {
             newAddressesToSkip.add(key)
           }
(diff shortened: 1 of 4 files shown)


packages/dd-trace/test/appsec/waf/index.spec.js
```diff
@@ -7,6 +7,14 @@ const Reporter = require('../../../src/appsec/reporter')
 const web = require('../../../src/plugins/util/web')
 
 describe('WAF Manager', () => {
+  const knownAddresses = new Set([
```
CarlesDD, line 10: Can we construct the Set with the constants from `packages/dd-trace/src/appsec/addresses.js` as it has been done in [packages/dd-trace/test/appsec/waf/waf_context_wrapper.spec.js](https://github.com/DataDog/dd-trace-js/pull/4656/files#diff-ac9108ef47852a2c1fffcd2f56c974b106a219bcaf1419bff66e7a2aaf889eccR8)?

uurien: I didn't want to modify the tests, and the tests in this file are using address strings directly, without using constants from `addresses.js`. The file `waf_context_wrapper.spec.js` is using the constants, so the Set is also using the constants.

packages/dd-trace/src/appsec/waf/waf_manager.js
```diff
@@ -39,7 +39,8 @@ class WAFManager {
         this.ddwaf.createContext(),
         this.wafTimeout,
         this.ddwafVersion,
-        this.rulesVersion
+        this.rulesVersion,
+        this.ddwaf.knownAddresses
```
simon-id, line 43: no test for this ? should we make one ?

packages/dd-trace/src/appsec/waf/waf_context_wrapper.js
```diff
@@ -27,31 +28,37 @@ class WAFContextWrapper {
 
     const payload = {}
     let payloadHasData = false
-    const inputs = {}
+    const persistentInputs = {}
+    const ephemeralInputs = {}
```
simon-id, line 32: yes but now that you moved the rest of the if blocks, those are way overscoped, you can move them in their respective `if( presistent !== null ...`

packages/dd-trace/src/appsec/waf/waf_manager.js
uurien: I added a simple test for this property.

simon-id (approved):

packages/dd-trace/src/appsec/waf/waf_context_wrapper.js
```diff
@@ -27,31 +28,39 @@ class WAFContextWrapper {
...
       for (const key of Object.keys(persistent)) {
-        // TODO: requiredAddresses is no longer used due to processor addresses are not included in the list. Check on
-        // future versions when the actual addresses are included in the 'loaded' section inside diagnostics.
-        if (!this.addressesToSkip.has(key)) {
-          inputs[key] = persistent[key]
+        if (!this.addressesToSkip.has(key) && this.knownAddresses.has(key)) {
+          persistentInputs[key] = persistent[key]
           if (preventDuplicateAddresses.has(key)) {
             newAddressesToSkip.add(key)
           }
```
iunanua, line 41: I don't get this check. `preventDuplicateAddresses` is a fixed Set, isn't it?

iunanua: So its values could be added when `newAddressesToSkip` is created in line 31.

uurien: I don't understand your point...
> `preventDuplicateAddresses` is a fixed Set, isn't it?

Yes, it is. 

> So its values could be added when newAddressesToSkip is created in line 31

No, if we do that, instead of prevent sending it twice in the same request, we will send it only if we are trying to send it in the first call. We should add it to skipped Set only when a payload with the address happens.

iunanua: true, forget it :P

iunanua (approved):

PR #4670, opened 2024-09-11
Automatic userID tracking and blocking

simon-id:
### What does this PR do?
Adds automatic User ID tracking when using passport.js

Changes:
- Add instrumentation for passport `deserializeUser`
- Add an IAST weak hash exclusion for express-session as it's using SHA1 for diff checking.
- Make the user tracking SDK call the WAF whenever an ID or login is passed.
- Make the user tracking SDK add the `_dd.appsec.user.collection_mode: 'sdk'` tag.
- Use `res.constructor.prototype.end()` instead of `res.end()` in `blocking.js` to bypass `express-session` monkeypatch of that method.
- Add automated user tracking and blocking for `passport` module.
- Add `instrum.user_auth.missing_user_id` telemetry metric.
- Remove rootSpan check from `trackLogin()` as it's already checked by the caller.
- Add `trackUser()` method, called right after `deserializeUser()` finishes.
- Update all relevant tests.

ST: https://github.com/DataDog/system-tests/pull/3747/


packages/datadog-instrumentations/src/passport.js
```diff
@@ -0,0 +1,87 @@
...
+}, Authenticator => {
+  shimmer.wrap(Authenticator.prototype, 'deserializeUser', wrapDeserializeUser)
+
+  shimmer.wrap(Authenticator.prototype, 'authenticate', function wrapAuthenticate (authenticate) {
+    return function wrappedAuthenticate (name) {
+      const middleware = authenticate.apply(this, arguments)
+
+      const strategy = this._strategy(name)
+
+      strategy._verify
```
simon-id, line 70: options.session can be used to know if it's login or session or not ?

.github/workflows/plugins.yml
```diff
@@ -878,6 +878,30 @@ jobs:
...
+      - uses: ./.github/actions/plugins/test
+
+  passport-local:
+    runs-on: ubuntu-latest
+    env:
+      PLUGINS: passport-local
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/plugins/test
+
```
simon-id, line 904: this might be superseded by  https://github.com/DataDog/dd-trace-js/pull/5151

packages/dd-trace/src/appsec/sdk/set_user.js
```diff
@@ -22,6 +24,19 @@ function setUser (tracer, user) {
   }
 
   setUserTags(user, rootSpan)
+  rootSpan.setTag('_dd.appsec.user.collection_mode', 'sdk')
+
+  const persistent = {}
+
+  if (user.id) {
```
uurien, line 31: This can not be falsy, it is already checked in line 15.

simon-id: indeed

packages/dd-trace/src/appsec/sdk/set_user.js
```diff
@@ -22,6 +24,19 @@ function setUser (tracer, user) {
...
+    persistent[addresses.USER_ID] = '' + user.id
+  }
```
simon-id, line 33:
```suggestion
  const persistent = {
    [addresses.USER_ID]: '' + user.id
  }
```

packages/dd-trace/src/appsec/sdk/set_user.js
```diff
@@ -22,6 +24,17 @@ function setUser (tracer, user) {
   }
 
   setUserTags(user, rootSpan)
+  rootSpan.setTag('_dd.appsec.user.collection_mode', 'sdk')
```
iunanua, line 27: should we have a constant for `_dd.appsec.user.collection_mode`? it is used in several places

packages/dd-trace/src/appsec/user_tracking.js
```diff
@@ -162,7 +159,36 @@ function trackLogin (framework, login, user, success, rootSpan) {
...
+    return
+  }
+
+  rootSpan.setTag('_dd.appsec.usr.id', userId)
+
+  const isSdkCalled = rootSpan.context()._tags['_dd.appsec.user.collection_mode'] === 'sdk'
+  // do not override SDK
+  if (!isSdkCalled) {
+    rootSpan.addTags({
+      'usr.id': userId,
```
iunanua, line 178: sometimes we use `'usr.id'` and others `[addresses.USER_ID]`

iunanua: one is a span tag and the other is an address but...

packages/dd-trace/src/appsec/sdk/set_user.js
simon-id: We have a lot of hardcoded tags everywhere in the codebase, and we don't have any enum file for appsec yet. I would definitely like to do that, but not sure it makes sense in this PR ? I'd rather make a subsequent PR where I remove all hardcoded tags from our codebase, and put them into an enum file. WDYT ?

packages/dd-trace/src/appsec/user_tracking.js
simon-id: yes, one is a tag, and the other is a waf address, they should not be mixed. Although I agree the tags could use an enum instead of being hardcoded. See my comment above

.github/workflows/plugins.yml
simon-id: yep, i'm removing this file and changing it to the new system

IlyasShabi:
LGTM

uurien (approved):

PR #4674, opened 2024-09-11
New automatic user event collection

simon-id:
#### TL;DR:
- support blocking in passport strategies instrumentation
- don't send passport strategies instrumentation events if there is an error
- add new user login events collection logic
- add a mode config to control the logic (+remote config)
- pass user data to WAF in user login SDK calls
- move some code around because god damn it was ugly

#### RTFM:
- passport instrumentation:
  - chore: simplify the instrumentation for `passport-local` and `passport-http` by sharing the wrappers between the two in `passport-utils.js` (by using `this._passReqToCallback` instead of the config object)
  - fix: handle the `err` argument in `passport-utils.js` when wrapping the callback of `verify()` to not publish auth events when a db error occured for example
  - refactor: remove `credentials` fields in passport verify event as they're useless. Instead add `success` and `login`
  - feat: add abort controller blocking support in the `datadog:passport:verify:finish` channel
  - refactor: use `this.name` to get strategy name instead of setting an hardcoded value

- passport-http and passport-local tests:
  - fix: pass `passReqToCallback: true` in the correct place to actually test this case, which requires having two strategies

- chore: delete passport-utlis tests as they were not useful and replace them with passport-http and passport-local tests

- RC for auto user events collection mode:
  - feat: send new RC capability
  - feat: add handler to configure collection mode
    - add a special logic to reject overlapping configurations

- SDK:
  - refactor: separate code between SDK and automated user events, it was getting too complicated to handle both in one spot
    - remove non SDK logic from the SDK files
  - refactor: extract WAF call from `trackEvent()` into its own function for better clarity
  - feat: pass user data to the WAF in SDK user event calls
    - add special logic to convert SDK API into WAF events, by passing user ID and login depending on if success or failure
  - feat: add `usr.login` to the metadata tags if not already provided, with the value of `usr.id`

- `index.js`
  - chore: at startup, set collection mode but do not override if already present, in case RC changed the value already
  - refactor: always subscribe `onPassportVerify()` because it can be enabled and disabled by RC
  - refactor: change `onPassportVerify()` a little to put the responsibility of providing the correct data on the publisher instead of the subscriber
  - feat: add blocking support in `onPassportVerify()`

- Telemetry:
  - feat: add a counter for login events with missing data

- Automated User Events:
  - feat: add a new file `user_tracking.js` to differentiate automated user events from SDK user events
    - remove `passport.js` and migrated some of the logic to `user_tracking.js` to make it framework independent and clearer
  - HOW IT WORKS:
    - 3 collection modes, set with `setCollectionMode()`, called by appsec index and remote config:
      - `anon`/`anonymous`/`safe`: hash the id and login
      - `ident`/`identification`/`extended`: don't hash the id and login
      - `disabled`: don't collect the user id and login
    - `getUserId()`: called by `trackLogin()`
      - we have a list of fields we try to get from the user object: `USER_ID_FIELDS`, if we find one we use it
      - we try to stringify the value for max compatibility
      - if it looks like an non-serializable object that starts with `[object `, we skip it
    - `trackLogin()`: called by appsec index in the `passport.verify` DC subscriber
      - make a list of tags to be added to rootspan depending on if login is successful or not
      - some tags overlap with the SDK so we don't set them if the SDK already did
      - in case of login success, if the user ID is present, we consider it as the "current" user by setting the tag `usr.id`, and also sending it to the WAF
      - in all cases we send the user login and the event type to the WAF and return the results so appsec index can handle the results and block if needed

- Config:
  - refactor: remove internal config `appsec.eventTracking.enabled`
  - feat: add new collection modes `identification` and `anonymous` to `appsec.eventTracking.mode`
  - feat: add new env var for colleciton modes `DD_APPSEC_AUTO_USER_INSTRUMENTATION_MODE`
  - feat: add retrocompatibility with the old collection modes from `DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING`

- tests: add missing coverage for existing and new features, fix various typos, incorrect tests, inconsistencies, etc...

- chore: update typings for new config


packages/datadog-instrumentations/src/passport-utils.js
```diff
@@ -5,33 +5,52 @@ const { channel } = require('./helpers/instrument')
...
-      return verify.apply(this, arguments)
+function wrapVerify (verify) {
+  return function wrappedVerify (req, username, password, verified) {
+    if (passportVerifyChannel.hasSubscribers) {
+      // replace the callback with our own wrapper to get the result
+      // if we ever need the type of strategy, we can get it from this.name
+      if (this._passReqToCallback) {
+        arguments[3] = wrapVerifiedAndPublish(username, verified)
+      } else {
+        arguments[2] = wrapVerifiedAndPublish(req, password) // shifted args
```
simon-id, line 29:
```suggestion
        arguments[3] = wrapVerifiedAndPublish(arguments[1], arguments[3])
      } else {
        arguments[2] = wrapVerifiedAndPublish(arguments[0], arguments[2]) // shifted args
```

packages/dd-trace/src/appsec/user_tracking.js
```diff
@@ -0,0 +1,168 @@
...
+      }
+    }
+
+    /* TODO: if one day we have this info
+    if (exists != null && shouldSetTag('appsec.events.users.login.failure.usr.exists')) {
+      newTags['appsec.events.users.login.failure.usr.exists'] = exists
+    }
+    */
+
+    persistent['server.business_logic.users.login.failure'] = null
```
CarlesDD, line 154:
```suggestion
    persistent[addresses.LOGIN_FAILURE] = null
```

index.d.ts
```diff
@@ -655,12 +655,24 @@ declare namespace tracer {
...
+         * *  'safe': deprecated alias for 'anonymous'
+         * 
+         * *  'identification': will collect user IDs and logins without redaction
+         * *  'ident': alias for 'identification'
+         * *  'extended': deprecated alias for 'identification'
+         * 
+         * *  'disabled': will not collect user IDs and logins
+         * 
+         * Unknown values will be considered as 'disabled'
+         * @default 'ident'
```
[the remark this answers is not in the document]
simon-id:
```suggestion
         * @default 'identification'
```

packages/dd-trace/src/appsec/user_tracking.js
```diff
@@ -0,0 +1,168 @@
...
+  if (!user) return
+
+  for (const field of USER_ID_FIELDS) {
+    let id = user[field]
+
+    // try to find a field that can be stringified
+    if (id && typeof id.toString === 'function') {
+      id = id.toString()
+
+      if (id.startsWith('[object ')) {
```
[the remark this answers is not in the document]
simon-id: i don't mind adding that little check tho, but yes, it's extremely paranoid hahaha

packages/dd-trace/src/appsec/user_tracking.js
```diff
@@ -0,0 +1,168 @@
...
+  const persistent = {
+    [addresses.USER_LOGIN]: login
+  }
+
+  const currentTags = rootSpan.context()._tags
+  const isSdkCalled = currentTags[`_dd.appsec.events.users.login.${success ? 'success' : 'failure'}.sdk`] === 'true'
+
+  // used to not overwrite tags set by SDK
+  function shouldSetTag (tag) {
+    return !(isSdkCalled && currentTags[tag])
```
uurien, line 105:
```suggestion
    return !isSdkCalled || !currentTags[tag]
```

I think that this is more readable, should set tag when sdk is not called or tag is not set, rather than should be called when not happening these two things: sdk is called and current tag is set.

packages/dd-trace/src/appsec/user_tracking.js
```diff
@@ -0,0 +1,168 @@
...
+      '_dd.appsec.events.users.login.success.auto.mode': collectionMode,
+      '_dd.appsec.usr.login': login
+    }
+
+    if (shouldSetTag('appsec.events.users.login.success.usr.login')) {
+      newTags['appsec.events.users.login.success.usr.login'] = login
+    }
+
+    if (userId) {
+      newTags['_dd.appsec.usr.id'] = userId
```
uurien, line 120: The spec says that this should be provided only when the user event has been collected through automated instrumentation, we shouldn't send this property via the sdk.

uurien: Nah, sorry, this methdo is not called from the SDK, my fault.

packages/dd-trace/src/appsec/user_tracking.js
simon-id: hahahaha honestly this is exactly what i instinctively wrote at first, but when i read it later i got confused and wasn't sure it was the correct logic, the two are logically equivalent but i rewrote it with the parenthesis because it felt more "english" to me:
```
if sdk is called and the tag already exist then do NOT add a new tag

vs

if sdk is not called or the tag doesn't already exist then add a new tag
```
which one feels more natural language ?

packages/dd-trace/src/appsec/user_tracking.js
simon-id: yes big part of this PR is to completely separate SDK and auto events, to avoid spaghetti

packages/dd-trace/src/appsec/user_tracking.js
```diff
@@ -0,0 +1,168 @@
...
+
+function trackLogin (framework, login, user, success, rootSpan) {
+  if (!collectionMode || collectionMode === 'disabled') return
+
+  if (!rootSpan) {
+    log.error('No rootSpan found in AppSec trackLogin')
+    return
+  }
+
+  if (!login || typeof login !== 'string') {
```
uurien, line 81: empty string is a valid login attemp.

simon-id: ok i will remove the check because it's try it's technically not invalid, but the current passport strategies we support (passport-local, passport-http basic), check if the login is empty before our instrumentation, so for now, an empty login should never happen

simon-id: Fixed in [956cb9b](https://github.com/DataDog/dd-trace-js/pull/4674/commits/956cb9b45e863272e73ff7cfab27a5b4492ad3ac)

packages/dd-trace/src/appsec/remote_config/index.js
```diff
@@ -23,9 +24,32 @@ function enable (config, appsec) {
...
+
+    rc.setProductHandler('ASM_FEATURES', (action, rcConfig, configId) => {
       if (!rcConfig) return
 
+      // this is put before other handlers because it can reject the config
+      if (typeof rcConfig.auto_user_instrum?.mode === 'string') {
+        if (action === 'apply' || action === 'modify') {
+          // check if there is already a config applied with this field
+          if (autoUserInstrumModeId && configId !== autoUserInstrumModeId) {
+            // eslint-disable-next-line no-throw-literal
```
iunanua, line 39: should we add an error? something like:
`log.error('[RC] Discarding config. Multiple auto_user_instrum.mode received')`

packages/dd-trace/src/appsec/remote_config/index.js
```diff
@@ -23,9 +24,32 @@ function enable (config, appsec) {
...
+          }
+
+          setCollectionMode(rcConfig.auto_user_instrum.mode)
+          autoUserInstrumModeId = configId
+        } else {
+          if (configId === autoUserInstrumModeId) {
+            setCollectionMode(config.appsec.eventTracking.mode)
+            autoUserInstrumModeId = null
+          }
+        }
```
iunanua, line 50:
```suggestion
        } else if (configId === autoUserInstrumModeId) {
          setCollectionMode(config.appsec.eventTracking.mode)
          autoUserInstrumModeId = null
        }
```

packages/dd-trace/src/appsec/remote_config/index.js
```diff
@@ -23,9 +24,32 @@ function enable (config, appsec) {
...
+    let autoUserInstrumModeId
+
+    rc.setProductHandler('ASM_FEATURES', (action, rcConfig, configId) => {
       if (!rcConfig) return
 
+      // this is put before other handlers because it can reject the config
+      if (typeof rcConfig.auto_user_instrum?.mode === 'string') {
+        if (action === 'apply' || action === 'modify') {
+          // check if there is already a config applied with this field
+          if (autoUserInstrumModeId && configId !== autoUserInstrumModeId) {
```
iunanua, line 38: this means that once a instrum mode is received, it cannot be changed, no?

packages/dd-trace/src/appsec/telemetry.js
```diff
@@ -172,6 +172,15 @@ function addRaspRequestMetrics (store, { duration, durationExt }) {
   store[DD_TELEMETRY_REQUEST_METRICS].raspEvalCount++
 }
 
+function incrementMissingUserLogin (framework, eventType) {
```
iunanua, line 175:
```suggestion
function incrementMissingUserLoginMetric (framework, eventType) {
```

packages/dd-trace/src/appsec/user_tracking.js
```diff
@@ -0,0 +1,168 @@
...
+function trackLogin (framework, login, user, success, rootSpan) {
+  if (!collectionMode || collectionMode === 'disabled') return
+
+  if (!rootSpan) {
+    log.error('No rootSpan found in AppSec trackLogin')
+    return
+  }
+
+  if (typeof login !== 'string') {
+    log.error('Invalid login provided to AppSec trackLogin')
```
iunanua, line 82:
```suggestion
    log.error('[ASM] Invalid login provided to AppSec trackLogin')
```

packages/dd-trace/src/appsec/user_tracking.js
```diff
@@ -0,0 +1,168 @@
...
+
+  const persistent = {
+    [addresses.USER_LOGIN]: login
+  }
+
+  const currentTags = rootSpan.context()._tags
+  const isSdkCalled = currentTags[`_dd.appsec.events.users.login.${success ? 'success' : 'failure'}.sdk`] === 'true'
+
+  // used to not overwrite tags set by SDK
+  function shouldSetTag (tag) {
```
iunanua, line 104: wouldn't it be clearer to define `shouldSetTag` outside of the `trackLogin` function?

iunanua: ok, I see. That way is a more performant way

packages/dd-trace/src/appsec/user_tracking.js
```diff
@@ -0,0 +1,168 @@
...
+const addresses = require('./addresses')
+const { keepTrace } = require('../priority_sampler')
+const { SAMPLING_MECHANISM_APPSEC } = require('../constants')
+const standalone = require('./standalone')
+const waf = require('./waf')
+
+// the RFC doesn't include '_id', but it's common in MongoDB
+const USER_ID_FIELDS = ['id', '_id', 'email', 'username', 'login', 'user']
+
+let collectionMode
```
iunanua, line 15: could it be assigned so you wouldn't need the `!collectionMode` in line 74?
```suggestion
let collectionMode = 'disabled'
```

packages/datadog-instrumentations/test/passport-http.spec.js
```diff
@@ -95,16 +125,17 @@ withVersions('passport-http', 'passport-http', version => {
...
       expect(res.status).to.equal(200)
       expect(res.data).to.equal('Granted')
-      expect(subscriberStub).to.be.calledOnceWithExactly(
-        {
-          credentials: { type: 'http', username: 'test' },
-          user: { _id: 1, username: 'test', password: '1234', email: 'testuser@ddog.com' }
-        }
-      )
+      expect(subscriberStub).to.be.calledOnceWithExactly({
+        framework: 'passport-basic',
```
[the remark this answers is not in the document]
simon-id: no, in the passport-http module there are two strategies: `basic` and `digest`. We want to be able to differentiate those two, so I called it passport-basic. Would you prefere something like `passport-http:basic` just in case someone creates a `passport-basic` module ?

packages/datadog-instrumentations/test/passport-http.spec.js
```diff
@@ -85,6 +103,18 @@ withVersions('passport-http', 'passport-http', version => {
       return agent.close({ ritmReset: false })
     })
 
+    it('should not call subscriber when an error occurs', async () => {
+      const res = await axios.get(`http://localhost:${port}/`, {
+        headers: {
+          // error:1234
+          Authorization: 'Basic ZXJyb3I6MTIzNAo='
```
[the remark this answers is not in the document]
simon-id: Ah whoops, that's my bad, I did a `echo "error:1234" | base64` and forgot it would include the newline. (it doesn't fail the test because we don't check for the password in an error case

packages/dd-trace/test/appsec/remote_config/index.spec.js
```diff
@@ -112,6 +125,57 @@ describe('Remote Config index', () => {
...
+        const configId = 'collectionModeId'
+
+        afterEach(() => {
+          listener('unnaply', rcConfig, configId)
+        })
+
+        it('should not update collection mode when not a string', () => {
+          listener('apply', { auto_user_instrum: { mode: 123 } }, configId)
+
+          expect(UserTracking.setCollectionMode).to.not.have.been.called
```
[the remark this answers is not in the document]
simon-id: Mmmh this is up to interpretation, if it's not a string, i consider the payload invalid, not just the mode, so i completely reject it. I only update to disabled if it's a string that's unknown.
image

packages/dd-trace/test/appsec/sdk/track_event.spec.js
```diff
@@ -129,23 +132,44 @@ describe('track_event', () => {
...
+          }
+        })
       })
     })
 
     describe('trackUserLoginFailureEvent', () => {
       it('should log warning when passed invalid userId', () => {
-        trackUserLoginFailureEvent(tracer, null, false)
-        trackUserLoginFailureEvent(tracer, [], false)
+        trackUserLoginFailureEvent(tracer, null, false, { key: 'value' })
```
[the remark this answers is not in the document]
simon-id: because it's written exactly like this for `trackUserLoginSuccessEvent` tests (example line 69), and it's also optional there anyway. Just wanted some consistency.

packages/dd-trace/src/appsec/user_tracking.js
simon-id: I don't think so, the reason why it's undefined at first is on purpose. When appsec/index.js calls this function, it must NOT overwrite remote config, and the way i did this was to only set the value if it's undefined. Because if it's not undefined, it was either set by RC or previously set by the same local config, in both cases we shouldn't override it.
TLDR it's to differentiate between "clean startup" and "remote config already set this value"

packages/dd-trace/src/appsec/user_tracking.js
simon-id: Not sure it's more performant, but if the function was outside this scope, it would have to look like this `shouldSetTag(currentTags, isSdkCalled, tag)` and since it's called a few times, I thought it would be too verbose.

simon-id: it's more readable with the closure imo

packages/dd-trace/src/appsec/user_tracking.js
```diff
@@ -0,0 +1,168 @@
...
+      return obfuscateIfNeeded(id)
+    }
+  }
+}
+
+function trackLogin (framework, login, user, success, rootSpan) {
+  if (!collectionMode || collectionMode === 'disabled') return
+
+  if (!rootSpan) {
+    log.error('No rootSpan found in AppSec trackLogin')
```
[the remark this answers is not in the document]
simon-id: fixed in [e9ceb27](https://github.com/DataDog/dd-trace-js/pull/4674/commits/e9ceb2740431b7268cfd48e8b4fda19557cdec89)

packages/dd-trace/src/appsec/user_tracking.js
simon-id: fixed in [e9ceb27](https://github.com/DataDog/dd-trace-js/pull/4674/commits/e9ceb2740431b7268cfd48e8b4fda19557cdec89)

packages/dd-trace/src/appsec/telemetry.js
```diff
@@ -172,6 +172,15 @@ function addRaspRequestMetrics (store, { duration, durationExt }) {
...
+  if (!enabled) return
+
+  appsecMetrics.count('instrum.user_auth.missing_user_login', {
+    framework,
+    event_type: eventType
```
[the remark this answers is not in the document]
simon-id: Very good point! I didn't know it worked like that. I've created a PR: https://github.com/DataDog/dd-go/pull/160921

packages/dd-trace/src/appsec/remote_config/index.js
simon-id: are you sure ? doesn't it mean that if a instrum mode is received, it needs to be unnaplied before a new one can be applied. But in the case of a `action === 'modify`, the configId is the say, so it should pass the check

packages/dd-trace/src/appsec/remote_config/index.js
simon-id: Fixed in [98cae84](https://github.com/DataDog/dd-trace-js/pull/4674/commits/98cae84a93d3efb150b93273d207bd15e619572d)

packages/dd-trace/src/appsec/user_tracking.js
simon-id: can i close this ?

packages/dd-trace/src/appsec/remote_config/index.js
iunanua: so, when a value changes in the UI, there is first an unapply action and then after that another one to apply the new value?
sorry don't know the details of how RC works.

simon-id: idk but it doesn't have to be, are you talking about the action === 'modify' case ?

packages/dd-trace/src/appsec/telemetry.js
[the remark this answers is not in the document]
simon-id: reopening this thread to remind us to actually merge that PR

packages/dd-trace/test/appsec/user_tracking.spec.js
```diff
@@ -0,0 +1,696 @@
...
+
+const assert = require('assert')
+
+const log = require('../../src/log')
+const telemetry = require('../../src/appsec/telemetry')
+const { SAMPLING_MECHANISM_APPSEC } = require('../../src/constants')
+const standalone = require('../../src/appsec/standalone')
+const waf = require('../../src/appsec/waf')
+
+describe.only('User Tracking', () => {
```
simon-id, line 11: good ol' `.only`
```suggestion
describe('User Tracking', () => {
```

uurien (dismissed):

iunanua (approved):

simon-id:
system tests are expected to fail. See this PR for green system-tests: https://github.com/DataDog/system-tests/pull/3666

packages/dd-trace/src/appsec/sdk/track_event.js
```diff
@@ -23,7 +24,13 @@ function trackUserLoginSuccessEvent (tracer, user, metadata) {
 
   setUserTags(user, rootSpan)
 
-  trackEvent('users.login.success', metadata, 'trackUserLoginSuccessEvent', rootSpan, 'sdk')
+  const login = user.login ?? user.id
+
+  metadata = Object.assign({ 'usr.login': login }, metadata)
```
iunanua, line 29:
```suggestion
  metadata = { 'usr.login': login, ...metadata }
```

iunanua (approved):

uurien (approved):

iunanua (approved):

PR #4676, opened 2024-09-12
Exploit Prevention LFI

iunanua:
### What does this PR do?
- Add new `AppsecFsPlugin` to mark child fs operations and to mark excluded operations when express is rendering views.
- Add LFI analyzer
- Discard fs child operations in IAST `PathTraversalAnalyzer`
- Include `ASM_RASP_LFI` RC capability

ST https://github.com/DataDog/system-tests/pull/3024

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/dd-trace/src/appsec/iast/analyzers/path-traversal-analyzer.js
```diff
@@ -29,7 +29,14 @@ class PathTraversalAnalyzer extends InjectionAnalyzer {
 
   onConfigure () {
     this.addSub('apm:fs:operation:start', (obj) => {
-      if (ignoredOperations.includes(obj.operation)) return
+      const store = storage.getStore()
+      const outOfReqOrChild = !store?.fs?.root
```
uurien, line 33: if `store.fs` is undefined, we are not going to check the path traversal.
I think that it can be undefined if the subscription order is not the expected order.

packages/dd-trace/src/appsec/iast/index.js
```diff
@@ -27,6 +28,7 @@ function enable (config, _tracer) {
   if (isEnabled) return
 
   iastTelemetry.configure(config, config.iast?.telemetryVerbosity)
+  enableFsPlugin('iast')
```
uurien, line 31: Could we extract this to a constant into `fs-plugin` file?

packages/dd-trace/src/appsec/rasp/lfi.js
```diff
@@ -0,0 +1,107 @@
...
+
+function shouldAnalyze (path, fs) {
+  if (!path) return
+
+  const notExcludedRootOp = !fs.opExcluded && fs.root
+  return notExcludedRootOp && (isAbsolute(path) || path.includes('../') || shouldAnalyzeURLFile(path, fs))
+}
+
+function shouldAnalyzeURLFile (path, fs) {
+  if (path.startsWith('file://')) {
```
uurien, line 99: I understood that waf will handle with this in the future, a TODO here to delete this check when it is supported?

packages/dd-trace/test/appsec/index.express.plugin.spec.js
```diff
@@ -95,6 +95,9 @@ withVersions('express', 'express', version => {
             rules: path.join(__dirname, 'api_security_rules.json'),
             apiSecurity: {
               enabled: true
+            },
+            rasp: {
```
uurien, line 99: why is this needed?

packages/dd-trace/test/appsec/index.spec.js
```diff
@@ -1062,6 +1063,8 @@ describe('IP blocking', function () {
       }
     }))
 
+    disableLfi()
```
uurien, line 1066: why? I think that it worth to add a comment in the code to understand this unrelated change.

packages/dd-trace/test/appsec/response_blocking.spec.js
```diff
@@ -55,6 +56,8 @@ describe('HTTP Response Blocking', () => {
         rules: path.join(__dirname, 'response_blocking_rules.json')
       }
     }))
+
+    disableLfi()
```
uurien, line 60: same here, I can imagine that without this disableLfi the tests are failing, but I don't understand why, write a comment here please

uurien (commented):
Please, change also the title of the PR to something like. "Exploit Prevention LFI" instead of "rasp lfi" (and use it in the squash and merge commit title

packages/dd-trace/src/appsec/iast/analyzers/path-traversal-analyzer.js
iunanua: yes, `AppsecFsPlugin` must be subscribed before the analyzer.

we could remove this dependency if `AppsecFsPlugin` would publish another event...

packages/dd-trace/test/appsec/response_blocking.spec.js
iunanua: as rasp is enabled by default, the test execution triggers some lfi checks executing `waf.run` and changing the expectation of the test.

But, instead of calling `disableLfi()` I have changed it to disable rasp with a config flag.

packages/dd-trace/test/appsec/index.express.plugin.spec.js
iunanua: it may not be necessary...

packages/dd-trace/src/appsec/rasp/lfi.js
iunanua: `shouldAnalyzeURLFile` only checks if url file paths are absolute or contain ../

packages/dd-trace/src/appsec/iast/analyzers/path-traversal-analyzer.js
uurien: and could we add a test to ensure that we are calling to `AppsecFsPlugin.enable` before `enableAllAnalyzers`?

packages/dd-trace/src/appsec/rasp/fs-plugin.js
```diff
@@ -0,0 +1,99 @@
...
+
+const enabledFor = {
+  [RASP_MODULE]: false,
+  [IAST_MODULE]: false
+}
+
+let fsPlugin
+
+function enterWith (fsProps, store = storage.getStore()) {
+  if (store && !store.fs?.opExcluded) {
```
CarlesDD, line 18: `!store.fs?.opExcluded` could be true in several situations:

- if `opExcluded` exists and its value is `false` (I think it is the intended check)
- if `opExcluded` does not exist in `fs` object
- if `fs` does not exist in `store` object

Are the two latter conditions ok to proceed with the if body? Or is it assure that they never will happen?

iunanua: yes, they are.
`enterWith` is NOT excluded if:
- `fs` doesn't exist -> probably it is a root fs operation so `enterWith` will create a new store with the `fs` attribute
- `fs.opExcluded` doesn't exist but `fs` does -> store has been created before by a root fs operation

uurien (approved):

uurien (dismissed):

packages/dd-trace/test/appsec/rasp/lfi.express.plugin.spec.js
```diff
@@ -82,12 +82,28 @@ describe('RASP - lfi', () => {
         }
       }
 
+      function getAppSync (fn, args, options) {
+        return (req, res) => {
+          try {
+            const result = fn(args)
+            options.onfinish?.(result)
+          } catch (e) {
+            if (e.message === 'DatadogRaspAbortError') {
```
uurien, line 91: I miss a test without this trycatch, to test that we are blocking the request when the exception is handled by express.

iunanua: https://github.com/DataDog/dd-trace-js/pull/4676/files#diff-4947cef62a3c8302aea45bb69bf3612b71b8cc4b3f404dd189d8c32b5dd5a526

uurien (approved):

CarlesDD (approved):

PR #4677, opened 2024-09-12
[test visibility] Add test source file to test suites

juan-fernandez:
### What does this PR do?

Add `test.source.file`, `test.source.start` and `test.codeowners` to test suites, in addition to tests, which already had those tags.

### Motivation

Test suites are test files and it's useful for users to be able to look at test files in Datadog UI. This allows customer to do this.

### Plugin Checklist

- [x] Unit tests.

### DIFF
--- a/integration-tests/cucumber/cucumber.spec.js
+++ b/integration-tests/cucumber/cucumber.spec.js
@@ -235,9 +235,6 @@ versions.forEach(version => {
                 }
               )
 
-              childProcess.stdout.pipe(process.stdout)
-              childProcess.stderr.pipe(process.stderr)
-
               childProcess.on('exit', () => {
                 receiverPromise.then(() => done()).catch(done)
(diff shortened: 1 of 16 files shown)


romainkomorndatadog (approved):
LGTM but I feel like we have an implicit "suite maps to whole file" assumption and I wonder if we should have something more explicit.

Maybe specifying the file without listing a start line? 🤔

juan-fernandez:
> LGTM but I feel like we have an implicit "suite maps to whole file" assumption and I wonder if we should have something more explicit.
> 
> Maybe specifying the file without listing a start line? 🤔

if the suite maps to part of a file, wouldn't it work to pass the start line? If a suite maps to multiple files, this won't work, but we mentioned that source files in test suites would be best effort and the libraries would take the decision whether to include the tag or not, right?

PR #4687, opened 2024-09-13
feat(dsm): add datastreams monitoring schema tracking

wconti27:
### What does this PR do?
Adds Schema Tracking support for Datatreams Monitoring.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/datastreams/schemas/schema_builder.js
+++ b/packages/dd-trace/src/datastreams/schemas/schema_builder.js
@@ -1,0 +1,125 @@
+const LRUCache = require('lru-cache')
+const { fnv64 } = require('../fnv')
+const { Schema } = require('./schema')
+
+const maxDepth = 10
+const maxProperties = 1000
+const CACHE = new LRUCache({ max: 32 })
+
+class SchemaBuilder {
+  constructor (iterator) {
+    this.schema = new OpenApiSchema()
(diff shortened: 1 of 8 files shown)


packages/dd-trace/src/datastreams/schemas/schema_builder.js
```diff
@@ -0,0 +1,125 @@
...
+
+function convertKey (key) {
+  if (key === 'enumValues') {
+    return 'enum'
+  }
+  return key
+}
+
+function jsonStringify (obj, indent = 2) {
+  // made to stringify json exactly similar to python / java in order for hashing to be the same
```
piochelepiotr, line 110: Oh, that's a good catch. Note that we can also remap things in the backend. But it's better if we are able to get the data right from the start.
Do you know if the order of fields will be the same between languages?

wconti27: No, its just based on the order the property was added. Should we sort them alphabetically?

piochelepiotr: I think the current order works, it should be added in a fixed order for a given schema.

Qard (approved):

PR #4695, opened 2024-09-17
[MLOB-1540] add configuration options for LLMObs

sabrenner:
### What does this PR do?
Adds configuration options for LLMObs to set on the global tracer config.

### Motivation
Part of the Node.js LLMObs SDK official release.

### DIFF
--- a/packages/dd-trace/src/config.js
+++ b/packages/dd-trace/src/config.js
@@ -239,7 +239,8 @@ class Config {
     options = {
       ...options,
       appsec: options.appsec != null ? options.appsec : options.experimental?.appsec,
-      iast: options.iast != null ? options.iast : options.experimental?.iast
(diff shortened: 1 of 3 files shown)


packages/dd-trace/src/config.js
```diff
@@ -239,7 +239,8 @@ class Config {
     options = {
       ...options,
       appsec: options.appsec != null ? options.appsec : options.experimental?.appsec,
-      iast: options.iast != null ? options.iast : options.experimental?.iast
+      iast: options.iast != null ? options.iast : options.experimental?.iast,
+      llmobs: options.llmobs != null ? options.llmobs : options.experimental?.llmobs
```
sabrenner, line 243: we're not sure if we want to label this as experimental yet. once i have a PR up for the type definitions, we'll know more, and can maybe change this, but for now we'll check if the LLMObs config is on the main config or `experimental`

PR #4696, opened 2024-09-17
[MLOB-1540] add llmobs configuration to global tracer config

sabrenner:
### What does this PR do?
Adds LLM Observability configuration options to the global tracer config.

There is one caveat in the behavior implemented in the config here: `config.llmobs.enabled` is computed intentionally with regards to environment variable first (`DD_LLMOBS_ENABLED`), then if the corresponding options object exists. This breaks the normal pattern of `options` taking precedence over `environment`, but is in parity with our [LLMObs Python SDK](https://github.com/DataDog/dd-trace-py/blob/cc824a2da7c39ce296624844ad8e8057ec12b701/ddtrace/llmobs/_llmobs.py#L155-L157).

Additionally, there is no consensus on if we'll want this to be placed under `experimental` yet or not. I've included in this PR a change that will coalesce `llmobs` from either the general or `experimental` options onto the general options, and is used as such for the rest of the config (as is done with `appsec` and `iast`).

### Motivation
First PR in a set of PRs to release a Node.js LLM Obs SDK within the tracer.

### DIFF
--- a/packages/dd-trace/src/config.js
+++ b/packages/dd-trace/src/config.js
@@ -339,7 +339,7 @@ class Config {
     this._applyDefaults()
     this._applyEnvironment()
     this._applyOptions(options)
-    this._applyCalculated(options)
+    this._applyCalculated()
     this._applyRemote({})
     this._merge()
 
@@ -399,7 +399,7 @@ class Config {
     }
 
     // TODO: test
-    this._applyCalculated(options)
+    this._applyCalculated()
     this._merge()
   }
 
@@ -425,10 +425,10 @@ class Config {
     return inAWSLambda || isGCPFunction || isAzureFunction
   }
 
-  _isLLMObsEnabled (options = {}) {
+  _isLLMObsEnabled () {
     return coalesce(
       process.env.DD_LLMOBS_ENABLED,
-      !!options.llmobs
+      !!this._optionsArg.llmobs
     )
   }
 
@@ -1045,7 +1045,7 @@ class Config {
   }
 
   // handles values calculated from a mixture of options and env vars
-  _applyCalculated (options) {
+  _applyCalculated () {
     const calc = setHiddenProperty(this, '_calculated', {})
 
     const {
@@ -1090,7 +1090,7 @@ class Config {
       calc['tracePropagationStyle.extract'] = calc['tracePropagationStyle.extract'] || defaultPropagationStyle
     }
 
-    this._setBoolean(calc, 'llmobs.enabled', this._isLLMObsEnabled(options))
+    this._setBoolean(calc, 'llmobs.enabled', this._isLLMObsEnabled())
   }
 
   _applyRemote (options) {
--- a/packages/dd-trace/src/config.js
+++ b/packages/dd-trace/src/config.js
@@ -284,11 +284,6 @@ class Config {
       options.appsec = {}
     }
 
-    const DD_LLMOBS_ENABLED = coalesce(
-      process.env.DD_LLMOBS_ENABLED,
-      !!options.llmobs
-    )
-
     const DD_INSTRUMENTATION_INSTALL_ID = coalesce(
       process.env.DD_INSTRUMENTATION_INSTALL_ID,
       null
@@ -325,10 +320,6 @@ class Config {
     // TODO: refactor
     this.apiKey = DD_API_KEY
 
-    this.llmobs = {
-      enabled: isTrue(DD_LLMOBS_ENABLED)
-    }
-
     // sent in telemetry event app-started
     this.installSignature = {
       id: DD_INSTRUMENTATION_INSTALL_ID,
@@ -348,7 +339,7 @@ class Config {
     this._applyDefaults()
     this._applyEnvironment()
     this._applyOptions(options)
-    this._applyCalculated()
+    this._applyCalculated(options)
     this._applyRemote({})
     this._merge()
 
@@ -408,7 +399,7 @@ class Config {
     }
 
     // TODO: test
-    this._applyCalculated()
+    this._applyCalculated(options)
     this._merge()
   }
 
@@ -432,6 +423,13 @@ class Config {
     const isGCPFunction = getIsGCPFunction()
     const isAzureFunction = getIsAzureFunction()
     return inAWSLambda || isGCPFunction || isAzureFunction
+  }
+
+  _isLLMObsEnabled (options = {}) {
+    return coalesce(
+      process.env.DD_LLMOBS_ENABLED,
+      !!options.llmobs
+    )
   }
 
   // for _merge to work, every config value must have a default value
@@ -508,6 +506,7 @@ class Config {
     this._setValue(defaults, 'isManualApiEnabled', false)
     this._setValue(defaults, 'llmobs.agentlessEnabled', false)
     this._setValue(defaults, 'llmobs.apiKey', undefined)
+    this._setBoolean(defaults, 'llmobs.enabled', false)
     this._setValue(defaults, 'llmobs.mlApp', undefined)
     this._setValue(defaults, 'ciVisibilitySessionName', '')
     this._setValue(defaults, 'logInjection', false)
@@ -1046,7 +1045,7 @@ class Config {
   }
 
   // handles values calculated from a mixture of options and env vars
-  _applyCalculated () {
+  _applyCalculated (options) {
     const calc = setHiddenProperty(this, '_calculated', {})
 
     const {
@@ -1090,6 +1089,8 @@ class Config {
       calc['tracePropagationStyle.inject'] = calc['tracePropagationStyle.inject'] || defaultPropagationStyle
       calc['tracePropagationStyle.extract'] = calc['tracePropagationStyle.extract'] || defaultPropagationStyle
     }
+
+    this._setBoolean(calc, 'llmobs.enabled', this._isLLMObsEnabled(options))
   }
 
   _applyRemote (options) {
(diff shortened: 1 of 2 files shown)


packages/dd-trace/test/config.spec.js
```diff
@@ -1979,6 +2010,36 @@ describe('Config', () => {
...
+      const config = new Config()
+      expect(config.llmobs.enabled).to.be.false
+    })
+
+    it('should enable llmobs with options and DD_LLMOBS_ENABLED is not set', () => {
+      const config = new Config({ llmobs: {} })
+      expect(config.llmobs.enabled).to.be.true
+    })
+
+    it('should have DD_LLMOBS_ENABLED take priority over options', () => {
```
Yun-Kim, line 2036: Is this expected? Env vars take precedence over programmatically set values?

Yun-Kim (commented):
LGTM on behalf of team mlobs other than one question about precedence

packages/dd-trace/test/config.spec.js
sabrenner: Normally no, but I believe this is how we do it in the Python SDK, right? `DD_LLMOBS_ENABLED=false` overrides `LLMObs.enable()` (which, here, is the same as just setting the `llmobs` option alltogether).

Yun-Kim: Ahh yes, thanks for reminding me 🤦 this is fine, just wanted to double check. Thanks

Yun-Kim (approved):

packages/dd-trace/src/config.js
```diff
@@ -284,6 +284,11 @@ class Config {
       options.appsec = {}
     }
 
+    const DD_LLMOBS_ENABLED = coalesce(
```
ida613, line 287:
```suggestion
```

packages/dd-trace/src/config.js
```diff
@@ -284,6 +284,11 @@ class Config {
...
+      process.env.DD_LLMOBS_ENABLED,
+      !!options.llmobs
+    )
```
ida613, line 290:
```suggestion
```

packages/dd-trace/src/config.js
```diff
@@ -320,6 +325,10 @@ class Config {
     // TODO: refactor
     this.apiKey = DD_API_KEY
 
+    this.llmobs = {
+      enabled: isTrue(DD_LLMOBS_ENABLED)
+    }
+
```
ida613, line 331:
```suggestion
```

packages/dd-trace/src/config.js
```diff
@@ -497,6 +506,9 @@ class Config {
     this._setValue(defaults, 'isGitUploadEnabled', false)
     this._setValue(defaults, 'isIntelligentTestRunnerEnabled', false)
     this._setValue(defaults, 'isManualApiEnabled', false)
+    this._setValue(defaults, 'llmobs.agentlessEnabled', false)
+    this._setValue(defaults, 'llmobs.apiKey', undefined)
```
ida613, line 510:
```suggestion
    this._setValue(defaults, 'llmobs.apiKey', undefined)
    this._setValue(defaults, 'llmobs.enabled', false)
```

ida613, line 510:
```suggestion
    this._setValue(defaults, 'llmobs.apiKey', undefined)
    this._setValue(defaults, 'llmobs.enabled', false)
```

packages/dd-trace/src/config.js
```diff
@@ -594,6 +606,8 @@ class Config {
       DD_INSTRUMENTATION_TELEMETRY_ENABLED,
       DD_INSTRUMENTATION_CONFIG_ID,
       DD_LOGS_INJECTION,
+      DD_LLMOBS_AGENTLESS_ENABLED,
+      DD_LLMOBS_ML_APP,
```
ida613, line 610:
```suggestion
      DD_LLMOBS_ENABLED,
      DD_LLMOBS_ML_APP,
```

packages/dd-trace/src/config.js
```diff
@@ -734,6 +748,8 @@ class Config {
     this._setArray(env, 'injectionEnabled', DD_INJECTION_ENABLED)
     this._setBoolean(env, 'isAzureFunction', getIsAzureFunction())
     this._setBoolean(env, 'isGCPFunction', getIsGCPFunction())
+    this._setBoolean(env, 'llmobs.agentlessEnabled', DD_LLMOBS_AGENTLESS_ENABLED)
```
ida613, line 751:
```suggestion
    this._setBoolean(env, 'llmobs.agentlessEnabled', DD_LLMOBS_AGENTLESS_ENABLED)
    this._setBoolean(env, 'llmobs.enabled', DD_LLMOBS_ENABLED)
```

packages/dd-trace/src/config.js
```diff
@@ -902,6 +918,9 @@ class Config {
     }
     this._setString(opts, 'iast.telemetryVerbosity', options.iast && options.iast.telemetryVerbosity)
     this._setBoolean(opts, 'isCiVisibility', options.isCiVisibility)
+    this._setBoolean(opts, 'llmobs.agentlessEnabled', options.llmobs?.agentlessEnabled)
```
ida613, line 921:
```suggestion
    this._setBoolean(opts, 'llmobs.agentlessEnabled', options.llmobs?.agentlessEnabled)
    this._setBoolean(opts, 'llmobs.enabled', options.llmobs)
```

packages/dd-trace/test/config.spec.js
```diff
@@ -261,6 +261,9 @@ describe('Config', () => {
     expect(config).to.have.nested.property('installSignature.id', null)
     expect(config).to.have.nested.property('installSignature.time', null)
     expect(config).to.have.nested.property('installSignature.type', null)
+    expect(config).to.have.nested.property('llmobs.mlApp', undefined)
+    expect(config).to.have.nested.property('llmobs.agentlessEnabled', false)
+    expect(config).to.have.nested.property('llmobs.apiKey', undefined)
```
ida613, line 266: can we also add tests for llmobs.enabled?

sabrenner: I added some tests for `llmobs.enabled` at the bottom of the config tests. but, i wanted to set `config.llmobs.enabled` with `DD_LLMOBS_ENABLED` overriding the existence of `options.llmobs`, ie environment overrides options for that one config property. with regard to the other changes you've proposed, that's why I had it originally computing `config.llmobs.enabled` in the constructor by coalescing the environment variable first, and then the existence of the options to compute `llmobs.enabled`. With regards to the tests at the bottom, and that desired behavior, is there a way I can do that in the config? it's my understanding that `options` will always override `environment`, which is what i don't want in this case for `llmobs.enabled`.

packages/dd-trace/src/config.js
sabrenner: with regards to my comment below (& updated the pr description too for clarity) - could i maybe do something like:

```javascript
this._setBoolean(opts, 'llmobs.enabled', this._env['llmobs.enabled'] ?? options.llmobs)
```

since it looks like although `options` take precedence over `environment`, `environment` is calculated _before_ `options`, so I can use `this._env` in computing `this._opts['llmobs.enabled']`. wdyt?

sabrenner: this way, i can effectively have the environment variable take precedence over the options and respect the general flow of the config structure

sabrenner:
@ida613 I've tried to address some of the concerns in setting `config.llmobs.enabled` in https://github.com/DataDog/dd-trace-js/pull/4696/commits/8fd4912c4c710799c8d12ee5a47a975a3b44a799. let me know what you think!

packages/dd-trace/src/config.js
sabrenner: i think resolved using `_applyCalculated` in https://github.com/DataDog/dd-trace-js/pull/4696/commits/8fd4912c4c710799c8d12ee5a47a975a3b44a799

packages/dd-trace/src/config.js
```diff
@@ -937,6 +950,15 @@ class Config {
...
     this._setBoolean(opts, 'traceId128BitLoggingEnabled', options.traceId128BitLoggingEnabled)
     this._setString(opts, 'version', options.version || tags.version)
+
+    // For LLMObs, we want the environment variable to take precedence over the options.
+    // This is reliant on environment config being set before options.
+    // This is to make sure the origins of each value are tracked appropriately for telemetry.
+    // We'll only set `llmobs.enabled` on the opts when it's not set on the environment, and options.llmobs is provided.
+    const llmobsEnabledEnv = this._env['llmobs.enabled']
+    if (llmobsEnabledEnv == null && options.llmobs) {
+      this._setBoolean(opts, 'llmobs.enabled', !!options.llmobs)
```
ida613, line 960:
```suggestion
      this._setBoolean(opts, 'llmobs.enabled', options.llmobs)
```
`[_setBoolean](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/config.js#L1111)` does `!!`

sabrenner: I tried without `!!`, but found the problem was it would skip the `if (value === undefined || value === null)` check, but then would also skip `isTrue` and `isFalse` because both of them evaluate the truthiness of `value` by stringifying it and then checking it against either `true` or `1` (for `isTrue`) or `false` or `0` (for `isFalse`). 

Since `options.llmobs` is an object, when it's stringified, it just becomes `[object Object]`, and it doesn't hit any of the checks 😢 i don't see in `_setBoolean` where it applies the `!!` automatically

ida613: make sense! thanks Sam :)

ida613 (approved):

PR #4699, opened 2024-09-18
[MLOB-1555] LLM Observability writers

sabrenner:
### What does this PR do?
Adds LLM Observability writers for span events (agentless and agent proxy) as well as evaluation metrics (which write directly to our public API).

#### Important Notes
- These writers will run on intervals separate from the main agent exporter, and in a future PR will be initialized in the appropriate spots to start those intervals (as defined in the constructor of the base writer). Because of this, these writers specifically won't interact with any tracer internal exporters, writers, or encoders.
- We need to make sure unicode special characters are encoded in their `\\u` form in payload strings. I wasn't sure if there was a cleaner way to do this, so any input on this is appreciated!

### Motivation
Merge in incremental change of LLMObs writers into the LLM Observability SDK release branch. 

The timeline of changes to merge looks like (in order):
- [x] Config
- [ ] Writers
- [ ] Tagger
- [ ] Span/Trace processor
- [ ] SDK + initialization
- [ ] Type definitions
- [ ] OpenAI integration

### DIFF
--- a/packages/dd-trace/src/llmobs/writers/base.js
+++ b/packages/dd-trace/src/llmobs/writers/base.js
@@ -1,0 +1,108 @@
+'use strict'
+
+const request = require('../../exporters/common/request')
+const { URL, format } = require('url')
+
+const logger = require('../../log')
+
+const { encodeUnicode } = require('../util')
+
+class BaseLLMObsWriter {
+  constructor ({ interval, timeout, endpoint, intake, eventType, protocol, port }) {
+    this._interval = interval || 1000 // 1s
+    this._timeout = timeout || 5000 // 5s
+    this._eventType = eventType
+
+    this._buffer = []
+    this._bufferLimit = 1000
+    this._bufferSize = 0
+
+    this._url = new URL(format({
+      protocol: protocol || 'https:',
+      hostname: intake,
+      port: port || 443,
+      pathname: endpoint
+    }))
+
+    this._headers = {
+      'Content-Type': 'application/json'
+    }
+
+    this._periodic = setInterval(() => {
+      this.flush()
+    }, this._interval).unref()
+
+    process.once('beforeExit', () => {
+      this.destroy()
+    })
+
+    this._destroyed = false
+
+    logger.debug(`Started ${this.constructor.name} to ${this._url}`)
+  }
+
+  append (event, byteLength) {
+    if (this._buffer.length >= this._bufferLimit) {
+      logger.warn(`${this.constructor.name} event buffer full (limit is ${this._bufferLimit}), dropping event`)
+      return
+    }
+
+    this._bufferSize += byteLength || Buffer.from(JSON.stringify(event)).byteLength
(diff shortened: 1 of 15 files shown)


packages/dd-trace/src/llmobs/writers/evaluations.js
```diff
@@ -0,0 +1,28 @@
+'use strict'
+
+const BaseWriter = require('./base')
+
+class LLMObsEvalMetricsWriter extends BaseWriter {
+  constructor (config) {
+    super({
+      endpoint: '/api/intake/llm-obs/v1/eval-metric',
```
Yun-Kim, line 8: (noticed this is the same as in the Python SDK), let's move this endpoint to the constants file for easier maintenance

packages/dd-trace/src/llmobs/writers/spans/agentless.js
```diff
@@ -0,0 +1,16 @@
+'use strict'
+
+const LLMObsBaseSpanWriter = require('./base')
+
+class LLMObsAgentlessSpanWriter extends LLMObsBaseSpanWriter {
+  constructor (config) {
+    super({
+      intake: `llmobs-intake.${config.site}`,
+      endpoint: '/api/v2/llmobs'
```
Yun-Kim, line 9: Same here, let's move this to the constants file

packages/dd-trace/src/llmobs/writers/base.js
```diff
@@ -0,0 +1,108 @@
...
+      this._destroyed = true
+    }
+  }
+
+  _encode (payload) {
+    return JSON.stringify(payload, (key, value) => {
+      if (typeof value === 'string') {
+        return encodeUnicode(value) // serialize unicode characters
+      }
+      return value
```
Yun-Kim, line 103: Just for clarification, can you explain what exactly's happening here? Does json.stringify() get called first then we run the encodeUnicode() helper on the result afterwards?

Yun-Kim (approved):
LGTM from team mlobs, just some small suggestions / clarification questions

packages/dd-trace/src/llmobs/writers/base.js
sabrenner: it gets run as `JSON.stringify` is happening. when passing a callback function to `JSON.stringify`, it'll execute that function over any values in the object. since we need to encode unicode characters (ie `–` &rarr; `\u2013`) for our decoder on ingestion, this function will make sure we encode those special characters with the correct unicode value (I think `json.dumps` does this for us on the Python SDK, but `JSON.stringify` doesn't do it by default here). There might be a better approach for this, will wait for Node.js folks input on that.

rochdev: Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#replacer

rochdev (approved):

PR #4700, opened 2024-09-18
capping version ranges for plugin test in ci

crysmags:
### What does this PR do?
This will cap versions ranges tested for plugin CI tests. The current implementation is to cap the version at the last successful version tested. The instrumentations have minimum versions which also are tested. 

This Pull Request does not cover, older major versions with new releases. 

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -265,8 +265,6 @@ jobs:
     strategy:
       matrix:
         node-version: [16]
-        // each comma should create a new test such that we test node 16
-        // against 2.6.12 and also node 16 against 3.0.7 & latest of major 3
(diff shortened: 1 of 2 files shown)


BridgeAR (commented):
I understand the intent, I actually do believe it is however also valuable to have the CI fail so that we at least A) know about the issue and B) can decide if it's an easy fix or if we just want to pin the version at that point.

BridgeAR:
I'll go ahead and close this for now since it has been open for quite a while.
Please feel free to reopen in case you would like to continue work on it. I do however believe this might not be what we want. I guess it would be a great topic to discuss again in the guild.

PR #4705, opened 2024-09-19
[test visibility] Remove beta flag from manual test API

juan-fernandez:
### What does this PR do?
* Change `DD_CIVISIBILITY_MANUAL_API_ENABLED` default value to `true`, so that it only acts as a kill switch.
* The manual API will be available unless `DD_CIVISIBILITY_MANUAL_API_ENABLED` is set to `false`. 

### Motivation
We're not making changes to it so it doesn't make sense to call it beta.

### Plugin Checklist
N/A

### DIFF
--- a/packages/dd-trace/test/config.spec.js
+++ b/packages/dd-trace/test/config.spec.js
@@ -1824,11 +1824,7 @@ describe('Config', () => {
         const config = new Config(options)
         expect(config).to.have.property('isIntelligentTestRunnerEnabled', false)
       })
-      it('should disable manual testing API by default', () => {
-        const config = new Config(options)
-        expect(config).to.have.property('isManualApiEnabled', false)
-      })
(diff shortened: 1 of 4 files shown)


Qard (dismissed):

packages/dd-trace/test/config.spec.js
```diff
@@ -1824,14 +1824,14 @@ describe('Config', () => {
         const config = new Config(options)
         expect(config).to.have.property('isIntelligentTestRunnerEnabled', false)
       })
-      it('should disable manual testing API by default', () => {
+      it('should enabled manual testing API by default', () => {
```
romainkomorndatadog, line 1827:
```suggestion
      it('should enable manual testing API by default', () => {
```

romainkomorndatadog (approved):

packages/dd-trace/test/config.spec.js
juan-fernandez: good catch!

anmarchenko (approved):

PR #4712, opened 2024-09-20
Sql injection Exploit Prevention implementation for mysql2 library

uurien:
### What does this PR do?

Instrument new methods in mysql2 to be able to detect the exploit when we are in a stack with app code. We need to have app code file and line in the stack to be able to share the exploited line.
In these new points, we support abortcontroller to stop sql execution when a exploit is detected.

### Motivation

Detect and stop sql injection attacks.

### Checklist

- [x] Unit tests.

### DIFF
--- a/packages/datadog-instrumentations/src/mysql2.js
+++ b/packages/datadog-instrumentations/src/mysql2.js
@@ -226,11 +226,12 @@ addHook({ name: 'mysql2', file: 'lib/pool.js', versions: ['>=1'] }, (Pool, versi
 // PoolNamespace.prototype.query does not exist in mysql2<2.3.0
 addHook({ name: 'mysql2', file: 'lib/pool_cluster.js', versions: ['>=2.3.0'] }, PoolCluster => {
   const startOuterQueryCh = channel('datadog:mysql2:outerquery:start')
+  const wrappedPoolNamespaces = new WeakSet()
 
   shimmer.wrap(PoolCluster.prototype, 'of', of => function () {
     const poolNamespace = of.apply(this, arguments)
 
-    if (startOuterQueryCh.hasSubscribers) {
+    if (startOuterQueryCh.hasSubscribers && !wrappedPoolNamespaces.has(poolNamespace)) {
       shimmer.wrap(poolNamespace, 'query', query => function (sql, values, cb) {
         if (typeof sql === 'object') sql = sql?.sql
 
@@ -265,9 +266,7 @@ addHook({ name: 'mysql2', file: 'lib/pool_cluster.js', versions: ['>=2.3.0'] }, 
 
         return query.apply(this, arguments)
       })
-    }
 
-    if (startOuterQueryCh.hasSubscribers) {
       shimmer.wrap(poolNamespace, 'execute', execute => function (sql, values, cb) {
         if (typeof sql === 'object') sql = sql?.sql
 
@@ -290,6 +289,8 @@ addHook({ name: 'mysql2', file: 'lib/pool_cluster.js', versions: ['>=2.3.0'] }, 
 
         return execute.apply(this, arguments)
       })
+
+      wrappedPoolNamespaces.add(poolNamespace)
     }
 
     return poolNamespace
--- a/packages/datadog-instrumentations/src/mysql2.js
+++ b/packages/datadog-instrumentations/src/mysql2.js
@@ -197,7 +197,11 @@ addHook({ name: 'mysql2', file: 'lib/pool.js', versions: ['>=1'] }, (Pool, versi
   })
 
   shimmer.wrap(Pool.prototype, 'execute', execute => function (sql, values, cb) {
-    if (!startOuterQueryCh.hasSubscribers || !sql) return execute.apply(this, arguments)
+    if (!startOuterQueryCh.hasSubscribers) return execute.apply(this, arguments)
+
+    if (typeof sql === 'object') sql = sql?.sql
+
+    if (!sql) return execute.apply(this, arguments)
 
     const abortController = new AbortController()
     startOuterQueryCh.publish({ sql, abortController })
--- a/packages/datadog-instrumentations/src/mysql2.js
+++ b/packages/datadog-instrumentations/src/mysql2.js
@@ -6,11 +6,15 @@ const {
   AsyncResource
 } = require('./helpers/instrument')
 const shimmer = require('../../datadog-shimmer')
+const semver = require('semver')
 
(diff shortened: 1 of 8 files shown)


packages/dd-trace/src/appsec/channels.js
```diff
@@ -24,5 +24,6 @@ module.exports = {
   setUncaughtExceptionCaptureCallbackStart: dc.channel('datadog:process:setUncaughtExceptionCaptureCallback:start'),
   pgQueryStart: dc.channel('apm:pg:query:start'),
   pgPoolQueryStart: dc.channel('datadog:pg:pool:query:start'),
+  mysql2OuterQueryStart: dc.channel('datadog:mysql2:outerquery:start'),
```
uurien, line 27: Open for suggestions for a better name...

packages/datadog-instrumentations/src/mysql2.js
```diff
@@ -28,6 +31,76 @@ addHook({ name: 'mysql2', file: 'lib/connection.js', versions: ['>=1'] }, Connec
     return asyncResource.bind(addCommand, this).apply(this, arguments)
   })
 
+  shimmer.wrap(Connection.prototype, 'query', query => function (sql, values, cb) {
+    if (!startOuterQueryCh.hasSubscribers) return query.apply(this, arguments)
+
+    const sqlIsString = typeof sql === 'string'
+
+    const sqlString = sqlIsString ? sql : sql?.sql
```
CarlesDD, line 39: In other parts of this PR, this check is done differently. I think it would be worthwhile to homogenize all of them.

[Check 1](https://github.com/DataDog/dd-trace-js/pull/4712/files#diff-30989a4234a7039a68994bf4e35c149c3ef115c24b65682c793ce6301467168bR79)
[Check 2](https://github.com/DataDog/dd-trace-js/pull/4712/files#diff-30989a4234a7039a68994bf4e35c149c3ef115c24b65682c793ce6301467168bR163)
[Check 3](https://github.com/DataDog/dd-trace-js/pull/4712/files#diff-30989a4234a7039a68994bf4e35c149c3ef115c24b65682c793ce6301467168bR230)
[Check 4](https://github.com/DataDog/dd-trace-js/pull/4712/files#diff-30989a4234a7039a68994bf4e35c149c3ef115c24b65682c793ce6301467168bR264)

packages/datadog-instrumentations/src/mysql2.js
```diff
@@ -79,3 +152,138 @@ addHook({ name: 'mysql2', file: 'lib/connection.js', versions: ['>=1'] }, Connec
...
+
+      return query.apply(this, arguments)
+    }
+  })
+
+  shimmer.wrap(Pool.prototype, 'execute', execute => function (sql, values, cb) {
+    if (!startOuterQueryCh.hasSubscribers) return execute.apply(this, arguments)
+
+    const abortController = new AbortController()
+    startOuterQueryCh.publish({ sql, abortController })
```
CarlesDD, line 202: In some other places, a check is done to assure there is actually a query (like [here](https://github.com/DataDog/dd-trace-js/pull/4712/files#diff-30989a4234a7039a68994bf4e35c149c3ef115c24b65682c793ce6301467168bR164)). Is there any reason to not do this check here? (and [here](https://github.com/DataDog/dd-trace-js/pull/4712/files#diff-30989a4234a7039a68994bf4e35c149c3ef115c24b65682c793ce6301467168bR164) and [here](https://github.com/DataDog/dd-trace-js/pull/4712/files#diff-30989a4234a7039a68994bf4e35c149c3ef115c24b65682c793ce6301467168bR164))

uurien: I changed all the cases to not call to publish without sql, just to prevent creating unnecessary AbortControllers.

packages/datadog-instrumentations/src/mysql2.js
```diff
@@ -79,3 +152,144 @@ addHook({ name: 'mysql2', file: 'lib/connection.js', versions: ['>=1'] }, Connec
...
+
+// PoolNamespace.prototype.query does not exist in mysql2<2.3.0
+addHook({ name: 'mysql2', file: 'lib/pool_cluster.js', versions: ['>=2.3.0'] }, PoolCluster => {
+  const startOuterQueryCh = channel('datadog:mysql2:outerquery:start')
+
+  shimmer.wrap(PoolCluster.prototype, 'of', of => function () {
+    const poolNamespace = of.apply(this, arguments)
+
+    if (startOuterQueryCh.hasSubscribers) {
+      shimmer.wrap(poolNamespace, 'query', query => function (sql, values, cb) {
```
iunanua, line 230: wondering if we could extract all the common logic to a method to avoid repetition, because it is almost? the same in 3 cases.

uurien: the methods are quite similar, but I'm not sure if it will make it harder to understand. The wrapped and unwrapped methods are different, and the code is not super long, that the reason why I decided to not extract it to a method.

packages/datadog-instrumentations/test/mysql2.spec.js
```diff
@@ -0,0 +1,718 @@
...
+      afterEach((done) => {
+        connection.end(() => done())
+      })
+
+      describe('Connection.prototype.query', () => {
+        describe('with string as query', () => {
+          describe('with callback', () => {
+            it('should abort the query on abortController.abort()', (done) => {
+              startCh.subscribe(abort)
+              const query = connection.query(sql, (err, _) => {
```
CarlesDD, line 77: Nit: Remove unused `_` parameter.

Applies for several cases along the code in this file.

packages/datadog-instrumentations/test/mysql2.spec.js
```diff
@@ -0,0 +1,718 @@
...
+                assert.propertyVal(err, 'message', 'Test')
+                sinon.assert.notCalled(apmQueryStart)
+
+                if (!shouldEmitEndAfterQueryAbort) done()
+              })
+
+              query.on('end', () => done())
```
CarlesDD, line 84:
```suggestion
              query.on('end', done)
```

CarlesDD: Multiple occurrences in this file

uurien: I don't like it, if `end` event send an object, `done(obj)` will be executed, and it is a test failure.

CarlesDD (approved):

iunanua (approved):

PR #4716, opened 2024-09-23
Add Plugin for @azure/functions

duncanpharvey:
### What does this PR do?

Adds plugin for [@azure/functions](https://www.npmjs.com/package/@azure/functions) package.

### Motivation

Create a root span for Azure Function invocations via HTTP requests.

https://datadoghq.atlassian.net/browse/SVLS-4929

### Plugin Checklist

- [x] ~Unit~ Integration tests.
- [x] TypeScript [tests][2].
- [x] API [documentation][3].
- [x] GitHub [jobs/workflows][4].

No unit tests but there is an integration test. Unit testing is cumbersome due to how Azure Functions can be tested locally. They must be started using the `func start` Azure Functions Core Tools cli command.

### Additional Notes

- Builds off the plugin work started by @gord02 in #4599 
- Uses [web utils](https://github.com/DataDog/dd-trace-js/blob/b079c6f93f567e9e464fffd8fe2624d6760fc4f2/packages/dd-trace/src/plugins/util/web.js) to avoid rewriting logic for http related span tags
- Only http related triggers are instrumented as part of this PR. Other triggers (CosmosDB, EventGrid, EventHub, Service Bus Queue, Service Bus Topic, Storage Blob, Storage Queue, and Timer) will be instrumented in future PRs.
  * https://learn.microsoft.com/en-us/javascript/api/@azure/functions/app?view=azure-node-latest


packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -0,0 +1,84 @@
...
+  configure (config) {
+    return super.configure(web.normalizeConfig(config))
+  }
+}
+
+function extractPath (url) {
+  const regex = /https?:\/\/[^/]+(\/.*$)/
+  const match = url.match(regex)
+  return match ? match[1] : ''
+}
```
tlhunter, line 78: I'd recommend using the built-in URL parsers instead of rolling one ourselves:

```js
extractPath('http://www.google.com/asdf?asdf')
// '/asdf?asdf'
url.parse('http://www.google.com/asdf?asdf').path // if search query is needed
// '/asdf?asdf'
(new URL('http://www.google.com/asdf?asdf')).pathname // if search query isn't needed
// '/asdf'
```

packages/dd-trace/src/plugins/util/web.js
```diff
@@ -427,11 +427,13 @@ function addRequestTags (context) {
...
   const url = extractURL(req)
 
+  const tags = span.context()._tags
+
   span.addTags({
     [HTTP_URL]: web.obfuscateQs(config, url),
     [HTTP_METHOD]: req.method,
     [SPAN_KIND]: SERVER,
-    [SPAN_TYPE]: WEB,
+    [SPAN_TYPE]: tags[SPAN_TYPE] === 'serverless' ? 'serverless' : WEB,
```
tlhunter, line 436: This seems unfortunate. Should there be a `packages/dd-trace/src/plugins/util/serverless.js` file instead?

rochdev: Why was this even changed at all for Serverless? It seems like the span type is one of the oldest properties of a span with a special meaning dating back to the very beginning of APM. The Serverless backend should respect those semantics and add a property if it needs more information.

.gitignore
```diff
@@ -125,3 +125,4 @@ packages/dd-trace/test/appsec/next/*/package.json
 packages/dd-trace/test/appsec/next/*/node_modules
 packages/dd-trace/test/appsec/next/*/yarn.lock
 !packages/dd-trace/**/telemetry/logs
+packages/datadog-plugin-azure-functions/test/integration-test/fixtures/node_modules
```
rochdev, line 128: I think this should already be the case by default?

index.d.ts
```diff
@@ -1229,6 +1230,12 @@ declare namespace tracer {
       [key: string]: boolean | Object | undefined;
     }
 
+    /**
+     * This plugin automatically instruments the
+     * azure.functions module.
```
rochdev, line 1235:
```suggestion
     * @azure/functions module.
```

packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -0,0 +1,84 @@
...
+      url: path
+    }
+
+    const context = web.patch(req)
+    context.span = ctx.currentStore.span
+    context.config = this.config
+    context.paths = [path]
+
+    // Use status for status code if available. Otherwise if no status is provided assume an internal server error
+    context.res = { statusCode: result.hasOwnProperty('status') ? result.status : 500 }
```
rochdev, line 64: Not sure about this default. At least for servers, a missing status code doesn't mean that there was an error. Why not just leave it empty instead of trying to interpret the meaning of a missing status?

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
```diff
@@ -0,0 +1,111 @@
...
+  return spawnProc(command, args, {
+    cwd,
+    env
+  }, stdioHandler)
+}
+
+function spawnProc (command, args, options = {}, stdioHandler, stderrHandler) {
+  const proc = spawn(command, args, { ...options, stdio: 'pipe' })
+  return new Promise((resolve, reject) => {
+    waitOn({
```
rochdev, line 77: Why is this needed? It adds a dependency and doesn't really seem necessary since the rest of the code just waits for output.

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
```diff
@@ -0,0 +1,111 @@
...
+      this.timeout(50000)
+      await sandbox.remove()
+    })
+
+    beforeEach(async () => {
+      agent = await new FakeAgent().start()
+    })
+
+    afterEach(async () => {
+      const azureFuncProc = await findProcess('name', 'func', true)
```
rochdev, line 36: This shouldn't be needed if the process is started within our code as we should then be able to get the PID as it's started.

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
```diff
@@ -0,0 +1,111 @@
...
+      const azureFuncProcPid = azureFuncProc[0]?.pid ?? null
+      azureFuncProcPid !== null && process.kill(azureFuncProcPid, 'SIGKILL')
```
rochdev, line 38: Could the process exit gracefully instead?

rochdev (changes requested):

.gitignore
duncanpharvey: Yes but this line overrides that default and includes `node_modules` for some test directories

https://github.com/DataDog/dd-trace-js/blob/6b405c7754ba5f67fa7620fa89e6e6d40c18fbfc/.gitignore#L120

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
duncanpharvey: Are you asking if `waitOn` is necessary? Without waiting for the resource to be up there is this connection error.

```
Error: connect ECONNREFUSED 127.0.0.1:7071
```

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
duncanpharvey: Running the `func start` command actually spawns another process. This was the best method I found to find this spawned Azure Function process to kill it after each test. I also tried using [tree-kill](https://www.npmjs.com/package/tree-kill) but couldn't get it working when running the tests locally on my mac.

```
#! /usr/bin/env node

const path = require('path');
const spawn = require('child_process').spawn;
const args = process.argv;

function main() {
    const bin = path.resolve(path.join(path.dirname(__filename), '..', 'bin'));
    const funcProc = spawn(bin + '/func', args.slice(2), {
        stdio: [process.stdin, process.stdout, process.stderr, 'pipe']
    });

    funcProc.on('exit', code => {
        process.exit(code);
    });
}

main();
```

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
duncanpharvey: I'm not sure why but the Azure Function process doesn't stop when using a `SIGTERM` signal. It only stops with the `SIGKILL` signal.

packages/dd-trace/src/plugins/util/web.js
duncanpharvey: We've been setting the span type to `serverless` on AWS Lambda functions for some time now. However this is currently happening in the Lambda layer. Azure Functions don't have the same concept of a layer where dependencies can be added so that's why it's being added as a plugin directly in the tracer.

https://github.com/DataDog/datadog-lambda-js/blob/e84f01785b2e761eba6f91e44dc277c1e7a78881/src/trace/listener.ts#L295-L296

duncanpharvey: I added a new `SERVERLESS` property and also moved this type logic to `util/serverless.js`. I also needed to copy a few of the helper methods over from `util/web.js` to get this working. Let me know if you had another solution in mind.

https://github.com/DataDog/dd-trace-js/blob/bcd230e7ed58786c70df4d08d1f6700ed7f2eceb/ext/types.js#L5

packages/datadog-plugin-azure-functions/src/index.js
duncanpharvey: Updated to leave the status empty 👍

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
rochdev: It seems the process is actually `func`? Or are you saying calling `func` also spawns a different `func`?

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
rochdev: What I meant is to not need to send it a signal and let it exit gracefully, although I don't know if it's meant to be able to do that. If it's not then this is fine.

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
rochdev: Why though? Can't the child process just signal to the parent process when it's ready instead of having to poll using a new dependency?

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
duncanpharvey: Correct, spawning `func` spawns a different `func`. The above code I shared is the code from Azure Functions Core Tools. So in the integration test `proc.pid` is a different value than `azureFuncProcPid`.

I'm happy to consider other methods that don't require adding another dev dependency! But this was the only reliable approach I could find while testing.

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
duncanpharvey: I removed the `wait-on` dependency by checking for when the output from the `func start` command shows that the endpoint is ready. Let me know if that's what you meant.

https://github.com/DataDog/dd-trace-js/blob/8a81f3ce5e4fd6c6f6f7a99084284d03b77f7c71/packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js#L92-L94

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
rochdev: It doesn't look detached, so wouldn't killing the parent process also kill the child process?

duncanpharvey: It is attached, but for some reason I don't understand killing the parent process (PID 26515) does not kill the child processes. I'll do some more exploration to see if I can find a method to kill the child processes that avoids adding a new dev dependency.

```
USER            PID  PPID  PGID   SESS JOBC STAT   TT       TIME COMMAND
duncan.harvey 26491 11483 26491      0    1 S+   s001    0:00.15 node /opt/homebrew/Cellar/yarn/1.22.22/libexec/bin/yarn.js test:integration:plugins
duncan.harvey 26493 26491 26491      0    1 S+   s001    0:00.46 /Users/duncan.harvey/.nvm/versions/node/v20.10.0/bin/node /Users/duncan.harvey/go/src/github.com/DataDog/dd-trace-js/node_modules/.bin/mocha -r packages/dd-trace/test/setup/mocha.js packages/datadog-plugin-@(azure-functions)/test/integration-test/**/*.spec.js
duncan.harvey 26515 26493 26491      0    1 S+   s001    0:00.06 /Users/duncan.harvey/.nvm/versions/node/v20.10.0/bin/node /Users/duncan.harvey/go/src/github.com/DataDog/dd-trace-js/packages/datadog-plugin-azure-functions/test/integration-test/fixtures/node_modules/.bin/func start
duncan.harvey 26516 26515 26491      0    1 S+   s001    0:01.38 /Users/duncan.harvey/go/src/github.com/DataDog/dd-trace-js/packages/datadog-plugin-azure-functions/test/integration-test/fixtures/node_modules/azure-functions-core-tools/bin/func start
duncan.harvey 26589 26516 26491      0    1 S+   s001    0:00.34 /Users/duncan.harvey/.nvm/versions/node/v20.10.0/bin/node /Users/duncan.harvey/go/src/github.com/DataDog/dd-trace-js/packages/datadog-plugin-azure-functions/test/integration-test/fixtures/node_modules/azure-functions-core-tools/bin/workers/node/dist/src/nodejsWorker.js --host 127.0.0.1 --port 56210 --workerId 1a8f50d0-e274-47f3-af07-664cf84029b2 --requestId 3f7963a5-de29-481f-acc3-7cce7af17ede --grpcMaxMessageLength 2147483647 --functions-uri http://127.0.0.1:56210/ --functions-worker-id 1a8f50d0-e274-47f3-af07-664cf84029b2 --functions-request-id 3f7963a5-de29-481f-acc3-7cce7af17ede --functions-grpc-max-message-length 2147483647
```

duncanpharvey: I was able to remove the `find-process` dependency by calling the nested `func` command directly. Now that I have that pid I can easily kill the process at the end of the test.

packages/dd-trace/src/plugins/util/web.js
rochdev: Could `serverless.js` just import web.js instead and only override the span type?

duncanpharvey: Since the `addRequestTags` helper function isn't exported, I needed to pass span type to it as a parameter. Now `serverless.js` just imports `web.js` and overrides the `SPAN_TYPE` property I added.

https://github.com/DataDog/dd-trace-js/blob/ebbe75e66045a3558a11be8ca58376749d6f881c/packages/dd-trace/src/plugins/util/serverless.js#L1-L7

 I think this is the best approach for making `serverless.js` as simple as possible with minimal changes to `web.js`. Let me know if you think otherwise or have more suggestions for improving the code.

rochdev (approved):

jcstorms1 (approved):

Issue #3719, opened 2023-10-13
Allow Datadog trace headers when using AWS IAM authentication


ryan-breidenbach:
Currently, dd-trace [explicitly omits trace HTTP headers](https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-plugin-http/src/client.js#L61-L63) if a request uses AWS IAM authentication. This behavior was introduced by [this change](https://github.com/DataDog/dd-trace-js/pull/205) that was made in 2018.

The comments in the change suggest that adding Datadog headers will break the authentication of the signature and cause `403` errors. However, this is not the case. When the [signature is created](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html), the headers used to calculate the signature are captured in the `SignedHeaders` component of the `Authorization` header. Any additional headers that are added after the signature is generated will be ignored.

We work with many applications that use IAM authentication for service-to-service communication. Not being able to see traces span all of our applications is hindering our ability to support and troubleshoot these applications.  I would like the ability to at least "opt-in" for allowing headers when using IAM authentication.

tlhunter:
I'm reverting the commit that I made to re-enable these headers on AWS requests:
https://github.com/DataDog/dd-trace-js/pull/3799

Both @rochdev and @astuyve believe the change will cause a lot of failures. Will need to build out better tests to figure this out.

astuyve:
Hey I'm sorry there was some confusion here on my end, unless the headers are explicitly part of the signature (present in `SignedHeaders`), it should be possible to add additional headers after signing without causing issues. I think #3796 would be okay, and we can test this in an integration test in something like Lambda to verify. But it seems like we had some reported issues, so I'd be curious to see what those are before making any further changes.

rochdev:
> But it seems like we had some reported issues

That's the main concern, we used to not handle this at all and we received several reports that applications were starting to error. Since there hasn't been any changes to the signature version, I would expect a regression of any cases that were not working before we introduced the check.

janos-kasa:
We were able to use this version and all requests correctly went through the IAM Authentication between our services, and DD was able to connect the trace.
I also support the configuration flag idea.

PR #4717, opened 2024-09-23
always enable tracing header injection for AWS requests

tlhunter:
### What does this PR do?
- always injects aws tracing headers
- see #3796, #205, #3719 for more information

### Motivation
- I don't believe this has caused issues for users in years
- users are complaining that this isn't the default behavior
- users are unable to programmatically change config in certain environments
  - technically this could be solved by adding another env var

### Plugin Checklist

- [x] Unit tests.
- [x] TypeScript [tests][2].
- [ ] API [documentation][3].


Qard (approved):

wconti27 (approved):

Romack:
Just want to point out that removing `enablePropagationWithAmazonHeaders` from the types file is a breaking change and should've been a major version update.  Any typescript projects that that have implemented this "workaround" (e.g. us) will receive the following error if/when they upgrade dd-trace to version 5.23.1

Object literal may only specify known properties, and 'enablePropagationWithAmazonHeaders' does not exist in type 'http'.

alxgrk:
Unfortunately, this change broke our production environment, because it made AWS requests fail. Specifically, it it the following request that fails:
```js
const {S3} = require('@aws-sdk/client-s3');
const awsS3Client = new S3();
const response = await awsS3Client
      .getObject({
        Bucket: 'my-bucket',
        Key: 'my-file',
        IfNoneMatch: 'my-etag',
      });
```
We tested both, with `aws-sdk@2.884.0` as well as with `@aws-sdk/client-s3@3.673.0` - same result.

From reading the different PRs/issues, it looks like since 2018 ([here](https://github.com/DataDog/dd-trace-js/pull/205)) the exception for signed AWS requests existed. In November 2023, it got removed ([here](https://github.com/DataDog/dd-trace-js/pull/3796)), but reverted on the same day ([here](https://github.com/DataDog/dd-trace-js/pull/3799)), because some users were facing issues. For some reason that I can not yet understand, it was assumed in [this PR](https://github.com/DataDog/dd-trace-js/pull/4717) that the errors won't show up again. Is there any reason that led to this assumption?
As a workaround, we reverted to 5.22.0, but of course this error prevents us from upgrading.

PS: I have also submitted a helpdesk request, but wanted to make this problem visible to others as well.

tlhunter:
@alxgrk from the research I had done it appeared that Amazon had fixed the bug which caused any arbitrary header to affect the signature. I didn't realize this was only fixed with SDK v3 and not with v2.

We'll work on reverting this change, perhaps just for SDK v2 and not for v3.

alxgrk:
thanks @tlhunter , sounds like a plan 👍

tmercswims:
This change broke our production as well. We still use aws-sdk v2, and get a signature mismatch error on requests to DynamoDB. Downgrading to 5.22.0 fixes the errors.

PR #4718, opened 2024-09-24
[MLOB-1556] LLM Observability tagger

sabrenner:
### What does this PR do?
Adds a tagger for the LLM Observability product. Responsible for setting temporary tags that will later be removed as part of the `span_processor` and `format` steps in a separate PR.

### Motivation
Adds more functionality to LLM Observability in preparation for a feature release in an upcoming PR, with the squashed commit of this PR being on of the few commits in that PR.

**Note**: will likely have some merge conflicts here once other PRs are merged

### DIFF
--- a/packages/dd-trace/src/llmobs/tagger.js
+++ b/packages/dd-trace/src/llmobs/tagger.js
@@ -93,10 +93,11 @@ class LLMObsTagger {
   }
 
   tagSpanTags (span, tags) {
+    // new tags will be merged with existing tags
     try {
       const currentTags = span.context()._tags[TAGS]
       if (currentTags) {
-        Object.assign(tags, currentTags)
+        Object.assign(tags, JSON.parse(currentTags))
       }
       span.setTag(TAGS, JSON.stringify(tags))
     } catch {
@@ -110,7 +111,6 @@ class LLMObsTagger {
         span.setTag(key, data)
       } else {
         try {
-          // this will help showcase unfinished promises being passed in as values
           span.setTag(key, JSON.stringify(data))
         } catch {
           const type = key === INPUT_VALUE ? 'input' : 'output'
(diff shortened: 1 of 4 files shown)


packages/dd-trace/src/llmobs/tagger.js
```diff
@@ -0,0 +1,206 @@
...
+    sessionId = sessionId || parentLLMObsSpan?.context()._tags[SESSION_ID]
+    if (sessionId) span.setTag(SESSION_ID, sessionId)
+
+    if (!mlApp) mlApp = parentLLMObsSpan?.context()._tags[ML_APP] || this._config.llmobs.mlApp
+    span.setTag(ML_APP, mlApp)
+
+    const parentId =
+      parentLLMObsSpan?.context().toSpanId() ||
+      span.context()._trace.tags[PROPAGATED_PARENT_ID_KEY] ||
+      'undefined'
```
Yun-Kim, line 54: Let's make `undefined` a constant

packages/dd-trace/test/llmobs/tagger.spec.js
```diff
@@ -0,0 +1,430 @@
...
+        'span.type': 'llm',
+        '_ml_obs.meta.span.kind': 'llm',
+        '_ml_obs.meta.ml_app': 'my-ml-app',
+        '_ml_obs.session_id': 'my-session',
+        '_ml_obs.llmobs_parent_id': '5678'
+      })
+    })
+
+    it('uses the propagated trace id if provided', () => {
+      spanContext._trace.tags['_dd.p.llmobs_trace_id'] = '-123'
```
Yun-Kim, line 127: Are we asserting trace ID here? I can't see it 😕

Yun-Kim (commented):
Are `setLLMObsSpanTags`, etc public user-facing SDK methods? Or are we only using them internally in our integrations/SDK?

packages/dd-trace/test/llmobs/tagger.spec.js
sabrenner: oops yeah it's because i removed the test for now but didn't remove this :p good catch!

sabrenner:
> Are setLLMObsSpanTags, etc public user-facing SDK methods? Or are we only using them internally in our integrations/SDK?

@Yun-Kim This class will be used on "private" (ie, `someObj._tagger`) variables. these variables will be accessible from the tracer itself, but most of our "private" variables are, they just aren't documented as they're not part of the public API. I think, sticking with that behavior for the tracer, this will only be used on underscore variables and not be documented publicly.

Yun-Kim (approved):
LGTM from the llmobs side of things. As long as we're being consistent with the tracer usage and LLMObs usage (and are intentional about what methods we provide public documentation for) I have no issues 👍

packages/dd-trace/test/llmobs/tagger.spec.js
```diff
@@ -0,0 +1,428 @@
...
+    })
+
+    it('logs when metrics is not JSON serializable', () => {
+      const metadata = unserializbleObject()
+      tagger.tagMetadata(span, metadata)
+      expect(logger.warn).to.have.been.calledOnce
+    })
+  })
+
+  describe('tagLLMIO', () => {
```
rochdev, line 187: These tests mostly only test that private methods are called when a public method is called. This doesn't provide a lot of confidence that the feature actually work. Could the output be tested instead? This could be the tags on the span itself, or even better, what will actually be sent (proving that the tags also properly cause the span to be split in 2 representations for tracing and llmobs).

rochdev (changes requested):

packages/dd-trace/test/llmobs/tagger.spec.js
sabrenner: ah yeah, that's a good idea. i think i'll opt for checking the tags directly. i plan on adding a more integration-style/E2E testing once I have the SDK PR ready (it'll be a big one 😓)

rochdev (approved):

tlhunter:
Hmm... Looks like this merged with failures?

rochdev:
> Hmm... Looks like this merged with failures?
Not sure how too as `all-green` is in the list of required checks 🤔

tlhunter:
LOL MY BAD it was not merged into master

rochdev:
Although it does look like unrelated failures that have already been merged on the main branch in the past few days.

> LOL MY BAD it was not merged into master

Oh yeah totally missed that too, I guess we don't use feature branches often 😅

sabrenner:
😭 sorry for the confusion here, i was planning on rebasing my feature branch after this merged. and yeah, won't go with a feature branch probably next time 😨

PR #4723, opened 2024-09-25
[test visibility] Early flake detection for mocha (parallel mode)

juan-fernandez:
### What does this PR do?

Add early flake detection logic when `--parallel` is passed to mocha.

Additionally, add the faulty threshold logic to mocha, which was missing: if too many tests are considered new, we'll bail out of EFD.

### Motivation
Allow users of the parallel mode of mocha to use early flake detection.

### Plugin Checklist

- [x] Unit tests.

### DIFF
--- a/packages/dd-trace/test/plugins/externals.json
+++ b/packages/dd-trace/test/plugins/externals.json
@@ -266,7 +266,7 @@
   "mocha": [
     {
       "name": "mocha",
-      "versions": [">=5.2.0"]
+      "versions": [">=5.2.0", ">=8.0.0"]
     },
     {
       "name": "mocha-each",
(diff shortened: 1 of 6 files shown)


packages/dd-trace/test/plugins/externals.json
```diff
@@ -266,7 +266,7 @@
   "mocha": [
     {
       "name": "mocha",
-      "versions": [">=5.2.0"]
+      "versions": [">=5.2.0", ">=8.0.0"]
```
juan-fernandez, line 269: ℹ️ not picked up automatically by `install_plugin_modules` because I've spread the instrumentation of mocha in multiple files. I could probably fix `install_plugin_modules.js` but this is just simpler

rochdev: This is actually being addressed in a separate effort to separate tests and test setup from the versions in the implementation.

rochdev (approved):

PR #4725, opened 2024-09-25
Implement Config Consistency

khanayan123:
### What does this PR do?

This pull request updates configuration defaults and behaviors to ensure alignment with standardized definitions across SDKs

changes:

- Introduce support for DD_TRACE_`<INTEGRATION>`_ENABLED, allowing the enablement or disablement of specific integrations based on the provided integration name.
- Set the default value for DD_TRACE_RATE_LIMIT to 100.
- Add DD_TRACE_ENABLED to the configuration object to enable transmission via telemetry.
- Ensure that leading and trailing white spaces are trimmed from items in DD_TRACE_HEADER_TAGS.
- Update the behavior of the version tag such that when a span is created with a service name different from the global service name, it will not inherit the global version value.
- Update configuration settings listed in the RFC to send telemetry using their respective environment variable names

### Motivation

The changes are driven by the need to meet the requirements outlined in the configuration consistency [RFC ](https://docs.google.com/document/d/1kI-gTAKghfcwI7YzKhqRv2ExUstcHqADIWA4-TZ387o/edit?usp=sharing)

### Additional Notes

This PR was tested against the system tests [here](https://github.com/DataDog/system-tests/pull/3144)

### DIFF
--- a/packages/datadog-instrumentations/src/helpers/register.js
+++ b/packages/datadog-instrumentations/src/helpers/register.js
@@ -25,7 +25,7 @@ const disabledInstrumentations = new Set(
 // Check for DD_TRACE_<INTEGRATION>_ENABLED environment variables
 for (const [key, value] of Object.entries(process.env)) {
   const match = key.match(/^DD_TRACE_(.+)_ENABLED$/)
-  if (match && value.toLowerCase() === 'false') {
+  if (match && (value.toLowerCase() === 'false' || value === '0')) {
     const integration = match[1].toLowerCase()
     disabledInstrumentations.add(integration)
   }
--- a/packages/datadog-instrumentations/src/helpers/register.js
+++ b/packages/datadog-instrumentations/src/helpers/register.js
@@ -22,6 +22,15 @@ const disabledInstrumentations = new Set(
   DD_TRACE_DISABLED_INSTRUMENTATIONS ? DD_TRACE_DISABLED_INSTRUMENTATIONS.split(',') : []
 )
 
+// Check for DD_TRACE_<INTEGRATION>_ENABLED environment variables
+for (const [key, value] of Object.entries(process.env)) {
+  const match = key.match(/^DD_TRACE_(.+)_ENABLED$/)
+  if (match && value.toLowerCase() === 'false') {
+    const integration = match[1].toLowerCase()
+    disabledInstrumentations.add(integration)
+  }
+}
+
 const loadChannel = channel('dd-trace:instrumentation:load')
 
 // Globals
--- a/packages/dd-trace/src/config.js
+++ b/packages/dd-trace/src/config.js
@@ -790,7 +790,6 @@ class Config {
     this._envUnprocessed['sampler.rules'] = DD_TRACE_SAMPLING_RULES
     this._setString(env, 'scope', DD_TRACE_SCOPE)
     this._setString(env, 'service', DD_SERVICE || DD_SERVICE_NAME || tags.service || OTEL_SERVICE_NAME)
-    console.log('env service', env.service)
     if (DD_SERVICE_MAPPING) {
       this._setValue(env, 'serviceMapping', fromEntries(
         process.env.DD_SERVICE_MAPPING.split(',').map(x => x.trim().split(':'))
@@ -929,7 +928,6 @@ class Config {
     this._setValue(opts, 'sampler.rateLimit', coalesce(options.rateLimit, ingestion.rateLimit))
     this._setSamplingRule(opts, 'sampler.rules', options.samplingRules)
     this._setString(opts, 'service', options.service || tags.service)
-    console.log('opts service', opts.service)
     this._setValue(opts, 'serviceMapping', options.serviceMapping)
     this._setString(opts, 'site', options.site)
     if (options.spanAttributeSchema) {
@@ -1149,7 +1147,7 @@ class Config {
     }
 
     if (typeof value === 'string') {
-      // Split by commas and trim each item
+      // Split by commas and trim each item in header tags mapping
       if (name === 'headerTags') {
         value = value.split(',').map(item => {
           // Trim each item and remove whitespace around the colon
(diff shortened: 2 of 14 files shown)


packages/dd-trace/src/opentracing/tracer.js
```diff
@@ -52,8 +52,16 @@ class DatadogTracer {
...
 
+    // as per spec, allow the setting of service name through tags
     const tags = {
-      'service.name': this._service
+      'service.name': options?.tags?.service ? String(options.tags.service) : this._service
+    }
+
+    // edge case handling, if a user manually sets a service name different from the tracer's service name
+    // then we need to ensure that the version is set to undefined
+    // in every other scenario version should be set to the tracer's version
```
mtoffl01, line 62: I assume by "manually" you mean when creating a custom span and passing a service name != global service name?
Does this apply to integration service names that differ from global service name (since it's not "manual" but does have the "different service name" effect)

packages/dd-trace/test/opentracing/tracer.spec.js
```diff
@@ -245,6 +245,39 @@ describe('Tracer', () => {
       expect(span.addTags).to.have.been.calledWith(fields.tags)
     })
 
+    it('If a user sets spans service name differs from the tracers, ensure the spans version is undefined.', () => {
```
mtoffl01, line 248:
```suggestion
    it('If span is granted a service name that differs from the global service name, ensure span's `version` tag is undefined.', () => {
```

mtoffl01 (dismissed):

packages/dd-trace/src/opentracing/tracer.js
tlhunter: @khanayan123 the comment says that we need to do something but why do we need to do it?

watson (commented):
Since this PR changes default values, one could argue it's a breaking change and should be labeled with `semver-major`. However, to be pragmatic about it, I think it should just be labeled it with `semver-minor` (currently it's labeled with `semver-patch`).

packages/dd-trace/src/config.js
```diff
@@ -707,6 +709,7 @@ class Config {
     this._setBoolean(env, 'dsmEnabled', DD_DATA_STREAMS_ENABLED)
     this._setBoolean(env, 'dynamicInstrumentationEnabled', DD_DYNAMIC_INSTRUMENTATION_ENABLED)
     this._setString(env, 'env', DD_ENV || tags.env)
+    this._setBoolean(env, 'traceEnabled', DD_TRACE_ENABLED)
```
simon-id, line 712: this will never ever be false because we stop any initialization of the library if DD_TRACE_ENABLED is false here: https://github.com/DataDog/dd-trace-js/blob/bba5f3ddb3e6d693aa74584f8c73d5b5cbf4ea95/packages/dd-trace/src/index.js#L9

packages/dd-trace/src/opentracing/tracer.js
khanayan123: Updated the comment, the reason we need to do it is to align with unified service tagging spec https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging/?tab=kubernetes

`DD_VERSION (version)
Environment Variable Name: DD_VERSION
Behavior: Adds version tag to all spans that have the same service name as the default service name (aka DD_SERVICE). Spans with a different service name do not get this tag.`

`Configure spans with DD_VERSION to add version to all spans that fall under the service that belongs to the tracer (generally DD_SERVICE). This means that if your service creates spans with the name of an external service, those spans do not receive version as a tag.`

`Adds version tag to all spans that have the same service name as the default service name (aka DD_SERVICE). Spans with a different service name do not get this tag.`

packages/dd-trace/src/config.js
khanayan123: I included this primarily to enable its transmission in the telemetry payload, if though what will be set will always be true.

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -22,6 +22,15 @@ const disabledInstrumentations = new Set(
   DD_TRACE_DISABLED_INSTRUMENTATIONS ? DD_TRACE_DISABLED_INSTRUMENTATIONS.split(',') : []
 )
 
+// Check for DD_TRACE_<INTEGRATION>_ENABLED environment variables
+for (const [key, value] of Object.entries(process.env)) {
+  const match = key.match(/^DD_TRACE_(.+)_ENABLED$/)
+  if (match && value.toLowerCase() === 'false') {
```
tlhunter, line 28: In practice customers use both `false` and `0`

packages/dd-trace/src/opentracing/tracer.js
```diff
@@ -52,8 +52,15 @@ class DatadogTracer {
...
 
+    // as per spec, allow the setting of service name through options
     const tags = {
-      'service.name': this._service
+      'service.name': options?.tags?.service ? String(options.tags.service) : this._service
+    }
+
+    // As per unified service tagging spec if a span is created with a service name different from the global
+    // service name it will not inherit the global version value
+    if (options?.tags?.service && options?.tags?.service !== this._service) {
```
tlhunter, line 62:
```suggestion
    if (options?.tags?.service && options.tags.service !== this._service) {
```

packages/dd-trace/src/config.js
```diff
@@ -546,7 +547,7 @@ class Config {
     this._setValue(defaults, 'tracePropagationStyle.otelPropagators', false)
     this._setValue(defaults, 'tracing', true)
     this._setValue(defaults, 'url', undefined)
-    this._setValue(defaults, 'version', pkg.version)
```
tlhunter, line 549: Is this a breaking change? What implications does it come with?

packages/dd-trace/src/opentracing/tracer.js
tlhunter: Since at this point we already know options.tags is an object

packages/dd-trace/src/config.js
```diff
@@ -1144,7 +1147,16 @@ class Config {
     }
 
     if (typeof value === 'string') {
-      value = value.split(',')
+      // Split by commas and trim each item in header tags mapping
+      if (name === 'headerTags') {
```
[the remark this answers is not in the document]
khanayan123: I think it's fair to remove white spaces around all values in a mapping, going to remove that check for headerTags

packages/dd-trace/src/config.js
tlhunter: Revert this and I think we can land it

khanayan123: reverted

tlhunter (approved):

PR #4728, opened 2024-09-26
[DI] Switch unit tests to Mocha instead of Tap

watson:
The reason for the change are that Tap has several bugs and drawbacks:

- It throws nasty native stack trace when assertions fail in certain async situations.
- Doesn't show nice diff of expected vs actual data in deeply nested object assertions.
- No support for `.only` when using the Mocha-syntax variant.
- Output from console.logs interfere with the Tap output. It's often interweaved, cut off i the middle of a string or mis-colored (probably because of multi-threading).


watson:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4728?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4728** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4728?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4727** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4727?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

package.json
```diff
@@ -20,8 +20,8 @@
     "test:appsec:ci": "nyc --no-clean --include \"packages/dd-trace/src/appsec/**/*.js\" --exclude \"packages/dd-trace/test/appsec/**/*.plugin.spec.js\" -- npm run test:appsec",
     "test:appsec:plugins": "mocha -r \"packages/dd-trace/test/setup/mocha.js\" \"packages/dd-trace/test/appsec/**/*.@($(echo $PLUGINS)).plugin.spec.js\"",
     "test:appsec:plugins:ci": "yarn services && nyc --no-clean --include \"packages/dd-trace/src/appsec/**/*.js\" -- npm run test:appsec:plugins",
-    "test:debugger": "tap packages/dd-trace/test/debugger/**/*.spec.js",
-    "test:debugger:ci": "npm run test:debugger -- --coverage --nyc-arg=--include=\"packages/dd-trace/src/debugger/**/*.js\"",
+    "test:debugger": "mocha -r 'packages/dd-trace/test/setup/mocha.js' 'packages/dd-trace/test/debugger/**/*.spec.js'",
```
juan-fernandez, line 23: (absolute nit) unrelated to this PR: using `:debugger` is a bit confusing to me given that `debugger` is a keyword and that I usually use `test:debug` to pass `--inspect-brk` to test sessions. 

Maybe `test:dynamic-inst` ? That's awful 😆 , so feel free to disregard

juan-fernandez (approved):

rochdev:
We've been switching in the other direction, so not sure this would actually be productive. I'm pretty sure most if not all of these issues are because of how we use it too, and not because of `tap` itself.

@watson Did you try to use the `test` module instead? Last time I tried it it worked pretty well, the only reason I didn't make the switch is that there were tons of issues with our own tests, but since for you it's a fresh test suite that is only a few weeks old, maybe it would be easier to port. Ideally, eventually we'd move everything to `test` (and subsequently to the built-in `node:test` once the features we need are available in all Node versions we support).

watson:
> Did you try to use the `test` module instead?

I did not. I thought we were married to the mocha style. But I'm happy to hear we're not and you're right many of the issues I have discovered are because we're using the mocha syntax. I'll try to switch and see if it fixes all the mentioned issues.

package.json
watson: I would have preferred to use something related to our product name as well, but it's called "debugger" in all the other languages for historical reasons, so I was told to use this name in Node.js as well. I'll ask around to see if we still think this is a good idea. But if we decide to rename it, I'll make a separate PR as it touches a lot of other areas of the code as well - not just these lines.

watson:
I'm at a bit of an impasse. We currently depend on an old version of tap, and re-writing these tests using that old version gives me a bunch of problems. It would be easier for me to wait using tap until we have upgraded to this new version which has support for using sinon. So since the DI tests doesn't currently work well with tap (and it's only going to get worse in upcoming PR's were I add more test), I'd like to stop using it until we upgrade the version of tap we use in the project.

rochdev (approved):

juan-fernandez (approved):

PR #4729, opened 2024-09-26
[DI] Improve performance of algorithm to find probe scriptId

watson:
**Warning: High risk of bikeshedding!**

This might very well be overkill, but while I was refactoring this code recently, I was thinking of ways to use a `Map` instead of an array to store the reference between the script URL and the script ID.

This is the algorithm that I came up with.

This is not a very hot code-path as it's only hit every time a new probe is added. So it's not really necessary. So in the end this comes down to code readability and maintainability: And I can't really determine if it makes the code more readable/maintainable or worse.

What do you think?

### DIFF
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -96,7 +96,7 @@ jobs:
         uses: ./.github/actions/testagent/logs
       - uses: codecov/codecov-action@v3
 
-  amqp10: # TODO: move rhea to its own job
+  amqp10:
     runs-on: ubuntu-latest
     services:
       qpid:
@@ -107,7 +107,7 @@ jobs:
         ports:
           - 5673:5672
     env:
-      PLUGINS: amqp10|rhea
+      PLUGINS: amqp10
       SERVICES: qpid
       DD_DATA_STREAMS_ENABLED: true
     steps:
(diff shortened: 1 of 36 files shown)


watson:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4729?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4729** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4729?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4727** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4727?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#4728](https://github.com/DataDog/dd-trace-js/pull/4728) <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4728?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`

<h2></h2>

rochdev (approved):
> So it's not really necessary. So in the end this comes down to code readability and maintainability: And I can't really determine if it makes the code more readable/maintainable or worse.

I think readability is down a bit and performance is up a bit, but not to an extent where I'd argue one way or the other, so if you prefer that version then LGTM.

watson:
There isn't much difference in the performance. Here's a benchmark of 10.000 entries (the same in both the array and the map) where the key being searched for is in the middle:

```
Array x 23,873 ops/sec ±3.26% (97 runs sampled)
Map x 25,972 ops/sec ±0.92% (98 runs sampled)
Fastest is Map
```

The main difference I think is that there's fewer and simpler objects being created in memory for the `Map` implementation.

PR #4730, opened 2024-09-26
Use async local storage in profiler

szegedi:
### What does this PR do?
Updates the wall/cpu profiler so that instead of capturing the current span, local root span, and endpoint on each async context enter, it now computes them once for a context the first time it is encountered, and stores them in an async local storage. This reduces the computational burden on context enters. It is now reduced to setting a single property on the sample context object, which now serves as an additional level of indirection for the profiler to reach the current async store value. 

### Motivation
We're getting closer to the ideal of profiler sampling mechanism being able to use an async local store directly. It's a cleaner separation, greatly reduced work on context switches, and opens the way for adding custom context features to Node.js profiler.

### DIFF
--- a/packages/dd-trace/src/profiling/profilers/wall.js
+++ b/packages/dd-trace/src/profiling/profilers/wall.js
@@ -14,6 +14,7 @@ const {
   getNonJSThreadsLabels,
   getThreadLabels
 } = require('./shared')
+const { AsyncLocalStorage } = require('async_hooks')
 
 const beforeCh = dc.channel('dd-trace:storage:before')
 const enterCh = dc.channel('dd-trace:storage:enter')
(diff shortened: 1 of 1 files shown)


BridgeAR:
Ping @szegedi

BridgeAR:
@szegedi PTAL, I believe this got superseded by https://github.com/DataDog/dd-trace-js/pull/5743

szegedi:
> @szegedi Is this something you plan to get to the finish line? Did the new approach end up working or is it not feasible?

It's somewhat orthogonal to the ACF work (which ended up working, FWIW :-) ) – the end goal here would've been to create an API for the customer to add business-logic specific tags to samples, similar to how profilers for Java and Go already do: https://docs.datadoghq.com/profiler/guide/isolate-outliers-in-monolithic-services/?tab=java#isolate-your-own-business-logic. But I probably need to rethink the logic anyhow.

PR #4734, opened 2024-09-27
Fix: Calculate correct Content-Length in requests to agent

watson:
If the payload being sent to the agent contained any two-or-more-byte characters, the calculation would be incorrect because `str.length` returns the number of characters and not the number of bytes in the string.
    
If the payload was JSON, this would make it malformed and the receiver would not be able to parse it (as it would be missing one or more bytes from the end of the payload).

### DIFF
--- a/packages/dd-trace/src/exporters/common/request.js
+++ b/packages/dd-trace/src/exporters/common/request.js
@@ -183,7 +183,7 @@ function request (data, options, callback) {
 }
 
 function byteLength (data) {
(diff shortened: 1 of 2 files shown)


tlhunter (approved):
Is this something we should benchmark?

watson:
Not sure what you're referring to? Do you mean benchmarking the difference between `str.length` and `Buffer.byteLength(str)`? That wouldn't make any sense IMO as there's no alternative. Also, I'm sure this isn't going to change anything in regards to performance.

PR #4737, opened 2024-09-30
Do not run fastify >=5 plugin tests in node <=18

juan-fernandez:
### What does this PR do?
Skip tests if the fastify version is >=5 and node is <=18, since it's not supported anymore:

https://github.com/fastify/fastify/releases/tag/v5.0.0: it's not clearly put as a breaking change but it seems like that's indeed what's happening: https://github.com/fastify/fastify/pull/5481

### Motivation
It's currently blocking the release.

### DIFF
--- a/packages/datadog-plugin-cucumber/test/index.spec.js
+++ b/packages/datadog-plugin-cucumber/test/index.spec.js
@@ -56,7 +56,7 @@ describe('Plugin', function () {
   let Cucumber
   this.timeout(10000)
   withVersions('cucumber', '@cucumber/cucumber', (version, _, specificVersion) => {
-    if ((NODE_MAJOR <= 16) && semver.satisfies(specificVersion, '>=10')) return
+    if (NODE_MAJOR <= 16 && semver.satisfies(specificVersion, '>=10')) return
 
     afterEach(() => {
       // > If you want to run tests multiple times, you may need to clear Node's require cache
--- a/packages/datadog-plugin-fastify/test/index.spec.js
+++ b/packages/datadog-plugin-fastify/test/index.spec.js
@@ -26,7 +26,7 @@ describe('Plugin', () => {
 
       withExports('fastify', version, ['default', 'fastify'], '>=3', getExport => {
         describe('without configuration', () => {
-          if ((NODE_MAJOR <= 18) && semver.satisfies(specificVersion, '>=5')) return
+          if (NODE_MAJOR <= 18 && semver.satisfies(specificVersion, '>=5')) return
 
           before(() => {
             return agent.load(['fastify', 'find-my-way', 'http'], [{}, {}, { client: false }])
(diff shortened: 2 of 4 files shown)


packages/datadog-plugin-fastify/test/index.spec.js
```diff
@@ -25,6 +26,8 @@ describe('Plugin', () => {
 
       withExports('fastify', version, ['default', 'fastify'], '>=3', getExport => {
         describe('without configuration', () => {
+          if ((NODE_MAJOR <= 18) && semver.satisfies(specificVersion, '>=5')) return
```
uurien, line 29: Why these extra params?
```suggestion
          if (NODE_MAJOR <= 18 && semver.satisfies(specificVersion, '>=5')) return
```

packages/datadog-plugin-fastify/test/integration-test/client.spec.js
```diff
@@ -8,14 +8,17 @@ const {
...
 
 describe('esm', () => {
   let agent
   let proc
   let sandbox
 
   // skip older versions of fastify due to syntax differences
-  withVersions('fastify', 'fastify', '>=3', version => {
+  withVersions('fastify', 'fastify', '>=3', (version, _, specificVersion) => {
+    if ((NODE_MAJOR <= 18) && semver.satisfies(specificVersion, '>=5')) return
```
uurien, line 20: Same here
```suggestion
    if ((NODE_MAJOR <= 18) && semver.satisfies(specificVersion, '>=5')) return
```

packages/datadog-plugin-fastify/test/index.spec.js
juan-fernandez: I just copy pasted from somewhere else without checking 😆 , good catch!

packages/datadog-plugin-cucumber/test/index.spec.js
```diff
@@ -56,7 +56,7 @@ describe('Plugin', function () {
   let Cucumber
   this.timeout(10000)
   withVersions('cucumber', '@cucumber/cucumber', (version, _, specificVersion) => {
-    if ((NODE_MAJOR <= 16) && semver.satisfies(specificVersion, '>=10')) return
+    if (NODE_MAJOR <= 16 && semver.satisfies(specificVersion, '>=10')) return
```
juan-fernandez, line 59: ℹ️ unrelated: removing unnecessary parenthesis

packages/datadog-plugin-undici/test/index.spec.js
```diff
@@ -23,7 +23,7 @@ describe('Plugin', () => {
   describe('undici-fetch', () => {
     withVersions('undici', 'undici', version => {
       const specificVersion = require(`../../../versions/undici@${version}`).version()
-      if ((NODE_MAJOR <= 16) && semver.satisfies(specificVersion, '>=6')) return
+      if (NODE_MAJOR <= 16 && semver.satisfies(specificVersion, '>=6')) return
```
juan-fernandez, line 26: ℹ️ unrelated: removing unnecessary parenthesis

uurien (approved):

PR #4738, opened 2024-09-30
[MLOB-1560] LLMObs Span Processor

sabrenner:
### What does this PR do?
Adds a span processor for the LLM Observability product. An instance of it is now a property on the tracer span processor, and it reads the temporary tags on the span and extracts them into an LLM Obs payload to append to the writer. Then, these tags are not read when formatting the span.

Additionally, makes an internal change to how these are stored on tags. Since `JSON.stringify` and `JSON.parse` can be expensive, it is better to only serialize at the writer flushing stage. However, by storing tags as plain objects and not stringifying at the time they are added, we do not get the benefit of checking if they are will not throw when stringified. This is fixed by adding some specific guards in-place at the tagger (for `metrics`, `documents`, and `messages`) and span processor (for `metadata`, which has ambiguous value types).

Future work is planned to try and replace tagging the span with using some kind of external, namespaced storage.

### Motivation
Follow-up PR in a series of PRs introducing an LLM Observability SDK into the Node.js tracer.

### DIFF
--- a/packages/dd-trace/src/exporters/agent/index.js
+++ b/packages/dd-trace/src/exporters/agent/index.js
@@ -42,12 +42,6 @@ class AgentExporter {
   }
 
   export (spans) {
-    const { llmobs } = this._config
-    if (llmobs.enabled && llmobs.agentlessEnabled) {
-      log.debug('LLMObs agentless mode enabled. Not sending APM spans to the agent.')
-      return
-    }
-
     this._writer.append(spans)
 
     const { flushInterval } = this._config
--- a/packages/dd-trace/src/exporters/agent/index.js
+++ b/packages/dd-trace/src/exporters/agent/index.js
@@ -42,6 +42,12 @@ class AgentExporter {
   }
 
   export (spans) {
+    const { llmobs } = this._config
+    if (llmobs.enabled && llmobs.agentlessEnabled) {
+      log.debug('LLMObs agentless mode enabled. Not sending APM spans to the agent.')
+      return
+    }
+
     this._writer.append(spans)
 
     const { flushInterval } = this._config
--- a/packages/dd-trace/src/format.js
+++ b/packages/dd-trace/src/format.js
@@ -30,8 +30,6 @@ const map = {
   'resource.name': 'resource'
 }
 
-const llmObsTagPrefix = '_ml_obs'
-
 function format (span) {
   const formatted = formatSpan(span)
 
@@ -162,9 +160,7 @@ function extractTags (trace, span) {
           break
         }
       default: // eslint-disable-line no-fallthrough
-        if (!tag.startsWith(llmObsTagPrefix)) { // don't add ml_obs-related tags
-          addTag(trace.meta, trace.metrics, tag, tags[tag])
-        }
+        addTag(trace.meta, trace.metrics, tag, tags[tag])
     }
   }
   setSingleSpanIngestionTags(trace, context._spanSampling)
--- a/packages/dd-trace/src/format.js
+++ b/packages/dd-trace/src/format.js
@@ -160,7 +160,9 @@ function extractTags (trace, span) {
           break
         }
       default: // eslint-disable-line no-fallthrough
-        addTag(trace.meta, trace.metrics, tag, tags[tag])
+        if (!tag.startsWith('_ml_obs')) { // don't add ml_obs-related tags
+          addTag(trace.meta, trace.metrics, tag, tags[tag])
+        }
     }
   }
   setSingleSpanIngestionTags(trace, context._spanSampling)
--- a/packages/dd-trace/src/format.js
+++ b/packages/dd-trace/src/format.js
@@ -30,6 +30,8 @@ const map = {
   'resource.name': 'resource'
 }
 
+const llmObsTagPrefix = '_ml_obs'
+
 function format (span) {
   const formatted = formatSpan(span)
 
@@ -160,7 +162,7 @@ function extractTags (trace, span) {
           break
         }
       default: // eslint-disable-line no-fallthrough
-        if (!tag.startsWith('_ml_obs')) { // don't add ml_obs-related tags
+        if (!tag.startsWith(llmObsTagPrefix)) { // don't add ml_obs-related tags
           addTag(trace.meta, trace.metrics, tag, tags[tag])
         }
     }
--- a/packages/dd-trace/src/llmobs/span_processor.js
+++ b/packages/dd-trace/src/llmobs/span_processor.js
@@ -46,7 +46,7 @@ class LLMObsSpanProcessor {
 
   // TODO: can we correlate the span + trace IDs with a namespaced object to
   // access LLMObs properties associated with the span?
-  process (span) {
+  process ({ span }) {
     if (!this._config.llmobs.enabled) return
     // if the span is not in our private tagger map, it is not an llmobs span
     if (!LLMObsTagger.tagMap.has(span)) return
--- a/packages/dd-trace/src/llmobs/span_processor.js
+++ b/packages/dd-trace/src/llmobs/span_processor.js
@@ -30,6 +30,7 @@ const AgentProxyWriter = require('./writers/spans/agentProxy')
 const { isLLMSpan } = require('./util')
 
 const tracerVersion = require('../../../../package.json').version
+const logger = require('../log')
 
 class LLMObsSpanProcessor {
   constructor (config) {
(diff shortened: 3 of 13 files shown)


packages/dd-trace/src/exporters/agent/index.js
```diff
@@ -42,6 +42,12 @@ class AgentExporter {
   }
 
   export (spans) {
+    const { llmobs } = this._config
```
sabrenner, line 45: Explaining this: we want `llmobs.agentlessEnabled` to have two uses (in parity with our existing Python SDK).
1. Use our agentless writer if specified (this is done in `llmobs/span_processor.js`)
2. Not export spans to the agent.

We expect (as has been so far with our Python SDK) a lot of new DD customers that aren't using APM. `agentlessEnabled` should allow customers to skip exporting spans to the agent. My question is, is it better to do it like this, or to create some kind of `LLMObsNoopExporter`, and have that be returned from `packages/dd-trace/src/exporter.js`?

packages/dd-trace/src/llmobs/span_processor.js
```diff
@@ -0,0 +1,147 @@
...
+    const meta = { 'span.kind': spanKind, input: {}, output: {} }
+    const input = {}
+    const output = {}
+
+    if (['llm', 'embedding'].includes(spanKind) && tags[MODEL_NAME]) {
+      meta.model_name = tags[MODEL_NAME]
+      meta.model_provider = (tags[MODEL_PROVIDER] || 'custom').toLowerCase()
+    }
+    if (tags[METADATA]) {
+      meta.metadata = JSON.parse(tags[METADATA])
```
sabrenner, line 66: [APM input needed] Coming back to the whole `JSON.stringify` and then now `JSON.parse`... I think on the tagger I stringified because tags _should_ be set as strings. However, if I won't be adding these LLMObs tags to the APM payload anyways, can I just store these JSON properties as objects on the span tags? Then, I won't have to do the whole stringify/parse game, and probably boost performance a little.

packages/dd-trace/src/exporters/agent/index.js
rochdev: If the config was on the whole library, I'd go with some sort of `NoopExporter` or an `AgentlessExporter` that would be a noop while this is not supported for tracing. But the option being on `llmobs` specifically seems to me that we should just not do anything for tracing. If an agent is present, then traces will be reported as normal. If they aren't present, the HTTP connection will fail to open and nothing will be flushed (which is already roughly equivalent to noop). I don't think an option that is explicitly for one product should touch anything in other products as that is an unexpected side-effect.

packages/dd-trace/src/llmobs/span_processor.js
sabrenner: Roch and I talked offline - we'll try and do this in follow-up work next week. For now, we'll store these on tags.

packages/dd-trace/src/llmobs/span_processor.js
```diff
@@ -0,0 +1,189 @@
...
+      if (seenObjects.has(value)) return true
+      seenObjects.add(value)
+      return false
+    }
+
+    const add = (obj, carrier) => {
+      for (const key in obj) {
+        const value = obj[key]
+        if (!Object.prototype.hasOwnProperty.call(obj, key)) continue
+        if (typeof value === 'bigint' || isCircular(value)) continue
```
sabrenner, line 157: adding in placeholder + log

packages/dd-trace/src/llmobs/span_processor.js
```diff
@@ -0,0 +1,195 @@
...
+
+    if (llmobs.enabled) {
+      const LLMObsSpanWriter = llmobs.agentlessEnabled ? AgentlessWriter : AgentProxyWriter
+      this._writer = new LLMObsSpanWriter(config)
+    }
+  }
+
+  process (span) {
+    if (!this._config.llmobs.enabled) return
+    if (!isLLMSpan(span)) return
```
lievan, line 49: nit: maybe rename this to isLLMObsSpan to make it clear we're checking for LLM Obs spans, and not the llm span kind

packages/dd-trace/src/llmobs/span_processor.js
```diff
@@ -0,0 +1,195 @@
...
+      ml_app: mlApp,
+      'dd-trace.version': tracerVersion,
+      error: Number(!!error) || 0,
+      language: 'javascript'
+    }
+    const errType = span.context()._tags[ERROR_TYPE] || error?.name
+    if (errType) tags.error_type = errType
+    if (sessionId) tags.session_id = sessionId
+    const existingTags = span.context()._tags[TAGS] || {}
+    if (existingTags) tags = { ...tags, ...existingTags }
```
lievan, line 190: will `existing_tags` override the tag values in `tags` or will they just be duplicated?  I _think_ the correct behavior should be override

packages/dd-trace/src/llmobs/tagger.js
```diff
@@ -78,32 +78,29 @@ class LLMObsTagger {
...
-    try {
-      span.setTag(METRICS, JSON.stringify(metrics))
-    } catch {
-      logger.warn('Failed to parse span metrics. Metrics key-value pairs must be JSON serializable.')
+    const filterdMetrics = {}
+    for (const [key, value] of Object.entries(metrics)) {
+      if (typeof value === 'number') {
+        filterdMetrics[key] = value
+      } else {
+        logger.warn(`Metric ${key} must be a number, instead got ${value}`)
```
lievan, line 90: nit; maybe put quotation marks around `${key}` / rephrase it like
"Value for metric '{metric_name}' must be a number, instead got .."

packages/dd-trace/src/format.js
```diff
@@ -160,7 +160,9 @@ function extractTags (trace, span) {
           break
         }
       default: // eslint-disable-line no-fallthrough
-        addTag(trace.meta, trace.metrics, tag, tags[tag])
+        if (!tag.startsWith('_ml_obs')) { // don't add ml_obs-related tags
```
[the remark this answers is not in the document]
lievan: Actually seems like it's mostly in the testing code, feel free to disregard this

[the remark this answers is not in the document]
sabrenner: I'll still make this a constant in `format`, might be easier to read that way

packages/dd-trace/src/llmobs/span_processor.js
sabrenner: yeah, existing tags will override new tags passed in. in JS, when you spread objects into other objects, they're applied from left to right, so duplicate properties in `existingTags` override properties in `tags`

lievan: nice thanks

lievan (approved):

packages/dd-trace/src/llmobs/tagger.js
```diff
@@ -188,97 +182,92 @@ class LLMObsTagger {
...
+              toolCallObj.name = name
+            }
 
-              if (args) {
-                if (typeof args !== 'object') {
-                  logger.warn('Tool arguments must be an object.')
-                  return undefined
-                }
-                toolCallObj.arguments = args
+            if (args) {
```
rochdev, line 234: Instead of repeating this condition several times it could be extracted to a function.

packages/dd-trace/src/span_processor.js
```diff
@@ -42,6 +44,7 @@ class SpanProcessor {
 
       for (const span of started) {
         if (span._duration !== undefined) {
+          this._llmobs.process(span)
```
rochdev, line 47: Could this be decoupled with a channel?

packages/dd-trace/src/llmobs/tagger.js
```diff
@@ -57,6 +58,7 @@ class LLMObsTagger {
     span.setTag(PARENT_ID_KEY, parentId)
   }
 
+  // TODO: similarly for the following `tag` methods, can we pass in a namespaced object instead of the span?
```
rochdev, line 61: It should be possible event short-term to have a WeakMap on the LLMObs side that would store tags outside of the span, for example something like `tagger.tagMap`. It could then be improved later on with the Core API but at least the tracing formatter wouldn't need to be modified to know about LLMObs.

rochdev (changes requested):

packages/dd-trace/src/llmobs/tagger.js
sabrenner: addressed in 80168f80aa4218c51a3bb7789ae49888d69f8627

packages/dd-trace/src/span_processor.js
sabrenner: yes, i think it can! for clarification, this will mean i'll initialize the LLMObs span processor in the SDK (next PR) and add `processor.format` as a subscriber to that channel when the SDK is initialized. It should actually give me better/direct access to the LLMObs span processor's writer

packages/dd-trace/src/llmobs/tagger.js
sabrenner: tried this in 4a19fec0ffc76a39a9104104beb8ff003f57db0b, let me know what you think 😄

rochdev (approved):

PR #4742, opened 2024-09-30
[MLOB-1524] feat(llmobs): Introduce LLM Observability SDK

sabrenner:
### What does this PR do?
Introduces an LLM Observability SDK into the tracer, with its own tracing API extending the core tracer, and span tagging and processing functionality to flush LLM Observability spans to LLM Observability.

The following PRs have been merged into the branch for this PR, and have been reviewed and approved:

(config) #4696 APM: @ida613 LLMObs: @Yun-Kim 
(writer) #4699 APM: @rochdev LLMObs: @Yun-Kim 
(tagger) #4718 APM: @rochdev LLMObs @Yun-Kim 
(span processing) #4738  APM: @rochdev LLMObs: @lievan 
(sdk api) #4773 APM: @rochdev LLMObs: @Yun-Kim @lievan @Kyle-Verhoog 

Submitting OpenAI spans will be a direct follow-up PR.

### Motivation
Introduce core LLM Observability SDK logic into the tracer.


packages/dd-trace/src/llmobs/channels.js
```diff
@@ -0,0 +1,10 @@
+'use strict'
+
+const { channel } = require('dc-polyfill')
+
+module.exports = {
```
rochdev, line 5: The main idea of channels is to not need to depend on shared objects directly like this and use names instead. What is the benefit of centrally defining channel objects?

packages/dd-trace/src/llmobs/constants.js
```diff
@@ -0,0 +1,46 @@
...
+module.exports = {
+  SPAN_KINDS: ['llm', 'agent', 'workflow', 'task', 'tool', 'embedding', 'retrieval'],
+  SPAN_KIND: '_ml_obs.meta.span.kind',
+  SESSION_ID: '_ml_obs.session_id',
+  METADATA: '_ml_obs.meta.metadata',
+  METRICS: '_ml_obs.metrics',
+  ML_APP: '_ml_obs.meta.ml_app',
+  PROPAGATED_PARENT_ID_KEY: '_dd.p.llmobs_parent_id',
+  PARENT_ID_KEY: '_ml_obs.llmobs_parent_id',
+  TAGS: '_ml_obs.tags',
```
rochdev, line 12: Nitpick (not a blocker) but I think splitting the constants in multiple files by subtypes (for example `tags.js`) would make them clearer and discovery easier.

packages/dd-trace/src/llmobs/channels.js
sabrenner: I think I was thinking it was more organized this way, but I see your point. But tbh I only use these channels in `index.js` where I subscribe/unsubscribe from everything, so I'll just move them there 👍

packages/dd-trace/src/llmobs/constants.js
sabrenner: Ah yes this would be good! I'll do this

packages/dd-trace/test/llmobs/sdk/index.spec.js
```diff
@@ -0,0 +1,1008 @@
...
+    LLMObsAgentProxySpanWriter.prototype.flush.resetHistory()
+
+    process.removeAllListeners('beforeExit')
+  })
+
+  after(() => {
+    sinon.restore()
+    llmobs.disable()
+    delete global._ddtrace
+    delete require.cache[require.resolve('../../../../dd-trace')]
```
rochdev, line 66: Why is this needed?

packages/dd-trace/test/llmobs/sdk/index.spec.js
```diff
@@ -0,0 +1,1008 @@
...
+        })
+
+        it('throws if the kind is invalid', () => {
+          expect(() => llmobs.trace({ kind: 'invalid' }, () => {})).to.throw()
+
+          expect(llmobs._tracer._processor.process).to.not.have.been.called
+          expect(LLMObsSpanProcessor.prototype.format).to.not.have.been.called
+        })
+
+        // need span kind optional for this
```
rochdev, line 192:
```suggestion
        // TODO: need span kind optional for this
```

packages/dd-trace/test/llmobs/sdk/index.spec.js
```diff
@@ -0,0 +1,1008 @@
...
+            .catch(done)
+
+          expect(span.finish).to.not.have.been.called
+
+          deferred.resolve()
+        })
+      })
+
+      describe('parentage', () => {
+        // it('starts a span with a distinct trace id', () => {
```
rochdev, line 255: Why is this commented out?

packages/dd-trace/test/llmobs/sdk/index.spec.js
```diff
@@ -0,0 +1,1008 @@
...
+        //       .to.exist.and.to.not.equal(span.context().toTraceId(true))
+        //   })
+        // })
+
+        it('sets span parentage correctly', () => {
+          llmobs.trace({ kind: 'workflow', name: 'test' }, outerLLMSpan => {
+            llmobs.trace({ kind: 'task', name: 'test' }, innerLLMSpan => {
+              expect(LLMObsTagger.tagMap.get(innerLLMSpan)['_ml_obs.llmobs_parent_id'])
+                .to.equal(outerLLMSpan.context().toSpanId())
+              // expect(innerLLMSpan.context()._tags['_ml_obs.trace_id'])
```
rochdev, line 267: Why is this commented out?

packages/dd-trace/test/llmobs/sdk/index.spec.js
```diff
@@ -0,0 +1,1008 @@
...
+
+                // apm span linkage
+                expect(innerLLMSpan.context()._parentId.toString(10)).to.equal(apmSpan.context().toSpanId())
+                expect(apmSpan.context()._parentId.toString(10)).to.equal(outerLLMSpan.context().toSpanId())
+              })
+            })
+          })
+        })
+
+        // it('starts different traces for llmobs spans as child spans of an apm root span', () => {
```
rochdev, line 293: Ditto.

packages/dd-trace/test/llmobs/sdk/index.spec.js
```diff
@@ -0,0 +1,1008 @@
...
+
+          expect(LLMObsTagger.tagMap.get(span)).to.deep.equal({
+            '_ml_obs.meta.span.kind': 'retrieval',
+            '_ml_obs.meta.ml_app': 'mlApp',
+            '_ml_obs.llmobs_parent_id': 'undefined',
+            '_ml_obs.meta.input.value': 'input'
+          })
+        })
+
+        // need span kind optional for this test
```
rochdev, line 439:
```suggestion
        // TODO: need span kind optional for this test
```

packages/dd-trace/test/llmobs/sdk/index.spec.js
```diff
@@ -0,0 +1,1008 @@
...
+        //   })
+
+        //   fn()
+
+        //   expect(span.context()._name).to.equal('unnamed-anonymous-function')
+        // })
+      })
+
+      describe('parentage', () => {
+        // it('starts a span with a distinct trace id', () => {
```
rochdev, line 453: A lot of tests seem to be commented out without an explanation. It would be better to `it.skip` them and add an explanation.

rochdev (changes requested):

Kyle-Verhoog (approved):
reviewed it to the best of my noob js and mlobs ability. Great job Sam 👏 🚢 

Let's try to get some tests into the shared testing framework ASAP to assert the similarity with the Python interface!

packages/dd-trace/src/config.js
```diff
@@ -501,6 +501,10 @@ class Config {
     this._setValue(defaults, 'isGitUploadEnabled', false)
     this._setValue(defaults, 'isIntelligentTestRunnerEnabled', false)
     this._setValue(defaults, 'isManualApiEnabled', false)
+    this._setValue(defaults, 'llmobs.agentlessEnabled', false)
+    this._setValue(defaults, 'llmobs.apiKey', undefined)
```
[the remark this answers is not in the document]
rochdev: I think it's because CI Visibility needs it defined as an environment variable because it changes its own exporter before even initializing the library. That is not necessarily an argument against a programmatic config though, as one doesn't prevent the other. For configuration options that apply library-wide, I'd definitely prefer that it's at the root, otherwise it gives the impression that you can mix and match API keys or even a hybrid agent+agentless.

packages/dd-trace/test/llmobs/sdk/index.spec.js
sabrenner: was able to remove it. I had it there because I was having issues between two sets of tests but was able to resolve it

packages/dd-trace/src/config.js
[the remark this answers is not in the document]
sabrenner: good point. happy to throw that change in this PR (have it at the root level), unless you'd rather have it in it's own PR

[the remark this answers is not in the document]
sabrenner: actually, I can push up a change that has LLMObs only rely on the current `config.apiKey` set from `DD_API_KEY`. I'll propose a follow-up PR to add it to the root level of the in-code config as well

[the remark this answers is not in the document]
sabrenner: addressed in 3e6945b57c39200bbbeb2d9331b9f6dfe20165bf (+ test fix after lol)

rochdev (approved):

PR #4743, opened 2024-10-01
Don't use deprecated url.parse function

watson:
The old code had several bugs/issues:

1. The inlined polyfill was lifted from Node.js core several years ago. Bugs have since been fixed in Node.js core that hasn't made it into our inlined version.
2. The check to see if `url.urlToHttpOptions` was a function had a bug (as `urlToHttpOptions` was never imported) so the check would always return `false`.
3. The fallback in case `url.urlToHttpOptions` wasn't present didn't load the polyfill, but instead the deprecated `url.parse`, so the polyfill was actually never used in that case.
4. Perf: The check to see if `url.urlToHttpOptions` was pressent was performed on every request instead of at load-time.
5. Perf: No need to use a RegEx to strip `unix:` from the URL when using named pipes on Windows.

## Note to reviewers

The minimum Node.js version supported by this tracer now ships with `url.urlToHttpOptions`, so from that point of view, the polyfill is no longer needed. However, we still maintain support for an old version of Cypress that ships with an older built-in version of Node.js that doesn't have `url.urlToHttpOptions`. So for that reason, we need to still have a polyfill.

### DIFF
--- a/packages/datadog-plugin-cypress/test/app/cypress/plugins/index.js
+++ b/packages/datadog-plugin-cypress/test/app/cypress/plugins/index.js
@@ -1,32 +1,3 @@
-// TODO: Remove `urlToHttpOptions` polyfill once we drop support for the older Cypress versions that uses a built-in
-// version of Node.js doesn't include that function.
-const url = require('url')
-if (url.urlToHttpOptions === undefined) {
-  url.urlToHttpOptions = (url) => {
-    const { hostname, pathname, port, username, password, search } = url
-    const options = {
-      __proto__: null,
-      ...url, // In case the url object was extended by the user.
-      protocol: url.protocol,
-      hostname: typeof hostname === 'string' && hostname.startsWith('[')
-        ? hostname.slice(1, -1)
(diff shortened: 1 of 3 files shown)


watson:
* **#4743** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4743?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

juan-fernandez (commented):
thanks for this!

The problem I see with this is that we only fix the unit test, but a customer using this version of cypress would still crash. 

Could we do a polyfill, like we do in `packages/dd-trace/src/iitm.js`? Where we check the version of node we're running and do something like:

```javascript 
// url-polyfill.js
if (semver.satisfies(process.versions.node, '$NODE_VERSION')) {
  module.exports = require('url')
} else {
  const url = require('url')
  url.urlToHttpOptions = ... // your polyfill
  return url
}
```

juan-fernandez (approved):
thanks!

PR #4745, opened 2024-10-01
[test visibility] Read `pull_request` and `pull_request_target` event info from GHA

juan-fernandez:
### What does this PR do?
Detect if we're in a `pull_request` or `pull_request_target` event by checking `GITHUB_BASE_REF`, and if we are, attempt to read the `event.json` payload made available by GitHub in `GITHUB_EVENT_PATH`. 

### Motivation
GitHub does not run workflows in the head of the branch about to be merged whenever the workflow is triggered through `pull_request` or `pull_request_target` events. It runs them in a merge commit. This is the one that we add to `git.commit.sha`. This is sometimes confusing because this commit does not exist in the history of the branch. What we did:

* Notice when this happens
* When it does, add a new set of tags:
  * `git.commit.head_sha`, which is the head SHA of the feature branch (branch to be merged).
  * `git.pull_request.base_branch_sha`, which is the head SHA of the base branch (branch you're merging into)
  * `git.pull_request.base_branch`, which is the name of the base branch 

### Plugin Checklist

- [x] Unit tests.


packages/dd-trace/src/plugins/util/ci.js
```diff
@@ -277,6 +289,16 @@ module.exports = {
           GITHUB_RUN_ATTEMPT
         })
       }
+      if (GITHUB_BASE_REF) { // `pull_request` or `pull_request_target` event
+        tags[GIT_PULL_REQUEST_BASE_BRANCH] = GITHUB_BASE_REF
+        try {
+          const eventContent = getGitHubEventPayload()
+          tags[GIT_PULL_REQUEST_BASE_BRANCH_SHA] = eventContent.pull_request.base.sha
```
uurien, line 296: the `getGitHubEventPayload` method is doing `return` when `GITHUB_EVENT_PATH` is not set. 
You are not checking here that `eventContent` has value, so this could crash if `GITHUB_EVENT_PATH` is not set.

uurien: and, could happen that the ci is running on master or a branch without PR? In that case, we could have the file in `env.GITHUB_EVENT_PATH`, but maybe without `pull_request` property.

uurien: ups, you can forget these messages, I haven't see the try catch ere 🤦

uurien (approved):

PR #4540, opened 2024-07-19
fix child_process propagation issues

bengl:
### What does this PR do?
- disables the `enter` argument to `startSpan()` which then disables the `storage.enterWith({ ...store, span })` call

### Motivation
- otherwise it appears that sometimes unrelated spans become child spans of a child process operation

Ref #4274

PR #4752, opened 2024-10-02
Fix child process not maintaining previous parent span after execution

uurien:
### What does this PR do?

Fix child process not maintaining previous parent span after execution

### Motivation

In some cases it was replacing the parent span by the new one, and set the next spans as children instead of a siblings.

### Plugin Checklist

- [x] Unit tests.


hoolioh (dismissed):

tlhunter:
Added additional tests from https://github.com/DataDog/dd-trace-js/pull/4540

tlhunter (approved):

PR #4753, opened 2024-10-02
add requirements json with native deps and denylist

rochdev:
### What does this PR do?

Add requirements JSON with native deps and denylist.

### Motivation

This file is used by SSI to determine the platforms where the library should be injected, and also as a denylist for specific commands that should be ignored.

The additional JSON files and the GitLab job are used by the shared pipeline to test the requirements JSON.

### Additional Notes

We should support 32-bit for Alpine but it was never added to [action-prebuildify](https://github.com/DataDog/action-prebuildify), so I added it to the denylist for now. Most Alpine users use Docker and the default Docker images use 64-bit, which might explain why this has never been a problem.


requirements.json
```diff
@@ -0,0 +1,85 @@
...
+      "arch": "arm64",
+      "supported": true,
+      "description": "From alpine 3.13"
+    },{
+      "arch": "x64",
+      "supported": true,
+      "description": "From alpine 3.13"
+    },{
+      "arch": "x86",
+      "supported": true,
```
andrewlock, line 40: In the description you say "We should support 32-bit for Alpine" - should this (and `arm`? be `false`)?

requirements.json
```diff
@@ -0,0 +1,85 @@
...
+      "id": "npm",
+      "description": "Ignore the npm CLI",
+      "os": null,
+      "cmds": [
+        "**/node",
+        "**/nodejs",
+        "**/ts-node",
+        "**/ts-node-*"
+      ],
+      "args": [{ "args":  ["*/npm-cli.js"], "position": 1}],
```
andrewlock, line 55: nit: on Windows this path separator will be incorrect, but as SSI is not windows-based yet, this is not an issue. I think the only solution with the current design would be to have a duplicate, windows-only rule 🤔 which almost certainly isn't worth it.

Also, is it possible to pass flags to `node` _prior_ to the `npm-cli.js` arg? e.g.

```
/path/to/node --some-flag /path/npm-cli.js
```

If so, then you might not want the `position` argument 🤔 But tbh, the denylist doesn't handle this particularly well right now

requirements.json
rochdev: I reverted that after posting the original PR, because we don't publish binaries for it, meaning that while some features won't work, the library itself does work so there wasn't really a reason to stop injection.

requirements.json
rochdev: > nit: on Windows this path separator will be incorrect, but as SSI is not windows-based yet, this is not an issue. I think the only solution with the current design would be to have a duplicate, windows-only rule 🤔 which almost certainly isn't worth it.

I thought about adding Windows but figured as you pointed out that it's not supported anyway right now. For Windows the rules would actually be quite different as even Node itself has a different binary name. I can add support if needed.

> Also, is it possible to pass flags to node prior to the npm-cli.js arg? e.g.

It would be possible when used directly in that way, but when using the `npm` CLI (which should be basically all usage) it should always be first.

> If so, then you might not want the position argument 🤔 But tbh, the denylist doesn't handle this particularly well right now

I only added it because I wasn't sure of the implications of leaving it out, but I can definitely remove it if there aren't any negative side-effects for doing so.

andrewlock: > I thought about adding Windows but figured as you pointed out that it's not supported anyway right now. For Windows the rules would actually be quite different as even Node itself has a different binary name. I can add support if needed.

Yeah, sounds good, I don't think there's any need to add this right now

> I only added it because I wasn't sure of the implications of leaving it out, but I can definitely remove it if there aren't any negative side-effects for doing so.

The negative side effects would be that we would block someone running something like:

```
/path/to/node somethingelse.js --some-arg blah/npm-cli.js
```

> It would be possible when used directly in that way, but when using the npm CLI (which should be basically all usage) it should always be first.

Given you state this, I think that leaving it as-is is correct 👍

requirements.json
andrewlock: Great, as long as you handle that in-library then there are no issues I think 👍

bengl (approved):

PR #4755, opened 2024-10-03
add support to api security sampling

IlyasShabi:
### What does this PR do?

- Implements a new API security sampling algorithm using an LRU cache with a 30s TTL
- Delays schema extraction until the end of the request processing

### Motivation
The current API security sampling method randomly selects 10% of requests. This PR introduces new algorithm based on request priority, utilizing an LRU cache with TTL for each { url, statusCode, method }

Delaying the schema extraction to the end of the request ensures that we have an accurate priority value

### Plugin Checklist

- [x] Unit tests.
- [x] TypeScript [tests][2].
- [ ] API [documentation][3].


index.d.ts
```diff
@@ -654,19 +654,14 @@ declare namespace tracer {
         mode?: 'safe' | 'extended' | 'disabled'
       },
       /**
-       * Configuration for Api Security sampling
+       * Configuration for Api Security
        */
       apiSecurity?: {
```
IlyasShabi, line 659: I preferred to keep `apiSecurity` for future usage and not only for sampling

packages/dd-trace/src/appsec/api_security_sampler_cache.js
```diff
@@ -0,0 +1,53 @@
+'use strict'
```
CarlesDD, line 1: Why not use `lru-cache`? It's a dep used in other places in the lib.

packages/dd-trace/src/appsec/index.js
```diff
@@ -202,7 +202,6 @@ function onRequestCookieParser ({ req, res, abortController, cookies }) {
 
 function onResponseBody ({ req, body }) {
   if (!body || typeof body !== 'object') return
-  if (!apiSecuritySampler.isSampled(req)) return
```
uurien, line 205: if we remove this here, we are going to call to the waf always with the response body, it is an overhead that we try to minimize, I think that the thing that we want to do here is:

```
if (!apiSecuritySampler.sampleRequest(req, res)) return;
```

packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -1,61 +1,47 @@
...
-    parsed = 0
-  } else {
-    parsed = Math.min(1, Math.max(0, parsed))
-  }
+  const rootSpan = web.root(req)
+  if (!rootSpan) return false
 
-  return parsed
-}
+  const isSampled = prioritySampler.isSampled(rootSpan)
```
uurien, line 27: This method set the priority sampler in the trace. Should we set the trace priority if, at the end, it was already sampled? Should we first check in our map, or in the trace?

imho, we should check first our map, and once that we have checked that we want to send it to the waf, check if the trace will be maintained or not.
if we want to maintain and prioirtySampler says that it is OK, update our map.

wdyt?
cc @iunanua

docs/test.ts
```diff
@@ -115,7 +115,6 @@ tracer.init({
     },
     apiSecurity: {
       enabled: true,
-      requestSampling: 1.0
```
simon-id, line 118: breaking change :/

packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -1,61 +1,97 @@
...
+
+function computeKey (req, res) {
+  const route = req.route?.path || req.url
+  const method = req.method.toLowerCase()
+  const statusCode = res.statusCode
+  const str = route + statusCode + method
+  return crypto.createHash('md5').update(str).digest('hex')
+}
+
+function parseSampleDelay (delay) {
```
iunanua, line 78: Usually config property values are resolved in the `config.js` file.
There you can set the default value, and the value read from env vars, etc

IlyasShabi: Got it, I think there is no need for this function then

packages/dd-trace/src/appsec/api_security_sampler.js
iunanua: it was discussed. I guess we can resolve the conversation.

integration-tests/standalone-asm.spec.js
```diff
@@ -103,7 +103,7 @@ describe('Standalone ASM', () => {
         assert.notProperty(meta, 'manual.keep')
         assert.notProperty(meta, '_dd.p.appsec')
 
-        assert.propertyVal(metrics, '_sampling_priority_v1', 1)
+        assert.propertyVal(metrics, '_sampling_priority_v1', 0)
```
iunanua, line 106: idk why this test affected by the PR changes. Let me debug it locally

packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -1,61 +1,87 @@
...
   enabled = false
+  sampledRequests?.clear()
 }
 
-function setRequestSampling (sampling) {
-  requestSampling = parseRequestSampling(sampling)
-}
+function sampleRequest (req, res, force = false) {
+  if (!enabled) return false
+  if (this.isSampled(req, res)) return false
```
uurien, line 27: why `this.`? Isn't this code will crash if:

```
const { sampleRequest } = require('..../api_security_sampler.js')
sampleRequest(req, res)
```


```suggestion
  if (isSampled(req, res)) return false
```

packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -1,61 +1,87 @@
...
+
+function isSampled (req, res) {
+  const key = computeKey(req, res)
+  if (!key) return false
+
+  return sampledRequests.has(key)
+}
+
+function computeKey (req, res) {
+  const route = web.getContext(req).paths.join('') || ''
```
simon-id, line 66: the interesting thing about getting the paths from web.js is that the result will depend on when you call it.
If this sampling is called in a router, it will be the route up to that router, but if it's called at the end of the request it will be the last router I think.
Is that a good or bad thing ?

simon-id (commented):
haven't reviewed it all yet but that should keep you busy:

integration-tests/standalone-asm.spec.js
iunanua, line 106: this PR should not affect standalone behaviour

```suggestion
        assert.propertyVal(metrics, '_sampling_priority_v1', 1)
```

packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -1,61 +1,87 @@
...
+  if (this.isSampled(req, res)) return false
 
-function parseRequestSampling (requestSampling) {
-  let parsed = parseFloat(requestSampling)
+  const rootSpan = web.root(req)
+  if (!rootSpan) return false
 
-  if (isNaN(parsed)) {
-    log.warn(`Incorrect API Security request sampling value: ${requestSampling}`)
+  const priority = getSpanPriority(rootSpan)
```
iunanua, line 32:
```suggestion
  let priority = getSpanPriority(rootSpan)
  if (!priority) {
    rootSpan._prioritySampler?.sample(rootSpan)
    priority = getSpanPriority(rootSpan)
  }
```

packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -1,61 +1,87 @@
...
+  if (priority === AUTO_REJECT || priority === USER_REJECT) {
+    return false
   }
 
-  return parsed
-}
-
-function sampleRequest (req) {
-  if (!enabled || !requestSampling) {
+  if (!priority && !rootSpan._prioritySampler?.isSampled(rootSpan)) {
```
iunanua, line 38: now we are calling `rootSpan._prioritySampler?.sample(rootSpan)` so this must be removed
```suggestion
 
```

iunanua, line 38: now we are calling rootSpan._prioritySampler?.sample(rootSpan) so this block must be removed
```javascript
// remove it
if (!priority && !rootSpan._prioritySampler?.isSampled(rootSpan)) {
  return false
}
```
(github UI doesn't allow to do a proper suggestion because it contains deleted lines)

```suggestion

```

packages/dd-trace/src/config.js
```diff
@@ -674,7 +674,7 @@ class Config {
       DD_API_SECURITY_ENABLED && isTrue(DD_API_SECURITY_ENABLED),
       DD_EXPERIMENTAL_API_SECURITY_ENABLED && isTrue(DD_EXPERIMENTAL_API_SECURITY_ENABLED)
     ))
-    this._setUnit(env, 'appsec.apiSecurity.requestSampling', DD_API_SECURITY_REQUEST_SAMPLE_RATE)
+    this._setValue(env, 'appsec.apiSecurity.sampleDelay', maybeFloat(DD_API_SECURITY_SAMPLE_DELAY))
```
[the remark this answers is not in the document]
IlyasShabi: The Previous `requestSampling` was a rate which ranged between 0 and 1.
```
this._setValue(obj, name, Math.min(Math.max(value, 0), 1))
```
`sampleDelay` represents a value in seconds

packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -1,61 +1,87 @@
...
+function computeKey (req, res) {
+  const route = web.getContext(req).paths.join('') || ''
+  const method = req.method
+  const status = res.statusCode
+
+  if (!method || !status) {
+    log.warn('Unsupported groupkey for API security')
+    return null
+  }
+  return method + route + status
```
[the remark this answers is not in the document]
IlyasShabi: Yes indeed simple string concatenation should be enough. Hashing might save some space, it adds extra processing and may not be worth the cost, especially if our data size is already manageable.

packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -1,61 +1,87 @@
...
 
-const sampledRequests = new WeakSet()
+let enabled
+let sampledRequests
 
 function configure ({ apiSecurity }) {
   enabled = apiSecurity.enabled
-  setRequestSampling(apiSecurity.requestSampling)
+  const delay = apiSecurity.sampleDelay || DEFAULT_DELAY
+  sampledRequests = new LRUCache({ max: MAX_SIZE, ttl: delay * 1000 })
```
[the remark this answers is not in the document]
IlyasShabi: I agree that `lru-cache` doesn’t have strong TTL guarantees, which is why I implemented it from scratch instead of using `@isaacs/ttlcache` (I chose to go with the custom approach).
IMHO I can create a separate PR to switch to `@isaacs/ttlcache` and update all the relevant areas such as https://github.com/DataDog/dd-trace-js/blob/ba53c01cf2bee861028fc5bc96ef292de3e3bbd6/packages/dd-trace/src/debugger/devtools_client/status.js#L21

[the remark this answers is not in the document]
simon-id: wdym you implemented it from scratch ? aren't you just using the lru-cache module ?

packages/dd-trace/src/config.js
```diff
@@ -842,7 +842,6 @@ class Config {
     tagger.add(tags, options.tags)
 
     this._setBoolean(opts, 'appsec.apiSecurity.enabled', options.appsec.apiSecurity?.enabled)
-    this._setUnit(opts, 'appsec.apiSecurity.requestSampling', options.appsec.apiSecurity?.requestSampling)
```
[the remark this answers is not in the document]
IlyasShabi: According to the RFC there is no programmatic option for this, just an env var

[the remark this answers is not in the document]
simon-id: Yeah the RFC rarely mention those as they're language dependant, it's really up to us to decide if we want to make them available or not. I'd say why not ?

docs/test.ts
IlyasShabi: We checked the values being used, and no one is using a custom sampling rate only the default. It's safe to remove it.

packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -1,61 +1,87 @@
...
-    log.warn(`Incorrect API Security request sampling value: ${requestSampling}`)
+  if (priority === AUTO_REJECT || priority === USER_REJECT) {
+    return false
+  }
 
-    parsed = 0
-  } else {
-    parsed = Math.min(1, Math.max(0, parsed))
+  if (force) {
+    return sample(req, res)
```
simon-id, line 43: I think you could replace this whole line by just

```suggestion
    sampledRequests.set(key)
```
I think it would be equivalent

packages/dd-trace/src/config.js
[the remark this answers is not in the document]
IlyasShabi: AFAIK backend set the 30s window for better accuracy. It’s mainly for backend use, not user configuration.
the env var is just for testing, and changing it could be risky

packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -1,61 +1,87 @@
...
+const DEFAULT_DELAY = 30 // 30s
 
-const sampledRequests = new WeakSet()
+let enabled
+let sampledRequests
 
 function configure ({ apiSecurity }) {
   enabled = apiSecurity.enabled
-  setRequestSampling(apiSecurity.requestSampling)
+  const delay = apiSecurity.sampleDelay || DEFAULT_DELAY
```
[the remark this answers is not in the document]
simon-id: actually i'm not sure it's a good idea to have two default values, one here and one in config.js.
If it's already in config.js like it should be, i don't really see why we should have it here too. I don't think we have a second default check like this anywhere else in our code ?

packages/dd-trace/src/appsec/api_security_sampler.js
simon-id: why was this comment resolved @IlyasShabi ? when resolving comments it's nice to put a conclusion in the thread so we know why we dismissed it ?

packages/dd-trace/src/appsec/api_security_sampler.js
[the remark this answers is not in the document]
iunanua: yes, you should trust config values here because, in theory, default values are set in `config.js` (or should they be)

packages/dd-trace/src/appsec/api_security_sampler.js
IlyasShabi: @simon-id I resolved this issue based on your proposal. I removed the `sample()` function and replaced it with your suggested approach, you can notice that the original function is now outdated.

packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -1,61 +1,81 @@
...
+  }
 
-    parsed = 0
-  } else {
-    parsed = Math.min(1, Math.max(0, parsed))
+  if (force) {
+    const key = computeKey(req, res)
+    if (!key || sampledRequests.has(key)) return false
+    sampledRequests.set(key)
+    return true
```
iunanua, line 46: not needed, no?
```suggestion
```

packages/dd-trace/src/appsec/api_security_sampler.js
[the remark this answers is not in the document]
IlyasShabi: Done

packages/dd-trace/src/appsec/api_security_sampler.js
iunanua: I think what @simon-id means is that there is a check at the beginning of the sampleRequest function to know if the req is sampled so if you have reached this point there is no need to check again if `sampledRequests.has(key)`. The above check ensures that it is not.

packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -1,61 +1,79 @@
...
+
+  if (priority === AUTO_REJECT || priority === USER_REJECT) {
+    return false
+  }
 
-    parsed = 0
-  } else {
-    parsed = Math.min(1, Math.max(0, parsed))
+  if (force) {
+    const key = computeKey(req, res)
```
iunanua, line 41: we could avoid to compute the key twice moving this call to the beginning of the `sampleRequest` and reuse the key in `isSampled(key)` and here you could simply set it `sampledRequests.set(key)` (as @simon-id suggested)

packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -1,61 +1,79 @@
...
+  return true
 }
 
-function sampleRequest (req) {
-  if (!enabled || !requestSampling) {
-    return false
-  }
+function isSampled (req, res) {
+  const key = computeKey(req, res)
+  if (!key) return false
```
iunanua, line 52: when there has been a problem computing the key, should we return true to avoid sampling the request?
```suggestion
  if (!key) return true
```

packages/dd-trace/src/appsec/api_security_sampler.js
IlyasShabi: Yes it's clear now, I did push a fix

packages/dd-trace/src/appsec/api_security_sampler.js
IlyasShabi: Done

packages/dd-trace/src/appsec/api_security_sampler.js
IlyasShabi: @iunanua Do you think this is still relevant after the last commit?

iunanua: nop, resolving :)

iunanua:
LGTM, let's wait for @simon-id opinion

iunanua (dismissed):

packages/dd-trace/test/config.spec.js
```diff
@@ -1154,7 +1154,7 @@ describe('Config', () => {
     process.env.DD_APPSEC_GRAPHQL_BLOCKED_TEMPLATE_JSON = BLOCKED_TEMPLATE_JSON_PATH // json and html here
     process.env.DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING = 'disabled'
     process.env.DD_API_SECURITY_ENABLED = 'false'
-    process.env.DD_API_SECURITY_REQUEST_SAMPLE_RATE = 0.5
+    process.env.DD_API_SECURITY_SAMPLE_DELAY = '10'
```
simon-id, line 1157: there is no more option for this test, no need to test it

packages/dd-trace/test/config.spec.js
```diff
@@ -1299,7 +1298,7 @@ describe('Config', () => {
     expect(config).to.have.nested.property('appsec.eventTracking.enabled', true)
     expect(config).to.have.nested.property('appsec.eventTracking.mode', 'safe')
     expect(config).to.have.nested.property('appsec.apiSecurity.enabled', true)
-    expect(config).to.have.nested.property('appsec.apiSecurity.requestSampling', 1.0)
+    expect(config).to.have.nested.property('appsec.apiSecurity.sampleDelay', 10)
```
simon-id, line 1301: no more options, no need to check it anymore

packages/dd-trace/test/appsec/index.express.plugin.spec.js
```diff
@@ -307,9 +307,9 @@ withVersions('express', 'express', version => {
       appsec.disable()
     })
 
-    describe('with requestSampling 1.0', () => {
+    describe('with sample delay 10', () => {
       beforeEach(() => {
-        config.appsec.apiSecurity.requestSampling = 1.0
+        config.appsec.apiSecurity.sampleDelay = 1.0
```
simon-id, line 312: uhm ? is it 10 or 1.0 ?

simon-id:
Code LGTM now, but haven't had time to fully review the tests. I found some problems

packages/dd-trace/test/config.spec.js
```diff
@@ -1402,7 +1399,7 @@ describe('Config', () => {
       },
       apiSecurity: {
         enabled: true,
-        requestSampling: 1.0
+        sampleDelay: 30
```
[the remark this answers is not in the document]
IlyasShabi: Done

packages/dd-trace/test/config.spec.js
IlyasShabi: Done

packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -1,61 +1,77 @@
...
   enabled = false
+  sampledRequests?.clear()
 }
 
-function setRequestSampling (sampling) {
-  requestSampling = parseRequestSampling(sampling)
-}
+function sampleRequest (req, res, force = false) {
+  if (!enabled) return false
+  if (isSampled(req, res)) return false
```
simon-id, line 25: bad copy paste ? :raised_eyebrow:

packages/dd-trace/test/config.spec.js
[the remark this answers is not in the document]
simon-id: of because the test fail if the object isn't complete ?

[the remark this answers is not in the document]
simon-id: `expect(config).to.have.deep.property` yeah i think that means it needs all the props ok

packages/dd-trace/src/appsec/api_security_sampler.js
IlyasShabi: I had some conflicts seems I duplicate the code

simon-id:
You got some system tests failures for API Security scenario, is it just the old tests that we should disable now ?

simon-id:
Code and tests LGTM, but let's resolve that CI failure

IlyasShabi:
In some scenarios, we need to specify that the sampling delay is 0 `"DD_API_SECURITY_SAMPLE_DELAY": "0.0"` which means we want to sample all requests. Unfortunately, the `@isaacs/ttlcache` library does not support `ttl: 0`, so we need to handle this ourselves.

Here are some failing system test scenarios:

- [APPSEC_API_SECURITY](https://github.com/DataDog/system-tests/blob/371287963f89a4bc4c7f30893e9e50a7daa3a860/utils/_context/_scenarios/__init__.py#L258)
-  [APPSEC_API_SECURITY_RC](https://github.com/DataDog/system-tests/blob/371287963f89a4bc4c7f30893e9e50a7daa3a860/utils/_context/_scenarios/__init__.py#L270)

Solutions:

- As implemented in the last commit, if the delay is 0, we instantiate a dummy TTL cache with basic functions to handle this edge case.
- Completely remove the `@isaacs/ttlcache` library and replace it with a custom implementation, similar to the python implementation

WDYT? @iunanua @simon-id

packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -1,61 +1,84 @@
...
+  set (key) { return undefined }
+  has (key) { return false }
+}
 
 function configure ({ apiSecurity }) {
   enabled = apiSecurity.enabled
-  setRequestSampling(apiSecurity.requestSampling)
+  sampledRequests = apiSecurity.sampleDelay === 0
+    ? new NoopTTLCache()
+    : new TTLCache({ max: MAX_SIZE, ttl: apiSecurity.sampleDelay * 1000 })
```
simon-id, line 23: since in config you do `maybeFloat()` to parse this field, if it's not a number, it will return undefined. What will this code do if sampleDelay is undefined ?

simon-id: I think it will be NaN because undefined * 1000 is NaN

simon-id: from TTLCache code
```javascript
    if (ttl !== undefined && !isPosIntOrInf(ttl)) {
      throw new TypeError(
        'ttl must be positive integer or Infinity if set'
      )
    }
```

simon-id: Should we let the startup throw ? or should we ignore with a default value ?

simon-id: ok apparently it takes the default value from config.js

simon-id (approved):

PR #4759, opened 2024-10-03
modernize eslint config

bengl:
### What does this PR do?

* Switch from the old eslintrc format to the newer format via: `npx @eslint/migrate-config .eslintrc.json`
* ECMAScript version is now set at 2022, in line with code supported in Node.js 16. This is needed for a bunch of ESM syntax like top-level await.
* Fixes:
    * ESM files are now covered.
    * Test globals and other test-specific config are now isolated to tests.
    * text_map.js has an invalid switch case. Fixed that in what I thought was the most reasonable way.

### Motivation

I was trying to add new local lint rules, but that's only supported by the modern config format. Along the way I noticed we weren't covering ESM files at all.


packages/dd-trace/src/opentracing/propagation/text_map.js
```diff
@@ -285,14 +285,17 @@ class TextMapPropagator {
...
           spanContext = this._extractB3SingleContext(carrier)
           break
         case 'b3':
+          if (this._config.tracePropagationStyle.otelPropagators) {
+            // TODO: should match "b3 single header" in next major
+            spanContext = this._extractB3SingleContext(carrier)
+          } else {
+            spanContext = this._extractB3MultiContext(carrier)
+          }
+          break
```
bengl, line 298: @khanayan123 You added the original lines here, so can you verify that this logic is correct?

integration-tests/forwarder-1723486776890.out
```diff
@@ -0,0 +1 @@
+library_entrypoint	{"metadata":{"language_name":"nodejs","language_version":"22.2.0","runtime_name":"nodejs","runtime_version":"22.2.0","tracer_version":"6.0.0-pre","pid":38665},"points":[{"name":"library_entrypoint.complete","tags":["injection_forced:false"]}]}
```
rochdev, line 1: Is this file expected?

bengl: nope. will remove.

package.json
```diff
@@ -109,6 +109,8 @@
     "tlhunter-sorted-set": "^0.1.0"
   },
   "devDependencies": {
+    "@eslint/eslintrc": "^3.1.0",
```
rochdev, line 112: It looks like the old dependency is also still installed if I understand this properly.

rochdev (changes requested):

package.json
bengl: @rochdev `@eslint/eslintrc` is not `@eslint/eslint`, if that's what you're talking about? `@eslint/eslintrc` is a compatibility layer allowing new-style config to be generated from old-style config.

See https://github.com/DataDog/dd-trace-js/pull/4759/files/d049b1eb11fc33c065fbe6a370318fbe74fe8080#diff-9601a8f6c734c2001be34a2361f76946d19a39a709b5e8c624a2a5a0aade05f2R7 and usages of it in that file.

rochdev: I misunderstood that upgrading eslint was part of this PR when it's not.

rochdev (approved):

eslint.config.mjs
```diff
@@ -0,0 +1,126 @@
...
+    },
+
+    settings: {
+      node: {
+        version: '>=16.0.0'
+      }
+    },
+
+    rules: {
+      'max-len': [2, 120, 2],
```
watson, line 61: Maybe it's out of scope for this PR, but the `max-len` rule has been deprecated since ESLint 8.53.0. We should instead use the [corresponding rule](https://eslint.style/rules/js/max-len) in [@stylistic/eslint-plugin-js](https://eslint.style/packages/js).

watson, line 61: I can't find any documentation for `max-len` saying that the config we use is valid, or what it means, making reading and understanding this line difficult. If I read the source code I can see that index 1 in the array is the line length for code and index 2 is the tab width.

The proper way to configure `max-len` (both the old deprecated version and the new from [@stylistic/eslint-plugin-js](https://eslint.style/packages/js) is to use a config object, e.g:

```suggestion
      'max-len': ['error', { code: 120, tabWidth: 2 }],
```

eslint.config.mjs
```diff
@@ -0,0 +1,126 @@
...
+      'no-var': 2,
+      'no-console': 2,
+      'prefer-const': 2,
+      'object-curly-spacing': [2, 'always'],
```
watson, line 65: Same as with `max-len`, this rule has been deprecated since ESLint 8.53.0. We should instead use the [corresponding rule](https://eslint.style/rules/js/object-curly-spacing) in [@stylistic/eslint-plugin-js](https://eslint.style/packages/js).

eslint.config.mjs
```diff
@@ -0,0 +1,126 @@
...
+      'import/no-extraneous-dependencies': 2,
+      'standard/no-callback-literal': 0,
+      'no-prototype-builtins': 0,
+      'n/no-restricted-require': [2, ['diagnostics_channel']],
+      'n/no-callback-literal': 0,
+
+      'object-curly-newline': ['error', {
+        multiline: true,
+        consistent: true
+      }],
```
watson, line 75: Same as with `max-len`, this rule has been deprecated since ESLint 8.53.0. We should instead use the [corresponding rule](https://eslint.style/rules/js/object-curly-newline) in [@stylistic/eslint-plugin-js](https://eslint.style/packages/js).

eslint.config.mjs
```diff
@@ -0,0 +1,126 @@
...
+
+      'import/no-absolute-path': 0,
+      'no-unused-expressions': 0
```
watson, line 78: I think it makes the file much more readable if you use the string literals for `off`, `warn`, and `error` instead of the numbers 0-2. So this block would be:

```suggestion
      'max-len': ['error', 120, 2],
      'no-var': 'error',
      'no-console': 'error',
      'prefer-const': 'error',
      'object-curly-spacing': ['error', 'always'],
      'import/no-extraneous-dependencies': 'error',
      'standard/no-callback-literal': 'off',
      'no-prototype-builtins': 'off',
      'n/no-restricted-require': ['error', ['diagnostics_channel']],
      'n/no-callback-literal': 'off',

      'object-curly-newline': ['error', {
        multiline: true,
        consistent: true
      }],

      'import/no-absolute-path': 'off',
      'no-unused-expressions': 'off'
```

And then the equivalent with the other `rules` objects below.

watson (commented):
I love it! Thanks for doing this ☺️

Besides the review comments that I just left, I think it would be a good idea if we sorted the rules in the `rules` objects alphabetically. That way all rules from for example `mocha` or `import` would all be grouped together and there would never be any doubt where a new rule should be added.

eslint.config.mjs
```diff
@@ -0,0 +1,126 @@
...
+      '**/dist',
+      '**/docs',
+      '**/out',
+      '**/node_modules',
+      '**/versions',
+      '**/acmeair-nodejs',
+      '**/vendor',
+      'integration-tests/esbuild/out.js',
+      'integration-tests/esbuild/aws-sdk-out.js',
+      'packages/dd-trace/src/appsec/blocked_templates.js'
```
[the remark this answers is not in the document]
bengl: The ignore list is pulled from the previous `.eslintignore` file. I'm deliberately not changing the ignore list in order to minimize impact of this change.

I'll add a TODO for adding comments explaining why each one is ignored.

eslint.config.mjs
```diff
@@ -0,0 +1,126 @@
...
+  }, ...compat.extends('eslint:recommended', 'standard', 'plugin:mocha/recommended'), {
+    plugins: {
+      mocha,
+      n
+    },
+
+    languageOptions: {
+      globals: {
+        ...globals.node,
+        ...globals.mocha,
```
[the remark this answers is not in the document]
bengl: That's correct. This is a copypasta error. Will fix.

packages/dd-trace/src/opentracing/propagation/text_map.js
khanayan123: This looks right to me

eslint.config.mjs
bengl: @watson do you know of a plugin or rule that would enforce this?

watson: Haha I haven't checked if one exists, but I wouldn't be surprised

eslint.config.mjs
```diff
@@ -0,0 +1,120 @@
...
+      'n/no-restricted-require': ['error', ['diagnostics_channel']],
+      'n/no-callback-literal': 'off',
+
+      '@stylistic/js/object-curly-newline': ['error', {
+        multiline: true,
+        consistent: true
+      }],
+
+      'import/no-absolute-path': 'off',
+      'no-unused-expressions': 'off'
```
watson, line 72: Nit: I still prefer that we sorted these:

```suggestion
      '@stylistic/js/max-len': ['error', { code: 120, tabWidth: 2 }],
      '@stylistic/js/object-curly-newline': ['error', { multiline: true, consistent: true }],
      '@stylistic/js/object-curly-spacing': ['error', 'always'],
      'import/no-absolute-path': 'off',
      'import/no-extraneous-dependencies': 'error',
      'n/no-callback-literal': 'off',
      'n/no-restricted-require': ['error', ['diagnostics_channel']],
      'no-console': 'error',
      'no-prototype-builtins': 'off',
      'no-unused-expressions': 'off',
      'no-var': 'error',
      'prefer-const': 'error',
      'standard/no-callback-literal': 'off'
```

eslint.config.mjs
```diff
@@ -0,0 +1,120 @@
...
+      'mocha/no-setup-in-describe': 'off',
+      'mocha/no-sibling-hooks': 'off',
+      'mocha/no-top-level-hooks': 'off',
+      'mocha/max-top-level-suites': 'off',
+      'mocha/no-identical-title': 'off',
+      'mocha/no-global-tests': 'off',
+      'mocha/no-exports': 'off',
+      'mocha/no-skipped-tests': 'off',
+      'n/handle-callback-err': 'off',
+      'no-loss-of-precision': 'off'
```
watson, line 106: Nit: same here:

```suggestion
      'mocha/max-top-level-suites': 'off',
      'mocha/no-exports': 'off',
      'mocha/no-global-tests': 'off',
      'mocha/no-identical-title': 'off',
      'mocha/no-mocha-arrows': 'off',
      'mocha/no-setup-in-describe': 'off',
      'mocha/no-sibling-hooks': 'off',
      'mocha/no-skipped-tests': 'off',
      'mocha/no-top-level-hooks': 'off',
      'n/handle-callback-err': 'off',
      'no-loss-of-precision': 'off'
```

watson (changes requested):
You missed a few updates to to the `max-len` rule rename in `integration-tests/debugger/index.spec.js` - most likely because they are in a file that was just merged into `master` and your local checked out version hasn't been updated to include this file.

eslint.config.mjs
watson: Did some quick googling today and I haven't been able find anything 🙃

watson (approved):

tlhunter (approved):

PR #4762, opened 2024-10-04
[test visibility] Add option to automatically report logs within tests when using `winston`

juan-fernandez:
### What does this PR do?

Add instrumentation to `winston` to automatically add a transport configuration that reports logs to datadog agentlessly, following the same instructions shown in the docs today for agentless logs: https://docs.datadoghq.com/logs/log_collection/nodejs/?tab=winston30#agentless-logging. This only happens if `LogSubmissionPlugin` is active, which only happens if `DD_AGENTLESS_LOG_SUBMISSION_ENABLED` is passed.

### Motivation
Same as we do for other languages like Java ([see docs](https://docs.datadoghq.com/tests/correlate_logs_and_tests?tab=cloudciprovideragentless#setup)), we want to be able to automatically report logs to Datadog's logs product if `DD_AGENTLESS_LOG_SUBMISSION_ENABLED` is passed. 

ℹ️ Only `winston` in `jest`, `mocha` and `cucumber` is supported at the moment.

```javascript
const { createLogger, format, transports } = require('winston')

const logger = createLogger({
  level: 'info',
  exitOnError: false,
  format: format.json(),
  transports: [
    new transports.Console()
  ]
})

describe('test', () => {
  it('should return true', () => {
    logger.log('info', 'Hello simple log!') // this log will be automatically submitted if `DD_AGENTLESS_LOG_SUBMISSION_ENABLED` is defined

    expect(true).to.be.true
  })
})
```

### Plugin Checklist

- [x] Unit tests.

### DIFF
--- a/integration-tests/automatic-log-submission.spec.js
+++ b/integration-tests/automatic-log-submission.spec.js
@@ -12,13 +12,22 @@ const {
 } = require('./helpers')
 const { FakeCiVisIntake } = require('./ci-visibility-intake')
 const webAppServer = require('./ci-visibility/web-app-server')
+const { NODE_MAJOR } = require('../version')
+
+const cucumberVersion = NODE_MAJOR <= 16 ? '9' : 'latest'
 
 describe('test visibility automatic log submission', () => {
   let sandbox, cwd, receiver, childProcess, webAppPort
   let testOutput = ''
 
   before(async () => {
-    sandbox = await createSandbox(['mocha', 'jest', 'winston', 'chai@4'], true)
+    sandbox = await createSandbox([
+      'mocha',
+      `@cucumber/cucumber@${cucumberVersion}`,
+      'jest',
+      'winston',
+      'chai@4'
+    ], true)
     cwd = sandbox.folder
     webAppPort = await getPort()
     webAppServer.listen(webAppPort)
@@ -48,6 +57,10 @@ describe('test visibility automatic log submission', () => {
     {
       name: 'jest',
       command: 'node ./node_modules/jest/bin/jest --config ./ci-visibility/automatic-log-submission/config-jest.js'
+    },
+    {
+      name: 'cucumber',
+      command: './node_modules/.bin/cucumber-js ci-visibility/automatic-log-submission-cucumber/*.feature'
     }
   ]
 
--- a/integration-tests/automatic-log-submission.spec.js
+++ b/integration-tests/automatic-log-submission.spec.js
@@ -113,11 +113,6 @@ describe('test visibility automatic log submission', () => {
       })
 
       it('does not submit logs when DD_AGENTLESS_LOG_SUBMISSION_ENABLED is not set', (done) => {
-        receiver
-          .gatherPayloadsMaxTimeout(({ url }) => url.includes('/api/v2/logs'), payloads => {
-            throw new Error('should not report logs')
-          }).catch(done)
-
         childProcess = exec(command,
           {
             cwd,
@@ -145,11 +140,6 @@ describe('test visibility automatic log submission', () => {
       })
 
       it('does not submit logs when DD_AGENTLESS_LOG_SUBMISSION_ENABLED is set but DD_API_KEY is not', (done) => {
-        receiver
-          .gatherPayloadsMaxTimeout(({ url }) => url.includes('/api/v2/logs'), payloads => {
-            throw new Error('should not report logs')
-          }).catch(done)
-
         childProcess = exec(command,
           {
             cwd,
--- a/integration-tests/automatic-log-submission.spec.js
+++ b/integration-tests/automatic-log-submission.spec.js
@@ -67,20 +67,43 @@ describe('test visibility automatic log submission', () => {
   testFrameworks.forEach(({ name, command }) => {
     context(`with ${name}`, () => {
       it('can automatically submit logs', (done) => {
-        const receiverPromise = receiver
+        let logIds, testIds
+
+        const logsPromise = receiver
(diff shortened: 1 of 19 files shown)


packages/dd-trace/src/plugins/log_plugin.js
```diff
@@ -54,7 +54,7 @@ module.exports = class LogPlugin extends Plugin {
   configure (config) {
     return super.configure({
       ...config,
-      enabled: config.enabled && config.logInjection
+      enabled: config.enabled && (config.logInjection || config.ciVisAgentlessLogSubmissionEnabled)
```
juan-fernandez, line 57: we also want to log injection if we're automatically submitting logs from test visibility

integration-tests/selenium/selenium.spec.js
```diff
@@ -18,7 +18,7 @@ const {
 } = require('../../packages/dd-trace/src/plugins/util/test')
 const { NODE_MAJOR } = require('../../version')
 
-const cucumberVersion = NODE_MAJOR <= 16 ? '9' : '10'
+const cucumberVersion = NODE_MAJOR <= 16 ? '9' : 'latest'
```
juan-fernandez, line 21: unrelated fix

packages/dd-trace/src/ci-visibility/log-submission/log-submission-plugin.js
```diff
@@ -0,0 +1,49 @@
+const Plugin = require('../../plugins/plugin')
+const log = require('../../log')
+
+function getWinstonLogSubmissionParameters (config) {
+  const { site, service } = config
+
+  const defaultParameters = {
+    host: `http-intake.logs.${site}`,
+    path: `/api/v2/logs?dd-api-key=${process.env.DD_API_KEY}&ddsource=winston&service=${service}`,
```
nikita-tkachenko-datadog, line 9: I think it is safer to include the API key as a request header (`dd-api-key`), rather than as part of the query string.

nikita-tkachenko-datadog: Although taking a closer look at this, I am not sure if this is what `winston` allows.

packages/dd-trace/src/ci-visibility/log-submission/log-submission-plugin.js
```diff
@@ -0,0 +1,49 @@
...
+    return defaultParameters
+  }
+
+  try {
+    const url = new URL(process.env.DD_AGENTLESS_LOG_SUBMISSION_URL)
+    return {
+      host: url.hostname,
+      port: url.port,
+      ssl: url.protocol === 'https:',
+      path: defaultParameters.path
```
nikita-tkachenko-datadog, line 23: Again, not sure if this is possible with `winston`, but gzipping the logs would make sense.

integration-tests/automatic-log-submission.spec.js
nikita-tkachenko-datadog, line 1: Does/should it test cases when the logs are written not by the test itself but by the logic that it covers?

nikita-tkachenko-datadog, line 1: Does/should it test cases when the logs are written not by the test itself but by the logic that it covers?

nikita-tkachenko-datadog (approved):

integration-tests/automatic-log-submission.spec.js
juan-fernandez: oh yeah, that's a good test. The log being in the test code or somewhere should be the same (what matters is the active span), but it's a good test, thanks 👍

packages/dd-trace/src/ci-visibility/log-submission/log-submission-plugin.js
juan-fernandez: safer because there's less likelihood of the API being logged somewhere, right? 

It is possible to set headers, yeah. I was following the instructions shown by APM for the APM-logs integration, which are more tested: https://docs.datadoghq.com/logs/log_collection/nodejs/?tab=winston30#agentless-logging. In this case I think it's worth doing the change though

packages/dd-trace/src/ci-visibility/log-submission/log-submission-plugin.js
juan-fernandez: this complicates the integration considerably and not even the APM-logs integration suggests it, so I'll keep it: https://docs.datadoghq.com/logs/log_collection/nodejs/?tab=winston30#agentless-logging

packages/dd-trace/src/ci-visibility/log-submission/log-submission-plugin.js
nikita-tkachenko-datadog: > safer because there's less likelihood of the API key being logged somewhere
yep, this is what I had in mind

integration-tests/automatic-log-submission.spec.js
juan-fernandez: done in https://github.com/DataDog/dd-trace-js/pull/4762/commits/1857aedc7c0d446e4c408c91b997a9012d4f2874

packages/dd-trace/src/ci-visibility/log-submission/log-submission-plugin.js
juan-fernandez: done in https://github.com/DataDog/dd-trace-js/pull/4762/commits/0d70afb92e060a5c11b08a5a1709d8f0e003e3d8

nikita-tkachenko-datadog (approved):

PR #4763, opened 2024-10-04
Keep a profiling context object in spans

szegedi:
### What does this PR do?
Keeps a profiling context object in spans. Allows for cheapest sampling context updates on async context switches. We already keep profiling data (web tags) associated with spans, this update rationalizes how we store data there.

### Motivation
Two reasons:
* An async context change in fast path now boils down to just `this._currentContext.ref = storage.getStore().span[ProfilingContext]`. It doesn't get cheaper than that.
* This creates a natural spot where to hang further [profiling-specific context data](https://docs.datadoghq.com/profiler/guide/isolate-outliers-in-monolithic-services/#isolate-your-own-business-logic).

### Additional Notes
This doesn't cause significant increase in memory use. Profiling context objects are small (3 properties) and their property values are object references to data available elsewhere in a span. Every context is only associated with its span until the span finishes, when the context is proactively removed, just as before with web tags.

Jira ticket: [PROF-10669]

[PROF-10669]: https://datadoghq.atlassian.net/browse/PROF-10669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/src/profiling/profilers/wall.js
+++ b/packages/dd-trace/src/profiling/profilers/wall.js
@@ -20,7 +20,7 @@ const enterCh = dc.channel('dd-trace:storage:enter')
 const spanFinishCh = dc.channel('dd-trace:span:finish')
 const profilerTelemetryMetrics = telemetryMetrics.manager.namespace('profilers')
 
-const MemoizedWebTags = Symbol('NativeWallProfiler.MemoizedWebTags')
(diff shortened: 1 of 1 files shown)


nsavoire (approved):
LGTM, this nicely simplifies the code !

PR #4764, opened 2024-10-08
Use static vulnerability hash source when the cookie name is too long

uurien:
### What does this PR do?

Use static vulnerability hash source when the cookie name is too long.

### Motivation

Prevent identifying multiple vulnerabilities when there is only one. For example, when the cookie name contains a hash or timestamp.

### Additional Notes

System tests PR: https://github.com/DataDog/system-tests/pull/3181

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2163,7 +2163,7 @@ declare namespace tracer {
      * Defines the pattern to ignore cookie names in the vulnerability hash calculation
      * @default ".{32,}"
      */
-    cookieFilterPattern?: boolean,
+    cookieFilterPattern?: string,
 
     /**
      * Whether to enable vulnerability deduplication
--- a/index.d.ts
+++ b/index.d.ts
@@ -1739,7 +1739,6 @@ declare namespace tracer {
      * on the tracer.
      */
     interface pino extends Integration {}
-  
     /**
      * This plugin automatically patches the [protobufjs](https://protobufjs.github.io/protobuf.js/)
      * to collect protobuf message schemas when Datastreams Monitoring is enabled.
@@ -2159,6 +2158,12 @@ declare namespace tracer {
      * @default 2
      */
     maxContextOperations?: number,
+
+    /**
+     * Defines the pattern to ignore cookie names in the vulnerability hash calculation
+     * @default ".{32,}"
+     */
+    cookieFilterPattern?: boolean,
 
     /**
      * Whether to enable vulnerability deduplication
--- a/packages/dd-trace/src/appsec/iast/analyzers/cookie-analyzer.js
+++ b/packages/dd-trace/src/appsec/iast/analyzers/cookie-analyzer.js
@@ -12,7 +12,11 @@ class CookieAnalyzer extends Analyzer {
   }
 
   onConfigure (config) {
-    this.cookieFilterRegExp = new RegExp(config.iast.cookieFilterPattern)
+    try {
+      this.cookieFilterRegExp = new RegExp(config.iast.cookieFilterPattern)
+    } catch {
+      this.cookieFilterRegExp = /.{32,}/
+    }
 
     this.addSub(
       { channelName: 'datadog:iast:set-cookie', moduleName: 'http' },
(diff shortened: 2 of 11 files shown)


packages/dd-trace/src/appsec/iast/analyzers/cookie-analyzer.js
```diff
@@ -11,7 +11,9 @@ class CookieAnalyzer extends Analyzer {
     this.propertyToBeSafe = propertyToBeSafe.toLowerCase()
   }
 
-  onConfigure () {
+  onConfigure (config) {
+    this.cookieFilterRegExp = new RegExp(config.iast.cookieFilterPattern)
```
CarlesDD, line 15: This could throw a `SyntaxError` that, if not managed, will lead to IAST to not be enabled. I think we should catch it here and manage it.

packages/dd-trace/test/appsec/iast/analyzers/insecure-cookie-analyzer.spec.js
CarlesDD, line 1: Missing a test for an invalid Regexp from the config.

CarlesDD (changes requested):

index.d.ts
```diff
@@ -2160,6 +2159,12 @@ declare namespace tracer {
      */
     maxContextOperations?: number,
 
+    /**
+     * Defines the pattern to ignore cookie names in the vulnerability hash calculation
+     * @default ".{32,}"
+     */
+    cookieFilterPattern?: boolean,
```
CarlesDD, line 2166: I think this should be a string, not a boolean.

CarlesDD (changes requested):

packages/dd-trace/test/appsec/iast/analyzers/insecure-cookie-analyzer.spec.js
```diff
@@ -43,6 +43,12 @@ describe('insecure cookie analyzer', () => {
         res.setHeader('set-cookie', ['key=value; HttpOnly', 'key2=value2; Secure'])
       }, INSECURE_COOKIE, 1)
 
+      testThatRequestHasVulnerability((req, res) => {
+        const cookieNamePrefix = (new Array(50)).join('0')
```
CarlesDD, line 47: Maybe it would be clearer `'0'.repeat(50)`

index.d.ts
uurien: OMG, true, also missed an update in `test.ts` to check the types

packages/dd-trace/test/appsec/iast/analyzers/insecure-cookie-analyzer.spec.js
CarlesDD: Same in [packages/dd-trace/test/appsec/iast/analyzers/no-samesite-cookie-analyzer.spec.js](https://github.com/DataDog/dd-trace-js/pull/4764/files#diff-fa8ab420938d6711dbcfd5529f14d60b609a80cdc367b5920962fb1a78234d00R70) and [packages/dd-trace/test/appsec/iast/analyzers/no-httponly-cookie-analyzer.spec.js](https://github.com/DataDog/dd-trace-js/pull/4764/files#diff-96883fba5aaa9ee6fc8d343d54f44d6e88d9d8face0dc2646ceca3a901670772R59)

packages/dd-trace/src/appsec/iast/analyzers/cookie-analyzer.js
```diff
@@ -11,7 +11,13 @@ class CookieAnalyzer extends Analyzer {
     this.propertyToBeSafe = propertyToBeSafe.toLowerCase()
   }
 
-  onConfigure () {
+  onConfigure (config) {
+    try {
+      this.cookieFilterRegExp = new RegExp(config.iast.cookieFilterPattern)
+    } catch {
+      this.cookieFilterRegExp = /.{32,}/
```
CarlesDD, line 18: Worth a log record when invalid pattern

CarlesDD (approved):

PR #4765, opened 2024-10-08
Add parallelism information to profiles

szegedi:
### What does this PR do?
Adds information about available parallelism and size of libuv thread pool to profiles.

### Motivation
Combined with timeline events activity, this can allow us to determine in the backend if the libuv thread pool is undersized and offer an analysis/insight for it.

### Additional Notes
Jira link: [PROF-10686]

[PROF-10686]: https://datadoghq.atlassian.net/browse/PROF-10686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/src/profiling/exporters/event_serializer.js
+++ b/packages/dd-trace/src/profiling/exporters/event_serializer.js
@@ -2,6 +2,22 @@ const os = require('os')
 const perf = require('perf_hooks').performance
 const version = require('../../../../../package.json').version
 
+const libuvThreadPoolSize = (() => {
+  const ss = process.env.UV_THREADPOOL_SIZE
+  if (ss === undefined) {
+    // Backend will apply the default size based on Node version.
+    return undefined
+  }
+  // libuv uses atoi to parse the value, which is almost the same as parseInt, except that parseInt
+  // will return NaN on invalid input, while atoi will return 0. This is handled at return.
+  const s = parseInt(ss)
+  // We dont' interpret the value further here in the library. Backend will interpret the number
+  // based on Node version. In all currently known Node versions, 0 results in 1 worker thread,
+  // negative values (because they're assigned to an unsigned int) become very high positive values,
+  // and the value is finally capped at 1024.
+  return isNaN(s) ? 0 : s
+})()
+
 class EventSerializer {
   constructor ({ env, host, service, version, libraryInjected, activation } = {}) {
     this._env = env
@@ -65,8 +81,7 @@ class EventSerializer {
           // proscribed by the Intake API, but that's an internal enum and is
           // not customer visible.
           engine: 'nodejs',
-          // Sent as a string, we'll let the backend figure out the effective numeric value.
-          libuv_threadpool_size: process.env.UV_THREADPOOL_SIZE,
+          libuv_threadpool_size: libuvThreadPoolSize,
           // strip off leading 'v'. This makes the format consistent with other
           // runtimes (e.g. Ruby) but not with the existing `runtime_version` tag.
           // We'll keep it like this as we want cross-engine consistency. We
--- a/packages/dd-trace/src/profiling/exporters/event_serializer.js
+++ b/packages/dd-trace/src/profiling/exporters/event_serializer.js
@@ -56,11 +56,17 @@ class EventSerializer {
           version
         },
         runtime: {
+          // os.availableParallelism only available in node 18.14.0/19.4.0 and above
+          available_processors: typeof os.availableParallelism === 'function'
+            ? os.availableParallelism()
+            : os.cpus().length,
           // Using `nodejs` for consistency with the existing `runtime` tag.
           // Note that the event `family` property uses `node`, as that's what's
           // proscribed by the Intake API, but that's an internal enum and is
           // not customer visible.
           engine: 'nodejs',
+          // Sent as a string, we'll let the backend figure out the effective numeric value.
+          libuv_threadpool_size: process.env.UV_THREADPOOL_SIZE,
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/profiling/exporters/agent.js
```diff
@@ -151,11 +151,17 @@ class AgentExporter {
           version
         },
         runtime: {
+          // os.availableParallelism only available in node 18.14.0/19.4.0 and above
+          available_processors: typeof os.availableParallelism === 'function'
```
nsavoire, line 155: Note that if container has a cpu limit enforced through cpu CFS shares, the value will be wrong (eg. with a CPU limit of 0.5CPU on a 96 core system, value will be 96).

packages/dd-trace/src/profiling/exporters/agent.js
```diff
@@ -151,11 +151,17 @@ class AgentExporter {
...
+            ? os.availableParallelism()
+            : os.cpus().length,
           // Using `nodejs` for consistency with the existing `runtime` tag.
           // Note that the event `family` property uses `node`, as that's what's
           // proscribed by the Intake API, but that's an internal enum and is
           // not customer visible.
           engine: 'nodejs',
+          // Sent as a string, we'll let the backend figure out the effective numeric value.
+          libuv_threadpool_size: process.env.UV_THREADPOOL_SIZE,
```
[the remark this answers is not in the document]
szegedi: yeah, that's fine – the property will be missing here then

packages/dd-trace/src/profiling/exporters/agent.js
szegedi: Oh I see. This would still be useful, 'cause it could give an upper limit for a suggestion for `UV_THREADPOOL_SIZE`. The main use I have for this is that I could provide a profiler analysis: if there's more concurrent DNS and filesystem operations (and in future, cryptography and compression) than we have workers in the libuv worker thread pool, we can suggest the customer increase the size of the pool _as long_ as there's space for some extra parallelism in the CPU cores. K8s won't limit parallelism as looking from outside the box, you don't see the threads, it just limits the overall utilization of the CPU, right? Theory being that DNS and filesystem operations cause the libuv threads to be blocked on IO, so they're idle.

szegedi: But I could also definitely have use for `kubernetes.cpu.requests` and `kubernetes.cpu.limits` too. I don't suppose I can grab those values easily from within the Node app?

packages/dd-trace/src/profiling/exporters/agent.js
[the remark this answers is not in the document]
nsavoire: why not send the default threadpool size when it is not overridden by `UV_THREADPOOL_SIZE` ?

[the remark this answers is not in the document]
szegedi: So, a hardcoded `4`? The comment above that line says it: "we'll let the backend figure out the effective numeric value." I think this is just me instinctively preferring having logic in the backend, 'cause we can more easily fix/modify that than libraries (and have users upgrade to new libraries.) E.g. if libuv in a future Node version changes either the default or the max thread pool size.

I might go for converting early to number though, as JavaScript's `parseInt` is almost compatible with C `atoi` that libuv uses to parse it so I wouldn't have to implement an `atoi` equivalent in Java in the backend.

packages/dd-trace/src/profiling/exporters/agent.js
```diff
@@ -89,6 +89,22 @@ function computeRetries (uploadTimeout) {
...
+    return undefined
+  }
+  // libuv uses atoi to parse the value, which is almost the same as parseInt, except that parseInt
+  // will return NaN on invalid input, while atoi will return 0. This is handled at return.
+  const s = parseInt(ss)
+  // We dont' interpret the value further here in the library. Backend will interpret the number
+  // based on Node version. In all currently known Node versions, 0 results in 1 worker thread,
+  // negative values (because they're assigned to an unsigned int) become very high positive values,
+  // and the value is finally capped at 1024.
+  return isNaN(s) ? 0 : s
```
nsavoire, line 105: Wow, great attention to detail here !

packages/dd-trace/src/profiling/exporters/agent.js
```diff
@@ -89,6 +89,22 @@ function computeRetries (uploadTimeout) {
   return [tries, Math.floor(uploadTimeout)]
 }
 
+const libuvThreadPoolSize = (() => {
+  const ss = process.env.UV_THREADPOOL_SIZE
+  if (ss === undefined) {
```
nsavoire, line 94: Sorry for the noise, I did not understand that you were delegating to the backend the computation of the default value.
My first take was that the library was in a better position to determine the default value, but since the node version is part of profile metadata, it might be better to let the backend handle this, since updating code in the backend is far more easier.

packages/dd-trace/src/profiling/exporters/agent.js
nsavoire: I don't believe there is a way to get kubernetes.cpu.requests and kubernetes.cpu.limits from node app.
My point is that this value will be hard to interpret.
If threads are mostly blocked on network / IO operations, it might be beneficial to increase thread pool size even if available parallelism is low.

nsavoire (approved):

szegedi:
Had to rebase it because it conflicted with #4980

nsavoire (approved):

PR #4772, opened 2024-10-11
Add support for exit spans in Code Origin for Spans

watson:
### What does this PR do?

**Note: This PR relates to an upcoming Datadog feature that's not generally available yet.**

Support for Code Origin for Spans was added in #4449. That PR laid the groundwork for supporting this new feature and added limited support for just Fastify entry-spans. This PR adds to that original PR by adding support for all exit-spans. That is, spans that record a request going out to a different service, e.g. a database request, an outgoing HTTP request etc.

### DIFF
--- a/packages/dd-trace/src/plugins/outbound.js
+++ b/packages/dd-trace/src/plugins/outbound.js
@@ -7,6 +7,7 @@ const {
   PEER_SERVICE_REMAP_KEY
 } = require('../constants')
 const TracingPlugin = require('./tracing')
+const { exitTag } = require('../../../datadog-code-origin')
 
 const COMMON_PEER_SVC_SOURCE_TAGS = [
   'net.peer.name',
@@ -23,6 +24,14 @@ class OutboundPlugin extends TracingPlugin {
     this.addTraceSub('connect', message => {
       this.connect(message)
     })
+  }
+
+  startSpan (...args) {
+    const span = super.startSpan(...args)
+    if (this._tracerConfig.codeOriginForSpans.enabled) {
+      span.addTags(exitTag(this.startSpan))
+    }
+    return span
   }
 
   getPeerService (tags) {
--- a/packages/dd-trace/src/plugins/outbound.js
+++ b/packages/dd-trace/src/plugins/outbound.js
@@ -7,7 +7,7 @@ const {
   PEER_SERVICE_REMAP_KEY
 } = require('../constants')
 const TracingPlugin = require('./tracing')
-const { exitTag } = require('../../../datadog-code-origin')
+const { exitTags } = require('../../../datadog-code-origin')
 
 const COMMON_PEER_SVC_SOURCE_TAGS = [
   'net.peer.name',
@@ -29,7 +29,7 @@ class OutboundPlugin extends TracingPlugin {
   startSpan (...args) {
     const span = super.startSpan(...args)
     if (this._tracerConfig.codeOriginForSpans.enabled) {
-      span.addTags(exitTag(this.startSpan))
+      span.addTags(exitTags(this.startSpan))
     }
     return span
   }
--- a/packages/dd-trace/test/plugins/outbound.spec.js
+++ b/packages/dd-trace/test/plugins/outbound.spec.js
@@ -3,6 +3,7 @@
 require('../setup/tap')
 
 const { expect } = require('chai')
+const { getNextLineNumber } = require('./helpers')
 const OutboundPlugin = require('../../src/plugins/outbound')
 
 describe('OuboundPlugin', () => {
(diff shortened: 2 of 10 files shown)


watson:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4772?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4772** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4772?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4449** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4449?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

packages/dd-trace/test/plugins/outbound.spec.js
```diff
@@ -157,4 +157,75 @@ describe('OuboundPlugin', () => {
...
+      expect(topFrame).to.have.property('line', lineNumber)
+    })
+  })
 })
+
+function getNextLineNumber () {
+  return String(Number(new Error().stack.split('\n')[2].match(/:(\d+):/)[1]) + 1)
+}
+
+function parseTags (tags) {
```
watson, line 212: Do we already have a utility function that does this that I can re-use? Seem very much overkill to write this code for this test, but it does make the actual test much simpler to write once the tags are converted to a proper object structure 🤔

packages/datadog-plugin-http/test/client.spec.js
```diff
@@ -55,7 +55,7 @@ describe('Plugin', () => {
         if (appListener) {
           appListener.close()
         }
-        return agent.close({ ritmReset: false })
+        return agent.close({ ritmReset: false, wipe: true })
```
watson, line 58: I don't like this change. However when running all the tests together using `PLUGINS="http" yarn test:plugins` we see the following problem:

If we don't wipe everything after the `client.spec.js` has finished running, the following test that runs (`code_origin.spec.js`) will fail because the tracer isn't reloaded with the new tracer config passed into the `agent.load()` call. This seems like a bug in how agent.load works, but I might be wrong.

juan-fernandez (dismissed):

packages/dd-trace/test/plugins/outbound.spec.js
szegedi: I suspect https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-core/src/utils/src/set.js might fit the bill.

packages/dd-trace/src/plugins/outbound.js
```diff
@@ -25,6 +26,14 @@ class OutboundPlugin extends TracingPlugin {
     })
   }
 
+  startSpan (...args) {
+    const span = super.startSpan(...args)
+    if (this._tracerConfig.codeOriginForSpans.enabled) {
+      span.addTags(exitTag(this.startSpan))
```
shatzi, line 32: nit: would suggest rename entryTag and exitTag method names. when reading the code it look like it add just one tag, which is misleading. I was thinking... createCodeOriginTagsForExitSpan(this.startSpan) but that might be a mouthful

shatzi (approved):

rochdev (dismissed):

packages/dd-trace/test/plugins/outbound.spec.js
rochdev: > Do we already have a utility function that does this that I can re-use?

Not sure we have one right now, but it could definitely be added to `datadog-core/src/util` instead of having the code in this file.

packages/dd-trace/src/plugins/outbound.js
watson: I agree they could probably be renamed from singular to plural 👍

But I think the method names should make sense locally to the module they live in. In this case, they live inside the `datadog-code-origin` module, so adding something like "code origin" to their name is repeating information that's already present in the module name. Similar to how the functions inside of the `path` module doesn't repeat the word `path` (e.g. it's just `join` instead of `pathJoin`).

If one wants to make them distinct in the importing module, one would either import the entire module:

```js
const codeOrigin = require('../../../datadog-code-origin')
// ...
codeOrigin.exitTags()
```

Rename them on import:

```js
const { exitTags: codeOriginExitTags } = require('../../../datadog-code-origin')
// ...
codeOriginExitTags()
```

Or you just let the reader look up where they are imported from - like today.

My general rule, unless there's a critical risk of misunderstanding what's going on, is to just leave their names as is, even though it's tiny bit harder to parse the code. I feel this is the normal way naming is done in most Node.js code.

watson: I've updated the code now with the plural method names

packages/dd-trace/test/plugins/outbound.spec.js
watson: I've updated the code now and moved the function to `datadog-core/src/util`. Let me know what you think.

shatzi (approved):

PR #4775, opened 2024-10-11
add namespace support for async storage

rochdev:
### What does this PR do?

Adds centralized named storages similar to channels.

### Motivation

Right now we are mainly relying on a single global storage with products reading or writing data to it according to their needs. This has a few downsides. First, it means that holding onto one piece of data ends up holding on everything else on that store. Second, there is no control over the scope of the individual properties of the store, as everything just ends up in whatever scope the main store is.

The goal of this PR is similar to why there is a `channel(name)` helper for diagnostics channel, to provide a central registry where it doesn't matter who initializes the storage or even if it was initialized at all, and to properly isolate the scope of each individual store.

### DIFF
--- a/packages/datadog-core/index.js
+++ b/packages/datadog-core/index.js
@@ -1,7 +1,5 @@
 'use strict'
 
-const { AsyncLocalStorage } = require('async_hooks')
-
-const storage = new AsyncLocalStorage()
+const storage = require('./src/storage')
 
 module.exports = { storage }
--- a/packages/datadog-core/src/storage.js
+++ b/packages/datadog-core/src/storage.js
@@ -1,0 +1,20 @@
+'use strict'
+
+const { AsyncLocalStorage } = require('async_hooks')
+
+const storages = Object.create(null)
+const legacyStorage = new AsyncLocalStorage()
+
+const storage = function (namespace) {
+  if (!storages[namespace]) {
+    storages[namespace] = new AsyncLocalStorage()
+  }
+  return storages[namespace]
+}
+
+storage.enterWith = legacyStorage.enterWith.bind(legacyStorage)
+storage.exit = legacyStorage.exit.bind(legacyStorage)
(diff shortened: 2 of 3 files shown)


rochdev:
> that means every one of our namespace has to do enterWith() and all the continuation fixes right

Not sure I understand the question. Continuation fixes are done with AsyncResource since we're fixing the library itself which is broken. This is not done at the AsyncLocalStorage level.

If you are referring instead to the use of `enterWith` when tracing, then this doesn't apply to other products since the scoping is different (for example AppSec wants a request-level scope, LLMObs wants an LLM span scope, etc). So then yes every product would have to maintain their own async stack, but that's how it should be, otherwise everything ends up being either attached to a span (or its corresponding store) or you need to maintain your own AsyncLocalStorage instance which has other problems (described below).

> The good thing about the current single ALS store is that I can just rely on it. I don't need to maintain it.

I don't understand this statement either. What additional maintenance burden is there between doing `const asmStorage = storage('asm')` versus `const asmStorage = new AsyncLocalStorage()`?

> If I needed my own namespace I'd just do new AsyncLocalStorage(), but I don't because that's not what we need ?

You can definitely do that and it's done in a few places today. The problem with this approach is that it means that in order to share that store with other products, either some interdependency has to be introduced by explicitly importing the other product, or the stores need to all be defined in a centralized place (which could be an option too but it pushes  product knowledge to that centralized place which should be generic). You also need to determine where to instantiate the storage, which is not necessarily a big problem but not having to worry about that means that the storage can be lazily instantiated to avoid wasting memory (or eventually even the whole async_hooks import when not needed)

Basically, this does exactly what is says in the description, which is to provide a channel-like API to storages for all the same reason that the API for that is `channel(name)` instead of `new Channel()`.

simon-id:
I guess this is fine, especially as part of a bigger project of separating core and products. As long as we can read the data we all need, like current span, current request etc, i'm fine with this :+1:

simon-id (approved):

PR #4780, opened 2024-10-15
[DI] Adhere to maxCollectionSize limit in snapshots

watson:
The `maxCollectionSize` limit affects the following types:
- `Array`
- `Map` / `WeakMap`
- `Set` / `WeakSet`
- All `TypedArray` types (`UInt8Array` etc)

This limit contols the maximum about of elements collected for those types. The default is 100.

If, for example, an array contains more than the allowed amount of elements, a `notCapturedReason` of `collectionSize` is added to the collected array and the original size of the array is also captured.

### DIFF
--- a/packages/dd-trace/src/debugger/devtools_client/snapshot/collector.js
+++ b/packages/dd-trace/src/debugger/devtools_client/snapshot/collector.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const { collectionSizeSym } = require('./symbols')
 const session = require('../session')
 
 const LEAF_SUBTYPES = new Set(['date', 'regexp'])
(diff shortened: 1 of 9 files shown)


watson:
* **#4780** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4780?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/dd-trace/src/debugger/devtools_client/snapshot/collector.js
```diff
@@ -14,56 +15,66 @@ module.exports = {
...
+    // Trim the result if it's too large.
+    // Collections doesn't contain private properties, so the code in this block doesn't have to deal with it.
+    const size = countEnumerableProperties(result)
+    if (size > opts.maxCollectionSize) {
+      result = result.slice(0, opts.maxCollectionSize)
+      result[collectionSizeSym] = size
+    }
+  } else if (privateProperties) {
+    result.push(...privateProperties)
+  }
```
watson, line 34: Note: It's on purpose that I do not take into account the scenario where `collection === true` and `privateProperties !== undefined`. This is because, currently, the only time `collection === true`, is when we're either getting the properties of an `Array` or a `TypedArray`. For those two object types, we know in advance that there's no private properties, so we can make the code simpler by not having to deal with that edge case.

If we wanted to deal with this, the code would look like this:

```suggestion
  if (collection) {
    const privatePropertiesSize = privateProperties?.length ?? 0
    const totalSize = countEnumerableProperties(result) + privatePropertiesSize

    if (totalSize > opts.maxCollectionSize) {
      // Prioritize private properties over regular properties
      const spaceForNonPrivateProperties = Math.max(opts.maxCollectionSize - privatePropertiesSize, 0)
      result = result.slice(0, spaceForNonPrivateProperties)
      result[collectionSizeSym] = totalSize

      if (privateProperties) {
        const remainingSize = opts.maxCollectionSize - result.length
        if (privatePropertiesSize > remainingSize) {
          privateProperties = privateProperties.slice(0, remainingSize)
        }
      }
    }
  }

  if (privateProperties) result.push(...privateProperties)
```

juan-fernandez (dismissed):

packages/dd-trace/src/debugger/devtools_client/snapshot/collector.js
```diff
@@ -14,56 +15,67 @@ module.exports = {
...
+    // Collections doesn't contain private properties, so the code in this block doesn't have to deal with it.
+    removeNonEnumerableProperties(result) // remove the `length` property
+    const size = result.length
+    if (size > opts.maxCollectionSize) {
+      result.splice(opts.maxCollectionSize)
+      result[collectionSizeSym] = size
+    }
+  } else if (privateProperties) {
+    result.push(...privateProperties)
+  }
```
watson, line 35: Note: It's on purpose that I do not take into account the scenario where `collection === true` and `privateProperties !== undefined`. This is because, currently, the only time `collection === true`, is when we're either getting the properties of an `Array` or a `TypedArray`. For those two object types, we know in advance that there's no private properties, so we can make the code simpler by not having to deal with that edge case.

If we wanted to deal with this, the code would look like this:

```suggestion
  if (collection) {
    const privatePropertiesSize = privateProperties?.length ?? 0
    const totalSize = countEnumerableProperties(result) + privatePropertiesSize

    if (totalSize > opts.maxCollectionSize) {
      // Prioritize private properties over regular properties
      const spaceForNonPrivateProperties = Math.max(opts.maxCollectionSize - privatePropertiesSize, 0)
      result = result.slice(0, spaceForNonPrivateProperties)
      result[collectionSizeSym] = totalSize

      if (privateProperties) {
        const remainingSize = opts.maxCollectionSize - result.length
        if (privatePropertiesSize > remainingSize) {
          privateProperties = privateProperties.slice(0, remainingSize)
        }
      }
    }
  }

  if (privateProperties) result.push(...privateProperties)
```

juan-fernandez (approved):

PR #4781, opened 2024-10-15
[ASM] multer instrumentation

iunanua:
### What does this PR do?
multer instrumentation

### Motivation

be able to handle multipart request bodies parsed with multer

ST: https://github.com/DataDog/system-tests/pull/3249

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


integration-tests/multer.spec.js
```diff
@@ -0,0 +1,69 @@
...
+        proc = await spawnProc(startupTestFile, { cwd, env, execArgv })
+      })
+
+      afterEach(async () => {
+        proc.kill()
+        await agent.stop()
+      })
+
+      it('should not block the request without an attack', async () => {
+        const form = new FormData()
```
uurien, line 47: Check that this will work in v4 CI, `FormData` class was added in `v16.15.0`

packages/datadog-instrumentations/src/multer.js
```diff
@@ -0,0 +1,37 @@
...
+
+const multerReadCh = channel('datadog:multer:read:finish')
+
+function publishRequestBodyAndNext (req, res, next) {
+  return shimmer.wrapFunction(next, next => function () {
+    if (multerReadCh.hasSubscribers && req) {
+      const abortController = new AbortController()
+      const body = req.body
+
+      multerReadCh.publish({ req, res, body, abortController })
```
IlyasShabi, line 14: Since body is already part of req, is there a reason we need to pass it separately?
(We do the same thing also in body-parser)

integration-tests/multer.spec.js
iunanua: tests run with `v16.20.2`

packages/dd-trace/src/appsec/index.js
```diff
@@ -299,6 +301,7 @@ function disable () {
 
   // Channel#unsubscribe() is undefined for non active channels
   if (bodyParser.hasSubscribers) bodyParser.unsubscribe(onRequestBodyParsed)
+  if (multerParser.hasSubscribers) multerParser.unsubscribe(onRequestBodyParsed)
```
iunanua, line 304: should we check if body has been sent to the waf before in a previous `onRequestBodyParsed`? (to avoid sending the same body multiple times)

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
```diff
@@ -26,15 +26,22 @@ class TaintTrackingPlugin extends SourceIastPlugin {
...
-          this._taintTrackingHandler(HTTP_REQUEST_BODY, req, 'body', iastContext)
-          iastContext.body = req.body
-        }
-      }
+      onRequestBody
+    )
+
+    this.addSub(
+      { channelName: 'datadog:multer:read:finish', tag: HTTP_REQUEST_BODY },
+      onRequestBody
```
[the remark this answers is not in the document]
iunanua: done [here](https://github.com/DataDog/dd-trace-js/pull/4781/files#diff-889a766dd4f26d0db3b15f8ca785862ed58be0bc32a9bf81f1ea943715d9b667)

packages/datadog-instrumentations/src/multer.js
simon-id: A while ago I asked the question: "Who should be responsible of finding the data ? the instrumentation publisher ? or the instrumentation subscriber ?"
Everyone agreed that it was better if it was the responsibility of the instrumentation publisher (ie: this file) to "find" the body property.
That allows us to write framework independent subscribers.
Makes sense to you now ?

iunanua: we could get it from the `req`, yes. But I guess it's a way to decouple them. Making unnecessary for the listeners to know that the body is a property of the `req`

simon-id: Example: we could have 3 different body parser libraries, that have different kind of APIs etc, but the instrumentation point all publish with the same standardized arguments, thus allowing us to only have one subscriber for all 3 libraries.
The instrumentation acts as a "translator", it is the part of the code with the most knowledge about the context, and thus should be responsible for abstracting that in a standard signature

IlyasShabi: Yes I got your point, thanks for clarification

packages/dd-trace/src/appsec/index.js
simon-id: could there be a case where multiple body middlewares parse it differently ? and thus we'd want to send them all to make sure we analyze all forms ?

integration-tests/multer.spec.js
```diff
@@ -0,0 +1,69 @@
...
+
+describe('Suspicious request blocking - multer', () => {
+  let sandbox, cwd, startupTestFile, agent, proc, env
+
+  ['1.4.4-lts.1', '1.4.5-lts.1'].forEach((version) => {
+    describe(`v${version}`, () => {
+      before(async () => {
+        sandbox = await createSandbox(['express', `multer@${version}`])
+        cwd = sandbox.folder
+        startupTestFile = path.join(cwd, 'multer/index.js')
```
uurien, line 21: as we are using `path.join`, I think that using multiple args instead of a `/` is safer for windows (maybe these tests are never executed in windows...)
```suggestion
        startupTestFile = path.join(cwd, 'multer', index.js')
```

packages/dd-trace/src/appsec/index.js
uurien: I think that we shouldn't. 

I'm thinking in juice-shop app... with `bodyParser.text()` middleware, and multer middleware. The good one is the last one, which will set the json object in `req.body`, we want to send the last to the waf.

packages/dd-trace/test/appsec/iast/taint-tracking/plugin.express.plugin.spec.js
```diff
@@ -8,6 +8,12 @@ function noop () {}
 describe('Taint tracking plugin sources express tests', () => {
   withVersions('express', 'express', '>=4.8.0', version => {
     prepareTestServerForIastInExpress('in express', version,
+      (expressApp, listener) => {
```
uurien, line 11: why is this tests here, and not as integration test using sandbox to test with the specific multer version?

packages/dd-trace/test/appsec/index.multer.plugin.spec.js
```diff
@@ -0,0 +1,82 @@
...
+const axios = require('axios')
+const path = require('path')
+const agent = require('../plugins/agent')
+const appsec = require('../../src/appsec')
+const Config = require('../../src/config')
+const { json } = require('../../src/appsec/blocked_templates')
+
+const multerReadCh = channel('datadog:multer:read:finish')
+
+withVersions('multer', 'multer', version => {
```
uurien, line 13: same here:
why is this tests here, and not as integration test using sandbox to test with the specific multer version?

packages/dd-trace/test/appsec/iast/taint-tracking/plugin.express.plugin.spec.js
iunanua: this one is to test the body tainting, I could include it inside [integration](https://github.com/DataDog/dd-trace-js/pull/4781/files#diff-0f99f954cb822f73a211aebfe34a4d7ad9305eeabc5f0c864e5162ca923566b8) test

packages/dd-trace/test/appsec/index.multer.plugin.spec.js
iunanua: this one is the original, to test multipart request blocking but now with the [integration](https://github.com/DataDog/dd-trace-js/pull/4781/files#diff-0f99f954cb822f73a211aebfe34a4d7ad9305eeabc5f0c864e5162ca923566b8) test is kind of duplicated

packages/dd-trace/src/appsec/index.js
iunanua: okay

packages/datadog-instrumentations/src/multer.js
```diff
@@ -0,0 +1,37 @@
...
+  name: 'multer',
+  file: 'lib/make-middleware.js',
+  versions: ['^1.4.4-lts.1']
+}, makeMiddleware => {
+  return shimmer.wrapFunction(makeMiddleware, makeMiddleware => function () {
+    const middleware = makeMiddleware.apply(this, arguments)
+
+    return shimmer.wrapFunction(middleware, middleware => function wrapMulterMiddleware (req, res, next) {
+      const nextResource = new AsyncResource('bound-anonymous-fn')
+      arguments[2] = nextResource.bind(publishRequestBodyAndNext(req, res, next))
```
simon-id, line 33: the middleware is always using (req, res, next) and never (err, req, res, next) ? i know we have some instrumentations that check for that

iunanua: this is what multer returns as middleware fn
```javascript
function makeMiddleware (setup) {
  return function multerMiddleware (req, res, next) {
    ...
  }
```

packages/dd-trace/test/appsec/iast/taint-tracking/plugin.express.plugin.spec.js
iunanua: moved

packages/dd-trace/test/appsec/index.multer.plugin.spec.js
iunanua: deleted

integration-tests/multer.spec.js
CarlesDD, line 1: I think it would be better to move these files to appsec folder (integration-test/appsec)

iunanua: You think well.
Moved

CarlesDD (approved):

IlyasShabi (approved):

packages/datadog-instrumentations/src/multer.js
uurien, line 33: why are we checking this `multerReadCh.hasSubscribers` in line 14 and not here? imo, we should prevent creating AsyncResource when appsec/iast is not enabled.
```suggestion
      if (multerReadCh.hasSubscribers) {
        const nextResource = new AsyncResource('bound-anonymous-fn')
        arguments[2] = nextResource.bind(publishRequestBodyAndNext(req, res, next))
      }

```

iunanua: i was inspired by [body-parser](https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-instrumentations/src/body-parser.js) instrumentation

uurien (approved):

PR #4784, opened 2024-10-15
use jsonpath rather than jsonpath-plus

bengl:
### What does this PR do?
See title
### Motivation
jsonpath-plus removed support for Node.js 16, which we need for our v4 release line.


bengl:
Closing since jsonpath doesn't seem to play nicely with bundlers

PR #4785, opened 2024-10-15
vendor jsonpath-plus

bengl:
We need the latest version (10.0.0) so that it doesn't have vulnerabilities, but we need it to be compatible with Node.js 16.0.0, so we needed to vendor it and make slight adjustments.

### DIFF
--- a/packages/dd-trace/src/payload-tagging/jsonpath-plus.js
+++ b/packages/dd-trace/src/payload-tagging/jsonpath-plus.js
@@ -2,13 +2,46 @@
 
 // NOTE(bengl): This file is taken directly from jsonpath-plus@10.0.0
 //
-// https://github.com/JSONPath-Plus/JSONPath/blob/a04dcbac760fed48760b09f387874a36f289c3f3/dist/index-node-cjs.index-node-cjs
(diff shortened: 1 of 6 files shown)


packages/dd-trace/src/payload-tagging/jsonpath-plus.js
```diff
@@ -0,0 +1,2061 @@
+'use strict';
+
+// NOTE(bengl): This file is taken directly from jsonpath-plus@10.0.0
+//
+// https://github.com/JSONPath-Plus/JSONPath/blob/a04dcbac760fed48760b09f387874a36f289c3f3/dist/index-node-cjs.index-node-cjs
+//
+// The only changes are:
+// - Replace Object.hasOwn with polyfill
```
watson, line 8: I would either add a short description of the reason and link to this PR or mention that we can revert this inlining once we drop support for Node.js 16

bengl: I also realized that vendoring is distribution, and distribution requires the inclusion of the original license (as the license dictates), so I've included that, and a bunch more detail, in the latest rev.

simon-id (approved):

rochdev:
Is the whole functionality of every possible syntax even needed or is our use case actually smaller and could be achieved more simply?

PR #4789, opened 2024-10-16
update path-to-regexp

bengl:
### What does this PR do?
Updates path-to-regexp to the latest version
### Motivation

The old version was very old and full of vulns.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "module-details-from-path": "^1.0.3",
     "msgpack-lite": "^0.1.26",
     "opentracing": ">=0.12.1",
-    "path-to-regexp": "^0.1.10",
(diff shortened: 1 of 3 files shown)


simon-id:
Ah this rings a bell, and it looks like the CI is failing so maybe that's the reason we never updated it ? it's worth a few retries

rochdev:
> maybe that's the reason we never updated it

[This](https://github.com/DataDog/dd-trace-js/pull/468) was the reason a long time ago, but I have no idea if it's still true today (hopefully not as there has been 8 majors since then)

Worth noting however that in any case the behaviour of the library changed over the years which is technically a breaking change as we would be changing the resource names by upgrading.

PR #4791, opened 2024-10-16
[ASM] Ssrf handle request options

iunanua:
### What does this PR do?
Format and analyze the uri in the options (if it is present)

### Motivation
instrumentation provides an incomplete uri when using `request` package

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/appsec/rasp/ssrf.js
+++ b/packages/dd-trace/src/appsec/rasp/ssrf.js
@@ -21,7 +21,7 @@ function disable () {
 function analyzeSsrf (ctx) {
   const store = storage.getStore()
   const req = store?.req
-  const uri = (ctx.args.options?.uri && url.format(ctx.args.options?.uri)) ?? ctx.args.uri
+  const uri = (ctx.args.options?.uri && url.format(ctx.args.options.uri)) ?? ctx.args.uri
 
   if (!req || !uri) return
 
--- a/packages/dd-trace/src/appsec/rasp/ssrf.js
+++ b/packages/dd-trace/src/appsec/rasp/ssrf.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const url = require('url')
+const { format } = require('url')
 const { httpClientRequestStart } = require('../channels')
 const { storage } = require('../../../../datadog-core')
 const addresses = require('../addresses')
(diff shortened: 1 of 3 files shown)


packages/dd-trace/test/appsec/rasp/ssrf.express.plugin.spec.js
```diff
@@ -162,6 +162,49 @@ describe('RASP - ssrf', () => {
...
+        withVersions('express', 'request', requestVersion => {
+          let requestToTest
+
+          beforeEach(() => {
+            requestToTest = require(`../../../../../versions/request@${requestVersion}`).get()
+          })
+
+          it('Should not detect threat', async () => {
+            app = (req, res) => {
+              requestToTest.get(`https://${req.query.host}`).on('response', () => {
```
simon-id, line 176: why is this not a threat btw ? the domain is directly coming from the user inputs no ?

packages/dd-trace/test/appsec/rasp/ssrf.express.plugin.spec.js
```diff
@@ -162,6 +162,49 @@ describe('RASP - ssrf', () => {
...
+                res.end('end')
+              })
+            }
+
+            axios.get('/?host=www.datadoghq.com')
+
+            return checkRaspExecutedAndNotThreat(agent)
+          })
+
+          it('Should detect threat doing a GET request', async () => {
```
simon-id, line 186: did you make sure this test was failing before you implemented the fix ?

iunanua: yes

image

packages/dd-trace/test/appsec/rasp/ssrf.express.plugin.spec.js
iunanua: the rasp doesn't detect the threat because we were sending an address with an incomplete uri. 
It should be `https://localhost/ifconfig.pro` but is `localhost` so it doesn't trigger the rasp rule

simon-id: oh ok ok

simon-id (dismissed):

CarlesDD (dismissed):

packages/dd-trace/src/appsec/rasp/ssrf.js
```diff
@@ -20,12 +21,12 @@ function disable () {
 function analyzeSsrf (ctx) {
   const store = storage.getStore()
   const req = store?.req
-  const url = ctx.args.uri
+  const uri = (ctx.args.options?.uri && url.format(ctx.args.options.uri)) ?? ctx.args.uri
```
CarlesDD, line 24: NIT: var is named `uri` but address is named `url`. I know it has changed due to requiring `url` module, but maybe we can name it `outgoingUrl` to match the address.

iunanua: yes, I had the same question, but as the origin value comes from a `.uri` property in both cases I decided to name it as `uri`
Anyway, I could require only `{ format } = require('url')` to avoid the naming conflict. I don't have any problem to change it.

iunanua: changed :stuck_out_tongue:

CarlesDD (approved):

PR #4792, opened 2024-10-16
Exploit prevention Shell injection

uurien:
### What does this PR do?

Modifies child_process instrumentation to add support for abortController and send file and fileArgs in separate properties, not only in command al together.
With this new extra info, call to the waf and block the operation and request if needed.

### Checklist

- [x] Unit tests.
- [x] Add capability
- [x] Add System tests: https://github.com/DataDog/system-tests/pull/3303
- [x] Add integration test to test that app is not crashing

### Additional Notes

[APPSEC-52417]

[APPSEC-52417]: https://datadoghq.atlassian.net/browse/APPSEC-52417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/datadog-instrumentations/src/child_process.js
+++ b/packages/datadog-instrumentations/src/child_process.js
@@ -75,6 +75,20 @@ function normalizeArgs (args, shell) {
   return childProcessInfo
 }
 
+function createContextFromChildProcessInfo (childProcessInfo) {
+  const context = {
+    command: childProcessInfo.command,
+    file: childProcessInfo.file,
+    shell: childProcessInfo.shell
+  }
+
+  if (childProcessInfo.fileArgs) {
+    context.fileArgs = childProcessInfo.fileArgs
+  }
+
+  return context
+}
+
 function wrapChildProcessSyncMethod (returnError, shell = false) {
   return function wrapMethod (childProcessMethod) {
     return function () {
@@ -86,14 +100,7 @@ function wrapChildProcessSyncMethod (returnError, shell = false) {
 
       const innerResource = new AsyncResource('bound-anonymous-fn')
       return innerResource.runInAsyncScope(() => {
-        const context = {
-          command: childProcessInfo.command,
-          file: childProcessInfo.file,
-          shell: childProcessInfo.shell
-        }
-        if (childProcessInfo.fileArgs) {
-          context.fileArgs = childProcessInfo.fileArgs
-        }
+        const context = createContextFromChildProcessInfo(childProcessInfo)
         const abortController = new AbortController()
 
         childProcessChannel.start.publish({ ...context, abortController })
@@ -130,14 +137,7 @@ function wrapChildProcessCustomPromisifyMethod (customPromisifyMethod, shell) {
 
     const childProcessInfo = normalizeArgs(arguments, shell)
 
-    const context = {
-      command: childProcessInfo.command,
-      file: childProcessInfo.file,
-      shell: childProcessInfo.shell
-    }
-    if (childProcessInfo.fileArgs) {
-      context.fileArgs = childProcessInfo.fileArgs
-    }
+    const context = createContextFromChildProcessInfo(childProcessInfo)
 
     const { start, end, asyncStart, asyncEnd, error } = childProcessChannel
     const abortController = new AbortController()
@@ -199,16 +199,8 @@ function wrapChildProcessAsyncMethod (ChildProcess, shell = false) {
 
       const innerResource = new AsyncResource('bound-anonymous-fn')
       return innerResource.runInAsyncScope(() => {
+        const context = createContextFromChildProcessInfo(childProcessInfo)
         const abortController = new AbortController()
-        const { command, file, shell, fileArgs } = childProcessInfo
-        const context = {
-          command,
-          file,
-          shell
-        }
-        if (fileArgs) {
-          context.fileArgs = fileArgs
-        }
 
         childProcessChannel.start.publish({ ...context, abortController })
 
--- a/packages/datadog-instrumentations/src/child_process.js
+++ b/packages/datadog-instrumentations/src/child_process.js
@@ -18,14 +18,33 @@ const names = ['child_process', 'node:child_process']
 
 // child_process and node:child_process returns the same object instance, we only want to add hooks once
 let patched = false
+
+function throwSyncError (error) {
+  throw error
+}
+
+function returnSpawnSyncError (error, context) {
+  context.result = {
+    error,
+    status: null,
+    signal: null,
+    output: null,
+    stdout: null,
+    stderr: null,
+    pid: 0
+  }
+
+  return context.result
+}
+
 names.forEach(name => {
   addHook({ name }, childProcess => {
     if (!patched) {
       patched = true
       shimmer.massWrap(childProcess, execAsyncMethods, wrapChildProcessAsyncMethod(childProcess.ChildProcess))
-      shimmer.wrap(childProcess, 'execSync', wrapChildProcessSyncMethod('execSync', true))
-      shimmer.wrap(childProcess, 'execFileSync', wrapChildProcessSyncMethod('execFileSync'))
-      shimmer.wrap(childProcess, 'spawnSync', wrapChildProcessSyncMethod('spawnSync'))
+      shimmer.wrap(childProcess, 'execSync', wrapChildProcessSyncMethod(throwSyncError, true))
+      shimmer.wrap(childProcess, 'execFileSync', wrapChildProcessSyncMethod(throwSyncError))
+      shimmer.wrap(childProcess, 'spawnSync', wrapChildProcessSyncMethod(returnSpawnSyncError))
     }
 
     return childProcess
@@ -56,7 +75,7 @@ function normalizeArgs (args, shell) {
   return childProcessInfo
 }
 
-function wrapChildProcessSyncMethod (methodName, shell = false) {
+function wrapChildProcessSyncMethod (returnError, shell = false) {
   return function wrapMethod (childProcessMethod) {
     return function () {
       if (!childProcessChannel.start.hasSubscribers || arguments.length === 0) {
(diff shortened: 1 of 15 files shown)


packages/datadog-instrumentations/src/child_process.js
```diff
@@ -84,18 +130,59 @@ function wrapChildProcessCustomPromisifyMethod (customPromisifyMethod, shell) {
...
 
-    return childProcessChannel.tracePromise(
-      customPromisifyMethod,
-      {
-        command: childProcessInfo.command,
-        shell: childProcessInfo.shell
-      },
-      this,
-      ...arguments)
+    const context = {
```
uurien, line 133: `tracePromise` does not support `abortController` to prevent call to the original function and reject a promise

packages/dd-trace/test/appsec/rasp/command_injection.express.plugin.spec.js
```diff
@@ -0,0 +1,432 @@
...
+const { getWebSpan, checkRaspExecutedAndHasThreat, checkRaspExecutedAndNotThreat } = require('./utils')
+const { assert } = require('chai')
+
+describe('RASP - command_injection', () => {
+  withVersions('express', 'express', expressVersion => {
+    let app, server, axios
+
+    async function testBlockingRequest () {
+      try {
+        await axios.get('/?dir=$(cat /etc/passwd 1>%262 ; echo .)')
```
uurien, line 17: `%26` is `&`, without encoding it axios was understanding that we had 2 params: `dir=$(cat /etc/passwd 1>` and `2 ; echo .)` (without value)

CarlesDD (commented):
I see some mesclun with naming. These changes are for detecting/blocking `shell injection` exploits, but files are named after command injection, as well as the [RASP rule type](https://github.com/DataDog/dd-trace-js/pull/4792/files#diff-0bf5127e7db5991ea4b6315349b6e85ef1cb397bb395e30cf6269dfd8c8fcbcdR15) 

Is it something we can change to keep the consistency or I am missing something?

uurien:
shell injection is a subset into command injection. For now, rule is called "command_injection", that is the reason that I used mostly command injection.
I used SHI for capability, and the tests. In the future when we support whole command injection, we will continue having only one `command_injection.js` file, and only one handler for it, but doing more things when `shell: false`. That is the reason why I decided to use command_injection for it, but the tests are only testing shell_injection, so the endpoints for the tests are SHI.

I'm not superhappy with the solution, but as you said, rule is called command-injection and the capability is called SHI... and currently it only works with shell.

packages/dd-trace/test/appsec/remote_config/index.spec.js
CarlesDD, line 1: Missing check for adding SHI capability when using default rules (line 301)
```
        expect(rc.updateCapabilities)
          .to.have.been.calledWithExactly(RemoteConfigCapabilities.ASM_RASP_SHI, true)
```

packages/dd-trace/test/appsec/remote_config/index.spec.js
```diff
@@ -340,6 +340,7 @@ describe('Remote Config index', () => {
           .to.have.been.calledWithExactly(RemoteConfigCapabilities.ASM_RASP_SQLI, true)
         expect(rc.updateCapabilities)
           .to.have.been.calledWithExactly(RemoteConfigCapabilities.ASM_RASP_LFI, true)
+          .to.have.been.calledWithExactly(RemoteConfigCapabilities.ASM_RASP_SHI, true)
```
CarlesDD, line 343: Maintain consistency by adding the `expect` and not concatenating it to the previous one.
```suggestion
         expect(rc.updateCapabilities)
          .to.have.been.calledWithExactly(RemoteConfigCapabilities.ASM_RASP_SHI, true)
```

packages/dd-trace/test/appsec/remote_config/index.spec.js
```diff
@@ -384,6 +385,7 @@ describe('Remote Config index', () => {
...
```
CarlesDD, line 388: Maintain consistency by adding the `expect` and not concatenating it to the previous one.
```suggestion
         expect(rc.updateCapabilities)
          .to.have.been.calledWithExactly(RemoteConfigCapabilities.ASM_RASP_SHI, true)
```

packages/datadog-instrumentations/src/child_process.js
```diff
@@ -63,14 +86,37 @@ function wrapChildProcessSyncMethod (shell = false) {
...
       return innerResource.runInAsyncScope(() => {
-        return childProcessChannel.traceSync(
-          childProcessMethod,
-          {
-            command: childProcessInfo.command,
-            shell: childProcessInfo.shell
-          },
-          this,
-          ...arguments)
+        const context = {
```
iunanua, line 89: couldn't you use `childProcessInfo` as `context`?

packages/datadog-instrumentations/src/child_process.js
```diff
@@ -112,9 +199,39 @@ function wrapChildProcessAsyncMethod (shell = false) {
...
+        }
+        if (fileArgs) {
+          context.fileArgs = fileArgs
+        }
+
+        childProcessChannel.start.publish({ ...context, abortController })
+
+        let childProcess
+        if (abortController.signal.aborted) {
+          childProcess = new ChildProcess()
```
iunanua, line 217: why do you need to create a new `ChildProcess` and emit `error` and `close` events? 
couldn't you just call `cb(error)`?

packages/datadog-instrumentations/src/child_process.js
uurien: Reasons why I decided not using `childProcessInfo` as context:
- In the previous version, it wasn't, it was a subset of properties of `childProcessInfo`, so I maintained it.
- I can't use directly the same object instance, (you can see that I am creating new `{ ...context, abortController }` each time, so less properties the best.

packages/datadog-instrumentations/src/child_process.js
uurien: These methods are expecting to return a ChildProcess, they fails asynchronously. We can not return an object with less properties or methods than the original, and we can't throw an Error when the expected behaviour is calling the callback.
With childProcess, customer would subscribe to `childProcess.on('error', ...)` or `childProcess.on('close')` without callback, we should support all the use cases.

packages/dd-trace/test/appsec/remote_config/index.spec.js
uurien: oh, it was not intentional,  I didn't know even that it could be joined 🤦

packages/datadog-instrumentations/src/child_process.js
iunanua: ok, I hadn't seen the `return childProcess` :S

packages/dd-trace/test/appsec/rasp/command_injection.integration.spec.js
```diff
@@ -0,0 +1,89 @@
...
+const Axios = require('axios')
+const { assert } = require('chai')
+
+// These test are here and not in the integration tests
+// because they require postgres instance
+describe('RASP - command_injection - integration', () => {
+  let axios, sandbox, cwd, appPort, appFile, agent, proc
+
+  before(async function () {
+    this.timeout(60000)
```
CarlesDD, line 15: is it necessary to wait 60 seconds in the `before` and `after`? is it not too much?

packages/dd-trace/test/appsec/rasp/command_injection.spec.js
```diff
@@ -0,0 +1,155 @@
...
+      start.publish(ctx)
+
+      sinon.assert.notCalled(waf.run)
+    })
+
+    it('should not analyze command_injection if no file', () => {
+      const ctx = {
+        fileArgs: ['arg0']
+      }
+      datadogCore.storage.getStore.returns({})
```
CarlesDD, line 134: There is no `req`, so the test cannot assure if it is not analyzed bc there is no file o there is no req.

packages/dd-trace/test/appsec/rasp/command_injection.integration.spec.js
```diff
@@ -0,0 +1,89 @@
...
+
+    sandbox = await createSandbox(
+      ['express'],
+      false,
+      [path.join(__dirname, 'resources')])
+
```
CarlesDD, line 21:
```suggestion
      [path.join(__dirname, 'resources')]
    )

```

packages/dd-trace/test/appsec/rasp/command_injection.integration.spec.js
```diff
@@ -0,0 +1,89 @@
+'use strict'
+
+const { createSandbox, FakeAgent, spawnProc } = require('../../../../../integration-tests/helpers')
+const getPort = require('get-port')
+const path = require('path')
+const Axios = require('axios')
+const { assert } = require('chai')
+
+// These test are here and not in the integration tests
+// because they require postgres instance
```
CarlesDD, line 10: It is not true for this test. It comes from the RASP SQLi integration test.

In the `integration-test` folder there are some RASP tests, so maybe this test should be moved there.

packages/dd-trace/test/appsec/rasp/command_injection.integration.spec.js
uurien: Most of the integration test has this timeout, these tests should copy app to temp dir, install dependencies with `npm install` (or yarn), start the server... I think that this is a safe number to prevent flaky tests and if this goes faster, it is not going to be waiting, it is a max timeout, not a `setTimeout` to make this long.

packages/dd-trace/test/appsec/rasp/command_injection.integration.spec.js
uurien: copy paste failure, I should remove the comment.
I think that I like the idea to have the `integration.spec.js` tests here, I think that in a separate PR we should move ssrf integration test to this directory also. To have all the rasp tests (integration and not integration) in the same directory.

packages/dd-trace/test/appsec/rasp/command_injection.spec.js
uurien: true, bad test. fixed.

CarlesDD (approved):

iunanua (approved):

PR #4804, opened 2024-10-18
remove try catch from iast plugin

IlyasShabi:
### What does this PR do?
Remove try/catch blocks from the `IastPlugin` since error handling is already managed in the parent class `Plugin`

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


CarlesDD (commented):
These changes result in a slight change in the behavior of the IAST plugin to unexpected errors.

Since errors will no longer be caught in the _wrapHandler catch, they will be caught in the [plugin catch](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/plugins/plugin.js#L80). 

This catch, besides logging the error, deactivates the plugin. This behavior is the one that diverges with respect to the one we had now. 

Is this what we want?

packages/dd-trace/src/appsec/iast/iast-plugin.js
```diff
@@ -60,24 +60,10 @@ class IastPlugin extends Plugin {
...
-
   _getTelemetryHandler (iastSub) {
     return () => {
-      try {
-        const iastContext = getIastContext(storage.getStore())
-        iastSub.increaseExecuted(iastContext)
-      } catch (e) {
-        log.error('[ASM] Error increasing handler executed metrics', e)
-      }
+      const iastContext = getIastContext(storage.getStore())
```
iunanua, line 65: If we don't catch here the error `IastPlugin` will be disabled by `Plugin`.
Because it doesn't differentiate between handlers, if a `Plugin`'s handler fails the whole `Plugin` is disabled.

I guess an error here is kind unlikely but i'd catch it without disabling the `IastPlugin`.

IlyasShabi: Yes this is exactly what @CarlesDD said, IMO:

1. Overriding Plugin's addSub to not add the failing subscription instead of disabling the whole plugin
2. we could discuss with APM about modifying `Plugin` class behavior to make it less aggressive (since this might benefit other plugins too)

iunanua: Yes but just to clarify, what I meant is that `IastPlugin` subscribes twice: one for `_wrapHandler` and another one for `_getTelemetryHandler`.
If we decide that is OK to disable the IastPlugin on an error, I'd only do for errors in `_wrapHandler` and not in `_getTelemetryHandler`.
`_getTelemetryHandler` only increments a metric so is something _optional_ for IastPlugin (that could continue running despite of an error in telemetry).

iunanua: ok ugaitz convinced me! :P let's disable the `IastPlugin` in both cases

iunanua (dismissed):

packages/dd-trace/src/appsec/iast/iast-plugin.js
IlyasShabi: @iunanua @CarlesDD @uurien Do you think it's good to disable it? If yes I will push a commit to fix linter

iunanua (approved):

uurien (approved):

PR #1227, opened 2021-01-28
Add `flush()` method to tracer

nunofgs:
### What does this PR do?
Allows users to manually flush any collected spans to the exporter agent. Very useful in scenarios where the app manually calls `process.exit()` such as the one [described by @fredefox here](https://github.com/DataDog/dd-trace-js/issues/708#issuecomment-707024807):

```
const tracer = DD.init();
const res = await App.run(tracer)
await tracer.flush();
process.exit(res);
```

### Motivation
Was implementing an app that handles SIGTERM in order to do a clean shutdown, and dd-trace did not export the span that contained the error to Datadog.

Issue #708, opened 2019-10-08
Traces not flushing on process exit


aheuermann:
**Describe the bug**
We are not seeing traces for our short-lived cron jobs we have added tracing to. These are kubernetes crons. A container is spun up specifically to run the job and then exits once complete or fails. I enabled debug and noticed that traces are being `encoded` but don't seem to be sent to the agent as I do not see any [`Request to the agent`](https://github.com/DataDog/dd-trace-js/blob/c94117d4f21093ee29981fa8a690a55362c17cdd/packages/dd-trace/src/exporters/agent/writer.js#L70) log messages. See logs below:

```
2019-10-08T21:30:12.582Z: Initializing tracer for service: jobs
2019-10-08T21:30:13.213Z: Starting Job
2019-10-08T21:30:13.364Z: Publishing messages ...
2019-10-08T21:30:13.844Z: Done publishing messages ...
2019-10-08T21:30:13.845Z: Job Complete
2019-10-08T21:30:15.950Z: "Encoding trace: [{""trace_id"":""4477388297506770617"",""span_id"":""972637913971971208"",""parent_id"":""4477388297506770617"", ...
2019-10-08T21:30:15.950Z: Adding encoded trace to buffer: ...
2019-10-08T21:30:15.950Z: Encoding trace: [{""trace_id"":""4477388297506770617"",""span_id"":""5820669086732705198"",""parent_id"":""4477388297506770617"", ...
2019-10-08T21:30:15.950Z: Adding encoded trace to buffer: ...
2019-10-08T21:30:15.950Z: Encoding trace: [{""trace_id"":""4477388297506770617"",""span_id"":""389139455014874781"",""parent_id"":""4477388297506770617"", ...
2019-10-08T21:30:15.950Z: Adding encoded trace to buffer: ...
2019-10-08T21:30:15.950Z: Encoding trace: [{""trace_id"":""4477388297506770617"",""span_id"":""1598632403633540690"",""parent_id"":""4477388297506770617"", ...
2019-10-08T21:30:15.950Z: Adding encoded trace to buffer: ...
2019-10-08T21:30:15.950Z: Encoding trace: [{""trace_id"":""4477388297506770617"",""span_id"":""6502698786863096"",""parent_id"":""4477388297506770617"", ...
2019-10-08T21:30:15.950Z: Adding encoded trace to buffer: ...
[...]

PR #4805, opened 2024-10-20
Add `flush` method to tracer

wu-json:
### What does this PR do?

Adds a `.flush()` method to the tracer, which allows users to manually flush collected spans to the exporter agent. This is useful for ensuring spans aren't lost when force exiting a process with `process.exit`.

### Motivation

My team recently migrated our Node.js crons from NewRelic to Datadog. In NewRelic land, we had access to `newrelic.shutdown({ collectPendingData: true })`, which allowed us to send all metrics and traces to NewRelic before killing the process with `process.exit` (we do not exit gracefully in our crons for reasons I will not dive into here). 

In Datadog land however, we found that such a method did not exist, which means we were not getting traces when crons completed. The only workaround is to set a delay before exiting the process, which is hacky and not guaranteed to work.

There are two existing threads with conversations requesting this feature as well.

- #708 
- #1227

The other PR that attempted to add this functionality was abandoned so I'm taking a stab at picking up where they left off.

### Checklist

- [x] Unit tests.
- [x] TypeScript [tests][2].

### Additional Notes

I really want this feature - happy to work with maintainers on making whatever edits we need to get this in.


simon-id:
FYI There is no point rebasing this PR on master anymore, as Juan recreated it to run the CI

juan-fernandez:
hey @wu-json thanks for the contribution. We've discussed this internally and while we like it, we're afraid to deviate from the public API other datadog libraries have. We're thinking of a way forward. Stay tuned 😄

wu-json:
> hey @wu-json thanks for the contribution. We've discussed this internally and while we like it, we're afraid to deviate from the public API other datadog libraries have. We're thinking of a way forward. Stay tuned 😄

@juan-fernandez Ah - understood. Let me know if I can be of use here after the direction is sorted out internally. Would be happy to help with the solution.

PR #4808, opened 2024-10-21
feat(kafkajs): add kafka cluster id to spans and dsm metrics

wconti27:
### What does this PR do?

Adds kafka cluster id to span tags and dsm metrics

### Plugin Checklist

- [x] Unit tests.
- [x] TypeScript [tests][2].
- [x] API [documentation][3].


packages/datadog-instrumentations/src/kafkajs.js
```diff
@@ -52,44 +52,48 @@ addHook({ name: 'kafkajs', file: 'src/index.js', versions: ['>=1.4'] }, (BaseKaf
...
-          return send.apply(this, arguments)
-        }
-
-        try {
-          const { topic, messages = [] } = arguments[0]
-          for (const message of messages) {
-            if (message !== null && typeof message === 'object') {
-              message.headers = message.headers || {}
-            }
+      return kafkaClusterIdPromise.then((clusterId) => {
```
bengl, line 58: If the promise has already resolved, it's unnecessary overhead to call `.then()` on it. Instead you can store the clusterId in its own variable (or just use `this._ddKafkaClusterId`), and if that variable is not set, _then_ you can call `.then()` on the promise.

packages/datadog-instrumentations/src/kafkajs.js
```diff
@@ -147,35 +155,63 @@ addHook({ name: 'kafkajs', file: 'src/index.js', versions: ['>=1.4'] }, (BaseKaf
...
+  const admin = kafka.admin()
+
+  if (!admin.describeCluster) {
+    return null
+  }
+
+  await admin.connect()
+  const clusterInfo = await admin.describeCluster()
+  const clusterId = clusterInfo?.clusterId
+  await admin.disconnect()
```
bengl, line 212: You can actually disconnect as soon as you've gotten `clusterInfo`, _and_ you don't have to `await` it, since you don't _need_ it to be disconnected to proceed.

That said, all `async/await` syntax in this file should be converted to promise chains, for performance reasons.

bengl (changes requested):

packages/datadog-instrumentations/src/kafkajs.js
wconti27: Okay, I changed the code to check if the returned data is a promise, if it is, it calls `.then()`, otherwise continues as normal without calling `.then()`

packages/datadog-instrumentations/src/kafkajs.js
wconti27: Done, converted this to use promise chains and the admin disconnects as soon as we've got `clusterInfo`

bengl (approved):

PR #4817, opened 2024-10-23
[DI] Refactor integration tests

watson:
Group Dynamic Instrumentation integration tests into separate files based on the type of target application that is needed.
    
As more and more complex tests needs to be written in the future, this will become more and more important, as the current single Fastify app quickly becomes too complicated. Without this refactor, tests needing a small change in the target app might suddenly break a lot of existing tests.

PR #4818, opened 2024-10-23
[DI] Drop snapshot if JSON payload is too large

watson:
The log track has a 1MB limit of the JSON payload. The client is not notified if the payload is too large, but it is simply never indexed.

This is a very crude approach. In the future a more sophsticated algorithm will be implemented that reduces the size of the snapshot instead of removing it completely.

### DIFF
--- a/packages/dd-trace/src/debugger/devtools_client/send.js
+++ b/packages/dd-trace/src/debugger/devtools_client/send.js
@@ -43,7 +43,11 @@ function send (message, logger, snapshot, cb) {
 
   if (Buffer.byteLength(json) > MAX_PAYLOAD_SIZE) {
     // TODO: This is a very crude way to handle large payloads. Proper pruning will be implemented later (DEBUG-2624)
-    delete payload['debugger.snapshot']
+    const line = Object.values(payload['debugger.snapshot'].captures.lines)[0]
+    line.locals = {
+      notCapturedReason: 'Snapshot was too large',
+      size: Object.keys(line.locals).length
+    }
     json = JSON.stringify(payload)
   }
 
--- a/packages/dd-trace/src/debugger/devtools_client/send.js
+++ b/packages/dd-trace/src/debugger/devtools_client/send.js
@@ -9,6 +9,8 @@ const { GIT_COMMIT_SHA, GIT_REPOSITORY_URL } = require('../../plugins/util/tags'
 
 module.exports = send
 
+const MAX_PAYLOAD_SIZE = 1024 * 1024 // 1MB
+
 const ddsource = 'dd_debugger'
 const hostname = getHostname()
 const service = config.service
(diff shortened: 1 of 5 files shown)


watson:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4818?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4818** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4818?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4817** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4817?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

packages/dd-trace/src/debugger/devtools_client/send.js
```diff
@@ -37,5 +39,13 @@ function send (message, logger, snapshot, cb) {
...
   }
 
-  request(JSON.stringify(payload), opts, cb)
+  let json = JSON.stringify(payload)
+
+  if (Buffer.byteLength(json) > MAX_PAYLOAD_SIZE) {
+    // TODO: This is a very crude way to handle large payloads. Proper pruning will be implemented later (DEBUG-2624)
+    delete payload['debugger.snapshot']
+    json = JSON.stringify(payload)
+  }
```
juan-fernandez, line 48: (nitpick) is this any thing different from this? 

```suggestion
  if (Buffer.byteLength(json) > MAX_PAYLOAD_SIZE) {
    // TODO: This is a very crude way to handle large payloads. Proper pruning will be implemented later (DEBUG-2624)
    delete payload['debugger.snapshot']
  }
  const json = JSON.stringify(payload)
```
it seems unnecessary to potentially `JSON.stringify` twice. Let me know if this makes sense 😄

juan-fernandez (dismissed):

juan-fernandez (approved):

packages/dd-trace/src/debugger/devtools_client/send.js
```diff
@@ -37,5 +39,13 @@ function send (message, logger, snapshot, cb) {
     'debugger.snapshot': snapshot
   }
 
-  request(JSON.stringify(payload), opts, cb)
+  let json = JSON.stringify(payload)
+
+  if (Buffer.byteLength(json) > MAX_PAYLOAD_SIZE) {
+    // TODO: This is a very crude way to handle large payloads. Proper pruning will be implemented later (DEBUG-2624)
+    delete payload['debugger.snapshot']
```
shatzi, line 46: kinda harsh :D 

remove 'debugger.snapshot' would cause probe-tracker to mark the probe as error: as the probe would trigger emitting start but there are not events marked with the probe-id and would default to error state.  

I would suggest remove 'debugger.snapshot.captures' or replace it with notCaptureReasons: 'snapshot was too large' - at least give the user some feedback on why the don't see the snapshot.

shatzi (dismissed):

packages/dd-trace/src/debugger/devtools_client/send.js
watson: Oh, good point. My intent was to only rmeove the `captures` part, but the JSON schema threw me for a loop since we're collecting a snapshot, the property is called `snapshot` and it's called "snapshot pruning" 🙄 I'll push a fix 👍

watson: Fixed 👍

shatzi (approved):
very cool idea to make the tests robust to line changes using `// BREAKPOINT`

PR #4826, opened 2024-10-25
[test visibility] Simple dynamic instrumentation - test visibility client 

juan-fernandez:
### What does this PR do?
This creates a `TestVisDynamicInstrumentation` class which includes a simplified version of the logic in `packages/dd-trace/src/debugger/index.js`:

* A `worker_thread` is started in `TestVisDynamicInstrumentation#start` that'll wait for messages from the parent process.
* The only available message right now is for adding a breakpoint at a given file and line. 
* If the breakpoint is hit, the worker will send the snapshot back to the parent process.

### Motivation
Continuing the work for the integration between test visibility and dynamic instrumentation started in https://github.com/DataDog/dd-trace-js/pull/4821

### Plugin Checklist

- [x] Unit tests.


packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
```diff
@@ -0,0 +1,106 @@
...
+let sessionStarted = false
+
+const scriptIds = []
+const scriptUrls = new Map()
+
+const breakpointIdToSnapshotId = new Map()
+const breakpointIdToProbe = new Map()
+
+function findScriptFromPartialPath (path) {
+  return scriptIds
```
juan-fernandez, line 17: ℹ️ this we could easily generalized for both DI and TV

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
```diff
@@ -0,0 +1,106 @@
...
+  const stack = callFrames.map((frame) => {
+    let fileName = scriptUrls.get(frame.location.scriptId)
+    if (fileName.startsWith('file://')) fileName = fileName.substr(7) // TODO: This might not be required
+    return {
+      fileName,
+      function: frame.functionName,
+      lineNumber: frame.location.lineNumber + 1, // Beware! lineNumber is zero-indexed
+      columnNumber: frame.location.columnNumber + 1 // Beware! columnNumber is zero-indexed
+    }
+  })
```
juan-fernandez, line 44: ℹ️ this can be extracted

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
```diff
@@ -0,0 +1,106 @@
+const { parentPort } = require('worker_threads')
+// TODO: move session to common place
+const session = require('../../../debugger/devtools_client/session')
```
juan-fernandez, line 3: session should be simple to move

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
```diff
@@ -0,0 +1,106 @@
...
+// TODO: move getLocalStateForCallFrame to common place
+const { getLocalStateForCallFrame } = require('../../../debugger/devtools_client/snapshot')
```
juan-fernandez, line 5: this might be a bit trickier but no reason not to do it

packages/dd-trace/src/debugger/devtools_client/state.js
```diff
@@ -32,8 +32,17 @@ module.exports = {
       .sort(([a], [b]) => a.length - b.length)[0]
   },
 
-  getScriptUrlFromId (id) {
```
juan-fernandez, line 35: `getScriptUrlFromId` was only used for formatting the stack, which is now done by `getStackFromCallFrames`

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/index.js
```diff
@@ -0,0 +1,92 @@
...
+    this.worker = null
+    this._readyPromise = new Promise(resolve => {
+      this._onReady = resolve
+    })
+  }
+
+  // Return 3 elements:
+  // 1. Snapshot ID
+  // 2. Promise that's resolved when the breakpoint is set
+  // 3. Promise that's resolved when the breakpoint is hit
```
juan-fernandez, line 23: I need to be able to sync on these two to be able to know when to start a test (a test can't start before the breakpoint is set) and when a dynamic instrumentation log can be created (if a breakpoint is hit). Maybe I could use callbacks for the same? Open to ideas here

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/index.js
```diff
@@ -0,0 +1,92 @@
...
+        probeIdToResolveBreakpointSet.set(probeId, resolve)
+      }),
+      new Promise(resolve => {
+        probeIdToResolveBreakpointHit.set(probeId, resolve)
+      })
+    ]
+  }
+
+  isReady () {
+    return this._readyPromise
```
juan-fernandez, line 45: this is just to make sure we don't attempt to set probes before the worker is online

packages/dd-trace/test/ci-visibility/dynamic-instrumentation/target-app/test-visibility-dynamic-instrumentation-script.js
```diff
@@ -0,0 +1,29 @@
+'use strict'
```
juan-fernandez, line 1: ℹ️ script we run in the test to grab local variable data

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
juan-fernandez: on second thought, `session` should probably go together with `inspector_promises_polyfill.js` and `state.js` since they seem to be self contained? I'm honestly OK with importing code directly here, as this code is evolving quickly. I don't know if it makes sense to invest on this now

juan-fernandez: thoughts @watson ?

vitor-de-araujo (approved):

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/index.js
```diff
@@ -0,0 +1,92 @@
...
+      } else if (type === BREAKPOINT_HIT) {
+        const { snapshot } = message
+        const { probe: { id: probeId } } = snapshot
+        const resolve = probeIdToResolveBreakpointHit.get(probeId)
+        if (resolve) {
+          resolve({ snapshot })
+          probeIdToResolveBreakpointHit.delete(probeId)
+        }
+      }
+    }).unref() // We also need to unref this message handler
```
watson, line 88: I assume no calling `.unref()` here gave you problems?

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
```diff
@@ -0,0 +1,88 @@
+const { parentPort } = require('worker_threads')
```
watson, line 1: No `'use strict'`?

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
watson: TL;DR: Leave `session.js` where it is for now.

This file currently acts a singleton, so it can't both be called from your code and from DI in the same worker thread. But I don't assume that's ever going to happen - just an observation.

If we were to move this to a generic shared location, there's a chance that others might use it for other things in the future. If so, this singleton behavior might throw someone for a loop. But since it's such simple code, I don't see any good reason to make this into a shared module. The `inspector_promises_polyfill.js` file on the other hand, should me moved to a central location.

So normally my conclusion to the above would be that we should each have our own version of `session.js` that just called a shared `inspector_promises_polyfill.js`. However, since the current `session.js` so used all over the place in my code, it would mean that all that would have to be refactored to allow a session object to be passed into the called functions. And I think that's too big of a change right now - hence my comment at the top about leaving it as-is for now.

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
watson: Especially if we use two different `session.js` files as I just laid out in my previous comment. If so, we need a way to pass in the current `session` object. So for now, I'd just leave this as-is

packages/dd-trace/test/ci-visibility/dynamic-instrumentation/dynamic-instrumentation.spec.js
```diff
@@ -0,0 +1,48 @@
...
+      childProcess.kill()
+    }
+  })
+
+  it('can grab local variables', (done) => {
+    // Dynamic Instrumentation - Test Visibility not currently supported for windows
+    if (process.platform === 'win32') {
+      done()
+      return
+    }
```
watson, line 24: Why not move this whole block to the root of the spec file before the first call to `describe`?

packages/dd-trace/test/ci-visibility/dynamic-instrumentation/dynamic-instrumentation.spec.js
```diff
@@ -0,0 +1,48 @@
...
+
+    childProcess.on('message', ({ snapshot: { language, stack, probe, captures }, snapshotId }) => {
+      assert.exists(snapshotId)
+      assert.exists(probe)
+      assert.exists(stack)
+      assert.equal(language, 'javascript')
+
+      assert.deepEqual(captures, {
+        lines: {
+          9: {
```
watson, line 35: Consider if you can re-use the [`getBreakpointInfo`](https://github.com/DataDog/dd-trace-js/blob/57f8a10ae8e521133ac929dc81a1c11e9ac52793/integration-tests/debugger/utils.js#L107-L124) function for you tests. That way, if you ever change the line your breakpoint is on, your tests just adapt automatically.

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
```diff
@@ -0,0 +1,88 @@
...
+let sessionStarted = false
+
+const breakpointIdToSnapshotId = new Map()
+const breakpointIdToProbe = new Map()
+
+session.on('Debugger.paused', async ({ params: { hitBreakpoints: [hitBreakpoint], callFrames } }) => {
+  const probe = breakpointIdToProbe.get(hitBreakpoint)
+  if (!probe) {
+    return session.post('Debugger.resume')
+  }
```
watson, line 23: Is this just an escape hatch in case the state is somehow messed up, is there a valid use-case for a breakpoint to exist, but no matching probe?

If this is just an escape hatch, I'd probably add a log line to log that you encountered this issue. But I'll leave that up you :)

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/index.js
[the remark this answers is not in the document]
juan-fernandez: thanks, let me try that

juan-fernandez: yeah, the test framework I used to run my tests got stuck and I had to `pkill node`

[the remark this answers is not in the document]
juan-fernandez: done in https://github.com/DataDog/dd-trace-js/pull/4826/commits/9dcc8df324999e9e489e9b6c249b39937456e9f0

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
juan-fernandez: that makes sense, thanks for the feedback

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
juan-fernandez: yeah I missed it, fixed in https://github.com/DataDog/dd-trace-js/pull/4826/commits/c92c40171e2f09cf928b2f7fd56457e0e2d9c18e

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
juan-fernandez: yeah it's a escape hatch. I encountered this situation while testing, but my code was all over the case, so I probably messed something up. Good idea to add a log line 👍

juan-fernandez: done in https://github.com/DataDog/dd-trace-js/pull/4826/commits/cd4b6fe40888043b45da080cb954447a25cd935a

packages/dd-trace/test/ci-visibility/dynamic-instrumentation/dynamic-instrumentation.spec.js
juan-fernandez: of course, that's better, thanks

juan-fernandez: done in https://github.com/DataDog/dd-trace-js/pull/4826/commits/ec347a265482e0de365a16525c6d55a0d3ee05b4

packages/dd-trace/test/ci-visibility/dynamic-instrumentation/dynamic-instrumentation.spec.js
juan-fernandez: the function needs some adjusting, as it's currently tailored for debugger tests (for example by the usage of `debugger/target-app/` and `__dirname`. The hardcoding of lines is no big issues right now, but I'll adapt the function once it becomes an issue 👍

vitor-de-araujo (approved):

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/index.js
watson: Odd... I never had this problem 🤔 @tlhunter @bengl Does this ring a bell?

PR #4827, opened 2024-10-25
Template injection vulnerability detection in handlebars and pug

IlyasShabi:
### What does this PR do?
Add support to server Template injection using handlebars and pug

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/dd-trace/test/appsec/iast/analyzers/template-injection-analyzer.handlebars.plugin.spec.js
```diff
@@ -0,0 +1,71 @@
...
+                    {{#with (string.sub.apply 0 codelist)}}
+                      {{this}}
+                    {{/with}}
+                  {{/each}}
+                {{/with}}
+              {{/with}}
+            {{/with}}
+          {{/with}}
+          `
+      goodSource = '<p>{{name}}</p>'
```
uurien, line 34: For vulnerability detection, the difference about the string is not its content, it is the source. You can use this `<p>{{name}</p>` for vulnerable and not vulnerable tests. It should be tainted in the vulnerable tests and not tainted in not vulnerable tests.

packages/datadog-instrumentations/src/handlebars.js
```diff
@@ -0,0 +1,22 @@
+'use strict'
+
+const shimmer = require('../../datadog-shimmer')
+const { channel, addHook } = require('./helpers/instrument')
+
+const handlebarsReadCh = channel('datadog:handlebars:compile:start')
+
+function wrapCompile (compile) {
+  return function () {
```
simon-id, line 9: fyi in a lot of places in the instrumentation the wrapped function is named, just for clarity like so:
```suggestion
  return function wrappedCompile () {
```

packages/datadog-instrumentations/src/handlebars.js
```diff
@@ -0,0 +1,22 @@
...
+    if (handlebarsReadCh.hasSubscribers) {
+      const source = arguments[0]
```
simon-id, line 11: i usually just use the named function argument, not sure why you're using the arguments var here ?

packages/datadog-instrumentations/src/pug.js
```diff
@@ -0,0 +1,22 @@
...
+    }
+    return compile.apply(this, arguments)
+  }
+}
+
+addHook({ name: 'pug', versions: ['>=2.0.4'] }, compiler => {
+  shimmer.wrap(compiler, 'compile', wrapCompile)
+  shimmer.wrap(compiler, 'compileClient', wrapCompile)
+  shimmer.wrap(compiler, 'compileClientWithDependenciesTracked', wrapCompile)
+  return compiler
```
uurien, line 21:
```suggestion
  shimmer.wrap(compiler, 'compileClientWithDependenciesTracked', wrapCompile)

  return compiler
```

packages/dd-trace/test/appsec/iast/analyzers/template-injection-analyzer.pug.plugin.spec.js
```diff
@@ -0,0 +1,66 @@
...
+        })
+    })
+
+    describe('compileClient', () => {
+      prepareTestServerForIast('template injection analyzer',
+        (testThatRequestHasVulnerability, testThatRequestHasNoVulnerability) => {
+          testThatRequestHasVulnerability(() => {
+            const store = storage.getStore()
+            const iastContext = iastContextFunctions.getIastContext(store)
+            const command = newTaintedString(iastContext, source, 'param', 'Request')
```
uurien, line 40: `command`? Looks like copy&pasted var name, isn't it really `template`?

packages/dd-trace/src/appsec/iast/vulnerabilities-formatter/evidence-redaction/sensitive-analyzers/injection-sensitive-analyzer.js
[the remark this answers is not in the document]
IlyasShabi: I did change the name to `long-content-sensitive-analyzer` but Im open to any other suggestion

packages/dd-trace/src/appsec/iast/vulnerabilities-formatter/evidence-redaction/sensitive-handler.js
```diff
@@ -5,13 +5,13 @@ const vulnerabilities = require('../../vulnerabilities')
...
 const { contains, intersects, remove } = require('./range-utils')
 
-const codeInjectionSensitiveAnalyzer = require('./sensitive-analyzers/code-injection-sensitive-analyzer')
 const commandSensitiveAnalyzer = require('./sensitive-analyzers/command-sensitive-analyzer')
 const hardcodedPasswordAnalyzer = require('./sensitive-analyzers/hardcoded-password-analyzer')
 const headerSensitiveAnalyzer = require('./sensitive-analyzers/header-sensitive-analyzer')
 const jsonSensitiveAnalyzer = require('./sensitive-analyzers/json-sensitive-analyzer')
 const ldapSensitiveAnalyzer = require('./sensitive-analyzers/ldap-sensitive-analyzer')
 const sqlSensitiveAnalyzer = require('./sensitive-analyzers/sql-sensitive-analyzer')
+const injectionSensitiveAnalyzer = require('./sensitive-analyzers/tainted-range-based-sensitive-analyzer')
```
uurien, line 14: As you have renamed the file, you should rename also the var name

.github/workflows/appsec.yml
```diff
@@ -122,7 +122,7 @@ jobs:
   express:
     runs-on: ubuntu-latest
     env:
-      PLUGINS: express|body-parser|cookie-parser|multer
+      PLUGINS: express|body-parser|cookie-parser|multer|handlebars|pug
```
CarlesDD, line 125: I think we should not place `handlebars` and `pug` plugin tests in the `express` job.

These tests are not related w/ express, since they are not testing the express integration (they don't do `app.set('view engine', 'pug')` or `app.set('view engine', 'hbs')`). A separate job for these plugin test would be more suitable.

WDYT?

CarlesDD:
PR title should be changed to mention both `handlebars` and `pug`.

packages/datadog-instrumentations/src/pug.js
```diff
@@ -0,0 +1,24 @@
+'use strict'
+
+const shimmer = require('../../datadog-shimmer')
+const { channel, addHook } = require('./helpers/instrument')
+
+const pugReadCh = channel('datadog:pug:compile:start')
```
CarlesDD, line 6: **Nit**: Var name could be improved by correctly indicating the operation to be published in the DC. In this case, as it will publish a message every time pug compiles, it could be named `pugCompileCh`.

packages/datadog-instrumentations/src/handlebars.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
+
+const shimmer = require('../../datadog-shimmer')
+const { channel, addHook } = require('./helpers/instrument')
+
+const handlebarsReadCh = channel('datadog:handlebars:compile:start')
```
CarlesDD, line 6: **Nit**: Var name could be improved by correctly indicating the operation to be published in the DC. In this case, as it will publish a message every time handlebars compiles, it could be named `handlebarsCompileCh`.

packages/datadog-instrumentations/src/pug.js
```diff
@@ -0,0 +1,24 @@
...
+      pugReadCh.publish({ source })
+    }
+
+    return compile.apply(this, arguments)
+  }
+}
+
+addHook({ name: 'pug', versions: ['>=2.0.4'] }, compiler => {
+  shimmer.wrap(compiler, 'compile', wrapCompile)
+  shimmer.wrap(compiler, 'compileClient', wrapCompile)
```
CarlesDD, line 20: Since `compileClient` [calls internally](https://github.com/pugjs/pug/blob/master/packages/pug/lib/index.js#L363) `compileClientWithDependenciesTracked`, and the latter is also instrumented, are we analyzing twice for SSTi when `compileClient` is called?

packages/dd-trace/test/appsec/iast/analyzers/template-injection-analyzer.handlebars.plugin.spec.js
```diff
@@ -0,0 +1,49 @@
...
+          testThatRequestHasVulnerability(() => {
+            const store = storage.getStore()
+            const iastContext = iastContextFunctions.getIastContext(store)
+            const template = newTaintedString(iastContext, source, 'param', 'Request')
+            lib.compile(template)
+          }, 'TEMPLATE_INJECTION')
+
+          testThatRequestHasNoVulnerability(() => {
+            const template = lib.compile(source)
+            template()
```
CarlesDD, line 28: `template()` is invoked in this test (no vuln), but is not invoked in the previous (when checking there is a vuln). Is there any reason for that?

packages/dd-trace/test/appsec/iast/analyzers/template-injection-analyzer.pug.plugin.spec.js
```diff
@@ -0,0 +1,65 @@
...
+          testThatRequestHasVulnerability(() => {
+            const store = storage.getStore()
+            const iastContext = iastContextFunctions.getIastContext(store)
+            const template = newTaintedString(iastContext, source, 'param', 'Request')
+            lib.compile(template)
+          }, 'TEMPLATE_INJECTION')
+
+          testThatRequestHasNoVulnerability(() => {
+            const template = lib.compile(source)
+            template()
```
CarlesDD, line 28: `template()` is invoked in this test (no vuln), but is not invoked in the previous (when checking there is a vuln). Is there any reason for that?

IlyasShabi: No reason for this just forgot to remove it

packages/datadog-instrumentations/src/pug.js
IlyasShabi: Done

packages/datadog-instrumentations/src/handlebars.js
IlyasShabi: Done

packages/datadog-instrumentations/src/pug.js
IlyasShabi: Done

.github/workflows/appsec.yml
IlyasShabi: I previously created separate jobs for `handlebars` and `pug`, but I encountered an error related to express, the only solution I found was to run them with express.
I tried something like `PLUGINS: express|handlebars|pug`  but that ends up running all the express tests again.

I've just pushed my previous commit, and you can see that the `AppSec / template` job is failing.

CarlesDD: I think the error is not related with express.

In the plugin test, the lib to be tested (hbs, pug) are [required](https://github.com/DataDog/dd-trace-js/pull/4827/files#diff-45d723a65b1a18f2cc2513af6b2acfebad56c4aa9c626be1cf38d22cb4f4925aR12) in the before fn, before calling the method `prepareTestServerForIast`. This method initializes everything necessary to be able to test the plugin, among other things, it initializes the tracer itself, which registers the hook of the require of the library, where we instrument these.

What is happening is that a require of the library is being done before the initialization of the tracer, and therefore, the first test fails because that library is not instrumented (you can see that the second test for the same library but a different version does pass).

The error would be solved by making the library require inside a `beforeEach` inside the `prepareTestServerForIast` cb.

CarlesDD: Like in this case: https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/test/appsec/iast/analyzers/sql-injection-analyzer.sequelize.plugin.spec.js#L27

IlyasShabi: Thanks for this clear explanation, Im gonna fix it

CarlesDD:
In Handlebars there is one more vulnerable point that this PR is not covering: `registerPartial`.

Let's take the following code as an example:

```javascript
  Handlebars.registerPartial('vulnerablePartial', req.query.partial).
  const templateSource = `{{> vulnerablePartial }}`
  const template = Handlebars.compile(templateSource)
```

This code is vulnerable to `SSTi`, but the tests I have run do not detect the vulnerability (and it is logical because `registerPartial` never calls `compile` or `precompile`).

packages/dd-trace/test/appsec/iast/analyzers/template-injection-analyzer.pug.plugin.spec.js
```diff
@@ -0,0 +1,79 @@
+'use strict'
+
+const { prepareTestServerForIast } = require('../utils')
+const { storage } = require('../../../../../datadog-core')
+const iastContextFunctions = require('../../../../src/appsec/iast/iast-context')
+const { newTaintedString } = require('../../../../src/appsec/iast/taint-tracking/operations')
+
+describe('template-injection-analyzer with pug', () => {
```
CarlesDD, line 8: In addition to these already tested methods, we also detected vulnerability when using the `render` method (which internally calls `handleTemplateCache` which in turn calls `compile`).

I think it would be worth adding a test with `render` to check that it still works in future versions, since we are not implementing it implicitly.

IlyasShabi: Done

packages/datadog-instrumentations/src/handlebars.js
```diff
@@ -0,0 +1,40 @@
...
+    if (handlebarsCompileCh.hasSubscribers) {
+      handlebarsCompileCh.publish({ source })
+    }
+
+    return compile.apply(this, arguments)
+  }
+}
+
+function wrapRegisterPartial (handlebarsEnvironment) {
+  return function wrappedCompile (name, partial) {
```
CarlesDD, line 20:
```suggestion
  return function wrappedRegisterPartial (name, partial) {
```

CarlesDD (approved):

uurien (approved):

packages/datadog-instrumentations/src/handlebars.js
```diff
@@ -0,0 +1,40 @@
...
+  return function wrappedCompile (source) {
+    if (handlebarsCompileCh.hasSubscribers) {
+      handlebarsCompileCh.publish({ source })
+    }
+
+    return compile.apply(this, arguments)
+  }
+}
+
+function wrapRegisterPartial (handlebarsEnvironment) {
```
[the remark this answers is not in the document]
IlyasShabi: Done

CarlesDD (approved):

PR #4830, opened 2024-10-28
create isNoop method on scope

bengl:
### What does this PR do?
Adds an isNoop() method on the Scope clas.

### Motivation

This enables `wrap()` and `trace()` to be written solely in terms of methods already available on publicly accessible classes.

### Additional Notes

Not documenting it _or_ testing it in isolation for now, since it's there just to enable internal things.

### DIFF
--- a/packages/dd-trace/src/scope.js
+++ b/packages/dd-trace/src/scope.js
@@ -58,6 +58,10 @@ class Scope {
   _isPromise (promise) {
     return promise && typeof promise.then === 'function'
   }
+
+  isNoop () {
+    return storage.getStore()?.noop
+  }
 }
 
 module.exports = Scope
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/scope.js
```diff
@@ -58,6 +58,10 @@ class Scope {
   _isPromise (promise) {
     return promise && typeof promise.then === 'function'
   }
+
+  isNoop () {
```
rochdev, line 62: A `noop` scope is an internal concept, but this exposes it publicly. If anything this should be moved out of scope completely and moved to its own store.

rochdev (changes requested):

bengl:
This is unnecessary. Closing.

PR #4835, opened 2024-10-29
eventNames net test

IlyasShabi:
### What does this PR do?
Remove the fix introduced in PR #3906, as it’s already addressed in Node.js by nodejs/node#51331

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/datadog-plugin-net/test/index.spec.js
+++ b/packages/datadog-plugin-net/test/index.spec.js
@@ -228,14 +228,8 @@ describe('Plugin', () => {
           socket.destroy()
 
           socket.once('close', () => {
-            setImmediate(() => {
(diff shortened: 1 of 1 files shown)


simon-id:
so if we try running the tests in node 21.2 it will break :upside_down_face: sometimes it's useful to run those tests in specific versions of node without having weird noise
but honestly no strong opinion

rochdev:
> so if we try running the tests in node 21.2 it will break

Odd versions of Node are not supported unless they are the current release (for example Node 23 is supported right now, but it will no longer be supported as soon as Node 24 is released)

simon-id (approved):

PR #4836, opened 2024-10-29
Support url.parse, url.URL.parse and new url.URL for IAST taint tracking

uurien:
### What does this PR do?

Add support to taint tracking string in iast when they are parsed URLs.

### Motivation

Improve iast vulnerabilities detection, especially SSRF

### Checklist

- [x] Unit tests.


packages/datadog-instrumentations/src/url.js
```diff
@@ -0,0 +1,85 @@
...
+  shimmer.wrap(url, 'URL', (URL) => {
+    return class extends URL {
+      constructor () {
+        super(...arguments)
+
+        if (!parseFinishedChannel.hasSubscribers) return
+
+        parseFinishedChannel.publish({
+          input: arguments[0],
+          base: arguments[1],
```
simon-id, line 58: I put the same comment in Ilyas' PR but why using `arguments[x]` instead of just named arguments ?

.github/workflows/plugins.yml
```diff
@@ -484,6 +484,17 @@ jobs:
...
+    strategy:
+      matrix:
+        node-version: ['18', '20', 'latest']
+    runs-on: ubuntu-latest
+    env:
+      PLUGINS: url
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/plugins/test
+
```
CarlesDD, line 497: Could it be placed further down in the file to maintain alphabetical order?

.github/workflows/plugins.yml
```diff
@@ -484,6 +484,17 @@ jobs:
         uses: ./.github/actions/testagent/logs
       - uses: codecov/codecov-action@v3
 
+  url:
+    strategy:
+      matrix:
+        node-version: ['18', '20', 'latest']
```
CarlesDD, line 490: Not sure it is working as expected. All plugin jobs for url - [url(18)](https://github.com/DataDog/dd-trace-js/actions/runs/11580977176/job/32240746741?pr=4836#step:3:304), [url(20)](https://github.com/DataDog/dd-trace-js/actions/runs/11580977176/job/32240747104?pr=4836#step:3:306) and [url(latest)](https://github.com/DataDog/dd-trace-js/actions/runs/11580977176/job/32240747522?pr=4836#step:3:302) - are running w/ node v18

uurien: not, this `matrix` is not necessary, as at the end I am using `./.github/actions/plugins/test` action, I don't need this matrix, internally it is already running in each node version. I'll replace it.

packages/datadog-instrumentations/src/url.js
uurien: I think that I am using it because I am already using `arguments` to call to `apply(this, arguments)` or `super(...arguments)`. But yes, it is cleaner if I use `input` and `base` variables.

packages/datadog-instrumentations/src/url.js
```diff
@@ -0,0 +1,84 @@
...
+
+const parseFinishedChannel = channel('datadog:url:parse:finish')
+const urlGetterChannel = channel('datadog:url:getter:finish')
+const instrumentedGetters = ['host', 'origin', 'hostname']
+
+addHook({ name: names }, function (url) {
+  shimmer.wrap(url, 'parse', (parse) => {
+    return function wrappedParse (input) {
+      const parsedValue = parse.apply(this, arguments)
+      if (!parseFinishedChannel.hasSubscribers) return parsedValue
```
juan-fernandez, line 15: will `parseFinishedChannel` have listeners if IAST is not enabled? I'm trying to think whether the url instrumentation should be enabled in Test Visibility or not

uurien: > will parseFinishedChannel have listeners if IAST is not enabled?

I don't think so, I dunno in the future, but for now it is going to be used only to be able to track the string when it converted to URL object. It doesn't look like an object that tracer will want to track.

CarlesDD (commented):
Note that the following case will not be correctly tainted:

```
const { URL } = require('url')
const baseURL = URL.parse(req.body.url)
const url = URL.parse('/path', baseURL)
```

This could be addressed in a future PR

CarlesDD (commented):
If host name of input contains Unicode characters, range's `parameterValue` of tainted value will be the ASCII converted value, not the original one.

`http://exаmple.com/` will be converted to `xn--exmple-4nf.com`. The latter is the value reported as valuePart in the evidence of a potential detected vulnerability.

CarlesDD (approved):

benconnito:
this commit seems to have broken mocking `got` with `nock` https://github.com/DataDog/dd-trace-js/issues/4896

PR #4841, opened 2024-10-30
Add exclusions for header injection vulnerability

CarlesDD:
### What does this PR do?
Adds new exclusions for the header injection vulnerability in order to reduce false positives:

- `Transfer-Encoding/Content-Encoding`: when sources come from `Accept-Encoding`
- `Pragma`: when sources come from `Cache-Control`

### Motivation
Reduce false positives in header injection vulnerability detections

### Additional Notes
Previously, the `set-cookie` header exclusion checked that the source of the range was both the value of a cookie and the name of a cookie. Since [cookie names are not being tainted,](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js#L105) this check has been removed.

`vary` header should  be ignored if the tainted portion is composed only from request header names, but since [header names are not being tainted](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js#L110), this check has not been implemented.

System Test PR: https://github.com/DataDog/system-tests/pull/3408

[APPSEC-55563](https://datadoghq.atlassian.net/browse/APPSEC-55563)

[APPSEC-55563]: https://datadoghq.atlassian.net/browse/APPSEC-55563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/src/appsec/iast/analyzers/header-injection-analyzer.js
+++ b/packages/dd-trace/src/appsec/iast/analyzers/header-injection-analyzer.js
@@ -6,7 +6,6 @@ const { getNodeModulesPaths } = require('../path-line')
 const { HEADER_NAME_VALUE_SEPARATOR } = require('../vulnerabilities-formatter/constants')
 const { getRanges } = require('../taint-tracking/operations')
 const {
-  HTTP_REQUEST_COOKIE_NAME,
   HTTP_REQUEST_COOKIE_VALUE,
   HTTP_REQUEST_HEADER_VALUE
 } = require('../taint-tracking/source-types')
@@ -45,13 +44,7 @@ class HeaderInjectionAnalyzer extends InjectionAnalyzer {
     if (this.isExcludedHeaderName(lowerCasedHeaderName) || typeof value !== 'string') return
 
     const ranges = getRanges(iastContext, value)
-    if (ranges?.length > 0) {
-      return !(this.isCookieExclusion(lowerCasedHeaderName, ranges) ||
-        this.isSameHeaderExclusion(lowerCasedHeaderName, ranges) ||
-        this.isAccessControlAllowExclusion(lowerCasedHeaderName, ranges))
-    }
-
-    return false
+    return ranges?.length > 0 && !this.shouldIgnoreHeader(lowerCasedHeaderName, ranges)
   }
 
   _getEvidence (headerInfo, iastContext) {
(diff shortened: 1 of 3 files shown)


packages/dd-trace/src/appsec/iast/analyzers/header-injection-analyzer.js
```diff
@@ -75,28 +68,52 @@ class HeaderInjectionAnalyzer extends InjectionAnalyzer {
...
+  /** Exclude when the header is reflected from the request */
   isSameHeaderExclusion (name, ranges) {
     return ranges.length === 1 && name === ranges[0].iinfo.parameterName?.toLowerCase()
   }
 
+  shouldIgnoreHeader (headerName, ranges) {
+    switch (headerName) {
+      case 'set-cookie':
+        /** Exclude set-cookie header if the source of all the tainted ranges are cookies */
+        return this.isAllRangesFromSource(ranges, HTTP_REQUEST_COOKIE_VALUE)
```
CarlesDD, line 104: Cookies names are not being tainted, so no need to check ranges comes from this source.

packages/dd-trace/test/appsec/iast/analyzers/header-injection.express.plugin.spec.js
```diff
@@ -270,6 +270,128 @@ describe('Header injection vulnerability', () => {
...
+                }
+              }).catch(done)
+            }
+          })
+
+          testThatRequestHasNoVulnerability({
+            testDescription: 'should not have HEADER_INJECTION vulnerability ' +
+              `when the header is "${headerName}" and the origin is a substring of accept-encoding header`,
+            fn: (req, res) => {
+              setHeaderFunction(headerName, req.headers['accept-encoding'].split(',').pop(), res)
```
uurien, line 383: I think that you are not testing what you want to test here.
In these type of tests, the rewriter is not modifying the code that is going to be executed, so in `split(',')` you will loose the tainted string.
Maybe, if you export this controller function to other file, move that file to tmp dir and require the new file in the test, it will work as expected.

We are doing it in some tests they need a rewriter, for example: https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/test/appsec/iast/analyzers/hardcoded-password-analyzer.spec.js#L108

CarlesDD: Yes, you are right.

CarlesDD: Fixed

uurien (approved):

PR #4845, opened 2024-10-30
upgrade to latest @azure/functions version

khanayan123:
### What does this PR do?
upgrades to @azure/functions version 4.6.0

### Motivation
npm audit failing in versions < 4.6.0 since those version have cookie 0.6.0 as a dependency


wconti27 (dismissed):

duncanpharvey (changes requested):
> registry.npmjs.org

Can you use `yarn` to install the dependencies instead of `npm install`? It should result in fewer changes to the lockfile

khanayan123:
> > registry.npmjs.org
> 
> Can you use `yarn` to install the dependencies instead of `npm install`? It should result in fewer changes to the lockfile

Done!

wconti27 (dismissed):

duncanpharvey (changes requested):
> > > registry.npmjs.org
> > 
> > 
> > Can you use `yarn` to install the dependencies instead of `npm install`? It should result in fewer changes to the lockfile
> 
> Done!

The registry in use is still `registry.npmjs.org`. Can you revert to the original lockfile then rerun `yarn`? Only `@azure/functions` and `cookie` should be changing.

duncanpharvey (approved):
LGTM!

PR #4853, opened 2024-11-01
add release proposal script for use locally

rochdev:
### What does this PR do?

Add release proposal script for use locally.

### Motivation

While `branch-diff` is great in functionality, it's very verbose to use as it requires a lot of options to be passed in that would never change for a single repo. This script automates that along with other steps that have to be done locally, and attempts to do so in a way that the script can be rerun multiple times until the proposal is done.

### DIFF
--- a/scripts/release/proposal.js
+++ b/scripts/release/proposal.js
@@ -46,7 +46,7 @@ const [lastMajor, lastMinor, lastPatch] = require('../../package.json').version.
 const lineDiff = capture(`${diffCmd} v${releaseLine}.x master`)
 const newVersion = lineDiff.includes('SEMVER-MINOR')
   ? `${releaseLine}.${lastMinor + 1}.${lastPatch}`
-  : `${releaseLine}.${lastMinor + 1}.${lastPatch + 1}`
+  : `${releaseLine}.${lastMinor}.${lastPatch + 1}`
 
 // Checkout new branch and output new changes.
 run(`git checkout v${newVersion}-proposal || git checkout -b v${newVersion}-proposal`)
--- a/scripts/release/proposal.js
+++ b/scripts/release/proposal.js
@@ -1,0 +1,128 @@
+'use strict'
+
+/* eslint-disable no-console */
+
+// TODO: Support major versions.
+
+const { execSync } = require('child_process')
+const fs = require('fs')
+const path = require('path')
+
+// Helpers for colored output.
+const log = msg => console.log(msg)
+const success = msg => console.log(`\x1b[32m${msg}\x1b[0m`)
+const error = msg => console.log(`\x1b[31m${msg}\x1b[0m`)
+const whisper = msg => console.log(`\x1b[90m${msg}\x1b[0m`)
+
+const currentBranch = capture('git rev-parse --abbrev-ref HEAD')
+const releaseLine = process.argv[2]
+
+// Validate release line argument.
+if (!releaseLine || releaseLine === 'help' || releaseLine === '--help') {
+  log('Usage: node scripts/release/prepare <release-line> [release-type]')
+  process.exit(0)
+} else if (!releaseLine?.match(/^\d+$/)) {
+  error('Invalid release line. Must be a whole number.')
+  process.exit(1)
+}
+
+// Make sure the release branch is up to date to prepare for new proposal.
+// The main branch is not automatically pulled to avoid inconsistencies between
+// release lines if new commits are added to it during a release.
+run(`git checkout v${releaseLine}.x`)
+run('git pull')
+
+const diffCmd = [
+  'branch-diff',
+  '--user DataDog',
+  '--repo dd-trace-js',
+  isActivePatch()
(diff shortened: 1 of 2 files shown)


scripts/release/proposal.js
```diff
@@ -0,0 +1,128 @@
...
+const success = msg => console.log(`\x1b[32m${msg}\x1b[0m`)
+const error = msg => console.log(`\x1b[31m${msg}\x1b[0m`)
+const whisper = msg => console.log(`\x1b[90m${msg}\x1b[0m`)
+
+const currentBranch = capture('git rev-parse --abbrev-ref HEAD')
+const releaseLine = process.argv[2]
+
+// Validate release line argument.
+if (!releaseLine || releaseLine === 'help' || releaseLine === '--help') {
+  log('Usage: node scripts/release/prepare <release-line> [release-type]')
```
bengl, line 22:
```suggestion
  log('Usage: node scripts/release/proposal <release-line> [release-type]')
```

scripts/release/proposal.js
```diff
@@ -0,0 +1,128 @@
...
+const fs = require('fs')
+const path = require('path')
+
+// Helpers for colored output.
+const log = msg => console.log(msg)
+const success = msg => console.log(`\x1b[32m${msg}\x1b[0m`)
+const error = msg => console.log(`\x1b[31m${msg}\x1b[0m`)
+const whisper = msg => console.log(`\x1b[90m${msg}\x1b[0m`)
+
+const currentBranch = capture('git rev-parse --abbrev-ref HEAD')
```
bengl, line 17:
```suggestion
const currentBranch = capture('git branch --show-current')
```

scripts/release/proposal.js
```diff
@@ -0,0 +1,128 @@
...
+const releaseLine = process.argv[2]
+
+// Validate release line argument.
+if (!releaseLine || releaseLine === 'help' || releaseLine === '--help') {
+  log('Usage: node scripts/release/prepare <release-line> [release-type]')
+  process.exit(0)
```
bengl, line 23:
```suggestion
  process.exit(1)
```

Since this is also not a successful exit

scripts/release/proposal.js
```diff
@@ -0,0 +1,128 @@
...
+    run(`echo "${proposalDiff}" | xargs git cherry-pick`)
+  } catch (err) {
+    error('Cherry-pick failed. Resolve the conflicts and run `git cherry-pick --continue` to continue.')
+    error('When all conflicts have been resolved, run this script again.')
+    process.exit(1)
+  }
+}
+
+// Update package.json with new version.
+run(`yarn version --no-git-tag-version --new-version ${newVersion}`)
```
bengl, line 79:
```suggestion
run(`npm version --git-tag-version=false x${newVersion}`)
```

scripts/release/proposal.js
```diff
@@ -0,0 +1,128 @@
...
+function ready () {
+  const notesDir = path.join(__dirname, '..', '..', '.github', 'notes')
+  const notesFile = path.join(notesDir, `${newVersion}.md`)
+  const lineDiff = capture(`${diffCmd} --markdown=true v${releaseLine}.x master`)
+
+  fs.mkdirSync(notesDir, { recursive: true })
+  fs.writeFileSync(notesFile, lineDiff)
+
+  success('Release proposal is ready.')
+  success(`Changelog at .github/notes/${newVersion}.md`)
```
bengl, line 125: I'm not a huge fan of leaving files around when we can avoid it.

My suggestion here would be to use [`git notes`](https://git-scm.com/docs/git-notes). While they don't show up in the UI (and haven't in 10 years! (_shakes fist at GitHub_)), They're a handy place to stash information that's preserved in git, and associated with a commit, so we can use that to store the release notes and if we want to automate even further, have a GitHub Action that copies the data from `git notes` into the PR description.

Alternatively... just log it instead of writing it to a file.

bengl (changes requested):

scripts/release/proposal.js
rochdev: It's only not successful in the first case but not the other 2.

scripts/release/proposal.js
rochdev: Logging it means it could be lost, and once the release is done it becomes more complicated to generate it as branch-diff would no longer find the same commits. I don't see a problem with leftover files in a gitignored folder to be honest. A lot of tools leave files around and you can just delete the folder if you want them removed. Using git notes makes this more akward and you need to check how to use it to be able to extract the release notes from it. Opening a file and copy pasting its content is a lot easier. Eventually we could also use the files as an actual changelog, or do this automatically, but for now it seemed like the easiest to use solution.

scripts/release/proposal.js
rochdev: I'd also argue that it's successful for the first case as well if the default command is `help`.

bengl: Eh, that's fine.

bengl (approved):

PR #4854, opened 2024-11-04
fix incompatibilities with node 16 for 4.49.0 release

rochdev:
### What does this PR do?

Fix incompatibilities with node 16 for 4.49.0 release. This means:

- Specifying TypeScript types explicitly to avoid TypeScript trying to load types that are incompatible with TypeScript 3.x which were added with the `@apollo/server` dependency.
- Skip `multer` tests as they use FormData which wasn't available in Node 16.

### Motivation

These issues are blocking the 4.49.0 release proposal.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -4,7 +4,7 @@
   "main": "typedoc.js",
   "scripts": {
     "build": "typedoc ../index.d.ts && ./add-redirects.sh",
-    "pretest": "tsc -p . && tsc test",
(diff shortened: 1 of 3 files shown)


uurien (approved):

iunanua:
I can't find a way to make a multipart request without adding a dependency and don't think it's worth it because tracer's node 16 support ends in one month...

iunanua (approved):

PR #4856, opened 2024-11-04
log template messages and errors

iunanua:
### What does this PR do?
Update `log` methods (`error` and `warn`, `info`, `debug`) to accept multiple arguments in order to use template messages.
[util.format](https://nodejs.org/api/util.html#utilformatformat-args) is used to format the templates

``` javascript
log.error(err)
log.error('template message %s %d', arg1, arg2)
log.error('template %s', arg1, err)
```

### Notes
Legacy cases are still supported
```javascript
log.error() // no message, no error
log.error({ message: 'error' }) // object with a message property
log.error({ stack: 'stack' }) // non Error object with a stack property
```
There is a case, used often with `log.debug()`, where a callback or delegate is passed to the logger: `log.debug(() => do_something_heavy())`. 
The delegate is executed by `log/writer.js` but not by `telemetry/log`.
So when executing `log.error(() => error)` the error will be written to the configured logger but won't be sent via telemetry.

### Motivation
- meet the 'no dynamic messages' telemetry requirement
- avoid calling twice to log an error and a message

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/dd-trace/src/log/log.js
```diff
@@ -0,0 +1,52 @@
...
+      formatted = format(message, ...args)
+    }
+    return formatted
+  }
+
+  static parse (...args) {
+    let message, cause, delegate
+
+    const lastArg = args[args.length - 1]
+    if (lastArg && typeof lastArg === 'object' && lastArg.stack) { // lastArg instanceof Error?
```
iunanua, line 27: could it check directly if `lastArg` is an instance of `Error`? or should we continue supporting an object with `stack` property?

packages/dd-trace/src/log/log.js
```diff
@@ -0,0 +1,52 @@
...
+      cause = args.pop()
+    }
+
+    const firstArg = args.shift()
+    if (firstArg) {
+      if (typeof firstArg === 'string') {
+        message = firstArg
+      } else if (typeof firstArg === 'object') {
+        message = String(firstArg.message || firstArg)
```
iunanua, line 36: is this case still valid? an object with message property...

packages/dd-trace/src/telemetry/logs/index.js
```diff
@@ -35,18 +35,21 @@ function onLog (log) {
...
-      message: msg
-    })
+  const { message, cause } = msg
+  if (!message && !cause) return
+
+  const telLog = {
+    level: 'ERROR',
+
+    // existing log.error(err) without message will be reported as 'Generic Error'
+    message: message || 'Generic Error'
```
iunanua, line 45: We are assigning the 'Generic Error' message to all errors logged without message. 
It allows us to not to update all the `log.error(err)` cases in the tracer but will be recommended to provide a message for each of the cases

packages/dd-trace/src/log/log.js
szegedi, line 36: Could we use `??` here instead? I believe you'd want `.message` to be nullish instead of falsey to fall back to the whole object.

packages/dd-trace/test/telemetry/logs/log-collector.spec.js
```diff
@@ -63,7 +63,7 @@ describe('telemetry log collector', () => {
       })).to.be.true
     })
 
-    it('should not include original message if first frame is not a dd frame', () => {
+    it('should include original message if first frame is not a dd frame', () => {
```
szegedi, line 66: Why is it okay that this got reversed?

packages/dd-trace/src/log/writer.js
```diff
@@ -51,32 +52,62 @@ function reset () {
   toggleSubscription(false)
 }
 
-function error (err) {
-  if (typeof err !== 'object' || !err) {
-    err = String(err)
-  } else if (!err.stack) {
-    err = String(err.message || err)
+function getErrorLog (err) {
+  if (err?.delegate && typeof err.delegate === 'function') {
```
szegedi, line 56:
```suggestion
  if (typeof err?.delegate === 'function') {
```

szegedi:
>The delegate is executed by log/writer.js but not by telemetry/log.
>So when executing log.error(() => error) the error will be written to the configured logger but won't be sent via telemetry.

This is super subtle and I wonder if it's a behavior we want to rely on. Nobody reading code would think that a delegate is used just for this effect; it's quite cryptic. I'd rather fully embrace the fact this API is – among other things – a facade for telemetry logs and have a dedicated function name for a local-only logging, e.g. `log.errorLocally`

packages/dd-trace/src/log/log.js
iunanua: it as it is [now](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/log/writer.js#L58)
but we could use `??`

packages/dd-trace/test/telemetry/logs/log-collector.spec.js
iunanua: we were using the Error message as telemetry log message, and we were replacing it by 'omitted' when first frame is not a dd frame.
But now as we are making mandatory to provide a message there is no need to replace it.

iunanua:
> > The delegate is executed by log/writer.js but not by telemetry/log.
> > So when executing log.error(() => error) the error will be written to the configured logger but won't be sent via telemetry.
> 
> This is super subtle and I wonder if it's a behavior we want to rely on. Nobody reading code would think that a delegate is used just for this effect; it's quite cryptic. I'd rather fully embrace the fact this API is – among other things – a facade for telemetry logs and have a dedicated function name for a local-only logging, e.g. `log.errorLocally`

This is a side effect, I'll try to explain it.

Delegates are used to "delay" the message composition after logger has checked if certain logging level is enabled:
```javascript
log.debug(() => `Setting new DSM Context: ${JSON.stringify(dataStreamsContext)}.`) // dataStreamsContext will only be stringifyed if debug level is enabled
```
They are mostly used for debugging logs. These logs are not sent via telemetry.
At the moment and there is no single case using `log.error` with a delegate.

To know if a certain logging level is enabled we check if a particular log channel [has subscribers](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/log/index.js#L63)

But as we want to enable log collection globally, log channels (at the moment error logs so at least `errorChannel`) will always have one subscriber (`telemetry/log`) regardless of whether `DD_TRACE_DEBUG` value is. 
Then there would be no point in doing `log.error(() => error)`

So I decided to execute the delegate only in `log/writer.js`.
Other option could be to execute it in each of the log channel listeners. But as it is a debugging case and they are not sent via telemetry I opted for the first option. That we can of course change and do what we all decide together :smile:

szegedi:
Oh I see, basically you're saying that delegates are only used to log debug messages that are deferred as a performance concern. Okay, I guess that makes sense.

packages/dd-trace/src/log/log.js
rochdev: I think we've had cases where it looked like an error but didn't extend from `Error`, especially from auto-instrumentation where we log errors from external libraries automatically. We could revisit it but it's out of scope of this PR.

packages/dd-trace/src/log/log.js
rochdev: I believe this is for errors.

packages/dd-trace/src/log/writer.js
```diff
@@ -51,32 +52,62 @@ function reset () {
...
+  if (cause) withNoop(() => logger.info(cause))
 }
 
-function warn (message) {
-  if (!logger.warn) return debug(message)
-  withNoop(() => logger.warn(message))
+function onDebug (log) {
+  const { formatted, cause } = getErrorLog(log)
+  if (formatted) withNoop(() => logger.debug(formatted))
+  if (cause) withNoop(() => logger.debug(cause))
```
rochdev, line 88: Does this change the current behaviour for loggers?

packages/dd-trace/src/telemetry/logs/index.js
rochdev: Not sure I understand. What is the current behaviour for this? Are we just logging nothing?

rochdev (approved):
A few questions/comments but otherwise LGTM.

packages/dd-trace/src/log/log.js
iunanua: done

packages/dd-trace/src/telemetry/logs/index.js
iunanua: It is the message for telemetry logs when there is no message.

We were sending the error message as telemetry log message but guys from telemetry complained about it due to cardinality and possible PII data problems.

Now, with the new log api, it is preferable whenever you log an error that is accompanied by a no dynamic message.
```javascript

// if you call log.error with a message and an error
// logger: 
//  - logs 'Error loading file'
//  - logs e
// telemetry:
//  - sends 'Error loading file' message + e.stack
log.error('Error loading file', e) 

```
But if a message is not provided `'Generic Error'` message is used

```javascript
// logger:
// - logs e
// telemetry:
// - sends 'Generic Error' message + e.stack
log.error(e) 
```

rochdev: Then this logic should be pushed to telemetry and not to the publisher.

rochdev (changes requested):

rochdev (dismissed):

iunanua:
@rochdev  changed an expression to make it work in node 12 because some [integration-guardrails(12)](https://github.com/DataDog/dd-trace-js/actions/runs/11888832200/job/33124176615?pr=4856) tests were failing.

rochdev (approved):

PR #4861, opened 2024-11-05
Use sampling on timeline events

szegedi:
### What does this PR do?
Instead of recording all events for the timeline, uses sampling to only select those events that occur during sampling instants. Sampling instants are produced by a Poisson point process with rate equal to the overall configured profiling sampling rate.

### Motivation
We'll be introducing some new event types, increasing their cardinality. Introducing sampling will reduce the number of recorded events, with higher likelihood of discarding short events.

### Additional Notes
JIRA issue [PROF-10852]

[PROF-10852]: https://datadoghq.atlassian.net/browse/PROF-10852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/src/profiling/profilers/events.js
+++ b/packages/dd-trace/src/profiling/profilers/events.js
@@ -292,19 +292,15 @@ class CompositeEventSource {
   }
 }
 
-class PoissonPointProcessSamplingStrategy {
-  constructor (options) {
-    // options.samplingInterval comes in microseconds, we need millis
-    this.samplingIntervalMillis = (options.samplingInterval ?? 1e6 / 99) / 1000
+function createPossionProcessSamplingFilter (samplingIntervalMillis) {
+  let nextSamplingInstant = performance.now()
+  let currentSamplingInstant = 0
+  setNextSamplingInstant()
 
-    this.nextSamplingInstant = performance.now()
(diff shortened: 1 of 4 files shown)


packages/dd-trace/src/profiling/profilers/events.js
```diff
@@ -292,29 +292,68 @@ class CompositeEventSource {
...
   constructor (options = {}) {
     this.type = 'events'
     this.eventSerializer = new EventSerializer()
 
-    const eventHandler = event => {
-      this.eventSerializer.addEvent(event)
+    const eventHandler = event => this.eventSerializer.addEvent(event)
+    const eventFilter = options.timelineSamplingEnabled
+      // options.samplingInterval comes in microseconds, we need millis
+      ? createPossionProcessSamplingFilter((options.samplingInterval ?? 1e6 / 99) / 1000)
```
nsavoire, line 339: I was surprised to see that we never define `options.samplingInterval` in the config and therefore rely on the defaults set in `WallProfiler` / `EventsProfiler`

nsavoire (approved):

PR #4862, opened 2024-11-05
temporary upper bound to the router instrumentation

wantsui:
### What does this PR do?

When you try to spin up an Express 5.x application with the tracer, there's an error about `fast_star`:
```
TypeError: Cannot read properties of undefined (reading 'fast_star')
```

It looks like Express 5.x uses Router 2.x which the instrumentation isn't ready to handle at the moment.

This PR adds a temporary upper bound so we only try to trace router 1.x so the error doesn't get thrown. 

One problem I'm seeing right now is that even though the test passes, the application throws errors like:
```
Found incompatible integration version: router@2.0.0
Error during ddtrace instrumentation of application, aborting.
Error: No original method use.
```

Researching an alternative with https://github.com/DataDog/dd-trace-js/pull/4872

### Motivation

Related to AIDM-438

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/datadog-instrumentations/src/router.js
+++ b/packages/datadog-instrumentations/src/router.js
@@ -170,7 +170,7 @@ function createWrapRouterMethod (name) {
 
 const wrapRouterMethod = createWrapRouterMethod('router')
 
-addHook({ name: 'router', versions: ['>=1'] }, Router => {
+addHook({ name: 'router', versions: ['>=1 <2.0.0'] }, Router => {
   shimmer.wrap(Router.prototype, 'use', wrapRouterMethod)
   shimmer.wrap(Router.prototype, 'route', wrapRouterMethod)
 


packages/datadog-instrumentations/src/router.js
```diff
@@ -170,7 +170,7 @@ function createWrapRouterMethod (name) {
 
 const wrapRouterMethod = createWrapRouterMethod('router')
 
-addHook({ name: 'router', versions: ['>=1'] }, Router => {
+addHook({ name: 'router', versions: ['>=1 <2.0.0'] }, Router => {
```
bouwkast, line 173: So just for my own clarity

We don't support Express v5+ _because_ we actually don't support Router v2+.
Which means that we first need to support v2 of Router and then we can see if that fixes Express v5, if not then we need to also support Express v5

wantsui: Yes, we'd need to properly support Router 2.x along with Express 5.x. In my local testing, I got spans from `express.request` with Express 5.x if we don't try to trace `router 2`, but not any of the express middleware spans. To fully support Express 5.x, we'd need to create middleware spans too.

wantsui:
Just updated the description of the PR that even though all the tests pass, except for 2 unrelated checklist type ones, the app itself is throwing an error:

```
Found incompatible integration version: router@2.0.0
Error during ddtrace instrumentation of application, aborting.
Error: No original method use.
```

I don't consider this to be a great outcome either. Researching an approach where we don't cap the upper bound and just try to fix Router 2 instead: https://github.com/DataDog/dd-trace-js/pull/4872/

wantsui:
Closing this PR in favor of https://github.com/DataDog/dd-trace-js/pull/4872

PR #4863, opened 2024-11-06
[test optimization] Fix active span being null in cypress

juan-fernandez:
### What does this PR do?
Do not run the telemetry tag addition if `this.activeTestSpan` is empty.  

### Motivation
Running this code without checking if `this.activeTestSpan` is dangerous.

### Plugin Checklist

Pending: I don't know exactly how to trigger this condition 😓 . This should nonetheless make the code safer. 

- [ ] Unit tests.

### DIFF
--- a/packages/datadog-plugin-cypress/src/cypress-plugin.js
+++ b/packages/datadog-plugin-cypress/src/cypress-plugin.js
@@ -654,55 +654,57 @@ class CypressPlugin {
         return this.activeTestSpan ? { traceId: this.activeTestSpan.context().toTraceId() } : {}
       },
       'dd:afterEach': ({ test, coverage }) => {
-        const { state, error, isRUMActive, testSourceLine, testSuite, testName, isNew, isEfdRetry } = test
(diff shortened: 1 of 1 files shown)


packages/datadog-plugin-cypress/src/cypress-plugin.js
```diff
@@ -654,55 +654,57 @@ class CypressPlugin {
         return this.activeTestSpan ? { traceId: this.activeTestSpan.context().toTraceId() } : {}
       },
       'dd:afterEach': ({ test, coverage }) => {
+        if (!this.activeTestSpan) {
+          log.warn('There is no active test span in dd:afterEach handler')
+          return null
+        }
         const { state, error, isRUMActive, testSourceLine, testSuite, testName, isNew, isEfdRetry } = test
-        if (this.activeTestSpan) {
```
juan-fernandez, line 658: no changes here, just returning earlier if `this.activeTestSpan` is null.

tonyredondo (approved):

PR #4866, opened 2024-11-06
skip warning if propagator is baggage

ida613:
### What does this PR do?
skip warning for the unconventional propagator [baggage](https://opentelemetry.io/docs/concepts/signals/baggage/). Lucas gave a [good explanation](https://github.com/DataDog/dd-trace-js/pull/4563#discussion_r1815786436) for having baggage as a propagator.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/opentracing/propagation/text_map.js
+++ b/packages/dd-trace/src/opentracing/propagation/text_map.js
@@ -324,7 +324,7 @@ class TextMapPropagator {
           spanContext = this._extractB3MultiContext(carrier)
           break
         default:
-          log.warn(`Unknown propagation style: ${extractor}`)
+          if (extractor !== 'baggage') log.warn(`Unknown propagation style: ${extractor}`)
       }
 
       if (this._config.tracePropagationStyle.extract.includes('baggage') && carrier.baggage) {


packages/dd-trace/src/opentracing/propagation/text_map.js
```diff
@@ -324,7 +324,7 @@ class TextMapPropagator {
           spanContext = this._extractB3MultiContext(carrier)
           break
         default:
```
tlhunter, line 326: Since you're already in a switch statement based on the extractor value you could make this a lot cleaner by doing this:
```js
        case 'baggage':
          break
        default:
          log.warn(`Unknown propagation style: ${extractor}`)
```

ida613: It's kind of awkward having baggage as a propagator, if we go with this cleaner solution (which also makes more sense to me), in the case we have `this._config.tracePropagationStyle.extract = ['baggage', 'datadog']`, we would end up with `spanContext` without `traceId` or `parentId` which the `datadog` propagation style clearly requires. Having baggage as a propagator simply means we are propagating baggage, but it should not override other propagation styles. We've decided on having baggage as a propagator since otel does so.

tlhunter (approved):

PR #4875, opened 2024-11-11
[serverless] Add S3 Span Pointers

nhulston:
### What does this PR do?
Adds span pointers in S3 for `putObject`, `copyObject`, and `completeMultipartUpload` requests.

Span pointers are similar to [Span Links](https://docs.datadoghq.com/tracing/trace_collection/span_links/), but for cases when it is impossible to pass the Trace ID and Span ID between the spans that need to be linked.

When the calculated hashes for the upstream and downstream lambdas match, the Datadog backend will automatically link the two traces together.

Screenshot 2024-11-11 at 1 29 11 PM
When clicking on the linked span, a new tab opens linking to the downstream Lambda function that was triggered by this S3 object update.

### Motivation
This feature already exists in Python, and I'm working on adding it to all other runtimes (Node, .NET, Java, Golang).

### Testing Guidelines

Easy: Checkout [this span](https://ddserverless.datadoghq.com/functions?cloud=aws&config_serverless-auto-span-links=true&entity_view=lambda_functions&fromUser=false&graphType=flamegraph&panel_end=1731688980000&panel_paused=false&panel_start=1731688080000&shouldShowLegend=true&sort=time&sp=%5B%7B%22p%22%3A%7B%22entityId%22%3A%22aws-lambda-functions%2Bnhulston-javascript-upstream-dev-hello%2Bus-east-1%2B425362996713%22%7D%2C%22i%22%3A%22lambda-panel%22%7D%2C%7B%22p%22%3A%7B%22traceID%22%3A%226833201988596926111%22%2C%22selectedSpanID%22%3A%226833201988596926111%22%7D%2C%22i%22%3A%22trace-panel%22%7D%5D&spanID=6833201988596926111&spanViewType=span-links&text_search=nhulston&traceID=6833201988596926111&start=1731685380000&end=1731688980000&paused=false), enable the feature flag, and you will see that it's pointing to the downstream Lambda.

Run [this Lambda function](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/nhulston-javascript-upstream-dev-hello?subtab=envVars&tab=testing) with the event payload
```json
{
  "shouldPutObject": false,
  "shouldCopyObject": false,
  "shouldMultipartUpload": false,
  "useV2": true
}
```
and change one of the bools to true. Also, try both with AWS SDK v2 and v3. Enable the span pointers feature flag, and check Datadog to ensure that the spans are properly linked.

I also added tests:
```shell
# Unit tests
yarn test packages/dd-trace/test/span_pointers.spec.js

# Integration tests
docker run -d -p 4566:4566 localstack/localstack
PLUGINS=aws-sdk SERVICES=localstack DD_DATA_STREAMS_ENABLED=true yarn test:plugins:ci 
```

### Additional Notes
- This handles the 'upstream' case, for Lambdas that make S3 requests. I have another PR in datadog-lambda-js to handle the 'downstream' case, for Lambdas that are triggered by S3 updates. https://github.com/DataDog/datadog-lambda-js/pull/587
- S3 and Dynamo are supported in Python. I will soon open another PR to add support for Dynamo.
- `packages/dd-trace/src/span_pointers.js` is a public API, since `datadog-lambda-js` will reuse the exported fields.


packages/datadog-plugin-aws-sdk/test/s3.spec.js
```diff
@@ -34,19 +35,39 @@ describe('Plugin', () => {
           return agent.load('aws-sdk')
         })
 
-        before(done => {
+        beforeEach(done => {
           AWS = require(`../../../versions/${s3ClientName}@${version}`).get()
 
-          s3 = new AWS.S3({ endpoint: 'http://127.0.0.1:4567', s3ForcePathStyle: true, region: 'us-east-1' })
+          s3 = new AWS.S3({ endpoint: 'http://127.0.0.1:4566', s3ForcePathStyle: true, region: 'us-east-1' })
```
nhulston, line 41: The port here was wrong

packages/datadog-plugin-aws-sdk/test/s3.spec.js
```diff
@@ -34,19 +35,39 @@ describe('Plugin', () => {
...
+          // Fix for LocationConstraint issue - only for SDK v2
+          if (s3ClientName === 'aws-sdk') {
+            s3.api.globalEndpoint = '127.0.0.1'
+          }
```
nhulston, line 45: Fixes an issue with localstack. I followed the 'quick workaround' mentioned here https://github.com/localstack/localstack/issues/8000#issuecomment-1488788710

nhulston, line 45: Fixes an issue with localstack. I followed the 'quick workaround' mentioned in [this LocalStack comment](https://github.com/localstack/localstack/issues/8000#issuecomment-1488788710)

packages/dd-trace/src/span_pointers.js
```diff
@@ -0,0 +1,46 @@
...
+})
+
+/**
+ * Generates a unique hash for an S3 object using its bucket name, key, and ETag
+ * https://github.com/DataDog/dd-span-pointer-rules/blob/main/AWS/S3/Object/README.md
+ * @param {string} bucketName - The name of the S3 bucket containing the object
+ * @param {string} objectKey - The full path/key of the object in the bucket
+ * @param {string} eTag - The ETag value from S3, which may be wrapped in quotes
+ * @returns {string|null} A hash uniquely identifying the S3 request, or null if missing parameters.
+ */
```
nhulston, line 26: Adding detailed comments here, since this function is a public API and will be used in `datadog-lambda-js`

packages/datadog-plugin-aws-sdk/test/s3.spec.js
```diff
@@ -34,20 +35,31 @@ describe('Plugin', () => {
...
-            done()
-          })
+        beforeEach(() => {
+          agent.reset()
         })
 
-        after(done => {
-          s3.deleteBucket({ Bucket: bucketName }, () => {
-            done()
-          })
```
nhulston, line 50: No need to do this, since we already call
```
after(async () => {
          await resetLocalStackS3()
```

packages/datadog-plugin-aws-sdk/src/services/s3.js
```diff
@@ -18,6 +25,33 @@ class S3 extends BaseAwsSdkPlugin {
       bucketname: params.Bucket
     })
   }
+
+  addSpanPointer (span, response) {
+    const request = response?.request
+    const operationName = request?.operation
+    if (!['putObject', 'copyObject', 'completeMultipartUpload'].includes(operationName)) {
```
purple4reina, line 32: I know that, at least in python, it's better to do a `if this == that or these == those` rather than asking if something is in an array. This is because the latter does an allocation which takes much more time. Not sure how js does it though!

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -93,6 +93,7 @@ class BaseAwsSdkPlugin extends ClientPlugin {
         this.responseExtractDSMContext(operation, params, response.data ?? response, span)
       }
       this.addResponseTags(span, response)
+      this.addSpanPointer(span, response)
```
apiarian-datadog, line 96: should this be plural, i.e. `addSpanPointers`?

also, in the boto3 version, we needed both the request parameters and the response object in some cases. is that encompassed in the `response` parameter?

apiarian-datadog: oh, i see further down that the request is a possible part of the response.

still should probably be plural, tho.

packages/datadog-plugin-aws-sdk/src/services/s3.js
```diff
@@ -18,6 +25,33 @@ class S3 extends BaseAwsSdkPlugin {
...
+    }
+
+    const bucketName = request?.params?.Bucket
+    const objectKey = request?.params?.Key
+    // AWS v2 (all 3 operations): `response.data.ETag`
+    // AWS v3 (putObject & completeMultipartUpload): `response.ETag`
+    // AWS v3 (copyObject): `response.CopyObjectResult.ETag`
+    const eTag = response?.data?.ETag || response?.ETag || response?.CopyObjectResult?.ETag
+
+    const pointerHash = generateS3PointerHash(bucketName, objectKey, eTag)
```
apiarian-datadog, line 44: there's a lot of question marks in the code above, so by the time we get here, `bucketName`, `objectKey`, and/or `eTag` may be undefined, right? should we be accounting for that here, maybe logging some kind of useful debug message for ourselves?

packages/datadog-plugin-aws-sdk/src/services/s3.js
```diff
@@ -18,6 +25,33 @@ class S3 extends BaseAwsSdkPlugin {
...
+    if (pointerHash) {
+      const attributes = {
+        'ptr.kind': S3_PTR_KIND,
+        'ptr.dir': SPAN_POINTER_DIRECTION.DOWNSTREAM,
+        'ptr.hash': pointerHash,
+        'link.kind': SPAN_LINK_KIND
+      }
+      span.addLink(ZERO_CONTEXT, attributes)
```
apiarian-datadog, line 52: i guess this is fine for now, but it might be nice to have an `addSpanPointer` method on the span that we can use here and in other situations? or do you want to keep that for a future refactor when we add the dynamo pointers?

packages/datadog-plugin-aws-sdk/test/s3.spec.js
```diff
@@ -74,6 +74,68 @@ describe('Plugin', () => {
           rawExpectedSchema.outbound
         )
 
+        describe('span pointers', () => {
```
apiarian-datadog, line 77: no test for multipart?

packages/dd-trace/src/span_pointers.js
```diff
@@ -0,0 +1,46 @@
+const crypto = require('crypto')
+const SpanContext = require('../../dd-trace/src/opentracing/span_context')
+const id = require('../../dd-trace/src/id')
+
+const SPAN_LINK_KIND = 'span-pointer'
+const S3_PTR_KIND = 'aws.s3.object'
+
+// Span context with ids of 0 is used for span pointers
+const ZERO_CONTEXT = new SpanContext({
```
apiarian-datadog, line 9: did you confirm that all the code that does span pointer serialization is okay with these zero values?

packages/dd-trace/src/span_pointers.js
```diff
@@ -0,0 +1,46 @@
...
+ * Generates a unique hash for an S3 object using its bucket name, key, and ETag
+ * https://github.com/DataDog/dd-span-pointer-rules/blob/main/AWS/S3/Object/README.md
+ * @param {string} bucketName - The name of the S3 bucket containing the object
+ * @param {string} objectKey - The full path/key of the object in the bucket
+ * @param {string} eTag - The ETag value from S3, which may be wrapped in quotes
+ * @returns {string|null} A hash uniquely identifying the S3 request, or null if missing parameters.
+ */
+function generateS3PointerHash (bucketName, objectKey, eTag) {
+  if (!bucketName || !objectKey || !eTag) {
+    return null
```
apiarian-datadog, line 29: maybe we debug log something here? (though there's probably more context upstream for that)

packages/dd-trace/src/span_pointers.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  if (eTag.startsWith('"') && eTag.endsWith('"')) {
+    eTag = eTag.slice(1, -1)
+  }
+  const dataToHash = `${bucketName}|${objectKey}|${eTag}`
+  const hash = crypto.createHash('sha256').update(dataToHash).digest('hex')
```
apiarian-datadog, line 36: might be worth extracting that into a helper function since we'll need something similar when we do the dynamodb hashing

apiarian-datadog (commented):
some notes, though mostly on the span pointers and their api. the javascript looks fine to my inexpert eyes

packages/datadog-plugin-aws-sdk/src/services/s3.js
purple4reina: Very interesting indeed! Thanks for sharing this!

packages/datadog-plugin-aws-sdk/src/base.js
nhulston: Since this is called during an S3 request, and each request only creates one span pointer, I think singular makes sense here. Although we can add multiple span pointers per Lambda execution, we only add one span pointer each time this function is called. What do you think?

packages/datadog-plugin-aws-sdk/src/services/s3.js
nhulston: Yes correct. I'm doing the checks inside of `generateS3PointerHash`, and returning a null hash if any parameter is undefined. 

Not sure if this is the best practice though, I'll see what the APM team thinks

I'll add debugs if `pointerHash` is null!

packages/datadog-plugin-aws-sdk/src/services/s3.js
nhulston: I agree. Something simple like this:
```
  addSpanPointer (attributes) {
    addLink(ZERO_CONTEXT, attributes)
  }
```

that way, we don't have to export ZERO_CONTEXT to the lambda library.

packages/dd-trace/src/span_pointers.js
nhulston: I'll extract any reused parts when I get to dynamodb

packages/dd-trace/src/span_pointers.js
nhulston: Yes, there's no validation for the trace/span id
https://github.com/DataDog/dd-trace-js/blob/5be631470047127fbb171162e94b3cb9fadf48a6/packages/dd-trace/src/format.js#L71-L91

packages/datadog-plugin-aws-sdk/test/s3.spec.js
nhulston: Added, but it's a little messy

packages/datadog-plugin-aws-sdk/src/base.js
apiarian-datadog: i still think it should be plural since this is a base class. the one implementation we currently have is for a single-pointer s3 requests, but we'll shortly have multi-pointer dynamodb requests. i think we should note that in the base api.

nhulston: Ah that makes sense

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -101,6 +102,10 @@ class BaseAwsSdkPlugin extends ClientPlugin {
     // implemented by subclasses, or not
   }
 
+  addSpanPointers (span, response) {
```
khanayan123, line 105: Is this empty method needed on the parent class?

packages/datadog-plugin-aws-sdk/src/services/s3.js
khanayan123: For better readability, I think it’s best to handle early exits and logging in this method instead of in generateS3PointerHash. This way, generateS3PointerHash can assume all necessary parameters are already provided. Handling the early exits and logging here will also eliminate the need for optional chaining, making the code easier to follow.

packages/datadog-plugin-aws-sdk/src/base.js
nhulston: Yes, since we call `this.addSpanPointers(span, response)` in `apm:aws:request:complete`. Seems silly right now since S3 is the only service using it, but we will be expanding span pointers to Dynamo very soon, and eventually other services.

packages/dd-trace/src/span_pointers.js
apiarian-datadog: thanks!

apiarian-datadog (dismissed):
looks good from the span pointer perspective! i'll leave the js-specific things to other reviewers, though.

khanayan123:
Mainly looks good to me just need some unit tests for addSpanPointer to increase code coverage in these files: 

packages/dd-trace/test/noop.spec.js
packages/dd-trace/test/opentracing/span.spec.js
packages/dd-trace/test/opentracing/span.spec.js

khanayan123 (approved):
Sorry, accidentally approved the changes. Just need unit tests in the above files and then it should be good

nhulston:
> Mainly looks good to me just need some unit tests for addSpanPointer to increase code coverage in these files:
> 
> packages/dd-trace/test/noop.spec.js
> packages/dd-trace/test/opentracing/span.spec.js
> packages/dd-trace/test/opentracing/span.spec.js

Added tests for two files, but I don't think it makes sense to add tests to `noop.spec.js`

packages/datadog-plugin-aws-sdk/src/services/s3.js
```diff
@@ -18,6 +25,39 @@ class S3 extends BaseAwsSdkPlugin {
...
+      log.debug('Unable to calculate span pointer hash because of missing parameters.')
+      return
+    }
+
+    const pointerHash = generateS3PointerHash(bucketName, objectKey, eTag)
+    const attributes = {
+      'ptr.kind': S3_PTR_KIND,
+      'ptr.dir': SPAN_POINTER_DIRECTION.DOWNSTREAM,
+      'ptr.hash': pointerHash,
+      'link.kind': SPAN_LINK_KIND
```
apiarian-datadog, line 57: now that i'm back here, it's a bit weird that we're telling the span pointer (through `addSpanPointer`) that it's kind is `SPAN_LINK_KIND`

packages/dd-trace/src/span_pointers.js
```diff
@@ -0,0 +1,33 @@
+const crypto = require('crypto')
+
+const SPAN_LINK_KIND = 'span-pointer'
```
apiarian-datadog, line 3: this should probably be called `SPAN_LINK_SPAN_POINTER_KIND` or something. it makes sense as `SPAN_LINK_KIND`  in this file, but it becomes confusing when it's imported elsewhere.

packages/dd-trace/src/span_pointers.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Generates a unique hash for an S3 object using its bucket name, key, and ETag
+ * https://github.com/DataDog/dd-span-pointer-rules/blob/main/AWS/S3/Object/README.md
+ * @param {string} bucketName - The name of the S3 bucket containing the object
+ * @param {string} objectKey - The full path/key of the object in the bucket
+ * @param {string} eTag - The ETag value from S3, which may be wrapped in quotes
+ * @returns {string} A hash uniquely identifying the S3 request.
+ */
+function generateS3PointerHash (bucketName, objectKey, eTag) {
```
apiarian-datadog, line 19: a little weird that we have an aws s3-specifc funciton here in a file that seems to be general span pointers code?

packages/dd-trace/src/opentelemetry/span.js
```diff
@@ -217,6 +218,14 @@ class Span {
     return this
   }
 
+  addSpanPointer (attributes) {
+    const zeroContext = new SpanContext({
+      traceId: id('0'),
+      spanId: id('0')
+    })
+    return this.addLink(zeroContext, attributes)
```
apiarian-datadog, line 226: arguably we could take care of the link kind stuff here so that callers don't have to know that particular aspect of span pointer storage details. for now they're embedded in the span links list, but in the future we might change that and it would be easier to do so if we didn't leak the span link abstraction out of this function.

apiarian-datadog (dismissed):
still looks fine, though i've added some structural comments after seeing how this is used in the lambda layer

packages/dd-trace/src/span_pointers.js
nhulston: Yes I agree, good catch!

packages/dd-trace/src/span_pointers.js
nhulston: I was going to put all the hashing functions in this file. And once I get to Dynamo, I can pull out any reused parts into this file or use a switch statement inside a single 'generateHash' function

packages/dd-trace/src/opentelemetry/span.js
nhulston: Agreed, I'm just going to make that change now because this is a good suggestion

packages/dd-trace/src/span_pointers.js
```diff
@@ -0,0 +1,27 @@
+const crypto = require('crypto')
+
+const S3_PTR_KIND = 'aws.s3.object'
+
+const SPAN_POINTER_DIRECTION = Object.freeze({
+  UPSTREAM: 'u',
+  DOWNSTREAM: 'd'
+})
```
khanayan123, line 8: NIT: can we move these constants to: packages/dd-trace/src/constants.js

packages/dd-trace/src/span_pointers.js
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Generates a unique hash from an array of strings by joining them with | before hashing.
+ * Used to uniquely identify AWS requests for span pointers.
+ * Expects S3 ETag to already have quotes removed!
+ * @param {string[]} components - Array of strings to hash
+ * @returns {string} A 32-character hash uniquely identifying the components
+ */
+function generatePointerHash (components) {
```
khanayan123, line 17: NIT: rather then adding a new file, would it make more sense for this function to be in:

packages/datadog-plugin-aws-sdk/src/base.js
Or
packages/dd-trace/src/util.js

khanayan123 (commented):
Some minor comments, everything else looks good

packages/dd-trace/src/span_pointers.js
nhulston: We can't do base.js because we need to access it from dd-trace in https://github.com/DataDog/datadog-lambda-js, but we can move it to util.js!

CarlesDD (approved):
Looks good from appsec prespective.

packages/datadog-plugin-aws-sdk/test/s3.spec.js
```diff
@@ -74,6 +74,138 @@ describe('Plugin', () => {
...
+              try {
+                const span = traces[0][0]
+                const links = JSON.parse(span.meta?.['_dd.span_links'] || '[]')
+
+                expect(links).to.have.lengthOf(1)
+                expect(links[0].attributes).to.deep.equal({
+                  'ptr.kind': S3_PTR_KIND,
+                  'ptr.dir': SPAN_POINTER_DIRECTION.DOWNSTREAM,
+                  'ptr.hash': '6d1a2fe194c6579187408f827f942be3',
+                  'link.kind': 'span-pointer'
```
apiarian-datadog, line 89: nano-nit: it might make sense to pull in a constant for this particular `link-kind` check?

packages/dd-trace/src/util.js
```diff
@@ -73,11 +74,25 @@ function hasOwn (object, prop) {
...
 
+/**
+ * Generates a unique hash from an array of strings by joining them with | before hashing.
+ * Used to uniquely identify AWS requests for span pointers.
+ * @param {string[]} components - Array of strings to hash
+ * @returns {string} A 32-character hash uniquely identifying the components
+ */
+function generatePointerHash (components) {
+  // If passing S3's ETag as a component, make sure any quotes have already been removed!
+  const dataToHash = components.join('|')
```
apiarian-datadog, line 85: do we need to make sure to encode this data correctly? or does the `update(...)` call below take care of that?

apiarian-datadog (approved):
looks good. just a super minor nit and a question

packages/datadog-plugin-aws-sdk/test/s3.spec.js
nhulston: Gonna hold off on that for now because any commit will remove all approvals. This is something I can add in the next PR

wconti27 (approved):

apiarian-datadog (approved):
still approved!

apiarian-datadog (approved):
still approved!

apiarian-datadog (approved):
still approved!

PR #4876, opened 2024-11-12
[test optimization] Add Dynamic Instrumentation to jest retries 

juan-fernandez:
### What does this PR do?

* If `DD_TEST_DYNAMIC_INSTRUMENTATION_ENABLED` is set to true, failed tests that are going to be retried add a breakpoint in the uppermost frame of the test error's stack trace. 
* If the line is executed on the retry, we'll automatically make a snapshot of the local variables and send it to datadog via the agentless log exporter.  
* Add `source-map` as dependency. It's 8.2KB minified: https://bundlephobia.com/package/source-map@0.7.4. It's necessary to handle compiled source code run by `jest`. 

Not done in this PR: 
* The integration should only work if explicitly set by the backend. We still need to update our settings endpoint to include this feature.
* The breakpoint should be removed on test end.

### Motivation
Allow users to automatically retry their tests with Dynamic Instrumentation automatically enabled

### Plugin Checklist

- [x] Unit tests.


packages/datadog-instrumentations/src/jest.js
```diff
@@ -236,7 +236,6 @@ function getWrappedEnvironment (BaseEnvironment, jestVersion) {
             name: removeEfdStringFromTestName(testName),
             suite: this.testSuite,
             testSourceFile: this.testSourceFile,
-            runner: 'jest-circus',
```
juan-fernandez, line 239: unrelated fix: this does not need to be reported from the instrumentation anymore, since the only runner is now `jest-circus`

packages/datadog-plugin-jest/src/index.js
```diff
@@ -360,7 +429,7 @@ class JestPlugin extends CiPlugin {
     } = test
 
     const extraTags = {
-      [JEST_TEST_RUNNER]: runner,
+      [JEST_TEST_RUNNER]: 'jest-circus',
```
juan-fernandez, line 432: unrelated fix

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
```diff
@@ -69,14 +71,20 @@ async function addBreakpoint (snapshotId, probe) {
...
-  const [path, scriptId] = script
+  const [path, scriptId, sourceMapURL] = script
 
   log.debug(`Adding breakpoint at ${path}:${line}`)
 
+  let generatedPosition = { line }
+
+  if (sourceMapURL && sourceMapURL.startsWith('data:')) {
+    generatedPosition = await processScriptWithInlineSourceMap({ file, line, sourceMapURL })
+  }
```
juan-fernandez, line 82: this handles inline sourcemaps so that we can add a line probe where the user wants, instead of the compiled code

daniel-mohedano (approved):

watson (approved):
The DI changes look good 👍

PR #4881, opened 2024-11-13
[ASM] Discard inferred spans when resolving the root span of a trace

iunanua:
### What does this PR do?
Visit upwards parent spans in order to discard possible inferred spans based on the `_inferred_span` tag

### Motivation
Fix an Appsec SDK `trackEvent` method problem in serverless environments. SDK root span resolution differs from that of the tracer and tags added by the SDK may be lost if they are added in a inferred span.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/appsec/sdk/utils.js
+++ b/packages/dd-trace/src/appsec/sdk/utils.js
@@ -1,8 +1,27 @@
 'use strict'
 
 function getRootSpan (tracer) {
-  const span = tracer.scope().active()
-  return span && span.context()._trace.started[0]
+  let span = tracer.scope().active()
+  if (!span) return
+
+  const context = span.context()
+  const started = context._trace.started
+
+  let parentId = context._parentId
+  while (parentId) {
+    const parent = started.find(s => s.context()._spanId === parentId)
+    const pContext = parent?.context()
+
+    if (pContext) {
+      parentId = pContext._parentId
+
+      if (!pContext._tags?._inferred_span) {
+        span = parent
+      }
+    }
+  }
+
+  return span
 }
 
 module.exports = {
--- a/packages/dd-trace/src/appsec/sdk/utils.js
+++ b/packages/dd-trace/src/appsec/sdk/utils.js
@@ -18,6 +18,8 @@ function getRootSpan (tracer) {
       if (!pContext._tags?._inferred_span) {
         span = parent
       }
+    } else {
+      break
     }
   }
 
--- a/packages/dd-trace/src/appsec/sdk/utils.js
+++ b/packages/dd-trace/src/appsec/sdk/utils.js
@@ -12,14 +12,12 @@ function getRootSpan (tracer) {
     const parent = started.find(s => s.context()._spanId === parentId)
     const pContext = parent?.context()
 
-    if (pContext) {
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/appsec/sdk/utils.js
```diff
@@ -1,8 +1,26 @@
 'use strict'
 
 function getRootSpan (tracer) {
```
simon-id, line 3: we get the rootspan everyone in appsec's code, is this the only place that needs to be updated ?

packages/dd-trace/test/appsec/sdk/utils.spec.js
```diff
@@ -0,0 +1,115 @@
...
+const Config = require('../../../src/config')
+
+describe('Appsec SDK utils', () => {
+  let tracer
+
+  before(() => {
+    tracer = new Tracer(new Config({
+      enabled: true
+    }))
+    tracer.init({})
```
CarlesDD, line 16: [constructor for tracer](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/proxy.js#L37) (`packages/dd-trace/src/proxy.js`) expects no parameters.

packages/dd-trace/src/appsec/sdk/utils.js
iunanua: Yes, the problem is only with the SDK.

packages/dd-trace/test/appsec/sdk/utils.spec.js
CarlesDD: You removed `init` params but not `constructor`'s

packages/dd-trace/src/appsec/sdk/utils.js
```diff
@@ -1,8 +1,27 @@
...
+
+  const context = span.context()
+  const started = context._trace.started
+
+  let parentId = context._parentId
+  while (parentId) {
+    const parent = started.find(s => s.context()._spanId === parentId)
+    const pContext = parent?.context()
+
+    if (pContext) {
```
IlyasShabi, line 15: If `pContext` is null this will lead to an infinite loop, I suggest to check it first:
```
if (!pContext) {
   break;
}
```

packages/dd-trace/test/appsec/sdk/utils.spec.js
iunanua: :no_good:

packages/dd-trace/test/appsec/sdk/utils.spec.js
CarlesDD, line 1: Nit: Could we swap the order for the assertions parameters to follow the API `equal(actual, expected)`?

IlyasShabi (dismissed):
LGTM

packages/dd-trace/src/appsec/sdk/utils.js
iunanua: fixed and added a new test case for it

CarlesDD (approved):

CarlesDD (approved):

PR #4884, opened 2024-11-13
Add handling when span id and trace id are undefined

ida613:
### What does this PR do?
When we only propagate [baggage](https://github.com/DataDog/dd-trace-js/pull/4563) and nothing else, trace id and span id could be undefined. This PR adds handling for this case. According to otel specification, baggages are attached to contexts rather than spans, but contexts are tied to spans in dd-trace-js. We are essentially using spans to propagate baggages, so trace id and span id could be missing.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/opentracing/span_context.js
+++ b/packages/dd-trace/src/opentracing/span_context.js
@@ -36,7 +36,6 @@ class DatadogSpanContext {
         ? this._trace.tags[TRACE_ID_128] + this._traceId.toString(16).padStart(16, '0')
         : this._traceId.toString(16).padStart(32, '0')
     }
-    if (this._traceId === undefined) return 'unavailable'
     return this._traceId.toString(10)
   }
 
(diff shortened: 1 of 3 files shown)


packages/dd-trace/src/opentracing/span_context.js
```diff
@@ -36,13 +36,15 @@ class DatadogSpanContext {
         ? this._trace.tags[TRACE_ID_128] + this._traceId.toString(16).padStart(16, '0')
         : this._traceId.toString(16).padStart(32, '0')
     }
+    if (this._traceId === undefined) return 'unavailable'
```
rochdev, line 39: Why not just handle this in the caller? (or even better not rely on the span context for this)

Also, this is not tested.

rochdev (changes requested):

packages/dd-trace/src/opentracing/span_context.js
ida613: can we rely on the user to do exception handling? They shouldn't call these two functions if they are only propagating baggage, but I'm not sure if I can trust them

ida613: on a second thought you are probably right, I'll close this PR

rochdev: If they are using OTel they would never be using our internal `SpanContext`.

PR #4895, opened 2024-11-15
make merging safer in release proposal script

rochdev:
### What does this PR do?

Make merging safer in release proposal script.

### Motivation

If the script is run, then the main branch is pull, and the script is run again, it will rebase the changes from the main branch. This is fine, but in doing so it starts diverging from the remote. This causes the next `git pull` from the script to create an unexpected merge commit after the version bump commit. The only situation where this is expected is when actively updating a release, in which case it's safe to ignore the remote changes and just continue with the local ones. Even if there were relevant remote changes (for example someone started a release, someone else continued it, and the first person picked it up again) it's still safe to ignore remote changes, we'll just end up writing new identical commits. The only downside is that in that last situation if a conflict was resolved on the remote, it will need to be resolved again, but I'm not sure how to fix that.


rochdev:
This fix was included in #4897

PR #4901, opened 2024-11-18
fix(config): keep the lookup value as passed

### LINKED ISSUE #4894: lookup function throws `The "options.lookup" property must be of type function. Received type string` in 4.49.0
I recently migrated from dd-trace 4.29 to 4.49
I'm using node 22.11.0. 

This code:
```
const ddtrace = require('dd-trace');
const dns = require("node:dns");
const tracer = ddtrace.init({
    serviceName: 'my-service',
    lookup: (...args) => {
        console.log('LOOKUP CALLED');
        return dns.lookup(...args)
    },
    hostname: 'whaafldjslkfds.com',

});


function main() {
    tracer.trace('trace', () => {
            console.log(`TRACED`);
        }
    );
    setTimeout(() => {
        console.log('Time to send the request to the agent')
    }, 2000)
}

main();
```

Is working and displaying `LOOKUP CALLED` in 4.29.0, but is throwing
`TypeError [ERR_INVALID_ARG_TYPE]: The "options.lookup" property must be of type function. Received type string ('(...args) => {`

In 4.49.0

If I should guess the culprit, I'd say [this line](https://github.com/DataDog/dd-trace-js/blob/a8896ee676c999d2b46cc4a2fa38b82dc9cc0c3e/packages/dd-trace/src/config.js#L955): `lookup` is not a string, it's a function.

blemoine:
The previous version was transforming the lookup
parameter to a string.

### What does this PR do?

Prevent the erroneous transformation of `lookup` parameter into a string. It must be kept as a function

### Motivation

The bug described in #4894 : nodejs throws an error if `lookup` is transformed into a string

Closes #4894


simon-id:
Thank you for the contribution. I've shared this PR to the relevant people, hopefully it'll get some attention!

blemoine:
It doesn't seem to have any appetite to merge it because the v4 branch is too old. I'm closing this.

tlhunter:
Remaking this PR in #5244 so CI runs

PR #4904, opened 2024-11-18
Support tainted strings coming from database for SQLi, SSTi and Code injection

uurien:
### What does this PR do?

Makes the iast able to detect vulnerabilities coming from database sources, tainting the first row in the sql database outputs.
If the data is used directly in a SQL, Template or Code eval a vulnerability is detected. 

### Motivation

Improve our vulnerability detections coverage.

### Checklist

- [x] Unit tests.

### Additional Notes

System tests: https://github.com/DataDog/system-tests/pull/3600


packages/dd-trace/src/appsec/iast/analyzers/injection-analyzer.js
```diff
@@ -1,19 +1,33 @@
...
     const ranges = getRanges(iastContext, value)
     return { value, ranges }
   }
+
+  _areRangesVulnerable (ranges) {
+    if (!ranges) return false
+
+    const nonRowRanges = ranges.filter(range => range.iinfo.type !== SQL_ROW_VALUE)
+
+    return nonRowRanges.length > 0
```
iunanua, line 29:
```suggestion
    return ranges?.some(range => range.iinfo.type !== SQL_ROW_VALUE)
```

packages/dd-trace/src/appsec/iast/analyzers/injection-analyzer.js
```diff
@@ -1,19 +1,33 @@
...
 class InjectionAnalyzer extends Analyzer {
   _isVulnerable (value, iastContext) {
     if (value) {
-      return isTainted(iastContext, value)
+      if (!isTainted(iastContext, value)) {
+        return false
+      }
+
+      return this._areRangesVulnerable(getRanges(iastContext, value))
     }
```
iunanua, line 14: this could be 
```javascript
if (value && isTainted(iastContext, value)) {
  return this._areRangesVulnerable(getRanges(iastContext, value))
}
return false
```

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
```diff
@@ -170,6 +181,37 @@ class TaintTrackingPlugin extends SourceIastPlugin {
...
+
+  _taintDatabaseResult (result, dbOrigin, iastContext = getIastContext(storage.getStore()), name) {
+    if (!iastContext) return result
+
+    let rowsToTaint = this.iastConfig?.dbRowsToTaint
+    if (typeof rowsToTaint !== 'number') {
+      rowsToTaint = 1
+    }
+
+    if (rowsToTaint === 0) return
```
iunanua, line 193: you could move this block to `onConfigure` method and resolve it only once setting a `this.rowsToTaint` var

```suggestion
    if (this.rowsToTaint === 0) return
```

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
```diff
@@ -170,6 +181,37 @@ class TaintTrackingPlugin extends SourceIastPlugin {
     this.taintHeaders(req.headers, iastContext)
     this.taintUrl(req, iastContext)
   }
+
+  _taintDatabaseResult (result, dbOrigin, iastContext = getIastContext(storage.getStore()), name) {
+    if (!iastContext) return result
+
+    let rowsToTaint = this.iastConfig?.dbRowsToTaint
+    if (typeof rowsToTaint !== 'number') {
```
IlyasShabi, line 189: I think that we should trust `dbRowsToTaint` is a number since it's coming from config ?

packages/datadog-instrumentations/src/sequelize.js
```diff
@@ -27,9 +27,14 @@ addHook({ name: 'sequelize', versions: ['>=4'] }, Sequelize => {
         dialect = this.dialect.name
       }
 
-      function onFinish () {
+      function onFinish (result) {
+        const type = options?.type || 'RAW'
+        if (type === 'RAW' && result && result.length > 1) {
```
IlyasShabi, line 32:
```suggestion
        if (type === 'RAW' && result?.length > 1) {
```

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
```diff
@@ -170,6 +181,37 @@ class TaintTrackingPlugin extends SourceIastPlugin {
...
+    if (Array.isArray(result)) {
+      for (let i = 0; i < result.length && i < rowsToTaint; i++) {
+        const nextName = name ? `${name}.${i}` : '' + i
+        result[i] = this._taintDatabaseResult(result[i], dbOrigin, iastContext, nextName)
+      }
+    } else if (result && typeof result === 'object') {
+      if (dbOrigin === 'sequelize' && result.dataValues) {
+        result.dataValues = this._taintDatabaseResult(result.dataValues, dbOrigin, iastContext, name)
+      } else {
+        Object.keys(result).forEach(key => {
```
IlyasShabi, line 204: I think using `for...of` is better for performance here

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
```diff
@@ -170,6 +181,37 @@ class TaintTrackingPlugin extends SourceIastPlugin {
...
+
+    if (rowsToTaint === 0) return
+
+    if (Array.isArray(result)) {
+      for (let i = 0; i < result.length && i < rowsToTaint; i++) {
+        const nextName = name ? `${name}.${i}` : '' + i
+        result[i] = this._taintDatabaseResult(result[i], dbOrigin, iastContext, nextName)
+      }
+    } else if (result && typeof result === 'object') {
+      if (dbOrigin === 'sequelize' && result.dataValues) {
```
IlyasShabi, line 201: IMO distinguishing between databases using `dbOrigin` in this function is unnecessary and could make the code more complex if we add more database subscriptions in the future.

I suggest sending `result.dataValues` directly from the sequelize subscription and removing this part from the function. wdyt?

uurien: It was my first approach, but it will be super expensive. If we are going to taint only one row, we don't need to iterate each row of the data just to taint only the first one. And this object could be complex, like 100 rows of:

```javascript
[
  { ..., dataValues: { id: 123, name: "name", children: [{.., dataValues: { /*another complex object */ }]}
]
```

This is the reason why I decided to check the database type here and not in the instrumentation, going against our good practices.

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
uurien: It is a simple and cheap check, why stop doing it? (and now it is done only once in `onConfigure`)

packages/dd-trace/src/appsec/iast/analyzers/injection-analyzer.js
```diff
@@ -1,19 +1,25 @@
...
 const { isTainted, getRanges } = require('../taint-tracking/operations')
+const { SQL_ROW_VALUE } = require('../taint-tracking/source-types')
 
 class InjectionAnalyzer extends Analyzer {
   _isVulnerable (value, iastContext) {
-    if (value) {
-      return isTainted(iastContext, value)
+    if (value && isTainted(iastContext, value)) {
+      return this._areRangesVulnerable(getRanges(iastContext, value))
     }
```
iunanua, line 10: here there are two calls to the native part, could be something like that?

```suggestion
    const ranges = value && getRanges(iastContext, value)  // `getRanges(iastContext, not_tainted_value)` returns undefined or an empty array?

    if (ranges) {
      return this._areRangesVulnerable(ranges)
    }
```

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
```diff
@@ -184,6 +205,32 @@ class TaintTrackingPlugin extends SourceIastPlugin {
...
   }
+
+  _taintDatabaseResult (result, dbOrigin, iastContext = getIastContext(storage.getStore()), name) {
+    if (!iastContext) return result
+
+    if (this._rowsToTaint === 0) return
+
+    if (Array.isArray(result)) {
+      for (let i = 0; i < result.length && i < this._rowsToTaint; i++) {
+        const nextName = name ? `${name}.${i}` : '' + i
```
iunanua, line 216:
```suggestion
  _taintDatabaseResult (result, dbOrigin, iastContext = getIastContext(storage.getStore()), name = '') {
    if (!iastContext) return result

    if (this._rowsToTaint === 0) return

    if (Array.isArray(result)) {
      for (let i = 0; i < result.length && i < this._rowsToTaint; i++) {
        const nextName = `${name}.${i}`
```

uurien: I don't like the change, and with this change, if name is empty string, the nextName will be `.1` in an array, and not just `1`. 
You are adding an extra `.` here. And it would be weird, if I apply same change in line 224
```javascript
for (const key in result) {
  const nextName = name ? `${name}.${key}` : key
  result[key] = this._taintDatabaseResult(result[key], dbOrigin, iastContext, nextName)
}
```

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
```diff
@@ -184,6 +205,32 @@ class TaintTrackingPlugin extends SourceIastPlugin {
     this.taintHeaders(req.headers, iastContext)
     this.taintUrl(req, iastContext)
   }
+
+  _taintDatabaseResult (result, dbOrigin, iastContext = getIastContext(storage.getStore()), name) {
+    if (!iastContext) return result
+
+    if (this._rowsToTaint === 0) return
```
iunanua, line 212: this should return `result`, no?

returning `undefined` would set `result[i] = undefined` or `result[key] = undefined`

packages/dd-trace/src/appsec/iast/analyzers/injection-analyzer.js
uurien: Yes, I did in this way to prevent creating arrays when it is not tainted,  but if it is not tainted, will create only an empty array, I think that it worth to change to your suggestion. Changing it.

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
iunanua: ah true.
another option could be `name = 'result'`

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
uurien: yeah it should. It doesn't really matter, because if `this._rowsToTaint` is 0, is not going to be into the recursive method, but I'll change it.

packages/dd-trace/test/appsec/iast/taint-tracking/sources/sql_row.pg.plugin.spec.js
```diff
@@ -0,0 +1,113 @@
...
+          await client.query(secondItem.query)
+
+          res.end()
+        }, 'SQL_INJECTION', null, 'Should not taint the second row of a query with default configuration')
+
+        testThatRequestHasNoVulnerability(async (req, res) => {
+          const result = await client.query('SELECT * from examples')
+          const firstItem = result.rows[0]
+
+          const childProcess = require('child_process')
```
IlyasShabi, line 66: We could move this declaration to the top of the file to avoid duplication

packages/dd-trace/test/appsec/iast/taint-tracking/sources/sql_row.sequelize.plugin.spec.js
```diff
@@ -0,0 +1,106 @@
...
+
+          await sequelize.query(examples[1].query)
+
+          res.end('OK')
+        }, 'SQL_INJECTION', null, 'Should not taint the second row of a query with default configuration')
+
+        testThatRequestHasNoVulnerability(async (req, res) => {
+          const examples = await Example.findAll()
+
+          const childProcess = require('child_process')
```
IlyasShabi, line 98: We could move this declaration to the top of the file to avoid duplication

uurien: This is intentional, this `require` is part of the test. If I do something like this in the top of the file, not instrumented method could be loaded because the require could happen before the instrumentation.

```
const { execSync } = require('child_process')
```

This is why I usually try to do the requires into the test, when I am sure that everything is instrumented.

IlyasShabi (approved):
LGTM

iunanua (approved):

PR #4912, opened 2024-11-19
[serverless] Add DynamoDB Span Pointers

nhulston:
*Large PR, but 1000+ lines are tests. Only ~250 lines are actual code from this PR. All these tests are necessary to test all the different possible cases. Please review commit by commit to make the review process easier.*

### What does this PR do?
Adds span pointers in DynamoDB for `putItem`, `updateItem`, `deleteItem`, `transactWriteItems`, and `batchWriteItem` requests.

For more info on the details of how span pointers work for each of these operations, see https://github.com/DataDog/dd-span-pointer-rules/blob/main/AWS/DynamoDB/Item/README.md.

Span pointers are similar to [Span Links](https://docs.datadoghq.com/tracing/trace_collection/span_links/), but for cases when it is impossible to pass the Trace ID and Span ID between the spans that need to be linked.

When the calculated hashes for the upstream and downstream lambdas match, the Datadog backend will automatically link the two traces together.

Screenshot 2024-11-19 at 4 41 31 PM
When clicking on the linked span, a new tab opens linking to the downstream Lambda function that was triggered by this Dynamo object update.

### Motivation
This feature already exists in Python, and I'm working on adding it to all other runtimes (Node, .NET, Java, Golang).

### Additional Notes
- This handles the 'upstream' case, for Lambdas that make Dynmo requests. I will have another PR in datadog-lambda-js to handle the 'downstream' case, for Lambdas that are triggered by Dynamo updates.
- My changes in `packages/dd-trace/src/util.js` are located there because it is a public API, since `datadog-lambda-js` will reuse the exported fields. 
- `UpdateItem` and `DeleteItem` require the primary key(s) as parameters, which makes our lives much easier. However, for PutItem, we have no way of determining which keys are the primary keys, so the user must set an environment variable specifying the primary key(s).
- I add the telemetry key to Datadog backend in https://github.com/DataDog/dd-go/pull/161069

### Testing Guidelines

Easy: Checkout [this span](https://ddserverless.datadoghq.com/functions?cloud=aws&config_serverless-auto-span-links=true&config_trace_show_hidden_metadata=true&entity_view=lambda_functions&fromUser=false&graphType=flamegraph&my_code=enabled&panel_end=1732052495436&panel_paused=false&panel_start=1732038095436&panel_tab=invocations&shouldShowLegend=true&sort=time&sp=%5B%7B%22p%22%3A%7B%22entityId%22%3A%22aws-lambda-functions%2Bnhulston-javascript-upstream-dynamo-dev-hello%2Bus-east-1%2B425362996713%22%7D%2C%22i%22%3A%22lambda-panel%22%7D%2C%7B%22p%22%3A%7B%22traceID%22%3A%222053206778893931889%22%2C%22selectedSpanID%22%3A%222053206778893931889%22%7D%2C%22i%22%3A%22trace-panel%22%7D%5D&spanID=2053206778893931889&spanViewType=span-links&text_search=nhulston-javascript&traceID=2053206778893931889&start=1732048352571&end=1732051952571&paused=false), enable the feature flag, and you will see that it's pointing to the downstream Lambda.

Thorough testing: Run [this Lambda function](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/nhulston-javascript-upstream-dynamo-dev-hello?subtab=envVars&tab=testing) with the event payload
```json
{
  "shouldUseV2": false,
  "twoKeyTable": false,
  "shouldPutItem": false,
  "shouldUpdateItem": false,
  "shouldDeleteItem": false,
  "shouldBatchWriteItems": false,
  "shouldTransactWriteItems": false
}
```
and change one of the bools to true. Also, try both with AWS SDK v2 and v3, and try with twoKeyTable vs oneKeyTable. Enable the span pointers feature flag, and check Datadog to ensure that the spans are properly linked.

I also added tests:
```shell
# Unit tests
yarn test packages/dd-trace/test/util.spec.js

# Integration tests
docker run -d -p 4566:4566 localstack/localstack
PLUGINS=aws-sdk SERVICES=localstack DD_DATA_STREAMS_ENABLED=true yarn test:plugins:ci 
```


packages/dd-trace/src/util.js
```diff
@@ -87,12 +87,83 @@ function generatePointerHash (components) {
...
 module.exports = {
   isTrue,
   isFalse,
   isError,
   globMatch,
   calculateDDBasePath,
   hasOwn,
-  generatePointerHash
+  generatePointerHash,
+  encodeValue,
```
nhulston, line 167: I had to export `encodeValue` to test it, but it's not used externally. Is this fine or is there something better practice I can do?

packages/datadog-plugin-aws-sdk/src/services/dynamodb.js
```diff
@@ -48,6 +51,155 @@ class DynamoDb extends BaseAwsSdkPlugin {
...
+   *
+   * @example
+   * // With env var DD_AWS_SDK_DYNAMODB_TABLE_PRIMARY_KEYS='{"UserTable":["userId","timestamp"]}'
+   * calculatePutItemHash(
+   *   'UserTable',
+   *   { userId: { S: "user123" }, timestamp: { N: "1234567" }, name: { S: "John" } },
+   *   { UserTable: new Set(['userId', 'timestamp']) }
+   * )
+   */
+  static calculatePutItemHash (tableName, item, primaryKeyConfig) {
```
nhulston, line 167: `UpdateItem` and `DeleteItem` require the primary key(s) as parameters, which makes our lives much easier. However, for PutItem, we have no way of determining which keys are the primary keys, so the user must set an environment variable specifying the primary key(s).

packages/datadog-plugin-aws-sdk/src/services/dynamodb.js
```diff
@@ -48,6 +51,155 @@ class DynamoDb extends BaseAwsSdkPlugin {
...
+   * @returns {Object|null} Parsed config from env var or null if empty/missing/invalid config.
+   */
+  static getPrimaryKeyConfig () {
+    const config = DynamoDb.dynamoPrimaryKeyConfig || {}
+    // Return cached config if it exists
+    if (Object.keys(config).length > 0) {
+      return config
+    }
+
+    const primaryKeysEnvVar = process.env.DD_AWS_SDK_DYNAMODB_TABLE_PRIMARY_KEYS
```
khanayan123, line 127: Can you add the reading and setting of this environment variable to a config object into packages/dd-trace/src/config.js?

packages/datadog-plugin-aws-sdk/src/services/dynamodb.js
```diff
@@ -48,6 +51,155 @@ class DynamoDb extends BaseAwsSdkPlugin {
...
+    if (!primaryKeysEnvVar) {
+      log.warn('Missing DD_AWS_SDK_DYNAMODB_TABLE_PRIMARY_KEYS env variable')
+      return null
+    }
+
+    try {
+      const parsedConfig = JSON.parse(primaryKeysEnvVar)
```
khanayan123, line 134: all of this config reading and parsing logic should go into packages/dd-trace/src/config.js and then this function would not be needed

packages/dd-trace/src/util.js
```diff
@@ -87,12 +87,83 @@ function generatePointerHash (components) {
...
+ * Encodes a DynamoDB attribute value to Buffer for span pointer hashing.
+ * @param {Object} valueObject - DynamoDB value in AWS format ({ S: string } or { N: string } or { B: Buffer })
+ * @returns {Buffer} Encoded value as Buffer, or empty Buffer if invalid input.
+ *
+ * @example
+ * encodeValue({ S: "user123" }) -> Buffer("user123")
+ * encodeValue({ N: "42" }) -> Buffer("42")
+ * encodeValue({ B: Buffer([1, 2, 3]) }) -> Buffer([1, 2, 3])
+ */
+function encodeValue (valueObject) {
```
khanayan123, line 100: I think alot of this util code is specific to the aws integration so I think it would make more sense to introduce a util.js file somewhere in packages/datadog-plugin-aws-sdk/src instead

packages/datadog-plugin-aws-sdk/src/services/dynamodb.js
```diff
@@ -72,8 +81,75 @@ class DynamoDb extends BaseAwsSdkPlugin {
...
+   * )
+   */
+  static calculatePutItemHash (tableName, item, primaryKeyConfig) {
+    if (!tableName || !item || !primaryKeyConfig) {
+      log.debug('Unable to calculate hash because missing required parameters')
+      return
+    }
+    const primaryKeySet = primaryKeyConfig[tableName]
+    if (!primaryKeySet || !(primaryKeySet instanceof Set) || primaryKeySet.size === 0 || primaryKeySet.size > 2) {
+      log.warn(`Invalid dynamo primary key config for table ${tableName}: ${JSON.stringify(primaryKeyConfig)}`)
```
apiarian-datadog, line 174: in order to make this more actionable we've recently added a note in the analogous python warning about the environment variable that needs to be updated: https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/_trace/utils_botocore/span_pointers/dynamodb.py#L236-L242

packages/dd-trace/test/util.spec.js
```diff
@@ -86,3 +86,177 @@ describe('generatePointerHash', () => {
...
+
+    it('handles binary (B) type correctly', () => {
+      const binaryData = Buffer.from([1, 2, 3])
+      const result = encodeValue({ B: binaryData })
+      expect(Buffer.isBuffer(result)).to.be.true
+      expect(result).to.deep.equal(binaryData)
+    })
+  })
+
+  describe('edge cases', () => {
```
apiarian-datadog, line 119: i'm worried that there's no way to tell an error from a value that happens to be an empty string or something

packages/dd-trace/test/util.spec.js
```diff
@@ -86,3 +86,177 @@ describe('generatePointerHash', () => {
...
+      const keySet = new Set(['timestamp', 'userId'])
+      const item = {
+        timestamp: { N: '1234567' },
+        userId: { S: 'user123' }
+      }
+      const result = extractPrimaryKeys(keySet, item)
+      expect(result).to.deep.equal(['timestamp', Buffer.from('1234567'), 'userId', Buffer.from('user123')])
+    })
+
+    it('handles and sorts composite key with Object input', () => {
```
apiarian-datadog, line 210: do these sort tests behave differently if the sorting step is omitted from our code? no big deal if not, i guess, just wondering what the underlying key order behavior actually is.

packages/datadog-plugin-aws-sdk/test/dynamodb.spec.js
```diff
@@ -0,0 +1,878 @@
...
+          return agent.close({ ritmReset: false })
+        })
+
+        describe('span pointers', () => {
+          beforeEach(() => {
+            DynamoDb.dynamoPrimaryKeyConfig = null
+            delete process.env.DD_AWS_SDK_DYNAMODB_TABLE_PRIMARY_KEYS
+          })
+
+          function testSpanPointers ({ expectedLength = 1, expectedHashes, operation }) {
```
apiarian-datadog, line 134: nit: shouldn't expected length be equal to the length of expected hashes if that is an array or 1 otherwise? does it need to be a parameter?

apiarian-datadog: or, maybe even better, we can either take `expectedHash` or `expectedHashes` as parameters, failing if we get both, and then do the right thing in this code? the tests below would read a little more clearly then.

packages/datadog-plugin-aws-sdk/test/dynamodb.spec.js
```diff
@@ -0,0 +1,878 @@
...
+          expect(result).to.be.undefined
+        })
+
+        it('handles missing attributes in item', () => {
+          const tableName = 'UserTable'
+          const item = { someOtherField: { S: 'value' } }
+          const keyConfig = { UserTable: new Set(['userId']) }
+
+          const actualHash = DynamoDb.calculatePutItemHash(tableName, item, keyConfig)
+          const expectedHash = generatePointerHash([tableName, 'userId', Buffer.from(''), '', ''])
```
apiarian-datadog, line 748: i would actually expect this to be no hash at all, since we didn't have the key values.

apiarian-datadog: if we hit this a lot, we would fail toward very aggressively linked spans, which is probably the wrong way to go.

packages/datadog-plugin-aws-sdk/test/dynamodb.spec.js
```diff
@@ -0,0 +1,878 @@
...
+        })
+
+        it('handles mixed valid and invalid key types', () => {
+          const keys = {
+            validKey: { S: 'test' },
+            invalidKey: { INVALID: 'value' }
+          }
+          const hash = DynamoDb.calculateHashWithKnownKeys('TestTable', keys)
+          expect(hash).to.equal(
+            generatePointerHash(['TestTable', 'invalidKey', Buffer.from(''), 'validKey', Buffer.from('test')])
```
apiarian-datadog, line 872: this should also probably not return a hash but indicate failure somehow.

apiarian-datadog (commented):
nice! i'm a little worried about the error handling flows.

also, can we add copies at least a couple of the "known hash" tests from either the rules repo or from the python tests? until we have system tests and self-monitoring covering all this stuff, i'd feel more confident that our hashing is consistent that way.

packages/datadog-plugin-aws-sdk/src/services/dynamodb.js
```diff
@@ -48,6 +51,47 @@ class DynamoDb extends BaseAwsSdkPlugin {
 
     return tags
   }
+
+  addSpanPointers (span, response) {
+    const request = response?.request
+    const operationName = request?.operation
```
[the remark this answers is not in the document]
nhulston: It seems unnecessary to me, as if the operationName is missing, we won't enter any case and will just exit the function. And if the request is missing, the operationName will be missing. WDYT?

packages/dd-trace/src/util.js
```diff
@@ -87,12 +87,83 @@ function generatePointerHash (components) {
...
+    return
+  }
+
+  if (keyNames.length === 1) {
+    return [keyNames[0], encodeValue(keyValuePairs[keyNames[0]]), '', '']
+  } else {
+    const [key1, key2] = keyNames.sort()
+    return [
+      key1,
+      encodeValue(keyValuePairs[key1]),
```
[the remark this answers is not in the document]
nhulston: Good catch

packages/datadog-plugin-aws-sdk/test/dynamodb.spec.js
nhulston: Yes good idea! I think it would be simplest to just remove the `expectedLength` parameter, and then we can just. do this:
```
let expectedLength = 0
if (expectedHashes) {
  expectedLength = Array.isArray(expectedHashes) ? expectedHashes.length : 1
}
```

packages/datadog-plugin-aws-sdk/test/dynamodb.spec.js
nhulston: Yes i agree, thanks for the suggestion

packages/dd-trace/test/util.spec.js
nhulston: Yes, the keys are not guaranteed to be in order, so we have to sort them first

packages/dd-trace/test/util.spec.js
nhulston: Good catch, adding logs and returning undefined for these edge cases so we know when an error has occurred

packages/dd-trace/src/util.js
```diff
@@ -87,12 +87,83 @@ function generatePointerHash (components) {
   return hash.substring(0, 32)
 }
 
+/**
+ * Encodes a DynamoDB attribute value to Buffer for span pointer hashing.
+ * @param {Object} valueObject - DynamoDB value in AWS format ({ S: string } or { N: string } or { B: Buffer })
```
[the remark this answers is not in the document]
nhulston: Yes. In SDK V3, the user is required to marshal their item before submitting the PutItem request.

In SDK V2, the user does not have to marshal the item, but if they don't, it will automatically be marshaled into this expected format before we receive it.

packages/datadog-plugin-aws-sdk/src/services/dynamodb.js
```diff
@@ -48,6 +51,47 @@ class DynamoDb extends BaseAwsSdkPlugin {
...
+   *
+   * @example
+   * calculateKeyBasedOperationsHash('UserTable', { userId: { S: "user123" }, timestamp: { N: "1234567" } })
+   */
+  static calculateHashWithKnownKeys (tableName, keys) {
+    if (!tableName || !keys) {
+      log.debug('Unable to calculate hash because missing parameters')
+      return
+    }
+    const keyValues = extractPrimaryKeys(keys, keys)
```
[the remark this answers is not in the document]
nhulston: I agree this is a little confusing. We could do something like this:
```javascript
const keyNamesSet = new Set(Object.keys(keysObject));
const keyValues = extractPrimaryKeys(keyNamesSet, keys)
```

which is definitely more readable. However, we're sacrificing a little bit of memory and runtime to do so, so I'm not sure this change is worth it. WDYT?

packages/datadog-plugin-aws-sdk/src/services/dynamodb.js
nhulston: Updated!

packages/dd-trace/src/util.js
```diff
@@ -87,12 +87,83 @@ function generatePointerHash (components) {
...
+        return Buffer.from(value)
+      case 'N':
+        return Buffer.from(value.toString())
+      case 'B':
+        return Buffer.isBuffer(value) ? value : Buffer.from(value)
+      default:
+        return Buffer.from('')
+    }
+  } catch (err) {
+    return Buffer.from('')
```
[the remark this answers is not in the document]
nhulston: We can't use the logger in this file (we run into an error). 

TODO for me is to come up with a solution to log something on Monday

packages/dd-trace/src/util.js
```diff
@@ -87,12 +87,83 @@ function generatePointerHash (components) {
...
+
+    switch (type) {
+      case 'S':
+        return Buffer.from(value)
+      case 'N':
+        return Buffer.from(value.toString())
+      case 'B':
+        return Buffer.isBuffer(value) ? value : Buffer.from(value)
+      default:
+        return Buffer.from('')
```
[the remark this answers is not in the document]
nhulston: We can't use the logger in this file (we run into an error). 

TODO for me is to come up with a solution to log something on Monday

packages/datadog-plugin-aws-sdk/src/services/dynamodb.js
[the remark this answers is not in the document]
apiarian-datadog: i generally lean towards readability over raw performance, but perhaps the tracer team has a different approach here?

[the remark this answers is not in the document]
nhulston: cc @khanayan123

packages/dd-trace/src/util.js
nhulston: Agreed, thanks for the suggestion

packages/datadog-plugin-aws-sdk/src/services/dynamodb.js
```diff
@@ -133,7 +133,10 @@ class DynamoDb extends BaseAwsSdkPlugin {
     }
     const primaryKeySet = primaryKeyConfig[tableName]
     if (!primaryKeySet || !(primaryKeySet instanceof Set) || primaryKeySet.size === 0 || primaryKeySet.size > 2) {
-      log.warn(`Invalid dynamo primary key config for table ${tableName}: ${JSON.stringify(primaryKeyConfig)}`)
+      log.warn(
+        `span pointers: failed to extract PutItem span pointer: table ${tableName} ` +
+        'not found in primary key names. Please set them through the DD_AWS_SDK_DYNAMODB_TABLE_PRIMARY_KEYS env var.'
```
apiarian-datadog, line 172: might still be nice to include what we parsed the environment variable, in case there's some parsing or other structural issue.

packages/datadog-plugin-aws-sdk/src/util.js
```diff
@@ -0,0 +1,89 @@
...
+const crypto = require('crypto')
+
+/**
+ * Generates a unique hash from an array of strings by joining them with | before hashing.
+ * Used to uniquely identify AWS requests for span pointers.
+ * @param {string[]} components - Array of strings to hash
+ * @returns {string} A 32-character hash uniquely identifying the components
+ */
+function generatePointerHash (components) {
+  // If passing S3's ETag as a component, make sure any quotes have already been removed!
```
apiarian-datadog, line 13: do we have a test case that checks his for our code?

apiarian-datadog (approved):
looking good! just left a few minor nits

packages/datadog-plugin-aws-sdk/src/util.js
nhulston: Are you referring to the ETag quotes? I'm not sure if there's an easy way to force that to happen, since `components` is just an array of Strings

packages/datadog-plugin-aws-sdk/src/services/dynamodb.js
nhulston: If there's a structural issue, it will get handled and logged in the `getPrimaryKeyConfig()` function.

However, I think we should remove the `primaryKeySet.size > 2` check because there could be a case where the table does exist in the primary key names, but we error anyway when it could have worked fine.

packages/datadog-plugin-aws-sdk/src/services/dynamodb.js
```diff
@@ -112,63 +114,23 @@ class DynamoDb extends BaseAwsSdkPlugin {
...
       log.debug('Unable to calculate hash because missing required parameters')
       return
     }
+    if (!primaryKeyConfig) {
+      log.warn('Missing DD_AWS_SDK_DYNAMODB_TABLE_PRIMARY_KEYS env variable')
+      return
+    }
     const primaryKeySet = primaryKeyConfig[tableName]
     if (!primaryKeySet || !(primaryKeySet instanceof Set) || primaryKeySet.size === 0 || primaryKeySet.size > 2) {
       log.warn(`Invalid dynamo primary key config for table ${tableName}: ${JSON.stringify(primaryKeyConfig)}`)
```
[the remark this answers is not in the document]
nhulston: If the `primaryKeyConfig` is missing, it means `getPrimaryKeyConfig` failed to find or parse it. 

Although this part is a little unclear -- i will update the log to tell them to set the keys:
```
    const configStr = this._tracerConfig?.aws?.dynamoDb?.tablePrimaryKeys
    if (!configStr) {
      log.warn('Missing DD_AWS_SDK_DYNAMODB_TABLE_PRIMARY_KEYS env variable')
      return
    }
```

packages/datadog-plugin-aws-sdk/src/util.js
nhulston: But we do test the function in `util.spec.js`

packages/datadog-plugin-aws-sdk/src/services/dynamodb.js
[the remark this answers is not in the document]
nhulston: I do feel like performance is more important than readability, since this will be ran on every Lambda execution that updates S3

packages/datadog-plugin-aws-sdk/src/util.js
apiarian-datadog: do we have tests for the code that handles etags they correctly strip the double quote bits?

packages/datadog-plugin-aws-sdk/src/services/dynamodb.js
apiarian-datadog: (i think the more information we and our customers have while debugging the better, in this case)

nhulston: Fair enough, I'll just update the log to say something more clear like "table {table} not found in primary key names _or the env var was invalid_"

packages/datadog-plugin-aws-sdk/src/util.js
nhulston: It's not so easy to do that without refactoring the S3 code.

I could open another PR to refactor the `s3.js` file and get the bucket/key/etag extracting in a separate function, and then unit test that function.

packages/datadog-plugin-aws-sdk/src/services/dynamodb.js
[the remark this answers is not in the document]
khanayan123: I would agree with readability in this case, it is a little confusing. I think we're missing an additional helper function prior to calling extractPrimaryKeys(keys, keys). Specifically, a function that prepares the key set to be passed into extractPrimaryKeys—rather than using the same object for both parameters WDYT?

khanayan123 (approved):

khanayan123 (approved):

PR #4913, opened 2024-11-20
Express 5 Instrumentation

IlyasShabi:
### What does this PR do?
This PR adds instrumentation for Express5 to support its new features and behavior. Some related changes were previously made by @wconti27  please check their commits for reference.

### Checklist
- [x] **Path params**: Instrumente handleRequest in Layer class in router library
- [x] **Query string**:  Directly instrumente query getter in express
- [x] **Response**: Added instrumentation for `json`, `jsonp`, and `render` in `response.js`
- [x] **Route**: Instrumente router prototype for `use` and `route` methods
- [x] Enable support for tests across different Express versions
- [x] Skip Path not named widcard for Express5 since they are not supported
- [x] Implement request blocking exception handling (to discuss)
- [x] Fix router params callback tests
- [x] Fix RASP lfi and ssrf tests (not related to Express 🤔 )
- [x] ~~Instrumente Node.js querystring used by express to parse query params~~
- [x] remove qs instrumentation


packages/datadog-instrumentations/src/express.js
```diff
@@ -69,6 +69,22 @@ addHook({ name: 'express', versions: ['>=4'] }, express => {
...
 
+addHook({ name: 'express', versions: ['>=5.0.0'] }, express => {
+  shimmer.wrap(express.application, 'handle', wrapHandle)
+  shimmer.wrap(express.Router.prototype, 'use', wrapRouterMethod)
+  shimmer.wrap(express.Router.prototype, 'route', wrapRouterMethod)
+
+  return express
+})
+
+addHook({ name: 'express', file: ['lib/response.js'], versions: ['>=5.0.0'] }, response => {
```
IlyasShabi, line 80: We can also do it for express 4 with same hook I think

packages/datadog-instrumentations/src/router.js
```diff
@@ -27,6 +27,9 @@ function createWrapRouterMethod (name) {
       const req = arguments[arguments.length > 3 ? 1 : 0]
       const next = arguments[lastIndex]
 
+      // explicitly call getter for express5
+      req.query
```
IlyasShabi, line 31: In Express5, `req.query` is not parsed until it is accessed, which can delay sending it to the WAF. This might result in blocking the request too late, after headers are already sent.

To fix this, explicitly call `req.query` at the start of the request to ensure it is parsed and sent to the WAF early.

packages/datadog-instrumentations/src/router.js
```diff
@@ -176,4 +179,34 @@ addHook({ name: 'router', versions: ['>=1'] }, Router => {
   return Router
 })
 
+function createWrapLayerMethod () {
+  // TODO: create a dedicate channel for this
+  const processParamsStartCh = channel('datadog:express:process_params:start')
```
IlyasShabi, line 184: create a new channel

packages/datadog-plugin-express/test/index.spec.js
```diff
@@ -5,10 +5,11 @@ const axios = require('axios')
 const { ERROR_MESSAGE, ERROR_STACK, ERROR_TYPE } = require('../../dd-trace/src/constants')
 const agent = require('../../dd-trace/test/plugins/agent')
 const plugin = require('../src')
+const semver = require('semver')
 
 const sort = spans => spans.sort((a, b) => a.start.toString() >= b.start.toString() ? 1 : -1)
 
-describe('Plugin', () => {
+describe('Plugin', function () {
```
IlyasShabi, line 12: I will check this file later

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
```diff
@@ -77,6 +78,15 @@ class TaintTrackingPlugin extends SourceIastPlugin {
       }
     )
 
+    this.addSub(
+      { channelName: 'datadog:query:read:finish', tag: HTTP_REQUEST_QUERY },
+      ({ query }) => {
+        if (query !== null && typeof query === 'object') {
```
IlyasShabi, line 84: Always use `query` directly instead of `req.query` to avoid an infinite loop

packages/dd-trace/src/appsec/rasp/utils.js
```diff
@@ -46,7 +46,8 @@ function handleResult (actions, req, res, abortController, config) {
   if (blockingAction) {
     const rootSpan = web.root(req)
     // Should block only in express
-    if (rootSpan?.context()._name === 'express.request') {
+    const name = rootSpan?.context()._name
+    if (name === 'express.request' || name === 'web.request') {
```
IlyasShabi, line 50: Question for the APM team: Why do `Express5` spans have a `web` prefix instead of an `express` prefix?

packages/dd-trace/test/appsec/index.express.plugin.spec.js
```diff
@@ -177,6 +175,7 @@ withVersions('express', 'express', version => {
         } catch (e) {
           assert.equal(e.response.status, 403)
           assert.deepEqual(e.response.data, JSON.parse(json))
+          // to be fixed
           sinon.assert.notCalled(paramCallbackSpy)
```
IlyasShabi, line 179: This is related to blocking request

CarlesDD (commented):
It is worth reviewing [this](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/appsec/index.js#L167-L169) and optimizing it to avoid invoking the query getter 3 times.

IlyasShabi:
Yes I forgot to push this change too @CarlesDD

packages/dd-trace/src/appsec/rasp/utils.js
IlyasShabi: Not related after route and use instrumentation

packages/dd-trace/test/plugins/externals.json
```diff
@@ -98,7 +98,7 @@
     },
     {
       "name": "express",
-      "versions": [">=4", ">=4.0.0 <4.3.0", ">=4.3.0"]
+      "versions": [">=4", ">=4.0.0 <4.3.0", ">=4.0.0 <5.0.0", ">=4.3.0 <5.0.0"]
```
IlyasShabi, line 101: https://github.com/fiznool/express-mongo-sanitize/issues/200

packages/datadog-instrumentations/src/express.js
```diff
@@ -129,7 +145,37 @@ addHook({ name: 'express', versions: ['>=4.0.0 <4.3.0'] }, express => {
...
+    shimmer.wrap(requestDescriptor, 'get', function (originalGet) {
+      return function wrappedGet () {
+        const query = originalGet.apply(this, arguments)
+
+        if (queryParserReadCh.hasSubscribers) {
+          const abortController = new AbortController()
+
+          queryParserReadCh.publish({ req: this, res: this.res, query, abortController })
+
+          if (abortController.signal.aborted) return
```
uurien, line 168: This is the right place to make iast work, because we need to taint the values each time the request is called, but is this the right place to use the abortController and call to the waf? Calling to the waf each time that `req.query` code is in the app, could be quite expensive.

packages/datadog-instrumentations/src/querystring.js
```diff
@@ -0,0 +1,24 @@
+'use strict'
+
+const { addHook, channel } = require('./helpers/instrument')
+const shimmer = require('../../datadog-shimmer')
+const names = ['querystring', 'node:querystring']
+
+const querystringParseCh = channel('datadog:querystring:parse:finish')
```
uurien, line 7: why do we need this? isn't this redundant?
We have instrumentation for req.query getter in express, why do we need to instrument this?

packages/datadog-instrumentations/src/router.js
```diff
@@ -169,11 +169,100 @@ function createWrapRouterMethod (name) {
...
+
+    if (!router._queryParsingWrapped) {
+      router._queryParsingWrapped = true
+      const originalHandle = router.handle
+      router.handle = function (req, res, next) {
+        req.query
+        // If query parsing was aborted, don't continue request handling
+        if (res.writableEnded) {
+          return
+        }
```
uurien, line 193: this is a bit hacky and feature related, I think that we should send a custom event in diagnostic_channels in this point, call to the waf and use the AbortController to decide if we should continue or not.

packages/datadog-instrumentations/src/router.js
```diff
@@ -169,11 +169,100 @@ function createWrapRouterMethod (name) {
...
+
+        return originalHandle.apply(this, arguments)
+      }
+    }
+
+    return router
+  }
+
+  WrappedRouter.prototype = originalRouter.prototype
```
uurien, line 202: `shimmer.wrapFunction` should deal with these things.

packages/datadog-instrumentations/src/querystring.js
simon-id: I was thinking maybe for IAST, but if you're also confused then idk hahaha

packages/datadog-instrumentations/src/express.js
simon-id: yeah maybe the best thing to do would be to just remove the waf call here, and instead call it in the router handle request hook

packages/datadog-instrumentations/src/querystring.js
IlyasShabi: This [test](https://github.com/DataDog/system-tests/blob/56b1a4753ecc30b8f93373f9a4e57f05a259b69d/tests/appsec/iast/source/test_parameter_value.py#L44) fails because in express v4 queries are read directly via [query middleware](https://github.com/expressjs/express/blob/4.21.1/lib/middleware/query.js#L27), and IAST subscribes to this event [here](https://github.com/DataDog/dd-trace-js/blob/9d0f9fa7e53c56a3e01d5524e7dd7b61354ca1cd/packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js#L49).

In express v5, query parsing uses both `qs` (same as express v4) and `querystring` ([code here](https://github.com/expressjs/express/blob/9e3dbb43744620b8b490c61c9269a168e0c0e3ae/lib/utils.js#L142)), so we need to instrument `querystring` to ensure the same functionality.

packages/datadog-instrumentations/src/express.js
```diff
@@ -129,7 +145,37 @@ addHook({ name: 'express', versions: ['>=4.0.0 <4.3.0'] }, express => {
...
+addHook({ name: 'express', versions: ['>=4.3.0 <5.0.0'] }, express => {
   shimmer.wrap(express.Router, 'process_params', wrapProcessParamsMethod(2))
   return express
 })
+
+addHook({ name: 'express', file: ['lib/express.js'], versions: ['>=5.0.0'] }, express => {
+  shimmer.wrapFunction(express, function (original) {
+    const app = original.call(this, arguments)
+    const requestProto = Object.getPrototypeOf(app.request)
+    const requestDescriptor = Object.getOwnPropertyDescriptor(requestProto, 'query')
```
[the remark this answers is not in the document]
IlyasShabi: I don't think this gonna work, because we need to execute createApplication first in order to have an app object [here](https://github.com/expressjs/express/blob/9e3dbb43744620b8b490c61c9269a168e0c0e3ae/lib/request.js#L219)

packages/datadog-instrumentations/src/router.js
```diff
@@ -169,11 +169,100 @@ function createWrapRouterMethod (name) {
 
 const wrapRouterMethod = createWrapRouterMethod('router')
 
-addHook({ name: 'router', versions: ['>=1'] }, Router => {
+addHook({ name: 'router', versions: ['>=1 <2'] }, Router => {
```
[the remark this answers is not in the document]
IlyasShabi: TBH no specific reason. I used full zeros in express to compare versions like `5.0.0` and `5.0.1`. For router, it seems stable for now. 
Open to any suggestions!

packages/datadog-instrumentations/src/express.js
IlyasShabi: I agree with you about how expensive this operation could be, but just to clarify:

1. In express v4, we don’t taint the query string using express’ [query method](https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-instrumentations/src/express.js#L92) but instead rely on the `qs` [parse function.](https://github.com/DataDog/dd-trace-js/blob/9d0f9fa7e53c56a3e01d5524e7dd7b61354ca1cd/packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js#L49)
2. IMO, we should harmonize how we taint query strings between express v4 and express v5. This could be achieved either by using both `qs` and `querystring` (though this might lead to false positives since we can’t confirm if the source is from express) or by keeping the getter instrumentation here and using it for IAST, as you suggested. 
We could also apply the same approach to express v4 and then (remove ?) the `qs` and `querystring` instrumentation altogether.

packages/datadog-instrumentations/src/router.js
```diff
@@ -169,11 +169,100 @@ function createWrapRouterMethod (name) {
...
+}
+
+addHook({
+  name: 'router', file: 'lib/layer.js', versions: ['>=2']
+}, Layer => {
+  shimmer.wrap(Layer.prototype, 'handleRequest', wrapHandleRequest)
+  return Layer
+})
+
+function wrapParam (original) {
```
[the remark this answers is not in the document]
simon-id: We've tried together and apparently only one i needed :+1: make sure to double check there is no edge case

packages/datadog-instrumentations/src/express.js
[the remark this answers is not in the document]
simon-id: we tried it together and apparently it works :shrug:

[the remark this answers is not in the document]
simon-id: image

packages/datadog-instrumentations/src/express.js
uurien: Definitely, we can remove the `qs` instrumentation to do it in the same way in express v4 and v5. 
I can't recall why are we instrumenting the qs and not reading the query in other position of the code, I think that we can move all to express instrumentation.

Not necessarily in this PR, if you see that could be done easily, go ahead, but we can change express v4 stuff in other PR.

packages/datadog-instrumentations/src/router.js
```diff
@@ -169,11 +169,103 @@ function createWrapRouterMethod (name) {
...
 })
 
+const queryParserReadCh = channel('datadog:query:read:finish')
+
+addHook({ name: 'router', versions: ['>=2'] }, Router => {
+  const WrappedRouter = shimmer.wrapFunction(Router, function (originalRouter) {
+    return function wrappedMethod (options) {
+      const router = originalRouter.call(this, options)
+
+      if (!router._queryParsingWrapped) {
```
IlyasShabi, line 186: Do not modify router object, use weakSet instead

packages/datadog-plugin-express/test/index.spec.js
```diff
@@ -1201,7 +1230,7 @@ describe('Plugin', () => {
         it('should support capturing groups in routes', done => {
           const app = express()
 
-          app.get('/:path(*)', (req, res) => {
```
IlyasShabi, line 1204: keep this for express 4 skip it for express 5

packages/dd-trace/test/appsec/iast/taint-tracking/sources/taint-tracking.express.plugin.spec.js
```diff
@@ -208,12 +209,8 @@ describe('Path params sourcing with express', () => {
...
       })
 
-      app.param((param, option) => {
-        return checkParamIsTaintedAndNext
-      })
-
-      app.param('parameter1')
-      app.param('parameter2')
+      app.param('parameter1', checkParamIsTaintedAndNext)
+      app.param('parameter2', checkParamIsTaintedAndNext)
```
CarlesDD, line 213: Could it be written like this?
```suggestion
      app.param(['parameter1', 'parameter2'], checkParamIsTaintedAndNext)
```
https://expressjs.com/en/5x/api.html#app.param

packages/dd-trace/test/plugins/externals.json
```diff
@@ -332,7 +332,7 @@
     },
     {
       "name": "express",
-      "versions": [">=4", ">=4.0.0 <4.3.0", ">=4.3.0"]
+     "versions": [">=4", ">=4.0.0 <4.3.0", ">=4.0.0 <5.0.0", ">=4.3.0 <5.0.0", ">=5.0.0"]
```
IlyasShabi, line 335: Wrong indentation
```suggestion
      "versions": [">=4", ">=4.0.0 <4.3.0", ">=4.0.0 <5.0.0", ">=4.3.0 <5.0.0", ">=5.0.0"]
```

packages/datadog-instrumentations/src/router.js
IlyasShabi: @uurien after checking express 5 codebase, we do call handle function only once not at each use(...)
I removed `_queryParsingWrapped` and seems to be good

packages/dd-trace/test/appsec/iast/taint-tracking/sources/taint-tracking.express.plugin.spec.js
IlyasShabi: Done

packages/datadog-instrumentations/src/express.js
iunanua: FYI, it is possible to configure the express5 query parser:

> The simple query parser is based on Node’s native query parser, [querystring](http://nodejs.org/api/querystring.html).
> The extended query parser is based on [qs](https://www.npmjs.org/package/qs).

`simple` is configured by default, but `extended` could be configured and it will use `qs` lib.

But it doesn't matter if we move the instrumentation to express...

packages/datadog-instrumentations/src/querystring.js
IlyasShabi: @iunanua I removed `qs` and `querystring` instrumentation so we do rely only on express

packages/datadog-instrumentations/src/express.js
```diff
@@ -69,6 +67,20 @@ addHook({ name: 'express', versions: ['>=4'] }, express => {
   return express
 })
 
+addHook({ name: 'express', versions: ['>=4.0.0 <5.0.0'] }, express => {
```
simon-id, line 70: btw, just an opinion kinda thing but i'm pretty sure this also works
```suggestion
addHook({ name: 'express', versions: ['4'] }, express => {
```

IlyasShabi: I prefer keeping `'>=4.0.0 <5.0.0'` as it's more explicit and makes the version range clear to any one reading the code.

rochdev: Not sure I understand this last comment as `4` is a lot easier to read then `>=4.0.0 <5.0.0`. You instantly know that `4` cover only v4.x, but with `>=4.0.0 <5.0.0` you need a second to parse it. The former is definitely easier to read.

packages/dd-trace/src/appsec/index.js
```diff
@@ -163,8 +165,9 @@ function incomingHttpEndTranslator ({ req, res }) {
     persistent[addresses.HTTP_INCOMING_COOKIES] = req.cookies
   }
 
-  if (req.query !== null && typeof req.query === 'object') {
-    persistent[addresses.HTTP_INCOMING_QUERY] = req.query
+  const query = req.query
+  if (query !== null && typeof query === 'object') {
+    persistent[addresses.HTTP_INCOMING_QUERY] = query
   }
```
simon-id, line 171: wait why is query still there since we instrument it anyway ?

packages/datadog-plugin-express/test/index.spec.js
```diff
@@ -214,34 +215,57 @@ describe('Plugin', () => {
...
+                expect(spans[index].parent_id.toString()).to.equal(rootSpan.span_id.toString())
+                expect(spans[index].meta).to.have.property('component', 'express')
+                index++
+
+                const handleResource = isExpress4 ? /^bound\s.*$/ : 'handle'
+                if (isExpress4) {
+                  expect(spans[index].resource).to.match(handleResource)
+                } else {
+                  expect(spans[index]).to.have.property('resource', handleResource)
+                }
```
uurien, line 257: Why the double `isExpress4` check?

```suggestion
                if (isExpress4) {
                  expect(spans[index].resource).to.match(/^bound\s.*$/)
                } else {
                  expect(spans[index]).to.have.property('resource', 'handle')
                }
```

packages/dd-trace/src/appsec/index.js
uurien: For the same reason why we are doing it for `req.query` and `req.cookies`? Do you think that we should delete this code? I think that it could be usefull only for frameworks that are not express.

packages/dd-trace/test/appsec/iast/analyzers/nosql-injection-mongodb-analyzer.mongoose.plugin.spec.js
```diff
@@ -27,11 +27,16 @@ describe('nosql injection detection in mongodb - whole feature', () => {
...
         mongoose = require(`../../../../../../versions/mongoose@${mongooseVersion}`).get()
 
-        mongoose.connect(`mongodb://localhost:27017/${dbName}`, {
+        await mongoose.connect(`mongodb://localhost:27017/${dbName}`, {
           useNewUrlParser: true,
           useUnifiedTopology: true
         })
 
+        if (mongoose.models.Test) {
+          delete mongoose.models?.Test
```
uurien, line 36: why?

packages/dd-trace/test/appsec/iast/taint-tracking/sources/taint-tracking.express.plugin.spec.js
```diff
@@ -46,9 +46,10 @@ describe('URI sourcing with express', () => {
       iast.disable()
     })
 
-    it('should taint uri', done => {
+    it('should taint uri', function (done) {
```
uurien, line 49: why this change from lambda to function?

packages/datadog-instrumentations/src/router.js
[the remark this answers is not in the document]
IlyasShabi: Yes, we need both of them. When using `app.param(param, fn)`, some tests are not working properly:
1. https://github.com/DataDog/dd-trace-js/blob/488fcbfdc495c050f967ade046b8cb592376c1d7/packages/dd-trace/test/appsec/index.express.plugin.spec.js#L168
2. https://github.com/DataDog/dd-trace-js/blob/c44d5392f3f1ee5aa8465f9d78c0de66125626ea/packages/dd-trace/test/appsec/iast/taint-tracking/sources/taint-tracking.express.plugin.spec.js#L185
The issue is because the callback function `fn` is executed inside processParams before reaching `layer.handleRequest`

packages/dd-trace/test/appsec/iast/analyzers/nosql-injection-mongodb-analyzer.mongoose.plugin.spec.js
IlyasShabi: We were running this test only with express v4.21.1. Now, it runs with all express versions, and this part was necessary to clean the Test model before creating a new one

uurien: oh, got it, as we are requiring same mongoose version for each express version, require is caching it. 👍

packages/dd-trace/src/appsec/index.js
simon-id: body and cookies are still in this subscriber because there is a million different body and cookie parser, and we only support 1, so in case there also here we send them to the WAF. But for querystring, there is only one parser no ? it's directly built into express.
We should not delete either body or cookie, but query yes

simon-id: @uurien 
> I think that it could be usefull only for frameworks that are not express.

This code was never designed for frameworks other than express. What other stupid framework would mutate the req object directly :laughing:

IlyasShabi: @simon-id  I thought that the only used framework for query and cookies is only express not all of them

simon-id: no that is true, we only support express, but express can have many different kind of body-parser or cookie-parser middlewares, so just in case, we read it at the end of the request too

IlyasShabi: This part is needed for nextjs, PR to add explanation https://github.com/DataDog/dd-trace-js/pull/4967/files

uurien (approved):

packages/datadog-instrumentations/src/router.js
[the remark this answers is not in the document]
IlyasShabi: We checked this together, there is no way to have a single instrumentation to process path parameters. However, we introduced a `WeakSet` to avoid calling the path parameters channel twice.

packages/datadog-instrumentations/src/router.js
```diff
@@ -169,11 +169,105 @@ function createWrapRouterMethod (name) {
...
 
   return Router
 })
 
+const queryParserReadCh = channel('datadog:query:read:finish')
+
+addHook({ name: 'router', versions: ['>=2'] }, Router => {
+  const WrappedRouter = shimmer.wrapFunction(Router, function (originalRouter) {
+    return function wrappedMethod (options) {
+      const router = originalRouter.call(this, options)
```
simon-id, line 184: i don't see a reason to use .call() here
```suggestion
    return function wrappedMethod () {
      const router = originalRouter.call(this, arguments)
```

packages/datadog-instrumentations/src/router.js
```diff
@@ -169,11 +169,105 @@ function createWrapRouterMethod (name) {
...
+          req,
+          res,
+          params: req?.params,
+          abortController
+        })
+
+        if (abortController.signal.aborted) return
+      }
+
+      return fn(req, res, next, param, name)
```
simon-id, line 259: why call the function manually ? also why are you adding the name argument at the end ? 
```suggestion
      return fn.apply(this, arguments)
```

simon-id: also you might need to double check that we're passing the correct `this` value

packages/datadog-instrumentations/src/router.js
IlyasShabi: Yes we can do it with an apply
```suggestion
    return function wrappedMethod (options) {
      const router = originalRouter.apply(this, arguments)
```

simon-id: you're not using options tho are you ?

simon-id: i have the habit of including the arguments only if they're really relevant to the instrumentation

packages/datadog-instrumentations/src/router.js
simon-id: > also why are you adding the name argument at the end ?

IlyasShabi: In fact this here is `undefined` I made some changes to fix its, feel free to suggest something better

IlyasShabi: > also why are you adding the name argument at the end ?

Since we are instrumenting `app.param(name, cb)` we need to change this [test](https://github.com/DataDog/dd-trace-js/pull/4913/files#diff-100808977ae7196a781c6303667a55d2455fab99fed41a31be6d326899ab9a31R80) and check if `req.params[name]` was tainted instead of name

packages/datadog-instrumentations/src/router.js
```diff
@@ -169,11 +169,109 @@ function createWrapRouterMethod (name) {
...
+
+        return originalFn.apply(this, args)
+      }
+    })
+
+    return original.apply(this, arguments)
+  }
+}
+
+addHook({
```
simon-id, line 270: actually i think putting back the named function is correct to pass the correct this value, sorry this is difficult code haha
```suggestion
      return function wrappedFn (req, res) {
        if (routerParamStartCh.hasSubscribers && Object.keys(req.params).length && !visitedParams.has(req.params)) {
          visitedParams.add(req.params)

          const abortController = new AbortController()

          routerParamStartCh.publish({
            req,
            res,
            params: req?.params,
            abortController
          })

          if (abortController.signal.aborted) return
        }

        return originalFn.apply(this, arguments)
```

packages/datadog-instrumentations/src/router.js
```diff
@@ -169,11 +169,109 @@ function createWrapRouterMethod (name) {
...
+            req,
+            res,
+            params: req?.params,
+            abortController
+          })
+
+          if (abortController.signal.aborted) return
+        }
+
+        return originalFn.apply(this, args)
```
simon-id, line 262: actually i think putting back the named function is correct to pass the correct this value, sorry this is difficult code haha
```suggestion
      return function wrappedFn (req, res) {
        if (routerParamStartCh.hasSubscribers && Object.keys(req.params).length && !visitedParams.has(req.params)) {
          visitedParams.add(req.params)

          const abortController = new AbortController()

          routerParamStartCh.publish({
            req,
            res,
            params: req?.params,
            abortController
          })

          if (abortController.signal.aborted) return
        }

        return originalFn.apply(this, arguments)
```

packages/dd-trace/test/appsec/iast/taint-tracking/sources/taint-tracking.express.plugin.spec.js
```diff
@@ -208,12 +208,7 @@ describe('Path params sourcing with express', () => {
         res.status(200).send()
```
simon-id, line 208: this test is now identical to the one above no ?

simon-id:
last two comments then LGTM

packages/dd-trace/test/appsec/iast/taint-tracking/sources/taint-tracking.express.plugin.spec.js
IlyasShabi: The last one make sense only for express v4

packages/datadog-instrumentations/src/router.js
IlyasShabi: Done

simon-id:
LGTM but let's wait for system tests to be ready for merge :+1:

wconti27 (approved):
LGTM!

simon-id (approved):

hoangnd25:
@IlyasShabi Removing qs hook seems to break express v4 with esbuild plugin.
Could you help me have a look

https://github.com/DataDog/dd-trace-js/issues/5056

PR #4927, opened 2024-11-21
Fix IAST standalone sampling priority propagation

CarlesDD:
### What does this PR do?
Moves the handling the standalone sampling priority from the end of the request - when vulnerabilites are sent - to the detection itself.

### Motivation
Set the correct priority for a downstream request whenever an IAST event occurs. Prior to these changes, since the priority was set at the end of the request, the priority of all downstream requests was set incorrectly, as tracer did not know at that time whether an IAST event had occurred.

### Additional Notes
[System Tests PR](https://github.com/DataDog/system-tests/pull/3529) to enable the test for these cases.

[APPSEC-55778](https://datadoghq.atlassian.net/browse/APPSEC-55778)

[APPSEC-55778]: https://datadoghq.atlassian.net/browse/APPSEC-55778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/src/appsec/iast/vulnerability-reporter.js
+++ b/packages/dd-trace/src/appsec/iast/vulnerability-reporter.js
@@ -17,9 +17,7 @@ let resetVulnerabilityCacheTimer
 let deduplicationEnabled = true
 
 function addVulnerability (iastContext, vulnerability) {
-  if (vulnerability?.evidence && vulnerability?.type &&
-    vulnerability?.location) {
-
+  if (vulnerability?.evidence && vulnerability?.type && vulnerability?.location) {
     if (deduplicationEnabled && isDuplicatedVulnerability(vulnerability)) return
 
     VULNERABILITY_HASHES.set(`${vulnerability.type}${vulnerability.hash}`, true)
--- a/packages/dd-trace/src/appsec/iast/vulnerability-reporter.js
+++ b/packages/dd-trace/src/appsec/iast/vulnerability-reporter.js
@@ -22,14 +22,28 @@ function addVulnerability (iastContext, vulnerability) {
 
     VULNERABILITY_HASHES.set(`${vulnerability.type}${vulnerability.hash}`, true)
 
-    if (iastContext?.rootSpan) {
-      keepTrace(iastContext.rootSpan, SAMPLING_MECHANISM_APPSEC)
-      standalone.sample(iastContext.rootSpan)
+    let span = iastContext?.rootSpan
 
+    if (!span && tracer) {
+      span = tracer.startSpan('vulnerability', {
+        type: 'vulnerability'
+      })
+
+      vulnerability.location.spanId = span.context().toSpanId()
+
+      span.addTags({
+        [IAST_ENABLED_TAG_KEY]: 1
+      })
+    }
+
+    keepTrace(span, SAMPLING_MECHANISM_APPSEC)
+    standalone.sample(span)
+
+    if (iastContext?.rootSpan) {
       iastContext[VULNERABILITIES_KEY] = iastContext[VULNERABILITIES_KEY] || []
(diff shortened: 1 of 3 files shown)


packages/dd-trace/src/appsec/iast/vulnerability-reporter.js
```diff
@@ -17,9 +17,15 @@ let resetVulnerabilityCacheTimer
...
-    vulnerability.location) {
-    if (iastContext && iastContext.rootSpan) {
+  if (vulnerability?.evidence && vulnerability?.type && vulnerability?.location) {
+    if (deduplicationEnabled && isDuplicatedVulnerability(vulnerability)) return
+
+    VULNERABILITY_HASHES.set(`${vulnerability.type}${vulnerability.hash}`, true)
+
+    if (iastContext?.rootSpan) {
+      keepTrace(iastContext.rootSpan, SAMPLING_MECHANISM_APPSEC)
+      standalone.sample(iastContext.rootSpan)
```
iunanua, line 27: Could you move these two out of the `if`, and remove them from `sendVulnerabilities` method (lines 57 and 58)?
```suggestion
    keepTrace(iastContext.rootSpan, SAMPLING_MECHANISM_APPSEC)
    standalone.sample(iastContext.rootSpan)
    
    if (iastContext?.rootSpan) {

```

CarlesDD: No I cannot. If these calls were moved outside the conditional, there would be no certainty that a span exists - it is necessary to take into account the cases of vulnerabilities outside the request, in which there is no span and it is created on the fly in the `sendVulnerabilities` method.

CarlesDD: In order to avoid having multiple calls to `keepTrace` and `standalone.sample`, span on the fly creation can be moved to `addVulnerability` method.

WDYT?

iunanua: ah true!
> span on the fly creation can be moved to addVulnerability method.

i like it!!

packages/dd-trace/src/appsec/iast/vulnerability-reporter.js
```diff
@@ -17,13 +17,33 @@ let resetVulnerabilityCacheTimer
...
+
+    keepTrace(span, SAMPLING_MECHANISM_APPSEC)
+    standalone.sample(span)
+
+    if (iastContext?.rootSpan) {
       iastContext[VULNERABILITIES_KEY] = iastContext[VULNERABILITIES_KEY] || []
       iastContext[VULNERABILITIES_KEY].push(vulnerability)
     } else {
-      sendVulnerabilities([vulnerability])
+      sendVulnerabilities([vulnerability], span, false)
```
iunanua, line 46: Instead of passing `isRootSpan` you could finish it here
```suggestion
      sendVulnerabilities([vulnerability], span)
      span.finish()
```

packages/dd-trace/src/appsec/iast/vulnerability-reporter.js
```diff
@@ -17,13 +17,34 @@ let resetVulnerabilityCacheTimer
...
+      })
+
+      vulnerability.location.spanId = span.context().toSpanId()
+
+      span.addTags({
+        [IAST_ENABLED_TAG_KEY]: 1
+      })
+    }
+
+    keepTrace(span, SAMPLING_MECHANISM_APPSEC)
```
iunanua, line 39: ah, one 'remote' thing. should we ensure there is a span before continue?

```suggestion
    if (!span) return

    keepTrace(span, SAMPLING_MECHANISM_APPSEC)
```

iunanua (approved):

PR #4937, opened 2024-11-25
Initial APM side for aws bedrock

yahya-mouman:
### What does this PR do?

This PR adds support for AWS Bedrock through the InvokeModel function.

Adds initial instrumentation support for client-bedrock-runtime >=3.422.0, specifically for:

- `InvokeModel` 

### Motivation

Tracing the Bedrock model invocation calls would help our customers get insights with APM and the LLM Observability product on how they use this service.

### Plugin Checklist

- [x] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
+++ b/packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
@@ -11,7 +11,7 @@ const PROVIDER = {
   META: 'META',
   STABILITY: 'STABILITY',
   MISTRAL: 'MISTRAL'
-};
+}
 
 const enabledOperations = ['invokeModel']
 
@@ -52,7 +52,7 @@ class BedrockRuntime extends BaseAwsSdkPlugin {
 }
 
 class Generation {
-  constructor({ message = '', finish_reason = '', choice_id = '' } = {}) {
+  constructor ({ message = '', finishReason = '', choiceId = '' } = {}) {
     // stringify message as it could be a single generated message as well as a list of embeddings
     this.message = typeof message === 'string' ? message : JSON.stringify(message) || ''
     this.finishReason = finishReason || ''
@@ -61,27 +61,37 @@ class Generation {
 }
 
 class RequestParams {
-  constructor({ prompt = '', temperature = undefined, top_p = undefined, max_tokens = undefined, stop_sequences = [], input_type = '', truncate = '', stream = '', n = undefined } = {}) {
+  constructor ({
+    prompt = '',
+    temperature = undefined,
+    topP = undefined,
+    maxTokens = undefined,
+    stopSequences = [],
+    inputType = '',
+    truncate = '',
+    stream = '',
+    n = undefined
+  } = {}) {
     // TODO: set a truncation limit to prompt
     // stringify prompt as it could be a single prompt as well as a list of message objects
     this.prompt = typeof prompt === 'string' ? prompt : JSON.stringify(prompt) || ''
     this.temperature = temperature !== undefined ? temperature : undefined
-    this.top_p = top_p !== undefined ? top_p : undefined
-    this.max_tokens = max_tokens !== undefined ? max_tokens : undefined
-    this.stop_sequences = stop_sequences || []
-    this.input_type = input_type || ''
+    this.topP = topP !== undefined ? topP : undefined
+    this.maxTokens = maxTokens !== undefined ? maxTokens : undefined
+    this.stopSequences = stopSequences || []
+    this.inputType = inputType || ''
     this.truncate = truncate || ''
     this.stream = stream || ''
     this.n = n !== undefined ? n : undefined
   }
 }
 
-function extractRequestParams(params, provider) {
+function extractRequestParams (params, provider) {
   const requestBody = JSON.parse(params.body)
   const modelId = params.modelId
 
   switch (provider) {
-    case PROVIDER.AI21:
+    case PROVIDER.AI21: {
       let userPrompt = requestBody.prompt
       if (modelId.includes('jamba')) {
         for (const message of requestBody.messages) {
@@ -90,29 +100,79 @@ function extractRequestParams(params, provider) {
           }
         }
       }
-      return new RequestParams({ prompt: userPrompt, temperature: requestBody.temperature, top_p: requestBody.top_p, max_tokens: requestBody.max_tokens, stop_sequences: requestBody.stop_sequences })
-    case PROVIDER.AMAZON:
+      return new RequestParams({
+        prompt: userPrompt,
+        temperature: requestBody.temperature,
+        topP: requestBody.top_p,
+        maxTokens: requestBody.max_tokens,
+        stopSequences: requestBody.stop_sequences
+      })
+    }
+    case PROVIDER.AMAZON: {
       if (modelId.includes('embed')) {
         return new RequestParams({ prompt: requestBody.inputText })
       }
       const textGenerationConfig = requestBody.textGenerationConfig || {}
-      return new RequestParams({ prompt: requestBody.inputText, temperature: textGenerationConfig.temperature, top_p: textGenerationConfig.topP, max_tokens: textGenerationConfig.maxTokenCount, stop_sequences: textGenerationConfig.stopSequences })
-    case PROVIDER.ANTHROPIC:
+      return new RequestParams({
+        prompt: requestBody.inputText,
+        temperature: textGenerationConfig.temperature,
+        topP: textGenerationConfig.topP,
+        maxTokens: textGenerationConfig.maxTokenCount,
+        stopSequences: textGenerationConfig.stopSequences
+      })
+    }
+    case PROVIDER.ANTHROPIC: {
       const prompt = requestBody.prompt || requestBody.messages
-      return new RequestParams({ prompt: prompt, temperature: requestBody.temperature, top_p: requestBody.top_p, max_tokens: requestBody.max_tokens_to_sample, stop_sequences: requestBody.stop_sequences })
-    case PROVIDER.COHERE:
+      return new RequestParams({
+        prompt,
+        temperature: requestBody.temperature,
+        topP: requestBody.top_p,
+        maxTokens: requestBody.max_tokens_to_sample,
+        stopSequences: requestBody.stop_sequences
+      })
+    }
+    case PROVIDER.COHERE: {
       if (modelId.includes('embed')) {
-        return new RequestParams({ prompt: requestBody.texts, input_type: requestBody.input_type, truncate: requestBody.truncate })
+        return new RequestParams({
+          prompt: requestBody.texts,
+          inputType: requestBody.input_type,
+          truncate: requestBody.truncate
+        })
       }
-      return new RequestParams({ prompt: requestBody.prompt, temperature: requestBody.temperature, top_p: requestBody.p, max_tokens: requestBody.max_tokens, stop_sequences: requestBody.stop_sequences, stream: requestBody.stream, n: requestBody.num_generations })
-    case PROVIDER.META:
-      return new RequestParams({ prompt: requestBody.prompt, temperature: requestBody.temperature, top_p: requestBody.top_p, max_tokens: requestBody.max_gen_len })
-    case PROVIDER.MISTRAL:
-      return new RequestParams({ prompt: requestBody.prompt, temperature: requestBody.temperature, top_p: requestBody.top_p, max_tokens: requestBody.max_tokens, stop_sequences: requestBody.stop, top_k: requestBody.top_k })
-    case PROVIDER.STABILITY:
+      return new RequestParams({
+        prompt: requestBody.prompt,
+        temperature: requestBody.temperature,
+        topP: requestBody.p,
+        maxTokens: requestBody.max_tokens,
+        stopSequences: requestBody.stop_sequences,
+        stream: requestBody.stream,
+        n: requestBody.num_generations
+      })
+    }
+    case PROVIDER.META: {
+      return new RequestParams({
+        prompt: requestBody.prompt,
+        temperature: requestBody.temperature,
+        topP: requestBody.top_p,
+        maxTokens: requestBody.max_gen_len
+      })
+    }
+    case PROVIDER.MISTRAL: {
+      return new RequestParams({
+        prompt: requestBody.prompt,
+        temperature: requestBody.temperature,
+        topP: requestBody.top_p,
+        maxTokens: requestBody.max_tokens,
+        stopSequences: requestBody.stop,
+        topK: requestBody.top_k
+      })
+    }
+    case PROVIDER.STABILITY: {
       return new RequestParams()
-    default:
+    }
+    default: {
       return new RequestParams()
+    }
   }
 }
 
@@ -121,33 +181,44 @@ function extractTextAndResponseReason (response, provider, modelName, shouldSetC
 
   try {
     switch (provider) {
-      case PROVIDER.AI21:
+      case PROVIDER.AI21: {
         if (modelName.includes('jamba')) {
           const generations = body.choices || []
           if (generations.length > 0) {
             const generation = generations[0]
-            return new Generation({ message: generation.message, finish_reason: generation.finish_reason, choice_id: shouldSetChoiceIds ? generation.id : undefined })
+            return new Generation({
+              message: generation.message,
+              finishReason: generation.finish_reason,
+              choiceId: shouldSetChoiceIds ? generation.id : undefined
+            })
           }
         }
         const completions = body.completions || []
         if (completions.length > 0) {
           const completion = completions[0]
-          return new Generation({ message: completion.data?.text, finish_reason: completion?.finishReason, choice_id: shouldSetChoiceIds ? completion?.id : undefined })
+          return new Generation({
+            message: completion.data?.text,
+            finishReason: completion?.finishReason,
+            choiceId: shouldSetChoiceIds ? completion?.id : undefined
+          })
         }
         return new Generation()
-      case PROVIDER.AMAZON:
+      }
+      case PROVIDER.AMAZON: {
         if (modelName.includes('embed')) {
           return new Generation({ message: body.embedding })
         }
         const results = body.results || []
         if (results.length > 0) {
           const result = results[0]
-          return new Generation({ message: result.outputText, finish_reason: result.completionReason })
+          return new Generation({ message: result.outputText, finishReason: result.completionReason })
         }
         break
-      case PROVIDER.ANTHROPIC:
-        return new Generation({ message: body.completion || body.content, finish_reason: body.stop_reason })
-      case PROVIDER.COHERE:
+      }
+      case PROVIDER.ANTHROPIC: {
+        return new Generation({ message: body.completion || body.content, finishReason: body.stop_reason })
+      }
+      case PROVIDER.COHERE: {
         if (modelName.includes('embed')) {
           const embeddings = body.embeddings || [[]]
           if (embeddings.length > 0) {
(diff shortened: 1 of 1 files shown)


packages/datadog-instrumentations/src/aws-sdk.js
```diff
@@ -167,7 +167,8 @@ function getChannelSuffix (name) {
     'sns',
     'sqs',
     'states',
-    'stepfunctions'
+    'stepfunctions',
+    'bedrock runtime'
```
sabrenner, line 171: nit: we can just have this check `bedrock`, since `bedrock runtime` includes `bedrock`

packages/datadog-instrumentations/src/aws-sdk.js
```diff
@@ -205,3 +206,8 @@ addHook({ name: 'aws-sdk', file: 'lib/core.js', versions: ['>=2.1.35'] }, AWS =>
   shimmer.wrap(AWS.Request.prototype, 'send', wrapRequest)
   return AWS
 })
+
+addHook({ name: '@aws-sdk/client-bedrock-runtime', file: 'dist-cjs/index.js', versions: ['>=3.693.0'] }, AWS => {
+  shimmer.wrap(AWS.Client.prototype, 'send', wrapRequest())
+  return AWS
+})
```
sabrenner, line 213: we shouldn't need this hook (i think), does the test instrumentation you have work without this?

packages/datadog-instrumentations/src/aws-sdk.js
sabrenner: Actually, disregard. I don't think I'm right here haha

packages/datadog-instrumentations/src/aws-sdk.js
yahya-mouman: I think it works still but why don't we need it ?

sabrenner: all AWS services in the `aws-sdk` or `@aws-sdk/{service}` modules should rely on the `@aws-sdk/smithy-client` or `AWS.Request` functions to make requests to AWS. Since we have those already patched, we don't want to patch the function that's calling the request/client `send` functions

sabrenner: also, as an aside - if we were patching here, we would just pass in `wrapRequest` instead of `wrapRequest()`. This is because `shimmer.wrap` takes in a function as its third argument, so it would look like:
```javascript
shimmer.wrap(AWS.Client.prototype, 'send', (originalSend) => wrapRequest(originalSend))
```
However, in JS, if the lambda function you provide just doex `(x) => foo(x)`, i.e. passing a value directly into another function, you can just pass that function in directly:
```javascript
shimmer.wrap(AWS.Client.prototype, 'send', wrapRequest)
```
Calling the function directly, in this case, doesn't pass the original `send` function into it

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,207 @@
...
+  static get id () { return 'bedrock runtime' }
+
+  generateTags (params, operation, response) {
+    let tags = {}
+    let modelName = ''
+    let modelProvider = ''
+    const modelMeta = params.modelId.split('.')
+    if (modelMeta.length === 2) {
+      [modelProvider, modelName] = modelMeta
+      modelProvider = modelProvider.toUpperCase()
```
sabrenner, line 23: any reason to have it uppercased here? i think it fails conditions when checking agains `_AMAZON`, `_AI21`, etc. as those constants' values are all lowercase

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,207 @@
...
+
+const BaseAwsSdkPlugin = require('../base')
+const log = require('../../../dd-trace/src/log')
+
+const _AI21 = 'ai21'
+const _AMAZON = 'amazon'
+const _ANTHROPIC = 'anthropic'
+const _COHERE = 'cohere'
+const _META = 'meta'
+const _STABILITY = 'stability'
```
sabrenner, line 11: we can leave out the `_` here

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,207 @@
...
+    const modelMeta = params.modelId.split('.')
+    if (modelMeta.length === 2) {
+      [modelProvider, modelName] = modelMeta
+      modelProvider = modelProvider.toUpperCase()
+    } else {
+      [, modelProvider, modelName] = modelMeta
+      modelProvider = modelProvider.toUpperCase()
+    }
+
+    const shouldSetChoiceIds = modelProvider === 'COHERE' && !modelName.includes('embed')
```
sabrenner, line 29:
```suggestion
    const shouldSetChoiceIds = modelProvider === 'cohere' && !modelName.includes('embed')
```
if we change the modelProvider to just be normal case (or even force it to lowercase)

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,207 @@
...
+  }
+  return {}
+}
+
+function extractTextAndResponseReason (response, provider, modelName, shouldSetChoiceIds) {
+  const body = JSON.parse(Buffer.from(response.body).toString('utf8'))
+  const textAndResponseReasons = []
+
+  try {
+    if (provider === 'AI21') {
```
sabrenner, line 112:
```suggestion
    if (provider === _AI21) {
```
let's match against the constant here instead (and remove the underscore once the constant name is changed)

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
sabrenner: if needed, maybe we can force `toLowerCase`. We can definitely keep it uppercase, but it has to be consistent with the casing of the constants above (which right now are lowercase)

packages/datadog-plugin-aws-sdk/test/bedrock.spec.js
```diff
@@ -0,0 +1,135 @@
...
+            const request = {
+              body: JSON.stringify(model.requestBody),
+              contentType: 'application/json',
+              accept: 'application/json',
+              modelId: `${model.ModelId}`
+            }
+
+            const response = {
+              body: JSON.stringify(model.responseBody)
+            }
```
sabrenner, line 106:
```suggestion
            const response = JSON.stringify(model.responseBody)
```
From what I can tell, we don't return a body of `{ body: response }`, it just ends up being the `response` itself. tried this locally for `amazon` models and that seemed the case, hopefully it's consistent for the other model providers 😅

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
yahya-mouman: I forced everything now to uppercase and replaced that with the constant for cohere

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
yahya-mouman: Yes !

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,236 @@
...
+      max_tokens: requestBody.max_tokens || '',
+      stop_sequences: requestBody.stop || [],
+      temperature: requestBody.temperature || '',
+      top_p: requestBody.top_p || '',
+      top_k: requestBody.top_k || ''
+    }
+  } else if (provider === STABILITY) {
+    // TODO: request/response formats are different for image-based models. Defer for now
+    return {}
+  }
```
sabrenner, line 124: I know we have this written out explicitly in the Python integration, but we can probably just leave this branch off. It'll hit the final `return {}` anyways.

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,236 @@
...
+    const temperature = requestBody.temperature || ''
+    const topP = requestBody.topP || ''
+    const maxTokens = requestBody.maxTokens || ''
+    const stopSequences = requestBody.stopSequences || []
+    let prompt = requestBody.prompt
+
+    if (modelId.includes('jamba')) {
+      for (const message of requestBody.messages) {
+        if (message.role === 'user') {
+          prompt = message.content // Return the content of the first user message
```
sabrenner, line 55: are we looking for the first user message or the last in the set of messages? we'll grab the last user message here, as we don't `break` from the loop in this condition, but maybe that's what we want (last message is most recent), but the comment suggests otherwise. double checking because i think this is new logic here that isn't in the Python integration

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,236 @@
...
+        textAndResponseReasons.push({
+          text: generation.text,
+          finish_reason: generation.stop_reason
+        })
+      })
+    } else if (provider === STABILITY) {
+      // TODO: request/response formats are different for image-based models. Defer for now
+    }
+  } catch (error) {
+    log.error('Unable to extract text/finish_reason from response body. Defaulting to empty text/finish_reason.')
```
sabrenner, line 197: i would probably have this as a `log.warn`, i think ideally we reserve a `log.error` for a user mis-configuring the plugin itself (via `tracer.use(...)`), and if it's a problem with the plugin interaction itself we can just `warn` it.

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,236 @@
...
+  try {
+    if (provider === AI21) {
+      const completions = body.completions || []
+      completions.forEach(completion => {
+        textAndResponseReasons.push({
+          text: completion.data ? completion.data.text : '',
+          finish_reason: completion.finishReason || '',
+          choice_id: shouldSetChoiceIds ? completion.id : undefined
+        })
+      })
```
sabrenner, line 141: looks like this logic differs a bit from our Python integration, where we just [grab the first completion and use that](https://github.com/DataDog/dd-trace-py/blob/a4f0e380a459542795f2d881895f18765211893b/ddtrace/contrib/internal/botocore/services/bedrock.py#L176-L181). my understanding is we'll support choices and messages through the Converse API, even if some of the models for `invokeModel` take in messages. let me know though, otherwise this logic looks good!! just wanna confirm

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,236 @@
...
+function extractTextAndResponseReason (response, provider, modelName, shouldSetChoiceIds) {
+  const body = JSON.parse(Buffer.from(response.body).toString('utf8'))
+  const textAndResponseReasons = []
+
+  try {
+    if (provider === AI21) {
+      const completions = body.completions || []
+      completions.forEach(completion => {
+        textAndResponseReasons.push({
+          text: completion.data ? completion.data.text : '',
```
sabrenner, line 137: JS reference -  can do `text: completion.data?.text || ''` here i think. the left hand side will evaluate to `undefined` if  `completion.data` is `null` or `undefined`

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,236 @@
...
+        })
+      })
+    } else if (provider === AMAZON && modelName.includes('embed')) {
+      textAndResponseReasons.push({
+        text: body.embedding || [],
+        finish_reason: '',
+        choice_id: undefined
+      })
+    } else if (provider === AMAZON) {
+      (body.results || []).forEach(result => {
```
sabrenner, line 149: same thing here, looks like we just grab the first result in the Python integration, probably good to keep it consistent.

packages/datadog-plugin-aws-sdk/test/bedrock.spec.js
```diff
@@ -0,0 +1,224 @@
...
+
+      const bedrockRuntimeClientName =
+        moduleName === '@aws-sdk/smithy-client' ? '@aws-sdk/client-bedrock-runtime' : 'aws-sdk'
+      describe('with configuration', () => {
+        before(() => {
+          return agent.load('aws-sdk')
+        })
+
+        before(done => {
+          AWS = require(`../../../versions/${bedrockRuntimeClientName}@${version}`).get()
```
sabrenner, line 32:
```suggestion
          const requireVersion = version === '3.0.0' ? '3.422.0' : '>=3.422.0'
          AWS = require(`../../../versions/${bedrockRuntimeClientName}@${requireVersion}`).get()
```
might be able to try this for CI, although there is probably a better way to do this once the `withVersions` changes land

packages/datadog-plugin-aws-sdk/test/bedrock.spec.js
```diff
@@ -0,0 +1,224 @@
...
+                  'aws.bedrock.request.prompt': prompt
+                })
+                expect(span.metrics).to.include({
+                  'aws.bedrock.request.temperature': temperature,
+                  'aws.bedrock.request.top_p': topP,
+                  'aws.bedrock.request.max_tokens': maxTokens
+                })
+              }).catch(done)
+              done()
+            })
```
sabrenner, line 218:
```suggestion
            agent.use(traces => {
              const span = traces[0][0]
              expect(span.meta).to.include({
                'aws.operation': 'invokeModel',
                'aws.bedrock.request.model': model.modelId.split('.')[1],
                'aws.bedrock.request.model_provider': `${model.provider}`,
                'aws.bedrock.request.prompt': prompt
              })
              expect(span.metrics).to.include({
                'aws.bedrock.request.temperature': temperature,
                'aws.bedrock.request.top_p': topP,
                'aws.bedrock.request.max_tokens': maxTokens
              })
            }).then(done, done)
            
            bedrockRuntimeClient.send(command, (err) => {
              if (err) return done(err)
            })
```
I would change it to something like this. the way it was before, having `done()` at the end of the bedrock callback would just end the test and pass it before the callback in `agent.use` executed. This way, the test will only end when either the `send` command fails, the `agent.use` callback fails, or the `agent.use` callback succeeds. Tried it out locally and it looks like a couple tests were silently failing (AI21 and Anthropic).

sabrenner (commented):
did a first pass, the core logic of the integration lgtm! just had a few questions about discrepancies we're introducing between the python integration and this one but it's more for my own edification. also hopefully some test changes suggestions that help a bit 🙏

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
yahya-mouman: Ah I initially aimed to get the first user message. But after reading the docs more I realized that it's the whole conversation with the model and we're going to be interested in the most recent user prompt instead of the first one. I need to change the comment.

yahya-mouman: After taking a look at the docs it appears that for A21 labs models there are 2 series the Jurassic and the Jamba types which have different input formats. In the python implementation we only support the Jurassic model arch type. So I added support here for the Jamba one as well.

packages/datadog-plugin-aws-sdk/test/bedrock.spec.js
yahya-mouman: It works ! Nice trick to bypass the limitation on `withVersions`

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
sabrenner: I think we can just grab the first item here instead of doing it later

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,254 @@
...
+
+const AI21 = 'AI21'
+const AMAZON = 'AMAZON'
+const ANTHROPIC = 'ANTHROPIC'
+const COHERE = 'COHERE'
+const META = 'META'
+const STABILITY = 'STABILITY'
+const MISTRAL = 'MISTRAL'
+
+const enabledOperations = ['invokeModel'] // TODO check if this is the correct operation
```
sabrenner, line 14:
```suggestion
const enabledOperations = ['invokeModel']
```
we should be good here now!

packages/datadog-instrumentations/src/aws-sdk.js
tlhunter: It does feel a little naughty to use a space in the channel name though I suppose there is no technical reason not to.

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,254 @@
...
+const BaseAwsSdkPlugin = require('../base')
+const log = require('../../../dd-trace/src/log')
+
+const AI21 = 'AI21'
+const AMAZON = 'AMAZON'
+const ANTHROPIC = 'ANTHROPIC'
+const COHERE = 'COHERE'
+const META = 'META'
+const STABILITY = 'STABILITY'
+const MISTRAL = 'MISTRAL'
```
tlhunter, line 12: It can be a code smell if there's a bunch of constants with the value being exactly the same as the variable name. It looks like these are something called "providers"? If so maybe either prefix them with `PROVIDER_` or toss them together like this:

```js
const PROVIDER = {
  AI21: 'AI21',
  // ...
};
// ...
if (provider === PROVIDER.AI21) {
```

This way it's super clear the purpose of the otherwise unrelated top level constants.

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,254 @@
...
+
+    return tags
+  }
+}
+
+function extractRequestParams (params, provider) {
+  const requestBody = JSON.parse(params.body)
+  const modelId = params.modelId
+
+  if (provider === AI21) {
```
tlhunter, line 56: Can you make this a switch / case since every branch is comparing the value of `provider`?

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,254 @@
...
+      stop_sequences: requestBody.stop_sequences || []
+    }
+  } else if (provider === COHERE && modelId.includes('embed')) {
+    return {
+      prompt: requestBody.texts,
+      input_type: requestBody.input_type || '',
+      truncate: requestBody.truncate || ''
+    }
+  } else if (provider === COHERE) {
+    return {
```
tlhunter, line 106: Up to you but this common object shape might make for a good class, maybe something like `RequestParams` based on the function name.

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,254 @@
...
+            choice_id: shouldSetChoiceIds ? generation.id : undefined
+          })
+        })
+      }
+      const completions = body.completions || []
+      completions.forEach(completion => {
+        textAndResponseReasons.push({
+          text: completion.data?.text || '',
+          finish_reason: completion.finishReason || '',
+          choice_id: shouldSetChoiceIds ? completion.id : undefined
```
tlhunter, line 159: These might also make a good class candidate.

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime.js
```diff
@@ -0,0 +1,254 @@
...
+        })
+      })
+    } else if (provider === ANTHROPIC) {
+      textAndResponseReasons.push({
+        text: body.completion || body.content || '',
+        finish_reason: body.stop_reason || '',
+        choice_id: undefined
+      })
+    } else if (provider === COHERE && modelName.includes('embed')) {
+      (body.embeddings || [[]]).forEach(embedding => {
```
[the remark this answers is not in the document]
tlhunter: Maybe it's intentional? I suppose line 218 also adds an empty entry, albeit with `text` being an empty string instead of an empty array.

In either case it might be beneficial to add a comment above `const textAndResponseReasons` to explain what goes in it and if it's fine to have empty entries or not in it.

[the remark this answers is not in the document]
yahya-mouman: Yes it is intentional as the Cohere model has a response struct set as a list of embeddings https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed.html#model-parameters-embed-request-response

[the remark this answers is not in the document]
yahya-mouman: I added more clarification in the recent refactoring commits that would stringify the fields if they are not strings.
Thanks for noting this.

tlhunter (approved):

sabrenner (approved):
This is really great stuff!! We can do the config changes (I/O character limit and sampling) in a follow-up PR next week once I merge in some shared utilities for the other LLM-based plugins. Left one small nit but can also be left for follow-up cleanup. Great work!

PR #4946, opened 2024-11-26
[test optimization] Do not init on package managers

juan-fernandez:
### What does this PR do?
Do not initialize test optimization (FKA test visibility) if the running process is a package manager, such as `yarn` or `npm`. 

So for example if you run your tests like this:

```json
  "scripts": {
    "test": "jest"
  }
```
```
yarn test
```

`yarn` actually spawns a _new_ process for `jest`, meaning that at least two `dd-trace` initializations happen.

### Motivation
We recommend initializing test optimization with `NODE_OPTIONS='-r dd-trace/ci/init'`. This means that _every_ node process will initialize the tracer. This is usually safe, but it leads to unnecessary code being run. By avoiding the most commonly used processes, that is, package managers, we already save a bunch of initializations.

### Plugin Checklist

- [x] Unit tests.

### DIFF
--- a/ci/init.js
+++ b/ci/init.js
@@ -71,8 +71,6 @@ if (shouldInit) {
   tracer.init(options)
   tracer.use('fs', false)
   tracer.use('child_process', false)
-} else {
-  console.log('not init')
 }
 
 module.exports = tracer
--- a/ci/init.js
+++ b/ci/init.js
@@ -1,10 +1,21 @@
 /* eslint-disable no-console */
 const tracer = require('../packages/dd-trace')
 const { isTrue } = require('../packages/dd-trace/src/util')
+const log = require('../packages/dd-trace/src/log')
 
 const isJestWorker = !!process.env.JEST_WORKER_ID
 const isCucumberWorker = !!process.env.CUCUMBER_WORKER_ID
 const isMochaWorker = !!process.env.MOCHA_WORKER_ID
+
+const packageManagers = [
+  'npm',
+  'yarn',
+  'pnpm'
+]
+
+const isPackageManager = () => {
+  return packageManagers.some(packageManager => process.argv[1]?.includes(`bin/${packageManager}`))
+}
 
 const options = {
   startupLogs: false,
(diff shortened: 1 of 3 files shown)


integration-tests/test-api-manual.spec.js
```diff
@@ -10,24 +10,20 @@ const {
...
   before(async () => {
     sandbox = await createSandbox([], true)
     cwd = sandbox.folder
-    webAppPort = await getPort()
-    webAppServer.listen(webAppPort)
   })
 
   after(async () => {
     await sandbox.remove()
-    await new Promise(resolve => webAppServer.close(resolve))
```
juan-fernandez, line 30: unrelated fixes in this file: the webserver is useless

integration-tests/test-optimization-startup.spec.js
```diff
@@ -0,0 +1,107 @@
...
+    )
+
+    childProcess.stdout.on('data', (chunk) => {
+      testOutput += chunk.toString()
+    })
+
+    childProcess.on('exit', () => {
+      assert.include(testOutput, 'dd-trace is not initialized in a package manager')
+      done()
+    })
```
vitor-de-araujo, line 105: Should we also have a test that it _doesn't_ skip in other situations?

vitor-de-araujo (approved):

integration-tests/test-optimization-startup.spec.js
juan-fernandez: yeah good point

juan-fernandez: done in https://github.com/DataDog/dd-trace-js/pull/4946/commits/ede950b4b5818433c33fb9ad850827176f11b405

vitor-de-araujo (approved):

ci/init.js
```diff
@@ -14,6 +25,11 @@ const options = {
 
 let shouldInit = true
 
+if (isPackageManager()) {
```
rochdev, line 28: @juan-fernandez Why do this only for CI? We pretty much never want the library as a whole to instrument package managers (or anything else in the denylist).

rochdev: > does APM or other products suggest users to init dd-trace via NODE_OPTIONS='-r dd-trace/init'?

Yes, the recommended initialization is with `NODE_OPTIONS`, and other approaches are discouraged (and will also be removed when the new public API lands). There are also some environments where programmatic init cannot be done and `NODE_OPTIONS` must be used.

> If the initialization is programmatic (as suggested in the docs:

Not sure why this is still the case today as we've been recommending `NODE_OPTIONS` for years now 😅

PR #4949, opened 2024-11-26
update guardrails to report telemetry in old node versions

rochdev:
### What does this PR do?

Update guardrails to report telemetry in old node versions.

### Motivation

Per spec, we should be reporting telemetry in all versions of Node that we could realistically encounter.


integration-tests/init.spec.js
```diff
@@ -84,30 +83,7 @@ function testRuntimeVersionChecks (arg, filename) {
       }
     }
 
-    if (!currentVersionCanLog) {
```
rochdev, line 87: Removed this test since it's effectively replaced by the "unsupported guardrails" tests for now until we rewrite these tests entirely.

.github/workflows/project.yml
```diff
@@ -34,7 +34,7 @@ jobs:
   integration-guardrails:
     strategy:
       matrix:
-        version: [12.0.0, 12, 14.0.0, 14, 16.0.0, 16, 18.0.0, 18.1.0, 20.0.0, 22.0.0]
```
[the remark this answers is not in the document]
rochdev: This was effectively testing that Node 12.0.0 is not crashing since the guardrails were incompatible with it. Now we test this in a different test suite that does `0.8` to `12` for that case. Actually, right now it does `0.8` to `10`, let me add `12.0.0` to it.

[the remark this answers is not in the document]
rochdev: Added to the other job.

packages/dd-trace/src/guardrails/telemetry.js
```diff
@@ -0,0 +1,78 @@
+'use strict'
```
rochdev, line 1: This file was copied as-is with only ECMAScript version related changes.

packages/dd-trace/src/guardrails/index.js
```diff
@@ -0,0 +1,67 @@
+'use strict'
```
rochdev, line 1: This file was copied as-is with only ECMAScript version related changes, as well as calling a callback instead of initializing directly.

bengl (approved):

packages/dd-trace/src/guardrails/telemetry.js
```diff
@@ -0,0 +1,78 @@
...
+    // This one can only be sent once, regardless of tags
+    return seen.includes('abort')
+  }
+  if (point.name === 'abort.integration') {
+    // For now, this is the only other one we want to dedupe
+    var compiledPoint = point.name + point.tags.join('')
+    return seen.includes(compiledPoint)
+  }
+  return false
+}
```
BridgeAR, line 46: @rochdev I just stumbled upon this while looking at some code.
It seems like `seen` is never changed throughout the code, so it's always going to return false. I wonder if that's critical or not. We should probably either remove this or fix it (after checking for includes, we add the entry to the seen array) and add tests?

PR #4951, opened 2024-11-27
(fix) Verifying span id is nonzero during extraction

mhlidd:
### What does this PR do?
Adds a check to ensure that Span Links are only created in compound header extraction when the `SpanID` is valid. 
### Motivation
This system-tests [PR](https://github.com/DataDog/system-tests/pull/3499/files#diff-2f9154dbb913f2bbd44191a5bcaeea03cb0ed080ab29fe2ca76e1fcef1b26044R123-R144) introduces a test to ensure that all libraries only create span links in distributed tracing header extraction when the `SpanID` is valid. This PR aims to implement this check in dd-trace-js.
### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/opentracing/propagation/text_map.js
+++ b/packages/dd-trace/src/opentracing/propagation/text_map.js
@@ -497,7 +497,7 @@ class TextMapPropagator {
 
   _extractGenericContext (carrier, traceKey, spanKey, radix) {
     if (carrier[traceKey] && carrier[spanKey]) {
-      if (invalidSegment.test(carrier[traceKey])) return null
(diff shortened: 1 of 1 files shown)


rochdev (changes requested):
Let's make sure to also add a test here.

rochdev:
Doesn't this break Synthetics? They always send an ID of `0`.

mhlidd:
After deliberation with the team, the consensus is to leave it as is in order to preserve synthetics

PR #4952, opened 2024-11-27
stop using native metrics for newer node versions

rochdev:
### What does this PR do?

Stop using native metrics for newer Node versions.

### Motivation

All the runtime metrics we need are now available out of the box from Node, so loading a native extension is no longer necessary to capture them.

### DIFF
--- a/packages/dd-trace/src/runtime_metrics.js
+++ b/packages/dd-trace/src/runtime_metrics.js
@@ -7,11 +7,18 @@ const os = require('os')
 const { DogStatsDClient } = require('./dogstatsd')
 const log = require('./log')
 const Histogram = require('./histogram')
-const { performance } = require('perf_hooks')
+const { performance, monitorEventLoopDelay } = require('perf_hooks')
 
+const { NODE_MAJOR, NODE_MINOR } = require('../../../version')
 const INTERVAL = 10 * 1000
 
+const allMetricsAvailable = NODE_MAJOR > 22 ||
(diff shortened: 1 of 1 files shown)


juan-fernandez (approved):

packages/dd-trace/src/runtime_metrics.js
```diff
@@ -7,11 +7,18 @@ const os = require('os')
...
 const log = require('./log')
 const Histogram = require('./histogram')
-const { performance } = require('perf_hooks')
+const { performance, monitorEventLoopDelay } = require('perf_hooks')
 
+const { NODE_MAJOR, NODE_MINOR } = require('../../../version')
 const INTERVAL = 10 * 1000
 
+const allMetricsAvailable = NODE_MAJOR > 22 ||
+  (NODE_MAJOR === 22 && NODE_MAJOR >= 8) ||
```
szegedi, line 16:
```suggestion
  (NODE_MAJOR === 22 && NODE_MINOR >= 8) ||
```

szegedi (approved):
This is great! One less native component to worry about.

szegedi (changes requested):
This is great! One less native component to worry about. I think there's one type tho'.

rochdev:
Turns out the Node implementation is invalid, so this PR cannot work. I will open an issue in Node, but I think their implementation will have to be completely rewritten for this to work.

rochdev:
Superseded by #4961 since `monitorLoopDelay` is broken.

PR #4953, opened 2024-11-28
use weakmap to avoid references from node to datadog stores

rochdev:
### What does this PR do?

Use WeakMap to avoid references from Node to Datadog stores.

### Motivation

When using AsyncLocalStorage, it attaches the store to the underlying resource directly. This means that if the resource is cloned, everything in all stores referenced by that resource will also be cloned, and some objects that may be referenced like sockets cannot be cloned, resulting in an error at best, and a segmentation fault at worst. Using a WeakMap makes it so that it's the map that references both a handle that is used as the store, and the real store stored in the map so that the store is no longer cloned.

This is an issue with `AsyncLocalStorage` and not with dd-trace so it should be fixed in Node, but for now we need to at least fix the immediate issue in the library.

### DIFF
--- a/packages/datadog-core/src/storage.js
+++ b/packages/datadog-core/src/storage.js
@@ -2,20 +2,36 @@
 
 const { AsyncLocalStorage } = require('async_hooks')
 
+class DatadogStorage {
+  constructor () {
+    this._storage = new AsyncLocalStorage()
+  }
+
+  enterWith (store) {
+    const handle = {}
+    stores.set(handle, store)
+    this._storage.enterWith(handle)
+  }
+
+  getStore () {
+    const handle = this._storage.getStore()
+    return stores.get(handle)
+  }
+}
+
 const storages = Object.create(null)
-const legacyStorage = new AsyncLocalStorage()
+const legacyStorage = new DatadogStorage()
 
 const storage = function (namespace) {
   if (!storages[namespace]) {
-    storages[namespace] = new AsyncLocalStorage()
+    storages[namespace] = new DatadogStorage()
   }
   return storages[namespace]
 }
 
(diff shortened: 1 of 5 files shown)


packages/datadog-core/src/storage.js
```diff
@@ -2,12 +2,47 @@
...
+  constructor () {
+    this._storage = new AsyncLocalStorage()
+  }
+
+  disable () {
+    this._storage.disable()
+  }
+
+  enterWith (store) {
+    const handle = {}
```
simon-id, line 15: symbol instead ?

rochdev: Sure, but why?

rochdev: Since it's a pretty small change I'll switch to a symbol with the name provided to `storage()`, but I'm still curious as to the benefit in general compared to an object. Not that I disagree with the change, but would still like to know the rationale.

szegedi: I _generally_ prefer symbols too as they can be useful in debugging and other forms of runtime inspection. That said, in _this case_ these objects are really just unique keys for a map without any other meaning, and even mutating them doesn't particularly matter, as they'll keep being compared by identity.

rochdev: Made the change. I think the debugging argument is still valid in this case as it's possible to see the list of stores on the underlying async resource.

rochdev: Turns out this is not supported on Node 18, reverting the change.

simon-id: wow sad

simon-id (approved):

PR #4954, opened 2024-11-28
Protect req.socket.remoteAddress in appsec reporter

uurien:
### What does this PR do?

Protect `req.socket.remoteAddress` for the use cases when `req.socket` is empty.

### Motivation

Detected logs with the failure.

### DIFF
--- a/packages/dd-trace/src/appsec/reporter.js
+++ b/packages/dd-trace/src/appsec/reporter.js
@@ -148,7 +148,9 @@ function reportAttack (attackData) {
     newTags['_dd.appsec.json'] = '{"triggers":' + attackData + '}'
   }
 
-  newTags['network.client.ip'] = req.socket.remoteAddress
+  if (req.socket) {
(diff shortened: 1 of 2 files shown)


CarlesDD (dismissed):

CarlesDD:
Worth adding a test?

IlyasShabi (dismissed):
[NIT] Is it possible to have a test for this ?

uurien:
@simon-id 
> how does a req not have a socket??

I saw this happening in the logs, and the doc says that this could be null: https://nodejs.org/docs/latest/api/http.html#messagesocket

> unless the user specified a socket type other than [<net.Socket>](https://nodejs.org/docs/latest/api/net.html#class-netsocket) or internally nulled

simon-id (approved):
a bit weird cuz i can't find a place where the socket would get removed in node core, but i don't have time to investigate this more

CarlesDD (approved):

PR #4961, opened 2024-12-02
use gc observer for gc runtime metrics when available

rochdev:
### What does this PR do?

Use `gc` observer for GC runtime metrics when available.

### Motivation

Node provides this out of the box since Node 16, so we shouldn't use our own native implementation.

### DIFF
--- a/packages/dd-trace/src/runtime_metrics.js
+++ b/packages/dd-trace/src/runtime_metrics.js
@@ -14,10 +14,12 @@ const INTERVAL = 10 * 1000
 
 // Node >=16 has PerformanceObserver with `gc` type, but <16.7 had a critical bug.
 // See: https://github.com/nodejs/node/issues/39548
-const hasGCMetrics = NODE_MAJOR >= 18 || (NODE_MAJOR === 16 && NODE_MINOR >= 7)
+const hasGCObserver = NODE_MAJOR >= 18 || (NODE_MAJOR === 16 && NODE_MINOR >= 7)
+const hasGCProfiler = NODE_MAJOR >= 20 || (NODE_MAJOR === 18 && NODE_MINOR >= 15)
 
 let nativeMetrics = null
 let gcObserver = null
+let gcProfiler = null
 
 let interval
 let client
@@ -37,7 +39,7 @@ const runtimeMetrics = module.exports = {
     try {
       nativeMetrics = require('@datadog/native-metrics')
 
-      if (hasGCMetrics) {
+      if (hasGCObserver) {
         nativeMetrics.start('loop') // Only add event loop watcher and not GC.
       } else {
         nativeMetrics.start()
@@ -51,7 +53,8 @@ const runtimeMetrics = module.exports = {
 
     time = process.hrtime()
 
-    startObservers()
+    startGCObserver()
+    startGCProfiler()
 
     if (nativeMetrics) {
       interval = setInterval(() => {
@@ -77,8 +80,6 @@ const runtimeMetrics = module.exports = {
     if (nativeMetrics) {
       nativeMetrics.stop()
     }
-
-    stopObservers()
 
     clearInterval(interval)
     reset()
@@ -153,6 +154,10 @@ function reset () {
   counters = {}
   histograms = {}
   nativeMetrics = null
+  gcObserver && gcObserver.disconnect()
+  gcObserver = null
+  gcProfiler && gcProfiler.stop()
+  gcProfiler = null
 }
 
 function captureCpuUsage () {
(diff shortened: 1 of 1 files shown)


packages/dd-trace/src/runtime_metrics.js
```diff
@@ -202,6 +222,29 @@ function captureHeapSpace () {
     client.gauge('runtime.node.heap.physical_size.by.space', stats[i].physical_space_size, tags)
   }
 }
+function captureGCMetrics () {
+  if (!gcProfiler) return
+
+  const profile = gcProfiler.stop()
```
szegedi, line 228: Interesting. I'm currently using `PerformanceObserver` for gathering GC events for profiler timeline view, I'll look into whether it'd make sense to switch to `GCProfiler` too. Presumably we can have multiple instances of the `GCProfiler` running?

packages/dd-trace/src/runtime_metrics.js
```diff
@@ -7,11 +7,19 @@ const os = require('os')
...
-const { performance } = require('perf_hooks')
+const { performance, PerformanceObserver } = require('perf_hooks')
 
+const { NODE_MAJOR, NODE_MINOR } = require('../../../version')
 const INTERVAL = 10 * 1000
 
+// Node >=16 has PerformanceObserver with `gc` type, but <16.7 had a critical bug.
+// See: https://github.com/nodejs/node/issues/39548
+const hasGCObserver = NODE_MAJOR >= 18 || (NODE_MAJOR === 16 && NODE_MINOR >= 7)
+const hasGCProfiler = NODE_MAJOR >= 20 || (NODE_MAJOR === 18 && NODE_MINOR >= 15)
```
szegedi, line 18: Shouldn't we test for major versions being `>= 17` and `>= 19`, respectively? Otherwise we disable these for people that might run Node versions 17 and 19 (I know it is unlikely, but not impossible.) 19.9.0 is still offered for download.

szegedi (approved):
Looks good overall. Going to approve it like this, but happy to reapprove later as well if you want to fix that pedantic issue with version numbers.

packages/dd-trace/src/runtime_metrics.js
```diff
@@ -306,3 +355,57 @@ function histogram (name, stats, tags) {
...
+}
+
+function startGCProfiler () {
+  if (gcProfiler || !hasGCProfiler) return
+
+  gcProfiler = new v8.GCProfiler()
+  gcProfiler.start()
+}
+
+function gcType (kind) {
```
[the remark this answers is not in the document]
rochdev: I didn't know there was a way to get the names directly from Node, but if that's possible it definitely better than a manually maintained list.

packages/dd-trace/src/runtime_metrics.js
rochdev: Yes it should be possible to run multiple instances, although maybe we could share one somehow. Not sure how beneficial that would be though. The main advantage to GCProfiler, although I have not benchmarked this directly, is that it only reports the data once when you stop it instead of potentially hundreds or thousands of times per second under high load. However, it does buffer for longer than `PerformanceObserver`, but it does so on the native side. I added it to avoid a lot of back and forth from JS to native as it seemed like a better trade-off.

packages/dd-trace/src/runtime_metrics.js
rochdev: We do not support odd versions of Node except when it's the current version, and any fixes that may have landed in 18 and even backported to 16 would be missing from 17 for example, so in my opinion it's safer to just leave them out.

szegedi: Understood, makes sense.

packages/dd-trace/src/runtime_metrics.js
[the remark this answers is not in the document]
szegedi: Yeah, they're all listed in https://nodejs.org/api/perf_hooks.html#garbage-collection-gc-details and follow a stable naming pattern, so I just wrote code to go over them all.

packages/dd-trace/src/runtime_metrics.js
szegedi: Interesting – thank you, I'll keep this in mind.

packages/dd-trace/src/runtime_metrics.js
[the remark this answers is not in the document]
rochdev: We use different classification for runtime metrics, and changing that would be a breaking change, so the mapping would still be needed unfortunately. I will keep the current mapping as it matches what the native addon is doing, and anyway we're no longer using the native code after Node 18 so any changes won't affect us.

PR #4974, opened 2024-12-05
fix guardrail on node version outside of ssi

rochdev:
### What does this PR do?

Fix guardrail on Node version outside of SSI.

### Motivation

When a version of Node is not supported, we should bailout regardless of SSI as we can otherwise cause errors and segmentation faults. This is not something that should happen outside of a user error, but it can happen, for example attempting to upgrade dd-trace without upgrading Node and without reading the release notes.

### DIFF
--- a/.github/workflows/project.yml
+++ b/.github/workflows/project.yml
@@ -49,14 +49,15 @@ jobs:
       matrix:
         version: ['0.8', '0.10', '0.12', '4', '6', '8', '10', '12.0.0']
     runs-on: ubuntu-latest
-    env:
-      DD_INJECTION_ENABLED: 'true'
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.version }}
       - run: node ./init
+      - run: node ./init
+        env:
+          DD_INJECTION_ENABLED: 'true'
 
   integration-ci:
     strategy:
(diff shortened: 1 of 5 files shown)


.github/workflows/project.yml
```diff
@@ -49,14 +49,15 @@ jobs:
...
-      DD_INJECTION_ENABLED: 'true'
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.version }}
       - run: node ./init
+      - run: node ./init
+        env:
+          DD_INJECTION_ENABLED: 'true'
```
szegedi, line 60: `DD_INJECTION_ENABLED` is normally not a boolean but rather a comma-separated list of keywords that identify features enabled through auto-instrumentation. When SSI is in the effect, a typical value will be `DD_INJECTION_ENABLED=tracer`. When SSI is not used, the env var will simply be absent.

szegedi (changes requested):

.github/workflows/project.yml
rochdev: Yes, but being a string means that `true` is also a valid value. The idea here was to not specify a real value and instead just provide any string.

szegedi (approved):

bengl (approved):

PR #4980, opened 2024-12-09
Add support for endpoint_counts

szegedi:
### What does this PR do?
The profiling UI can show a `CPU - Average Time Per Call For Top Endpoints` if it is given an `endpoint_counts` JSON property in the uploaded `event.json`. We now add this property.

### Motivation
We want to get the Node.js functionality to match https://docs.datadoghq.com/profiler/guide/isolate-outliers-in-monolithic-services as much as possible.

### Additional Notes
This will also need a change in `logs_backend` before it starts showing up.

JIRA: [PROF-11013]

[PROF-11013]: https://datadoghq.atlassian.net/browse/PROF-11013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/src/profiling/exporters/event_serializer.js
+++ b/packages/dd-trace/src/profiling/exporters/event_serializer.js
@@ -1,0 +1,76 @@
+const os = require('os')
+const perf = require('perf_hooks').performance
+const version = require('../../../../../package.json').version
+
+class EventSerializer {
+  constructor ({ env, host, service, version, libraryInjected, activation } = {}) {
+    this._env = env
+    this._host = host
+    this._service = service
+    this._appVersion = version
+    this._libraryInjected = !!libraryInjected
+    this._activation = activation || 'unknown'
+  }
+
+  typeToFile (type) {
+    return `${type}.pprof`
+  }
+
+  getEventJSON ({ profiles, start, end, tags = {}, endpointCounts }) {
+    return JSON.stringify({
+      attachments: Object.keys(profiles).map(t => this.typeToFile(t)),
+      start: start && start.toISOString(),
+      end: end.toISOString(),
+      family: 'node',
+      version: '4',
+      tags_profiler: [
+        'language:javascript',
+        'runtime:nodejs',
+        `runtime_arch:${process.arch}`,
+        `runtime_os:${process.platform}`,
+        `runtime_version:${process.version}`,
+        `process_id:${process.pid}`,
+        `profiler_version:${version}`,
+        'format:pprof',
+        ...Object.entries(tags).map(([key, value]) => `${key}:${value}`)
+      ].join(','),
+      endpoint_counts: endpointCounts,
+      info: {
+        application: {
+          env: this._env,
+          service: this._service,
+          start_time: new Date(perf.nodeTiming.nodeStart + perf.timeOrigin).toISOString(),
+          version: this._appVersion
+        },
+        platform: {
+          hostname: this._host,
+          kernel_name: os.type(),
+          kernel_release: os.release(),
+          kernel_version: os.version()
+        },
+        profiler: {
+          activation: this._activation,
+          ssi: {
+            mechanism: this._libraryInjected ? 'injected_agent' : 'none'
+          },
+          version
+        },
+        runtime: {
+          // Using `nodejs` for consistency with the existing `runtime` tag.
+          // Note that the event `family` property uses `node`, as that's what's
+          // proscribed by the Intake API, but that's an internal enum and is
+          // not customer visible.
+          engine: 'nodejs',
+          // strip off leading 'v'. This makes the format consistent with other
+          // runtimes (e.g. Ruby) but not with the existing `runtime_version` tag.
+          // We'll keep it like this as we want cross-engine consistency. We
+          // also aren't changing the format of the existing tag as we don't want
+          // to break it.
+          version: process.version.substring(1)
+        }
+      }
+    })
+  }
+}
+
+module.exports = { EventSerializer }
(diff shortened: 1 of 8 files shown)


packages/dd-trace/src/profiling/exporters/event_serializer.js
nsavoire, line 1: Nice refactoring !

packages/dd-trace/src/profiling/exporters/event_serializer.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  typeToFile (type) {
+    return `${type}.pprof`
+  }
+
+  getEventJSON ({ profiles, start, end, tags = {}, endpointCounts }) {
+    return JSON.stringify({
+      attachments: Object.keys(profiles).map(t => this.typeToFile(t)),
+      start: start && start.toISOString(),
```
nsavoire, line 22: This is strange to check for `start` being defined.
I guess that's for the file exporter case, but maybe we should also ensure to always pass a `start` to file export.

packages/dd-trace/src/profiling/profiler.js
```diff
@@ -82,6 +86,9 @@ class Profiler extends EventEmitter {
         this._logger.debug(`Started ${profiler.type} profiler in ${threadNamePrefix} thread`)
       }
 
+      this._spanFinishListener = this._onSpanFinish.bind(this)
```
nsavoire, line 89: Should we enable endpoint count when endpoint collection is enabled ?
Also the similiraties of the code make make wonder if this does not belong to wall profiler.

packages/dd-trace/src/profiling/profiler.js
```diff
@@ -137,6 +147,25 @@ class Profiler extends EventEmitter {
     }
   }
 
+  _onSpanFinish (span) {
+    const tags = span.context()._tags
+    if (tags[SPAN_TYPE] !== WEB) return
+
+    const endpointName = tags[RESOURCE_NAME] || [
```
nsavoire, line 154: Maybe reuse code from wall profiler ?

packages/dd-trace/src/profiling/profiler.js
```diff
@@ -137,6 +147,25 @@ class Profiler extends EventEmitter {
...
+      tags[HTTP_METHOD],
+      tags[HTTP_ROUTE]
+    ].filter(v => v).join(' ')
+    if (!endpointName) return
+
+    let counter = this.endpointCounts.get(endpointName)
+    if (counter === undefined) {
+      counter = { count: 1 }
```
nsavoire, line 162: The use of an object with a `count` field as value instead of directly a number is to allow for extensibility ?

packages/dd-trace/src/profiling/profiler.js
```diff
@@ -137,6 +147,25 @@ class Profiler extends EventEmitter {
...
+    const tags = span.context()._tags
+    if (tags[SPAN_TYPE] !== WEB) return
+
+    const endpointName = tags[RESOURCE_NAME] || [
+      tags[HTTP_METHOD],
+      tags[HTTP_ROUTE]
+    ].filter(v => v).join(' ')
+    if (!endpointName) return
+
+    let counter = this.endpointCounts.get(endpointName)
```
nsavoire, line 160: Are we guarenteed that a web span cannot have another web span with the same endpoint as child ?

nsavoire (dismissed):
LGTM, just a few questions / nitpicks !

packages/dd-trace/src/profiling/profiler.js
szegedi: No, I just didn't want to always have to call `set` after a `get` on the Map.
```
const count = this.endpointCounts.get(endpointName) || 0
this.endpointCounts.set(endpointName, count++)
```
Of course, if I used an object instead of a Map, I wouldn't have to do this.

packages/dd-trace/src/profiling/profiler.js
szegedi: Yeah, there are similarities, but then I'd also need to somehow communicate the data from the wall profiler to here, maybe by generalizing the return value from "profile ()". I was thinking about it and concluded that endpoint counting can be a general higher-level profiler concern – it's unlikely any particular profile provider would return it differently. You are right that this should be gated by whether endpoint collection is enabled. I might also extract the web span utilities into a separate file so we don't repeat ourselves.

packages/dd-trace/src/profiling/profiler.js
szegedi: Yep, will do.

packages/dd-trace/src/profiling/profiler.js
szegedi: I added some code to check if this is the outermost web span

packages/dd-trace/src/profiling/exporters/event_serializer.js
szegedi: yeah, it's for the file exporter. Fixed.

packages/dd-trace/src/profiling/profiler.js
szegedi: Okay, extracted commonalities into a separate file.

packages/dd-trace/src/profiling/profiler.js
nsavoire: Ok, I get it ;)
Get returns you a reference that you can update contrary to a number.

packages/dd-trace/src/profiling/profiler.js
nsavoire: I did a quick experiment with a simple nextjs app and indeed we get to 2 web spans with same endpoint per request, probably one from http plugin and one from nextjs plugin.

nsavoire (approved):

packages/dd-trace/src/profiling/profiler.js
szegedi: Thanks for checking this!

nsavoire: I remembered seeing some weird things with nextjs when I added endpoint code 😄 
I read the conversation about endpoint determination on slack, and I agree we need to come up with a cleaner way to gather the information in the tracer.

PR #4983, opened 2024-12-09
Fix ESM instrumentation when importing only methods

uurien:
### What does this PR do?

Instrument `default` and each exported method in esm modules.

### Motivation

`import-in-the-middle` exports a `default` object and a method/property by each exported method in the original library. If `default` is present, we are instrumenting only the default object, and the methods are not instrumented.

### DIFF
--- a/packages/datadog-instrumentations/src/child_process.js
+++ b/packages/datadog-instrumentations/src/child_process.js
@@ -17,7 +17,7 @@ const execAsyncMethods = ['execFile', 'spawn']
 const names = ['child_process', 'node:child_process']
 
 // child_process and node:child_process returns the same object instance, we only want to add hooks once
-let patched = false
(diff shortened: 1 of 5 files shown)


packages/datadog-instrumentations/src/child_process.js
```diff
@@ -17,7 +17,7 @@ const execAsyncMethods = ['execFile', 'spawn']
 const names = ['child_process', 'node:child_process']
 
 // child_process and node:child_process returns the same object instance, we only want to add hooks once
-let patched = false
+const patched = new WeakSet()
```
uurien, line 20: addHook could be executed twice with different object as `childProcess`, in that case, we want to instrument both.

```
import { method } from 'child-process'
// and
import childProcess from 'child-process'
```

PR #4991, opened 2024-12-11
[DI] Handle async errors in mocha tests

watson:
If an async error is thrown in mocha tests, mocha doesn't see it. Best case, the test will just time out, worst case, it will pass.

In our case, there actually was an error in one of the integration tests, which has gone unnoticed (see 2nd commit).

## Example

The following Mocha will never fail even if `result` isn't `42`:

```js
it('should do something', function (done) {
  doSomething((result) => {
    assert.equal(result, 42)
    done()
  })
})
```

The reason is that `assert.equal` throws asynchronously because it's inside the `doSomething` callback, and hence mocha doesn't catch it. I'm aware that in this case, Mocha will instead time out because `done` isn't called, but there's situations where your test-code contains multiple calls to `done` and so, the next call to `done` will end the test and it will not time out (as is evident with the 2nd commit in this PR).

In any case, the normal way to "fix" this is:

```js
it('should do something', function (done) {
  doSomething((result) => {
    try {
      assert.equal(result, 42)
    } catch (err) {
      return done(err)
    }
    done()
  })
})
```

However, this quickly becomes tedious, if your Mocha tests has a lot of these callbacks. With the addition of the `failOnException` helper function in this PR, the above code can instead be written like so:

```js
it('should do something', function (done) {
  doSomething(failOnException(done, (result) => {
    assert.equal(result, 42)
    done()
  }))
})
```

### DIFF
--- a/integration-tests/debugger/basic.spec.js
+++ b/integration-tests/debugger/basic.spec.js
@@ -201,7 +201,7 @@ describe('Dynamic Instrumentation', function () {
         }, {
           ddsource: 'dd_debugger',
           service: 'node',
-          debugger: { diagnostics: customErrorDiagnosticsObj ?? { probeId, version: 0, status: 'ERROR' } }
+          debugger: { diagnostics: customErrorDiagnosticsObj ?? { probeId, probeVersion: 0, status: 'ERROR' } }
         }]
 
         t.agent.on('debugger-diagnostics', failOnException(done, ({ payload }) => {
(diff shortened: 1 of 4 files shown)


watson:
* **#4991** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4991?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/4991?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`

<h2></h2>

juan-fernandez (approved):

PR #5001, opened 2024-12-12
telemetry: increment .count when deduping telemetry logs

tlhunter:
### What does this PR do?
- we currently deduplicate telemetry logs
- however we don't track when this deduplicaiton happens
- this change modifies telemetry logs to have a .count field which increments each time this happens

### Motivation
- AIDM-391
- this way we'll know if a message happens once or a thousand times

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/telemetry/logs/log-collector.js
+++ b/packages/dd-trace/src/telemetry/logs/log-collector.js
@@ -3,7 +3,7 @@
 const log = require('../../log')
 const { calculateDDBasePath } = require('../../util')
 
-const logs = new Map()
+const logs = new Map() // hash -> log
 
 // NOTE: Is this a reasonable number?
 let maxEntries = 10000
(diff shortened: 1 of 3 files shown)


bengl (approved):

packages/dd-trace/src/telemetry/logs/log-collector.js
```diff
@@ -73,8 +73,10 @@ const logCollector = {
       }
       const hash = createHash(logEntry)
       if (!logs.has(hash)) {
-        logs.set(hash, logEntry)
+        logs.set(hash, errorCopy(logEntry))
```
iunanua, line 76: For the record: `errorCopy` function is OK but it is not needed :)

`logCollector` expects a `logEntry` not an `Error`.

 `logEntry` is an object that meets telemetry logs payload schema: `message`, `level`, `count`, `tags`, `stack_trace` and `tracer_time` fields. So you could simply add here `logEntry.count = 1` instead calling `errorCopy(logEntry)`

tlhunter: Ah! I had done it that way to prevent issues where a userland error would then get mutated to have the .count property. Is that something that can never happen?

PR #5007, opened 2024-12-12
[DI] Batch outgoing http requests

watson:
### What does this PR do?

Adds a queue in front of the outgoing http requests from the Dynamic Instrumentation devtools worker thread.

Two queues are added:

- One for requests to `/debugger/v1/diagnostics`
- One for requests to `/debugger/v1/input`

Both queues are configured in the same manner.

When adding a payload to any of the queues, it's buffered until either a 1 second timer ends or the size of the queue exceeds 5MB after which the queue is flushed. Once flushed, all payloads waiting in the queue are sent to the HTTP endpoint in a single HTTP request.

### Motivation

Without this queue the tracer could produce up to 5.000 request per second. This would overload the tracer.

### Note to reviewers

Hide whitespace changes for a better diff

### DIFF
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -35,6 +35,10 @@ jobs:
             range: '>=5.12.1'
             aerospike-image: ce-6.4.0.3
             test-image: ubuntu-latest
+          - node-version: 22
+            range: '>=6.0.0'
+            aerospike-image: ce-6.4.0.3
+            test-image: ubuntu-latest
     runs-on: ${{ matrix.test-image }}
     services:
       aerospike:
--- a/integration-tests/debugger/snapshot.spec.js
+++ b/integration-tests/debugger/snapshot.spec.js
@@ -16,10 +16,10 @@ describe('Dynamic Instrumentation', function () {
           assert.deepEqual(Object.keys(captures.lines), [String(t.breakpoint.line)])
 
           const { locals } = captures.lines[t.breakpoint.line]
-          const { request, fastify, getSomeData } = locals
+          const { request, fastify, getUndefined } = locals
           delete locals.request
           delete locals.fastify
-          delete locals.getSomeData
+          delete locals.getUndefined
 
           // from block scope
           assert.deepEqual(locals, {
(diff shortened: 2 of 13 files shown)


packages/dd-trace/src/debugger/devtools_client/queue.js
```diff
@@ -0,0 +1,36 @@
...
+    this._partialJson = null
+  }
+
+  _flush () {
+    const json = `${this._partialJson}]`
+    this._reset()
+    this._onFlush(json)
+  }
+
+  add (str, size = Buffer.byteLength(str)) {
```
shatzi, line 23: I think it better to pass just the json and remove the size param. 
If the user give wrong values it break the logic of the queue

packages/dd-trace/src/debugger/devtools_client/queue.js
```diff
@@ -0,0 +1,36 @@
+'use strict'
+
+class JSONQueue {
```
shatzi, line 3: nit: maybe JSONBuffer is better as this is not really a queue API

shatzi (approved):
This is great. thank you.

currently we only creating snapshots probes. but once we have log probes we should have another take on this to handle chatty probes. Would be great if we can unify the spec for that before we implement.

packages/dd-trace/src/debugger/devtools_client/queue.js
watson: This is purely a performance optimization. In the snapshot code, the size is already calculated since we need to know if it's larger than 1MB. So to not calculate it twice we allow it to be passed in. It's not user generated input.

packages/dd-trace/src/debugger/devtools_client/queue.js
watson: Yeah I couldn't decide if I should call it a buffer or a queue... was considering both names. In Node.js there's a `Buffer` class which is different, so the term is a bit overloaded in Node.js. But I'll consider changing it... you're probably right

watson: I've renamed it from `Queue` to `JSONBuffer` now 👍

shatzi (approved):

szegedi (approved):
Looks good to me; nice solution!

PR #5012, opened 2024-12-13
ESM support for iast

uurien:
### What does this PR do?

With application is configured with support for ESM (using --loader or initialize.mjs) and IAST is enabled, the app files are rewritten in the same way that they are rewritten when they are required.

### Motivation

Support ESM for IAST taint tracking.

### Plugin Checklist

- [x] Unit tests.

### Additional Notes

This will work only when `Module.register` is supported: >= 20.8.0 or ^18.19.0

[APPSEC-11290]

[APPSEC-11290]: https://datadoghq.atlassian.net/browse/APPSEC-11290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
```diff
@@ -114,11 +138,70 @@ function enableRewriter (telemetryVerbosity) {
...
+    }
+
+    try {
+      Module.register('./rewriter-esm.mjs', {
+        parentURL: pathToFileURL(__filename),
+        transferList: [port2],
+        data
+      })
+    } catch (e) {
+      log.error('[ASM] Error enabling ESM Rewriter', e)
```
iunanua, line 200: should we call `port1.off`?

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
```diff
@@ -114,11 +138,70 @@ function enableRewriter (telemetryVerbosity) {
...
+function isEsmConfigured () {
+  const hasLoaderArg = !!process.execArgv?.find(
+    arg => arg.includes('--loader') || arg.includes('--experimental-loader')
+  )
+  if (hasLoaderArg) return true
+
+  const nodeOptions = process.env.NODE_OPTIONS
+  const hasLoaderNodeOptions = nodeOptions &&
+    (nodeOptions.includes('--loader') || nodeOptions.includes('--experimental-loader'))
+  if (hasLoaderNodeOptions) return true
```
iunanua, line 157: could it be `isFlagPresent('--loader') || isFlagPresent('--experimental-loader')`?

uurien: Yes, and also I will fix `isFlagPresent` to check it in `NODE_OPTIONS` that is not doing

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
uurien: Good point, I checked the api, and I think that is better if we call to `port1.close()` and `port2.close()` methods. changed.

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
uurien: My fault, it is already checking `NODE_OPTIONS` 🤦 
Changed.

packages/dd-trace/src/appsec/iast/taint-tracking/constants.js
```diff
@@ -0,0 +1,6 @@
+'use strict'
+
+module.exports = {
+  LOG: 'LOG',
+  REWRITTEN: 'REWRITTEN'
```
iunanua, line 5: these constants could be more descriptive:
```suggestion
  LOG_MESSAGE: 'LOG_MESSAGE',
  REWRITTEN_MESSAGE: 'REWRITTEN_MESSAGE'
```

iunanua (approved):

PR #5021, opened 2024-12-16
Enable 128 bit trace id log injection by default

khanayan123:
### What does this PR do?
Enables 128 bit trace id log injection by default & refactors code to change 128 bit trace id generation to default.

Changes:

- Both features (128Bit Generation and 128Bit ids in logs) are true by default
- If 128Bit Generation is set to false, then 128Bit ids in logs follows suit and evaluates to false. However, it can be enabled independently if desired.

Other edge case behaviors will stay the same, such as:
- If 128Bit Generation is set to false and 128Bit ids in logs is set to true, only the uint 64bit trace ID will be logged, unless we're dealing with an edge case where we've inherited a trace context with a 128bit trace ID.

### Motivation
Standardizing configuration across SDKs to a unified and consistent definition


rochdev:
@bantonsson I remember this being a breaking change for log injection, do you know if there have been any new developments on that front that would make this non-breaking?

bantonsson (commented):
@rochdev So the breaking part that going to 128 bits can have is the observed behavior from the customer perspective. If the customer is shipping logs somewhere else than Datadog, and the format of the `TraceID` changes then there could be breakage. The Datadog log ingestion and correlation handles both 64 and 128 bit ids as far as I know.

khanayan123:
> @rochdev So the breaking part that going to 128 bits can have is the observed behavior from the customer perspective. If the customer is shipping logs somewhere else than Datadog, and the format of the `TraceID` changes then there could be breakage. The Datadog log ingestion and correlation handles both 64 and 128 bit ids as far as I know.

@mabdinur should we classify this as a breaking change on the jira card?

mabdinur:
> > @rochdev So the breaking part that going to 128 bits can have is the observed behavior from the customer perspective. If the customer is shipping logs somewhere else than Datadog, and the format of the `TraceID` changes then there could be breakage. The Datadog log ingestion and correlation handles both 64 and 128 bit ids as far as I know.
> 
> @mabdinur should we classify this as a breaking change on the jira card?

Officially we only support log correlation within the Datadog Ecosystem. As long as this change does not break any existing Datadog functionality I don't think it's a breaking change. We do not officially support customer's manually extracting trace data from logs and doing their own correlations. I think this out of the scope of our version policy. 

I think calling out this breakage in a release note is good enough. Requiring this change to go out in a major release will significantly impact the roll out of this enhancement/fix to Datadog users.

rochdev (approved):

PR #5023, opened 2024-12-17
repo: mandatory issue templates (AIDM-422)

tlhunter:
### What does this PR do?
- makes it mandatory to create issues using a template
- creates a bug report template
- creates a feature request template

### Motivation
- creating a consistent issue reporting experience across all tracers
- here's a screenshot of what the create issue dialog will now look like

Screenshot 2024-12-16 at 16 03 42

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -1,0 +1,64 @@
+name: "Bug Report (Low Priority)"
+description: "Create a public Bug Report. Note that these may not be addressed as quickly as the helpdesk and that looking up account information will be difficult."
+title: "[BUG]: "
+labels: bug
+body:
+  - type: input
+    attributes:
+      label: Tracer Version(s)
+      description: "Version(s) of the tracer affected by this bug"
+      placeholder: 1.2.3, 4.5.6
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,64 @@
+name: "Bug Report (Low Priority)"
+description: "Create a public Bug Report. Note that these may not be addressed as quickly as the helpdesk and that looking up account information will be difficult."
```
uurien, line 2: Is this text visible when the bug report is being created? Looks like there is a `markdown` type for body to show data when someone is creating an issue:

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#markdown

```yaml
body:
  - type: markdown
    attributes:
      value: "Create a public Bug Report. Note that these may not be addressed as quickly as the helpdesk and that looking up account information will be difficult."
```

uurien: Oh, I didn't see the attached screenshot before. LGTM, resolving this and approving the PR.

uurien (approved):

PR #5033, opened 2024-12-18
ignore engines for package size job

rochdev:
### What does this PR do?

Ignore engines for package size job.

### Motivation

It fails on Node 16 otherwise because the new eslint dependencies don't support it.

### DIFF
--- a/.github/workflows/package-size.yml
+++ b/.github/workflows/package-size.yml
@@ -20,7 +20,7 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: '18'
-      - run: yarn
+      - run: yarn --ignore-engines
       - name: Compute module size tree and report
         uses: qard/heaviest-objects-in-the-universe@v1
         with:


.github/workflows/package-size.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: '18'
-      - run: yarn
+      - run: yarn --ignore-engines
```
uurien, line 23: I the PR description you've said:

> It fails on Node 16 otherwise because the new eslint dependencies don't support it.

But this ci job is running only with `node-version: '18'` when is this job failing?

rochdev: On the main branch yes, but on v4.x it runs on Node 16. I guess there is no reason not to use Node 18 everywhere for this specifically, so I could change the version instead of that makes more sense, although not sure how since it's already changed in a larger commit that is now landed.

rochdev: I guess I could bump it further to for example 20+

rochdev:
Closing in favour of #5040

PR #5044, opened 2024-12-19
env var to disable all middleware spans

bengl:
### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/datadog-plugin-express/test/integration-test/client.spec.js
```diff
@@ -36,6 +36,7 @@ describe('esm', () => {
     })
 
     it('is instrumented', async () => {
+      process.env.DD_TRACE_MIDDLEWARE_ENABLED = true
```
bengl, line 39:
```suggestion
      delete process.env.DD_TRACE_MIDDLEWARE_ENABLED
```

packages/dd-trace/src/config.js
```diff
@@ -520,6 +520,7 @@ class Config {
     this._setValue(defaults, 'lookup', undefined)
     this._setValue(defaults, 'inferredProxyServicesEnabled', false)
     this._setValue(defaults, 'memcachedCommandEnabled', false)
+    this._setValue(defaults, 'middleware', true)
```
tlhunter, line 523: This name seems too simple and easy to confuse the purpose. How about `enableMiddlewareTracing`?

bengl: I prefer this name, yes.

tlhunter (approved):

tlhunter (dismissed):

tlhunter (approved):

tlhunter (dismissed):

packages/datadog-plugin-express/test/integration-test/client.spec.js
```diff
@@ -36,6 +36,7 @@ describe('esm', () => {
     })
 
     it('is instrumented', async () => {
+      delete process.env.DD_TRACE_MIDDLEWARE_ENABLED
```
rochdev, line 39: Altering environment variables should be done in `beforeEach` and `afterEach` to make sure that the previous value is restored.

packages/dd-trace/src/config.js
```diff
@@ -674,6 +675,7 @@ class Config {
       DD_TRACE_HEADER_TAGS,
       DD_TRACE_LEGACY_BAGGAGE_ENABLED,
       DD_TRACE_MEMCACHED_COMMAND_ENABLED,
+      DD_TRACE_MIDDLEWARE_ENABLED,
```
rochdev, line 678: The environment variable doesn't match the name of the programmatic option.

packages/dd-trace/src/plugins/util/web.js
```diff
@@ -570,11 +572,16 @@ function getHooks (config) {
...
-function getMiddlewareSetting (config) {
-  if (config && typeof config.middleware === 'boolean') {
-    return config.middleware
-  } else if (config && config.hasOwnProperty('middleware')) {
-    log.error('Expected `middleware` to be a boolean.')
+function getMiddlewareSetting (config, tracerConfig) {
+  if (config) {
+    if (typeof config.middleware === 'boolean') {
+      return config.middleware
+    } else if (config.hasOwnProperty('middlewareTracingEnabled')) {
```
rochdev, line 579: This should be done in the plugin manager, consistent with other similar options. While it could be argued that this is not ideal, changing how this works is out of scope of this PR.

packages/dd-trace/src/config.js
rochdev: It also doesn't have a corresponding test.

rochdev (changes requested):

packages/dd-trace/src/config.js
bengl: > The environment variable doesn't match the name of the programmatic option.

Please provide suggested values for both. @tlhunter @rochdev please coordinate on this because right now you folks have both made naming suggestions.

> It also doesn't have a corresponding test.

I don't understand this comment. In this same review, you commented on the corresponding test. What additional testing are you looking for?

bengl: > The environment variable doesn't match the name of the programmatic option.

Please provide suggested values for both. @tlhunter @rochdev please coordinate on this because right now you folks have both made naming suggestions.

> It also doesn't have a corresponding test.

I don't understand this comment. In this same review, you commented on the corresponding test. What additional testing are you looking for?

rochdev: > Please provide suggested values for both.

I don't mind the name, but it should map between programmatic and env var.

> I don't understand this comment. In this same review, you commented on the corresponding test.

The plugin test yes, but there isn't a test for the option itself (config.spec.js)

bengl: >I don't mind the name, but it should map between programmatic and env var.

So renaming the env var to `DD_TRACE_MIDDLEWARE_TRACING_ENABLED` would be sufficient? That's a lot of `TRACE` in the name, but I guess that's fine.

rochdev: > That's a lot of TRACE in the name, but I guess that's fine.

We did name the library `dd-trace` after all 😅

rochdev (approved):

PR #5061, opened 2024-12-31
refactor custom metrics to be self-contained

bengl:
This had no real business being in `Proxy#init()`, so I moved it to where it makes more sense.

### DIFF
--- a/packages/dd-trace/src/dogstatsd.js
+++ b/packages/dd-trace/src/dogstatsd.js
@@ -190,6 +190,14 @@ class CustomMetrics {
   constructor (config) {
     const clientConfig = DogStatsDClient.generateClientConfig(config)
     this.dogstatsd = new DogStatsDClient(clientConfig)
+
+    this._boundFlush = this.flush.bind(this)
+
+    // TODO(bengl) this magic number should be configurable
+    this._flushInterval = setInterval(this._boundFlush, 10 * 1000)
+    this._flushInterval.unref()
+
+    process.once('beforeExit', this._boundFlush)
   }
 
   increment (stat, value = 1, tags) {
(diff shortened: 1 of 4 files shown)


packages/dd-trace/src/dogstatsd.js
```diff
@@ -190,6 +190,14 @@ class CustomMetrics {
   constructor (config) {
     const clientConfig = DogStatsDClient.generateClientConfig(config)
     this.dogstatsd = new DogStatsDClient(clientConfig)
+
+    this._boundFlush = this.flush.bind(this)
```
uurien, line 194: Why are `_boundFlush` and `_flushInterval` object level properties, and not just `const boundFlush` and `const flushInterval`?

packages/dd-trace/src/dogstatsd.js
```diff
@@ -190,6 +190,14 @@ class CustomMetrics {
...
+
+    // TODO(bengl) this magic number should be configurable
```
rochdev, line 196: This must match the timing of the agent, so it could be a footgun to make it configurable.

BridgeAR: I guess the comment is not really blocking this from landing?

packages/dd-trace/src/dogstatsd.js
```diff
@@ -190,6 +190,14 @@ class CustomMetrics {
...
+    this._flushInterval = setInterval(this._boundFlush, 10 * 1000)
+    this._flushInterval.unref()
+
+    process.once('beforeExit', this._boundFlush)
```
BridgeAR, line 200:
```suggestion
    const flush = this.flush.bind(this)

    // TODO(bengl) this magic number should be configurable
    setInterval(flush, 10 * 1000).unref()

    process.once('beforeExit', flush)
```

BridgeAR (approved):
LGTM % comment. I would definitely prefer not to add any properties that are not absolutely needed

PR #5070, opened 2025-01-06
Send tags, including git metadata, to RC endpoint

watson:
### What does this PR do?

Adds DD tags to the JSON data sent to the `/v0.7/config` endpoint used by Remote Config. The tags include git metadata: Repository and commit SHA of the service being instrumented.

### Motivation

The git metadata is needed by DI to show the source code of a running service in the UI if source code integration is enabled.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/appsec/remote_config/manager.js
+++ b/packages/dd-trace/src/appsec/remote_config/manager.js
@@ -9,6 +9,7 @@ const log = require('../../log')
 const { getExtraServices } = require('../../service-naming/extra-services')
 const { UNACKNOWLEDGED, ACKNOWLEDGED, ERROR } = require('./apply_states')
 const Scheduler = require('./scheduler')
+const { GIT_REPOSITORY_URL, GIT_COMMIT_SHA } = require('../../plugins/util/tags')
 
 const clientId = uuid()
 
@@ -32,6 +33,14 @@ class RemoteConfigManager extends EventEmitter {
       hostname: config.hostname || 'localhost',
       port: config.port
     }))
+
+    const tags = config.repositoryUrl
+      ? {
+          ...config.tags,
+          [GIT_REPOSITORY_URL]: config.repositoryUrl,
+          [GIT_COMMIT_SHA]: config.commitSHA
+        }
+      : config.tags
 
     this._handlers = new Map()
     const appliedConfigs = this.appliedConfigs = new Map()
@@ -67,7 +76,8 @@ class RemoteConfigManager extends EventEmitter {
           service: config.service,
           env: config.env,
           app_version: config.version,
-          extra_services: []
+          extra_services: [],
+          tags: Object.entries(tags).map((pair) => pair.join(':'))
         },
         capabilities: DEFAULT_CAPABILITY // updated by `updateCapabilities()`
       },
--- a/packages/dd-trace/test/appsec/remote_config/manager.spec.js
+++ b/packages/dd-trace/test/appsec/remote_config/manager.spec.js
@@ -98,7 +98,8 @@ describe('RemoteConfigManager', () => {
           service: config.service,
           env: config.env,
           app_version: config.version,
-          extra_services: []
+          extra_services: [],
+          tags: ['runtime-id:runtimeId']
         },
         capabilities: 'AA=='
       },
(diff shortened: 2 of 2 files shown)


watson:
* **#5070** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5070?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5070?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

shatzi (dismissed):

packages/dd-trace/src/appsec/remote_config/manager.js
```diff
@@ -33,6 +34,10 @@ class RemoteConfigManager extends EventEmitter {
       port: config.port
     }))
 
+    const tags = { ...config.tags }
+    tags[GIT_REPOSITORY_URL] = config.repositoryUrl
+    tags[GIT_COMMIT_SHA] = config.commitSHA
```
simon-id, line 39: less HiddenClass created ?
```suggestion
    const tags = {
      ...config.tags,
      [GIT_REPOSITORY_URL]: config.repositoryUrl,
      [GIT_COMMIT_SHA]: config.commitSHA
    }
```

packages/dd-trace/test/appsec/remote_config/manager.spec.js
simon-id, line 1: would be nice to have a test that actually check this works ?

packages/dd-trace/src/appsec/remote_config/manager.js
```diff
@@ -67,7 +72,8 @@ class RemoteConfigManager extends EventEmitter {
           service: config.service,
           env: config.env,
           app_version: config.version,
-          extra_services: []
+          extra_services: [],
+          tags: Object.entries(tags).map((pair) => pair.join(':')).join(',')
```
simon-id, line 76: no escaping uh ? sounds like a tag with a colon or a comma will break the parsing in the backend one day ?

packages/dd-trace/test/appsec/remote_config/manager.spec.js
```diff
@@ -98,7 +98,8 @@ describe('RemoteConfigManager', () => {
           service: config.service,
           env: config.env,
           app_version: config.version,
-          extra_services: []
+          extra_services: [],
+          tags: 'runtime-id:runtimeId,git.repository_url:,git.commit.sha:'
```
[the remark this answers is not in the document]
watson: They don't need to, but I would say that it shouldn't matter. I did it this way thinking the logic to add them would be simpler. I could of course have a filter to remove them if the value was falsy. Do you think that's worth it?

packages/dd-trace/src/appsec/remote_config/manager.js
watson: Fixed 👍

packages/dd-trace/test/appsec/remote_config/manager.spec.js
watson: Fixed 👍

packages/dd-trace/src/appsec/remote_config/manager.js
watson: I took a second look at where I copied this logic from (the Python tracer) and I actually copied it incorrectly. I've pushed a fix now. The right encoding seems to be:

Original
```js
{ a: 1, b: 2 }
```

Encoded:
```js
['a:1', 'b:2']
```

This is still not optimal as a colon in either the tag name of value might mess things up. But it's better. I'm not sure why we try to invent our own encoding when JSON could have encoded this for us without a problem 🤷

simon-id: Well, Nuremberg defense it is then :shrug: i'll resolve the comment

packages/dd-trace/src/appsec/remote_config/manager.js
```diff
@@ -33,6 +34,14 @@ class RemoteConfigManager extends EventEmitter {
       port: config.port
     }))
 
+    const tags = config.repositoryUrl
+      ? {
+          ...config.tags,
+          [GIT_REPOSITORY_URL]: config.repositoryUrl,
+          [GIT_COMMIT_SHA]: config.commitSHA
```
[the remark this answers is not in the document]
watson: Shouldn't `config.repositoryUrl` etc always take precedence over `tags` in this case? What would be the scenario where git metadata was in the tags and not in the `config` object?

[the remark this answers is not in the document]
simon-id: nvm i was mistaken

simon-id (approved):

PR #5072, opened 2025-01-07
[test optimization] [SDTEST-1355] Fix DI issues with auto test retries

juan-fernandez:
### What does this PR do?

Fix different issues with the integration between Dynamic Instrumentation and Test Optimization.

### Motivation

* Line probes were not cleaned up after the test finished. This caused an issue where we tried to add a line probe that was already there.
* We did not wait for hit probes to return their information to the main process. This caused a race condition where we assigned the snapshot to the wrong test. 
* We were working under the assumption that the line probe would always be located at the top of the test error stack trace. This is not true since the user code might be at a different place.

Unrelated to this PR: I fixed the way we reported `TestingLibraryElementError`. Instead of showing the error message in the stack, we clean up the stack. 

### Plugin Checklist

- [x] Unit tests.

### DIFF
--- a/packages/dd-trace/src/ci-visibility/dynamic-instrumentation/index.js
+++ b/packages/dd-trace/src/ci-visibility/dynamic-instrumentation/index.js
@@ -21,7 +21,6 @@ class TestVisDynamicInstrumentation {
   }
 
   removeProbe (probeId) {
-    // breakpointRemoveChannel
     return new Promise(resolve => {
       this.breakpointRemoveChannel.port2.postMessage(probeId)
 
@@ -29,10 +28,9 @@ class TestVisDynamicInstrumentation {
     })
   }
 
-  // Return 3 elements:
-  // 1. Snapshot ID
+  // Return 2 elements:
+  // 1. Probe ID
   // 2. Promise that's resolved when the breakpoint is set
-  // 3. Promise that's resolved when the breakpoint is hit
   addLineProbe ({ file, line }, onHitBreakpoint) {
     const probeId = randomUUID()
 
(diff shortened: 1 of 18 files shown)


nikita-tkachenko-datadog (approved):

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/index.js
```diff
@@ -101,15 +109,19 @@ class TestVisDynamicInstrumentation {
...
-        resolve({ snapshot })
-        probeIdToResolveBreakpointHit.delete(probeId)
+      const onHit = this.onHitBreakpointByProbeId.get(probeId)
+      if (onHit) {
+        onHit({ snapshot })
       }
     }).unref()
 
-    this.worker.on('error', (err) => log.error('ci-visibility DI worker error', err))
-    this.worker.on('messageerror', (err) => log.error('ci-visibility DI worker messageerror', err))
```
juan-fernandez, line 112: this was already handled above

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
```diff
@@ -81,7 +105,7 @@ async function addBreakpoint (snapshotId, probe) {
     try {
       lineNumber = await processScriptWithInlineSourceMap({ file, line, sourceMapURL })
     } catch (err) {
-      log.error(err)
+      log.error('Error processing script with inline source map')
```
watson, line 108: Are you sure you don't want to report the original error as well? Ok if you don't, I just wanted to make sure it wasn't an oversight 😊

```suggestion
      log.error('Error processing script with inline source map', err)
```

watson (approved):
I reviewed changes to the following files only:
- `packages/dd-trace/src/ci-visibility/dynamic-instrumentation/index.js`
- `packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js`

I have only a small question (see comment), but besides that it look good 👍

watson (approved):

PR #5083, opened 2025-01-08
switch to npm registry for yarn

rochdev:
### What does this PR do?

Switch to npm registry for yarn.

### Motivation

It looks like the Yarn registry may not be rate limiting `npm view` calls as we've seen a few packages that have many versions start to fail. This is an attempt to see if switching to the official registry could solve the problem.

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,0 +1,1 @@
+registry: https://registry.npmjs.org/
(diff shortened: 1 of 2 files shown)


BridgeAR:
Closing as not needed and discussed with @rochdev

PR #5089, opened 2025-01-09
[serverless] lazy load SDKs

duncanista:
### What does this PR do?

Lazy loads SDKs when they're enabled

### Motivation
These SDKs should not be loaded unless the product is explicitly enabled.
Screenshot 2025-01-09 at 3 10 19 PM

This is increasing overhead by 25ms

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### Additional Notes

⚠️  This is missing unit test updates

### DIFF
--- a/packages/dd-trace/src/proxy.js
+++ b/packages/dd-trace/src/proxy.js
@@ -9,13 +9,11 @@ const DynamicInstrumentation = require('./debugger')
 const telemetry = require('./telemetry')
 const nomenclature = require('./service-naming')
 const PluginManager = require('./plugin_manager')
-const AppsecSdk = require('./appsec/sdk')
 const dogstatsd = require('./dogstatsd')
 const NoopDogStatsDClient = require('./noop/dogstatsd')
 const spanleak = require('./spanleak')
 const { SSIHeuristics } = require('./profiling/ssi-heuristics')
 const appsecStandalone = require('./appsec/standalone')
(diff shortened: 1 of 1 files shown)


packages/dd-trace/src/proxy.js
```diff
@@ -89,7 +86,7 @@ class Tracer extends NoopProxy {
       }
 
       if (config.remoteConfig.enabled && !config.isCiVisibility) {
-        const rc = remoteConfig.enable(config, this._modules.appsec)
+        const rc = require('./appsec/remote_config').enable(config, this._modules.appsec)
```
rochdev, line 89: This file is required in other places. Should the lazy loading happen in the module instead instead?

joeyzhao2018: would this const be assigned each time for normal runs besides cold start runs?

rochdev:
> AppSec SDK needs to be loaded even when appsec is not enabled

@simon-id Can you explain why that is the case? I remember this being discussed in the past but I don't remember the reasoning.

simon-id:
the SDK methods are mostly just settings specific tags on the rootSpan from the arguments passed to them (and calling the WAF when available).
They're completely standalone to DD_APPSEC_ENABLED=true

duncanista:
Closing since this has been addressed by other work done in the tracer.

PR #5109, opened 2025-01-14
chore: add some shared llm-type plugin utilities

sabrenner:
### What does this PR do?
Adds utilities for llm-based APM tracing plugins for span tags: truncation/normalization and prompt/completion sample rate. 

Note that for `openai`, we changed the name of one config to better respect the namespacing of each config. Additionally, we are not adding the prompt/completion sample rate for `openai` in this PR - if needed, it can be done in a separate PR.

**Aside:** I considered making an `LlmPlugin` type extending `TracingPlugin` that would just have these methods on their prototype and read from the tracer config in the constructor, but this won't work well for AWS bedrock (which is already a base AWS plugin type), which we'll be updating after this PR lands. It additionally also would've been weird for LangChain (as the handlers are a property of the plugin and not an extension, so the functionality would've had to have been passed as a constructor arg).

### Motivation
We'll be adding more APM integrations for llm-based libraries, and they'll need these helpers to reduce code duplication.


packages/dd-trace/test/plugins/util/llm.spec.js
```diff
@@ -0,0 +1,64 @@
...
+      })
+    })
+
+    it('should normalize text to 100 characters', () => {
+      const text = 'a'.repeat(256)
+      expect(utils.normalize(text)).to.equal('a'.repeat(100) + '...')
+    })
+
+    it('should sample prompt completion at 50%', () => {
+      sinon.stub(Math, 'random').returns(0.4)
```
tlhunter, line 53: These stubs should probably be in an adjacent beforeEach/afterEach. As it is if the expect on line 54 throws then it might make it hard to debug weird failures later in the test suite.

sabrenner: fixed

tlhunter (approved):

PR #5120, opened 2025-01-16
fix startup benchmark

bengl:
Some dependencies don't work correctly with `require` anymore.


ddyurchenko (approved):
All good, there are no errors in Benchmarking Platform logs anymore.

PR #5124, opened 2025-01-17
Do not init the tracer by default in Worker threads

uurien:
### What does this PR do?

Prevent the tracer init in Worker threads by default. 

### Motivation

- Can cause unexpected and untested behaviours: Two tracers with the same service name but different ids
-  In `initialize.mjs` we are initializing the tracer only in the main thread: https://github.com/DataDog/dd-trace-js/blob/master/initialize.mjs#L52

### DIFF
--- a/ci/init.js
+++ b/ci/init.js
@@ -2,6 +2,7 @@
 const tracer = require('../packages/dd-trace')
 const { isTrue } = require('../packages/dd-trace/src/util')
 const log = require('../packages/dd-trace/src/log')
+const { isMainThread } = require('worker_threads')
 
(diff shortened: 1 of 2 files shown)


juan-fernandez:
@uurien could you to add this logic to `ci/init.js` as well? I'm just curious whether the newest DI + Test optimization logic would work with this.

Also: are we worried at all about this being a breaking change?

watson:
@uurien Most of the Dynamic Instrumentation code runs in Worker Thread. It doesn't need to instrument the code in the Worker Thread, but it expects to access certain parts of the tracer - e.g. the logger. I can see that debugger tests are failing. Let me know if you need my input or review on this PR.

PR #5126, opened 2025-01-17
upgrade(tracer): Support libdatadog's library_config module

BaptisteFoy:
### What does this PR do?
Supports libdatadog's `library_config` module to configure the tracing libraries from a file on disk that may remotely managed.

### Motivation
Being able to configure tracing libraries from a file / the UI

### Additional Notes
**This PR requires a release of `libdatadog-nodejs` & a bump in package.json**


packages/dd-trace/src/config.js
```diff
@@ -236,6 +237,25 @@ function reformatSpanSamplingRules (rules) {
 
 class Config {
   constructor (options = {}) {
+    // Read & apply the YAML configuration
+    const libconfig = libdatadog.maybeLoadWASM('library_config')
+    if (libconfig != null) {
```
rochdev, line 242: Not sure I understand this condition. My understanding is that the native config would be replacing the config we currently do in process. It doesn't quite make sense to me then to allow it to be missing, it should always be present so that we can remove the JS code.

packages/dd-trace/src/config.js
```diff
@@ -236,6 +237,25 @@ function reformatSpanSamplingRules (rules) {
 
 class Config {
   constructor (options = {}) {
+    // Read & apply the YAML configuration
```
rochdev, line 240: All of this new code could be in a method instead of right in the constructor.

package.json
```diff
@@ -81,7 +81,7 @@
     "node": ">=18"
   },
   "dependencies": {
-    "@datadog/libdatadog": "^0.4.0",
+    "@datadog/libdatadog": "file:../libdatadog-nodejs",
```
BaptisteFoy, line 84: Will be fixed before merge once we release `libdatadog-nodejs`

packages/dd-trace/src/config.js
BaptisteFoy: Indeed, I just migrated it 👍

packages/dd-trace/src/config.js
BaptisteFoy: > My understanding is that the native config would be replacing the config we currently do in process.
The current way of configuring the tracing library will not be removed. We will simply add new configuration sources.

This is mostly a failsafe to avoid crashing the app if anything happens.

szegedi (changes requested):
Mostly looks good, made some suggestions.

When `@datadog/libdatadog` gets released, add `package.json` and `yarn.lock` updates to this PR – I know you mentioned in the PR description that it needs to be done; I'm just emphasizing that it should be done as part of this PR.

packages/dd-trace/src/config.js
```diff
@@ -19,6 +19,7 @@ const telemetryMetrics = require('./telemetry/metrics')
 const { getIsGCPFunction, getIsAzureFunction } = require('./serverless')
 const { ORIGIN_KEY, GRPC_CLIENT_ERROR_STATUSES, GRPC_SERVER_ERROR_STATUSES } = require('./constants')
 const { appendRules } = require('./payload-tagging/config')
+const libdatadog = require('@datadog/libdatadog')
```
rochdev, line 22: This should not always be loaded as it won't be available in serverless environments. This probably means we need to either revisit how libdatadog is packaged or how unwanted files can be excluded, but in the meantime it should be loaded only when needed and when available.

See https://github.com/DataDog/datadog-lambda-js/blob/9e4af128689ec9832f491e676a487f5f9fa99d6d/Dockerfile#L37

packages/dd-trace/src/config.js
```diff
@@ -244,13 +247,31 @@ class Config {
...
+      process.env.OTEL_LOG_LEVEL === 'debug' || undefined,
+      localFileConfigEntries.DD_TRACE_DEBUG,
+      logConfig.enabled
+    ))
     this.logger = coalesce(options.logger, logConfig.logger)
-    this.logLevel = coalesce(options.logLevel, logConfig.logLevel)
+    this.logLevel = coalesce(
+      options.logLevel,
+      fleetFileConfigEntries.DD_TRACE_LOG_LEVEL,
+      process.env.DD_TRACE_LOG_LEVEL,
```
rochdev, line 261: This is now defined in 2 places as the logger reads it directly.

rochdev: Similar concern for the above as well.

packages/dd-trace/src/config.js
```diff
@@ -390,6 +413,85 @@ class Config {
     }
   }
 
+  _getStableConfigPaths () {
```
rochdev, line 416: There are a lot of new helpers and new code here all related to stable config, should it be extracted maybe to a separate file that is used by `config.js`?

packages/dd-trace/src/config.js
```diff
@@ -1330,9 +1476,33 @@ class Config {
...
+      this._fleetStableConfig,
+      this._env,
+      this._localStableConfig,
+      this._calculated,
+      this._defaults
+    ]
+    const origins = [
+      'remote_config',
+      'code',
+      'fleet_stable_config',
```
rochdev, line 1491: Are these real values? Just want to make sure as I believe there is a schema for the origin.

packages/dd-trace/src/config.js
rochdev: cc @khanayan123 since I know you've been looking into splitting up this file.

packages/dd-trace/test/config.spec.js
```diff
@@ -2372,4 +2372,89 @@ describe('Config', () => {
       expect(taggingConfig).to.have.property('maxDepth', 7)
     })
   })
+
+  context('library config', () => {
```
rochdev, line 2376: I feel like there are a few tests missing here, for example the priority order of the new sources with the old sources, error cases, etc. Only 3 tests seems a little small for the amount of code added.

rochdev (changes requested):

packages/dd-trace/src/config.js
BaptisteFoy: It could work to some extent on serverless, but it's not in the current scope. If it's possible I'll add conditional loading to avoid serverless.

packages/dd-trace/src/config.js
BaptisteFoy: These values have been defined in the technical implementation RFC and will be the same across libraries. Is the schema in the backend or in this repository?

rochdev: In the backend, but if the values are defined by the RFC I would expect the schema to include them 👍

packages/dd-trace/src/config.js
BaptisteFoy: Split in [split into two files](https://github.com/DataDog/dd-trace-js/pull/5126/commits/d683435d48353dd4b55f4617e1d4817f8b77860f)

packages/dd-trace/src/config.js
BaptisteFoy: Done in [address review](https://github.com/DataDog/dd-trace-js/pull/5126/commits/4e83b16fa4e1e31c7f3ef42c9e4f71f13287c12f)

packages/dd-trace/src/config.js
BaptisteFoy: Done in [address review](https://github.com/DataDog/dd-trace-js/pull/5126/commits/4e83b16fa4e1e31c7f3ef42c9e4f71f13287c12f)

packages/dd-trace/test/config.spec.js
BaptisteFoy: There is already a test about config priority, I have added more sources. I have also added the following tests:
- Invalid yaml
- Unknown key

See [add tests](https://github.com/DataDog/dd-trace-js/pull/5126/commits/343c9c2bc6af5c7bacc98623277039c3b55f4c72)

szegedi (changes requested):

packages/dd-trace/src/config.js
```diff
@@ -244,13 +247,24 @@ class Config {
...
+      this.stableConfig.fleetEntries?.DD_TRACE_DEBUG,
+      this.stableConfig.localEntries?.DD_TRACE_DEBUG
+    )
     this.logger = coalesce(options.logger, logConfig.logger)
-    this.logLevel = coalesce(options.logLevel, logConfig.logLevel)
-
+    this.logLevel = log.getLogLevel(
+      options.logLevel,
+      this.fleetEntries?.DD_TRACE_LOG_LEVEL,
+      this.localEntries?.DD_TRACE_LOG_LEVEL
```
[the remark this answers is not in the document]
BaptisteFoy: Good catch, thanks!

packages/dd-trace/src/config_stable.js
```diff
@@ -0,0 +1,158 @@
...
+        localConfigPath = '/etc/datadog-agent/application_monitoring.yaml'
+        fleetConfigPath = '/etc/datadog-agent/managed/datadog-agent/stable/application_monitoring.yaml'
+        break
+      case 'darwin':
+        localConfigPath = '/opt/datadog-agent/etc/application_monitoring.yaml'
+        fleetConfigPath = '/opt/datadog-agent/etc/managed/datadog-agent/stable/application_monitoring.yaml'
+        break
+      case 'win32':
+        localConfigPath = 'C:\\ProgramData\\Datadog\\application_monitoring.yaml'
+        fleetConfigPath = 'C:\\ProgramData\\Datadog\\managed\\datadog-agent\\stable\\application_monitoring.yaml'
```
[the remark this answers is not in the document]
BaptisteFoy: I just copied what exists on the shared library ([here](https://github.com/DataDog/libdatadog/blob/main/library-config/src/lib.rs#L500)) to avoid any drift between languages.

Using an env var is an interesting idea but if we implement it it should be the same everywhere.

packages/dd-trace/src/config.js
szegedi: It doesn't [seem to include them yet](https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v2/SchemaDocumentation/Schemas/conf_key_value.md). Since @paullegranddc is attached to the RFC I'd expect this'll happen sooner rather than later.

szegedi (changes requested):
We're getting there – looks mostly good, one more nit.

packages/dd-trace/src/config_stable.js
```diff
@@ -0,0 +1,159 @@
...
+    this.fleetEntries = {}
+    this.wasm_loaded = false
+
+    const { localConfigPath, fleetConfigPath } = this._getStableConfigPaths()
+    if (!fs.existsSync(localConfigPath) && !fs.existsSync(fleetConfigPath)) {
+      // Check if files exist, if not bail out early to avoid unnecessary library loading
+      return
+    }
+
+    // libdatadog isn't always available (e.g serverless) so we need to handle that case gracefully
```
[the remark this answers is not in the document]
BaptisteFoy: Done & tested in [address review](https://github.com/DataDog/dd-trace-js/pull/5126/commits/7de0d2a4e538e4912ee5681fddea6de48960df82)

szegedi (dismissed):

szegedi (approved):

packages/dd-trace/src/config.js
```diff
@@ -244,13 +247,24 @@ class Config {
...
+    this.logLevel = log.getLogLevel(
+      options.logLevel,
+      this.stableConfig.fleetEntries?.DD_TRACE_LOG_LEVEL,
+      this.stableConfig.localEntries?.DD_TRACE_LOG_LEVEL
+    )
     log.use(this.logger)
     log.toggle(this.debug, this.logLevel)
 
+    // Process stable config warnings, if any
+    for (const warning of this.stableConfig.warnings) {
```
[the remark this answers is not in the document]
BaptisteFoy: I don't think so; the logger isn't loaded in `stableConfig`.

packages/dd-trace/src/config.js
```diff
@@ -825,11 +851,7 @@ class Config {
     }
     this._setString(env, 'port', DD_TRACE_AGENT_PORT)
     const profilingEnabledEnv = coalesce(DD_EXPERIMENTAL_PROFILING_ENABLED, DD_PROFILING_ENABLED)
-    const profilingEnabled = isTrue(profilingEnabledEnv)
-      ? 'true'
-      : isFalse(profilingEnabledEnv)
-        ? 'false'
-        : profilingEnabledEnv === 'auto' ? 'auto' : undefined
+    const profilingEnabled = normalizeProfilingEnabledValue(profilingEnabledEnv)
```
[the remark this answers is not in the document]
BaptisteFoy: We can simplify the code a bit and remove `profilingEnabledEnv` as an intermediate variable. Though the input of `normalizeProfilingEnabledValue` isn't the same in the two places it's used (it's not an environment variable in `stableConfig`) so we can't do what you suggest.

packages/dd-trace/src/config_stable.js
```diff
@@ -0,0 +1,165 @@
...
+    this._setBoolean(obj, 'logInjection', DD_LOGS_INJECTION)
+    const profilingEnabledEnv = DD_PROFILING_ENABLED
+    const profilingEnabled = normalizeProfilingEnabledValue(profilingEnabledEnv)
+    this._setString(obj, 'profiling.enabled', profilingEnabled)
+    this._setBoolean(obj, 'runtimeMetrics', DD_RUNTIME_METRICS_ENABLED)
+    this._setString(obj, 'service', DD_SERVICE)
+    this._setString(obj, 'version', DD_VERSION)
+  }
+
+  _setBoolean (obj, name, value) {
```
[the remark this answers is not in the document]
BaptisteFoy: Done in [address review](https://github.com/DataDog/dd-trace-js/pull/5126/commits/c17602a6eee331d9ef6b32e853ee85f15903791e)

packages/dd-trace/src/config_stable.js
```diff
@@ -0,0 +1,165 @@
...
+    if (libconfig !== undefined) {
+      let localConfig = ''
+      try {
+        localConfig = fs.readFileSync(localConfigPath, 'utf8')
+      } catch (err) {
+        if (err.code !== 'ENOENT') {
+          this.warnings.push(`Error reading local config file: ${localConfigPath}`)
+        }
+      }
+      let fleetConfig = ''
```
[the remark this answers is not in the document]
BaptisteFoy: Done in [address review](https://github.com/DataDog/dd-trace-js/pull/5126/commits/c17602a6eee331d9ef6b32e853ee85f15903791e)

packages/dd-trace/src/config_stable.js
```diff
@@ -0,0 +1,165 @@
...
+      try {
+        fleetConfig = fs.readFileSync(fleetConfigPath, 'utf8')
+      } catch (err) {
+        if (err.code !== 'ENOENT') {
+          this.warnings.push(`Error reading fleet config file: ${fleetConfigPath}`)
+        }
+      }
+
+      if (localConfig || fleetConfig) {
```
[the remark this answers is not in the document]
BaptisteFoy: Done in [address review](https://github.com/DataDog/dd-trace-js/pull/5126/commits/c17602a6eee331d9ef6b32e853ee85f15903791e)

packages/dd-trace/src/config_stable.js
```diff
@@ -0,0 +1,165 @@
...
+      this.wasm_loaded = true
+    } catch (e) {
+      this.warnings.push('Can\'t load libdatadog library')
+      return
+    }
+
+    // Note: we use maybeLoad because there may be cases where the library is not available and we
+    // want to avoid breaking the application. In those cases, we will not have the file-based configuration.
+    const libconfig = libdatadog.maybeLoad('library_config')
+    if (libconfig !== undefined) {
```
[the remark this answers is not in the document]
BaptisteFoy: Done in [address review](https://github.com/DataDog/dd-trace-js/pull/5126/commits/c17602a6eee331d9ef6b32e853ee85f15903791e)

packages/dd-trace/src/config.js
```diff
@@ -1330,9 +1352,33 @@ class Config {
...
+      this._fleetStableConfig,
+      this._env,
+      this._localStableConfig,
+      this._calculated,
+      this._defaults
+    ]
+    const origins = [
+      'remote_config',
+      'code',
+      this.stableConfig.FLEET_STABLE_CONFIG_ORIGIN,
```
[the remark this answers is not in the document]
BaptisteFoy: Done in [address review](https://github.com/DataDog/dd-trace-js/pull/5126/commits/c17602a6eee331d9ef6b32e853ee85f15903791e)

packages/dd-trace/src/config_stable.js
```diff
@@ -0,0 +1,165 @@
...
+
+class StableConfig {
+  constructor () {
+    this.warnings = [] // Logger hasn't been initialized yet, so we can't use log.warn
+    this.localEntries = {}
+    this.fleetEntries = {}
+    this.wasm_loaded = false
+
+    if (isInServerlessEnvironment()) {
+      // Bail out early if we're in a serverless environment, stable config isn't supported
```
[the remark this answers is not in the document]
BaptisteFoy: I tried something in [don't load config_stable.js on serverless](https://github.com/DataDog/dd-trace-js/pull/5126/commits/1cbf8a5fe9b80b0332fb4d0d53ce5f9e63018ef1); please tell me if this what you expect.

I'm semi-convinced by my approach because it forces us to check whether or not the `stableConfig` object is defined at every call.

packages/dd-trace/src/config.js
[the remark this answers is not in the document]
szegedi: I suggested this approach in [earlier](https://github.com/DataDog/dd-trace-js/pull/5126#discussion_r1954650219) review specifically because logging is only configured after `stableConfig` initializes, since the values it reads can affect logging.

szegedi (dismissed):

rochdev (approved):

khanayan123 (approved):
No concerns on my end

PR #5137, opened 2025-01-20
(chore)APM: Refactor Bedrock Integration

yahya-mouman:
### What does this PR do?

This PR adds the following changes to the AWS Bedrock APM Integration
- Adds parsing for more variations of the model id in bedrock
- Channel suffix space removal
- Plugin rename from bedrock to bedrockruntime
- Structure refactor adding a utils file with all the parsing logic

Please also review https://github.com/DataDog/dd-trace-js/pull/5116
As it contains some fixture refactors to the testing code

### Motivation

The addition of the LLMObs plugin for this integration makes it necessary to refactor the code and share utilities to avoid code duplication.

### Plugin Checklist

- [x] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


CODEOWNERS
```diff
@@ -59,6 +59,7 @@
 /packages/datadog-plugin-langchain/ @DataDog/ml-observability
 /packages/datadog-instrumentations/src/openai.js @DataDog/ml-observability
 /packages/datadog-instrumentations/src/langchain.js @DataDog/ml-observability
+/packages/datadog-plugin-aws-sdk/src/services/bedrockruntime @DataDog/ml-observability
```
sabrenner, line 62: we can probably add ourselves to the test file as well

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/index.js
```diff
@@ -0,0 +1,14 @@
+const CompositePlugin = require('../../../../dd-trace/src/plugins/composite')
+const BedrockRuntimeTracing = require('./tracing')
+class BedrockRuntimePlugin extends CompositePlugin {
+  static get id () {
+    return 'bedrock'
```
sabrenner, line 5:
```suggestion
    return 'bedrockruntime'
```

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/tracing.js
```diff
@@ -0,0 +1,65 @@
...
+    const operation = request.operation
+    if (!enabledOperations.includes(operation)) {
+      return false
+    }
+
+    return super.isEnabled(request)
+  }
+
+  generateTags (params, operation, response) {
+    let tags = {}
```
sabrenner, line 21:
```suggestion
```

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
```diff
@@ -176,11 +191,13 @@ function extractRequestParams (params, provider) {
   }
 }
 
-function extractTextAndResponseReason (response, provider, modelName, shouldSetChoiceIds) {
+function extractTextAndResponseReason (response, provider, modelName, shouldSetChoiceIds = undefined) {
   const body = JSON.parse(Buffer.from(response.body).toString('utf8'))
-
+  if (shouldSetChoiceIds === undefined) {
```
sabrenner, line 196:
```suggestion
  if (!shouldSetChoiceIds) {
```
if taking the above suggestion as well. not sure if `shouldSetChoiceIds` is used elsewhere in this method too, but if so it should also change accordingly

sabrenner (commented):
just a couple comments - this is a really nice cleanup, awesome job!

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
```diff
@@ -176,11 +191,13 @@ function extractRequestParams (params, provider) {
   }
 }
 
-function extractTextAndResponseReason (response, provider, modelName, shouldSetChoiceIds) {
+function extractTextAndResponseReason (response, provider, modelName, shouldSetChoiceIds = undefined) {
```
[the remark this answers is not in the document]
yahya-mouman: My thinking there is I'm only setting an override option to the shouldSetChoiceIds where if anything is provided (true or false) it would override inferring it inside the function.
With the absence of an override, the provider is used alongside the model name to know if we should set the choice ids or not.
If i set the default value to false I won't be able to cover the case where the in function logic is stating that we should set choice ids, while we want to not take the inferred value into consideration and just not send any id regardless of the provider.
All of this is completely optional IMO as we could just infer the shouldSetChoiceIds inside the function and not pass it at all.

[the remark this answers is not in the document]
sabrenner: ah i think i see what you're saying. i would say yeah we should just infer it inside the function (it looks like the only place we use this for now we don't pass this argument in). if we need to fix this again for LLMObs that's OK, but we can just infer inside the function for now.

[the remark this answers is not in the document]
yahya-mouman: Alright will be removing the override parameter in my next commit

sabrenner (approved):

packages/datadog-instrumentations/src/aws-sdk.js
```diff
@@ -155,6 +155,8 @@ function getMessage (request, error, result) {
 }
 
 function getChannelSuffix (name) {
+  // some resource identifiers have spaces between ex: bedrock runtime
+  name = name.split(' ').join('')
```
tlhunter, line 159:
```suggestion
  name = name.replaceAll(' ', '')
```
This should be quicker as it skips the intermediary array representation.

tlhunter (dismissed):

sabrenner (approved):

packages/datadog-instrumentations/src/aws-sdk.js
```diff
@@ -155,6 +155,8 @@ function getMessage (request, error, result) {
 }
 
 function getChannelSuffix (name) {
+  // some resource identifiers have spaces between ex: bedrock runtime
+  name = name.replaceAll(' ', '')
```
tlhunter, line 159: It looks like `name` can be undefined which causes this line to throw an error.

PR #5145, opened 2025-01-22
Instrument dd-trace-api

bengl:
This is the initial PR that adds support for the (soon to be released) `dd-trace-api` module.

The `dd-trace-api` module works by sending all API calls through diagnostics channels, including the name of the function being called, all the arguments, and the `this` object. appropriate functions are passed through for any return values that need to be proxied.

What's proxying? Here, it refers to having a facade object presented by `dd-trace-api` that corresponds to a "real" object created by `dd-trace`. The facade allows `dd-trace-api` to never expose an internal object to the caller.

Testing is done primarily by checking that diagnostics channels map correctly to the internal function calls. Additional testing will take place in `dd-trace-api` to ensure the whole system works end-to-end.

### DIFF
--- a/packages/datadog-plugin-dd-trace-api/src/index.js
+++ b/packages/datadog-plugin-dd-trace-api/src/index.js
@@ -76,23 +76,35 @@ module.exports = class DdTraceApiPlugin extends Plugin {
       })
     }
 
-    handleEvent('init')
-    // TODO(bengl) for API calls like this one that return an object created
-    // internally, care needs to be taken to ensure we're not breaking the
-    // calling API. We don't expect spans to change much, but if they do, this
-    // needs to be taken into account.
+    // handleEvent('configure')
     handleEvent('startSpan')
     handleEvent('wrap')
     handleEvent('trace')
     handleEvent('inject')
     handleEvent('extract')
     handleEvent('getRumData')
-    handleEvent('setUser')
     handleEvent('profilerStarted')
     handleEvent('span:context')
     handleEvent('span:setTag')
     handleEvent('span:addTags')
     handleEvent('span:finish')
     handleEvent('span:addLink')
+    handleEvent('scope')
+    handleEvent('scope:activate')
+    handleEvent('scope:active')
+    handleEvent('scope:bind')
+    handleEvent('appsec:blockRequest')
+    handleEvent('appsec:isUserBlocked')
+    handleEvent('appsec:setUser')
+    handleEvent('appsec:tracerCustomEvent')
+    handleEvent('appsec:trackUserLoginFailureEvent')
+    handleEvent('appsec:trackUserLoginSuccessEvent')
+    handleEvent('dogstatsd:decrement')
+    handleEvent('dogstatsd:distribution')
+    handleEvent('dogstatsd:flush')
+    handleEvent('dogstatsd:gauge')
+    handleEvent('dogstatsd:histogram')
+    handleEvent('dogstatsd:increment')
+    handleEvent('use')
   }
 }
--- a/packages/datadog-plugin-dd-trace-api/src/index.js
+++ b/packages/datadog-plugin-dd-trace-api/src/index.js
@@ -1,0 +1,50 @@
+'use strict'
+
+const Plugin = require('../../dd-trace/src/plugins/plugin')
+
+const objectMap = new WeakMap()
+
+class DdTraceApiPlugin extends Plugin {
+  static get id () {
+    return 'dd-trace-api'
+  }
+
+  constructor (...args) {
+    super(...args)
+
+    const tracer = this._tracer
+
+    const handleEvent = (name, fn) => {
+      // For v1, APIs are 1:1 with their internal equivalents, so we can just
+      // call the internal method directly. That's what we do here unless we
+      // want to override. As the API evolves, this may change.
+      this.addSub(`datadog-api:v1:${name}`, ({ self, args, ret, dummy }) => {
+        if (!fn && name.includes(':')) {
+          name = name.split(':').pop()
+        }
+
+        try {
+          ret.value = fn ? fn(args, self) : self[name](...args)
+          if (dummy) {
+            objectMap.set(dummy, ret.value)
+            ret.value = dummy
+          }
+        } catch (e) {
+          ret.error = e
+        }
+      })
+    }
+
+    handleEvent('init')
+    // TODO(bengl) for API calls like this one that return an object created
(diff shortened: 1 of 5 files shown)


packages/datadog-plugin-dd-trace-api/src/index.js
```diff
@@ -0,0 +1,117 @@
...
+      objectMap.set(proxyVal, tracer)
+      objectMap.set(proxyVal.appsec, tracer.appsec)
+      objectMap.set(proxyVal.dogstatsd, tracer.dogstatsd)
+    })
+
+    const handleEvent = (name) => {
+      const counter = apiMetrics.count('dd_trace_api.called', [
+        `name:${name.replace(':', '.')}`,
+        'api_version:v1',
+        `injection_enabled:${process.env.DD_INJECTION_ENABLED ? 'yes' : 'no'}`
```
tlhunter, line 31: Can the `process.env.DD_INJECTION_ENABLED` check be hoisted to the outermost scope? I'm assuming it doesn't change throughout the lifetime of a process and it doesn't seem to be set during tests.

Also hopefully future config changes will make the config file easier to read in files like this one without needing to grab env vars everywhere.

bengl: This particular env var is typically only used when config hasn't been initialized yet, so it's never been part of our config system IIRC. I'll hoist it as requested.

packages/datadog-plugin-dd-trace-api/src/index.js
```diff
@@ -0,0 +1,120 @@
...
+                  const proxyVal = revProxy[j]()
+                  objectMap.set(proxyVal, fnArgs[j])
+                  fnArgs[j] = proxyVal
+                }
+              }
+              // TODO do we need to apply(this, ...) here?
+              return orig(...fnArgs)
+            }
+          }
+        }
```
sabrenner, line 69: what is this block trying to do here? 

i have been reviewing this in the context of the `dd-trace-api-js` [implementation](https://github.com/DataDog/dd-trace-api-js/blob/bengl/next/index.js), and everything in these handlers around this block makes sense to me (in terms of setting and grabbing the proxy obj value from `self`, and calling the mapped function on the tracer), but I'm unsure what this block is doing. 

in what case would we check the arguments to the function against the `objectMap`? what is `revProxy` (i see on the publisher side & tests it's just defaulted to an empty list, which doesn't help here 😅)? why do we need to wrap a `function` argument (and, maybe _it's_ arguments)?

sabrenner (commented):
awesome stuff, almost all lgtm! i think having some context before reviewing helped (when we talked a few weeks ago), and looking along with the `dd-trace-api-js` helped a bunch. started a review because i thought i'd have more questions but really just the one 😅

packages/datadog-plugin-dd-trace-api/src/index.js
bengl: It's preserving the object map across all function calls that happen through our message bus (i.e. diagnostics channel). That means objects on the API side will never be _real_ objects returned from _real_ API calls.

We want to map these to "real" objects from the internal/dd-trace API so that a "real" value is never exposed to the user via dd-trace-api. I sometimes call this concept "the firewall". We need the firewall because we never want to expose an internal API to the user by accident, or else the whole premise of keeping the API separate just falls apart.

The object map has "dummy" objects as keys, and "real" objects as values. To maintain the firewall, we map them later on (see what we do later on in code with `ret.value`). When an API call is made the `proxy` property contains a function that returns a "dummy" object to map to. This is how we populate the object map.

Now, suppose one of these dummy objects, having already been mapped, is passed in to one of our APIs (take `tracer.inject(span, carrier)` as an example, where span was returned from one of our APIs so it's a mapped object). In order to make sure that the underlying "real API" only receives "real" objects, we need to map all of these through the object map. This would give us some nice code that looks something like this:

```js
for (let i = 0; i < args.length; i++) {
  if (objectMap.has(args[i])) {
    args[i] = objectMap.get(args[i])
  }
}
```

That would preserve our firewall with respect to arguments passed to these API calls, and indeed, constitutes the first few lines you reference here.

Some of our API calls can take in callbacks. In this case, we need to do everything in reverse; instead of mapping return values, we need to map function arguments. A simple way to think of why this is necessary is to consider that callback arguments are effectively return values (google "continuation passing style"). An easy example API here is `tracer.trace('name', options, () => { /* do stuff to trace */ })`.

So, to deal with this in our loop, we first check if any given argument is a function. That means it's a callback so we have to jump into this reverse-mapping procedure. We replace the function with a wrapper function that passes any relevant args through relevant `revProxy` array functions, which need to give us dummy objects to pass through, much like `proxy` does for regular return values. We then map then in the object map, again, just like we do for return values. Then we're good to call the original/wrapped function.

In summary (and as a tl;dr, if you want), we have to map return values, and we have to unmap them when they're passed in as arguments. For callbacks, we have to do the reverse of those two operations. And that's about it.

sabrenner: This actually makes a lot of sense!! I think the overarching idea i wasn't connecting here was the whole "firewall" approach, which i get now in obfuscating the actual return values of things like span, scope, etc. from the tracer core api. i actually find that `tracer.inject` example a simple and understandable one for this, WRT passing these objects back to other tracer APIs!

it further helped seeing how it's actually implemented using `defaultFun` to signal this as the `proxy` value, and then through stuff like `getSpan` returns a dummy span with only the methods/properties we want to expose, and so on in `dd-trace-api-js`.

i think _maybe_ a small comment here could help future readers like myself - maybe just linking to the explanation above?

sabrenner (dismissed):
the approach and philosophy behind what's going on here makes sense to me!

packages/datadog-plugin-dd-trace-api/src/index.js
```diff
@@ -0,0 +1,120 @@
...
+    this.addSub('datadog-api:v1:tracerinit', ({ proxy }) => {
+      const proxyVal = proxy()
+      objectMap.set(proxyVal, tracer)
+      objectMap.set(proxyVal.appsec, tracer.appsec)
+      objectMap.set(proxyVal.dogstatsd, tracer.dogstatsd)
+    })
+
+    const handleEvent = (name) => {
+      const counter = apiMetrics.count('dd_trace_api.called', [
+        `name:${name.replace(':', '.')}`,
```
tlhunter, line 32: is this intentionally not `.replaceAll()`?

tlhunter (dismissed):

tlhunter (approved):

Issue #4987, opened 2024-12-10
Getting TraceID from an OpenTracing DatadogSpanContext can throw "Cannot read properties of undefined (reading 'toString')


davidalpert:
Calling this `DatadogSpanContext.toTraceId()` method on what appears to be a valid span can sometimes error out with an undefined inner `this._traceId` value:

https://github.com/DataDog/dd-trace-js/blob/497c0b0d4333b6a7e7a434fb4de275595e5950c9/packages/dd-trace/src/opentracing/span_context.js#L33-L40

### Suggested Solution

- While I am not sure what might cause this scenario, a quick fix would be to make the dot property access of `this._traceId` a bit more defensive:

    ```
     toTraceId (get128bitId = false) {
        if (this._traceId && get128bitId) {
          return this._traceId.toBuffer().length <= 8 && this._trace.tags[TRACE_ID_128]
            ? this._trace.tags[TRACE_ID_128] + this._traceId.toString(16).padStart(16, '0')
            : this._traceId.toString(16).padStart(32, '0')
        }
        return this._traceId?.toString(10)
      }
    ```

### Replication Steps

- Using:

    ```
        "dd-trace": "5.25.0",
    ```
- Initialize the tracer on startup:

    ```
    import tracer from 'dd-trace';
    import { version } from '../../package.json';
    import { buildVars } from '#root/.build/buildVars';
    
    tracer.init({
      service: process.env.DD_SERVICE || 'my-fe-app',
      version: version ? `${version}-${buildVars.GIT_SHA}` : '0.0.0',
      tags: {
        pid: process.pid,
      },
      runtimeMetrics: true,
      logInjection: true,
    });
    ```

- Try to defensively get the active span's `traceID` at runtime

    ```nodejs
    import tracer from 'dd-trace';
    
    ...
    
        const span = tracer.scope().active();
        const traceID = span?.context()?.toTraceId() ?? 'n/a';

    ...
    ```

### Expected result

- `traceID` is either a valid Trace ID from the current span or `undefined` in which case we default to `'n/a/` based on the code above

### Actual result

- We regularly encounter an error whereby we seem to have a defined `span` (i.e. not `null` or `undefined`) with an empty or missing `DatadogSpanContext._traceID` property such that when we call `toTraceId()` on it we get the following error:

    ```
    .../my_app/node_modules/dd-trace/packages/dd-trace/src/opentracing/span_context.js:39
         return this._traceId.toString(10)
                                          ^
[...]

PR #5147, opened 2025-01-23
make DatadogSpanContext.toTraceId and DatadogSpanContext.toSpanId more defensive

ida613:
### What does this PR do?
Add handling for the cases _traceId or _spanId is undefined as a possible fix for this [issue](https://github.com/DataDog/dd-trace-js/issues/4987)

### Motivation
[This line](https://github.com/DataDog/dd-trace-js/blob/c28765a66d5d7d2b2f97fc0cbcf179344926b71a/packages/dd-trace/src/opentracing/propagation/text_map.js#L348) technically makes it possible to have a DatadogSpanContext object with no _traceId or _spanId, even though I'm not sure when this case will be reached, other than when the only propagation style is `baggage` in the [tracer config object](https://github.com/DataDog/dd-trace-js/blob/c28765a66d5d7d2b2f97fc0cbcf179344926b71a/packages/dd-trace/src/config.js#L564-L565). Some customers reported they started seeing crashes due to this issue after their services have been running for a while, so I think there is probably some edge cases that are currently unaccounted for. This PR is an attempt at remedying this.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/opentracing/span_context.js
+++ b/packages/dd-trace/src/opentracing/span_context.js
@@ -49,14 +49,14 @@ class DatadogSpanContext {
         ? this._trace.tags[TRACE_ID_128] + this._traceId.toString(16).padStart(16, '0')
         : this._traceId.toString(16).padStart(32, '0')
     }
-    return this._traceId.toString(10)
+    return this._traceId?.toString(10)
   }
 
   toSpanId (get128bitId = false) {
     if (get128bitId) {
       return this._spanId.toString(16).padStart(16, '0')
     }
-    return this._spanId.toString(10)
+    return this._spanId?.toString(10)
   }
 
   toTraceparent () {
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/opentracing/span_context.js
```diff
@@ -49,14 +49,14 @@ class DatadogSpanContext {
         ? this._trace.tags[TRACE_ID_128] + this._traceId.toString(16).padStart(16, '0')
         : this._traceId.toString(16).padStart(32, '0')
     }
-    return this._traceId.toString(10)
+    return this._traceId?.toString(10)
```
rochdev, line 52: These should guarantee a string. Also, it should not be possible for IDs to be undefined to begin with, so this is working around a bug. If that's needed for OpenTelemetry, let's move that logic only to OpenTelemetry, or translate to something valid on our side.

rochdev (changes requested):

PR #5151, opened 2025-01-23
Add more YAML verification

bengl:
* Validate YAML files using actionlint
* Check that tests in `packages/datadog-instrumentations/tests/` are actually run.
* Check that tests in `packages/datadog-plugin-*/tests/` are actually run.
* Ensure all tests have the same triggers.


.github/workflows/project.yml
```diff
@@ -169,3 +169,22 @@ jobs:
       - uses: ./.github/actions/node/setup
       - uses: ./.github/actions/install
       - run: node scripts/verify-ci-config.js
+
+      - name: actionlint
+        id: actionlint  #optional, id required only when outputs are used in the workflow steps later
+        uses: raven-actions/actionlint@v2
```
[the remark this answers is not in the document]
watson: ??

packages/datadog-instrumentations/test/multer.spec.js
```diff
@@ -15,7 +15,7 @@ withVersions('multer', 'multer', version => {
     })
 
     before((done) => {
-      const express = require('../../../versions/express').get()
+      const express = require('express')
```
simon-id, line 18: why this change ? shouldn't multer decide what version it's supposed to use from the external.json file thingy ?

simon-id: wouldn't that just use the express we have in our package.json ?

scripts/verify-ci-config.js
```diff
@@ -30,6 +45,9 @@ function checkYaml (yamlPath) {
     if (!job.env || !job.env.PLUGINS) continue
 
     const pluginName = job.env.PLUGINS
+    if (!yamlPath.includes('appsec')) {
```
simon-id, line 48: why not check appsec ones too ?

simon-id:
few comments otherwise LGTM

scripts/verify-ci-config.js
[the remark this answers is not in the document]
bengl: yes it is.

packages/datadog-instrumentations/test/multer.spec.js
bengl: _sheepish admission_ i don't know how to use external.json correctly. What needs to change here to make that happen?

Note: This multer test was _not actually running at all_ prior to this PR. That's part of what's being corrected.

scripts/verify-ci-config.js
bengl: That can probably be checked. I was concentrating on the issue at hand, which was non-appsec plugin tests.

packages/datadog-instrumentations/test/multer.spec.js
simon-id: I also am not 100% how to use external.json, BUT from what I understand, if a plugin/instrumentation test needs to depend on another library for the setup, it needs to be put in the externals.json like in my PR here: https://github.com/DataDog/dd-trace-js/pull/4670/files#diff-c79dfa6aa60b124902dd80d1f3f70cfaea8caa71b801e93d7cc23b38d1e4527d

simon-id: at least this is how I understand the external.json file, but it's always been blurry for me tbh

.github/workflows/project.yml
```diff
@@ -169,3 +169,36 @@ jobs:
       - uses: ./.github/actions/node/setup
       - uses: ./.github/actions/install
       - run: node scripts/verify-ci-config.js
+
+  actionlint:
```
watson, line 173: Wouldn't it be prudent to put this in its own yaml file, so that the risk of a syntax error in that yaml file was as low as possible? That way a syntax error in `project.yml` would be detectable

scripts/verify-ci-config.js
bengl: Actually, looking into it. `test:appsec:plugins` is somewhat non-trivial. Are you fine with this being handled in a subsequent PR?

.github/workflows/project.yml
bengl: Sure I can do that.

bengl: Done

packages/datadog-instrumentations/test/multer.spec.js
bengl: I changed this so it's now in externals and referring to it via `versioons`.

simon-id (approved):

PR #5152, opened 2025-01-24
[MLOB-2098] feat(llmobs): record bedrock token counts

sabrenner:
### What does this PR do?
Attempts to record token counts from headers for AWS Bedrock Runtime requests (for `InvokeModelCommand`).

#### APM Patching Changes
Changes AWS patching to emit response headers on deserialization. Some clients drop headers when deserializing (like BedrockRuntime) that could be useful to us. I tried to do this a few different ways, but the most direct way seemed to be patching the deserializer for the command being executed by the client, since:
- the `deserialize` method is where the raw response is directly passed in before headers are stripped
- the `deserialize` method is not on the `prototype` of the `Command` class. While it is sometimes on the prototype of classes that inherit from `Command`, this changes by version and is not guaranteed for all commands. It does seem, however, that the `deserialize` property is available at the time a command is executed via `client.send`, so we patch it on the instance there.
- A small check is included to make sure `deserialize` is a function

#### LLMObs Changes
We extract the request ID and token counts to map against each other, since we don't have a shared context object being passed around. We then associate the request ID from the final response against our map, and grab the token counts from there. However, we'll first check if we were able to extract tokens from the actual response body (as some models return them) in case the header values were empty.

Additionally, changes some Bedrock shared utilities and testing fixture.

### Motivation
Full support for counting tokens from models served through AWS Bedrock from `@aws-sdk/client-bedrock-runtime`. This is only for the LLMObs plugin but we are adding the publishers in patching.


packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
```diff
@@ -24,11 +24,23 @@ const PROVIDER = {
 }
 
 class Generation {
-  constructor ({ message = '', finishReason = '', choiceId = '' } = {}) {
+  constructor ({
+    message = '',
+    finishReason = '',
+    choiceId = '',
+    role,
```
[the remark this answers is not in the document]
sabrenner: it will get set as `undefined` as-is if it's not passed on the object

yahya-mouman (dismissed):
Great work 🚀, I'm seeing some unrelated failing tests but it should be alright.

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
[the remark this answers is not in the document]
yahya-mouman: ah yes forgot that the whole object is optional !

packages/datadog-instrumentations/src/helpers/hooks.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
   '@apollo/gateway': () => require('../apollo'),
   'apollo-server-core': () => require('../apollo-server-core'),
   '@aws-sdk/smithy-client': () => require('../aws-sdk'),
+  '@aws-sdk/client-bedrock-runtime': () => require('../aws-sdk'),
```
sabrenner, line 8:
```suggestion
```

packages/datadog-instrumentations/src/aws-sdk.js
```diff
@@ -40,6 +40,16 @@ function wrapRequest (send) {
   }
 }
 
+function wrapDeserialize (deserialize, channelSuffix) {
+  const headersCh = channel(`apm:aws:response:deserialize:${channelSuffix}`)
+
+  return function (response) {
+    headersCh.publish({ headers: response.headers })
```
IlyasShabi, line 47: I think you need to check `if (headersCh.hasSubscribers) `

sabrenner: fixed 😄

packages/dd-trace/src/llmobs/plugins/bedrockruntime.js
```diff
@@ -30,6 +32,17 @@ class BedrockRuntimeLLMObsPlugin extends BaseLLMObsPlugin {
...
     })
+
+    this.addSub('apm:aws:response:deserialize:bedrockruntime', ({ headers }) => {
+      const requestId = headers['x-amzn-requestid']
+      const inputTokenCount = headers['x-amzn-bedrock-input-token-count']
+      const outputTokenCount = headers['x-amzn-bedrock-output-token-count']
+
+      requestIdsToTokens[requestId] = {
+        inputTokensFromHeaders: inputTokenCount && parseInt(inputTokenCount),
+        outputTokensFromHeaders: outputTokenCount && parseInt(outputTokenCount)
```
IlyasShabi, line 43: Because `parseInt` could return `NaN`, I think you should check if `inputTokenCount` and `outputTokenCount` are valid numbers before adding a new item to `requestIdsToTokens`

sabrenner: good point - I do default this below when doing
```javascript
const inputTokens = usage.inputTokens || inputTokensFromHeaders || 0
```
which would default to 0 if the first is `undefined`, and the second is `NaN` from the operation above. I'm personally OK with keeping the check as-is, because even though they might switch away from this header, the idea is it should always be a number or string number, or non-existent, if you also think this is OK! if not though, i'll make the change

IlyasShabi: Im okay with your approach too

IlyasShabi (approved):
LGTM, I'd recommend getting another approval from your team

yahya-mouman (approved):

PR #5155, opened 2025-01-27
Fix typo in `DD_INJECTION_ENABLED`

### LINKED ISSUE #5146: Invalid env variable name
Likely a typo https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-shimmer/src/shimmer.js#L76C30-L76C50

juan-fernandez:
### What does this PR do?
Use `DD_INJECTION_ENABLED`

### Motivation
Fixes #5146

### Plugin Checklist
Ideally this should've been caught by unit tests, but I'm not sure how this is tested

- [ ] Unit tests.

### DIFF
--- a/packages/datadog-shimmer/src/shimmer.js
+++ b/packages/datadog-shimmer/src/shimmer.js
@@ -73,7 +73,7 @@ function isPromise (obj) {
   return obj && typeof obj === 'object' && typeof obj.then === 'function'
 }
 
-let safeMode = !!process.env.DD_INEJCTION_ENABLED
+let safeMode = !!process.env.DD_INJECTION_ENABLED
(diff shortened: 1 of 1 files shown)


BridgeAR (commented):
Code LGTM. It can be tested similar to the current tests in packages/datadog-shimmer/test/shimmer.spec.js:231

Instead of using the setter for the tests, we have to change the environment variables and also reset the file to not being loaded in the require cache. Then, require the simmer module again with the now reset environment variable. Feel free to reach out to me on slack in case you would like to do some pair programming

bengl:
We've removed safe mode for now, so this is no longer relevant. Closing.

PR #5157, opened 2025-01-27
use sandbox for express plugin test

rochdev:
### What does this PR do?

Use sandbox for express plugin test.

### Motivation

This PR is trying a simple strategy for running plugin tests on the final artifact of dd-trace instead of on the local code.


BridgeAR:
@rochdev is this something you would still like to work upon?

rochdev:
> is this something you would still like to work upon?

@BridgeAR Yes, eventually all tests should be using the sandbox similar to the integration tests. However, before that happens it needs to be streamlined. This hasn't been prioritized yet.

BridgeAR:
I think we should get back to this, when we really need it.

Using the sandbox has an overhead and I do not remember issues around code having a negative impact on each other. If I am not mistaken, it might even detect such problems with the current setup just as much as it could hide issues.

PR #5162, opened 2025-01-28
chore(tracing): graphql error support

wconti27:
### What does this PR do?
Adds error support for GraphQL using span events.

### Plugin Checklist

- [x] Unit tests.
- [x] TypeScript [tests][2].
- [x] API [documentation][3].

### DIFF
--- a/packages/datadog-plugin-graphql/src/utils.js
+++ b/packages/datadog-plugin-graphql/src/utils.js
@@ -1,0 +1,41 @@
+function extractErrorIntoSpanEvent (config, span, exc) {
+  const attributes = {}
+
+  if (exc.name) {
+    attributes.type = exc.name
+  }
+
+  if (exc.stack) {
+    attributes.stacktrace = exc.stack
+  }
+
+  if (exc.locations) {
+    attributes.locations = []
+    for (const location of exc.locations) {
+      attributes.locations.push(`${location.line}:${location.column}`)
+    }
+  }
+
+  if (exc.path) {
+    attributes.path = exc.path
+  }
+
+  if (exc.message) {
+    attributes.message = exc.message
+  }
+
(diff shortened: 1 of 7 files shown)


packages/datadog-plugin-graphql/src/utils.js
```diff
@@ -0,0 +1,40 @@
...
+
+  if (exc.locations) {
+    attributes.locations = []
+    for (const location of exc.locations) {
+      attributes.locations.push(`${location.line}:${location.column}`)
+    }
+  }
+
+  if (exc.path) {
+    attributes.path = exc.path
```
marcotc, line 20: We need to stringify the elements in `exc.path` because span events attributes do not support heterogeneous arrays (arrays with a mix of integers and strings), and `exc.path` can be `["query", 0, "id]` when there are arrays in the GraphQL path.

packages/datadog-plugin-graphql/src/utils.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  if (exc.message) {
+    attributes.message = exc.message
+  }
+
+  if (config.graphqlErrorExtensions) {
+    for (const ext of config.graphqlErrorExtensions) {
+      if (exc.extensions?.[ext]) {
+        attributes[`extensions.${ext}`] = exc.extensions[ext]
```
marcotc, line 30: `exc.extensions[ext]` needs to be stringified, was we want to limit the potential for unbounded nesting in user-provided extensions.

tlhunter (dismissed):

packages/datadog-plugin-graphql/src/utils.js
wconti27: Okay, I'll fix that 👍

tlhunter (approved):

PR #5164, opened 2025-01-28
Code injection instrumented metric

IlyasShabi:
### What does this PR do?
Support instrumented sink metric sink for code injection

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/test/appsec/iast/code_injection.integration.spec.js
+++ b/packages/dd-trace/test/appsec/iast/code_injection.integration.spec.js
@@ -1,12 +1,13 @@
 'use strict'
 
-const { createSandbox, FakeAgent, spawnProc } = require('../../../../../integration-tests/helpers')
 const getPort = require('get-port')
 const path = require('path')
 const Axios = require('axios')
+const { assert } = require('chai')
+const { createSandbox, FakeAgent, spawnProc } = require('../../../../../integration-tests/helpers')
 
 describe('IAST - code_injection - integration', () => {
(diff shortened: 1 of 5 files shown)


packages/dd-trace/test/appsec/iast/code_injection.integration.spec.js
```diff
@@ -53,24 +42,79 @@ describe('IAST - code_injection - integration', () => {
   async function testVulnerabilityRepoting (url) {
     await axios.get(url)
 
-    return agent.assertMessageReceived(({ headers, payload }) => {
-      expect(payload[0][0].metrics['_dd.iast.enabled']).to.be.equal(1)
-      expect(payload[0][0].meta).to.have.property('_dd.iast.json')
+    let iastTelemetryReceived = false
+    const checkTelemetry = agent.assertTelemetryReceived(({ headers, payload }) => {
```
uurien, line 46: I don't understand why these tests are working.
The request is finished when we are subscribing to the agent, why are we able to check the logs that happened before? Could this be a source of flakyness?
I think that we should set these asserts before the `axios.get(url)`

uurien: Ok, I can see more code implemented like this, I think that it is working because there is at least small delay betewen the vuln is detected and reported and the logs are send to the agent.

uurien (approved):

uurien (approved):

PR #5165, opened 2025-01-29
lazily load SDK APIs

bengl:
See title

### DIFF
--- a/packages/dd-trace/src/proxy.js
+++ b/packages/dd-trace/src/proxy.js
@@ -10,13 +10,34 @@ const telemetry = require('./telemetry')
 const nomenclature = require('./service-naming')
 const PluginManager = require('./plugin_manager')
 const remoteConfig = require('./appsec/remote_config')
-const AppsecSdk = require('./appsec/sdk')
 const dogstatsd = require('./dogstatsd')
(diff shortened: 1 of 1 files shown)


packages/dd-trace/src/proxy.js
```diff
@@ -217,8 +238,16 @@ class Tracer extends NoopProxy {
         const prioritySampler = appsecStandalone.configure(config)
         this._tracer = new DatadogTracer(config, prioritySampler)
         this.dataStreamsCheckpointer = this._tracer.dataStreamsCheckpointer
-        this.appsec = new AppsecSdk(this._tracer, config)
-        this.llmobs = new LLMObsSDK(this._tracer, this._modules.llmobs, config)
+        if (config.appsec.enabled) {
+          this.appsec = new lazyClasses.AppsecSdk(this._tracer, config)
+        } else {
+          this.appsec = new lazyClasses.NoopAppsecSdk()
+        }
```
simon-id, line 245: AppSec SDK needs to be available even if AppSec is disabled. I think this is breaking that

simon-id: (the noopAppsecSdk is used when tracing is disabled)

PR #5167, opened 2025-01-29
Lazy load Remote Config in proxy.js

simon-id:
### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/appsec/index.js
+++ b/packages/dd-trace/src/appsec/index.js
@@ -2,7 +2,6 @@
 
 const log = require('../log')
 const RuleManager = require('./rule_manager')
-const remoteConfig = require('./remote_config')
 const {
   bodyParser,
   cookieParser,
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/appsec/index.js
```diff
@@ -53,7 +52,7 @@ function enable (_config) {
 
     RuleManager.loadRules(_config.appsec)
 
-    remoteConfig.enableWafUpdate(_config.appsec)
+    global._ddtrace?._modules?.rc?.module?.enableWafUpdate(_config.appsec)
```
rochdev, line 55: Wouldn't diagnostics channel be more appropriate at that point?

rochdev: Otherwise the common pattern in other places would be a noop, although I'm starting to lean towards DC more and more as a way to decouple things in a way that is more well, decoupled, and also makes for easier bundling and detection of private access.

packages/dd-trace/src/proxy.js
```diff
@@ -25,7 +24,7 @@ class LazyModule {
 
   enable (...args) {
     this.module = this.provider()
-    this.module.enable(...args)
+    return this.module.enable(...args)
```
rochdev, line 27: Why the added return?

simon-id:
superseded by https://github.com/DataDog/dd-trace-js/pull/5256

PR #5168, opened 2025-01-29
update tap to 18

rochdev:
### What does this PR do?

Update tap to 18.

### Motivation

So that we can use `tap replay --reporter=junit` to export JUnit reports in addition to the current report.

### Additional Notes

We have to use tap 18.5.0 and mocha-globals 1.13.0 specifically because 18.6.0/1.14.0 dropped support for Node 18.


rochdev:
Closing since it turns out we would need a more recent version as the last version with Node 18 support is broken.

PR #5179, opened 2025-01-30
downgrade eslint to support node 18

rochdev:
### What does this PR do?

Downgrade eslint to support Node 18

### Motivation

Linting otherwise doesn't work on some supported versions which makes working on them a pain.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,8 +116,8 @@
   },
   "devDependencies": {
     "@apollo/server": "^4.11.0",
-    "@eslint/eslintrc": "^3.1.0",
-    "@eslint/js": "^9.11.1",
+    "@eslint/eslintrc": "^2.1.4",
+    "@eslint/js": "^8.57.0",
     "@msgpack/msgpack": "^3.0.0-beta3",
     "@stylistic/eslint-plugin-js": "^2.8.0",
     "@types/node": "^16.0.0",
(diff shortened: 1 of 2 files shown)


watson:
As far as I understand ESLint v9 works on the following Node.js releases: `^18.18.0`, `^20.9.0`, or `>=21.1.0`. I assume you're primarily making this change to allow running ESLint on for example Node.js 18.0.0? If so, what's the use-case for doing that?

rochdev:
> I assume you're primarily making this change to allow running ESLint on for example Node.js 18.0.0? If so, what's the use-case for doing that?

These are versions we support, so we have to work on them sometimes. When that happens, `yarn install` stops working because the dependency cannot be installed.

```
error @eslint/eslintrc@3.1.0: The engine "node" is incompatible with this module. Expected version "^18.18.0 || ^20.9.0 || >=21.1.0". Got "18.0.0"
error Found incompatible module.
```

Having to ignore engines is annoying and could even fail if any post-install script is present.

watson:
Hmm I see... I would really like to avoid having to downgrade ESLint if we can get around this edge case using `--ignore-engines`. I understand that this might be an issue in some cases, as you also bring up. However, as long as ESLint doesn't has a post-install script and you're not required to actually run ESLint on Node.js 18.0.0, I think this is the better of the two options.

The reason is, that I think we benefit from the newer ESLint because of [the improved rules](https://eslint.org/blog/2024/04/eslint-v9.0.0-released/) and by downgrading we would therefore be worse off.

rochdev:
Closing in favour of #5183

PR #5183, opened 2025-01-31
add yarnrc with ignore-engines config

rochdev:
### What does this PR do?

Add `.yarnrc` with `ignore-engines true` config.

### Motivation

When using old Node versions, ESLint otherwise fails to install.

### Additional Notes

Supersedes #5179

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,0 +1,1 @@
+ignore-engines true


watson:
It would be nice if we could remove this file in CI. That way we'd be notified if we by accident installed a module that didn't support any of the versions we test in CI

rochdev:
> It would be nice if we could remove this file in CI. That way we'd be notified if we by accident installed a module that didn't support any of the versions we test in CI

I don't think it's necessary for 2 reasons:

1) We already explicitly use `--ignore-engines` in CI, so this is basically just redundant and only also applies it when working locally.
2) We could have an explicit job that checks that. Short of that, we have integration tests that use a sandbox that install dd-trace like a user would (which does not include the file)

bengl (approved):

PR #5185, opened 2025-01-31
[test optimization] Fix `cy.window` for multi origin tests

juan-fernandez:
### What does this PR do?
Make it so that `dd-trace` does not crash cypress tests that do multi origin testing.

### Motivation
Tests that do multi origin testing, such as (got from [cypress blog](https://www.cypress.io/blog/cypress-9-6-0-easily-test-multi-domain-workflows-with-cy-origin)):

```javascript
it('navigates', () => {
  cy.visit('/')
  cy.get('h1').contains('My Homepage')
  cy.origin('www.acme.com', () => {
  cy.visit('/history/founder')
    cy.get('h1').contains('About our Founder, Marvin Acme') // 👍
  })
})
```

can't safely access `cy.window()` on `after` hooks. 

This change makes it so that we try to grab the first's origin window. Additionally, we `try / catch` the reading of `window` so that it's safer. There's no way to `try / catch` the `cy.window()` operation, so this is what we're left with.

### Plugin Checklist

- [x] Unit tests.

### DIFF
--- a/integration-tests/cypress/cypress.spec.js
+++ b/integration-tests/cypress/cypress.spec.js
@@ -75,7 +75,7 @@ moduleTypes.forEach(({
   describe(`cypress@${version} ${type}`, function () {
     this.retries(2)
     this.timeout(60000)
-    let sandbox, cwd, receiver, childProcess, webAppPort
+    let sandbox, cwd, receiver, childProcess, webAppPort, secondWebAppServer
 
     if (type === 'commonJS') {
       testCommand = testCommand(version)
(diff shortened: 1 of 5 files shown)


integration-tests/cypress/cypress.spec.js
```diff
@@ -1636,5 +1640,56 @@ moduleTypes.forEach(({
         })
       })
     })
+
+    // cy.origin is not available in old versions of Cypress
+    if (version === 'latest') {
```
tonyredondo, line 1645: how is this going to behave for new releases? why not pin it to an specific version or a >=?
It would be nice to document why `latest` and not a version number.

tonyredondo (approved):

integration-tests/cypress/cypress.spec.js
juan-fernandez: newer cypress releases will be tested. I could add a specific test for cypress `9.6.0`, where `cy.origin` was added, but I don't think it's worth it. We normally test our integrations in the oldest version we support and the newest one released

PR #5193, opened 2025-02-03
ci: pin github actions by hash and update via dependabot

xopham:
### What does this PR do?

- **Add dependabot for github actions**
- **Pin all actions by hash**

### Motivation

Pinning 3rd-party GitHub Actions by commit SHA makes them less vulnerable to compromise of the 3rd party. To avoid outdating and non-verbosity, versions are commented after the SHA and updating via dependabot is introduced that will automatically update the commented version tag as well.

In case of a false commit SHA, this change could break the corresponding workflow. Typically, this does not cause major interruptions, but it can for example affect a release pipeline and require restart causing delays.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      gh-actions-packages:
+        patterns:
+          - "*"
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -11,7 +11,7 @@ jobs:
   actionlint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - uses: ./.github/actions/node/setup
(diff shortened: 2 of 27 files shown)


simon-id (approved):
Seen reasonable enough :+1: I'm approving but please wait until you get approval from other teams as I'm not the only stakeholder of this repo :)

bengl (approved):

rochdev:
@xopham This change is extremely noisy as it repeats values in dozens if not hundreds of places. It's also unclear whether the update strategy even works as we already have some inconsistencies in versions for the same action.

Any way to centralize those in some local composite action instead that would then be updated in a single place? I tried to do it in #5537, and while it works great for things like `codecov` it doesn't seem to work for `checkout` since a checkout is actually needed before being able to run a local action to begin with. Maybe those could be moved to some sort of trusted branch of repository that is centralized? Open to ideas, but I definitely don't like having these hashes all over the YAMLs.

xopham:
@rochdev I am not certain what you mean by noisy. Do you consider the updating noisy because the same dependency is used in multiple places? The repetition is because the dependencies are called in different places and there is no centralized `*.lock` file or anything for GitHub Actions, but they are always "imported" by their full name and version. That's just how GHA is designed to work. I do not think that GitHub Actions has a proper modularization approach and you can try to build composite actions or reusable workflows for common patterns, but it's not aimed at centralizing dependencies and probably leads to complications.
However, the updating is not a problem because all values in all places are properly handled, as can be seen in #5533 . Both dependabot and renovate bot know how to handle the GHA updates (including the commented version).
I would personally recommend to not update these overly often, hence the [monthly interval for dependabot](https://github.com/DataDog/dd-trace-js/blob/master/.github/dependabot.yml#L14), but that is a matter of taste.

rochdev:
It touches a lot of files which could result in things like conflicts etc. Ideally we'd centralize _most_ versions (which I already did for things like installing Node) but I don't know how that could be achieved for `checkout` since the code needs to be checked out to use any local action. We could commit it elsewhere but then the version would no longer be deterministic, so it might just be something we'll have to accept.

PR #5197, opened 2025-02-03
update ci nightly schedules to run 3 times

rochdev:
### What does this PR do?

Update CI nightly schedules to run 3 times.

### Motivation

1) Find flaky tests faster.
2) Figure out if a nightly failure is because of flakiness or if something is broken by comparing the 3 runs.

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -5,7 +5,9 @@ on:
   push:
     branches: [master]
   schedule:
-    - cron: "0 4 * * *"
+    - cron: 0 4 * * *
+    - cron: 20 4 * * *
+    - cron: 40 4 * * *
 
 jobs:
   actionlint:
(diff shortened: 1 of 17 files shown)


watson (approved):
It's a little unclear to me where we'd see the "result" of these runs. The GitHub UI is really bad at getting an overview over cron jobs runs. Do we have a dashboard?

In any case, the PR looks good 👍

simon-id (approved):

rochdev:
> It's a little unclear to me where we'd see the "result" of these runs. The GitHub UI is really bad at getting an overview over cron jobs runs. Do we have a dashboard?

Indeed the UI is not great at summarizing multi-workflow projects like ours. For now the noise will mostly come from emails when a run fails (same as today) but later on we'll get the data in a dashboard. There is ongoing work to streamline reporting/alerting, but I'm setting this up in advance so that we can start working on flakiness right away. I guess for now I'll act as the forwarder when jobs fail.

PR #5203, opened 2025-02-04
simplify storage.js, adding a bunch of comments for clarity

bengl:
### What does this PR do?
* Make DatadogStorage extend from AsyncLocalStorage
* Add comments everywhere

### Motivation
This file is a confusing chunk of code, and it's worth spending a few minutes now to explain it as best as possible.

### DIFF
--- a/packages/datadog-core/src/storage.js
+++ b/packages/datadog-core/src/storage.js
@@ -2,66 +2,81 @@
 
 const { AsyncLocalStorage } = require('async_hooks')
 
-class DatadogStorage {
-  constructor () {
-    this._storage = new AsyncLocalStorage()
-  }
-
-  disable () {
-    this._storage.disable()
-  }
-
+/// This is exactly the same as AsyncLocalStorage, with the exception that it
+/// uses a WeakMap to store the store object. This is because ALS stores the
+/// store object as a property of the resource object, which causes all sorts
+/// of problems with logging and memory. We substitute the `store` object with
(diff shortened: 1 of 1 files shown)


rochdev (approved):

packages/datadog-core/src/storage.js
```diff
@@ -2,66 +2,81 @@
...
-  disable () {
-    this._storage.disable()
-  }
-
+/// This is exactly the same as AsyncLocalStorage, with the exception that it
+/// uses a WeakMap to store the store object. This is because ALS stores the
+/// store object as a property of the resource object, which causes all sorts
+/// of problems with logging and memory. We substitute the `store` object with
+/// a "handle" object, which is used as a key in a WeakMap, where the values
+/// are the real store objects.
```
watson, line 10: I haven't seen this tripple slash commenting style before - is that on purpose?

packages/datadog-core/src/storage.js
```diff
@@ -2,66 +2,81 @@
...
     stores.set(handle, store)
-    this._storage.enterWith(handle)
-  }
-
-  exit (callback, ...args) {
-    this._storage.exit(callback, ...args)
+    super.enterWith(handle)
   }
 
+  // This is method is a passthrough to the real `getStore()`, so that, when we
```
watson, line 18:
```suggestion
  // This method is a passthrough to the real `getStore()`, so that, when we
```

PR #5205, opened 2025-02-05
[DI] Add source map support

watson:
### What does this PR do?

Resolve probe location using a source map if available.

### Motivation

A lot of users will transpile their source code before deploying. For example if they develop in TypeScript. Without source map support, these users will not be able to use Dynamic Instrumentation reliably, as the probe file/line might not match a file/line in the running app.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


watson:
* **#5205** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5205?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5205?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
```diff
@@ -98,17 +99,21 @@ async function addBreakpoint (probe) {
...
-      lineNumber = await processScriptWithInlineSourceMap({ file, line, sourceMapURL })
+      lineNumber = await getSourceMappedLine(url, source, line, sourceMapURL)
     } catch (err) {
-      log.error('Error processing script with inline source map', err)
+      log.error('Error processing script with source map', err)
+    }
+    if (lineNumber === null) {
+      log.error('Could not find generated position for %s:%s', url, line)
+      lineNumber = line
     }
```
watson, line 117: Hopefully the new algorithm is more stable in regards to finding the right line. If we still can't find it, shouldn't this code just throw so it's aborted?

juan-fernandez: I wanted to avoid throwing so that the next test may have a chance to still set a probe

watson: Shouldn't you then at least return early and not try to attach the breakpoint to a line number is most likely incorrect?

watson: I know you try-catch around the code that adds the breakpoint takes care of that for you, but if you don't think it will work, I think it makes the purpose of the code more readable if we return early in that case

juan-fernandez: > Shouldn't you then at least return early and not try to attach the breakpoint to a line number that is most likely incorrect?

you're right. I remember finding a situation where the source map was there, `processScriptWithInlineSourceMap` failed, but setting the probe in the original line number worked 😓. I'm definitely missing something here

packages/datadog-plugin-cucumber/src/index.js
```diff
@@ -251,7 +252,12 @@ class CucumberPlugin extends CiPlugin {
           const { file, line, stackIndex } = probeInformation
           this.runningTestProbe = { file, line }
           this.testErrorStackIndex = stackIndex
-          // TODO: we're not waiting for setProbePromise to be resolved, so there might be race conditions
+          const waitUntil = Date.now() + BREAKPOINT_SET_GRACE_PERIOD_MS
+          while (Date.now() < waitUntil) {
+            // TODO: To avoid a race condition, we should wait until `probeInformation.setProbePromise` has resolved.
+            // However, Cucumber doesn't have a mechanism for waiting asyncrounously here, so for now, we'll have to
+            // fall back to a fixed syncronous delay.
+          }
```
juan-fernandez, line 260: same comment as https://github.com/DataDog/dd-trace-js/pull/5205/files#r1954345426

juan-fernandez (approved):
thanks for worrying about test optimization code and for finding this nice trick 😄 . LGTM!

PR #5206, opened 2025-02-05
Remove legacyStorage in favor of namespaced storages

bm1549:
### What does this PR do?

- Removes `legacyStorage` in favor of a namespaced `'legacy'` storage
- Adds JSDoc comments to `storage.js` so my IDE knows how to code for me

### Motivation

This gets us one step closer to getting rid of `getHandle` as we are now fully converted to namespaced storages

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/datadog-core/index.js
+++ b/packages/datadog-core/index.js
@@ -1,5 +1,5 @@
 'use strict'
 
-const { storage, SPAN_NAMESPACE } = require('./src/storage')
+const { storage } = require('./src/storage')
 
-module.exports = { storage, SPAN_NAMESPACE }
+module.exports = { storage }
--- a/packages/datadog-core/index.js
+++ b/packages/datadog-core/index.js
@@ -1,5 +1,5 @@
 'use strict'
 
-const storage = require('./src/storage')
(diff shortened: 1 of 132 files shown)


rochdev (changes requested):
First thank you for taking on this massive PR to address tech debt.

With that said, one of the goals of this API is to decouple storages. Using a shared environment variable actually introduces coupling. It also differs of usage in other places for the similar `channel` API. A more consistent pattern would be to simply use a string, and possibly combining that with a single instance per file (or even an instance shared in multiple files of the same component). This would also help with performance by avoiding multiple lookups at runtime. Basically, a storage instance should be ideally looked up only at most once per file, and it should not pollute a core module with information about the caller.

It's also worth noting that removing `getHandle` would require an actual span store, which this isn't. In that sense even the namespace name is incorrect. It would be called something like "default" or "global", but I think calling it "legacy" explicitly would be better to make it obvious that its usage going forward is discouraged and alternatives are to be considered.

rochdev (approved):

bm1549:
Bypassing merge protections bc mongoose is a broken test (cc @rochdev )

Issue #5111, opened 2025-01-15
[BUG]: Express tracing errors then stops sending traces


vernak2539:
### Tracer Version(s)

5.29.0, v5.29.1, v5.30.0

### Node.js Version(s)

22.11.0

### Bug Report

I originally put this in https://github.com/DataDog/dd-trace-js/issues/5092, but separated into it's own issue.

Since upgrading to the affected versions, we're finding that tracing stops after a bit of time. It's quite arbitrary, but it seems to be after 4-6 hours. When our application is redeployed, the tracing begins again. I've put a screenshot below to illustrate. While the screenshot shows the `express.middleware` operation this also happens for the `express.request` operation.

![Image](https://github.com/user-attachments/assets/8f176e3a-6f2d-46f5-9a85-da2430fdbaac)

Versions less than the affected versions (I've tested 5.13.0, 5.21.0, 5.26.0, 5.28.0) don't seem to have this issue.

I've just enabled the logs and debug modes for our app to find out more, but initially there was no indication output as to why the tracing stopped.

I should also note that this happened in our staging and production environments. While our staging environment doesn't always have traffic, our production environment does. Additionally, both staging and production have health check endpoints that are hit by our ALBs, which should show up in tracing. During the times of no tracing, not even the health check endpoint is visible in tracing.

I've been able to find the logs, which are below. It seems the `dd-trace` module is encountering errors for some reason. The log:

```
2025-01-13T22:08:17.196Z DD_TRACE-INFO: Disabling plugin: undefined
```

Feels like the cause for tracing to be stopped.

Maybe it's related to https://github.com/DataDog/dd-trace-js/issues/4987. But since https://github.com/DataDog/dd-trace-js/issues/4987 deals with different affected versions, maybe it's different.

**Note**: The log `Package 'next' was loaded before dd-trace! This may break instrumentation.` is present in working versions. The `next.request` operation works fine in previous versions and in the affected versions. Our **start command** below which should load the instrumentation before next (we use a express custom server with next):

```
$ DD_TRACE_DEBUG=true DD_TRACE_LOG_FILE=ddtrace_logs.log NODE_ENV=production node --import ./src/node-instrumentation.mjs src/server/index.mjs
```

### Reproduction Code

```js
// ./src/server/node-instrumentation.mjs 

import tracer from 'dd-trace';

  tracer.init({
    hostname: process.env.STATSD_HOST,
    service: 'app',
    startupLogs: true,
    debug: true,
[...]

PR #5209, opened 2025-02-05
remove ability to propagate baggage on its own to avoid app crashes

ida613:
### What does this PR do?
Fixes this [issue](https://github.com/DataDog/dd-trace-js/issues/5111) at the expense of not allowing customers to propagate baggage on its own.
This is a quick fix for now to avoid crashing customer applications, we will add a more involved solution to provide customers with the full functionality of baggage in a later PR.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/opentracing/propagation/text_map.js
+++ b/packages/dd-trace/src/opentracing/propagation/text_map.js
@@ -598,7 +598,7 @@ class TextMapPropagator {
   }
 
   _extractBaggageItems (carrier, spanContext) {
-    if (!carrier) return
+    if (!spanContext) return
     const baggages = carrier.baggage.split(',')
     for (const keyValue of baggages) {
       if (!keyValue.includes('=')) {
--- a/packages/dd-trace/src/opentracing/propagation/text_map.js
+++ b/packages/dd-trace/src/opentracing/propagation/text_map.js
@@ -325,9 +325,7 @@ class TextMapPropagator {
       if (context === null) {
         context = extractedContext
         if (this._config.tracePropagationExtractFirst) {
-          if (this._hasPropagationStyle('extract', 'baggage') && carrier.baggage) {
-            this._extractBaggageItems(carrier, context)
-          }
+          this._extractBaggageItems(carrier, context)
           return context
         }
       } else {
@@ -347,9 +345,7 @@ class TextMapPropagator {
       }
     }
 
-    if (this._hasPropagationStyle('extract', 'baggage') && carrier.baggage) {
-      this._extractBaggageItems(carrier, context)
-    }
+    this._extractBaggageItems(carrier, context)
 
     return context || this._extractSqsdContext(carrier)
   }
@@ -598,6 +594,8 @@ class TextMapPropagator {
   }
 
   _extractBaggageItems (carrier, spanContext) {
+    if (!this._hasPropagationStyle('extract', 'baggage')) return
+    if (!carrier.baggage) return
     if (!spanContext) return
     const baggages = carrier.baggage.split(',')
     for (const keyValue of baggages) {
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/opentracing/propagation/text_map.js
```diff
@@ -325,6 +325,9 @@ class TextMapPropagator {
       if (context === null) {
         context = extractedContext
         if (this._config.tracePropagationExtractFirst) {
+          if (this._hasPropagationStyle('extract', 'baggage') && carrier.baggage) {
+            if (context) this._extractBaggageItems(carrier, context)
```
rochdev, line 329: Since the check for `context` is now always needed, let's move it to `_extractBaggageItems` instead of nesting conditions.

packages/dd-trace/test/opentracing/propagation/text_map.spec.js
```diff
@@ -451,7 +451,7 @@ describe('TextMapPropagator', () => {
       expect(spanContextD._baggageItems).to.deep.equal({})
     })
 
-    it('should extract baggage when it is the only propagation style', () => {
+    it('should not extract baggage when it is the only propagation style', () => {
```
rochdev, line 454: Let's skip the test instead since the expected behaviour is not to skip extraction, it's just a temporary limitation.

ida613: @bm1549 what do you think?

bm1549: More test coverage is nearly always better imo

What about if we just add a comment on top that describes that this is because of a temporary limitation? Alternatively, we can comment the test if we really don't want to run it, but that's my least preferred option

@rochdev - if none of that is acceptable we can delete the test

rochdev: > What about if we just add a comment on top that describes that this is because of a temporary limitation?

Good enough, I just want to avoid a situation where adding the feature would fail a test, and then a reviewer (or other implementor) is not sure if they're breaking something that was supposed to work differently (because of a test testing a bug)

bm1549: On a similar note, I wanted to avoid the situation where we re-introduce the bug without fixing the edge case and not see this test failing to stop us

packages/dd-trace/src/opentracing/propagation/text_map.js
ida613: updated

packages/dd-trace/test/opentracing/propagation/text_map.spec.js
ida613: added a comment

bm1549 (dismissed):

packages/dd-trace/src/opentracing/propagation/text_map.js
```diff
@@ -596,6 +598,7 @@ class TextMapPropagator {
   }
 
   _extractBaggageItems (carrier, spanContext) {
+    if (!carrier) return
```
rochdev, line 601: Shouldn't this be `!spanContext`?

packages/dd-trace/src/opentracing/propagation/text_map.js
```diff
@@ -325,6 +325,9 @@ class TextMapPropagator {
       if (context === null) {
         context = extractedContext
         if (this._config.tracePropagationExtractFirst) {
+          if (this._hasPropagationStyle('extract', 'baggage') && carrier.baggage) {
```
rochdev, line 328: I just noticed that this condition could also be moved to `_extractBaggageItems`.

rochdev (approved):

PR #5212, opened 2025-02-05
chore(graphql): only stringify graphql error extension attributes in span event if not a native type

wconti27:
Only stringify attributes if they are not of type number or boolean. Fixes new System Test added by @marcotc 

### What does this PR do?
Small fix

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/datadog-plugin-graphql/src/utils.js
```diff
@@ -26,8 +26,15 @@ function extractErrorIntoSpanEvent (config, span, exc) {
 
   if (config.graphqlErrorExtensions) {
     for (const ext of config.graphqlErrorExtensions) {
+      const value = exc.extensions[ext]
+
       if (exc.extensions?.[ext]) {
```
tlhunter, line 31: Why does the `exc.extensions?.[ext]` check on line 31 have a defensive `?.` but the one on 29 doesn't?
This must mean the case where `exc.extensions` doesn't exist must not be tested.

At any rate shouldn't line 31 look like this?

```js
if (value) {
```

packages/datadog-plugin-graphql/src/utils.js
```diff
@@ -26,8 +26,15 @@ function extractErrorIntoSpanEvent (config, span, exc) {
...
-        attributes[`extensions.${ext}`] = exc.extensions[ext].toString()
+        // We should only stringify the value if it is not of type number or boolean
+        if (typeof value === 'number' || typeof value === 'boolean') {
+          attributes[`extensions.${ext}`] = value
+        } else {
+          attributes[`extensions.${ext}`] = value.toString() // Convert to string
```
tlhunter, line 36: What values do you expect `value` to be? An instance of `Error`? It might be safer to use `String(value)` but maybe it doesn't matter. `String(null)` is better than `null.toString()` however your truthy check already avoids that issue.

marcotc: If `String(value)` presents the most human readable value for an arbitrary object (compared to `value.toString()`), we should go with that.

tlhunter: It's more of a defensive thing but it's probably overkill. Like in the following case:

```js
x = null
x.toString() // throws
String(x) // safe

x = {}
x.toString() // safe
String(x) // safe
```

It's probably overkill because I can't actually think of a case where it will fail with the existing `if` check. Objects have a default `toString` method but even if they don't the `String()` conversion will fail anyway:

```
> x = Object.create(null)
[Object: null prototype] {}
> x.toString()
Uncaught TypeError: x.toString is not a function
> String(x)
Uncaught TypeError: Cannot convert object to primitive value
    at String (<anonymous>)
> typeof x.toString
'undefined'
> y = {}
{}
> typeof y.toString
'function'
```

packages/datadog-plugin-graphql/src/utils.js
wconti27: Good catch, I placed that line in the wrong place, fixing.

packages/datadog-plugin-graphql/src/utils.js
wconti27: The values can be any type. These are just additional attributes to error objects, which a user can define. I will change this code to use `String(value)` instead.

tlhunter (approved):

marcotc (approved):

PR #5213, opened 2025-02-06
Update DD_TAGS Parsing to Meet Spec

khanayan123:
### What does this PR do?
Updates DD_TAGS for comma and space-separated tags (including empty values) to align with the agents handling of DD_TAGS

Note:
This PR is marked as a semver-minor since the updates to the tagger should not be considered a breaking change

### Motivation
Config consistency across SDKs. Updated spec for DD_TAG outlined here: https://docs.google.com/document/d/1kI-gTAKghfcwI7YzKhqRv2ExUstcHqADIWA4-TZ387o/edit?usp=sharing


slimee:
We would love this one to be merged! 
❤️ 🙏🏽 ❤️

khanayan123:
Failing system test is due to not adding new config to telemetry intake. Gonna wait till that PR is merged in to move this PR to ready to review

packages/dd-trace/src/config.js
```diff
@@ -673,6 +674,7 @@ class Config {
       DD_TRACE_CLIENT_IP_HEADER,
       DD_TRACE_DYNAMODB_TABLE_PRIMARY_KEYS,
       DD_TRACE_ENABLED,
+      DD_TRACE_EXPERIMENTAL_ENABLED,
```
rochdev, line 677: This seems to be missing a word? What experimental feature is enabled?

packages/dd-trace/src/format.js
```diff
@@ -212,7 +212,9 @@ function extractError (trace, error) {
 function addTag (meta, metrics, key, value, nested) {
   switch (typeof value) {
     case 'string':
+      // edge case to align want to align with the agent by allowing DD_TAGS with empty values
```
rochdev, line 215: This should be one line down.

packages/dd-trace/src/tagger.js
```diff
@@ -5,25 +5,44 @@ const log = require('./log')
...
   if (!carrier || !keyValuePairs) return
 
   if (Array.isArray(keyValuePairs)) {
-    return keyValuePairs.forEach(tags => add(carrier, tags))
+    return keyValuePairs.forEach(tags => add(carrier, tags, parseOtelTags, parseSpaceSeparatedTags))
   }
   try {
     if (typeof keyValuePairs === 'string') {
-      const segments = keyValuePairs.split(',')
+      let segments
```
rochdev, line 27: The tagger is already pretty heavy, I think it would be worth having a benchmark for it if there isn't already one, so see how much overhead we're adding here. Ideally, this whole logic would be reworked to actually improve the overhead even though it does more things.

packages/dd-trace/src/tagger.js
```diff
@@ -5,25 +5,44 @@ const log = require('./log')
...
+          : keyValuePairs.split(separator)
+      } else {
+        segments = keyValuePairs.split(',')
+      }
+
       for (const segment of segments) {
         const separatorIndex = parseOtelTags ? segment.indexOf('=') : segment.indexOf(':')
-        if (separatorIndex === -1) continue
+        if (separatorIndex === -1) {
+          if (parseSpaceSeparatedTags) {
```
rochdev, line 41: It would probably be more efficient to have an entirely different code path when this option is on. Benchmarks would show us that though (although it's worth noting that benchmarks don't always show all the possible decisions that v8 can take for a piece of code)

rochdev (changes requested):
Did we consider normalizing in the opposite direction, where the agent logic would become simpler instead of the client libraries logic becoming more complicated?

packages/dd-trace/src/config.js
khanayan123: Currently this is a catch all flag to enable all experimental features hidden behind the flag, but there is a proposal out right now to evolve this to also allow for targeting  of which features to enable

packages/dd-trace/src/tagger.js
khanayan123: It would be difficult to implement a completely separate code path without duplicating a significant amount of the existing code

packages/dd-trace/src/tagger.js
khanayan123: I added a benchmark in benchmark/core.js if more is needed do let me know

khanayan123:
> Did we consider normalizing in the opposite direction, where the agent logic would become simpler instead of the client libraries logic becoming more complicated?

No because this is a tracer tracer led initiative to standardize config settings on a common definition. For DD_TAGS we used an exiting spec that was compatible with both comma & space separated formats

packages/dd-trace/src/tagger.js
rochdev: I don't think those benchmarks are actually used by any check, it would be better to add a sirun benchmark that adds a tag through the public API.

packages/dd-trace/src/tagger.js
rochdev: Yeah that is definitely not ideal, but it's better speed-wise to avoid going through unneeded code, even if it means duplication. However, if the benchmarks don't show any significant slowdown it would definitely be fine to keep it all together. Basically I'm not blocking on separating, it's just one possible way to tackle the problem.

This type of optimization is something @BridgeAR excels at, so pinging him as well if he also wants to take a look, especially since tagger is called in pretty much the hottest code path of the library (multiple times for every span).

rochdev: Turns out space support is only needed when the environment variable is read, so the tagger doesn't need to be changed, support can be added only in the config class.

packages/datadog-plugin-express/test/index.spec.js
```diff
@@ -1349,10 +1349,8 @@ describe('Plugin', () => {
               expect(spans[0]).to.have.property('resource', 'GET')
               expect(spans[0].meta).to.have.property('http.status_code', '404')
               expect(spans[0].meta).to.have.property('component', 'express')
-              expect(spans[0].meta).to.not.have.property('http.route')
-
-              done()
-            })
+              expect(spans[0].meta).to.have.property('http.route', '')
```
rochdev, line 1352: Does it make sense to report this when empty? If there is a use case for that in the general case it's fine, but our own auto-instrumentation should not be reporting empty tags as these will show up in the UI.

packages/dd-trace/src/plugins/util/git.js
```diff
@@ -328,22 +328,33 @@ function getGitMetadata (ciMetadata) {
...
+  const entries = [
+    [GIT_REPOSITORY_URL],
+    filterSensitiveInfoFromRepository(repositoryUrl || sanitizedExec('git', ['ls-remote', '--get-url'])),
+    [GIT_COMMIT_AUTHOR_DATE], authorDate,
+    [GIT_COMMIT_AUTHOR_NAME], ciAuthorName || authorName,
+    [GIT_COMMIT_AUTHOR_EMAIL], ciAuthorEmail || authorEmail,
+    [GIT_COMMIT_COMMITTER_DATE], committerDate,
+    [GIT_COMMIT_COMMITTER_NAME], committerName,
+    [GIT_COMMIT_COMMITTER_EMAIL], committerEmail,
+    [GIT_TAG], tag
```
juan-fernandez, line 348: I understand these are pairs, but why put the tags in a single element array? 🤔  We could do something like

```javascript
const entries = [
    [GIT_REPOSITORY_URL, filterSensitiveInfoFromRepository(repositoryUrl || sanitizedExec('git', ['ls-remote', '--get-url']))],
    [GIT_COMMIT_AUTHOR_DATE, authorDate],
    [GIT_COMMIT_AUTHOR_NAME, ciAuthorName || authorName],
    [GIT_COMMIT_AUTHOR_EMAIL, ciAuthorEmail || authorEmail],
    [GIT_COMMIT_COMMITTER_DATE, committerDate],
    [GIT_COMMIT_COMMITTER_NAME, committerName],
    [GIT_COMMIT_COMMITTER_EMAIL, committerEmail],
    [GIT_TAG, tag]
]

return Object.fromEntries(entries.filter(([, value]) => value !== undefined))
```

and get rid of the `for` below

juan-fernandez (changes requested):
minor comment in https://github.com/DataDog/dd-trace-js/pull/5213/files#r1975216077. Only other concern is being able to empty strings to our backend, which I think is fine, but I wonder if this is a side effect you're aware of

packages/dd-trace/src/plugins/util/git.js
BridgeAR: It's faster to use the simple array as less intermediate array allocations are needed and also less iterations (filter is n and fromEntries is also n. So 2n instead of 1).

BridgeAR:
@juan-fernandez 
> Only other concern is being able to empty strings to our backend, which I think is fine, but I wonder if this is a side effect you're aware of

Yes, it was originally changed by chance as it seems. I would normally expect empty values to go through as it might be intentional. We did not verify how other libraries handle it, so we could do that first.

packages/dd-trace/src/config.js
khanayan123: More so for allowing breaking features before a major release

zacharycmontoya:
> @juan-fernandez
> 
> > minor comment in https://github.com/DataDog/dd-trace-js/pull/5213/files#r1975216077. Only other concern is being able to empty strings to our backend, which I think is fine, but I wonder if this is a side effect you're aware of
> 
> @zacharycmontoya tested sending empty tags to the backend & verified it does support it

Yes I looked into tags with empty strings as values, and searching/querying for these tags is generally supported in the Datadog product so this is acceptable for us to generate. It's also arguably a better user experience when they configure `DD_TAGS=cluster:${CLUSTER_NAME}` and the CLUSTER_NAME environment variable evaluates to nothing. In this case, the user will see `cluster:""` in the UI which is more noticeable than `cluster` not appearing in the UI.

packages/dd-trace/src/config.js
```diff
@@ -714,8 +714,8 @@ class Config {
     const env = setHiddenProperty(this, '_env', {})
     setHiddenProperty(this, '_envUnprocessed', {})
 
-    tagger.add(tags, OTEL_RESOURCE_ATTRIBUTES, true)
-    tagger.add(tags, DD_TAGS)
+    parseSpaceSeparatedTags(tags, handleOtel(OTEL_RESOURCE_ATTRIBUTES))
+    parseSpaceSeparatedTags(tags, DD_TAGS)
```
BridgeAR, line 718: Nit: after looking at it again, we could change `parseSpaceSeparatedTags` to only change the format and move `tagger.add()` out again. That way the method only does a single thing which is always good :)

packages/dd-trace/src/plugins/util/web.js
```diff
@@ -476,7 +476,8 @@ function addResponseTags (context) {
   const { req, res, paths, span, inferredProxySpan } = context
 
   if (paths.length > 0) {
-    span.setTag(HTTP_ROUTE, paths.join(''))
+    const pathStr = paths.join('')
+    if (pathStr.length > 0) span.setTag(HTTP_ROUTE, paths.join(''))
   }
```
BridgeAR, line 481:
```suggestion
  const route = paths.join('')
  if (route) {
    span.setTag(HTTP_ROUTE, route)
  }
```

Nit: that way join is only called once and I expect the average case to be that paths contains at least one entry. So we would always do the length check anyway.

BridgeAR (dismissed):
This looks very good to me and fixes a couple of system tests for Node.js :)

BridgeAR (approved):
Still LGTM

rochdev (approved):
Code LGTM, but why did some seemingly unrelated tests have to be changed? (for example IAST)

packages/dd-trace/src/plugins/util/git.js
juan-fernandez: I'm not sure why to use an array at all? 

I also think readability is more important than turning 2n into n, given that this is executed _once_ per test session (per `yarn test` command). In any case, could we avoid `Object.fromEntries` and the array with something like:

```javascript
const entries = [
   GIT_REPOSITORY_URL,
   filterSensitiveInfoFromRepository(repositoryUrl || sanitizedExec('git', ['ls-remote', '--get-url'])),
   GIT_COMMIT_AUTHOR_DATE, authorDate,
   GIT_COMMIT_AUTHOR_NAME, ciAuthorName || authorName,
   GIT_COMMIT_AUTHOR_EMAIL, ciAuthorEmail || authorEmail,
   GIT_COMMIT_COMMITTER_DATE, committerDate,
   GIT_COMMIT_COMMITTER_NAME, committerName,
   GIT_COMMIT_COMMITTER_EMAIL, committerEmail,
   GIT_TAG, tag
]

for (let i = 0; i < entries.length; i += 2) {
  const tag = entries[i]
  const value = entries[i + 1]
  if (value) {
    tags[tag] = value
  }
}
```

khanayan123:
> Code LGTM, but why did some seemingly unrelated tests have to be changed? (for example IAST)

Some flaky tests were just improved to run better

packages/dd-trace/test/appsec/iast/code_injection.integration.spec.js
```diff
@@ -27,7 +27,7 @@ describe('IAST - code_injection - integration', () => {
 
   after(async function () {
     this.timeout(60000)
-    await sandbox.remove()
+    await sandbox?.remove()
```
BridgeAR, line 30: Drive-by fix: if the before would fail, sandbox is going to be undefined. This came up in a CI, so we just fixed that and increased the timeout a tad.

juan-fernandez (approved):
thanks

PR #5214, opened 2025-02-06
Clean up ESLint config

watson:
### What does this PR do?
    
- Allow the use of Node.js APIs in Node.js v18.0.0 and above
- Remove disabled rules that were too agressive (replace with inline ignore comments or fix lint errors)
- Scope mocha rules to only apply to test files
- Rename `package.json` script `lint-fix` to `lint:fix`
- Use modern style of extending imported flat config files
- Delete unused `.rslintrc.json` files
  - Nested ESLint config files are not supported with the new flat config file standard
  - They should now all be covered by the root `eslint.config.mjs` file
- Name all config groups (useful for debugging)

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


watson:
* **#5214** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5214?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5214?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

package.json
```diff
@@ -117,7 +117,7 @@
   "devDependencies": {
     "@apollo/server": "^4.11.0",
     "@eslint/eslintrc": "^3.1.0",
-    "@eslint/js": "^9.11.1",
+    "@eslint/js": "^8.57.1",
```
watson, line 120: This version of is actually not compatible with the major version of ESLint we use (v8). Once we upgrade to v9, we can upgrade this as well

packages/dd-trace/test/encode/0.5.spec.js
```diff
@@ -35,7 +35,7 @@ describe('encode 0.5', () => {
         example: 1
       },
       start: 123123123123123120,
-      duration: 456456456456456456,
+      duration: 4564564564564564,
```
watson, line 38: Changed because I enabled `no-loss-of-precision` rule

benchmark/sirun/appsec-iast/insecure-bank.js
```diff
@@ -1,5 +1,5 @@
 const http = require('http')
```
IlyasShabi, line 1: We have a lot of files without `"use strict"`. Do you think it matters to add an ESLint rule for this?

watson: I've noticed that as well. I will make a follow up PR to lint for that

IlyasShabi (approved):
LGTM

benchmark/sirun/appsec-iast/insecure-bank.js
watson: Upcoming draft PR: https://github.com/DataDog/dd-trace-js/pull/5218

eslint.config.mjs
```diff
@@ -48,47 +57,43 @@ export default [
 
     settings: {
       node: {
-        version: '>=16.0.0'
+        // Used by `eslint-plugin-n` to determine the minimum version of Node.js to support.
+        // Normally setting this in the `package.json` engines field is enough, but when we have more than one active
+        // major release line at the same time, we need to specify the lowest version here to ensure backporting will
+        // not fail.
+        version: '>=18.0.0'
```
simon-id, line 64: didn't we agree we'd keep v4 "backportable" in case we need to, and thus keep v16.0.0 for now ?

rochdev: > didn't we agree we'd keep v4 "backportable" in case we need to, and thus keep v16.0.0 for now ?

Given a few recent commits completely dropping support for it, I think the hope is that we'd be able to only backport critical fixes if absolutely necessary, so this PR wouldn't apply.

simon-id (approved):

PR #5216, opened 2025-02-06
Enable recommended rules for eslint-plugin-n

watson:
### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


watson:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5216?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5215** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5215?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5216** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5216?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5216?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5214** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5214?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

benchmark/sirun/plugin-bluebird/index.js
```diff
@@ -4,6 +4,7 @@ if (Number(process.env.USE_TRACER)) {
   require('../../..').init()
 }
 
+// eslint-disable-next-line n/no-missing-require
```
rochdev, line 7: Most occurrences of this are in benchmarks and tests. Should that rule maybe not apply to those?

watson: Good point. I've disabled them 👍

index.js
```diff
@@ -1,3 +1,5 @@
 'use strict'
 
+// TODO: It shouldn't be necessary to disable n/no-unpublished-require - Research
+// eslint-disable-next-line n/no-unpublished-require
```
watson, line 4: Will be fixed in https://github.com/DataDog/dd-trace-js/pull/5274

esbuild.js
```diff
@@ -1,3 +1,5 @@
 'use strict'
 
+// TODO: It shouldn't be necessary to disable n/no-unpublished-require - Research
+// eslint-disable-next-line n/no-unpublished-require
```
watson, line 4: Will be fixed in https://github.com/DataDog/dd-trace-js/pull/5274

packages/datadog-plugin-fetch/test/integration-test/server.mjs
```diff
@@ -1,3 +1,5 @@
+/* eslint n/no-unsupported-features/node-builtins: ['error', { ignores: ['fetch'] }] */
```
tlhunter, line 1: The docs are a little confusing. It seems to suggest that the engines field of package.json is used to check node compat (in our case v18). But those same docs also seem to suggest this plugin is stuck in the world of 13.2.0?

https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unsupported-features/node-builtins.md

Overall it seems bad that we need to add such a line to the top of files to use a built in global. Like we'll make a change, push it, CI fails, add the rule, push again.

watson: I agree it's easy to be confused. We are not using `eslint-plugin-node`, but `eslint-plugin-n`, which exists exactly because `eslint-plugin-node` is no longer maintained. The right docs for this is:

https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/no-unsupported-features/node-builtins.md

There is a few places we use features newer than v18.0.0, but we feature detect, e.g. `if (globalThis.fetch) { ... }`. The ESLint plugin is not clever enough to detect this and complains, but I think that's better than we accidentally used that API without the surrounding if-block.

There's also a few places in tests where we use need to add these ESLint comments. And you could argue we might just turn this plugin off in tests 🤷 Or alternatively configure it to be the latest v18.x in tests instead of v18.0.0 as it is in the rest of our published source. That way we could get a rid of a few of these comments for those locations. Do you think that would be worth while, is it good enough as it is?

BridgeAR: I guess limiting the linter for everything unrelated to tests would be good

BridgeAR (dismissed):
LGTM. Limiting the linter for everything not test related is probably a good idea. I don't think the plugin provides a huge benefit in there.

watson:
@BridgeAR I discovered a problem regarding `eslint-plugin-n` and `eslint-config-standard`. It seems that `eslint-config-standard` bundles `eslint-plugin-n`, and so any scope we set for which files `eslint-plugin-n` should cover is overridden the moment we enable `eslint-config-standard`. I've long wanted to ditch `eslint-config-standard` since it's using a lot of deprecated rules and otherwise doesn't follow modern ESLint standards. Maybe this is the final straw. But in any case, figuring out how to deal with this is probably a prerequisite for this PR.

BridgeAR:
That makes sense. Standard is sadly not properly maintained since a long time. +1 on removing it.
Should we just do it in this PR?

BridgeAR:
Seems like there is an issue running the linter:

ReferenceError: n is not defined
    at file:///home/runner/work/dd-trace-js/dd-trace-js/eslint.config.mjs?mtime=1740645489770:55:8

init.js
```diff
@@ -2,9 +2,13 @@
 
 /* eslint-disable no-var */
 
+// TODO: It shouldn't be necessary to disable n/no-unpublished-require - Research
+// eslint-disable-next-line n/no-unpublished-require
 var guard = require('./packages/dd-trace/src/guardrails')
 
 module.exports = guard(function () {
+  // TODO: It shouldn't be necessary to disable n/no-unpublished-require - Research
+  // eslint-disable-next-line n/no-unpublished-require
```
watson: Will be fixed in https://github.com/DataDog/dd-trace-js/pull/5320

packages/datadog-plugin-fetch/test/integration-test/server.mjs
watson: I updated the rules, so that in tests we're allowed to use experimental Node.js APIs. This meant that I could get rid of most of these comments in tests files.

eslint.config.mjs
```diff
@@ -328,6 +341,18 @@ export default [
       }],
       'import/no-extraneous-dependencies': 'error',
       'n/no-restricted-require': ['error', ['diagnostics_channel']],
+      'n/hashbang': 'off', // TODO: Enable this rule once we have a plan to address it
+      'n/no-process-exit': 'off', // TODO: Enable this rule once we have a plan to address it
+      'n/no-unsupported-features/node-builtins': ['error', {
+        ignores: [
+          'Response',
```
watson, line 348: Can we also add `Request` to this list? If so, we can remove a few ESLint ignore comments.

BridgeAR: We could have a look at it later?

BridgeAR (approved):
LGTM

eslint.config.mjs
watson: Added in #6059

PR #5227, opened 2025-02-07
fix logging when extracting from w3c traceparent

rochdev:
### What does this PR do?

Fix logging when extracting from w3c traceparent.

### Motivation

When W3C support was added, the logging was not updated, so right now an empty object is logged.

### Additional Notes

I also added logging the propagation styles, so that if headers are received from multiple sources we know which ones are used.


BridgeAR (commented):
Could you please add a test as well?

BridgeAR (approved):
LGTM, thank you for adding the generic tests!

PR #5229, opened 2025-02-07
add span events as a top level field for v0.4 encoding

wconti27:
### What does this PR do?
Formatting Changes in this PR:
- Changes variable naming in format.js to be less confusing

Actual Changes:
- Updates format.js to format span events as a top-level field all the time.
- Updates encoder classes to format span events given the encoder type. v0.5 encodes the events within the tags. V0.4 encodes the events within the tags if `DD_TRACE_NATIVE_SPAN_EVENTS=true`, else it encodes the span events within tags.
- Adds `DD_TRACE_NATIVE_SPAN_EVENTS` env var that defaults to `false`

### Motivation
Matches Otel Implementation. Being done in parallel across languages.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


rochdev (changes requested):
Checking the `/info` endpoint doesn't work because it needs to be done on a per-connection basis since each request could otherwise end up on a different agent with a different version and different capabilities, and by that point it's too late to make the decision in most cases because the data has already been generated/formatted. Because of that, the only way this can be implemented for 0.4 is with a tag, and if we want a more efficient data model it needs to be on a newer protocol version that is configured manually by the user.

tlhunter:
@rochdev do you mean that a customer might setup a round robin proxy in front of multiple agents of different versions?

Do we officially support swapping out agent versions on this fly like this?

rochdev:
It's not a weird edge case, and as mentioned above, we hit an issue within 24h of trying to implement the `/info` endpoint. Doing this right requires a lot of complexity that doesn't even handle every use case properly, whereas alternatives are much simpler and will work 100% of the time, so I don't see why we should go with the subpar option.

wconti27:
> Checking the `/info` endpoint doesn't work because it needs to be done on a per-connection basis since each request could otherwise end up on a different agent with a different version and different capabilities, and by that point it's too late to make the decision in most cases because the data has already been generated/formatted. Because of that, the only way this can be implemented for 0.4 is with a tag, and if we want a more efficient data model it needs to be on a newer protocol version that is configured manually by the user.

@rochdev AFAIK this was settled with @marcotc, and the decision was that as long as there's a config option allowing us to turn this feature off, all is well. Can you re-review, I added the configuration so now this feature can be turned off if desired.

packages/dd-trace/src/encode/0.5.js
```diff
@@ -11,6 +11,11 @@ function formatSpan (span) {
 }
 
 class AgentEncoder extends BaseEncoder {
+  constructor (writer, limit = SOFT_LIMIT) {
+    super(writer, limit)
+    this._format = 'v0.5'
+  }
```
tlhunter, line 17:
```suggestion
  constructor () {
    super(...arguments)
    this._format = 'v0.5'
  }
```

packages/dd-trace/src/encode/0.5.js
```diff
@@ -1,7 +1,7 @@
 'use strict'
 
 const { truncateSpan, normalizeSpan } = require('./tags-processors')
-const { AgentEncoder: BaseEncoder } = require('./0.4')
+const { AgentEncoder: BaseEncoder, SOFT_LIMIT } = require('./0.4')
```
tlhunter, line 4:
```suggestion
const { AgentEncoder: BaseEncoder } = require('./0.4')
```

packages/dd-trace/src/encode/0.4.js
```diff
@@ -319,4 +329,4 @@ class AgentEncoder {
   }
 }
 
-module.exports = { AgentEncoder }
+module.exports = { AgentEncoder, SOFT_LIMIT }
```
tlhunter, line 332:
```suggestion
module.exports = { AgentEncoder }
```

packages/dd-trace/src/format.js
```diff
@@ -30,11 +30,13 @@ const map = {
...
 
-function format (span) {
+function format (span, exporter) {
+  const encoder = exporter?._writer?._encoder
+
   const formatted = formatSpan(span)
 
   extractSpanLinks(formatted, span)
-  extractSpanEvents(formatted, span)
+  extractSpanEvents(formatted, span, encoder, exporter?._agentSupportsTopLevelSpanEvents)
```
tlhunter, line 39: Since this value is only ever read from the outside should we really prefix it with an underscore as that signifies it's a private value?

packages/dd-trace/src/exporters/agent/index.js
```diff
@@ -3,6 +3,7 @@
 const { URL, format } = require('url')
 const log = require('../../log')
 const Writer = require('./writer')
+const { fetchAgentInfo } = require('../common/util')
 
 class AgentExporter {
```
tlhunter, line 8: There's sort of three states here. The first is that we don't know if the agent supports top level span events. The second is that we know it doesn't, and the third is that we know it does. So maybe we set a default value of null which is falsey to help with debugging.
```suggestion
class AgentExporter {
  _agentSupportsTopLevelSpanEvents = null
```
We need to decide if the _ prefix makes sense before applying this change.

tlhunter: Also, any idea if there is a risk of shenanigans if messages are sent before the time it takes to get agent info?

packages/dd-trace/src/exporters/agent/index.js
```diff
@@ -14,6 +15,21 @@ class AgentExporter {
...
+          if (err) {
+            this._agentSupportsTopLevelSpanEvents = false
+          } else {
+            this._agentSupportsTopLevelSpanEvents = agentInfo?.span_events === true
+          }
+        })
+      } catch {
+      // pass
+      }
+    }
```
tlhunter, line 31: Overall I have concerns about the exporter class triggering an agent lookup. It seems like with this pattern we could end up with a situation where many different corners of the tracer make simultaneous calls to the info endpoint instead of sharing the results. E.g. CI Vis obviously makes this call at some point too. Making a bunch of redundant HTTP requests would likely annoy users.

An easy fix might be to wrap the `fetchAgentInfo` function in something which consults a cache to see if the info response already exists. The cache could have an expiration of 1 second or so. When building such a cache it's easy to accidentally allow two parallel lookups to still trigger two outgoing requests so you'd need to account for that too. E.g. first cache lookup fails and an outbound request is generated. Then second cache lookup fails and a second outbound request is generated before the first finishes. Instead you'd need to note that an outbound request is in progress and await that resolving. There might be a library that handles this for us.

It would also be worth exposing a function to force the cache to clear. Maybe that would help with Roch's concern about switching agents.

tlhunter (changes requested):

packages/dd-trace/src/config.js
```diff
@@ -593,6 +593,7 @@ class Config {
     this._setValue(defaults, 'instrumentation_config_id', undefined)
     this._setValue(defaults, 'trace.aws.addSpanPointers', true)
     this._setValue(defaults, 'trace.dynamoDb.tablePrimaryKeys', undefined)
+    this._setValue(defaults, 'fetchAgentInfo', true)
```
rochdev, line 596: This should only be true by default in SSI as discussed. This is because in SSI we have better guarantees for what agent is installed where and with what version. In other environments, agents can be installed elsewhere, can change versions whenever, can be in the middle of a rolling update with multiple versions at the same time, etc.

packages/dd-trace/src/exporters/agent/index.js
wconti27: I don't think there's a problem if trace data is sent before the agent info is returned. In this case, the property would be considered falsey, so we would not do the top-level encoding.

rochdev (changes requested):

packages/dd-trace/src/exporters/agent/index.js
```diff
@@ -14,6 +15,21 @@ class AgentExporter {
       port
     }))
 
+    this._agentSupportsTopLevelSpanEvents = false
+    if (this._config.fetchAgentInfo) {
+      try {
+        fetchAgentInfo(this._url, (err, agentInfo) => {
```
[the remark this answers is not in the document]
marcotc: We are removing fetchAgentInfo, as per latest RFC changes for Node.

packages/dd-trace/src/format.js
```diff
@@ -103,18 +105,26 @@ function extractSpanEvents (trace, span) {
...
     }
   }
-  if (events.length > 0) { trace.meta.events = JSON.stringify(events) }
+  if (events.length > 0) {
+    if (encoder?._format === 'v0.4' && agentSupportsTopLevelSpanEvents) {
+      // the agent supports formattedSpan.span_events natively as a top-level field as of 7.63.0
+      formattedSpan.span_events = events
+    } else {
+      // v0.5 case && backwards compatibility for v0.4 where the agent is older than 7.63.0
+      formattedSpan.meta.events = JSON.stringify(events)
```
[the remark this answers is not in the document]
marcotc: > Also, this could conflict extremely easily with user tags given the very generic name. It should probably be something like meta[_dd.events].

This unfortunately was already implemented a few years ago, across all tracers, setting the tag to `"events"`, we are just not changing it here. :/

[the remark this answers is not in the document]
marcotc: > What happens for users that have 0.4? Are they going to have a degraded experience? Wouldn't that be an argument for making this work properly with tags?

I'm assuming you mean v0.5 here: yes, we should add agent support to whitelist this tag, specially for v0.5 users. And probably move it to something like `_dd.span_events` instead of `events`, like you suggested. I'll bring this up as follow up work, to improve support for v0.5 users.

packages/dd-trace/src/agent_discovery/agent_discovery.js
```diff
@@ -0,0 +1,169 @@
...
+      }
+    })
+  }
+
+  /**
+   * Starts the periodic fetching of DataDog Agent information.
+   */
+  _startFetching() {
+    // Initial fetch
+    this._fetchAgentInfo()
```
marcotc, line 104: It this a blocking fetch, that could slow down startup? I'm assuming not.

packages/dd-trace/src/config.js
```diff
@@ -761,6 +762,7 @@ class Config {
       DD_TRACING_ENABLED,
       DD_VERSION,
       DD_TRACE_INFERRED_PROXY_SERVICES_ENABLED,
+      DD_FETCH_AGENT_INFO,
```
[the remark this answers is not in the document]
marcotc: Let me check if we have a standard name already in place for languages that implement this feature

packages/dd-trace/src/agent_discovery/agent_discovery.js
tlhunter: You'd basically need to write C++ to get a blocking HTTP request in Node.js

packages/dd-trace/src/config.js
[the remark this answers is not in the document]
marcotc: We should rename it to `DD_TRACE_AGENT_FEATURE_DISCOVERY`.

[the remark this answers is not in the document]
wconti27: removed this env var, no need for it since agent discovery is not being added with this PR

packages/dd-trace/src/format.js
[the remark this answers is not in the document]
wconti27: this code no longer exists.

packages/dd-trace/src/agent_discovery/agent_discovery.js
wconti27: this code no longer exists.

packages/dd-trace/src/exporters/agent/index.js
wconti27: this is no longer relevant given that the setting is now an env configuration

packages/dd-trace/src/exporters/agent/index.js
wconti27: No longer relevant as we are not doing agent discovery

packages/dd-trace/src/config.js
wconti27: no longer relevant, we are just using a config for whether to encode span events top level

packages/dd-trace/src/exporters/agent/index.js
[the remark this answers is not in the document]
wconti27: this was removed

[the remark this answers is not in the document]
wconti27: this code no longer exists.

packages/dd-trace/test/encode/0.4.spec.js
```diff
@@ -184,7 +184,7 @@ describe('encode', () => {
     })
   })
 
-  it('should encode span events', () => {
+  it('should encode span events within tags as a fallback to encoding as a top level field', () => {
```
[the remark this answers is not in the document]
wconti27: Does this still apply? The only scenarios I see are:
- top level encoding is enabled, check that span events are encoded top level correctly
- top level encoding is disabled (default), check that span events within tags correctly.

packages/dd-trace/src/encode/0.4.js
```diff
@@ -8,8 +8,18 @@ const coalesce = require('koalas')
 
 const SOFT_LIMIT = 8 * 1024 * 1024 // 8MB
 
-function formatSpan (span) {
-  return normalizeSpan(truncateSpan(span, false))
+function formatSpan (span, encoder) {
+  span = normalizeSpan(truncateSpan(span, false))
+  // ensure span events are encoded as tags if agent doesn't support top level encoding
```
rochdev, line 13: Why not just pass the config at that point? It would be one less level of indirection.

wconti27: makes sense, fixing

packages/dd-trace/src/encode/0.4.js
```diff
@@ -319,4 +339,72 @@ class AgentEncoder {
...
+        return {
+          type: 4,
+          array_value: convertedArray
+        }
+      } else {
+      // If all elements were unsupported, return undefined
+        return undefined
+      }
+    } else {
+      log.warn(
```
rochdev, line 395: What if this happens on every span? It will be way too verbose for the WARN level. Could it be memoized somehow? I think there is memoization in some other places in dd-trace which hopefully can be reused.

packages/dd-trace/src/encode/0.4.js
```diff
@@ -319,4 +339,72 @@ class AgentEncoder {
...
+        .filter((convertedVal) => convertedVal !== undefined)
+
+      // Only include array_value if there are valid elements
+      if (convertedArray.length > 0) {
+        return {
+          type: 4,
+          array_value: convertedArray
+        }
+      } else {
+      // If all elements were unsupported, return undefined
```
rochdev, line 391:
```suggestion
        // If all elements were unsupported, return undefined
```

packages/dd-trace/src/encode/0.4.js
```diff
@@ -319,4 +339,72 @@ class AgentEncoder {
   }
 }
 
+function formatSpanEvents (span) {
+  for (const spanEvent of span.span_events) {
+    if (spanEvent.attributes) {
+      for (const [key, value] of Object.entries(spanEvent.attributes)) {
+        const newValue = convertSpanEventAttributeValues(key, value)
+        if (newValue !== undefined) spanEvent.attributes[key] = newValue
+        else delete spanEvent.attributes[key] // delete from attributes if undefined
```
rochdev, line 348: if/else is easier to read multiline.

packages/dd-trace/src/encode/0.4.js
wconti27: Lowered to DEBUG log level and added memoization, including tests to ensure memoization works.

packages/dd-trace/src/encode/0.4.js
wconti27: fixed.

tlhunter (approved):

PR #5230, opened 2025-02-07
Inject trace info as comment to MongoDB operation when dbm propagation is enabled. 

lu-zhengda:
### What does this PR do?
This PR introduces `DBM propagation mode` support for MongoDB operations, aligning with the existing functionality in relational database such as PostgreSQL and MySQL. DBM propagation enables linking APM traces with DBM samples through trace comment injection, providing enhanced observability into database operations.

Key Features:
- Trace Propagation via Comments:
The tracer now supports injecting trace metadata into MongoDB commands using the comment field.
- Configurable Propagation Mode:
Users can configure the behavior via the dbmPropagationMode setting:
    - `disabled` – No trace metadata is injected.
    - `service` – Injects only service-level information.
    - `full` – Injects both service and span-level information.

Example Usage:

Set `dbmPropagationMode` in tracer config
```
// Load the dd-trace library for tracing
const tracer = require("dd-trace").init({
  service: "express-mongodb-app", // Your service name
  env: "dev", // Your environment
  ......  // Your other tracer configs
  dbmPropagationMode: "full",
});
```

With DBM propagation enabled, MongoDB commands will include a trace comment similar to:
image
Screenshot 2025-02-07 at 2 49 38 PM

### Motivation
Improved Traceability:
With trace information attached, MongoDB operation samples in DBM can now be linked to APM traces, allowing users to analyze query performance in the context of distributed tracing.
https://datadoghq.atlassian.net/browse/DBMON-4990

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


dalextorres (dismissed):

packages/dd-trace/src/plugins/database.js
```diff
@@ -82,6 +82,40 @@ class DatabasePlugin extends StoragePlugin {
...
+      return command
+    }
+
+    let dbmTraceComment = ''
+
+    const dbmService = this.getDbmServiceName(span, serviceName)
+    const servicePropagation = this.createDBMPropagationCommentService(dbmService, span)
+
+    if (mode === 'service') {
+      dbmTraceComment = `${servicePropagation}`
```
bengl, line 98: This doesn't need to be in a template literal like this. It should suffice to just set it exactly to `servicePropagation`. If we don't know that it's always a string, you can coerce it directly using `String()` if that's what you want to do.
```suggestion
      dbmTraceComment = servicePropagation
```

bengl (changes requested):

packages/dd-trace/src/plugins/database.js
```diff
@@ -82,6 +82,40 @@ class DatabasePlugin extends StoragePlugin {
     }
   }
 
+  injectDbmCommand (span, command, serviceName) {
```
wconti27, line 85: Can we have this new function merged with `injectDbmQuery`, that way we aren't duplicating a bunch of code lines and can maintain 1 function for DBM propagation injection?

packages/dd-trace/src/plugins/database.js
```diff
@@ -82,6 +82,40 @@ class DatabasePlugin extends StoragePlugin {
...
+    } else if (mode === 'full') {
+      span.setTag('_dd.dbm_trace_injected', 'true')
+      const traceparent = span._spanContext.toTraceparent()
+      dbmTraceComment = `${servicePropagation},traceparent='${traceparent}'`
+    }
+
+    if (command.comment) {
+      // if the command already has a comment, append the dbm trace comment
+      if (typeof command.comment === 'string') {
+        command.comment += `,${dbmTraceComment}`
```
wconti27, line 108: Just because of curiosity, why are we appending the DBM comment to the end of the command? As opposed to previous MySQL injection where we prepend it to the beginning of the comment.

wconti27 (changes requested):
The tests and plugin change looks good to me. But we should put the injection functionality within the same function as the existing DBM propagation.

packages/dd-trace/src/plugins/database.js
lu-zhengda: This is due the how MongoDB handles database operations as compared to relational database like mysql. MongoDB database commands are sent as object instead of SQL string. In mysql we chose to prepend comment to the beginning of the SQL statement because we want to avoid appending comment after optional trailing `;`. We don't have such edge cases for MongoDB because `comment` is an attribute in the command object. I chose append so that the tracer comment appears after user comment.

packages/dd-trace/src/plugins/database.js
lu-zhengda: The reason behind a new `injectDbmCommand` is because mongodb commands are `object` instead of SQL in `string`. I can merge the common comment generation into one function to avoid duplicated code.

packages/dd-trace/src/plugins/database.js
tlhunter: Can you add a test that ensures we still work even when the user provides a comment? It looks like the comments can be a string or an array. Can you create a test for both situations? It should be sufficient to add this to only one of the two test files.

Is whatever DBM doing on the backend compatible with both string and array approaches? I'm sure it is but just want to make sure.

Also, with the comment concatenation approach is there any user-supplied comment pattern that can break parsing of the DBM comment? E.g. if the user sets a comment to `foo=` or `,,,` or `asdf"`?

Does the backend only consider the _first_ or _last_ occurrence of DBM comment fields? If it's just the first then the comment should be prepended, and if it's the last it should be appended like you're doing here. E.g. if the user comment is `ddpv=fooboar` does the real `ddpv=1.2.3` appended by the tracer still work as expected?

packages/dd-trace/src/plugins/database.js
wconti27: The refactor looks good to me, thanks for abstracting the common code!

packages/dd-trace/src/plugins/database.js
```diff
@@ -63,23 +63,54 @@ class DatabasePlugin extends StoragePlugin {
     return tracerService
   }
 
-  injectDbmQuery (span, query, serviceName, isPreparedStatement = false) {
+  createDbmComment (span, serviceName, isPreparedStatement = false) {
```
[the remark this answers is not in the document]
wconti27: Good catch, this makes sense to me. Shouldn't be too much effort to make this refactor.

packages/datadog-plugin-mongodb-core/src/index.js
```diff
@@ -24,6 +25,7 @@ class MongodbCorePlugin extends DatabasePlugin {
         'out.port': options.port
       }
     })
+    ops = this.injectDbmCommand(span, ops, service)
```
tlhunter, line 28: The way `injectDbmCommand` is written it seems to mutate an object provided by the user. It's a common bug pattern where we add a feature that mutates a user object, then we release it, and then we get a bug report because someone is seeing their object mutated, then we have to release a fix.

Am I right in assuming this is mutating the object in userland? If not, can you write a test that ensures the user's object isn't mutated? And if so, can you modify `injectDbmCommand` so that it creates a new object and returns that to be shipped off to the database instead of mutating the user's object?

It might be as simple as this:

```js
  injectDbmCommand (span, command, serviceName) {
    const dbmTraceComment = this.createDbmComment(span, serviceName)

    if (!dbmTraceComment) {
      return command
    }
    newCommand = { ...command }
    // newCommand.comment = ...
    return newCommand
  }
```

tlhunter (changes requested):
Left some comments requesting changes

packages/datadog-plugin-mongodb-core/src/index.js
lu-zhengda: good call

docker-compose.yml
```diff
@@ -37,7 +37,7 @@ services:
     ports:
       - "127.0.0.1:6379:6379"
   mongo:
-    image: circleci/mongo:3.6
+    image: circleci/mongo:4.4
```
lu-zhengda, line 40: bump mongo image to 4.4 because prior to 4.4 only str comments are allowed. newer mongo version allows us to test array comments.

packages/dd-trace/src/plugins/database.js
lu-zhengda: > with the comment concatenation approach is there any user-supplied comment pattern that can break parsing of the DBM comment? E.g. if the user sets a comment to foo= or ,,, or asdf"
Any user supplied comments that do not follow the `key='value'` style will be safely ignored. DBM backend extracts the key='value' pair from the comment and leaves the comment untouched.

lu-zhengda: > Is whatever DBM doing on the backend compatible with both string and array approaches?

DBM backend unmarshalls both [string and array](https://github.com/DataDog/dd-go/blob/prod/database-monitoring/apps/dbm-events-processor/decoder/payloads.go#L144) into an array.

wconti27 (approved):
LGTM, thanks for making those changes!

packages/datadog-plugin-mongodb-core/src/index.js
```diff
@@ -34,6 +36,30 @@ class MongodbCorePlugin extends DatabasePlugin {
...
+    }
+
+    // create a copy of the command to avoid mutating the original
+    const dbmTracedCommand = { ...command }
+
+    if (dbmTracedCommand.comment) {
+      // if the command already has a comment, append the dbm trace comment
+      if (typeof dbmTracedCommand.comment === 'string') {
+        dbmTracedCommand.comment += `,${dbmTraceComment}`
+      } else if (Array.isArray(command.comment)) {
```
tlhunter, line 54:
```suggestion
      } else if (Array.isArray(dbmTracedCommand.comment)) {
```

PR #5233, opened 2025-02-07
handles _thenUnwrap for OpenAI's custom APIPromise class

josemussa:
### What does this PR do?
Tracing is failing for requests that include a call to the OpenAI `this.client.beta.chat.completions.parse` method.  

This happens because OpenAI uses a custom promise class. The Datadog team is already aware of this issue, as it was previously addressed for the OpenAI `this.client.chat.completions.create` method. I'm reusing that logic to resolve the current issue.  

This PR implements the same unwrap logic and response handler that is used for the `parse` method, but for the `_thenUnwrap` method.

### Motivation  
When using the OpenAI Node SDK and calling `await this.client.beta.chat.completions.parse`, traces completely break.  

This happens because OpenAI uses a custom promise. This issue was previously fixed for `this.client.chat.completions.create` (the non-beta version). To address this, I abstracted that fix and applied it to `_thenUnwrap`.

The issue is reported here https://github.com/openai/openai-node/issues/1307 

### Plugin Checklist

- [x] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/datadog-instrumentations/src/openai.js
+++ b/packages/datadog-instrumentations/src/openai.js
@@ -329,7 +329,7 @@ for (const shim of V4_PACKAGE_SHIMS) {
 }
 
 function handleUnwrappedAPIPromise (apiProm, ctx, stream, n) {
-  apiProm.then(([{ response, options }, body]) => {
+  return apiProm.then(([{ response, options }, body]) => {
     if (stream) {
       if (body.iterator) {
         shimmer.wrap(body, 'iterator', wrapStreamIterator(response, options, n, ctx))
--- a/packages/datadog-instrumentations/src/openai.js
+++ b/packages/datadog-instrumentations/src/openai.js
@@ -298,44 +298,24 @@ for (const shim of V4_PACKAGE_SHIMS) {
         return ch.start.runStores(ctx, () => {
           const apiProm = methodFn.apply(this, arguments)
 
+          // wrapping `_thenUnwrap` is necessary to handle responses
+          // when OpenAI makes use of this method from the APIPromise class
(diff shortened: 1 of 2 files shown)


josemussa:
Hi @sabrenner, I saw that you’re the main author of this file—would you be open to reviewing this PR when you have a moment? Thanks a lot!

sabrenner:
Hi @josemussa! Yes, I'll be taking a look at this today, thanks for the ping!

sabrenner:
This change LGTM, although I'll refactor the test a bit. Unfortunately CI for community PRs isn't great here haha - I will cherry pick your commits into a new PR to run against CI and merge 🙌 

just also wanna note as an FYI that since it's technically on the beta client, we generally won't be as quick to update support if its behavior changes at all from what we assume, and once the functionality that is supported here (I'm guessing structured output) is passed onto the official client, we'll advertise tracing through the official client instead.

josemussa:
Thanks, @sabrenner! Appreciate you handling the CI situation and the refactor. Let me know if anything else is needed! 🚀

sabrenner:
Hi @josemussa, the PR with the cherry picks has been merged! We should be tracing those beta chat completions now. Will follow up with the release it lands in (it might not be the next patch release, but probably 5.38.0 instead)

josemussa:
Great news! Thank you @sabrenner 🙇

sabrenner:
@josemussa This just landed in [5.38.0](https://github.com/DataDog/dd-trace-js/releases/tag/v5.38.0), feel free to try it out!

PR #5241, opened 2025-02-10
Add missing decrement method to NoopDogStatsDClient

booleangate:
### What does this PR do?

Adds the missing `decrement` method to the `NoopDogStatsDClient` class.

### Motivation

The `NoopDogStatsDClient` gets used in unit tests in consumers of this package.  However, it is missing the `decrement` method which causes those tests to fail because trying to call this method will throw an exception.

### Additional notes

I also manually tested that this works by applying the same change to the version of this package used by my project and confirmed that tests calling `tracer.dogstatsd.decrement` no longer threw exceptions.

### DIFF
--- a/packages/dd-trace/src/noop/dogstatsd.js
+++ b/packages/dd-trace/src/noop/dogstatsd.js
@@ -1,6 +1,8 @@
 module.exports = class NoopDogStatsDClient {
   increment () { }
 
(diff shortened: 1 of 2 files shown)


BridgeAR (approved):
LGTM, thank you for the PR! Is there a reason this is still a draft?

PR #5253, opened 2025-02-12
[MLOB-2233] feat(openai, llmobs): patch `_thenUnwrap` of chat completion api promise instances

sabrenner:
### What does this PR do?
Closes #5233 

Patches the `APIPromise` instance's `_thenUnwrap` returned for `chat.completion` calls.

This is done through instance patching, as opposed to prototype patching, which is normal for this OpenAI instrumentation, but is becoming a bit hectic. Even though this support is temporary, we will want to rework how this instrumentation is done in general. More notes below.

### Motivation
Temporary tracing support for the `client.beta.chat.completions` endpoint by request. This is a one-off, and this support will be removed once the functionality on the beta client (mainly structured output) is on the main client itself.

This is the only `beta` method we will support tracing for.

### Additional Notes
We can't just patch `apiProm._thenUnwrap` and call it a day because `_thenUnwrap` returns a _new_ `APIPromise`, and we a) cannot return a normal `Promise` in it's place, so we b) instead opt to patch the `parse` method on that new `APIPromise`.

This instrumentation really needs a refactor. We're doing a lot of nested instance patching that could be done better by maybe patching the base `APIPromise` along with the `APIClient` to hoist the base API information. I will try to do this in a follow-up PR, but for now, I believe just modifying the current implementation is the best approach.


packages/datadog-plugin-openai/test/index.spec.js
```diff
@@ -3724,6 +3727,55 @@ describe('Plugin', () => {
           }
         })
       }
+
+      if (semver.intersects('>=4.59.0', version)) {
```
tlhunter, line 3731: Is openai >=4.59.0 being installed and tested during CI? I suppose it is since it's the latest but without the withVersions magic used in this file it might get tricky later when you want to test many versions of the package.

tlhunter (approved):

Kyle-Verhoog (approved):

PR #5254, opened 2025-02-12
lazy load runtime metrics only when needed

rochdev:
### What does this PR do?

Lazy load runtime metrics only when needed.

### Motivation

Right now when runtime metrics are disabled, the code is still loaded which impacts startup time.

### DIFF
--- a/packages/dd-trace/src/runtime_metrics/index.js
+++ b/packages/dd-trace/src/runtime_metrics/index.js
@@ -3,19 +3,19 @@
 let runtimeMetrics
 
 const noop = runtimeMetrics = {
-  start () {},
-  stop () {},
-  track () {},
-  boolean () {},
-  histogram () {},
-  count () {},
-  gauge () {},
-  increment () {},
-  decrement () {}
+  start: () => {},
+  stop: () => {},
+  track: () => {},
+  boolean: () => {},
+  histogram: () => {},
+  count: () => {},
+  gauge: () => {},
+  increment: () => {},
+  decrement: () => {}
 }
 
 module.exports = {
-  start (config) {
+  start: (config) => {
     if (!config?.runtimeMetrics) return
 
     runtimeMetrics = require('./runtime_metrics')
--- a/packages/dd-trace/src/runtime_metrics/index.js
+++ b/packages/dd-trace/src/runtime_metrics/index.js
@@ -1,0 +1,39 @@
+'use strict'
+
+let runtimeMetrics
+
+const noop = runtimeMetrics = {
+  start () {},
+  stop () {},
+  track () {},
+  boolean () {},
+  histogram () {},
+  count () {},
+  gauge () {},
+  increment () {},
+  decrement () {}
+}
+
+module.exports = {
+  start (config) {
+    if (!config?.runtimeMetrics) return
+
+    runtimeMetrics = require('./runtime_metrics')
+    runtimeMetrics.start(config)
+  },
+
+  stop: () => {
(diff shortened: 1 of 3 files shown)


packages/dd-trace/src/runtime_metrics/index.js
```diff
@@ -0,0 +1,39 @@
...
+  },
+
+  stop: () => {
+    if (runtimeMetrics === noop) return
+
+    runtimeMetrics.stop()
+    runtimeMetrics = noop
+  },
+
+  track: (...args) => runtimeMetrics.track(...args),
```
bengl, line 32: Rather than barreling functions through like this, maybe it's better to do a prototype switch.

tlhunter:
Does this affect user applications with runtime metrics disabled but who make use of custom metrics?

packages/dd-trace/src/runtime_metrics/index.js
rochdev: Made the change 👍

rochdev:
> Does this affect user applications with runtime metrics disabled but who make use of custom metrics?

No because custom metrics expose the underlying DogStatsD client directly and doesn't rely on runtime metrics and have their own scheduler in `proxy.js`.

packages/dd-trace/src/runtime_metrics/index.js
rochdev: Any remaining concerns?

packages/dd-trace/src/runtime_metrics/index.js
```diff
@@ -0,0 +1,35 @@
+'use strict'
+
+let runtimeMetrics
+
+const noop = runtimeMetrics = {
+  start: () => {},
```
BridgeAR, line 6:
```suggestion
```

This is actually always defined so it won't be needed

packages/dd-trace/src/runtime_metrics/index.js
```diff
@@ -0,0 +1,35 @@
...
+  stop: () => {},
+  track: () => {},
+  boolean: () => {},
+  histogram: () => {},
+  count: () => {},
+  gauge: () => {},
+  increment: () => {},
+  decrement: () => {}
```
BridgeAR, line 14: Super not important nit: any reason not to use

```suggestion
  decrement() {}
```

If we talk about super optimizing startup time, this will safe byes loaded and that need to be read :D

BridgeAR (dismissed):
LGTM. Just left two nits

bengl (approved):

packages/dd-trace/src/runtime_metrics/index.js
```diff
@@ -0,0 +1,34 @@
...
+    Object.setPrototypeOf(module.exports, runtimeMetrics)
+
+    runtimeMetrics.start(config)
+  },
+
+  stop () {
+    runtimeMetrics.stop()
+
+    Object.setPrototypeOf(module.exports, runtimeMetrics = noop)
+  }
```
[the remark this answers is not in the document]
rochdev: Yeah this was really just an optimization to free up memory if the feature is disabled. Some features cannot be disabled at runtime so that made less sense for those other PRs. Happy to remove it if you think keeping the code simpler is a better trade-off.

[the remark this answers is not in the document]
rochdev: I'll merge it like this for now, but happy to open a new PR if you think the optimization is not useful.

PR #5255, opened 2025-02-12
lazy load telemetry metrics only when needed

rochdev:
### What does this PR do?

Lazy load telemetry only when needed.

### Motivation

Right now when telemetry is disabled, the code is still loaded which impacts startup time.

### Additional Notes

For now I only handled this for tracing specifically. There are other places that use telemetry internals directly, for example AppSec, but this will be handled in a second pass later on as right now we are focusing on the default.


BridgeAR:
@rochdev there are quite some conflicts here. Is this something you still want to finish? I would suggest to otherwise close it for now?

rochdev:
Let's close it, most of the benefits were from lazy loading `telemetry/telemetry` and not `telemetry/metrics`, and the former has already landed.

PR #5256, opened 2025-02-12
lazy load public api + profiling + crashtracking + span stats

rochdev:
### What does this PR do?

Lazy load non-tracing public API, profiling, crashtracking and span stats.

### Motivation

These are all costly to import and increase startup time even when not needed.


packages/dd-trace/src/proxy.js
```diff
@@ -33,6 +27,20 @@ class LazyModule {
   }
 }
 
+function lazyProxy (obj, property, getClass, ...args) {
```
bengl, line 30: why not just...

```js
function lazyProxy(obj, property, getClass, ...args) {
  const configurable = true
  const writable = true
  const enumerable = true
  Reflect.defineProperty(obj, property, {
    get () {
    	const RealClass = getClass()
    	const value = new RealClass(...args)
    	Reflect.defineProperty(obj, property, {
    		value,
    		configurable,
    		writable,
    		enumerable
    	})
    	return value
    },
    configurable,
    writable,
    enumerable
  })
}

**Review by bengl — CHANGES_REQUESTED** [review]

File: packages/dd-trace/src/proxy.js (hunk 1 above)
Comment (rochdev) [reply packages/dd-trace/src/proxy.js:30]: There are scenarios where this doesn't work, for example any property defined in the constructor. It's also unclear that this is much faster as we've already established in the past that descriptors can be pretty slow.

Comment (rochdev) [reply packages/dd-trace/src/proxy.js:30]: Oh wait I think I misunderstood what you're doing, basically you're loading when the property from the tracer is read, not the property of the instance. Maybe that makes more sense, let me think about it.

@simon-id I took the proxy idea from your draft PR for AppSec, what was the reasoning for using a Proxy?

Comment (rochdev) [reply packages/dd-trace/src/proxy.js:30]: Updated with a property descriptor instead of a proxy.

File: packages/dd-trace/src/proxy.js [hunk 2]
```diff
@@ -218,11 +233,11 @@ class Tracer extends NoopProxy {
         this._modules.llmobs.enable(config)
       }
       if (!this._tracingInitialized) {
-        const prioritySampler = appsecStandalone.configure(config)
+        const prioritySampler = config.appsec.standalone?.enabled && require('./appsec/standalone').configure(config)
         this._tracer = new DatadogTracer(config, prioritySampler)
         this.dataStreamsCheckpointer = this._tracer.dataStreamsCheckpointer
-        this.appsec = new AppsecSdk(this._tracer, config)
-        this.llmobs = new LLMObsSDK(this._tracer, this._modules.llmobs, config)
+        lazyProxy(this, 'appsec', () => require('./appsec/sdk'), this._tracer, config)
```
simon-id, line 239: this will be lazy loaded even outside of serverless, thus incurring a synchronous cost at runtime. You literally asked to not do this yourself.

You could manually trigger the proxy when not in serverless like that:
```javascript
if (!config._isInServerlessEnvironment()) {
  this.appsec
}
```

packages/dd-trace/src/proxy.js
```diff
@@ -33,6 +27,22 @@ class LazyModule {
   }
 }
 
+function lazyProxy (obj, property, getClass, ...args) {
+  const attributes = { configurable: true, enumerable: true }
```
simon-id, line 31: is it worth reusing an object and using spread operators when you could just hardcode them ?

packages/dd-trace/src/proxy.js
simon-id: Eh no reason except it's less verbose, the proxy destructs itself after first use anyway. Property descriptor should work just the same i assume

packages/dd-trace/src/config.js
```diff
@@ -823,7 +826,11 @@ class Config {
       this._envUnprocessed.peerServiceMapping = DD_TRACE_PEER_SERVICE_MAPPING
     }
     this._setString(env, 'port', DD_TRACE_AGENT_PORT)
-    const profilingEnabledEnv = coalesce(DD_EXPERIMENTAL_PROFILING_ENABLED, DD_PROFILING_ENABLED)
+    const profilingEnabledEnv = coalesce(
+      DD_EXPERIMENTAL_PROFILING_ENABLED,
+      DD_PROFILING_ENABLED,
+      this._isInServerlessEnvironment() ? 'false' : undefined
```
szegedi, line 832: Ooh, I see what you did here. Clever 😃.

For consistency, I'd probably move these to `_applyCalculated`. I know it's not _particularly_ interesting since you said we don't report telemetry but it's still conceptually cleaner. You could just put this in `_applyCalculated()`:
```javascript
    if (this._isInServerlessEnvironment()) {
      this._setValue(calc, 'profiling.enabled', 'false')
      this._setValue(calc, 'crashtracking.enabled', false)
    }
```

szegedi (changes requested):

packages/dd-trace/src/proxy.js
rochdev: Probably not, originally there were more attributes so it made more sense, I can change it.

packages/dd-trace/src/config.js
rochdev: Ironically, I moved them here _for consistency_, as other options are also calculated in this way in `_applyEnv`. For now let's keep it there and we can revisit in a separate PR if needed, although I kind of think it makes sense to be in `_applyEnv` as Serverless is an environment.

packages/dd-trace/src/proxy.js
rochdev: Yeah, after thinking more about it I figured it's better to be consistent, but I don't wanna cause a regression either. Will add a condition.

rochdev: Done.

packages/dd-trace/src/proxy.js
rochdev: Changed it.

packages/dd-trace/src/config.js
szegedi: I thought "Env" in "_applyEnv" meant "environment variables". But you're right that some uses of specifically `_isInServerlessEnvironment` have already snuck in, so by that reasoning I can accept these being there as well. If someone gets sufficiently irritated by it, they can fix it later.

szegedi (dismissed):
Looks good for profiling.

simon-id (dismissed):

sabrenner (approved):
lgtm for llmobs

packages/dd-trace/src/proxy.js
rochdev: @bengl Is the new code fine?

rochdev: @bengl Any remaining concerns?

packages/dd-trace/src/proxy.js
```diff
@@ -33,6 +27,21 @@ class LazyModule {
...
+  const get = () => {
+    const RealClass = getClass()
+    const value = new RealClass(...args)
+
+    Reflect.defineProperty(obj, property, { value, configurable: true, enumerable: true })
+
+    return value
+  }
+
+  if (config?._isInServerlessEnvironment?.() === false) return get()
```
bengl, line 40: So wait, if we're not in severless, we return a value, otherwise we just define a property? That seems odd.

bengl (changes requested):

packages/dd-trace/src/proxy.js
rochdev: The value doesn't need to be returned, I just wanted to return early and returning the statement makes it a one-liner. I can split this into 4 lines if you think it makes the code easier to understand. The important bits is to call `get()` so that the prototype is overridden and then return early.

rochdev: Refactored the code to make the distinction clearer, remove the early return, and no longer return a value unnecessarily.

bengl (approved):

PR #5258, opened 2025-02-12
lazy load telemetry entrypoint only when needed

rochdev:
### What does this PR do?

Lazy load telemetry entrypoint only when needed.

### Motivation

Right now when telemetry is disabled, the code is still loaded which impacts startup time.

### Additional Notes

For now I only handled the entrypoint, but there are also modules using the internals directly which will be addressed in another PR.


packages/dd-trace/src/telemetry/index.js
```diff
@@ -1,393 +1,40 @@
...
-    'sampler.rateLimit': 'DD_TRACE_RATE_LIMIT',
-    queryStringObfuscation: 'DD_TRACE_OBFUSCATION_QUERY_STRING_REGEXP',
-    version: 'DD_VERSION',
-    env: 'DD_ENV',
-    service: 'DD_SERVICE',
-    clientIpHeader: 'DD_TRACE_CLIENT_IP_HEADER',
-    'grpc.client.error.statuses': 'DD_GRPC_CLIENT_ERROR_STATUSES',
-    'grpc.server.error.statuses': 'DD_GRPC_SERVER_ERROR_STATUSES'
-  }
+module.exports = {
```
simon-id, line 13: you're mixing arrow and non arrow function ?

simon-id: i don't see any `this` so why not use only non-arrow functions ?

rochdev: Fixed, and I also did the same fix for #5254

simon-id (approved):

packages/dd-trace/src/telemetry/index.js
```diff
@@ -1,393 +1,40 @@
...
-  for (const change of changes) {
-    const name = nameMapping[change.name] || change.name
+    telemetry.stop()
+    telemetry = noop
+  },
 
-    names.push(name)
-    const { origin, value } = change
-    const entry = { name, value, origin }
+  updateIntegrations: (...args) => telemetry.updateIntegrations(...args),
```
BridgeAR, line 28: I am uncertain why this is handled separately. Does removing this change anything?

packages/dd-trace/test/telemetry/index.spec.js
```diff
@@ -14,6 +14,85 @@ const DEFAULT_HEARTBEAT_INTERVAL = 60000
...
+    expect(telemetry.stop).to.have.been.calledOnce
+  })
+
+  it('should proxy when config is updated', () => {
+    const config = { telemetry: { enabled: true } }
+    const changes = []
+
+    proxy.updateConfig(changes, config)
+
+    expect(telemetry.updateConfig).to.have.been.calledWith(changes, config)
```
BridgeAR, line 92: Please make the test more robust by checking for the whole telemetry object being replaced

BridgeAR (approved):
LGTM % comments

packages/dd-trace/test/telemetry/index.spec.js
rochdev: Turned out that test and its underlying functionality were not even needed to begin with.

BridgeAR (dismissed):
LGTM

bengl (approved):

PR #5259, opened 2025-02-12
lazy load span leak detection only when used

rochdev:
### What does this PR do?

Lazy load span leak detection only when used.

### Motivation

This is only a debugging tool and should only be loaded when used.

### DIFF
--- a/packages/dd-trace/src/opentracing/span.js
+++ b/packages/dd-trace/src/opentracing/span.js
@@ -13,7 +13,6 @@ const log = require('../log')
 const { storage } = require('../../../datadog-core')
 const telemetryMetrics = require('../telemetry/metrics')
 const { channel } = require('dc-polyfill')
-const spanleak = require('../spanleak')
 const util = require('util')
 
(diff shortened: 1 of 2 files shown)


tlhunter (approved):

BridgeAR (dismissed):

packages/dd-trace/src/opentracing/span.js
```diff
@@ -99,7 +98,10 @@ class DatadogSpan {
 
       unfinishedRegistry.register(this, operationName, this)
     }
-    spanleak.addSpan(this, operationName)
+
+    if (tracer?._config?.spanLeakDebug > 0) {
```
BridgeAR, line 102: Please add a TODO entry to replace the tests that cause the tracer from being nullish.
The tracer should AFAIK never be nullish effectively.

rochdev: Added a TODO, is there any other remaining concern?

BridgeAR (approved):

PR #5260, opened 2025-02-12
move otel to datadog context conversion on otel side

rochdev:
### What does this PR do?

Move OTel to Datadog context conversion on OTel side.

### Motivation

OTel can be quite intrusive and heavy, so it's best kept isolated.


BridgeAR (approved):
LGTM

PR #5262, opened 2025-02-13
fix(telemetry): only report configuration with known values

dmehala:
### What does this PR do?
This PR ensures that telemetry configuration is only reported when it has valid values. Since the `value` field is mandatory for configuration entries (as specified in the [schema](https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v2/SchemaDocumentation/Schemas/conf_key_value.md)), reporting entries without values results in invalid payloads that our intake do not or entirely process.

### Motivation
To reduce the number of invalid payloads and ensure the tracer conforms to the expected schema.

### Plugin Checklist
I've done none of the thing below because I'm a clueless cpp engineer but feel free to take over.

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/telemetry/index.js
+++ b/packages/dd-trace/src/telemetry/index.js
@@ -338,7 +338,7 @@ function updateConfig (changes, config) {
   for (const change of changes) {
     const { origin, value } = change
     if (value === undefined) {
-      continue;
+      continue
     }
 
     const name = nameMapping[change.name] || change.name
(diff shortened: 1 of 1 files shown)


packages/dd-trace/src/telemetry/index.js
```diff
@@ -336,10 +336,14 @@ function updateConfig (changes, config) {
   const names = [] // list of config names whose values have been changed
 
   for (const change of changes) {
+    const { origin, value } = change
+    if (value === undefined) {
+      continue;
```
simon-id, line 341:
```suggestion
      continue
```

PR #5263, opened 2025-02-13
ESLint: Require await inside async functions

watson:
### What does this PR do?

Enable the `require-await` ESLint rule. The rule isn't enabled in tests, where the extra created promises doesn't harm anyone.

### Motivation

If a function is marked as async, V8 will create a few extra promises behind the scenes when awaiting it. If the function doesn't itself contain any awaits, these extra promises just adds overhead that isn't needed.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### Additional Notes

Be aware: This might break some things that expect the function to return a promise. We'll have to go over each of the places where the PR removes the `async` keyword and check if the function returns something that should be a promise, but isn't.

There might be a few places, especially places were we use `shimmer.wrap` where we might actually want to keep the `async` keyword!

### DIFF
--- a/benchmark/sirun/get-results.js
+++ b/benchmark/sirun/get-results.js
@@ -134,7 +134,7 @@ async function getResults (gitCommit) {
 async function main () {
   const ref = process.argv.length > 2 ? process.argv[2] : 'HEAD'
   const gitCommit = execSync(`git rev-parse ${ref}`).toString().trim()
-  console.log(JSON.stringify(getResults(gitCommit), null, 4))
+  console.log(JSON.stringify(await getResults(gitCommit), null, 4))
 }
 
 module.exports = {
--- a/packages/datadog-instrumentations/src/apollo-server-core.js
+++ b/packages/datadog-instrumentations/src/apollo-server-core.js
@@ -10,7 +10,7 @@ addHook({ name: 'apollo-server-core', file: 'dist/runHttpQuery.js', versions: ['
   const HttpQueryError = runHttpQueryModule.HttpQueryError
 
   shimmer.wrap(runHttpQueryModule, 'runHttpQuery', function wrapRunHttpQuery (originalRunHttpQuery) {
-    return async function runHttpQuery () {
+    return function runHttpQuery () {
       if (!requestChannel.start.hasSubscribers) {
         return originalRunHttpQuery.apply(this, arguments)
       }
--- a/packages/datadog-instrumentations/src/apollo-server-core.js
+++ b/packages/datadog-instrumentations/src/apollo-server-core.js
@@ -10,7 +10,6 @@ addHook({ name: 'apollo-server-core', file: 'dist/runHttpQuery.js', versions: ['
   const HttpQueryError = runHttpQueryModule.HttpQueryError
 
   shimmer.wrap(runHttpQueryModule, 'runHttpQuery', function wrapRunHttpQuery (originalRunHttpQuery) {
-    // The original function is async, but no need to mark it as such as long as it returns a promise
     return function runHttpQuery () {
       if (!requestChannel.start.hasSubscribers) {
         return originalRunHttpQuery.apply(this, arguments)
--- a/packages/datadog-instrumentations/src/apollo-server-core.js
+++ b/packages/datadog-instrumentations/src/apollo-server-core.js
@@ -10,6 +10,7 @@ addHook({ name: 'apollo-server-core', file: 'dist/runHttpQuery.js', versions: ['
   const HttpQueryError = runHttpQueryModule.HttpQueryError
 
   shimmer.wrap(runHttpQueryModule, 'runHttpQuery', function wrapRunHttpQuery (originalRunHttpQuery) {
+    // The original function is async, but no need to mark it as such as long as it returns a promise
     return function runHttpQuery () {
       if (!requestChannel.start.hasSubscribers) {
         return originalRunHttpQuery.apply(this, arguments)
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -87,7 +87,8 @@ export default [
       'no-console': 'error',
       'no-prototype-builtins': 'off', // Override (turned on by @eslint/js/recommnded)
       'no-unused-expressions': 'off', // Override (turned on by standard)
-      'no-var': 'error' // Override (set to warn in standard)
+      'no-var': 'error', // Override (set to warn in standard)
+      'require-await': 'error'
     }
   },
   {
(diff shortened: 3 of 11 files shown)


watson:
* **#5263** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5263?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5263?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

IlyasShabi (approved):

rochdev:
Actually, can we instead have a rule to _block async/await entirely_? Promises too. We've had a non-written rule about never using promises in the codebase since day 1 because it can slow down `async_hooks` significantly.

watson:
@rochdev I wouldn't mind doing that, but I feel that's a different PR

benchmark/sirun/get-results.js
```diff
@@ -131,7 +131,7 @@ async function getResults (gitCommit) {
   return buildData
 }
 
-async function main () {
+function main () {
```
bengl, line 134: I assume this was never caught because we never run this file directly, so `main()` is never called 😆.

The actual issue here is that `getResults()` should have an `await` in front of it.

bengl (changes requested):
LGTM except that one request change, and I'd like to make sure @juan-fernandez reviews the test-framework instrumentation code.

benchmark/sirun/get-results.js
watson: Fixed 👍

BridgeAR (dismissed):
I used the PR to familiarize myself with a bit more of the code and I looked very deeply into each usage. All of these changes are LGTM

packages/datadog-instrumentations/src/apollo-server-core.js
```diff
@@ -10,6 +10,7 @@ addHook({ name: 'apollo-server-core', file: 'dist/runHttpQuery.js', versions: ['
   const HttpQueryError = runHttpQueryModule.HttpQueryError
 
   shimmer.wrap(runHttpQueryModule, 'runHttpQuery', function wrapRunHttpQuery (originalRunHttpQuery) {
+    // The original function is async, but no need to mark it as such as long as it returns a promise
```
BridgeAR, line 13: Should we really add these comments? While reviewing it, I also thought about it and it seemed fine without, since the method should always return what ever it originally does.
Obviously we may not change that, but it should also never be the case, right?
I am just in favor of not having comments as they are easily outdated, forgotten and often do not help a lot.

juan-fernandez: in this particular case it'll help test opt devs 😄 , as it helps highlighting that async operations are indeed allowed in this wrapper function. Removing it means you either have to remember or check the test framework's original code

juan-fernandez (commented):
test optimization's perspective here.

I'm not against the change but I don't find it particularly useful: the `async`s need to be replaced by a comment or it'll make it harder to know that an async operation is possible in the wrapper. This makes legibility worse IMO. 

As for the performance improvements: I understand the improvement, but is it really applying? These wrapper functions are async because the original ones are async. Example: https://github.com/jestjs/jest/blob/93ca9d3cf6db4b88393ab02b8d0007bcdf30e4bb/packages/jest-reporters/src/CoverageReporter.ts#L107 

is wrapping an `async` function into a sync one what we want here? 

This said, again, I don't oppose the change 😄

tlhunter:
> Actually, can we instead have a rule to block async/await entirely? Promises too. We've had a non-written rule about never using promises in the codebase since day 1 because it can slow down async_hooks significantly.

async/await is totally fine in tests

rochdev:
> async/await is totally fine in tests

Absolutely, by codebase I meant the source code, not the tests, and even that has exceptions today (for example I believe at least profiling has been using promises for a while) but the main reason it was blocked in most places is that historically it would cause issues with `async_hooks`. Since we instrument promises, creating promises means we're paying double the cost (actually more because of GC). As pointed out though, it's technically out of scope of the PR in the sense that I was asking for something else, but I think it would make sense to make sure through linting that we're not using promises at all.

packages/datadog-instrumentations/src/apollo-server-core.js
watson: I was on the fence about this as well. The reason I ended up adding the comments in the end was because of the risk involved. If we ever decided to return something else than the return value of the original function, we needed to ensure that it was wrapped in a promise. If we didn't, something would break. So I went with the better-safe-than-sorry approach 🤷 

> I am just in favor of not having comments as they are easily outdated

Unless the original function is changed to not be async anymore, the comment is never outdated, right?

szegedi: I'd argue it's not necessary for the comment to elaborate on what did the function used to be before the change. "Returns a promise" feels enough to me. Or "Returns a promise, can be awaited.".

If you want a bit more formalism, we can even put a JSDoc annotation on it?
```
/** @returns {Promise<…>} */
```
(I don't know what value type is the returned promise in question supposed to carry, hence ellipsis.)

BridgeAR: I would still remove the comments while I don't feel strongly enough about it that this is in any way blocking!
If you rather keep them, that's fine to me. I'll mark this therefore as resolved.
I just tend to rather use less comments as they just easily get outdated and are my experience also says they are often ignored.

BridgeAR (approved):

packages/datadog-instrumentations/src/apollo-server-core.js
watson: After thinking more about this, I have come to the conclusion, that the comments can be removed. I've updated the PR 👍 

While not all these functions where I added the comments are covered by CI, it's not my fault, and any lack of coverage could result in all sorts of errors going undetected - and if the functions were tested, any chance in async-ness would be flagged.

juan-fernandez: > After thinking more about this, I have come to the conclusion, that the comments can be removed. I've updated the PR 👍

could you keep the test optimization's files comments?

bengl (approved):

packages/datadog-instrumentations/src/apollo-server-core.js
juan-fernandez: I noticed the comments are gone. It's OK, I can add them back in a different PR 😄

watson: @juan-fernandez I think I went over the tests for those and if any of those functions no longer returned a promise, the tests would break if I'm not mistaken - it was based on that that I decided to remove them as it was covered by tests. WDYT?

juan-fernandez: that's good to know. The part where I know that I can slip in an async operation is missing though 😄. It's not extremely necessary but I see no disadvantage in having a small comment

PR #4112, opened 2024-02-27
Fix unhandled rejection in kakfa commit tracking

jbertran:
### What does this PR do?

Fix unhandled promise rejection in kafkajs producer instrumentation.

Taken from #4089 (thanks @Gilbert142!), removed the additional, unnecessary branch - we can do all the work inside the async hook resolution, since we don't have anything to do with commits in case of error.

I've added a test which, for some reason, always passes, despite reproducing the upstream error described in #4039   - an easy way to check is to apply the following patch:

```diff
diff --git a/packages/datadog-instrumentations/src/kafkajs.js b/packages/datadog-instrumentations/src/kafkajs.js
index 4e66f4cf0..9b063fb00 100644
--- a/packages/datadog-instrumentations/src/kafkajs.js
+++ b/packages/datadog-instrumentations/src/kafkajs.js
@@ -81,6 +81,8 @@ addHook({ name: 'kafkajs', file: 'src/index.js', versions: ['>=1.4'] }, (BaseKaf
             })
           )

+          result.then(() => {})
+
           return result
         } catch (e) {
           producerErrorCh.publish(e)
```

Whether or not this is there, the test succeeds, and looking at output yields a caught Kafka error, and no unhandled rejection. I'm probably being thick or missing something - second pair of 👀 welcome

### Plugin Checklist

- [x] Unit tests.

### Security 
Datadog employees:
- [ ] If this PR touches code that signs or publishes builds or packages, or handles credentials of any kind, I've requested a review from `@DataDog/security-design-and-guidance`.
- [x] This PR doesn't touch any of that.

Closes #4089 
Fixes #4039

PR #5270, opened 2025-02-13
Fix KafkaJS instrumentation of orphaned promise

mastermatt:
### What does this PR do?
Refactors querying the cluster ID in the KafkaJS instrumentation, so that unhandled rejections can't occur if the underlying client is unable to connect to the broker. 

### Motivation
Bug added in https://github.com/DataDog/dd-trace-js/pull/4808, released under 5.25.0 
Similar in effect to the bug which was resolved with https://github.com/DataDog/dd-trace-js/pull/4112

The addition of the logic which added the clusterId to publish and consume spans created orphaned promises in `producer` and `consumer` methods which don't become re-exposed until the `send` or `run` methods are called respectively. This results in unhandled rejections, and by default, exits the process if those secondary methods are note called in time and admin methods fail.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### Additional Notes

Reproduction:
```js
require('dd-trace/init')

const { Kafka } = require('kafkajs')
const { config } = require('@/config')

process.on('unhandledRejection', err => {
  console.error('unhandledRejection', err)
  throw err
})

const main = async () => {
  const producer = new Kafka({
    ...config.KAFKA,
    retry: { retries: 0 },
  }).producer()

  await producer.connect() // the error thrown here will be caught safely
  await producer.send({
    messages: [{ value: 'hello world' }],
    topic: 'my.topic',
  })
}

main().catch(err => {
  console.error('boom', err)
})
```
If the above file is run when no Kafka broker is running/accessible, the expected behavior is for the `connect()` to timeout, throw an error, and be caught and logged at the bottom with a "boom" message.

Current behavior will log the expected error from the rejected `connect()`, however, there will also be an unhandled rejection log with a stack trace that originates from the synchronous `producer()` call. 

I'd like to call out that this file has exhibited the same bug twice in the last year and there doesn't seem to be any test file for the KafkaJS instrumentation. 
I'm willing to add tests go along with this PR, but I'd like guidance on test strategy from the DD team.
The PR which added the bug only updated a test suite in `datadog-plugin-kafkajs`, and only covered happy paths. It's unclear to me if any new tests should live in `datadog-plugin-kafkajs` or `datadog-instrumentations/test`.

cc: @wconti27 @bengl @juan-fernandez

### DIFF
--- a/packages/datadog-instrumentations/src/kafkajs.js
+++ b/packages/datadog-instrumentations/src/kafkajs.js
@@ -52,60 +52,57 @@ addHook({ name: 'kafkajs', file: 'src/index.js', versions: ['>=1.4'] }, (BaseKaf
     const send = producer.send
     const bootstrapServers = this._brokers
 
-    const kafkaClusterIdPromise = getKafkaClusterId(this)
-
     producer.send = function () {
       const wrappedSend = (clusterId) => {
-        const innerAsyncResource = new AsyncResource('bound-anonymous-fn')
-
-        return innerAsyncResource.runInAsyncScope(() => {
-          if (!producerStartCh.hasSubscribers) {
-            return send.apply(this, arguments)
-          }
-
-          try {
-            const { topic, messages = [] } = arguments[0]
-            for (const message of messages) {
-              if (message !== null && typeof message === 'object') {
-                message.headers = message.headers || {}
-              }
+        try {
+          const { topic, messages = [] } = arguments[0]
+          for (const message of messages) {
+            if (message !== null && typeof message === 'object') {
+              message.headers = message.headers || {}
             }
-            producerStartCh.publish({ topic, messages, bootstrapServers, clusterId })
-
-            const result = send.apply(this, arguments)
-
-            result.then(
-              innerAsyncResource.bind(res => {
-                producerFinishCh.publish(undefined)
-                producerCommitCh.publish(res)
-              }),
-              innerAsyncResource.bind(err => {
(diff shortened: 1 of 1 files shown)


mastermatt:
I have no idea who to ping to get this reviewed. @watson simply because you have recent activity on the repo and I'm taking a shot in the dark.

BridgeAR:
@mastermatt thanks for the PR and the ping, I am going to try to take a look at it tomorrow. It is strongly appreciated that you opened the PR with the fix. We can't run our CI on external contributions, so we have to port this over.

About the tests: adding a test is definitely required. We do have unit tests and also integration tests in datadog-plugin-kafkajs and that seems the best place for additional tests.

BridgeAR (commented):
Sorry for my review taking much longer than expected.

This definitely looks very good and you are absolutely correct that the error is not properly handled in case of an error. I'll open a PR to fix the issue soon (we can't accept a PR as is, we have to port things and we need to add tests as well).

I am happy to add you as reviewer, if you like.

BridgeAR:
I work on it in https://github.com/DataDog/dd-trace-js/pull/5837. There seems to be another issue about potentially manipulating input data and I optimize the code for higher performance as well.

BridgeAR:
@mastermatt I want to apologize for taking this long to fix this, I just wanted to give you an update that I landed a fix for the issue and it will be released in our upcoming release.

PR #5282, opened 2025-02-17
ESLint: Run on the latest supported Node.js version

watson:
### What does this PR do?

This changes the GitHub action CI workflow that runs the linter, to use the latest supported Node.js version instead of the oldest (currently version 22 instead of version 18).

### Motivation

The linting should produce the same result no matter which version it runs on, so there's no reason to use an older Node.js version for this.  If using a newer version, we might even get some benefits like faster runs or better error messages (thought I haven't verified this).

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/.github/workflows/project.yml
+++ b/.github/workflows/project.yml
@@ -151,7 +151,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-      - uses: ./.github/actions/node/setup
+      - uses: ./.github/actions/node/latest
       - uses: ./.github/actions/install
       - run: yarn lint
 


watson:
* **#5282** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5282?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5282?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

BridgeAR (commented):
I agree that running on a new version would be beneficial. I suggest to use LTS as eslint itself might not support latest

watson:
> I suggest to use LTS as eslint itself might not support latest

The `latest` action in one we control and is currently hardcoded to version `22`. We decide when to bump this. You can see it here: https://github.com/DataDog/dd-trace-js/blob/master/.github/actions/node/latest/action.yml

BridgeAR:
@watson true, while we used 19 and 17 before. Is there a reason why we don't directly specify the Node.js LTS version?

watson:
@BridgeAR A large portion of our GitHub actions already uses the `latest` workflow. Whether or not we should create a separate `latest-lts` action and use that instead for these, I think is a separate, but very valid, discussion. I think that's better suited for another, separate PR.

watson:
Closing - This was implemented as part of #5322

PR #5310, opened 2025-02-24
Fix issues detected by TS type checks

BridgeAR:
### What does this PR do?

I am looking into using TS to detect issues in the repo. These are some of the things that came up.
The options were not used or the JSDoc were off, so this corrects that.

### DIFF
--- a/packages/dd-trace/src/config.js
+++ b/packages/dd-trace/src/config.js
@@ -179,8 +179,7 @@ function validateNamingVersion (versionString) {
  * If a blank path is provided a null is returned to signal that the feature is disabled.
  * An empty array means the feature is enabled but that no rules need to be applied.
  *
- * @param {string} input
- * @returns {[string]|null}
+ * @param {string | string[]} input
  */
 function splitJSONPathRules (input) {
   if (!input) return null
@@ -270,8 +269,7 @@ class Config {
     }
     const PROPAGATION_STYLE_INJECT = propagationStyle(
       'inject',
-      options.tracePropagationStyle,
-      this._getDefaultPropagationStyle(options)
+      options.tracePropagationStyle
     )
 
     validateOtelPropagators(PROPAGATION_STYLE_INJECT)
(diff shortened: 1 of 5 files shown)


packages/dd-trace/src/config.js
```diff
@@ -270,8 +269,7 @@ class Config {
     }
     const PROPAGATION_STYLE_INJECT = propagationStyle(
       'inject',
-      options.tracePropagationStyle,
-      this._getDefaultPropagationStyle(options)
```
bm1549, line 274: Similar concern - this is likely not the correct behavior (even if it equivalent)

@khanayan123 (recent modifier) what is the impact of not including this?

packages/dd-trace/src/datastreams/processor.js
```diff
@@ -13,16 +13,14 @@ const log = require('../log')
...
 const ENTRY_PARENT_HASH = Buffer.from('0000000000000000', 'hex')
 
-const HIGH_ACCURACY_DISTRIBUTION = 0.0075
-
 class StatsPoint {
   constructor (hash, parentHash, edgeTags) {
     this.hash = hash.readBigUInt64BE()
     this.parentHash = parentHash.readBigUInt64BE()
     this.edgeTags = edgeTags
-    this.edgeLatency = new LogCollapsingLowestDenseDDSketch(HIGH_ACCURACY_DISTRIBUTION)
```
bm1549, line 23: Although this is equivalent behavior, I do have some concerns that this is not the correct behavior

@ida613 (author) what is the downside of lower accuracy for sketches?

bm1549 (commented):
Changes LGTM since they are equivalent, just had some concerns about intended behaviors that could be fixed after the fact

packages/dd-trace/src/config.js
BridgeAR: This was fixed in a different way later on by @simon-id in https://github.com/DataDog/dd-trace-js/pull/4580

packages/dd-trace/src/datastreams/processor.js
BridgeAR: I believe the difference does not matter much for now, since we never received complains about this.

bm1549: Sounds good - it can definitely be changed after the fact, so resolving my comment

bm1549 (approved):

PR #5316, opened 2025-02-25
Various improvements to profiling sample label generator

szegedi:
### What does this PR do?
* Allows for `context.context` not being defined
* Bail out faster when it's not present
* Handle `context.asyncId`

### Motivation
A version of pprof-nodejs in the works streamlines few things. As a consequence,  `context.context` can sometimes be undefiined. also, `context.asyncId` will start appearing.

### Additional Notes
Jira: [PROF-11371]

### DIFF
--- a/packages/dd-trace/src/profiling/profilers/wall.js
+++ b/packages/dd-trace/src/profiling/profilers/wall.js
@@ -286,13 +286,22 @@ class NativeWallProfiler {
 
     const labels = { ...getThreadLabels() }
 
-    const { context: { ref }, timestamp } = context
-    const { spanId, rootSpanId, webTags, endpoint } = ref ?? {}
-
     if (this._timelineEnabled) {
       // Incoming timestamps are in microseconds, we emit nanos.
-      labels[END_TIMESTAMP_LABEL] = timestamp * 1000n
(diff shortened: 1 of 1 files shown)


BridgeAR (commented):
Code is LGTM. Can you please add a test as well?

packages/dd-trace/src/profiling/profilers/wall.js
```diff
@@ -286,14 +286,23 @@ class NativeWallProfiler {
...
-      labels[END_TIMESTAMP_LABEL] = timestamp * 1000n
+      labels[END_TIMESTAMP_LABEL] = context.timestamp * 1000n
+    }
+
+    const asyncId = context.asyncId
+    if (asyncId !== undefined && asyncId !== -1) {
+      labels['async id'] = asyncId
     }
 
+    const ref = context.context?.ref
```
BridgeAR, line 299: I guess a comment to explain why the context might be undefined would be useful in this case. I tried to think of even better ways like having a variable name for different context types or something the like, but I couldn't come up with anything great so far

nsavoire (approved):

BridgeAR (approved):

PR #5317, opened 2025-02-25
lazy load fetch instrumentation only when needed

rochdev:
### What does this PR do?

Lazy load fetch instrumentation only when needed.

### Motivation

This was a significant part of the remaining startup time. I'm not sure why as this instrumentation doesn't do much, but it still makes things a lot faster.

### DIFF
--- a/packages/dd-trace/src/serverless.js
+++ b/packages/dd-trace/src/serverless.js
@@ -60,9 +60,18 @@ function getIsAzureFunction () {
   return isAzureFunction
 }
 
+function isInServerlessEnvironment () {
+  const inAWSLambda = process.env.AWS_LAMBDA_FUNCTION_NAME !== undefined
+  const isGCPFunction = getIsGCPFunction()
+  const isAzureFunction = getIsAzureFunction()
+
+  return inAWSLambda || isGCPFunction || isAzureFunction
+}
+
 module.exports = {
   maybeStartServerlessMiniAgent,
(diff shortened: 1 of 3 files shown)


packages/dd-trace/src/serverless.js
```diff
@@ -60,9 +60,18 @@ function getIsAzureFunction () {
   return isAzureFunction
 }
 
+function isInServerlessEnvironment () {
+  const inAWSLambda = process.env.AWS_LAMBDA_FUNCTION_NAME !== undefined
+  const isGCPFunction = getIsGCPFunction()
+  const isAzureFunction = getIsAzureFunction()
+
+  return inAWSLambda || isGCPFunction || isAzureFunction
```
szegedi, line 68: How about we have this folded into the final expression, so it short-circuits. (`getIsGCPFunction` could be rewritten so it short circuits too, possibly first evaluating `isNewerGCPFunction` as I imagine that's more frequent.) I figure microoptimizations are on topic in a PR that deals with startup time :-)

szegedi (approved):
LGTM, some very speculative suggestions for saving few nanos here and there.

packages/dd-trace/src/serverless.js
rochdev: Worth looking into, but for the purpose of this PR, existing code was just copied over so I would consider it out of scope.

PR #5321, opened 2025-02-26
Add first capabilities tagging

Mariovido:
### What does this PR do?

This PR enables the library capabilities which let the user know which features are supported by their setup (programming language / test framework / Datadog library version).

### Motivation

Apart from the mentioned in the previous point we would also like to use this information to prevent users from enabling the features that are not supported by the repository. Also we want to remove the hardcoded constants inside the **web_ui** repository

### Plugin Checklist

- [x] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/integration-tests/cucumber/cucumber.spec.js
+++ b/integration-tests/cucumber/cucumber.spec.js
@@ -46,9 +46,9 @@ const {
   DD_TEST_IS_USER_PROVIDED_SERVICE,
   TEST_MANAGEMENT_ENABLED,
   TEST_MANAGEMENT_IS_QUARANTINED,
-  TAG_TEST_IMPACT_ANALYSIS,
-  TAG_EARLY_FLAKE_DETECTION,
-  TAG_AUTO_TEST_RETRIES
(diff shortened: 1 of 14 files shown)


juan-fernandez (changes requested):
this looks good! I want to rethink our whole approach to `setMetadataTags` though, so I'm blocking for a bit 😄

juan-fernandez:
@Mariovido I've done some changes in https://github.com/DataDog/dd-trace-js/pull/5321/commits/f3f7bf98a7ddbb991e6d48520f715bd530af93bb: 

* slightly simplify test logic (mostly just some really minor improvements)
* renamed the tags to better follow our current standard
* added some `receiver.setSettings` to the tests to make sure we follow the response from the backend 
* changed `setMetadataTags` to `addMetadataTags`. This way the logic is greatly simplified: we just allow our event types to be passed and we just merge whatever we had with the input 

if all tests pass and this looks good to you, we should be good to merge 😄 . Great job!

juan-fernandez (approved):

PR #5322, opened 2025-02-26
CI: Clean up Node.js installation in GitHub actions

watson:
### What does this PR do?

Clean up Node.js installation in GitHub actions
    
- Remove actions installing old Node.js versions no longer used (`14` and `16`)
- Rename `latest` action to `active-lts` to clearly signal intent (a separate `latest` or `current` action that tests non-lts, can be created in a follow up commit)
- Remove `18` action, and replace any usage of it with `oldest-maintenance-lts`
- Remove `setup` action and replace any usage of it with `active-lts`
- Move `install` action to after the first Node.js install action that actually needs it, instead of first installing one Node.js version, only to install another right after
- Change any direct usage of `actions/setup-node` to use our own Node.js installation actions instead, when a matrix is not used.
- A few places where Node.js 18 was used for testing, now use `active-lts` as this was most likely the intent (👉  **reviewers should verify!** 👈 )

### Motivation

- Align how Node.js is installed in our GitHub actions
- Reduce runtime of GitHub actions by not running unnecessary steps
- Make it easier to upgrade to newer major Node.js versions when they are released (with the use of labels like `active-lts` instead of hardcoded version numbers)

### DIFF
--- a/.github/actions/node/14/action.yml
+++ b/.github/actions/node/14/action.yml
@@ -1,8 +1,0 @@
-name: Node 14
-description: Install Node 14
-runs:
-  using: composite
-  steps:
-    - uses: actions/setup-node@v4
-      with:
-        node-version: '14'
--- a/.github/actions/node/16/action.yml
+++ b/.github/actions/node/16/action.yml
@@ -1,8 +1,0 @@
-name: Node 16
-description: Install Node 16
-runs:
-  using: composite
-  steps:
-    - uses: actions/setup-node@v4
-      with:
-        node-version: '16'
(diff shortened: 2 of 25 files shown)


watson:
* **#5322** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5322?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5322?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

BridgeAR (approved):
LGTM, this is IMO a great improvement over the status quo

PR #5326, opened 2025-02-26
fix(dd-trace): expose addSpanProcessor and getActiveSpanProcessor in …

BridgeAR:
…tracer_provider

External PR https://github.com/DataDog/dd-trace-js/pull/5131


bengl (approved):

rochdev:
Have we validated the validity of exposing those? For example it seems from a quick Google search it seems that `addSpanProcessor` might be deprecated in OpenTelemetry?

BridgeAR:
@rochdev that is a very good point! I did not see that it is deprecated before. Now that I checked again I saw that.

@longlho it seems like it would be possible to pass through constructor options instead of adding the additional API here. See https://github.com/open-telemetry/opentelemetry-js/pull/5134 and https://github.com/open-telemetry/opentelemetry-js/pull/5192
Is that sufficient?

Due to the deprecation, I think it's best to not implement these for now.

PR #5333, opened 2025-02-27
Login events SDK v2

uurien:
### What does this PR do?

Implements new SDK methods to send the login username used to login in a different field of user id in the login success/failure tracking methods.

### Motivation

Request login for loginSuccess and loginFailure event, in loginFailure do not request user.id anymore that doesn't make sense when usually user does not exist.

### Plugin Checklist

- [x] Unit tests.
- [x] TypeScript [tests][2].
- [ ] API [documentation][3].

### Additional Notes

ST: https://github.com/DataDog/system-tests/pull/4486

[APPSEC-56916]

[APPSEC-56916]: https://datadoghq.atlassian.net/browse/APPSEC-56916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/test.ts
```diff
@@ -472,6 +472,26 @@ tracer.appsec.trackUserLoginFailureEvent('user_id', false, meta)
 tracer.appsec.trackCustomEvent('event_name')
 tracer.appsec.trackCustomEvent('event_name', meta)
 
+tracer.appsec.v2.trackUserLoginSuccess("login")
+tracer.appsec.v2.trackUserLoginSuccess("login", user)
+tracer.appsec.v2.trackUserLoginSuccess("login", user, meta)
+tracer.appsec.v2.trackUserLoginSuccess("login", null, meta)
```
simon-id, line 478: i'm really not sure i enjoy the v2. here.
It's not really a v2 of the appsec sdk, it's a v2 only for those specific functions. For future breaking change would we put them into this v2 namespace ? should we add new feature to the v2 namespace ?
it just look a bit not well future-proofed you know.

tracer.appsec.trackUserLoginSuccessV2() at least sounds very specific, we made a breaking change on this function only, idk, not super convinced

packages/dd-trace/src/appsec/sdk/set_user.js
```diff
@@ -11,6 +11,12 @@ function setUserTags (user, rootSpan) {
   }
 }
 
+function setUserTagsSdk (user, rootSpan) {
```
simon-id, line 14: we can use this function in user_blocking.js too then

simon-id: actually i just realized, setUserTags() is not even used outside of this function anymore ? it's probably a relic of how automated user tracking used to use the sdk. We can probably just add the `rootSpan.setTag('_dd.appsec.user.collection_mode', 'sdk')` to the `setUserTags()` instead of making a new function no ?

docs/test.ts
simon-id: also lol, try running this method with `process.env.DD_TRACING_ENABLED = '0'` :laughing: can you guess what will happen ?

packages/dd-trace/src/appsec/sdk/track_event.js
```diff
@@ -2,12 +2,27 @@
...
+const appsecMetrics = telemetryMetrics.manager.namespace('appsec')
+
+function increaseSdkEventMetric (eventType, version) {
+  const tags = {
+    event_type: eventType,
+    sdk_version: version
+  }
+
+  appsecMetrics.count('sdk.event', tags).inc(1)
+}
```
simon-id, line 21: should we have this in the telemetry file instead of here ? looks a bit out of plac eno ?

packages/dd-trace/src/appsec/sdk/track_event.js
```diff
@@ -2,12 +2,27 @@
...
+
+/**
+ * @deprecated in favor of trackUserLoginSuccessV2
+ */
```
simon-id, line 25: not sure those are super useful as those functions are not used anywhere really. We have proxy functions that are used for the SDK insteads, and you've already set them as deprecated in the typings file, i think that's enough, no need it here too

packages/dd-trace/src/appsec/sdk/track_event.js
```diff
@@ -99,6 +234,10 @@ function runWaf (eventName, user) {
 
 module.exports = {
   trackUserLoginSuccessEvent,
+  trackUserLoginSuccessV2,
+  trackUserLoginFailureV2,
   trackUserLoginFailureEvent,
```
simon-id, line 239: funny ordering
```suggestion
module.exports = {
  trackUserLoginSuccessEvent,
  trackUserLoginFailureEvent,
  trackUserLoginSuccessV2,
  trackUserLoginFailureV2,
```

packages/dd-trace/src/appsec/sdk/track_event.js
uurien: don't you think it adds value when in the future, when you want to modify the new (or old) method to find easily what is the one that need to be modified?

packages/dd-trace/src/appsec/sdk/track_event.js
uurien: merged Ilyas changes from master, and removed this code

packages/dd-trace/src/appsec/sdk/set_user.js
uurien: yes, I didn't realize that it was used only from sdk, I thought that it was used somewhere else, changing it.

packages/dd-trace/src/appsec/sdk/noop.js
```diff
@@ -1,6 +1,16 @@
...
+class NoopAppsecSdkV2 {
+  trackUserLoginSuccess () {}
+
+  trackUserLoginFailure () {}
+}
+
 class NoopAppsecSdk {
+  constructor () {
+    this.v2 = new NoopAppsecSdkV2()
+  }
```
simon-id, line 12: this is still so weird to have the appsecSdkV2 be INSIDE of appsecSdkV1. I'm not gonna block the PR for that but it doesn't radiate hapiness

uurien: I really don't like the idea of having:

```javascript
tracer.appsec.trackUserLoginSuccessEvent(...)
tracer.appsec.trackUserLoginSuccessV2(...)
tracer.appsec.trackUserLoginFailureEvent(...)
tracer.appsec.trackUserLoginFailureV2(...)
```

imo is cleaner if we have:
```javascript
tracer.appsec.trackUserLoginSuccessEvent(...)
tracer.appsec.trackUserLoginFailureEvent(...)
tracer.appsec.v2.trackUserLoginSuccess(...)
tracer.appsec.v2.trackUserLoginFailure(...)
```

uurien: Maybe, something like `eventTrackingV2` as namespace could be acceptable also:
```javascript
tracer.appsec.trackUserLoginSuccessEvent(...)
tracer.appsec.trackUserLoginFailureEvent(...)
tracer.appsec.eventTrackingV2.trackUserLoginSuccess(...)
tracer.appsec.eventTrackingV2.trackUserLoginFailure(...)
```

simon-id: what about `tracer.appsecV2.trackUserLoginSuccess()` ?

uurien: I am suggesting these names, instead of appsecV2, also because this comment:

https://github.com/DataDog/dd-trace-js/pull/5333#discussion_r2001199730

> It's not really a v2 of the appsec sdk, it's a v2 only for those specific functions

docs/test.ts
```diff
@@ -473,6 +473,26 @@ tracer.appsec.trackUserLoginFailureEvent('user_id', false, meta)
 tracer.appsec.trackCustomEvent('event_name')
 tracer.appsec.trackCustomEvent('event_name', meta)
 
+tracer.appsec.v2.trackUserLoginSuccess("login")
```
simon-id, line 476: you're using a mix of double and single quotes, this should probably be a lint rule one day

docs/test.ts
```diff
@@ -473,6 +473,26 @@ tracer.appsec.trackUserLoginFailureEvent('user_id', false, meta)
...
+tracer.appsec.v2.trackUserLoginFailure("login", true)
+tracer.appsec.v2.trackUserLoginFailure("login", meta)
+tracer.appsec.v2.trackUserLoginFailure("login", false, meta)
+
+tracer.appsec.trackUserLoginSuccessEvent(user, meta)
+
+tracer.appsec.trackUserLoginFailureEvent('user_id', true)
+tracer.appsec.trackUserLoginFailureEvent('user_id', true, meta)
+tracer.appsec.trackUserLoginFailureEvent('user_id', false)
+tracer.appsec.trackUserLoginFailureEvent('user_id', false, meta)
```
simon-id, line 494: bad copy paste ? those are already done a bit higher no ?

uurien: yes... removed

docs/test.ts
uurien: > this should probably be a lint rule one day

Yes there is a big mix of `"` and `'` in this file, replacing this one, thanks.

packages/dd-trace/src/appsec/sdk/track_event.js
```diff
@@ -107,5 +224,9 @@ function runWaf (eventName, user) {
 module.exports = {
   trackUserLoginSuccessEvent,
   trackUserLoginFailureEvent,
-  trackCustomEvent
+  trackUserLoginSuccessV2,
+  trackUserLoginFailureV2,
+  trackCustomEvent,
```
simon-id, line 229: i made a mistake in my previous suggestion
```suggestion
  trackCustomEvent,
  trackUserLoginSuccessV2,
  trackUserLoginFailureV2,
```

packages/dd-trace/src/appsec/sdk/track_event.js
```diff
@@ -63,7 +69,112 @@ function trackCustomEvent (tracer, eventName, metadata) {
...
+      setUserTags(user, rootSpan)
+      metadata.usr = user
+    }
+  }
+
+  trackEvent('users.login.success', metadata, 'v2.trackUserLoginSuccess', rootSpan)
+
+  runWaf('users.login.success', wafData)
+
+  incrementSdkEventMetric('login_success', 'v2')
```
simon-id, line 114: in the v1 method, the sdkEventMetric is called before the rootSpan check. Here it's not. Is that intended ?

packages/dd-trace/src/appsec/sdk/track_event.js
```diff
@@ -63,7 +69,112 @@ function trackCustomEvent (tracer, eventName, metadata) {
...
+    'usr.login': login,
+    'usr.exists': exists ? 'true' : 'false',
+    ...metadata
+  }
+
+  trackEvent('users.login.failure', metadata, 'v2.trackUserLoginFailure', rootSpan)
+
+  runWaf('users.login.failure', wafData)
+
+  incrementSdkEventMetric('login_failure', 'v2')
```
simon-id, line 146: same thing here

packages/dd-trace/src/appsec/sdk/track_event.js
```diff
@@ -63,7 +69,112 @@ function trackCustomEvent (tracer, eventName, metadata) {
...
+
+  const rootSpan = getRootSpan(tracer)
+  if (!rootSpan) {
+    log.warn('[ASM] Root span not available in v2.trackUserLoginFailure')
+    return
+  }
+
+  const wafData = { login }
+
+  if (typeof exists === 'object' && typeof metadata === 'undefined') {
```
simon-id, line 131: what's the reason for using typeof undefined instead of this ?
```suggestion
  if (typeof exists === 'object' && metadata === undefined) {
```

packages/dd-trace/src/appsec/sdk/track_event.js
```diff
@@ -63,7 +69,112 @@ function trackCustomEvent (tracer, eventName, metadata) {
...
+    metadata = exists
+    exists = false
```
simon-id, line 133: if the user doesn't provide `exists`, it doesn't mean that it doesn't exist

uurien: yes, it is the expected behaviour. from the spec:

> The value of the exists parameter must be set in the appsec.events.users.login.failure.usr.exists tag in the root span. If this parameter is not provided, it should default to false.

packages/dd-trace/src/appsec/sdk/track_event.js
```diff
@@ -63,7 +69,112 @@ function trackCustomEvent (tracer, eventName, metadata) {
...
+      truncated: true
+    }
+  }
+
+  const result = {}
+  let truncated = false
+  for (const key of Object.keys(fields)) {
+    const value = fields[key]
+
+    if (value && typeof value === 'object') {
```
simon-id, line 161: little optimization
```suggestion
    if (value !== null && typeof value === 'object') {
```

packages/dd-trace/src/appsec/sdk/track_event.js
```diff
@@ -63,7 +69,112 @@ function trackCustomEvent (tracer, eventName, metadata) {
...
+      truncated = truncated || inheritTruncated
+
+      if (flatValue) {
+        for (const flatKey of Object.keys(flatValue)) {
+          result[`${key}.${flatKey}`] = flatValue[flatKey]
+        }
+      }
+    } else {
+      if (value !== undefined) {
+        result[key] = value
```
simon-id, line 172: should we toString the value ?

packages/dd-trace/src/appsec/sdk/track_event.js
```diff
@@ -63,7 +69,112 @@ function trackCustomEvent (tracer, eventName, metadata) {
...
+    'usr.exists': exists ? 'true' : 'false',
+    ...metadata
+  }
+
+  trackEvent('users.login.failure', metadata, 'eventTrackingV2.trackUserLoginFailure', rootSpan)
+
+  runWaf('users.login.failure', wafData)
+}
+
+function flattenFields (fields, sdkMethodName, depth = 0) {
```
simon-id, line 149: sdkMethodName is only used to be passed to itself no ?

simon-id: by that i mean it's not used anywhere, and the linter is not detecting it because it's recursive

index.d.ts
```diff
@@ -869,13 +869,48 @@ declare namespace tracer {
     flush(): void
   }
 
+  export interface EventTrackingV2 {
+    /**
+     * Links a successful login event to the current trace. Will link the passed user to the current trace with Appsec.setUser() internally.
+     * @param {string} login the login data (username, email...) used by the user to authenticate
```
simon-id, line 875:
```suggestion
     * @param {string} login the login key (username, email...) used by the user to authenticate.
```

index.d.ts
```diff
@@ -869,13 +869,48 @@ declare namespace tracer {
...
+     * @param {User} user Properties of the authenticated user. Accepts custom fields. Can be null.
+     * @param {[key: string]: string} metadata Custom fields to link to the login success event.
+     */
+    trackUserLoginSuccess(login: string, user?: User | null, metadata?: any): void;
```
simon-id, line 879: why any ?
```suggestion
    trackUserLoginSuccess(login: string, user?: User | null, metadata?: { [key: string]: string }): void;
```

index.d.ts
```diff
@@ -869,13 +869,48 @@ declare namespace tracer {
...
+
+    /**
+     * Links a successful login event to the current trace. Will link the passed user to the current trace with Appsec.setUser() internally.
+     * @param {string} login the login data (username, email...) used by the user to authenticate
+     * @param {string} userId Identifier of the authenticated user. Accepts custom fields. Can be null.
```
simon-id, line 884:
```suggestion
     * @param {string} userId Identifier of the authenticated user.
```

index.d.ts
```diff
@@ -869,13 +869,48 @@ declare namespace tracer {
...
+     * @param {[key: string]: string} metadata Custom fields to link to the login success event.
+     */
+    trackUserLoginSuccess(login: string, userId: string, metadata?: any): void;
```
simon-id, line 887:
```suggestion
    trackUserLoginSuccess(login: string, userId?: string, metadata?: any): void;
```

packages/dd-trace/test/appsec/sdk/track_event.spec.js
```diff
@@ -56,6 +57,8 @@ describe('track_event - Internal API', () => {
     })
 
     trackUserLoginSuccessEvent = trackEvents.trackUserLoginSuccessEvent
+    trackUserLoginSuccessV2 = trackEvents.trackUserLoginSuccessV2
+    trackUserLoginFailureV2 = trackEvents.trackUserLoginFailureV2
     trackUserLoginFailureEvent = trackEvents.trackUserLoginFailureEvent
     trackCustomEvent = trackEvents.trackCustomEvent
```
simon-id, line 63:
```suggestion
    trackUserLoginSuccessEvent = trackEvents.trackUserLoginSuccessEvent
    trackUserLoginFailureEvent = trackEvents.trackUserLoginFailureEvent
    trackCustomEvent = trackEvents.trackCustomEvent
    trackUserLoginSuccessV2 = trackEvents.trackUserLoginSuccessV2
    trackUserLoginFailureV2 = trackEvents.trackUserLoginFailureV2
```

packages/dd-trace/src/appsec/sdk/track_event.js
simon-id: oh nvm it's done a bit later by the other function

packages/dd-trace/test/appsec/sdk/track_event.spec.js
```diff
@@ -331,9 +335,509 @@ describe('track_event - Internal API', () => {
...
+
+        expect(log.warn).to.not.have.been.called
+        expect(setUserTags).to.not.have.been.called
+
+        expect(rootSpan.addTags).to.have.been.calledOnceWithExactly(
+          {
+            'appsec.events.users.login.success.track': 'true',
+            '_dd.appsec.events.users.login.success.sdk': 'true',
+            'appsec.events.users.login.success.usr.login': 'login'
+          })
```
simon-id, line 402: i know this is just a styling nit, but why not just use the normal formatting that's used everywhere else ?
```suggestion
        expect(rootSpan.addTags).to.have.been.calledOnceWithExactly({
          'appsec.events.users.login.success.track': 'true',
          '_dd.appsec.events.users.login.success.sdk': 'true',
          'appsec.events.users.login.success.usr.login': 'login'
        })
```

simon-id:
Otherwise LGTM

index.d.ts
uurien: it really accepts anything, non string values are converted to strings and deeper object are also accepted.

packages/dd-trace/src/appsec/sdk/track_event.js
uurien: true, it was there because the log was originally into the method. removing

index.d.ts
simon-id: Please change it in the other occurences too

index.d.ts
simon-id: Please change it in the other occurences too

index.d.ts
simon-id: oh actually i think this one is alright

packages/dd-trace/test/appsec/sdk/track_event.spec.js
simon-id: I'm pretty sure you have the wrong indentation now 
image

simon-id:
also i think the tests are failing because of the log message not using the right prefix

packages/dd-trace/test/appsec/sdk/track_event.spec.js
uurien: linter is failing, so I guess so

uurien:
> also i think the tests are failing because of the log message not using the right prefix

Yes, there is no need to review if the ci is not green 😅

simon-id (approved):

PR #5334, opened 2025-02-27
ESLint: Require 'use strict' header

watson:
### What does this PR do?

Require [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) on all CommonJS modules.


simon-id:
I've reviewed all 344 files in the previous PR, I don't think i have the strength to do it again :/

packages/dd-trace/test/appsec/next/pages-dir/server.js
```diff
@@ -1,5 +1,3 @@
-'use strict'
```
tlhunter, line 1: wat

tlhunter (approved):

BridgeAR (approved):
LGTM. Nice startup perf improvements!
It is great to see that the strict mode is faster during parsing / compiling.

PR #5335, opened 2025-02-27
Report general tags and metrics

IlyasShabi:
### What does this PR do?
Adding general tags and metrics [APPSEC-56393](https://datadoghq.atlassian.net/browse/APPSEC-56393)

1. `_dd.appsec.block.failed`
2. `_dd.appsec.truncated.string_length`
3. `_dd.appsec.truncated.container_size`
4. `_dd.appsec.truncated.container_depth`
5. `_dd.appsec.rasp.timeout`
6. `_dd.appsec.waf.timeouts`
7. `_dd.appsec.waf.error`
8. `_dd.appsec.rasp.error`

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### Additional Notes

[APPSEC-56393]: https://datadoghq.atlassian.net/browse/APPSEC-56393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dd-trace/src/plugins/util/ip_extractor.js
```diff
@@ -28,12 +27,23 @@ const privateCIDRs = [
...
+  privateIPMatcher = new BlockList()
 
-  privateIPMatcher.addSubnet(address, parseInt(prefix), net.isIPv6(address) ? 'ipv6' : 'ipv4')
+  for (const cidr of privateCIDRs) {
+    const [address, prefix] = cidr.split('/')
+
+    privateIPMatcher.addSubnet(address, parseInt(prefix), net.isIPv6(address) ? 'ipv6' : 'ipv4')
+  }
+} catch (e) {
+  // Fallback for older Node.js versions without BlockList
```
simon-id, line 42: I implemented a whole polyfill for blocklist back when I added this feature for this exact reason. I then removed it once we stopped supporting versions of node old enough to not have blocklist.
Optional chaining was added around the same time as blocklist in node, and you're not gonna write try catch for every optional chaining you see right ?
I'm thus very confused as to why you're trying to add this back.

packages/dd-trace/src/appsec/rasp/index.js
```diff
@@ -86,7 +87,21 @@ function blockOnDatadogRaspAbortError ({ error }) {
...
+
+      rootSpan?.addTags({
+        'appsec.blocked': 'true'
+      })
+    } catch (err) {
+      rootSpan?.addTags({
+        '_dd.appsec.block.failed': 1
+      })
+
+      log.error('[ASM] Blocking error', err)
```
simon-id, line 103: this whole try catch is duplicated with the one in appsec/index.js isn't it ? why not just add this try catch directly in the block function ? so we don't have to change the code in those two places ?

packages/dd-trace/src/appsec/waf/waf_context_wrapper.js
```diff
@@ -17,8 +17,8 @@ class WAFContextWrapper {
     this.wafTimeout = wafTimeout
     this.wafVersion = wafVersion
     this.rulesVersion = rulesVersion
-    this.addressesToSkip = new Set()
     this.knownAddresses = knownAddresses
+    this.addressesToSkip = new Set()
     this.cachedUserIdActions = new Map()
   }
```
simon-id, line 23: superfluous change no ? what's the rationale behind it ?

IlyasShabi: No functional change intended in this modification, the reordering was just to separate the creation of new instances from the other argument attribution

simon-id: right, i guess it's nicer indeed

packages/dd-trace/src/appsec/blocking.js
```diff
@@ -100,29 +100,43 @@ function getBlockingData (req, specificType, actionParameters) {
...
+    rootSpan?.addTags({
+      '_dd.appsec.block.failed': 1
+    })
+    log.error('[ASM] Blocking error', err)
+
+    blocked = false
+  }
 
-  abortController?.abort()
+  return blocked
```
simon-id, line 139: no need for a variable, just add return true in the try, and return false in the catch

packages/dd-trace/src/appsec/blocking.js
```diff
@@ -100,29 +100,43 @@ function getBlockingData (req, specificType, actionParameters) {
...
 
-  // this is needed to call the original end method, since express-session replaces it
-  res.constructor.prototype.end.call(res, body)
+    // this is needed to call the original end method, since express-session replaces it
+    res.constructor.prototype.end.call(res, body)
 
-  responseBlockedSet.add(res)
+    rootSpan.addTags({
+      'appsec.blocked': 'true'
+    })
```
simon-id, line 124: i just realized this should be setTag() as we're only modifying one tag. Probably coming from a previous change.

packages/dd-trace/src/appsec/blocking.js
```diff
@@ -100,29 +100,43 @@ function getBlockingData (req, specificType, actionParameters) {
...
+
+    responseBlockedSet.add(res)
+    abortController?.abort()
+
+    blocked = true
+  } catch (err) {
+    rootSpan?.addTags({
+      '_dd.appsec.block.failed': 1
+    })
```
simon-id, line 133:
```suggestion
    rootSpan?.setTag('_dd.appsec.block.failed', 1)
```

packages/dd-trace/src/appsec/graphql.js
```diff
@@ -94,14 +95,22 @@ function beforeWriteApolloGraphqlResponse ({ abortController, abortData }) {
...
 
-    rootSpan.setTag('appsec.blocked', 'true')
+      rootSpan.setTag('appsec.blocked', 'true')
 
-    abortController?.abort()
+      abortController?.abort()
+    } catch (err) {
+      rootSpan.addTags({
+        '_dd.appsec.block.failed': 1
+      })
```
simon-id, line 110:
```suggestion
      rootSpan.setTag('_dd.appsec.block.failed', 1)
```

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -209,6 +236,14 @@ function finishRequest (req, res) {
     rootSpan.setTag('_dd.appsec.rasp.rule.eval', metrics.raspEvalCount)
   }
 
+  if (metrics?.raspTimeouts) {
+    rootSpan.setTag('_dd.appsec.rasp.timeout', metrics.raspTimeouts)
+  }
+
+  if (metrics?.raspErrorCode) {
+    rootSpan.setTag('_dd.appsec.rasp.error', metrics.raspErrorCode)
+  }
```
simon-id, line 245: tiny nit, but this could be just above `_dd.appsec.rasp.rule.eval`, so we can have the same order for both waf and rasp

packages/dd-trace/src/appsec/blocking.js
```diff
@@ -100,29 +100,36 @@ function getBlockingData (req, specificType, actionParameters) {
...
-  responseBlockedSet.add(res)
+    rootSpan.setTag('appsec.blocked', 'true')
 
-  abortController?.abort()
+    responseBlockedSet.add(res)
+    abortController?.abort()
+
+    return true
+  } catch (err) {
+    rootSpan?.setTag('_dd.appsec.block.failed', 1)
```
simon-id, line 128: I think it's possible the 1 should be a string here, otherwise the tag goes into metrics instead of tags. Try to confirm it with other implementer of the spec

packages/dd-trace/src/appsec/graphql.js
```diff
@@ -94,14 +95,20 @@ function beforeWriteApolloGraphqlResponse ({ abortController, abortData }) {
     const rootSpan = web.root(req)
     if (!rootSpan) return
```
simon-id, line 96: shouldn't this be included as a blocking failure ?

packages/dd-trace/src/appsec/telemetry/rasp.js
```diff
@@ -5,10 +5,21 @@ const { DD_TELEMETRY_REQUEST_METRICS } = require('./common')
...
+
+  if (wafTimeout) {
+    store[DD_TELEMETRY_REQUEST_METRICS].raspTimeouts++
+  }
+
+  if (errorCode) {
+    store[DD_TELEMETRY_REQUEST_METRICS].raspErrorCode = Math.max(
+      errorCode,
+      store[DD_TELEMETRY_REQUEST_METRICS].raspErrorCode ?? errorCode
+    )
```
simon-id, line 21: is the max() thing defined by the spec ? it doesn't look like an obvious thing to do. If it's defined by the spec then no problem

packages/dd-trace/src/appsec/waf/waf_context_wrapper.js
```diff
@@ -77,48 +77,86 @@ class WAFContextWrapper {
...
-      const blockTriggered = !!getBlockingAction(result.actions)
+    if (!result) {
+      // Binding or other waf unexpected errors
+      metrics.errorCode = -127
 
-      // SPECIAL CASE FOR USER_ID
-      // TODO: make this universal
-      if (userId && ruleTriggered && blockTriggered) {
-        this.setUserIdCache(userId, result)
+      return null
```
simon-id, line 102: no need to do an explicit null return, this function doesn't return anything anyway

packages/dd-trace/src/appsec/waf/waf_context_wrapper.js
```diff
@@ -77,48 +77,86 @@ class WAFContextWrapper {
...
+    } else {
+      if (typeof result.errorCode === 'number' && result.errorCode < 0) {
```
simon-id, line 104: no need for else if there is a return just before

packages/dd-trace/src/appsec/waf/waf_context_wrapper.js
```diff
@@ -77,48 +77,86 @@ class WAFContextWrapper {
...
+    this.addressesToSkip = newAddressesToSkip
 
-      const ruleTriggered = !!result.events?.length
+    const end = process.hrtime.bigint()
+    metrics.durationExt = parseInt(end - start) / 1e3
 
-      const blockTriggered = !!getBlockingAction(result.actions)
+    if (!result) {
+      // Binding or other waf unexpected errors
+      metrics.errorCode = -127
```
simon-id, line 100: what's the point of adding errorcode to metrics if metrics are discarded anyway ?

packages/dd-trace/src/appsec/graphql.js
IlyasShabi: If the rootSpan do not exist how we could send the tags?

packages/dd-trace/src/appsec/telemetry/rasp.js
IlyasShabi: Yes the max it's defined by the spec in case we run the WAF multiple times during a request and we got different error codes.
We call max only if the errorCode is defined

packages/dd-trace/src/appsec/waf/waf_context_wrapper.js
IlyasShabi: you are right, the is a left over from previous impl

packages/dd-trace/src/appsec/waf/waf_context_wrapper.js
IlyasShabi: I removed return null

packages/dd-trace/src/appsec/waf/waf_context_wrapper.js
IlyasShabi: I removed return null

packages/dd-trace/src/appsec/telemetry/rasp.js
simon-id: yes but you call max even when currentRaspErrorCode is undefined, with my ternary operator you wouldn't

packages/dd-trace/src/appsec/graphql.js
simon-id: it would still print the error log

simon-id: but it's fine whatever

packages/dd-trace/src/appsec/waf/waf_context_wrapper.js
```diff
@@ -77,48 +77,84 @@ class WAFContextWrapper {
 
     if (!payloadHasData) return
 
-    try {
-      const start = process.hrtime.bigint()
+    const start = process.hrtime.bigint()
```
simon-id, line 80: same thing here, no need to have metrics called after the hrtime start

packages/dd-trace/src/appsec/telemetry/rasp.js
IlyasShabi: I pushed an if/else approach which is more readable

packages/dd-trace/src/appsec/telemetry/index.js
simon-id, line 1: no tests for the telemetry changes ?

packages/dd-trace/src/appsec/graphql.js
simon-id, line 1: no test for the graphql change ?

packages/dd-trace/src/appsec/telemetry/index.js
IlyasShabi: I don't know how I forget this commit

packages/dd-trace/src/appsec/graphql.js
IlyasShabi: Done

packages/dd-trace/test/appsec/reporter.spec.js
simon-id, line 1: no test updated to cover the wafErrorCode, wafTimeouts, raspErrorCode, and raspTimeouts that were added in reporter.js ?

packages/dd-trace/src/appsec/telemetry/index.js
simon-id: where ? you didn't add any tests in telemetry, you added them to reporter, which stubs telemetry, i don't see any tests that check the Math.max() error code, I don't see any test that checks the timeouts and error codes for rasp and waf are actually added to telemeetry no ?

IlyasShabi: Should be good now

packages/dd-trace/test/appsec/telemetry/rasp.spec.js
```diff
@@ -38,7 +38,7 @@ describe('Appsec Rasp Telemetry metrics', () => {
         appsecTelemetry.updateRaspRequestsMetricTags({
           duration: 42,
           durationExt: 52
-        }, req, 'rule-type')
+        }, req, { type: 'rule-type' })
```
simon-id, line 41: not your fault, but it's awful to see that these tests were wrong, but not failing :/

packages/dd-trace/test/appsec/blocking.spec.js
```diff
@@ -58,20 +58,22 @@ describe('blocking', () => {
...
     it('should log warn and not send blocking response when headers have already been sent', () => {
       res.headersSent = true
-      block(req, res, rootSpan)
+      const blocked = block(req, res, rootSpan)
 
+      expect(blocked).to.be.false
       expect(log.warn).to.have.been
         .calledOnceWithExactly('[ASM] Cannot send blocking response when headers have already been sent')
-      expect(rootSpan.addTags).to.not.have.been.called
+      expect(rootSpan.setTag).to.have.been.called
```
simon-id, line 66:
```suggestion
      expect(rootSpan.setTag).to.have.been.calledOnceWithExactly('appsec.blocked', 'true')
```

simon-id, line 66: called with what ? you never check that it's been called with the block failed tag ?

simon-id:
```suggestion
      expect(rootSpan.setTag).to.have.been.calledOnceWithExactly('_dd.appsec.block.failed', 1)
```

packages/dd-trace/test/appsec/waf/waf_context_wrapper.spec.js
```diff
@@ -4,16 +4,30 @@ const proxyquire = require('proxyquire')
...
 describe('WAFContextWrapper', () => {
   const knownAddresses = new Set([
     addresses.HTTP_INCOMING_QUERY,
     addresses.HTTP_INCOMING_GRAPHQL_RESOLVER
   ])
 
+  let reportMetricsStub
+
+  before(() => {
+    reportMetricsStub = sinon.stub(Reporter, 'reportMetrics')
```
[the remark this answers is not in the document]
simon-id: stub !== spy btw

packages/dd-trace/test/appsec/reporter.spec.js
```diff
@@ -150,15 +150,37 @@ describe('reporter', () => {
...
+      const metrics = { durationExt: 42, totalRuntime: 1337000 }
+
+      Reporter.reportMetrics(metrics)
+
+      expect(telemetry.updateWafRequestsMetricTags).not.to.have.been.called
+      expect(telemetry.updateRaspRequestsMetricTags).not.to.have.been.called
+    })
+
+    it('should set duration metrics from result.totalRuntime', () => {
+      const metrics = { durationExt: 42, totalRuntime: 1337000 }
```
simon-id, line 165: pretty sure totalRuntime here is completely wrong, it's called `duration` instead. And the fact that your test is not failing, means you're not testing the feature correctly :/

packages/dd-trace/test/appsec/blocking.spec.js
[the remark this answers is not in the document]
IlyasShabi: We do have a test when headers have already been sent https://github.com/DataDog/dd-trace-js/pull/5335/files#diff-c366e875523cd49112a82d0f44ef089846b3cc7b4bf5a5f8a1b1d9c225410a13R66

packages/dd-trace/test/appsec/reporter.spec.js
IlyasShabi: It's a leftover from a previous impl. 
This test file only checks whether we call `updateRaspRequestsMetricTags` or `updateWafRequestsMetricTags` correctly, depending on whether `reportMetrics` is called with `raspRule`. 
In other words, you can add any data in `metrics` and verify that it functions as expected.
I pushed the fix

packages/dd-trace/test/appsec/waf/waf_context_wrapper.spec.js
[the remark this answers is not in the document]
simon-id: i don't understand what you mean, i'm saying you don't need the alias

packages/dd-trace/test/appsec/blocking.spec.js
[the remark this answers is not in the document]
simon-id: yes i was thinking to make the blocking crash unexpectedly instead of this test which is our own error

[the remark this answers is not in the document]
simon-id: but you know what, it's fine like that whatever

packages/dd-trace/test/appsec/waf/waf_context_wrapper.spec.js
```diff
@@ -27,14 +41,45 @@ describe('WAFContextWrapper', () => {
     wafContextWrapper.run(payload)
 
     expect(ddwafContext.run).to.have.been.calledOnceWithExactly(payload, 1000)
+    expect(reportMetricsStub).to.have.been.calledOnce
   })
 
-  it('Should send ephemeral addresses every time', () => {
+  it('Should send HTTP_INCOMING_QUERY twice if waf run returns null', () => {
```
simon-id, line 47: not sure what this test is trying to do with the twice thing ? why not call it just once ?

simon-id, line 47:
```suggestion
  it('Should send HTTP_INCOMING_QUERY twice if waf run fails', () => {
```

packages/dd-trace/test/appsec/waf/waf_context_wrapper.spec.js
```diff
@@ -27,14 +41,45 @@ describe('WAFContextWrapper', () => {
...
     const ddwafContext = {
       run: sinon.stub()
```
simon-id, line 49: to make it clear that it's not because it returns undefined, but because it throws
```suggestion
      run: sinon.stub().throws(new Error('test'))
```

packages/dd-trace/test/appsec/waf/waf_context_wrapper.spec.js
```diff
@@ -27,14 +41,45 @@ describe('WAFContextWrapper', () => {
...
+      persistent: {
+        [addresses.HTTP_INCOMING_QUERY]: { key: 'value' }
+      }
+    }
+
+    wafContextWrapper.run(payload)
+    wafContextWrapper.run(payload)
+
+    expect(ddwafContext.run).to.have.been.calledTwice
+    expect(ddwafContext.run).to.have.been.calledWithExactly(payload, 1000)
```
simon-id, line 63: The test title says you're making sure it's sent twice, then why don't you actually check it's sent twice
```suggestion
    expect(ddwafContext.run.firstCall).to.have.been.calledWithExactly(payload, 1000)
    expect(ddwafContext.run.secondCall).to.have.been.calledWithExactly(payload, 1000)
```

packages/dd-trace/test/appsec/waf/waf_context_wrapper.spec.js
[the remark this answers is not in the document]
IlyasShabi: Honestly I don't see a diff, but okay I can do it

packages/dd-trace/test/appsec/waf/waf_context_wrapper.spec.js
```diff
@@ -27,14 +41,45 @@ describe('WAFContextWrapper', () => {
...
+
+    const firstCall = reportMetricsStub.getCall(0).args[0]
+    expect(firstCall).to.have.property('errorCode')
+    expect(firstCall.errorCode).to.equal(-127)
+
+    const secondCall = reportMetricsStub.getCall(1).args[0]
+    expect(secondCall).to.have.property('errorCode')
+    expect(secondCall.errorCode).to.equal(-127)
```
simon-id, line 71:
```suggestion
    const firstCall = Reporter.reportMetrics.firstCall.args[0]
    expect(firstCall).to.have.property('errorCode', -127)

    const secondCall = Reporter.reportMetrics.secondCall.args[0]
    expect(secondCall).to.have.property('errorCode', -127)
```

packages/dd-trace/test/appsec/waf/waf_context_wrapper.spec.js
IlyasShabi: I do check `reportMetrics` instead

packages/dd-trace/test/appsec/waf/waf_context_wrapper.spec.js
```diff
@@ -75,7 +119,7 @@ describe('WAFContextWrapper', () => {
 
     wafContextWrapper.run(payload)
 
-    expect(ddwafContext.run).to.have.not.been.called
```
simon-id, line 78: to have not been called is grammatically more correct but ye apparently whoever wrote those tests didn't get it right, and this is why i hate expect() xD

packages/dd-trace/test/appsec/waf/waf_context_wrapper.spec.js
simon-id, line 1: one last thing, it would be nice to have a specific test that checks context.run() is calling getMetrics with a non-default error code, truncation metrics, blocktriggered, ruleTriggered, and all the metrics that aren't time duration (those are annoying to check)

packages/dd-trace/test/appsec/waf/waf_context_wrapper.spec.js
simon-id: yes but here the test is to make sure `addressesToSkip` is not updated after a failed waf call, metrics have nothing to do with this, so you need to make sure the waf is called twice with the address, to make sure `addressesToSkip` wasn't updated in the firsrt call

simon-id: I thought this was already what you meant by writing this test, otherwise i have no idea why you called it "should call querystring twice"

IlyasShabi: I kept the title and check `ddwafContext.run` calls, it's not related to my PR but nice to have it

packages/dd-trace/test/appsec/waf/waf_context_wrapper.spec.js
IlyasShabi: Done

simon-id (approved):

PR #5338, opened 2025-02-27
Improve release proposal script

watson:
### What does this PR do?

Improve release proposal script.
    
This introduces two significant changes:
    
1. Don't suppress important errors. Some errors generated by the commands in the script should be ignored as they are just noise. Others however are important clues to setup and permission issues.
2. Improve requirements check, to ensure branch-diff has the expected permisisons.

### Motivation

If the branch-diff permissions were not configured correctly (or expired), no warning would be given and the resulting release diff would be incorrect.

### Additional Notes

The change in the default behavior from originally hiding errors to now showing them, was done before I decided to add the extra requirement checks. Technically, now that these checks are added, the output of STDERR is no longer that important, and could technically be reverted. But we never know what other errors might occur in the future, so showing them is proabably a good idea.

### DIFF
--- a/scripts/release/helpers/requirements.js
+++ b/scripts/release/helpers/requirements.js
@@ -10,7 +10,7 @@ const { capture, fatal, run } = require('./terminal')
 // Check that the `git` CLI is installed.
 function checkGit () {
   try {
-    run('git --version', true)
+    run('git --version', false)
   } catch (e) {
     fatal(
       'The "git" CLI could not be found.',
@@ -22,7 +22,7 @@ function checkGit () {
 // Check that the `branch-diff` CLI is installed.
 function checkBranchDiff () {
   try {
-    run('branch-diff --version', true)
+    run('branch-diff --version', false)
   } catch (e) {
     const link = [
       'https://datadoghq.atlassian.net/wiki/spaces/DL/pages/3125511269/Node.js+Tracer+Release+Process',
@@ -66,7 +66,7 @@ function checkGitHub () {
   }
 
   try {
-    run('gh --version', true)
+    run('gh --version', false)
   } catch (e) {
     fatal(
       'The "gh" CLI could not be found.',
--- a/scripts/release/helpers/requirements.js
+++ b/scripts/release/helpers/requirements.js
@@ -2,14 +2,15 @@
 
 /* eslint-disable @stylistic/js/max-len */
 
+const { join } = require('path')
+const { existsSync, readFileSync } = require('fs')
+const { default: getApplicationConfigPath } = require('application-config-path')
 const { capture, fatal, run } = require('./terminal')
-
-const requiredScopes = ['public_repo', 'read:org']
 
 // Check that the `git` CLI is installed.
 function checkGit () {
   try {
-    run('git --version')
+    run('git --version', true)
   } catch (e) {
     fatal(
       'The "git" CLI could not be found.',
@@ -21,7 +22,7 @@ function checkGit () {
 // Check that the `branch-diff` CLI is installed.
 function checkBranchDiff () {
   try {
-    run('branch-diff --version')
+    run('branch-diff --version', true)
   } catch (e) {
     const link = [
       'https://datadoghq.atlassian.net/wiki/spaces/DL/pages/3125511269/Node.js+Tracer+Release+Process',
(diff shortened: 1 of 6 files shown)


watson:
* **#5338** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5338?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5338?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

scripts/release/helpers/requirements.js
```diff
@@ -32,53 +33,72 @@ function checkBranchDiff () {
...
+function checkGitHubScopes (token, requiredScopes, source) {
   const url = 'https://api.github.com'
   const headers = [
     'Accept: application/vnd.github.v3+json',
-    `Authorization: Bearer ${process.env.GITHUB_TOKEN || process.env.GH_TOKEN}`,
+    `Authorization: Bearer ${token}`,
     'X-GitHub-Api-Version: 2022-11-28'
   ].map(h => `-H "${h}"`).join(' ')
 
-  const lines = capture(`curl -sS -I ${headers} ${url}`).trim().split(/\r?\n/g)
```
simon-id, line 70: why remove .trim() ? you sure it's useless ?

simon-id (approved):
one tiny comment otherwise i'm preapproving

scripts/release/helpers/terminal.js
```diff
@@ -54,12 +54,19 @@ function checkpoint (question) {
...
+function capture (cmd, ignoreErrors = false) {
   if (flags.debug) {
     log(`${GRAY}> ${cmd}${RESET}`)
   }
 
-  const output = execSync(cmd, { encoding: 'utf8', stdio: 'pipe' }).toString().trim()
+  const opts = {
+    encoding: 'utf8',
+    // execSync will only return the value of STDOUT.
+    // The easiest way to get access STDERR is not set `stdio`, which will force STDERR to be piped to `process.stderr`
```
rochdev, line 65: This was done on purpose because any output not controlled by the script can cause issues (because of the logic that rewrites lines as progress advances). Any way to instead store this somewhere else and output errors at the very end? (or even better, formatted).

rochdev (changes requested):
Left a non-blocking comment about potential output issues, but thinking more about it I think it warrants to be blocking.

scripts/release/helpers/requirements.js
watson: the output of `capture` is already trimmed (you added that in #4897 😉)

scripts/release/helpers/terminal.js
watson: I wish that was possible. I wanted to do that as well, but I couldn't find a way to do that in a sync way. Maybe I've missed something 🤔

watson: Ok, you inspired me to ask ChatGPT and I discovered an option to `spawnSync` that I didn't know: `shell: true`, which will make `spawnSync` interpret the given command similar to `execSync`. This will allow us to get access to both STDOUT and STDERR separately.

I've pushed a commit to try and address your comments. Let me know what you think 😃

scripts/release/proposal.js
```diff
@@ -50,40 +50,39 @@ try {
...
   const notesFile = path.join(notesDir, `${newVersion}.md`)
 
-  pass(`${isMinor ? 'minor' : 'patch'} (${lastVersion} -> ${newVersion})`)
+  pass(`${isMinor ? 'minor' : 'patch'} (${DD_MAJOR}.${DD_MINOR}.${DD_PATCH} -> ${newVersion})`)
 
   start('Checkout release proposal branch')
 
   // Checkout new or existing branch.
-  run(`git checkout v${newVersion}-proposal || git checkout -b v${newVersion}-proposal`)
+  run(`git checkout --quiet v${newVersion}-proposal &> /dev/null || git checkout --quiet -b v${newVersion}-proposal`)
```
rochdev, line 81: Why is this still needed with the latest version?

watson: I'm pretty sure I don't need the `--quiet` flag (do you want me to remove it?), but if I don't redirect the STDERR from the first command it will be picked up by the new error output detection code and the script aborted.

rochdev: Ok then the change is fine. I just wanted to make sure that it won't actually output, as long as that's true then it's fine 👍

rochdev (approved):

PR #5347, opened 2025-03-03
Custom metrics aggregation

rochdev:
### What does this PR do?

Add pre-aggregation support to custom metrics.

### Motivation

Right now only runtime metrics are pre-aggregated, and every data point is sent for custom metrics. This is problematic because there might be too many data points which might go above the threshold of the agent. This normalizes the aggregation logic to apply to both runtime metrics and custom metrics.

### DIFF
--- a/packages/dd-trace/src/dogstatsd.js
+++ b/packages/dd-trace/src/dogstatsd.js
@@ -305,7 +305,11 @@ class MetricsAggregationClient {
   }
 }
 
-// This is a simplified user-facing proxy to the underlying DogStatsDClient instance
+/**
+ * This is a simplified user-facing proxy to the underlying DogStatsDClient instance
+ *
+ * @implements {DogStatsD}
+ */
 class CustomMetrics {
   constructor (config) {
     const clientConfig = DogStatsDClient.generateClientConfig(config)
--- a/packages/dd-trace/src/dogstatsd.js
+++ b/packages/dd-trace/src/dogstatsd.js
@@ -6,6 +6,7 @@ const dgram = require('dgram')
 const isIP = require('net').isIP
 const log = require('./log')
 const { URL, format } = require('url')
+const Histogram = require('./histogram')
 
 const MAX_BUFFER_SIZE = 1024 // limit from the agent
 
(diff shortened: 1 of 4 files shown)


packages/dd-trace/test/dogstatsd.spec.js
```diff
@@ -376,61 +377,76 @@ describe('dogstatsd', () => {
...
       expect(udp4.send).to.have.been.called
-      expect(udp4.send.firstCall.args[0].toString()).to.equal('test.histogram:10|h\n')
+      expect(udp4.send.firstCall.args[0].toString()).to.equal([
+        'test.histogram.min:10|g',
+        'test.histogram.max:10|g',
+        'test.histogram.sum:20|c',
+        'test.histogram.total:20|c',
+        'test.histogram.avg:10|g',
+        'test.histogram.count:2|c',
+        'test.histogram.median:10.074696689511441|g',
```
tlhunter, line 447: Is this a bug in the histogram calculation?

packages/dd-trace/src/dogstatsd.js
```diff
@@ -193,6 +194,117 @@ class DogStatsDClient {
   }
 }
 
+// TODO: Handle arrays of tags and tags translation.
+class MetricsAggregationClient {
```
tlhunter, line 198:
```suggestion
class MetricsAggregationClient {
  _counters = {}
  _gauges = {}
  _histograms = {}
```

packages/dd-trace/src/dogstatsd.js
```diff
@@ -193,6 +194,117 @@ class DogStatsDClient {
...
+  decrement (name, count = 1, tag) {
+    this.count(name, -count, tag)
+  }
+
+  _captureGauges () {
+    Object.keys(this._gauges).forEach(name => {
+      this._gauges[name].forEach((value, tag) => {
+        this._client.gauge(name, value, tag && [tag])
+      })
+    })
```
tlhunter, line 270: I see the code didn't do it this way before your refactor but I'm wondering if the gauges should be reset?

```js
    this._gauges = {}
```

tlhunter: Maybe not since they represent a value that remains steady when nothing changes?

packages/dd-trace/test/custom-metrics.spec.js
```diff
@@ -53,7 +53,7 @@ describe('Custom Metrics', () => {
       if (stdout) console.log(stdout)
       if (stderr) console.error(stderr)
 
-      expect(metricsData.split('#')[0]).to.equal('page.views.data:1|c|')
+      expect(metricsData.split('#')[0]).to.equal('page.views.data:1|g|')
```
tlhunter, line 56: Does this mean all increments switched from a count metric to a gauge metric? 

```js
tracer.dogstatsd.increment('page.views.data')
```

rochdev: Similar to runtime metrics, instead of sending the increment to the agent/backend and having them handle actually applying it, it is now applied upfront.

packages/dd-trace/test/dogstatsd.spec.js
rochdev: My understanding is that this is normal as the histogram doesn't store values directly, it stores approximations, so while the values are expected to be close to the originals, there is no guarantee of being exact.

packages/dd-trace/src/dogstatsd.js
rochdev: This is already covered in the constructor through `reset()`

PR #5357, opened 2025-03-04
Only load async_hooks if tracing is enabled

watson:
The act of requiring the `async_hooks` module has side effects, which can lead to a rare race condition in Node.js 18 under some very specific circomstances. It's not 100% understood yet, but the race condition seems to only manifest itself in worker threads where tracing is disabled.
    
This commit works around that issue, by not loading `async_hooks` from within the logger, since the logger might be used in threads where tracing is disabled.


watson:
* **#5357** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5357?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5357?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

BridgeAR:
@watson is this still needed / do you still want to finish this? The Node.js 18 bug is fixed in the latest version, so we are not running into that anymore.

watson:
I still think it's a win if we can avoid loading async hooks if we don't need it, as I still think there's an overhead if we do.

PR #5362, opened 2025-03-04
fix release proposal failing when stderr has output

rochdev:
### What does this PR do?

Fix release proposal failing when stderr has output.

### Motivation

It's common for `stderr` to have output even without any error, which would end up failing the script. By instead relying on the status code, we can simplify the code and only abort on actual errors.

### DIFF
--- a/scripts/release/helpers/terminal.js
+++ b/scripts/release/helpers/terminal.js
@@ -23,8 +23,8 @@ let timer
 let current
 
 // Output a command to the terminal and execute it.
-function run (cmd, treatStderrAsFailure = true) {
-  capture(cmd, treatStderrAsFailure)
+function run (cmd) {
+  capture(cmd)
 }
 
 // Ask a question in terminal and return the response.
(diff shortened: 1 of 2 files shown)


scripts/release/helpers/terminal.js
```diff
@@ -76,25 +76,11 @@ function capture (cmd, treatStderrAsFailure = true) {
...
-    }
-  } else if (treatStderrAsFailure && stderr) {
-    if (flags['no-abort-on-error']) {
-      log(`${RED}${stderr}${RESET}`)
-    } else {
-      fatal(
-        stderr,
-        'Aborting due to error! Use --no-abort-on-error to ignore and continue.'
-      )
-    }
```
watson, line 97: Removing this will unfortunately get us back to the original problem, which prompted me to make this change in the first place: The `branch-diff` tool will exit with a status code zero, but output errors on STDERR if there's a permission issue.

While I have also updated the requirements check to check that the expected permissions are in place, it would still be nice to have some sort of guardrail in place for when `branch-diff` runs into issues.

However, instead of defaulting to throwing on output to STDERR, maybe we should change the default to ignoring it, and only for branch-diff commands we can enable it.

Alternatively, or in combination with this, we could also change the default to only output STDERR, but not abort on it.

What do you think about all this?

PR #5365, opened 2025-03-05
add new baggage APIs which act on currently active baggage

ida613:
### What does this PR do?
Provides users with a set of global functions which act on the currently active baggage of [implicitly propagated context](https://opentelemetry.io/docs/specs/otel/baggage/api/#context-interaction)

### Motivation
From the baggage RFC:
> Generally speaking, the Baggage API should be exposed on the Context object or class on APM SDKs where the context itself is exposed. On SDKs where context is implicit, the Baggage API can be exposed through static/global functions that act on the currently active baggage.

Added return statements because
> If any language implements baggage as an immutable object or data structure (as defined in the OpenTelemetry Baggage API specification), the functions above that modify baggage (set and remove) will need to return the new (modified) baggage object. However, if baggage is mutable and modified in place, those functions don't need to return a value.

### Plugin Checklist

- [x] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/dd-trace/src/opentracing/baggage.js
```diff
@@ -0,0 +1,40 @@
...
+}
+
+function getAllBaggageItems () {
+  return storage('baggage').getStore()
+}
+
+function removeBaggageItem (keyToRemove) {
+  const newBaggage = {}
+  const oldBaggage = storage('baggage').getStore()
+  for (const [key, value] of Object.entries(oldBaggage)) {
```
ida613, line 22: use destructuring to delete key-value pair

BridgeAR (changes requested):
The transport itself seems fine, I just don't see how this is integrated and exposed to users. That is something that still needs to be addressed :)

ida613:
@BridgeAR sorry about misunderstanding you before! I have added the APIs to NoopProxy to make them accessible to the public. Please let me know if this solution is fine and I'll update the test :)

BridgeAR (changes requested):
Thanks for exposing the API now!
While it would work that way, it should not be on the `no op` proxy class. The name is an indicator for no operation, so all methods on that class should just do nothing. Instead, the API should probably be exposed on the main Tracer proxy.
Can you please also add tests that actually use the API as an integration test and also add documentation for it?

ida613:
@BridgeAR thank you for your review Ruben! :) This is awkward, technically baggage is separate from tracing and does not require tracing to be enabled to work

BridgeAR:
@ida613 that is fair! I feel I am actually not the right person to suggest where to add the functionality in that case. I just know that it should not be the NoopProxy.

@rochdev do you have an opinion about that?

bengl:
The methods should be added to the proxy in `dd-trace/src/proxy.js`. No-op versions of this should be added to the NoopTracer proxy (i.e. the one currently in this file).

rochdev:
> This is awkward, technically baggage is separate from tracing and does not require tracing to be enabled to work

Unfortunately the original design of the library was only a tracer (hence the name dd-trace) and the default export ended up being the tracer itself. This means that anything, whether or not related to tracing, needs to be on the tracer to be exposed. Of course that could be directly on the proxy instead of the underlying tracer, but from a public API perspective they are the effectively the same. The underlying implementation however should indeed not be coupled to tracing.

ida613:
Made changes as Bryan suggested! Will work on tests now

ida613:
@BridgeAR @rochdev I've updated the tests. Let me know if there is anything else to update :)

BridgeAR (commented):
The code is looking good to me!

I believe we still need some documentation for it. Otherwise it would be pretty much "in the wild" :D
I believe our TS definition is pretty much that documentation, if I am not mistaken.

ida613:
@BridgeAR added the TS definition! let me know if it's good to go :)

index.d.ts
```diff
@@ -138,6 +138,15 @@ interface Tracer extends opentracing.Tracer {
    * LLM Observability SDK
    */
   llmobs: tracer.llmobs.LLMObs;
+
+  /**
+   * Baggage
```
BridgeAR, line 143: Could you please add some further description about the usage and link to https://opentelemetry.io/docs/concepts/signals/baggage/?

index.d.ts
```diff
@@ -138,6 +138,15 @@ interface Tracer extends opentracing.Tracer {
...
+   */
+  setBaggageItem (key: string, value: string): {[key: string]: string};
```
BridgeAR, line 145: Nit (I find these easier to read, it can also stay though, if you prefer that)

```suggestion
  setBaggageItem (key: string, value: string): Record<string, string>;
```

index.d.ts
```diff
@@ -138,6 +138,15 @@ interface Tracer extends opentracing.Tracer {
...
+  getBaggageItem (key: string): string | undefined;
+  getAllBaggageItems (): {[key: string]: string};
+  removeBaggageItem (key: string): {[key: string]: string};
+  removeAllBaggageItems (): {[key: string]: string};
```
BridgeAR, line 149:
```suggestion
  getAllBaggageItems (): Record<string, string>;
  removeBaggageItem (key: string): Record<string, string>;
  removeAllBaggageItems (): Record<string, string>;
```

BridgeAR (approved):
The code is LGTM as far as I can tell, thanks for following up on these!

I left a suggestion for the description which would be nice. I don't know all that much about the otel implementation and and the text_map and it would be nice if @rochdev could have a brief look as well as second review.

BridgeAR (approved):
LGTM
Thank you for the hard work, @ida613! :)

PR #5368, opened 2025-03-05
fix release script esm error on node 20

rochdev:
### What does this PR do?

Fix release script ESM error on Node 20.

### Motivation

The `application-config-path` dependency is an ESM module which cannot be used with `require` in some versions of Node, resulting in errors. Since it's only a few lines to find where the config path is, I just vendored the code.


watson (changes requested):
When vendoring code like this, I think it's best to include a comment in the top of the file with a reference to the original code

rochdev:
@watson Added a comment, let me know if that works!

watson (approved):

package.json
```diff
@@ -122,7 +122,6 @@
     "@msgpack/msgpack": "^3.0.0-beta3",
     "@stylistic/eslint-plugin-js": "^3.0.1",
     "@types/node": "^16.0.0",
-    "application-config-path": "^1.0.0",
```
BridgeAR, line 125:
```suggestion
    "application-config-path": "0.1.1",
```

BridgeAR (changes requested):
I would rather just pin 0.1.1. It was a CJS module and nothing else besides adding TS definitions were added in the meanwhile. I do not feel comfortable inlining the code as is and it is easier for us to just use the older version.

rochdev:
> I would rather just pin 0.1.1. It was a CJS module and nothing else besides adding TS definitions were added in the meanwhile. I do not feel comfortable inlining the code as is and it is easier for us to just use the older version.

Why? I mean I can make the change in this case as it still solves the immediate issue, but we shouldn't feel uncomfortable removing dependencies. In fact it's an explicit ask from the serverless team to move towards as close as possible to zero dependencies and inline everything in order to reduce package size and cold start overhead.

rochdev:
@BridgeAR I made the change to use an old version of the library instead, any further concerns?

BridgeAR (approved):
Thanks! :)

About the why: the way the license reference was in there is a way that I would not have felt comfortable with. I would normally have expected it to still be listed in our license file and also to include the full license in the file that you copied. Since it was a 100% inclusion of that, not being something that is commonly written, I think it's best to just stick to the older version.

PR #5369, opened 2025-03-05
[MLOB-2340] feat(vertexai): add `@google-cloud/vertexai` instrumentation 

sabrenner:
### What does this PR do?
Adds **APM** `@google-cloud/vertexai` instrumentation for the following functions:
- `model.generateContent`/`model.generateContentStream`
- `chat.sendMessage`/`chat.sendMessageStream`

For both streamed responses, spans are finished when the stream is consumed.

### Motivation
Furthering auto-instrumentation for different LLM providers. A PR for submitting these span events to LLM Observability will follow.


sabrenner:
working on adding config rules to `system-tests` for the two new config values added

tlhunter (approved):

packages/datadog-plugin-google-cloud-vertexai/src/index.js
```diff
@@ -0,0 +1,195 @@
...
+
+    const history = instance.historyInternal
+    let contents = typeof request === 'string' || Array.isArray(request) ? request : request.contents
+    if (history) {
+      contents = [...history, ...(Array.isArray(contents) ? contents : [contents])]
+    }
+
+    const generationConfig = instance.generationConfig || {}
+    for (const key of Object.keys(generationConfig)) {
+      const transformedKey = key.replace(/([a-z0-9])([A-Z])/g, '$1_$2').toLowerCase()
```
Yun-Kim, line 70: curious why we need to transform the keys here for the tags?

packages/datadog-plugin-google-cloud-vertexai/src/index.js
```diff
@@ -0,0 +1,195 @@
...
+      tags[`vertexai.request.generation_config.${transformedKey}`] = JSON.stringify(generationConfig[key])
```
Yun-Kim, line 71: do we need to JSON.stringify each config setting value? Can't we just directly cast as a string instead?

Yun-Kim (commented):
Mostly clarification questions but otherwise great work!

packages/datadog-instrumentations/src/google-cloud-vertexai.js
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    const ctx = {
+      request,
+      instance: this,
+      resource: [this.constructor.name, generateStream.name].join('.'),
+      stream: true
+    }
+
+    return vertexaiTracingChannel.start.runStores(ctx, () => {
```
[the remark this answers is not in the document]
sabrenner: There's some overview docs here: https://nodejs.org/api/diagnostics_channel.html#channelrunstorescontext-fn-thisarg-args, as well as other functions in `diagnostics_channel` (which `dc-polyfill` uses under the hood, but polyfills some functions that are incompatible with older node versions or broken on certain node versions).

My best explanation of our use here is it enters a given property on the `ctx` object (in this case, it binds the return value from the `bindStart` method in the plugin file) to a given `AsyncLocalStorage` instance (which is what we use for context management across async call stacks). This is done in [this block here](https://github.com/DataDog/dd-trace-js/blob/52279e0e0fbf4777f4be48c11659df6dc7b47f19/packages/dd-trace/src/plugins/plugin.js#L33-L37), which takes the return value from `transform` (which here is what is returned from `bindStart`) and returns it as the bound function for this specific `runStores` event. The idea is you might have multiple `AsyncLocalStorage` stores that listen to this event and want to take in the context and enter something into its store as a result on the start of a function.

What it lets us do is generically enter DD stores into our global `AsyncLocalStore` instance, so that parentage is maintained in sub-calls (e.g. the HTTP span that is generated as a result of this VertexAI one).

packages/datadog-instrumentations/src/google-cloud-vertexai.js
```diff
@@ -0,0 +1,102 @@
...
+  if (err) {
+    ctx.error = err
+    vertexaiTracingChannel.error.publish(ctx)
+  }
+
+  ctx.result = { response }
+
+  if (publishEndEvent) vertexaiTracingChannel.end.publish(ctx)
+
+  vertexaiTracingChannel.asyncEnd.publish(ctx)
```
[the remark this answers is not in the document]
sabrenner: `end` and `asyncEnd` are two separate events here. for the tracing one we only care about `asyncEnd`, as that's when the end result of an async operation actually happens. here's how i think about it:
```javascript
const promise = someAsyncFunctionThatNeedsToBeAwaited() // the `end` event triggers after this function call
const resolved = await promise // the `asycnEnd` event triggers after the promise is awaited/resolved
```
it's not too relevant here, as we only use the `asyncEnd` function in the plugin to read everything after the call is actually resolved. but, we'll use the `end` event in our LLMObs plugin to restore our LLMObs store instance (I can explain better over Zoom likely, but there are some improvements we could make that would let us use our own store binding, but would require a bit of a refactor in the plugin system).

so yeah, we're fine with publishing both twice. we definitely wouldn't want to publish the same event twice, though (publishing `asyncEnd` twice, for example)

packages/datadog-instrumentations/src/google-cloud-vertexai.js
```diff
@@ -0,0 +1,102 @@
...
+
+addHook({
+  name: '@google-cloud/vertexai',
+  file: 'build/src/models/generative_models.js',
+  versions: ['>=1']
+}, exports => {
+  const GenerativeModel = exports.GenerativeModel
+
+  shimmer.wrap(GenerativeModel.prototype, 'generateContent', wrapGenerate)
+  shimmer.wrap(GenerativeModel.prototype, 'generateContentStream', wrapGenerateStream)
```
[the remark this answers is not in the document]
sabrenner: the only functions provided by the vertexai sdk for node.js are async, no sync versions 😄

packages/datadog-plugin-google-cloud-vertexai/src/index.js
sabrenner: this was more for consistency (after working w/ our shared tests). for example, in node.js you can set a `generationConfig` as:
```javascript
const vertexModel = vertexai.getGenerativeModel({
  model: 'gemini-1.5-flash-002',
  generationConfig: {
    maxOutputTokens: 1
  }
});
```
whereas on python it would be `max_output_tokens`. setting the key directly from the config itself would look like `vertexai.request.generation_config.maxOutputTokens` instead of `...generation_config.max_output_tokens`. so, this is really just for consistency (even though I guess the resource names are different here, `generateContent` vs `generate_content`, etc.). i don't have much of a preference, let me know what you think!

packages/datadog-plugin-google-cloud-vertexai/src/index.js
sabrenner: it wouldn't work for object-type configurations. For example, `responseSchema`. casting it to a string would yield `[object Object]`, while `JSON.stringify` would stringify it as `'{"type":...,"properties":...,"required":...'` (or however it may look):
```
> const obj = { a: 5 }
> String(obj)
'[object Object]'
> JSON.stringify(obj)
'{"a":5}'
```
i actually think besides `responseSchema` everything would be fine casting as a string. I don't think the `JSON.stringify` has too much of a perf impact over just casting to a String directly. let me know your thoughts, we could just special-case `responseSchema` off of the tags!

packages/datadog-plugin-google-cloud-vertexai/src/index.js
sabrenner: i guess we could also hard-code a mapping, which could maybe be a little easier to read

packages/datadog-plugin-google-cloud-vertexai/src/index.js
Yun-Kim: Ahh thanks for explaining. This is fine for me as-is if you don't think the performance impact is that significant.

packages/datadog-instrumentations/src/google-cloud-vertexai.js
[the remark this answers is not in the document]
Yun-Kim: Will definitely benefit from a more in-depth node tracing session in the future (no need for one now), but this explanation makes sense to me!

Yun-Kim (approved):

PR #5374, opened 2025-03-06
Implement Config Chaining

khanayan123:
### What does this PR do?
- Enhances configuration telemetry to capture data from all sources, sending updates only when changes occur.
- Adds a seq_id to each payload, allowing the backend to assume the configuration with the highest seq_id is the one in active use by the tracer.

### Motivation
POC for configuration visibility.

### Note:
Requires system test PR to be merged in first: https://github.com/DataDog/system-tests/pull/4303


packages/dd-trace/src/telemetry/telemetry.js
```diff
@@ -332,16 +334,16 @@ function updateConfig (changes, config) {
...
 
   for (const change of changes) {
     const name = nameMapping[change.name] || change.name
 
-    names.push(name)
+    updatedTuples.add(`${name}|${change.origin}`)
     const { origin, value } = change
-    const entry = { name, value, origin }
+    const entry = { name, value, origin, seq_id: seqId }
+    seqId++
```
watson: Nit / personal preference:

```suggestion
    const entry = { name, value, origin, seq_id: seqId++ }
```

yarn.lock
```diff
@@ -96,7 +96,7 @@
   resolved "https://registry.yarnpkg.com/@babel/helper-plugin-utils/-/helper-plugin-utils-7.26.5.tgz#18580d00c9934117ad719392c4f6585c9333cc35"
   integrity sha512-RS+jZcRdZdRFzMyr+wcsaqOmld1/EqTghfaBGQQd/WnRdzdlvSZ//kF7U8VQTxf1ynZ4cjUcYgjVGx13ewNPMg==
 
-"@babel/helper-string-parser@^7.25.9", "@babel/helper-string-parser@^7.27.1":
+"@babel/helper-string-parser@^7.27.1":
```
watson: Why are these changes to `yarn.lock` included?

packages/dd-trace/src/telemetry/telemetry.js
```diff
@@ -357,8 +359,8 @@ function updateConfig (changes, config) {
     configuration.push(entry)
   }
 
-  function isNotModified (entry) {
-    return !names.includes(entry.name)
+  function isNotModified (oldEntry) {
+    return !updatedTuples.has(`${oldEntry.name}|${oldEntry.origin}`)
```
watson: If you change `updatedTuples` to a `Map` instead, it's more performant to do this check by a factor of around 5 (probably renamed the variable as well):

```suggestion
    return updatedTuples.get(oldEntry.name) !== oldEntry.origin
```

packages/dd-trace/test/telemetry/index.spec.js
```diff
@@ -464,22 +450,37 @@ describe('Telemetry extended heartbeat', () => {
...
-          '[{"service":"*","sampling_rate":1},{"service":"svc*","resource":"*abc","name":"op-??","tags":{"tag-a":"ta-v*","tag-b":"tb-v?","tag-c":"tc-v"},"sample_rate":0.5}]',
-          origin: 'code'
-        }
-      ])
+    const expectedConfigListWithSamplingRules = expectedConfigList.concat([
+      {
+        name: 'DD_TRACE_SAMPLING_RULES',
+        value: '[{"service":"*","sampling_rate":1},' +
+        '{"service":"svc*","resource":"*abc","name":"op-??",' +
+        '"tags":{"tag-a":"ta-v*","tag-b":"tb-v?","tag-c":"tc-v"},"sample_rate":0.5}]',
```
watson:
```suggestion
        value: '[{"service":"*","sampling_rate":1},' +
          '{"service":"svc*","resource":"*abc","name":"op-??",' +
          '"tags":{"tag-a":"ta-v*","tag-b":"tb-v?","tag-c":"tc-v"},"sample_rate":0.5}]',
```

rochdev (approved):

integration-tests/telemetry.spec.js
```diff
@@ -60,5 +61,23 @@ describe('telemetry', () => {
...
+        const { configuration } = msg.payload.payload
+        const logInjectionEntries = configuration.filter(entry => entry.name === 'DD_LOG_INJECTION')
+
+        if (logInjectionEntries.length !== 3) return
+
+        const [first, second, third] = logInjectionEntries
+
+        if (first.value !== false || first.origin !== 'default') return
+        if (second.value !== true || second.origin !== 'env_var' || second.seq_id <= first.seq_id) return
+        if (third.value !== false || third.origin !== 'code' || third.seq_id <= second.seq_id) return
```
watson: This is all very brittle and might easily make the tests gang if either the order in the array is not maintained, or if any of the 3 elements does not contain what is expected. A more robust implementation would be:

```suggestion
        assertObjectContains(configuration, [
          {
            name: 'DD_LOG_INJECTION',
            value: false,
            origin: 'default'
          },
          {
            name: 'DD_LOG_INJECTION',
            value: true,
            origin: 'env_var'
          },
          {
            name: 'DD_LOG_INJECTION',
            value: false,
            origin: 'code'
          }
        ])
```

You can get `assertObjectContains` from `./helpers` above.

If there's a risk that the first payload received doesn't contain this, you can guard it with a `if (configuration.some(entry => entry.name === 'DD_LOG_INJECTION')) {...}` but that would make the test potentially hang if it doesn't send this, so it's best if it isn't nessecary.

packages/dd-trace/src/telemetry/telemetry.js
```diff
@@ -334,12 +336,12 @@ function updateConfig (changes, config) {
   const host = createHostObject()
 
   const changed = configWithOrigin.size > 0
+  const configuration = []
```
watson: This doesn't seem to be used anywhere?

packages/dd-trace/src/config.js
```diff
@@ -1465,43 +1462,50 @@ class Config {
...
+    return [
+      { container: this._remote, origin: 'remote_config', unprocessed: this._remoteUnprocessed },
+      { container: this._options, origin: 'code', unprocessed: this._optsUnprocessed },
+      { container: this._fleetStableConfig, origin: 'fleet_stable_config', unprocessed: {} },
+      { container: this._env, origin: 'env_var', unprocessed: this._envUnprocessed },
+      { container: this._localStableConfig, origin: 'local_stable_config', unprocessed: {} },
+      { container: this._calculated, origin: 'calculated', unprocessed: {} },
+      { container: this._defaults, origin: 'default', unprocessed: {} }
+    ]
+  }
```
watson: Instead of re-building this complex object in memory every time we need it, I would simply make this constant outside the `Config` object:

```js
const sourcesOrder = [
  { containerProperty: '_remote', origin: 'remote_config', unprocessedProperty: '_remoteUnprocessed' },
  { containerProperty: '_options', origin: 'code', unprocessedProperty: '_optsUnprocessed' },
  { containerProperty: '_fleetStableConfig', origin: 'fleet_stable_config' },
  { containerProperty: '_env', origin: 'env_var', unprocessedProperty: '_envUnprocessed' },
  { containerProperty: '_localStableConfig', origin: 'local_stable_config' },
  { containerProperty: '_calculated', origin: 'calculated' },
  { containerProperty: '_defaults', origin: 'default' }
]
```

Then I would loop over that when I need to access the containers. That would simplify the rest of the code. It's a bit messy to lay out in individual review comments, so here's a diff of the changes I'd do in the entire file:

```diff
diff --git a/packages/dd-trace/src/config.js b/packages/dd-trace/src/config.js
index 97ec8ae46..dfe9721e9 100644
--- a/packages/dd-trace/src/config.js
+++ b/packages/dd-trace/src/config.js
@@ -231,6 +231,16 @@ function reformatSpanSamplingRules (rules) {
   })
 }
 
+const sourcesOrder = [
+  { containerProperty: '_remote', origin: 'remote_config', unprocessedProperty: '_remoteUnprocessed' },
+  { containerProperty: '_options', origin: 'code', unprocessedProperty: '_optsUnprocessed' },
+  { containerProperty: '_fleetStableConfig', origin: 'fleet_stable_config' },
+  { containerProperty: '_env', origin: 'env_var', unprocessedProperty: '_envUnprocessed' },
+  { containerProperty: '_localStableConfig', origin: 'local_stable_config' },
+  { containerProperty: '_calculated', origin: 'calculated' },
+  { containerProperty: '_defaults', origin: 'default' }
+]
+
 class Config {
   constructor (options = {}) {
     if (!isInServerlessEnvironment()) {
@@ -1462,20 +1472,20 @@ class Config {
   // for telemetry reporting, `name`s in `containers` need to be keys from:
   // https://github.com/DataDog/dd-go/blob/prod/trace/apps/tracer-telemetry-intake/telemetry-payload/static/config_norm_rules.json
   _merge () {
-    // Use reverse order for merge (lowest priority first)
-    const sources = [...this._getAllSources()].reverse()
-
     const changes = []
 
     for (const name in this._defaults) {
-      for (const { container, origin, unprocessed } of sources) {
+      // Use reverse order for merge (lowest priority first)
+      for (let i = sourcesOrder.length - 1; i >= 0; i--) {
+        const { containerProperty, origin, unprocessedProperty } = sourcesOrder[i]
+        const container = this[containerProperty]
         const value = container[name]
-        if ((value !== null && value !== undefined) || container === this._defaults) {
+        if (value != null || container === this._defaults) {
           this._setAndTrackChange({
             name,
             value,
             origin,
-            unprocessedValue: unprocessed[name],
+            unprocessedValue: unprocessedProperty === undefined ? undefined : this[unprocessedProperty][name],
             changes
           })
         }
@@ -1485,35 +1495,12 @@ class Config {
     updateConfig(changes, this)
   }
 
-  // Single source of truth for all config sources (highest priority first)
-  _getAllSources () {
-    return [
-      { container: this._remote, origin: 'remote_config', unprocessed: this._remoteUnprocessed },
-      { container: this._options, origin: 'code', unprocessed: this._optsUnprocessed },
-      { container: this._fleetStableConfig, origin: 'fleet_stable_config', unprocessed: {} },
-      { container: this._env, origin: 'env_var', unprocessed: this._envUnprocessed },
-      { container: this._localStableConfig, origin: 'local_stable_config', unprocessed: {} },
-      { container: this._calculated, origin: 'calculated', unprocessed: {} },
-      { container: this._defaults, origin: 'default', unprocessed: {} }
-    ]
-  }
-
-  _getContainersAndOriginsOrdered () {
-    const sources = this._getAllSources()
-    const containers = sources.map(s => s.container)
-    const origins = sources.map(s => s.origin)
-
-    return { containers, origins }
-  }
-
   getOrigin (name) {
-    const { containers, origins } = this._getContainersAndOriginsOrdered()
-
-    for (let i = 0; i < containers.length; i++) {
-      const container = containers[i]
+    for (const { containerProperty, origin } of sourcesOrder) {
+      const container = this[containerProperty]
       const value = container[name]
       if (value != null || container === this._defaults) {
-        return origins[i]
+        return origin
       }
     }
   }
```

watson (changes requested):

packages/dd-trace/src/telemetry/telemetry.js
```diff
@@ -307,7 +307,7 @@ function formatMapForTelemetry (map) {
 
 const nameMapping = {
   sampleRate: 'DD_TRACE_SAMPLE_RATE',
-  logInjection: 'DD_LOGS_INJECTION',
+  logInjection: 'DD_LOG_INJECTION',
```
rochdev: Is this expected? All our public docs reference `DD_LOGS_INJECTION`.

integration-tests/telemetry.spec.js
```diff
@@ -60,5 +61,31 @@ describe('telemetry', () => {
         }
       }, null, 'app-dependencies-loaded', 1)
     })
+
+    it('Assert configuration chaining data is sent', (done) => {
+      agent.assertTelemetryReceived(msg => {
+        if (msg.payload.request_type !== 'app-started') return
+
+        const { configuration } = msg.payload.payload
+        assertObjectContains(configuration, [
```
watson: You need to import `assertObjectContains` from the `./helpers` module at the top of the file

packages/dd-trace/src/config.js
```diff
@@ -8,8 +8,6 @@ const log = require('./log')
 const pkg = require('./pkg')
 const coalesce = require('koalas')
 const tagger = require('./tagger')
-const get = require('../../datadog-core/src/utils/src/get')
-const has = require('../../datadog-core/src/utils/src/has')
```
watson, line 12: I just realized that this file is the only one that uses `get`, `has` and `set`, so these two files and their tests should actually be deleted as part fo this PR

packages/dd-trace/src/telemetry/telemetry.js
khanayan123: Yes, i have created a jira ticket to fix this. Because of the circular dependency between this PR and system tests I'am going to fix this in a subsequent PR

watson (dismissed):
Nice work!

PR #5375, opened 2025-03-06
dbm: tedious (sql server) service mode

tlhunter:
### What does this PR do?
- adds DBM service mode tracing to tedious (MS SQL Server)

### Motivation
- users want it

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/datadog-instrumentations/src/tedious.js
+++ b/packages/datadog-instrumentations/src/tedious.js
@@ -16,7 +16,7 @@ addHook({ name: 'tedious', versions: ['>=1.0.0'] }, tedious => {
       return makeRequest.apply(this, arguments)
     }
 
-    const queryOrProcedure = getQueryOrProcedure(request)
+    const [queryOrProcedure, queryParent, queryField] = getQueryOrProcedure(request)
 
     if (!queryOrProcedure) {
       return makeRequest.apply(this, arguments)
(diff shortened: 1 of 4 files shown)


packages/datadog-plugin-tedious/test/index.spec.js
```diff
@@ -53,7 +52,6 @@ describe('Plugin', () => {
           config.password = MSSQL_PASSWORD
         }
 
-        connectionIsClosed = false
```
tlhunter, line 56: This was never written to

packages/dd-trace/src/plugins/database.js
```diff
@@ -63,7 +63,7 @@ class DatabasePlugin extends StoragePlugin {
     return tracerService
   }
 
-  createDbmComment (span, serviceName, isPreparedStatement = false) {
+  createDbmComment (span, serviceName, disableFullMode = false) {
```
tlhunter, line 66: Renaming this to be more universal

packages/datadog-instrumentations/src/tedious.js
```diff
@@ -28,7 +28,9 @@ addHook({ name: 'tedious', versions: ['>=1.0.0'] }, tedious => {
     const connectionConfig = this.config
 
     return asyncResource.runInAsyncScope(() => {
-      startCh.publish({ queryOrProcedure, connectionConfig })
+      const payload = { queryOrProcedure, connectionConfig }
+      startCh.publish(payload)
+      queryParent[queryField] = payload.sql
```
sabrenner, line 33: maybe we could pass `queryParent` and `queryField` on the payload to the start event so we can do this on the subscriber side when doing `injectDbmQuery`, instead of the instrumentation having to know about it and set it here (i think it should be fine, since they're all references, but correct me if i'm wrong). i don't mind much either way. wdyt?

sabrenner (approved):
logic here lgtm, might be worth getting a look from DBM folks too to see if the content of the modified query looks ok!

packages/datadog-instrumentations/src/tedious.js
tlhunter: You might be onto something. Ideally `getQueryOrProcedure` could do the mutation but the publish expects the original object.

PR #3259, opened 2023-06-20
Telemetry metrics for OTel

Qard:
### What does this PR do?

Adds Telemetry Metrics API and uses it to track types of spans created and finished.

### Motivation

We want visibility into OTel adoption so we should record which way spans are created.

PR #5382, opened 2025-03-10
Remove version tag in telemetry metrics

smola:
### What does this PR do?
Remove `version` tag from telemetry metrics.

### Motivation
Introduced in https://github.com/DataDog/dd-trace-js/pull/3259 as part of OTel metrics. However, `version` is not a valid tag in our Instrumentation Telemetry metrics definition (not OTel).

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


BridgeAR (commented):
Should we just change the tag name to a supported one instead? While we'd only need the version once when starting, so removing it here is probably a good idea in general.

zacharycmontoya (approved):
LGTM but you may want to wait on an approval from a regular dd-trace-js contributor

smola:
@BridgeAR Telemetry intake already tags every metric with `language_version` based on telemetry headers.

PR #5383, opened 2025-03-10
Fix reporting of profiling_enabled in startup log

szegedi:
### What does this PR do?
Fixes the reporting of `profiling_enabled` value in startup log.

### Motivation
It was always reporting true due to the way the `profiling.enabled` value is not a boolean, but rather a three-valued string. The problem was basically that `!!'false' === true`

### Additional Notes
Jira: [PROF-11492]

[PROF-11492]: https://datadoghq.atlassian.net/browse/PROF-11492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/src/startup-log.js
+++ b/packages/dd-trace/src/startup-log.js
@@ -86,7 +86,8 @@ function tracerInfo () {
 
   out.log_injection_enabled = !!config.logInjection
   out.runtime_metrics_enabled = !!config.runtimeMetrics
-  out.profiling_enabled = !!(config.profiling || {}).enabled
+  const profilingEnabled = config.profiling?.enabled
+  out.profiling_enabled = profilingEnabled === 'true' || profilingEnabled === 'auto'
   Object.assign(out, getIntegrationsAndAnalytics())
 
   out.appsec_enabled = !!config.appsec.enabled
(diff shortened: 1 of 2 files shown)


BridgeAR (changes requested):
Could you please add a test for this? :)

IlyasShabi (dismissed):

nsavoire (approved):

tlhunter (dismissed):

rochdev (approved):

BridgeAR (approved):
Thanks for adding the test. The test does seem a bit brittle, when it comes to changes, while I guess it does not hurt too much either.

I would have probably just called `tracerInfo()` directly (it is also exported) and just checked that. That way the need to rely on console.info being called with this part as start would be gone.

szegedi:
> I would have probably just called tracerInfo() directly (it is also exported) and just checked that. That way the need to rely on console.info being called with this part as start would be gone.

That's fair – I ended up with this approach as I tried to keep the testing methodology consistent with the other test was already in the file.

PR #5386, opened 2025-03-10
Add consistent sampling via knuth's method

bm1549:
### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/dd-trace/src/sampler.js
```diff
@@ -18,8 +22,9 @@ class Sampler {
   /**
    * @returns {boolean}
    */
-  isSampled () {
-    return this._rate === 1 || Math.random() < this._rate
+  isSampled (context) {
+      return this._rate === 1 ||
+          ((context.toTraceIdNumber(false) * KNUTH_FACTOR) % MAX_TRACE_ID) <= this._sampling_id_threshold
```
BridgeAR, line 27: I fail to understand how this shall work.

As far as I understand the Knuth sampling, it has a reservoir of samples and it replaces those during a processing time window depending on the input. This is done randomly with a decreasing chance of replacing items already in the reservoir. Here, the chance seems constant, if I am not mistaken. I read it as if it never takes into account how many we have already processed in a specific time window.

I might also misunderstand how it should work.

PR #5392, opened 2025-03-11
Adds JSDoc types to much of the sampling code

bm1549:
### What does this PR do?

Adds JSDoc types to the sampling code so that it's easier to work in

### Motivation

I was investigating how we could support Knuth-method sampling in #5386 and split this PR out from it

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/dd-trace/src/priority_sampler.js
```diff
@@ -39,12 +39,28 @@ const DEFAULT_KEY = 'service:,env:'
...
+/**
+ * from config.js
+ * @typedef { sampleRate: number, provenance: string, rateLimit: number, rules: SamplingRule[] } SamplingConfig
+ *
+ * empirically defined
+ * @typedef {2|-1|1|0} SamplingPriority
+ */
 class PrioritySampler {
+  /**
+   * @param env {string}
```
BridgeAR, line 51: Nit: Would you mind renaming this to `environment`? I feel we could generally use less short versions to make things more explicit :)

packages/dd-trace/src/priority_sampler.js
```diff
@@ -150,7 +202,15 @@ class PrioritySampler {
       : this._getPriorityByAgent(context)
   }
 
-  _getPriorityFromTags (tags) {
+  /**
+   *
+   * @param tags {Map<string, Object>}
```
BridgeAR, line 207: I don't know exactly what JSDoc refers to when it being an Object but at least in TS it's not ideal to use. Should we use something like `Record<string | symbol, unknown>`?

packages/dd-trace/src/priority_sampler.js
```diff
@@ -166,6 +226,13 @@ class PrioritySampler {
     }
   }
 
+  /**
+   *
+   * @param context {DatadogSpanContext}
+   * @param rule {SamplingRule}
+   * @returns {SamplingPriority}
+   * @private
```
BridgeAR, line 234: Instead of marking something as private, should we not just use a proper private JS property (`#privateProperty`), if it is not currently somewhere "misused" internally? :)

BridgeAR (dismissed):
LGTM

One suggestion: instead of directly writing JSDoc, we could also write TS documentation in a separate file for each class and then just mark the class that is implements the interface. That way we can be more specific about types than it's possible with JSDoc.

packages/dd-trace/src/priority_sampler.js
bm1549: Usually I'd agree, but `env` is a standard across all of Datadog ([src](https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging/?tab=kubernetes)) so it's more understandable in its short-form

packages/dd-trace/src/priority_sampler.js
bm1549: Unfortunately we've already broken this by calling it [here](https://github.com/DataDog/dd-trace-js/blob/d88b6625a6ada85334b8585d302083084dabecf6/packages/dd-trace/test/standalone/tracesource_priority_sampler.spec.js#L51)

I can remove the `@private` declaration

packages/dd-trace/src/priority_sampler.js
```diff
@@ -166,7 +223,13 @@ class PrioritySampler {
...
   }
 
-  _getPriorityByRule (context, rule) {
+  /**
+   *
+   * @param context {DatadogSpanContext}
+   * @param rule {SamplingRule}
+   * @returns {SamplingPriority}
+   */
+  #_getPriorityByRule (context, rule) {
```
BridgeAR, line 232:
```suggestion
  #getPriorityByRule (context, rule) {
```

packages/dd-trace/src/priority_sampler.js
```diff
@@ -185,7 +254,13 @@ class PrioritySampler {
...
   }
 
-  _getPriorityByAgent (context) {
+  /**
+   *
+   * @param context {DatadogSpanContext}
+   * @returns {SamplingPriority}
+   * @private
+   */
+  #_getPriorityByAgent (context) {
```
BridgeAR, line 263:
```suggestion
  #getPriorityByAgent (context) {
```

And also the other entries :)

The underscore is a legacy way where private properties didn't exist to "mark" something as "internal".

BridgeAR (approved):
LGTM! Just left a nit. Feel free to ignore it.

packages/dd-trace/src/priority_sampler.js
simon-id, line 1: not a hard blocker, but it would be interesting to ask roch or bryan, in case some people already use our private fields in the wild, and if this is going to break them. I know it's not a breaking change since it's not part of the public API, but you never know

rochdev: I'm not too worried about the sampler internal APIs being used in the wild. Most usage I've seen of private properties is on `span.context()`.

PR #5395, opened 2025-03-12
fix(vitest): fix coverage not being reported when value is zero

giamir:
### What does this PR do?
This PR ensures that code coverage is reported also in cases is zero. Since zero is a falsy value in js the way the conditional was setup before this PR was preventing zero values to be reported.

### Motivation
There are valid cases where test coverage is zero and should be reported like a monorepo project where a workspace is not tested at all.

### DIFF
--- a/packages/datadog-plugin-vitest/src/index.js
+++ b/packages/datadog-plugin-vitest/src/index.js
@@ -300,7 +300,7 @@ class VitestPlugin extends CiPlugin {
         this.testModuleSpan.setTag('error', error)
         this.testSessionSpan.setTag('error', error)
       }
-      if (testCodeCoverageLinesTotal) {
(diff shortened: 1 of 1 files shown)


BridgeAR (changes requested):
Thanks for the PR! The change itself is LGTM.
Could you please also add a test case?

BridgeAR:
This landed is now on the main branch. See https://github.com/DataDog/dd-trace-js/pull/5681

I cherry-picked the commit in that branch (we can't run our CI on external contributions) and added the test.

It will be released soon. Thanks again for your contribution!

giamir:
Thanks for addressing this and apologies about not managing to find the space to write that unit test myself.

juan-fernandez:
thanks for this change @giamir ! 😄 (sry, I couldn't respond earlier)

PR #5396, opened 2025-03-12
Fix logging null when debugging and sending data without any error

FredericEspiau:
### What does this PR do?

If you have debugging on, when you send data without a tracing agent, then you would log `null` every time, because there is no error

### Motivation

Being able to debug easier

### Additional Notes

I didn't write a test because I couldn't find a test for this kind of stuff


rochdev (approved):

packages/dd-trace/src/telemetry/send-data.js
```diff
@@ -100,7 +100,11 @@ function sendData (config, application, host, reqType, payload = {}, cb = () =>
         path: '/api/v2/apmtelemetry'
       }
       if (backendUrl) {
-        request(data, backendOptions, (error) => { log.error('Error sending telemetry data', error) })
+        request(data, backendOptions, (error) => {
+          if (error) {
+            log.error('Error sending telemetry data', error);
```
simon-id, line 105:
```suggestion
            log.error('Error sending telemetry data', error)
```

FredericEspiau: Done 🙏

simon-id (approved):

simon-id:
Recreated in https://github.com/DataDog/dd-trace-js/pull/5480

PR #5399, opened 2025-03-12
lazy load ip extraction

rochdev:
### What does this PR do?

Lazy load IP extraction.

### Motivation

This is an AppSec feature which is disabled by default, and it has significant startup time cost because the code it uses from `net` is fairly complicated. Locally I get a 15% startup overhead reduction from not importing that single file.


simon-id (commented):
so this is going to be a blocking overhead on the first request, what's the latency impact you measured at runtime in absolute numbers ?

rochdev:
> so this is going to be a blocking overhead on the first request, what's the latency impact you measured at runtime in absolute numbers ?

It's not as `normalizeConfig` is called when the library is initialized.

simon-id (approved):

PR #5404, opened 2025-03-12
remove fs dependency from lambda hook

rochdev:
### What does this PR do?

Remove `fs` dependency from Lambda hook.

### Motivation

This was forcing `fs` and its integration to load, thus impacting startup time at worst and showing them incorrectly in the cold start trace at best.

### Additional Notes

It's possible that I may be missing something and there might have been a specific reason to use `fs` here, but it didn't seem like it so I just went ahead with the change.

Ideally this would be rewritten entirely to be like any other instrumentation instead of having this second system work in parallel, but that will require a lot more work, so in the meantime I just addressed the immediate issue shown in the cold start trace.

### DIFF
--- a/packages/dd-trace/src/lambda/runtime/ritm.js
+++ b/packages/dd-trace/src/lambda/runtime/ritm.js
@@ -7,7 +7,6 @@
  */
 'use strict'
 
-const fs = require('fs')
 const path = require('path')
 
 const log = require('../../log')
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/lambda/runtime/ritm.js
```diff
@@ -60,23 +59,18 @@ function _extractModuleNameAndHandlerPath (handler) {
...
-  } else if (fs.existsSync(lambdaStylePath + '.cjs')) {
-    lambdaFilePath += '.cjs'
-  }
-  return lambdaFilePath
+function _getLambdaFilePaths (lambdaStylePath) {
+  return [
+    `${lambdaStylePath}.js`,
+    `${lambdaStylePath}.mjs`,
+    `${lambdaStylePath}.cjs`
+  ]
```
duncanista, line 73: So normally, customers would only have one of this, is it OK to just leave it as is rn?

rochdev: Yes, it will register a hook for all paths, and only the one that is actually used will hit.

duncanista (approved):

PR #5408, opened 2025-03-13
Improve iast mongodb nosql detection removing some false positives

uurien:
### What does this PR do?

* Does not detect a vulnerability when the tainted string is detected in a known safe mongodb query property, like `$eq` or `$in`
* Tainted value should be the same received from the query parameter or request body, if it has been modified, it is not vulnerable, because query syntax can't be modified.
* Fix some tests that weren't not executed before.

### Motivation

Reduce the false positives detected for mongodb databases.

### Additional Notes

[APPSEC-56443]

[APPSEC-56443]: https://datadoghq.atlassian.net/browse/APPSEC-56443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dd-trace/src/appsec/iast/analyzers/nosql-injection-mongodb-analyzer.js
```diff
@@ -119,29 +126,25 @@ class NosqlInjectionMongodbAnalyzer extends InjectionAnalyzer {
...
+          isVulnerable = true
 
-          if (filteredRanges.length > 0) {
-            rangesByKey[nextLevelKeys.join('.')] = filteredRanges
-            allRanges.push(...filteredRanges)
-          }
+          rangesByKey[nextLevelKeys.join('.')] = ranges
+          allRanges.push(range)
         }
-      }, [], 4)
```
CarlesDD, line 144: Why object depth analysis has changed from 4 to 10 (the default one on `iterateMongodbQueryStrings`)?

packages/dd-trace/src/appsec/iast/analyzers/nosql-injection-mongodb-analyzer.js
```diff
@@ -103,7 +106,11 @@ class NosqlInjectionMongodbAnalyzer extends InjectionAnalyzer {
     })
   }
 
-  _isVulnerableRange (range) {
+  _isVulnerableRange (range, value) {
+    const rangeIsWholeValue = range.start === 0 && range.end === value?.length
```
IlyasShabi, line 110: I think it's more safe to check `range` first or:

```suggestion
    const rangeIsWholeValue = range?.start === 0 && range?.end === value?.length
```

uurien: true, as it was checked before, we should continue checking.

packages/dd-trace/src/appsec/iast/analyzers/nosql-injection-mongodb-analyzer.js
```diff
@@ -103,7 +106,11 @@ class NosqlInjectionMongodbAnalyzer extends InjectionAnalyzer {
     })
   }
 
-  _isVulnerableRange (range) {
+  _isVulnerableRange (range, value) {
+    const rangeIsWholeValue = range?.start === 0 && range?.end === value?.length
+
+    if (!rangeIsWholeValue) return false
```
CarlesDD, line 112: There is no test for this new check. We should add it.

CarlesDD (changes requested):

packages/dd-trace/src/appsec/iast/analyzers/nosql-injection-mongodb-analyzer.js
```diff
@@ -162,4 +165,25 @@ class NosqlInjectionMongodbAnalyzer extends InjectionAnalyzer {
   }
 }
 
+function iterateMongodbQueryStrings (target, fn, levelKeys = [], depth = 10, visited = new Set()) {
+  if (target !== null && typeof target === 'object') {
+    if (visited.has(target)) return
+
+    visited.add(target)
```
BridgeAR, line 172: While a set seems the obvious thing to use here to limit the number of lookups, it's likely faster to use an array. The reason is that the default lookup will normally not reach a depth of 10. There is probably a break even point somewhere in-between and it might be interesting to check for that.

Allocating a set and creating the hash is just more expensive than allocating an array and iterating over very few entries.

This will however only work in case we also release the values when moving up. Do we actually have to match all former entries or only only ones that are available in the current leaf?

uurien: You mean using `[].includes(target)` instead of `Set.has(target)`? Yeah, it should work also, I dunno if the difference will be noticed, but I think that we can change it...

> This will however only work in case we also release the values when moving up. Do we actually have to match all former entries or only only ones that are available in the current leaf?

I'm not sure if I understand, the goal of this is to prevent circular dependencies, but also to prevent  checking the same twice, with objects like:

```
{
  a: objectA,
  b: {
      a: objectA
  }
}
```
We don't need to iterate again in `b.a: objectA`

CarlesDD (approved):
LGTM

PR #5414, opened 2025-03-13
fix runtime metrics histogram sending invalid data

### LINKED ISSUE #5389: [BUG]: runtime.node.event_loop.delay.avg metric disappeared
### Tracer Version(s)

5.41.1

### Node.js Version(s)

v22.14.0

### Bug Report

Since the [v5.41.1](https://redirect.github.com/DataDog/dd-trace-js/compare/v5.41.0...v5.41.1), the metric `runtime.node.event_loop.delay.avg` disappeared from our dashboard.

The metrics comes back when we switch back to v5.41.0.

### Reproduction Code

* monitor a node process with dd-trace-js
* add `runtime.node.event_loop.delay.avg` metric to a dashboard
* with dd-trace-js v5.41.0, the metric is visible in the dashboard
* upgrade dd-trace-js to 5.41.1, the metric is no longer sent
* downgrade dd-trace-js to 5.41.0, the metric is sent again.

### Error Logs

_No response_

### Tracer Config

_No response_

### Operating System

_No response_

### Bundling

Unsure

rochdev:
### What does this PR do?

Fix runtime metrics histogram sending invalid data.

### Motivation

When moving to the `MetricsAggregationClient` for aggregations, some of the code was not ported properly, and previous calls to the internal runtime metrics `histogram()` function (which expects a stats object) ended up calling the client directly instead (which expects a numeric value that will be computed internally as a histogram).

Fixes #5389

### Additional Notes

While working on this, I also uncovered another bug where `MetricsAggregationClient` would increment and decrement as gauges by default, when it should have been counters. Only runtime metrics expect `increment` and `decrement` to be gauges because of how it's used in the library.


packages/dd-trace/src/runtime_metrics/runtime_metrics.js
```diff
@@ -282,12 +282,19 @@ function captureNativeMetrics () {
...
 
-  if (tags.length > 0) {
-    for (const tag of tags) {
-      client.histogram(name, stats, tag)
-    }
-  } else {
-    client.histogram(name, stats)
+  if (tags.length === 0) {
+    tags.push(undefined)
+  }
```
tlhunter, line 287: This seems weird and like it might need a comment

tlhunter (approved):

PR #5419, opened 2025-03-14
[Debugger] improve snapshot performance

BridgeAR:
The sync code got a tad optimized (dense instead of sparse arrays and some other minor improvements). The async parts now also run in parallel.
These changes do not make a big difference effectively due to the overhead of the inspector calls themselves.

### DIFF
--- a/integration-tests/debugger/redact.spec.js
+++ b/integration-tests/debugger/redact.spec.js
@@ -2,48 +2,47 @@
 
 const { assert } = require('chai')
 const { setup } = require('./utils')
+const { once } = require('node:events')
 
 // Default settings is tested in unit tests, so we only need to test the env vars here
 describe('Dynamic Instrumentation snapshot PII redaction', function () {
   describe('DD_DYNAMIC_INSTRUMENTATION_REDACTED_IDENTIFIERS=foo,bar', function () {
     const t = setup({ env: { DD_DYNAMIC_INSTRUMENTATION_REDACTED_IDENTIFIERS: 'foo,bar' } })
 
-    it('should respect DD_DYNAMIC_INSTRUMENTATION_REDACTED_IDENTIFIERS', function (done) {
+    it('should respect DD_DYNAMIC_INSTRUMENTATION_REDACTED_IDENTIFIERS', async function () {
       t.triggerBreakpoint()
 
-      t.agent.on('debugger-input', ({ payload: [{ 'debugger.snapshot': { captures } }] }) => {
-        const { locals } = captures.lines[t.breakpoint.line]
-
-        assert.deepPropertyVal(locals, 'foo', { type: 'string', notCapturedReason: 'redactedIdent' })
-        assert.deepPropertyVal(locals, 'bar', { type: 'string', notCapturedReason: 'redactedIdent' })
-        assert.deepPropertyVal(locals, 'baz', { type: 'string', value: 'c' })
-
-        // existing redaction should not be impacted
-        assert.deepPropertyVal(locals, 'secret', { type: 'string', notCapturedReason: 'redactedIdent' })
-
-        done()
-      })
+      const promise = once(t.agent, 'debugger-input')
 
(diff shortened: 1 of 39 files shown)


integration-tests/debugger/redact.spec.js
```diff
@@ -2,48 +2,47 @@
...
   describe('DD_DYNAMIC_INSTRUMENTATION_REDACTED_IDENTIFIERS=foo,bar', function () {
     const t = setup({ env: { DD_DYNAMIC_INSTRUMENTATION_REDACTED_IDENTIFIERS: 'foo,bar' } })
 
-    it('should respect DD_DYNAMIC_INSTRUMENTATION_REDACTED_IDENTIFIERS', function (done) {
+    it('should respect DD_DYNAMIC_INSTRUMENTATION_REDACTED_IDENTIFIERS', async function () {
       t.triggerBreakpoint()
 
-      t.agent.on('debugger-input', ({ payload: [{ 'debugger.snapshot': { captures } }] }) => {
-        const { locals } = captures.lines[t.breakpoint.line]
+      const promise = once(t.agent, 'debugger-input')
```
BridgeAR, line 15: Note: I personally like this way of writing tests. That way it's somewhat easier for me to follow the logic and the events are always definitely going to cause the error to fail in case of an error.

integration-tests/debugger/snapshot.spec.js
```diff
@@ -208,11 +208,11 @@ describe('Dynamic Instrumentation', function () {
...
 
-          assert.deepEqual(Object.keys(locals), [
-            // Up to 3 properties from the local scope
-            'request', 'nil', 'undef',
+          assert.deepStrictEqual(Object.keys(locals), [
             // Up to 3 properties from the closure scope
-            'fastify', 'getUndefined'
+            'fastify', 'getUndefined',
+            // Up to 3 properties from the local scope
+            'request', 'nil', 'undef'
```
BridgeAR, line 215: The order should not be important, right?

BridgeAR:
Seems like this is triggering the AsyncHooks issue we faced recently constantly on 18.
I guess we have to consider waiting for a new major release where we drop support for Node.js 18.

packages/dd-trace/src/debugger/devtools_client/snapshot/processor.js
```diff
@@ -86,26 +92,26 @@ function getObjectValue (obj, maxLength) {
...
+    case 'generator':
+      // Use `subtype` instead of `className` to make it obvious it's a generator
+      return toObject(obj.subtype, obj.properties, maxLength)
     case 'arraybuffer':
       return toArrayBuffer(obj.className, obj.properties, maxLength)
+    case 'weakmap':
+      return toMap(obj.className, obj.properties, maxLength)
+    case 'weakset':
+      return toSet(obj.className, obj.properties, maxLength)
+      // case 'iterator': // TODO: I've not been able to trigger this subtype
```
watson, line 114:
```suggestion
    // case 'iterator': // TODO: I've not been able to trigger this subtype
```

watson (dismissed):
Once the `.github` changes have been merged in a separate PR, this looks very good 👍

packages/dd-trace/src/debugger/devtools_client/snapshot/processor.js
```diff
@@ -18,8 +18,12 @@ function processProperties (props, maxLength) {
   const result = {}
 
   for (const prop of props) {
-    // TODO: Hack to avoid periods in keys, as EVP doesn't support that. A better solution can be implemented later
-    result[prop.name.replaceAll('.', '_')] = getPropertyValue(prop, maxLength)
+    // TODO: Hack to avoid periods in keys, as EVP doesn't support that. A better solution can be implemented later)
```
watson, line 21:
```suggestion
    // TODO: Hack to avoid periods in keys, as EVP doesn't support that. A better solution can be implemented later
```

watson (approved):

PR #5423, opened 2025-03-17
[Code Origin for Spans] Add support for Express entry spans

watson:
### What does this PR do?

Add support for Express entry spans in Code Origin for Spans.

This is a follow up to PR #4449, which added support for Fastify entry spans.

### Motivation

Express is the most popular web framework in the Node.js ecosystem. Having support for this framework is vital for adoption.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


watson:
* **#5423** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5423?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5423?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/datadog-plugin-express/src/code_origin.js
```diff
@@ -0,0 +1,32 @@
...
+      const tags = layer[kCodeOriginForSpansTagsSym]
+      if (!tags) return
+      const context = web.getContext(req)
+      context.span?.addTags(tags)
+    })
+
+    this.addSub('apm:express:route:added', ({ topOfStackFunc, layer }) => {
+      if (Object.hasOwn(layer, kCodeOriginForSpansTagsSym)) return
+
+      layer[kCodeOriginForSpansTagsSym] = entryTags(topOfStackFunc, 1)
```
watson, line 27: After testing this on a real life Express app, I found that the top frame this `1` is skipping (`shimmer.wrapFunction`) isn't always present, in which case the code is skipping the stack frame we actually care about.

More research is needed to make this code more resilient. However, I'll mark this PR as "Draft" and table this work to focus on other more pressing issues.

watson: For some reason, the app used in tests no longer has the issue that required us to skip the top frame. I've not been able to reproduce the issue either. So without any more to go by, I've removed the "skip-top-frame-fix" and now just always pick the actual top frame as we would normally expect to do. I think this might be due to some of the recent changes to shimmer or how Express is being instrumented. If that's the case, we are safe to move on.

tylfin (approved):

wconti27 (approved):
Code change looks good to me!

packages/datadog-plugin-express/src/code_origin.js
```diff
@@ -0,0 +1,30 @@
+'use strict'
+
+const { entryTags } = require('../../datadog-code-origin')
+const Plugin = require('../../dd-trace/src/plugins/plugin')
+const web = require('../../dd-trace/src/plugins/util/web')
+
+class ExpressCodeOriginForSpansPlugin extends Plugin {
```
tlhunter, line 7: If I'm understanding our wiki page right we'll ultimately need one of these code origin classes / files for all web servers and for all outbound / client integrations?

packages/datadog-instrumentations/src/router.js
```diff
@@ -153,6 +154,10 @@ function createWrapRouterMethod (name) {
         this.stack = [{ handle: this.stack }]
       }
 
+      if (routeAddedChannel.hasSubscribers) {
+        routeAddedChannel.publish({ topOfStackFunc: methodWithTrace, layer: this.stack[0] })
```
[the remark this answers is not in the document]
watson: Sorry, I didn't see your comment before merging the PR. I'll try to address it here, but let me know if this is a show stopper if it can't be dealt with in a follow up PR if needed.

The `topOfStackFunc` function is basically used for making the stack trace generation more performant, as it's used as the 2nd argument to `Error.captureStackTrace`. It would probably work without it, but it would be generating some extra stack frames that we don't need to find the closest user-land stack frame.

I don't like this name either, or the fact that we need to pass it around. It's an ugly API / internal implementation detail, but I'm not sure how to have the performance gains without this 🤷

PR #5435, opened 2025-03-18
fix connection pool error in mongoose tests

rochdev:
### What does this PR do?

Fix connection pool error in mongoose tests.

### Motivation

There have been a lot of flakiness around Mongoose tests, usually with either a `PoolClosedError` or `MongoExpiredSessionError` error. I'm not sure the exact underlying cause, but switching to `beforeEach` for the connection instead of `before` seems to resolve it after ~100 local runs.

### DIFF
--- a/packages/datadog-plugin-mongoose/test/index.spec.js
+++ b/packages/datadog-plugin-mongoose/test/index.spec.js
@@ -29,11 +29,11 @@ describe('Plugin', () => {
         })
       }
 
-      before(() => {
+      beforeEach(() => {
         return agent.load(['mongodb-core'])
       })
 
-      before(async () => {
+      beforeEach(async () => {
         tracer = require('../../dd-trace')
 
         mongoose = require(`../../../versions/mongoose@${version}`).get()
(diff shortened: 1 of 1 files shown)


rochdev:
> I am a bit puzzled that this improves the situation.

This is actually a pretty common occurrence, and when you think about the errors, it actually makes sense. Basically, this PR changes 2 main things:

1) Ensures that there is a new agent for every test. This prevents previous tests from continuing to send data to the agent which can sometimes cause weird side-effects like an unexpected span received.
2) Ensures that there is a new connection for every test. Similar to the point above, this means that anything that could have been in-flight in the previous test won't cause side-effects in the current test as it uses a brand new connection, and it also ensures that the previous test is not leaking anyway since its connection is destroyed.

> Did you check if it's actually a specific one of these needing to be called each time?

I would assume `connect()` since all the errors are from a connection problem (pool and session errors). It also seems to only happen for the peer service test, but I wasn't able to find anything in that test specifically that is any different than other tests, it might just be what one of the previous tests is doing that has a side-effect on this one. It's generally better practice to just use a brand new connection and agent for each test for better isolation anyway, the only reason `before` is used in a few places was to make the tests run faster, but it doesn't have that much of an impact and can cause this kind of issue, so it's best to avoid it whenever a problem arises.

BridgeAR (approved):
LGTM

PR #5436, opened 2025-03-19
fix mysql/mysql2/pg dbm trace id tests

rochdev:
### What does this PR do?

Fix mysql/mysql2/pg DBM trace ID tests.

### Motivation

The tests were trying to guess the start of the trace ID instead of using the real one, which was often a byte off. This PR changes the code to check against the actual trace ID.

### DIFF
--- a/packages/datadog-plugin-mysql/test/index.spec.js
+++ b/packages/datadog-plugin-mysql/test/index.spec.js
@@ -460,7 +460,7 @@ describe('Plugin', () => {
         it('query text should contain traceparent', done => {
           let queryText = ''
           agent.use(traces => {
-            const expectedTimePrefix = Math.floor(clock.now / 1000).toString(16).padStart(8, '0').padEnd(16, '0')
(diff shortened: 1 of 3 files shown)


BridgeAR (changes requested):
This now keeps unused variables around (the clock was only necessary for this call).
Instead of using the ID, we can likely also fix this by just moving the `const clock = sinon.useFakeTimers(new Date())` part to the first line before we call `agent.use()`. That way the time is mocked right away.

rochdev:
> Instead of using the ID, we can likely also fix this by just moving the const clock = sinon.useFakeTimers(new Date()) part to the first line before we call agent.use(). That way the time is mocked right away.

I did tried that and it was still failing, not sure why but that seems unreliable. And to be honest, the real expectation is that the trace ID matches what's in the trace, so the test shouldn't make this kind of assumption to begin with.

BridgeAR (approved):

PR #5444, opened 2025-03-19
chore(llmobs): add span.finished telemetry metric

Yun-Kim:
### What does this PR do?

Adds the `span.finished` telemetry count metric to LLMObs spans before submitting to Datadog, in the `LLMObsSpanProcessor.process()` method. Also adds the following span store key-value pairs:
- `_ml_obs.integration` to keep track of which integration a span originates from.
- `_ml_obs.decorator` to keep track whether a span is created from a decorator.

### Motivation

Verified telemetry metrics are being submitted using a sample nodejs app.

Screenshot 2025-03-20 at 1 08 02 PM

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/dd-trace/src/llmobs/plugins/base.js
```diff
@@ -43,7 +43,7 @@ class LLMObsPlugin extends TracingPlugin {
       llmobsStorage.enterWith({ span })
       ctx.llmobs.parent = parent
 
-      this._tagger.registerLLMObsSpan(span, { parent, ...registerOptions })
+      this._tagger.registerLLMObsSpan(span, { parent, integration: this.constructor.id, ...registerOptions })
```
sabrenner, line 46: we'll also want to add this [here](https://github.com/DataDog/dd-trace-js/blob/2c5e17c2808bac1ec89612c4fd11893d21fb9ce1/packages/dd-trace/src/llmobs/plugins/bedrockruntime.js#L53-L59) - we had to special case bedrock bc the way its instrumented doesn't use the nice tracing channel events (start, end, asyncEnd, etc.) but instead custom event names. should just be able to hardcode `bedrock` there (or feel free to set it up as a constant!)

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -1,12 +1,48 @@
...
 
 function incrementLLMObsSpanStartCount (tags, value = 1) {
   llmobsMetrics.count('span.start', tags).inc(value)
 }
 
+function incrementLLMObsSpanFinishedCount (span, value = 1) {
+  const mlObsTags = LLMObsTagger.tagMap.get(span)
+
+  const isRootSpan = mlObsTags[PARENT_ID_KEY] !== ROOT_PARENT_ID
+  const hasSessionId = mlObsTags[SESSION_ID] !== null
```
sabrenner, line 25:
```suggestion
  const hasSessionId = mlObsTags[SESSION_ID] != null
```
this will match for `undefined` and `null`, where as strict equal `!==` will only pass for `null`, but not for `undefined` (gotta love js)

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -1,12 +1,48 @@
...
+  const integration = mlObsTags[INTEGRATION]
+  const autoInstrumented = integration !== null
```
sabrenner, line 27:
```suggestion
  const autoInstrumented = integration != null
```

packages/dd-trace/src/llmobs/span_processor.js
```diff
@@ -48,6 +50,7 @@ class LLMObsSpanProcessor {
 
     try {
       const formattedEvent = this.format(span)
+      telemetry.incrementLLMObsSpanFinishedCount(span)
```
sabrenner, line 53: it's possible we'll increment this count, but the appending could fail if for some reason there's a json-serializable issue when calling `_writer.append` (as that will `JSON.stringify` and get the byte length, as noted in the comment below it's rare but i don't wanna rule it out). are we OK with incrementing this metric if the appending fails, or do we only want to increment after we have successfully appended?

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -1,12 +1,48 @@
 'use strict'
 
+const {
+  SPAN_KIND,
+  MODEL_PROVIDER,
+  PARENT_ID_KEY,
+  SESSION_ID,
+  INTEGRATION
```
sabrenner, line 8:
```suggestion
  INTEGRATION,
  ROOT_PARENT_ID
```

packages/dd-trace/src/llmobs/span_processor.js
Yun-Kim: Yup, we're fine to increment it here, we'll have separate metrics for spans submitted and payload errors so we won't miss any gaps here

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -1,12 +1,54 @@
...
 const llmobsMetrics = telemetryMetrics.manager.namespace('mlobs')
 
 function incrementLLMObsSpanStartCount (tags, value = 1) {
   llmobsMetrics.count('span.start', tags).inc(value)
 }
 
+function incrementLLMObsSpanFinishedCount (span, value = 1) {
+  const mlObsTags = LLMObsTagger.tagMap.get(span)
+
+  const isRootSpan = mlObsTags[PARENT_ID_KEY] === ROOT_PARENT_ID
```
sabrenner, line 26: saw you changed this in the last commit from `!==` to `===`

just confirming we'll need to change on the python side too?
```python
is_root_span = span._get_ctx_item(PARENT_ID_KEY) != ROOT_PARENT_ID
```

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -1,12 +1,54 @@
...
+  const hasSessionId = mlObsTags[SESSION_ID] != null
+  const integration = mlObsTags[INTEGRATION]
+  const autoInstrumented = integration != null
+  const decorator = mlObsTags[DECORATOR] != null
```
sabrenner, line 30: nit (no action needed): i think this can be a straight `!mlObsTags[DECORATOR]`, since checking boolean values in this case we'll want to have `null/undefined` and `false` evaluate the same here.

like how we check
```python
decorator = span._get_ctx_item(DECORATOR) is True
```
instead of checking `is not None`

It would matter only if we set `mlObsTags[DECORATOR] = false` somewhere, as `false != null` evaluates to `true`. but since we either set `decorator` to `true`, or not at all (ie `null`/`undefined`), the check is fine - just more a note for consistency!

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -1,12 +1,54 @@
...
+  const spanKind = mlObsTags[SPAN_KIND]
+  const modelProvider = mlObsTags[MODEL_PROVIDER]
+
+  const tags = {
+    autoinstrumented: Number(autoInstrumented),
+    has_session_id: Number(hasSessionId),
+    is_root_span: Number(isRootSpan),
+    span_kind: spanKind,
+    integration: integration || 'N/A'
```
sabrenner, line 39: do we want the error tag here too?
```python
("error", str(span.error)),
```

sabrenner (approved):
just a couple clarifying questions but otherwise lgtm, nice job!!

packages/dd-trace/src/llmobs/telemetry.js
Yun-Kim: ugh i didn't realize this mistake was also in the python side. Yeah the root span check needs to verify that the parent ID is set to `undefined`, I'll make the corresponding fix on the python side too

packages/dd-trace/src/llmobs/telemetry.js
Yun-Kim: Great, will make the change. Thanks for the suggestion!

packages/dd-trace/src/llmobs/telemetry.js
Yun-Kim: ahh great catch!

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -1,12 +1,59 @@
...
 
+function incrementLLMObsSpanFinishedCount (span, value = 1) {
+  const mlObsTags = LLMObsTagger.tagMap.get(span)
+  const spanTags = span.context()._tags
+
+  const isRootSpan = mlObsTags[PARENT_ID_KEY] === ROOT_PARENT_ID
+  const hasSessionId = mlObsTags[SESSION_ID] != null
+  const integration = mlObsTags[INTEGRATION]
+  const autoInstrumented = integration != null
+  const decorator = !mlObsTags[DECORATOR]
```
sabrenner, line 33: this is totally my bad but I think it should be this. or we can do `=== true`. apologies! i think i was misreading a bit before between looking at this and python. (`!!` coalesces it into a boolean)
```suggestion
  const decorator = !!mlObsTags[DECORATOR]
```

sabrenner (dismissed):

sabrenner (approved):

PR #5445, opened 2025-03-19
fix amqplib flaky dsm tests

rochdev:
### What does this PR do?

Fix `amqplib` flaky DSM tests by moving each test to its own queue thus making sure they are isolated and deterministic.

### Motivation

Those tests were very brittle, with consumer tests relying on messages from previous producer tests to work, and messages piling up in the queue, all of which was causing significant flakiness.


wconti27 (approved):
The changes look good to me, thanks for taking this on.

PR #5446, opened 2025-03-19
fix aws-sdk kinesis flaky dsm test

rochdev:
### What does this PR do?

Fix `aws-sdk` kinesis flaky DSM test. I also removed a redundant call to `nowStub.restore()` which is already done in an `afterEach` block.

### Motivation

This test has been flaky for a while. The only difference with the other tests is that it mocks `Date.now()`, so my theory is that this impacts `aws-sdk` somehow and causes it to fail intermittently. Since we don't really care about the result from the server, I simply removed the error check which should solve the issue assuming my theory is correct.


wconti27 (approved):
LGTM, thanks for the fix!

PR #5447, opened 2025-03-19
fix wrong test error message when expected span was not received

rochdev:
### What does this PR do?

Fix wrong test error message when expected span was not received.

### Motivation

Now that the official MsgPack library is used for decoding traces in the test agent, some values are `BigInt` instead of numbers which is not supported by `JSON.stringify`. This ends up throwing an error which gets outputted instead of the real underlying error that was thrown by the test.

### DIFF
--- a/packages/dd-trace/test/plugins/helpers.js
+++ b/packages/dd-trace/test/plugins/helpers.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const { AssertionError } = require('assert')
+const { inspect } = require('util')
 const { AsyncResource } = require('../../../datadog-instrumentations/src/helpers/instrument')
 
 const Nomenclature = require('../../src/service-naming')
@@ -15,7 +16,6 @@ function resolveNaming (namingSchema) {
 
 function expectSomeSpan (agent, expected, timeout) {
   return agent.use(traces => {
-    const replacer = (k, v) => typeof v === 'bigint' ? Number(v) : v
     const scoredErrors = []
     for (const trace of traces) {
       for (const span of trace) {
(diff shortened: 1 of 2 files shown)


packages/dd-trace/test/plugins/helpers.js
```diff
@@ -15,6 +15,7 @@ function resolveNaming (namingSchema) {
 
 function expectSomeSpan (agent, expected, timeout) {
   return agent.use(traces => {
+    const replacer = (k, v) => typeof v === 'bigint' ? Number(v) : v
```
simon-id, line 18: loss of precision, if some people debug values they're gonna be confused as to why the error is showing a weird value no ?

packages/dd-trace/test/plugins/helpers.js
```diff
@@ -33,7 +34,7 @@ function expectSomeSpan (agent, expected, timeout) {
     const error = scoredErrors.sort((a, b) => a.score - b.score)[0].err
     // We'll append all the spans to this error message so it's visible in test
     // output.
-    error.message += '\n\nCandidate Traces:\n' + JSON.stringify(traces, null, 2)
+    error.message += '\n\nCandidate Traces:\n' + JSON.stringify(traces, replacer, 2)
```
BridgeAR, line 37: Does the error need to be JSON formatted? If not, lets use `util.inspect()`

```suggestion
    error.message += '\n\nCandidate Traces:\n' + util.inspect(traces)
```

packages/dd-trace/test/plugins/helpers.js
BridgeAR: If it has to be JSON formatted, there is no way around that. We could return a string instead. But that might also be confusing?

BridgeAR (approved):

PR #5450, opened 2025-03-20
add matrix strategy to debugger ci job

rochdev:
### What does this PR do?

Add matrix strategy to debugger CI job.

### Motivation

These are slow and can total over 10 minutes when combined in a single job.


BridgeAR (commented):
The reason for changing it before was that the versions are now going to not be aligned anymore. This was clearly specified with the former implementation.

I guess we could have a variable defined somewhere to specify the matrix in a central spot?

watson (approved):
Thanks for doing this 💯

PR #5458, opened 2025-03-21
Add rasp telemetry metrics

IlyasShabi:
### What does this PR do?
[APPSEC-57045](https://datadoghq.atlassian.net/browse/APPSEC-57045)
- [x] rasp.rule.eval
- [x] rasp.rule.match
- [x] rasp.timeout
- [x] rasp.rule.duration
- [x] rasp.duration
- [x] rasp.duration_ext
- [x] rasp.error
- [x] rasp.rule.skipped

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### Additional Notes

[APPSEC-57045]: https://datadoghq.atlassian.net/browse/APPSEC-57045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dd-trace/src/appsec/rasp/utils.js
```diff
@@ -51,15 +53,18 @@ function handleResult (actions, req, res, abortController, config) {
...
 
       // TODO Delete this when support for node 16 is removed
       if (!abortController.signal.reason) {
         abortController.signal.reason = abortError
       }
+      return
     }
   }
+
+  updateRaspRuleMatchMetricTags(req, raspRule, false, false)
```
simon-id, line 67: don't you also need to call this in line 50 ? `  if (!abortController || abortOnUncaughtException) return
`

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -291,6 +292,7 @@ module.exports = {
   reportMetrics,
   reportAttack,
   reportWafUpdate: incrementWafUpdatesMetric,
+  reportRaspRuleSkipped: updateRaspRuleSkippedMetricTags,
```
simon-id, line 295: not sure this is super useful, you can just call the telemetry function from waf_context_wrapper.js directly no ?

packages/dd-trace/src/appsec/telemetry/rasp.js
```diff
@@ -26,25 +32,89 @@ function addRaspRequestMetrics (store, { duration, durationExt, wafTimeout, erro
...
+  }
 
-  if (metrics.ruleTriggered) {
-    appsecMetrics.count('rasp.rule.match', tags).inc(1)
+  if (raspRule.variant) {
+    tags.rule_variant = raspRule.variant
+  }
+
+  const block = getRuleMatchBlockingStatus(blockTriggered, blocked)
+  appsecMetrics.count('rasp.rule.match', { ...tags, block }).inc(1)
```
simon-id, line 94: why have spread operator here ??? just add `block: getRuleMatchBlockingStatus()` in the `tags` declaration no ?

packages/dd-trace/src/appsec/reporter.js
IlyasShabi: I think it's good to keep calling Reporter from `waf_context_wrapper.js` as an abstraction

packages/dd-trace/src/appsec/rasp/utils.js
```diff
@@ -45,21 +47,25 @@ function handleResult (actions, req, res, abortController, config) {
...
+        abortController.abort(abortError)
 
-      // TODO Delete this when support for node 16 is removed
-      if (!abortController.signal.reason) {
-        abortController.signal.reason = abortError
+        // TODO Delete this when support for node 16 is removed
+        if (!abortController.signal.reason) {
+          abortController.signal.reason = abortError
+        }
+        return
```
simon-id, line 63: the return is not neccesary now that you changed the if

packages/dd-trace/src/appsec/telemetry/rasp.js
```diff
@@ -26,25 +32,89 @@ function addRaspRequestMetrics (store, { duration, durationExt, wafTimeout, erro
...
+    appsecMetrics.count('rasp.error', errorTags).inc(1)
+  }
+
   if (metrics.wafTimeout) {
     appsecMetrics.count('rasp.timeout', tags).inc(1)
   }
+}
+
+function trackRaspRuleMatch (store, raspRule, blockTriggered, blocked) {
+  if (!store[DD_TELEMETRY_REQUEST_METRICS].ruleTriggered) return
```
simon-id, line 81: if we reach this function with ruleTriggered == false, that means something went wrong sure, but i'm not sure that means we should competely ignore the fact we had a rulematch no ?

packages/dd-trace/src/appsec/telemetry/rasp.js
```diff
@@ -26,25 +32,89 @@ function addRaspRequestMetrics (store, { duration, durationExt, wafTimeout, erro
...
+
+  const tags = {
+    waf_version: store[DD_TELEMETRY_REQUEST_METRICS].wafVersion,
+    event_rules_version: store[DD_TELEMETRY_REQUEST_METRICS].rulesVersion,
```
simon-id, line 85: might be worth aliasing `store[DD_TELEMETRY_REQUES_METRICS]` as you use it 3 times and it's long and messy to read while spelled out like this

packages/dd-trace/src/appsec/rasp/utils.js
IlyasShabi: The return is still necessary because we dont want to call updateRaspRuleMatchMetricTags here but after blocking action https://github.com/DataDog/dd-trace-js/blob/3bb5c154b15b7a44b2d51fafebb6112bbde7e939/packages/dd-trace/src/appsec/rasp/index.js#L91

packages/dd-trace/src/appsec/telemetry/rasp.js
IlyasShabi: According to the RFC and  previous implementation `rasp.rule.match` metric should be reporter only if a rule was triggered.
Reaching this function with `ruleTriggered == false` doesn't mean that something went wrong

packages/dd-trace/src/appsec/telemetry/rasp.js
IlyasShabi: Done

packages/dd-trace/test/appsec/telemetry/rasp.spec.js
```diff
@@ -50,26 +65,105 @@ describe('Appsec Rasp Telemetry metrics', () => {
...
+          waf_version: '1.0.0',
+          event_rules_version: '2.0.0'
+        })
+        expect(track).to.have.been.calledWith(42)
+
+        expect(distribution).to.have.been.calledWith('rasp.duration', {
+          waf_version: '1.0.0',
+          event_rules_version: '2.0.0'
+        })
+        expect(track).to.have.been.calledWith(42)
```
simon-id, line 105: the second `expect(track).to.have.been.calledWith(42)` can never fail. Because it's just checking if it was ever called with 42, which is already tested by the first assert. I don't think that's what you're trying to do here

packages/dd-trace/test/appsec/telemetry/rasp.spec.js
```diff
@@ -115,6 +209,85 @@ describe('Appsec Rasp Telemetry metrics', () => {
...
+          block: 'irrelevant'
+        })
+        expect(inc).to.have.been.called
+      })
+
+      it('should not increment any metric if ruleTriggered is not set', () => {
+        const newReq = {}
+        appsecTelemetry.updateRaspRuleMatchMetricTags(newReq, raspRule, true, true)
+
+        expect(count).to.not.have.been.calledWith('rasp.rule.match')
```
simon-id, line 267:
```suggestion
        expect(count).to.not.have.been.called
```

packages/dd-trace/test/appsec/rasp/utils.spec.js
```diff
@@ -3,20 +3,33 @@
...
+      updateRaspRuleMatchMetricTags: sinon.stub()
     }
 
     utils = proxyquire('../../../src/appsec/rasp/utils', {
       '../../plugins/util/web': web,
-      '../stack_trace': stackTrace
+      '../stack_trace': stackTrace,
+      '../telemetry': telemetry,
+      '../blocking': {
+        getBlockingAction: sinon.stub().callsFake((actions) => actions?.blocking_action)
```
simon-id, line 31:
```suggestion
        getBlockingAction: sinon.spy((actions) => actions?.blocking_action)
```

packages/dd-trace/test/appsec/rasp/command_injection.spec.js
simon-id, line 1: weird that command injection tests had to be updated but not all the other rasp types ?

simon-id:
otherwise LGTM

packages/dd-trace/test/appsec/telemetry/rasp.spec.js
```diff
@@ -115,6 +209,85 @@ describe('Appsec Rasp Telemetry metrics', () => {
...
+      it('should not increment any metric if req is not provided', () => {
+        appsecTelemetry.updateRaspRuleMatchMetricTags(null, raspRule, true, true)
+
+        expect(count).to.not.have.been.called
+        expect(inc).to.not.have.been.called
+      })
+    })
+
+    describe('updateRaspRuleSkippedMetricTags', () => {
+      it('should increment rasp.rule.skipped with reason', () => {
```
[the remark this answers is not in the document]
IlyasShabi: I added tests with `rule_variant`

packages/dd-trace/test/appsec/rasp/command_injection.spec.js
IlyasShabi: Because only `command_injection.spec.js` is checking `utils.handleResult`

packages/dd-trace/test/appsec/telemetry/rasp.spec.js
IlyasShabi: The idea of this test is to make sure that each time we call `updateRaspRequestsMetricTags`, it tracks `rasp.rule.duration` with the current duration, and `rasp.duration` with the total durations, it's just part of making sure both calls are correctly tracked.
You can check the diff in `updateRaspRequestsMetricTags` second call

packages/dd-trace/test/appsec/rasp/command_injection.spec.js
simon-id: oh well, that's sad, but not your fault

simon-id (approved):

PR #5461, opened 2025-03-21
chore(langchain): fix test flakiness

sabrenner:
### What does this PR do?
Removes problematic LangChain test causing flakiness, likely caused by the need to reload the tracer fresh to read different config values from the environment. The actual behavior of these tests are asserted [here](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/test/plugins/util/llm.spec.js).

### Motivation
Fix test flakiness.

### DIFF
--- a/packages/datadog-plugin-langchain/test/index.spec.js
+++ b/packages/datadog-plugin-langchain/test/index.spec.js
@@ -40,9 +40,11 @@ describe('Plugin', function () {
         return agent.load('langchain')
       })
 
-      afterEach(() => {
+      afterEach(async () => {
+        console.log('afterEach agent.close')
         // wiping in order to read new env vars for the config each time
-        return agent.close({ ritmReset: false, wipe: true })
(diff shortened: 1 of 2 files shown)


packages/datadog-plugin-langchain/test/index.spec.js
```diff
@@ -16,7 +16,7 @@ const openAiBaseCompletionInfo = { base: 'https://api.openai.com', path: '/v1/co
 const openAiBaseChatInfo = { base: 'https://api.openai.com', path: '/v1/chat/completions' }
 const openAiBaseEmbeddingInfo = { base: 'https://api.openai.com', path: '/v1/embeddings' }
 
-describe('Plugin', () => {
+describe('Plugin', function () {
```
sabrenner, line 19:
```suggestion
describe('Plugin', () => {
```

rochdev (approved):

PR #5462, opened 2025-03-21
add sdk.event to communicate a login or custom event

IlyasShabi:
### What does this PR do?
Count telemetry metric when the application is using the SDK to communicate a login or custom event [APPSEC-57046](https://datadoghq.atlassian.net/browse/APPSEC-57046)

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### Additional Notes

[APPSEC-57046]: https://datadoghq.atlassian.net/browse/APPSEC-57046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/src/appsec/sdk/track_event.js
+++ b/packages/dd-trace/src/appsec/sdk/track_event.js
@@ -16,6 +16,8 @@ function trackUserLoginSuccessEvent (tracer, user, metadata) {
     return
   }
 
+  incrementSdkEventMetric('login_success')
+
   const rootSpan = getRootSpan(tracer)
   if (!rootSpan) {
     log.warn('[ASM] Root span not available in trackUserLoginSuccessEvent')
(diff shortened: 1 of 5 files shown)


packages/dd-trace/src/appsec/sdk/track_event.js
```diff
@@ -15,6 +16,8 @@ function trackUserLoginSuccessEvent (tracer, user, metadata) {
     return
   }
 
+  incrementSdkEventMetric('login_success')
```
simon-id, line 19: so this will run even if the function fails because we can't find the rootspan, but it won't run if it's used with wrong arguments. Is that the semantic the backend and RFC expects ? did it precises it should be sent even when it fails ?

simon-id: especially since in the other functions (login failure and custom event), it is called after the whole function is finished

simon-id:
1 question otherwise LGTM

packages/dd-trace/src/appsec/sdk/track_event.js
IlyasShabi: We could keep it like this as discussed

simon-id (approved):

PR #5466, opened 2025-03-21
feat(llmobs): evaluation metrics can be submitted by agent proxy

sabrenner:
### What does this PR do?
This PR does a couple of things:
1. LLMObs evaluation metrics can be submitted through an agent proxy
2. `submitEvaluation` will not check for a present API key. If it is not set for agentless data submission, it will now be thrown by the module logic instead.
3. Adds a new feature to detect if an agent is running and has the appropriate endpoint to determine `agentlessEnabled` if it is not explicitly set by the user

### Motivation
MLOB-2444


packages/dd-trace/src/llmobs/writers/base.js
```diff
@@ -18,28 +23,36 @@ class BaseLLMObsWriter {
...
+
+    // We only need uncaughtException since:
+    // - unhandledRejection is for Promises and will eventually trigger uncaughtException if unhandled
+    // - uncaughtExceptionMonitor is just for monitoring and doesn't prevent the exception
+    this._uncaughtExceptionHandler = (error) => {
+      this.flush(() => {
+        throw error // Re-throw to preserve Node's default error handling
+      })
+    }
+    process.once('uncaughtException', this._uncaughtExceptionHandler)
```
sabrenner, line 47: this one might be out of scope as the fix might need to be more robust...

packages/dd-trace/src/llmobs/index.js
```diff
@@ -46,7 +52,20 @@ function enable (config) {
...
+      throw new Error(
+        'Cannot send LLM Observability data without a running agent or without both a Datadog API key and site.\n' +
+        'Ensure these configurations are set before running your application.'
+      )
+    }
+
+    evalWriter?.setAgentless(useAgentless)
+    spanWriter?.setAgentless(useAgentless)
+
+    telemetry.recordLLMObsEnabled(startTime, config)
```
sabrenner, line 67: moved this telemetry metric to be inside the agent strategy callback - do we want to define our llmobs startup/enabled metric to include this async op, or just do it at the end of this function synchronously?

packages/dd-trace/src/llmobs/writers/util.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    const endpoints = agentInfo.endpoints
+    const hasEndpoint = Array.isArray(endpoints) && endpoints.some(endpoint => endpoint === EVP_PROXY_AGENT_BASE_PATH)
+    setWritersAgentlessValue(!hasEndpoint)
+  })
+}
+
+function parseResponseAndLog (err, code, eventsLength, url, eventType) {
+  if (code === 403 && err.message.includes('API key is invalid')) {
```
sabrenner, line 36: i've decided to special-case this here to make the message a bit clearer/more actionable, and still send the telemetry event as `request_error`. I have telemetry submission here too but now that i think about it, if the api key and site don't match, and they're likely using agentless telemetry too, i doubt the telemetry event would even make it to our end...

docs/test.ts
```diff
@@ -574,7 +574,7 @@ const enabled = llmobs.enabled
 // manually enable
 llmobs.enable({
   mlApp: 'mlApp',
-  agentlessEnabled: true
+  agentlessEnabled: false
```
Yun-Kim, line 577: this is just for docs? Any reason we're changing this value on a docs snippet?

packages/dd-trace/src/llmobs/constants/writers.js
```diff
@@ -1,12 +1,16 @@
...
-  EVP_PROXY_AGENT_ENDPOINT: 'evp_proxy/v2/api/v2/llmobs',
+  EVP_PROXY_AGENT_BASE_PATH: '/evp_proxy/v2/',
   EVP_SUBDOMAIN_HEADER_NAME: 'X-Datadog-EVP-Subdomain',
-  EVP_SUBDOMAIN_HEADER_VALUE: 'llmobs-intake',
-  AGENTLESS_SPANS_ENDPOINT: '/api/v2/llmobs',
-  AGENTLESS_EVALULATIONS_ENDPOINT: '/api/intake/llm-obs/v1/eval-metric',
+
+  SPANS_EVENT_TYPE: 'span',
+  SPANS_INTAKE: 'llmobs-intake',
+  SPANS_ENDPOINT: '/api/v2/llmobs',
```
Yun-Kim, line 9: lol smart. i'm going to do the same on python side too

packages/dd-trace/src/llmobs/index.js
Yun-Kim: I think we should probably include this async op, just because this is a pretty important part of enabling (writers)

packages/dd-trace/src/llmobs/index.js
```diff
@@ -46,7 +52,20 @@ function enable (config) {
...
+  setAgentStrategy(config, useAgentless => {
+    if (useAgentless && !(config.apiKey && config.site)) {
+      throw new Error(
+        'Cannot send LLM Observability data without a running agent or without both a Datadog API key and site.\n' +
+        'Ensure these configurations are set before running your application.'
+      )
+    }
+
+    evalWriter?.setAgentless(useAgentless)
+    spanWriter?.setAgentless(useAgentless)
```
Yun-Kim, line 65: Any specific reason we are setting the writers to be agentless here instead of just creating new writers once with the correct agentless mode flag?

Yun-Kim (commented):
Looks good for the most part, just some clarifying questions!

docs/test.ts
sabrenner: while this file is just testing our types are valid TS when interpreted, it does actually do/execute everything in this file. i changed it because this call will start llmobs and try to start our writers, so we don't want the file to complain about a missing api key.

packages/dd-trace/src/llmobs/index.js
sabrenner: we want to be able to enqueue data to the writers while the async operation to get the agent info is happening. it's unlikely but possible some data is enqueued between llmobs enabled/starting the operation to get the agent info and then us finally getting it back, and making new writers at that point would essentially erase the buffers they had accumulated.

Yun-Kim (approved):

PR #5468, opened 2025-03-21
chore(llmobs): add span size telemetry metrics

Yun-Kim:
### What does this PR do?

This PR adds telemetry metrics for `span.raw_size` and `span.size`.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -53,7 +61,45 @@ function incrementLLMObsSpanFinishedCount (span, value = 1) {
   llmobsMetrics.count('span.finished', tags).inc(value)
 }
 
+function submitLLMObsRawSpanSize (event, rawEventSize) {
+  const spanKind = event.meta?.spanKind || ''
```
sabrenner, line 65:
```suggestion
  const spanKind = event.meta['span.kind'] || ''
```
this is set here: https://github.com/DataDog/dd-trace-js/blob/3bff8507eb8e6a6c549a620d16dd04cd3ff56129/packages/dd-trace/src/llmobs/span_processor.js#L73
i think the key should always exist, but the value could be undefined if we implement span kind optional type stuff so the fallback is good!

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -53,7 +61,45 @@ function incrementLLMObsSpanFinishedCount (span, value = 1) {
...
+    autoinstrumented: Number(autoinstrumented),
+    error: error ? 1 : 0,
+    integration: integration || 'N/A'
+  }
+
+  llmobsMetrics.distribution('span.raw_size', tags).track(rawEventSize)
+}
+
+function submitLLMObsSpanSize (event, eventSize) {
+  const spanKind = event.meta?.spanKind || ''
```
sabrenner, line 81: ditto

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -53,7 +61,45 @@ function incrementLLMObsSpanFinishedCount (span, value = 1) {
...
+
+  const tags = {
+    span_kind: spanKind,
+    autoinstrumented: Number(autoinstrumented),
+    error: error ? 1 : 0,
+    integration: integration || 'N/A',
+    truncated: Number(truncated)
+  }
+
+  llmobsMetrics.distribution('span.size', tags).track(eventSize)
```
sabrenner, line 97: could be worth extracting some common logic out from these functions, i see we have smth similar in python where we have a shared function for grabbing the tags (which could be helpful if you add more telemetry for span events with base tags). we could also possibly pass in a `truncated` parameter and just have one function that sets the truncated tag and determines which telemetry event name to use. purely stylistic tbh - your call!

packages/dd-trace/src/llmobs/writers/spans/base.js
```diff
@@ -18,11 +19,18 @@ class LLMObsSpanWriter extends BaseWriter {
 
   append (event) {
     const eventSizeBytes = Buffer.from(JSON.stringify(event)).byteLength
+    telemetry.submitLLMObsRawSpanSize(event, eventSizeBytes)
+
+    let processedEventSizeBytes = eventSizeBytes
```
sabrenner, line 24: this made me realize we should actually be passing this `processedEventSizeBytes` event size to the `super.append` call below instead of `eventSizeBytes`, and we probs should've had this re-computation already 😱 idt it's been an issue at all but is something we should fix

 i can make this change in a PR i have for other writers-based changes!

sabrenner (commented):
nice job!! just a style thing + small change and then something you added that made me realize we were potentially passing the wrong byte size to the `super.append` call, but it's something i can fix bc it will have related test changes and is out of scope here i think. otherwise everything lgtm!

packages/dd-trace/src/llmobs/telemetry.js
Yun-Kim: Ahh great suggestion, will simplify this!

packages/dd-trace/src/llmobs/telemetry.js
Yun-Kim: was about to ask why use bracket notation but then realized why lol

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -18,6 +18,27 @@ const LLMObsTagger = require('./tagger')
 
 const llmobsMetrics = telemetryMetrics.manager.namespace('mlobs')
 
+function extractIntegrationFromTags (tags) {
+  if (!tags || !Array.isArray(tags)) return null
```
sabrenner, line 22:
```suggestion
  if (!Array.isArray(tags)) return null
```
nit: `Array.isArray` checks nullish

sabrenner (dismissed):

sabrenner (approved):

PR #5469, opened 2025-03-21
chore(llmobs): add telemetry metrics for dropped span/eval payloads

Yun-Kim:
### What does this PR do?

This PR adds `dropped_{span/eval}_event` telemetry metrics for the LLMObs SDK, providing information regarding how many span and evaluation events are dropped and not submitted to LLM Observability.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -85,9 +85,17 @@ function recordLLMObsSpanSize (event, eventSize, shouldTruncate) {
   llmobsMetrics.distribution('span.size', tags).track(eventSize)
 }
 
+function recordDroppedPayload (numEvents, eventType, error) {
+  if (eventType !== 'span' && eventType !== 'evaluation_metric') return
+  const metricName = eventType === 'span' ? 'dropped_span_event' : 'dropped_eval_event'
+  const tags = { error: error }
```
sabrenner, line 91:
```suggestion
  const tags = { error }
```

sabrenner, line 91:
```suggestion
  const tags = { error }
```

sabrenner (commented):
lgtm! will approve after the lint fix - do we also want this metric on `buffer_full`? https://github.com/DataDog/dd-trace-js/blob/8f744b05ffdea2bc410c07061e7b5d667db0ba60/packages/dd-trace/src/llmobs/writers/base.js#L46-L49

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -102,10 +102,18 @@ function recordLLMObsSpanSize (event, eventSize, shouldTruncate) {
   llmobsMetrics.distribution('span.size', tags).track(eventSize)
 }
 
+function recordDroppedPayload (numEvents, eventType, error) {
+  if (eventType !== 'span' && eventType !== 'evaluation_metric') return
```
sabrenner, line 106:
```suggestion
  if (eventType !== 'span' || eventType !== 'evaluation_metric') return
```
missed this before but i think this is right 😛

Yun-Kim: wait we're trying to return early if the event type is neither `span` or `evaluation_metric`, is && not the correct operator here?

sabrenner (approved):

PR #5470, opened 2025-03-21
chore(llmobs): add telemetry metrics for llmobs.annotate

Yun-Kim:
### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/llmobs/sdk.js
+++ b/packages/dd-trace/src/llmobs/sdk.js
@@ -213,49 +213,60 @@ class LLMObs extends NoopLLMObs {
       span = this._active()
     }
 
-    if (!span) {
-      throw new Error('No span provided and no active LLMObs-generated span found')
-    }
-    if (!options) {
-      throw new Error('No options provided for annotation.')
-    }
+    let err = ''
 
-    if (!LLMObsTagger.tagMap.has(span)) {
-      throw new Error('Span must be an LLMObs-generated span')
-    }
-    if (span._duration !== undefined) {
-      throw new Error('Cannot annotate a finished span')
-    }
-
-    const spanKind = LLMObsTagger.tagMap.get(span)[SPAN_KIND]
-    if (!spanKind) {
-      throw new Error('LLMObs span must have a span kind specified')
-    }
-
-    const { inputData, outputData, metadata, metrics, tags } = options
-
-    if (inputData || outputData) {
-      if (spanKind === 'llm') {
-        this._tagger.tagLLMIO(span, inputData, outputData)
-      } else if (spanKind === 'embedding') {
-        this._tagger.tagEmbeddingIO(span, inputData, outputData)
-      } else if (spanKind === 'retrieval') {
-        this._tagger.tagRetrievalIO(span, inputData, outputData)
-      } else {
-        this._tagger.tagTextIO(span, inputData, outputData)
+    try {
+      if (!span) {
+        err = 'invalid_span_no_active_spans'
+        throw new Error('No span provided and no active LLMObs-generated span found')
       }
-    }
+      if (!options) {
+        err = 'invalid_options'
+        throw new Error('No options provided for annotation.')
+      }
 
-    if (metadata) {
-      this._tagger.tagMetadata(span, metadata)
-    }
+      if (!LLMObsTagger.tagMap.has(span)) {
+        err = 'invalid_span_type'
+        throw new Error('Span must be an LLMObs-generated span')
+      }
+      if (span._duration !== undefined) {
+        err = 'invalid_finished_span'
+        throw new Error('Cannot annotate a finished span')
+      }
 
-    if (metrics) {
-      this._tagger.tagMetrics(span, metrics)
-    }
+      const spanKind = LLMObsTagger.tagMap.get(span)[SPAN_KIND]
+      if (!spanKind) {
+        err = 'invalid_no_span_kind'
+        throw new Error('LLMObs span must have a span kind specified')
+      }
 
-    if (tags) {
-      this._tagger.tagSpanTags(span, tags)
+      const { inputData, outputData, metadata, metrics, tags } = options
+
+      if (inputData || outputData) {
+        if (spanKind === 'llm') {
+          err = this._tagger.tagLLMIO(span, inputData, outputData)
+        } else if (spanKind === 'embedding') {
+          err = this._tagger.tagEmbeddingIO(span, inputData, outputData)
+        } else if (spanKind === 'retrieval') {
+          err = this._tagger.tagRetrievalIO(span, inputData, outputData)
+        } else {
+          err = this._tagger.tagTextIO(span, inputData, outputData)
+        }
+      }
+
+      if (metadata) {
+        this._tagger.tagMetadata(span, metadata)
+      }
+
+      if (metrics) {
+        err = this._tagger.tagMetrics(span, metrics)
+      }
+
+      if (tags) {
+        this._tagger.tagSpanTags(span, tags)
+      }
+    } finally {
+      telemetry.recordLLMObsAnnotate(span, err)
     }
   }
 
--- a/packages/dd-trace/src/llmobs/tagger.js
+++ b/packages/dd-trace/src/llmobs/tagger.js
@@ -86,23 +86,27 @@ class LLMObsTagger {
   // TODO: similarly for the following `tag` methods,
   // how can we transition from a span weakmap to core API functionality
   tagLLMIO (span, inputData, outputData) {
-    this._tagMessages(span, inputData, INPUT_MESSAGES)
-    this._tagMessages(span, outputData, OUTPUT_MESSAGES)
+    errIn = this._tagMessages(span, inputData, INPUT_MESSAGES)
+    errOut = this._tagMessages(span, outputData, OUTPUT_MESSAGES)
+    return errIn || errOut
   }
 
   tagEmbeddingIO (span, inputData, outputData) {
-    this._tagDocuments(span, inputData, INPUT_DOCUMENTS)
-    this._tagText(span, outputData, OUTPUT_VALUE)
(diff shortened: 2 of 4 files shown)


packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -289,36 +307,43 @@ class LLMObs extends NoopLLMObs {
...
     const timestampMs = options.timestampMs || Date.now()
     if (typeof timestampMs !== 'number' || timestampMs < 0) {
+      telemetry.recordSubmitEvaluation(options, 'invalid_timestamp')
       throw new Error('timestampMs must be a non-negative integer. Evaluation metric data will not be sent')
     }
 
     const { label, value, tags } = options
     const metricType = options.metricType?.toLowerCase()
     if (!label) {
+      telemetry.recordSubmitEvaluation(options, 'invalid_metric_label')
```
lievan, line 333: should we also use a try/catch block logic for recording this metric with errors to be consistent with `annotate`?

Same with `export_span`

packages/dd-trace/src/llmobs/index.js
```diff
@@ -94,11 +94,15 @@ function handleLLMObsParentIdInjection ({ carrier }) {
...
+  let err = ''
   try {
     spanWriter.flush()
     evalWriter.flush()
   } catch (e) {
+    err = 'writer_flush_error'
     log.warn(`Failed to flush LLMObs spans and evaluation metrics: ${e.message}`)
+  } finally {
+    telemetry.recordUserFlush(err)
   }
```
sabrenner, line 106:
```suggestion
  } 
  
  telemetry.recordUserFlush(err)
```
shouldn't need the finally here i think since we swallow the error anyways

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -213,74 +213,92 @@ class LLMObs extends NoopLLMObs {
...
-        this._tagger.tagLLMIO(span, inputData, outputData)
-      } else if (spanKind === 'embedding') {
-        this._tagger.tagEmbeddingIO(span, inputData, outputData)
-      } else if (spanKind === 'retrieval') {
-        this._tagger.tagRetrievalIO(span, inputData, outputData)
-      } else {
-        this._tagger.tagTextIO(span, inputData, outputData)
+      if (inputData || outputData) {
+        if (spanKind === 'llm') {
+          err = this._tagger.tagLLMIO(span, inputData, outputData)
```
sabrenner, line 247: have you tried this out in a dummy app that triggers these scenarios (anything calling the tagger)? i think this is actually always guaranteed to throw from the tagger (since this tagger instance hard fails). my proposal:

i'm guessing for now we don't care about tagging/annotating errors from auto-instrumentation, so what we can do is instead of returning an error from these functions, we can catch the error and rethrow it (since the overhead shouldn't matter if we're going to crash their program anyways). then, when throwing the error in the tagger, we can add a property on it with the reason, ie:

```javascript
_handleFailure (msg, errorTag) { // assuming errorTag is easily set by the caller, something like `invalid_embedding_io`
    if (this.softFail) {
      // assuming we won't care about telemetry for misuse in autoinstrumentation/soft fail cases
      log.warn(msg)
    } else {
      const error = new Error(msg)
      // defining it like this shouldn't make it show up in the error stack
      Object.defineProperty(error, 'ddErrorTag', { get () { return errorTag }  })
      throw new Error(msg)
    }
  }
```

then in a new `catch` block in annotate, we can grab the error reason and set it, and re-throw. 

optionally, instead of setting an `err` variable, you could also get rid of the `finally` and call telemetry with the error in the `catch` before re-throwing and if there's no error, call telemetry at the end without setting the error. 

we could also then remove the outer try-catch, add telemetry when we know we're throwing directly from `annotate` (similar to `exportSpan` and `submitEvaluation`), and just `try`/`catch` around the tagger calls.

lmk what you think about this!

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -355,7 +381,7 @@ class LLMObs extends NoopLLMObs {
       timestamp_ms: timestampMs,
       tags: Object.entries(evaluationTags).map(([key, value]) => `${key}:${value}`)
     }
-
+    telemetry.recordSubmitEvaluation(options, '')
```
sabrenner, line 384:
```suggestion
    telemetry.recordSubmitEvaluation(options)
```
doing this in js is valid syntax and will define the second arg as `undefined` 😄

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -109,11 +109,60 @@ function recordDroppedPayload (numEvents, eventType, error) {
   llmobsMetrics.count(metricName, tags).inc(numEvents)
 }
 
+function recordLLMObsAnnotate (span, err, value = 1) {
+  const mlObsTags = LLMObsTagger.tagMap.get(span)
```
sabrenner, line 113:
```suggestion
  const mlObsTags = LLMObsTagger.tagMap.get(span) || {}
```
i think there's a possibility that if the span was not an llmobs span, the tags will be undefined

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -109,11 +109,60 @@ function recordDroppedPayload (numEvents, eventType, error) {
...
+  const spanKind = mlObsTags[SPAN_KIND] || 'N/A'
+  const isRootSpan = mlObsTags[PARENT_ID_KEY] === ROOT_PARENT_ID
+
+  const tags = {
+    error: Number(!err),
```
sabrenner, line 118:
```suggestion
    error: Number(!!err),
```
i think this is what we want as `!''` or `!undefined` --> `true` with the current logic which i don't think is right

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -109,11 +109,60 @@ function recordDroppedPayload (numEvents, eventType, error) {
...
+    span_kind: spanKind,
+    is_root_span: Number(isRootSpan)
+  }
+  if (err) tags.error_type = err
+  llmobsMetrics.count('annotations', tags).inc(value)
+}
+
+function recordUserFlush (err, value = 1) {
+  const tags = { error: Number(!err) }
```
sabrenner, line 127:
```suggestion
  const tags = { error: Number(!!err) }
```

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -109,11 +109,60 @@ function recordDroppedPayload (numEvents, eventType, error) {
...
+  llmobsMetrics.count('user_flushes', tags).inc(value)
+}
+
+function recordExportSpan (span, err, value = 1) {
+  const mlObsTags = LLMObsTagger.tagMap.get(span)
+  const spanKind = mlObsTags[SPAN_KIND] || 'N/A'
+  const isRootSpan = mlObsTags[PARENT_ID_KEY] === ROOT_PARENT_ID
+
+  const tags = {
+    error: Number(!err),
```
sabrenner, line 138:
```suggestion
    error: Number(!!err),
```

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -109,11 +109,60 @@ function recordDroppedPayload (numEvents, eventType, error) {
...
+    span_kind: spanKind,
+    is_root_span: Number(isRootSpan)
+  }
+  if (err) tags.error_type = err
+  llmobsMetrics.count('spans_exported', tags).inc(value)
+}
+
+function recordSubmitEvaluation (options, err, value = 1) {
+  const tags = {
+    error: Number(!err),
```
sabrenner, line 148:
```suggestion
    error: Number(!!err),
```

sabrenner (commented):
mostly lgtm!! added a proposal for a different approach for doing the annotation errors, lmk if it doesn't make sense and sorry it's a bit of a read 😭 otherwise just a couple other suggestions but please lmk if i misread anything!

i also did my best to check the actual error tags themselves and i think they all look right to me. once we have a consensus on the annotate error approach, i'll give each of the scenarios a shot to see if the right telemetry metric is emitted 😄

packages/dd-trace/src/llmobs/sdk.js
Yun-Kim: Yeah, gonna do this. Thanks for the catch!

packages/dd-trace/src/llmobs/sdk.js
sabrenner: nice the change looks great! let me know if the approach doesn't work w local testing. 

also, I think we can just remove the err = from these calls here now (for all IO annotations)

packages/dd-trace/src/llmobs/tagger.js
```diff
@@ -313,11 +313,15 @@ class LLMObsTagger {
...
   // any public-facing LLMObs APIs using this tagger should not soft fail
   // auto-instrumentation should soft fail
-  _handleFailure (msg) {
+  _handleFailure (msg, errorTag) {
     if (this.softFail) {
       log.warn(msg)
     } else {
-      throw new Error(msg)
+      const err = new Error(msg)
+      if (!errorTag) {
```
sabrenner, line 321:
```suggestion
      if (errorTag) {
```
i might be reading it wrong but i think we want to add it if it does exist? lmk if i'm not getting it 😅

packages/dd-trace/src/llmobs/sdk.js
Yun-Kim: Local testing was great (except for uncovering that error metrics are not submitted/flushed in case of unhandled exceptiosn), attached screenshots of metrics (error metrics were flushed by handling exceptions manually).

ncybul (approved):

PR #5471, opened 2025-03-24
Exclude nextjs 12.3.5

IlyasShabi:
https://github.com/vercel/next.js/issues/77461


.github/workflows/appsec.yml
```diff
@@ -205,14 +205,16 @@ jobs:
         version:
           - 18
           - latest
-        range: ['>=10.2.0 <11', '>=11.0.0 <13', '11.1.4', '>=13.0.0 <14', '13.2.0', '>=14.0.0 <=14.2.6', '>=14.2.7 <15', '>=15.0.0']
+        range: ['>=10.2.0 <11', '>=11.0.0 <12.3.5', '11.1.4', '>12.3.5 <13', '>=13.0.0 <14', '13.2.0', '>=14.0.0 <=14.2.6', '>=14.2.7 <15', '>=15.0.0']
```
simon-id, line 208:
```suggestion
        range: ['>=10.2.0 <11', '>=11.0.0 <12.3.5', '11.1.4', '>=12.3.5 <13', '>=13.0.0 <14', '13.2.0', '>=14.0.0 <=14.2.6', '>=14.2.7 <15', '>=15.0.0']
```

packages/datadog-instrumentations/src/next.js
```diff
@@ -214,7 +214,7 @@ addHook({
 
 addHook({
   name: 'next',
-  versions: ['>=11.1'],
+  versions: ['>=11.1 <12.3.5', '>=13'],
```
rochdev, line 217: This will mean that we have a gap of support in the next release.

rochdev (changes requested):

packages/datadog-instrumentations/src/next.js
simon-id: yes as discussed i don't see a way to have the CI pass without doing that. This is just a temporary CI fix and i'm monitoring the issue on nextjs repo to revert that asap

simon-id: you said it yourself, if the tip of the range isn't available, it doesn't work. And if the test ranges are different from the instrumentation ranges, the CI checker complains too

rochdev: We should never drop a version from the instrumentation unless it's completely broken (meaning the instrumented library itself cannot work to begin with), but not if it's fixable, and my understanding is that the broken Next release is fixable by release the missing dependencies, in which case a version would be out that our next release doesn't support.

> yes as discussed i don't see a way to have the CI pass without doing that. This is just a temporary CI fix and i'm monitoring the issue on nextjs repo to revert that asap

A temporary fix should not make the main branch unreleasable, which this would do.

simon-id: I understand, this isn't blocking me, you can close the PR and find a better fix if you want

rochdev:
This PR is no longer needed as a new fixed version of Next has now been released: https://github.com/vercel/next.js/issues/77461#issuecomment-2749675995

PR #5473, opened 2025-03-24
add missing shared config for clientIpHeader

ida613:
### What does this PR do?

`clientIpHeader` was missing from shared config, meaning that users couldn't remap the client IP header to the `http.client_ip` span tag. This PR adds the shared config and adds test cases for the config itself along with test cases within the http plugin to ensure functionality works as intended.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/dd-trace/test/plugin_manager.spec.js
```diff
@@ -338,4 +338,70 @@ describe('Plugin Manager', () => {
       expect(Four.prototype.configure).to.have.been.calledWithMatch({ enabled: false })
     })
   })
+
+  describe('_getSharedConfig', () => {
```
rochdev, line 342: Can a test be added to test the behaviour instead? The fact that we test this with mocked unit tests is why this wasn't caught in the first place.

rochdev (changes requested):

packages/dd-trace/test/plugin_manager.spec.js
rochdev: I meant testing with an actual web server whether the config applies, for example in one of our plugin tests.

wconti27: Done 🏁

packages/datadog-plugin-http/test/server.spec.js
```diff
@@ -255,6 +255,53 @@ describe('Plugin', () => {
...
         })
       })
+
+      describe('with client IP header configuration', () => {
+        beforeEach(() => {
+          return agent.load('http', {
+            client: false,
+            server: {
+              clientIpEnabled: true,
+              clientIpHeader: 'x-forwarded-for'
```
ida613, line 265: can we also add a test with `cliantIpHeader:'X-Forwarded-For'`? `x-forwarded-for` is a part of the [ipHeaderList](https://github.com/DataDog/dd-trace-js/blob/b984e6321bcbe9a9d72e5ced93edd3c0a805b9bf/packages/dd-trace/src/plugins/util/ip_extractor.js#L6), so it will be added without setting `clientIpHeader`

packages/dd-trace/test/plugin_manager.spec.js
rochdev: I would also remove this test at that point as I don't think it bring any more value that the newly added `http` test.

packages/datadog-plugin-http/test/server.spec.js
rochdev: Agreed, we should be testing with a header that doesn't have an existing meaning.

wconti27: I changed the test to use a custom header key name instead of `x-forwarded-for`. In order to ensure that the config is loaded with the tracer config -- using `agent.load(integration, pluginConfig, tracerConfig)` -- I moved the tests to a lesser used server plugin `koa`, since we need to configure and load the necessary tracer config immediately within the test suite (since the tracer and its config cannot be reinitialized during tests AFAIK)

packages/datadog-plugin-koa/test/index.spec.js
```diff
@@ -216,6 +225,54 @@ describe('Plugin', () => {
...
+
+            axios.get(`http://localhost:${port}/user`, {
+              headers: {
+                'x-custom-client-ip-header': '8.8.8.8'
+              }
+            }).catch(done)
+          })
+        })
+
+        it('should not add client IP tag when header is missing', done => {
```
ida613, line 254: Can you also add a test for the case when your request has a header that is neither in `ipHeaderList` nor set by `clientIpHeader`, and check that the header is not added to the span?

wconti27: done

packages/dd-trace/test/plugin_manager.spec.js
wconti27: done, should the web test be removed as well? And just keep the koa plugin test?

rochdev (approved):

packages/dd-trace/test/plugin_manager.spec.js
rochdev: Nah that one is fine, it's an actual util used elsewhere and not just an internal function.

packages/dd-trace/test/plugins/util/web.spec.js
```diff
@@ -206,6 +206,39 @@ describe('plugins/util/web', () => {
         })
       })
 
+      it('should add custom client ip tag to the span when it is configured', () => {
+        req.headers['X-Forwad-For'] = '8.8.8.8'
```
simon-id, line 210: is the misspelling intended here ?

packages/datadog-plugin-koa/test/index.spec.js
simon-id, line 1: i'm not sure why you're changing the koa plugin tests ? ipHeader is not a koa or web framework thing. It's used directly in the web.js file, so the http plugin. Why is it tested here ?

rochdev: Looks like the PR was merged despite your comment. I agree it probably makes more sense to have an `http` test for this, but since all web frameworks inherit the `http` span anyway I guess any integration works. Since it's now merged, I personally am not too worried, but if you really think it should be moved to `http` that can be done as a follow-up PR.

simon-id: It's alright. It would have just been better to have a comment to explain this to the future maintainer of this file

PR #5475, opened 2025-03-24
chore(llmobs): Add llmobs product enabled telemetry metrics

Yun-Kim:
### What does this PR do?

This PR adds two telemetry metrics, `product_enabled` and `init_time` to provide the LLMObs team visibility into how often (and how long) the LLMObs SDK enablement happens, with useful tags such as `agentless` (if agentless mode) and `auto` (if enabled via env vars or inline code setup).

Note that the `auto` tag is based on whether `config._env["llmobs.enabled"]` is set, which refers to the `DD_LLMOBS_ENABLED` env var. This isn't a 100% perfect indicator that env var setup is happening, as a user could technically have those env vars set but still initialize the tracer (and therefore llmobs) manually in code. I've chosen to interpret this case as a form of auto enablement as auto enablement via env vars takes precedence over manual config options, and also because it's pretty difficult to otherwise determine if auto enablement occurred.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -53,7 +53,25 @@ function incrementLLMObsSpanFinishedCount (span, value = 1) {
   llmobsMetrics.count('span.finished', tags).inc(value)
 }
 
+function recordLLMObsEnabled (startTime, config, value = 1) {
+  const initTimeMs = performance.now() - startTime
+  // There isn't an easy way to determine if a user automatically enabled LLMObs via
+  // in-code or command line setup. We'll use the presence of DD_LLMOBS_ENABLED env var
+  // as a rough heuristic, but note that this isn't perfect since
+  // a user may have env vars but enable manually in code.
+  const autoEnabled = Number(!!config._env['llmobs.enabled'])
```
sabrenner, line 62:
```suggestion
  const autoEnabled = Number(!!config._env?.['llmobs.enabled'])
```
i think this is why tests are failing, bc a lot of tests just use a small dummy config option. actually something i can/should do is make a fixture for it, and can do that as part of my refactor. i think for now doing the `?.` here is good enough for tests, and shouldn't be needed in actual environments using llmobs.

the real change is better tests, but def out of scope here so this should do it 🙏

sabrenner (approved):

PR #5476, opened 2025-03-25
fix memory leak in runtime.node.heap.* metrics

### LINKED ISSUE #5394: [BUG]: Datadog 5.41.0 causing high memory consumption
### Tracer Version(s)

5.41.0

### Node.js Version(s)

22 -> [docker image sha256:9459e243f620fff19380c51497493e91db1454f5a30847efe5bc5e50920748d5](https://hub.docker.com/layers/library/node/22.0.0-alpine/images/sha256-3a4802e64ab5181c7870d6ddd8c824c2efc42873baae37d1971451668659483b?context=explore)

### Bug Report

We've been observing high memory usage from Datadog since Saturday, Mar 8.
The stack runs on Next14 and Node 22 from the docker image mentioned above.

Please look at the following Datadog graphs from before and after 8th march.

9th March
See the timing for "shimmer.js" and anonymousL#57

Image

After pinning version to 5.39.0
See the absence of any significant block from datadog.js

Image

### Reproduction Code

_No response_

### Error Logs

_No response_

### Tracer Config

_No response_

### Operating System

_No response_

### Bundling

Next.js

### LINKED ISSUE #5405: [BUG]:
### Tracer Version(s)

5.41.1

### Node.js Version(s)

22.14

### Bug Report

Upgrading 5.41.1 to broke a number of our metrics. We're having a hard time tracking down exactly what happened, but something in 5.41.1 definitely caused some of our counter metrics to increase by an order of magnitude. This PR looks a little sus to me, but I'm not sure.

In the chart below, we deployed a PR where the only substantive change was upgrading from 5.41.0 to 5.41.1. 

We're currently downgrading the version to confirm this is the cause.

Image

### Reproduction Code

This is our code to increment a metric.

```typescript
export type MetricTags = Partial<Record<MetricTag, string | number>>;

export function increment(
  namespace: MetricNamespace,
  metric: MetricName,
  amount: number,
  tags?: MetricTags, // Accept string | undefined to make it easier for the client
) {
  ddTracer.dogstatsd.increment(getStat(namespace, metric), amount, tags);
}


```

### Error Logs

_No response_

### Tracer Config

_No response_

### Operating System

_No response_

### Bundling

Unsure

rochdev:
### What does this PR do?

Fix memory leak in `runtime.node.heap.*` metrics.

### Motivation

These metrics were still reported using the syntax of the previous code where an array of tags would be passed, but the new code expects only a single tag. Since the tag name is then used as a key to a map to store the value for each tag, this would end up adding new entries each time since unlike strings 2 identical arrays don't share reference. This would cause an exponential increase in the metrics reported over time, thus skewing the results and using more and more CPU and memory over time.

Fixes #5394

### DIFF
--- a/packages/dd-trace/test/runtime_metrics.spec.js
+++ b/packages/dd-trace/test/runtime_metrics.spec.js
@@ -262,6 +262,32 @@ suiteDescribe('runtimeMetrics', () => {
         }
       }))
     })
+
+    it('should collect individual metrics only once every 10 seconds', (done) => {
+      runtimeMetrics.stop()
+      runtimeMetrics.start(config)
+
+      global.gc()
+
+      setImmediate(() => setImmediate(() => { // Wait for GC observer to trigger.
+        clock.tick(60 * 60 * 1000)
+
+        try {
+          // If a metric is leaking, it will leak expontentially because it will
+          // be sent one more time each flush, in addition to the previous
+          // flushes that also had the metric multiple times in them, so after
+          // 1 hour even if a single metric is leaking it would get over
+          // 64980 calls on its own without any other metric. A slightly lower
+          // value is used here to be on the safer side.
+          expect(client.gauge.callCount).to.be.lt(60000)
+          expect(client.increment.callCount).to.be.lt(60000)
+
+          done()
+        } catch (e) {
+          done(e)
+        }
+      }))
+    })
   })
 
   describe('when started', () => {
--- a/packages/dd-trace/test/runtime_metrics.spec.js
+++ b/packages/dd-trace/test/runtime_metrics.spec.js
@@ -146,7 +146,7 @@ suiteDescribe('runtimeMetrics', () => {
       }
     }
 
-    setImmediate = globalThis.setImmediate
+    setImmediate = require('timers/promises').setImmediate
     clock = sinon.useFakeTimers()
 
     runtimeMetrics.start(config)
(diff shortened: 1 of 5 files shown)


packages/dd-trace/test/runtime_metrics.spec.js
```diff
@@ -262,6 +262,32 @@ suiteDescribe('runtimeMetrics', () => {
...
+          // 64980 calls on its own without any other metric. A slightly lower
+          // value is used here to be on the safer side.
+          expect(client.gauge.callCount).to.be.lt(60000)
+          expect(client.increment.callCount).to.be.lt(60000)
+
+          done()
+        } catch (e) {
+          done(e)
+        }
+      }))
```
BridgeAR, line 289: Nit, I would like to prevent usage of callbacks in our tests as much as possible. They are error prone (can sometimes lead to timeouts instead of early errors) and it's more difficult to read these.

Using the promise based `setImmediate()` will allow this: https://nodejs.org/api/timers.html#timerspromisessetimmediatevalue-options

```suggestion
      // Wait for GC observer to trigger.
      await setImmediate()
      await setImmediate()
      clock.tick(60 * 60 * 1000)

      // If a metric is leaking, it will leak expontentially because it will
      // be sent one more time each flush, in addition to the previous
      // flushes that also had the metric multiple times in them, so after
      // 1 hour even if a single metric is leaking it would get over
      // 64980 calls on its own without any other metric. A slightly lower
      // value is used here to be on the safer side.
      expect(client.gauge.callCount).to.be.lt(60000)
      expect(client.increment.callCount).to.be.lt(60000)
```

BridgeAR (approved):
LGTM. Thanks for fixing this!

Just left a suggestion that would be nice to include but that does not block the PR.

packages/dd-trace/test/runtime_metrics.spec.js
rochdev: Switched to promise-based timers 👍

BridgeAR (approved):

BridgeAR:
Thank you! :)

BridgeAR:
The benchmark does seem to be a bit flaky, no?

PR #5480, opened 2025-03-25
Fix logging null when debugging and sending data without any error (recreates #5396)

simon-id:
Recreates #5396 
Make sure an error exists before logging it an error message


BridgeAR (approved):

BridgeAR:
@simon-id I saw too late that you didn't cherry-pick the commit. So the attribution for the change is somewhat off.
Please always cherry pick for that :)

simon-id:
I've used the `Co-authored-by:` trailer so it still count equally in github crediting system

PR #5484, opened 2025-03-25
feat: monitor io-valkey

Nico385412:
### What does this PR do?
This PR will make datadog able to trace valkey

### Motivation
It's about amost year now that redis change his licence and valkey was born right after this.

Since then, time has passed and we start having some mismatch between the two and some commands in redis are not compatible anymore (or not yet) with valkey
For exemple HEXPIRE
- Implemented in ([Redis 7.4.0](https://redis.io/docs/latest/commands/hexpire/))
- Still discussed in [Valkey Issue](https://github.com/valkey-io/valkey/issues/640)

For a while, ioredis was compatible with redis but in the future ... meh

So i would like to instrument the official lib for valkey and help the community.

### Plugin Checklist

- [x] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/datadog-plugin-iovalkey/test/naming.js
+++ b/packages/datadog-plugin-iovalkey/test/naming.js
@@ -1,0 +1,19 @@
+const { resolveNaming } = require('../../dd-trace/test/plugins/helpers')
+
+const rawExpectedSchema = {
+  outbound: {
+    v0: {
+      opName: 'redis.command',
+      serviceName: 'test-redis'
+    },
+    v1: {
+      opName: 'redis.command',
+      serviceName: 'test'
+    }
+  }
+}
+
+module.exports = {
+  rawExpectedSchema,
+  expectedSchema: resolveNaming(rawExpectedSchema)
+}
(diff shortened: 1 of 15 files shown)


packages/datadog-plugin-iovalkey/test/naming.js
```diff
@@ -0,0 +1,19 @@
+const { resolveNaming } = require('../../dd-trace/test/plugins/helpers')
+
+const rawExpectedSchema = {
+  outbound: {
+    v0: {
+      opName: 'redis.command',
```
tlhunter, line 6: I need to figure out if it's better to have a `valkey.command` option here. The two services are very similar but I'm not sure what the UI implications are. It seems that long term we'd be better off with `valkey.command` but in the short term using `redis.command` probably "just works".

tlhunter:
Thanks for this PR! I'm going to ask around about the resource names. FYI I'll need to duplicate this PR to "bless" it with CI.

PR #5485, opened 2025-03-25
chore: add tracing macrobenchmark with runtime metrics enabled

zacharycmontoya:
### What does this PR do?
Adds another recurring macrobenchmark job to run tracing with runtime metrics enabled.

### Motivation
We're investigating enabling runtime metrics by default, so this recurring job will help validate that the feature does not incur significant overhead for tracing users.

### DIFF
--- a/.gitlab/macrobenchmarks.yml
+++ b/.gitlab/macrobenchmarks.yml
@@ -65,3 +65,9 @@ only-tracing:
   extends: .macrobenchmarks
   variables:
     DD_BENCHMARKS_CONFIGURATION: only-tracing
+
+only-tracing-with-runtime-metrics-enabled:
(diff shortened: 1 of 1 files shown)


BridgeAR (commented):
The change is LGTM, I just don't yet know how these benchmarks work and why the pipeline failed.

Do you mind explaining it to me? That is likely faster than me having to go through everything on my own :)

zacharycmontoya:
The issue with the CI was apparently a race condition between a job in the `benchmarks` stage and a job in the `benchmarks-pr-comment` stage. After manually re-running the affected benchmark job and then the `check-big-regressions`, this was resolved.

BridgeAR (approved):

PR #5488, opened 2025-03-26
[DI] Add support for probe condition

watson:
### What does this PR do?

Add support for having a condition on a probe.

A condition should return either `true` or `false`. It can however also throw an error. The error can be thrown in two places:
1. When compiling the AST down to the JavaScript source code (an issue was detected at compilation-time)
2. When running the condition in the context of the breakpoint (an issue was detected at run-time)

Compilation-time exceptions are reported as an error event on the probe to the DI diagnostics backend. Run-time exceptions are currently just swallowed and the result of the condition is going to be the same as if it evaluated `false` (i.e. it's not going to tigger the breakpoint).

To be 100% compatible with how conditions work in the other tracers, we should technically record the run-time errors and expose those to the user (as a help in debugging the condition). However, due to how conditions are implemented in this PR (by using the `condition` property on the `Debugger.setBreakpoint` Chrome DevTools Protocol API), it's not possible to know if a condition failed, or just returned `false`.

For now, this is an ok compromise, due to the increased performance gained by using this API.

### Motivation

1. Feature parity with the other tracers
2. Improved performance for probes on hot code-paths that doesn't need to trigger all the time

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/integration-tests/debugger/basic.spec.js
+++ b/integration-tests/debugger/basic.spec.js
@@ -422,6 +422,21 @@ describe('Dynamic Instrumentation', function () {
           when: { json: { eq: [{ getmember: [{ getmember: [{ ref: 'request' }, 'params'] }, 'name'] }, 'invalid'] } }
         }))
       })
+
+      it('should report error if condition is invalid', function (done) {
+        t.agent.on('debugger-diagnostics', ({ payload }) => {
+          payload.forEach(({ debugger: { diagnostics } }) => {
+            if (diagnostics.status === 'ERROR') {
+              assert.strictEqual(diagnostics.exception.message, 'Cannot compile expression: original dsl')
+              done()
+            }
+          })
+        })
+
+        t.agent.addRemoteConfig(t.generateRemoteConfig({
+          when: { dsl: 'original dsl', json: { ref: 'this is not a valid ref' } }
+        }))
+      })
     })
 
     describe('race conditions', function () {
--- a/integration-tests/debugger/basic.spec.js
+++ b/integration-tests/debugger/basic.spec.js
@@ -400,7 +400,7 @@ describe('Dynamic Instrumentation', function () {
         })
 
         t.agent.addRemoteConfig(t.generateRemoteConfig({
-          when: { eq: [{ getmember: [{ getmember: [{ ref: 'request' }, 'params'] }, 'name'] }, 'bar'] }
+          when: { json: { eq: [{ getmember: [{ getmember: [{ ref: 'request' }, 'params'] }, 'name'] }, 'bar'] } }
         }))
       })
 
@@ -419,7 +419,7 @@ describe('Dynamic Instrumentation', function () {
         })
 
         t.agent.addRemoteConfig(t.generateRemoteConfig({
-          when: { eq: [{ getmember: [{ getmember: [{ ref: 'request' }, 'params'] }, 'name'] }, 'invalid'] }
+          when: { json: { eq: [{ getmember: [{ getmember: [{ ref: 'request' }, 'params'] }, 'name'] }, 'invalid'] } }
         }))
       })
     })
--- a/integration-tests/debugger/basic.spec.js
+++ b/integration-tests/debugger/basic.spec.js
@@ -391,6 +391,39 @@ describe('Dynamic Instrumentation', function () {
       })
     })
 
+    describe('condition', function () {
+      beforeEach(t.triggerBreakpoint)
+
+      it('should trigger when condition is met', function (done) {
+        t.agent.on('debugger-input', (x) => {
+          done()
+        })
+
+        t.agent.addRemoteConfig(t.generateRemoteConfig({
+          when: { eq: [{ getmember: [{ getmember: [{ ref: 'request' }, 'params'] }, 'name'] }, 'bar'] }
+        }))
+      })
+
+      it('should not trigger when condition is not met', function (done) {
+        t.agent.on('debugger-diagnostics', ({ payload }) => {
+          payload.forEach((event) => {
+            if (event.debugger.diagnostics.status === 'INSTALLED') {
+              // Can't know if the probe didn't trigger, so just wait a bit and see if the test fails in the mean time
+              setTimeout(done, 2000)
+            }
+          })
+        })
+
+        t.agent.on('debugger-input', () => {
+          assert.fail('Should not trigger when condition is not met')
+        })
+
+        t.agent.addRemoteConfig(t.generateRemoteConfig({
+          when: { eq: [{ getmember: [{ getmember: [{ ref: 'request' }, 'params'] }, 'name'] }, 'invalid'] }
+        }))
+      })
+    })
+
     describe('race conditions', function () {
       it('should remove the last breakpoint completely before trying to add a new one', function (done) {
         const rcConfig2 = t.generateRemoteConfig()
--- a/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
+++ b/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
@@ -2,6 +2,7 @@
 
 const { getGeneratedPosition } = require('./source-maps')
 const session = require('./session')
+const compileCondition = require('./condition')
 const { MAX_SNAPSHOTS_PER_SECOND_PER_PROBE, MAX_NON_SNAPSHOTS_PER_SECOND_PER_PROBE } = require('./defaults')
 const { findScriptFromPartialPath, probes, breakpoints } = require('./state')
 const log = require('../../log')
@@ -47,12 +48,15 @@ async function addBreakpoint (probe) {
     url, lineNumber, columnNumber, probe.id, probe.version
   )
 
+  const condition = probe.when && compileCondition(probe.when)
+
   const { breakpointId } = await session.post('Debugger.setBreakpoint', {
     location: {
       scriptId,
       lineNumber: lineNumber - 1, // Beware! lineNumber is zero-indexed
       columnNumber
-    }
+    },
+    condition
   })
 
   probes.set(probe.id, breakpointId)
--- a/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
+++ b/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
@@ -48,7 +48,7 @@ async function addBreakpoint (probe) {
     url, lineNumber, columnNumber, probe.id, probe.version
   )
 
-  const condition = probe.when && compileCondition(probe.when)
+  const condition = probe.when?.json && compileCondition(probe.when.json)
 
   const { breakpointId } = await session.post('Debugger.setBreakpoint', {
     location: {
--- a/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
+++ b/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
@@ -48,7 +48,12 @@ async function addBreakpoint (probe) {
     url, lineNumber, columnNumber, probe.id, probe.version
   )
 
-  const condition = probe.when?.json && compileCondition(probe.when.json)
+  let condition
+  try {
+    condition = probe.when?.json && compileCondition(probe.when.json)
+  } catch (err) {
+    throw new Error(`Cannot compile expression: ${probe.when.dsl}`, { cause: err })
+  }
 
   const { breakpointId } = await session.post('Debugger.setBreakpoint', {
     location: {
--- a/packages/dd-trace/src/debugger/devtools_client/condition.js
+++ b/packages/dd-trace/src/debugger/devtools_client/condition.js
@@ -88,15 +88,15 @@ function compile (node) {
             return String.prototype.includes.call(obj, elm)
           } else if (Array.isArray(obj)) {
             return Array.prototype.includes.call(obj, elm)
-          } else if (obj instanceof Object.getPrototypeOf(Int8Array)) {
+          } else if (${isTypedArray('obj')}) {
             return Object.getPrototypeOf(Int8Array.prototype).includes.call(obj, elm)
-          } else if (obj instanceof Set) {
+          } else if (${isInstanceOfCoreType('Set', 'obj')}) {
             return Set.prototype.has.call(obj, elm)
-          } else if (obj instanceof WeakSet) {
+          } else if (${isInstanceOfCoreType('WeakSet', 'obj')}) {
             return WeakSet.prototype.has.call(obj, elm)
-          } else if (obj instanceof Map) {
+          } else if (${isInstanceOfCoreType('Map', 'obj')}) {
             return Map.prototype.has.call(obj, elm)
-          } else if (obj instanceof WeakMap) {
+          } else if (${isInstanceOfCoreType('WeakMap', 'obj')}) {
             return WeakMap.prototype.has.call(obj, elm)
           } else {
             throw new TypeError('Variable does not support contains')
@@ -146,22 +146,31 @@ function iterateOn (fnName, variable, callbackCode) {
 }
 
 function isString (variable) {
-  return `typeof ${variable} === 'string' || ${variable} instanceof String`
+  return `(typeof ${variable} === 'string' || ${variable} instanceof String)`
 }
 
 function isIterableCollection (variable) {
-  return `${isArrayOrTypedArray(variable)} || ${variable} instanceof Set || ${variable} instanceof WeakSet`
+  return `(${isArrayOrTypedArray(variable)} || ${isInstanceOfCoreType('Set', variable)} || ` +
+    `${isInstanceOfCoreType('WeakSet', variable)})`
 }
 
 function isArrayOrTypedArray (variable) {
-  return `(Array.isArray(${variable}) || ${variable} instanceof Object.getPrototypeOf(Int8Array))`
+  return `(Array.isArray(${variable}) || ${isTypedArray(variable)})`
+}
+
+function isTypedArray (variable) {
+  return isInstanceOfCoreType('TypedArray', variable, `${variable} instanceof Object.getPrototypeOf(Int8Array)`)
+}
+
+function isInstanceOfCoreType (type, variable, fallback = `${variable} instanceof ${type}`) {
+  return `(process[Symbol.for('datadog:node:util:types')]?.is${type}?.(${variable}) ?? ${fallback})`
 }
 
 function getSize (variable) {
   return `((val) => {
     if (${isString('val')} || ${isArrayOrTypedArray('val')}) {
       return ${guardAgainstPropertyAccessSideEffects('val', '"length"')}
-    } else if (val instanceof Set || val instanceof Map) {
+    } else if (${isInstanceOfCoreType('Set', 'val')} || ${isInstanceOfCoreType('Map', 'val')}) {
       return ${guardAgainstPropertyAccessSideEffects('val', '"size"')}
     } else {
       throw new TypeError('Cannot get length or size of string/collection')
@@ -171,13 +180,13 @@ function getSize (variable) {
 
 function accessProperty (variable, keyOrIndex, allowMapAccess) {
   return `((val, key) => {
-    if (val instanceof Set || val instanceof WeakSet) {
+    if (${isInstanceOfCoreType('Set', 'val')} || ${isInstanceOfCoreType('WeakSet', 'val')}) {
       throw new Error('Accessing a Set or WeakSet is not allowed')
-    } else if (val instanceof Map) {
+    } else if (${isInstanceOfCoreType('Map', 'val')}) {
       ${allowMapAccess
         ? 'return Map.prototype.get.call(val, key)'
         : 'throw new Error(\'Accessing a Map is not allowed\')'}
-    } else if (val instanceof WeakMap) {
+    } else if (${isInstanceOfCoreType('WeakMap', 'val')}) {
       ${allowMapAccess
         ? 'return WeakMap.prototype.get.call(val, key)'
         : 'throw new Error(\'Accessing a WeakMap is not allowed\')'}
@@ -189,7 +198,10 @@ function accessProperty (variable, keyOrIndex, allowMapAccess) {
 
 function guardAgainstPropertyAccessSideEffects (variable, propertyName) {
   return `((val, key) => {
-    if (${maybeProxy('val')} || Object.getOwnPropertyDescriptor(val, key)?.get !== undefined) {
+    if (
+      ${isInstanceOfCoreType('Proxy', 'val', 'true')} ||
+      Object.getOwnPropertyDescriptor(val, key)?.get !== undefined
+    ) {
       throw new Error('Possibility of side effect')
     } else {
       return val[key]
@@ -201,7 +213,7 @@ function guardAgainstCoercionSideEffects (variable) {
   return `((val) => {
     if (
       typeof val === 'object' && val !== null && (
-        ${maybeProxy('val')} ||
+        ${isInstanceOfCoreType('Proxy', 'val', 'true')} ||
         val[Symbol.toPrimitive] !== undefined ||
         val.valueOf !== Object.prototype.valueOf ||
         val.toString !== Object.prototype.toString
@@ -212,10 +224,6 @@ function guardAgainstCoercionSideEffects (variable) {
       return val
     }
   })(${variable})`
-}
-
-function maybeProxy (variable) {
-  return `(process[Symbol.for('datadog:isProxy')]?.(${variable}) ?? true)`
 }
 
 function assertString (variable) {
--- a/packages/dd-trace/src/debugger/devtools_client/condition.js
+++ b/packages/dd-trace/src/debugger/devtools_client/condition.js
@@ -134,30 +134,30 @@ function isArrayOrTypedArray (variable) {
 }
 
 function accessProperty (variable, keyOrIndex, allowMapAccess) {
-  return `(() => {
-    if (${variable} instanceof Set || ${variable} instanceof WeakSet) {
+  return `((val, key) => {
+    if (val instanceof Set || val instanceof WeakSet) {
       throw new Error('Accessing a Set or WeakSet is not allowed')
-    } else if (${variable} instanceof Map || ${variable} instanceof WeakMap) {
+    } else if (val instanceof Map || val instanceof WeakMap) {
       ${allowMapAccess
-        ? `return ${callMethodOnPrototype(variable, 'get', keyOrIndex)}`
+        ? `return ${callMethodOnPrototype('val', 'get', 'key')}`
         : 'throw new Error(\'Accessing a Map or WeakMap is not allowed\')'}
     } else {
-      return ${guardAgainstPropertyAccessSideEffects(variable, keyOrIndex)}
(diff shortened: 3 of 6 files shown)


watson:
* **#5488** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5488?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5488?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,124 @@
...
+  const [type, value] = Object.entries(node)[0]
+
+  if (type === 'not') {
+    return `!(${compile(value)})`
+  } else if (type === 'isEmpty') {
+    return `(${compile(value)}).length === 0`
+  } else if (type === 'isDefined') {
+    return `(() => {
+        try {
+          ${value}
```
shatzi, line 19: should we add `;` or `()` around the value here?

packages/dd-trace/src/debugger/devtools_client/breakpoints.js
```diff
@@ -47,12 +48,15 @@ async function addBreakpoint (probe) {
     url, lineNumber, columnNumber, probe.id, probe.version
   )
 
+  const condition = probe.when && compileCondition(probe.when)
```
shatzi, line 51: what if compileCondition throws? is there a try catch handling of this addBreakpoint?

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,124 @@
...
+}
+
+function iterateOn (fnName, variable, callbackCode) {
+  return `${isCollection(variable)}
+    ? Array.from(${variable}).${fnName}((it) => ${callbackCode})
+    : Object.entries(${variable}).${fnName}(([key, value]) => ${callbackCode})`
+}
+
+function isCollection (variable) {
+  return `Array.isArray(${variable}) || ${variable} instanceof Set || ${variable} instanceof WeakSet`
```
shatzi, line 105: this will cause multiple evaluation of ${variable} -> which could be a whole expression.
I think this ok for now but we might want to address in the future in expression that uses the input variable multiple times.

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,124 @@
...
+          ${value}
+          return true
+        } catch {
+          return false
+        }
+      })()`
+  } else if (type === 'instanceof') {
+    return `${compile(value[0])} instanceof ${value[1]}`
+  } else if (type === 'len' || type === 'count') {
+    return `(${compile(value)}).length`
```
shatzi, line 28: 1. does this also works for Map? Sets? I think we need to call size on those.
2. do we need to use throwOnSideEfffects if length is a getter?

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,124 @@
...
+    return `(() => {
+        try {
+          ${value}
+          return true
+        } catch {
+          return false
+        }
+      })()`
+  } else if (type === 'instanceof') {
+    return `${compile(value[0])} instanceof ${value[1]}`
```
shatzi, line 26: should this be under Array.isArray(value) at line 39 ?

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,124 @@
...
+      case 'ne': return `${args[0]} !== ${args[1]}`
+      case 'gt': return `${args[0]} > ${args[1]}`
+      case 'ge': return `${args[0]} >= ${args[1]}`
+      case 'lt': return `${args[0]} < ${args[1]}`
+      case 'le': return `${args[0]} <= ${args[1]}`
+      case 'any': return iterateOn('some', ...args)
+      case 'all': return iterateOn('every', ...args)
+      case 'and': return `(${args.join(') && (')})`
+      case 'or': return `(${args.join(') || (')})`
+      case 'startsWith': return `${args[0]}.startsWith(${args[1]})`
```
shatzi, line 52: should we put `()` round ${args[0]} here?

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,124 @@
...
+        })()`
+      case 'matches': return `(() => {
+        if (typeof ${args[0]} === 'string') {
+          if (${args[1]} instanceof RegExp) {
+            return ${callMethodOnPrototype(args[1], 'test', args[0])}
+          } else {
+            return ${args[0]}.match(${args[1]}) !== null
+          }
+        } else {
+          throw new TypeError('Variable ${args[0]} is not a string')
```
shatzi, line 76: or Regex

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,124 @@
...
+      return 'key'
+    } else if (value === '@value') {
+      return 'value'
+    } else {
+      return value
+    }
+  } else if (Array.isArray(value)) {
+    const args = value.map(compile)
+    switch (type) {
+      case 'eq': return `${args[0]} === ${args[1]}`
```
shatzi, line 42: we might just need to add `()` round those expression and then avoid adding `()` everywhere when we use the expression.

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,124 @@
...
+          throw new TypeError('Variable ${args[0]} is not a string')
+        }
+      })()`
+      case 'filter': return `${isCollection(args[0])}
+        ? Array.from(${args[0]}).filter((it) => ${args[1]})
+        : Object.entries(${args[0]}).filter(([key, value]) => ${args[1]}).reduce((acc, [key, value]) => {
+          acc[key] = value
+          return acc
+        }, {})`
+      case 'substring': return `${args[0]}.substring(${args[1]}, ${args[2]})`
```
shatzi, line 85: I think this is the right direction here for now. just use the native implementation.

do we need to check that ${args[0]} is a string first? if not we just calling a substring method on an arbitrary object.

packages/dd-trace/test/debugger/devtools_client/condition.spec.js
```diff
@@ -0,0 +1,261 @@
...
+    ['foo', 'bar']
+  ],
+  [
+    { filter: [{ ref: 'collection' }, { not: { isEmpty: { ref: '@it' } } }] },
+    { collection: new Set(['foo', 'bar', '']) },
+    ['foo', 'bar']
+  ],
+  [
+    { filter: [{ ref: 'collection' }, { not: { isEmpty: { ref: '@value' } } }] },
+    { collection: { 1: 'foo', 2: 'bar', 3: '' } },
```
shatzi, line 120: does the spec allow the user to have '@it' iterator over entities? not sure if that something we need to support..

packages/dd-trace/test/debugger/devtools_client/condition.spec.js
```diff
@@ -0,0 +1,261 @@
...
+    new Error('Aborting because accessing a Map or WeakMap is not allowed')
+  ],
+  [
+    { getmember: [{ ref: 'obj' }, 'getter'] },
+    { obj: Object.create(Object.prototype, { getter: { get () { return 'x' } } }) },
+    new Error('Aborting because of possible side effects')
+  ],
+
+  // References with operations
+  [{ len: { ref: 'foo' } }, { foo: 'hello' }, 5],
```
shatzi, line 49: lets add test for 'len' 'size' with getters defined

packages/dd-trace/test/debugger/devtools_client/condition.spec.js
```diff
@@ -0,0 +1,261 @@
...
+  ],
+  [
+    { contains: [{ ref: 'obj' }, 'missing'] },
+    { obj: { foo: 'bar' } },
+    new TypeError('Variable obj does not support contains')
+  ],
+  // TODO: Ensure there's no side-effects due to Symbol.match
+  [{ matches: [{ ref: 'foo' }, '[0-9]+'] }, { foo: '42' }, true],
+  [{ matches: [{ ref: 'foo' }, { ref: 'regex' }] }, { foo: '42', regex: /[0-9]+/ }, true],
+  [{ matches: [{ ref: 'foo' }, { ref: 'regex' }] }, { foo: '42', regex: overloadMethod(/[0-9]+/, 'test') }, true],
```
shatzi, line 166: lets add test when foo is not a string.
and then maybe also overloadMethod of match on foo.

packages/dd-trace/test/debugger/devtools_client/condition.spec.js
```diff
@@ -0,0 +1,261 @@
...
+  [
+    { instanceof: [{ ref: 'bar' }, 'CustomObject'] },
+    { bar: new CustomObject(), CustomObject },
+    true
+  ]
+  // TODO: Ensure there's no side-effects due to proxies
+]
+
+const definedTestCases = [
+  [{ isDefined: 'foo' }, 'const foo = undefined', false],
```
shatzi, line 197: not sure why this is false...

shatzi (approved):
awesome progress

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: The array `length` property cannot be redefined as a function or getter, so we should be safe there. However, maps and sets can have their `size` property replaced with a getter, so we should guard against that as well as supporting those with the `len`/`count` operation. Good call!

packages/dd-trace/test/debugger/devtools_client/condition.spec.js
watson: The 2nd element in the array is code that's added after evaluating the the condition, so it's in in scope yet, hence it's undefined.

packages/dd-trace/test/debugger/devtools_client/condition.spec.js
watson: I've been wondering the same thing. I came to the conclusion that shouldn't be supported based on [this discussion](https://docs.google.com/document/d/1g76oXz9A5NV4y8Do8eSwmUXOKdZxVE61kbQSJGpBZ-E/edit?tab=t.0#bookmark=id.jtx0oaxna0rj) from January, but maybe I misunderstood it.

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: No, I think the error is correct: `args[0]` should always be a string. `args[1]` might be a regex or a string.

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: Isn't this always a reference to a variable? If so, it shouldn't be required, or?

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: Is it because `value[1]` should also be compiled?

packages/dd-trace/test/debugger/devtools_client/condition.spec.js
watson: Done 👍

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: Done 👍

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: Done 👍

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: Done 👍

packages/dd-trace/src/debugger/devtools_client/condition.js
shatzi: I see your point. 99% of the time is will just be a reference. it could be a complex one like map[index].field but those are get translated into a method call with () around it.

packages/dd-trace/src/debugger/devtools_client/condition.js
shatzi: nope. because you access it as an array...

packages/dd-trace/src/debugger/devtools_client/condition.js
shatzi: yup. my bad

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -15,17 +16,40 @@ function compile (node) {
...
+        Array.isArray(val) ||
+        val instanceof Object.getPrototypeOf(Int8Array)
+      ) {
+        return val.length
+      } else if (val instanceof Set || val instanceof Map) {
+        return ${throwOnSideEffect('val', '"size"')}
+      } else {
+        throw new TypeError('Variable does not support len/count')
+      }
+    })(${compile(value)})`
```
shatzi, line 52: nice! love how you solved the compile multiple times.

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -96,29 +130,37 @@ function callMethodOnPrototype (variable, methodName, ...args) {
...
+  })()`
+}
+
+function throwOnSideEffect (variable, propertyName) {
+  return `(() => {
+    if (Object.getOwnPropertyDescriptor(${variable}, ${propertyName})?.get === undefined) {
+      return ${variable}[${propertyName}]
     } else {
-      throw new Error('Aborting because of possible side effects')
+      throw new Error('Posibility of side effect')
```
shatzi, line 163: all those errors should be more helpful for the user to understand the problem.
something like `Possibility of side effect accessing when evaluate ${stringifyExpression(node)}`

However, we might want to address this on future PRs

watson: This only truly becomes relevant when implementing template strings, as any errors thrown in conditions will just treat the condition as "not met" - but the error itself will be swallowed.

I'll see if I can improve it in this PR, or I will defer it till I implement support for template strings.

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: Under the `Array.isArray(value)` if-block below, I also compile the second element of the `value` array. If we don't want that in this case, it shouldn't be part of that if-block

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: Thank you 😊

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: Fixed 👍

packages/dd-trace/test/debugger/devtools_client/condition.spec.js
watson: Consensus from Zoom: We'll keep the current implementation

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  throw new TypeError(`Unknown AST node type: ${type}`)
+}
+
+function iterateOn (fnName, variable, callbackCode) {
+  return `(($dd_val) => {
+    return ${isIterableCollection('$dd_val')}
+      ? Array.from($dd_val).${fnName}(($dd_it) => ${callbackCode})
+      : Object.entries($dd_val).${fnName}(([$dd_key, $dd_value]) => ${callbackCode})
```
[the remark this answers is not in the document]
watson: The `fnName` argument is not user provided input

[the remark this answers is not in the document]
watson: This is the same as the comment above. Resolving to reduce noise

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,199 @@
...
+  })(${variable})`
```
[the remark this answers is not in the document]
watson: The `ref` value is sanitized (see `assertIdentifier`), so I believe this issue has been mitigated.

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,199 @@
...
+    }
+  }
+
+  throw new TypeError(`Unknown AST node type: ${type}`)
+}
+
+function iterateOn (fnName, variable, callbackCode) {
+  return `(($dd_val) => {
+    return ${isIterableCollection('$dd_val')}
+      ? Array.from($dd_val).${fnName}(($dd_it) => ${callbackCode})
```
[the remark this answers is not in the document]
watson: The `ref` value is sanitized (see `assertIdentifier`), so I believe this issue has been mitigated.

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,237 @@
...
+    } else if (value === '@key') {
+      return '$dd_key'
+    } else if (value === '@value') {
+      return '$dd_value'
+    } else {
+      return assertIdentifier(value)
+    }
+  } else if (Array.isArray(value)) {
+    const args = value.map(compile)
+    switch (type) {
```
BridgeAR, line 73: Is this going to be a hot code path?

If yes, we could check for the length of each string to skip many checks in average (length === 2 || length === 3 ...).

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,237 @@
...
+            return Array.prototype.includes.call(obj, elm)
+          } else if (obj instanceof Object.getPrototypeOf(Int8Array)) {
+            return Object.getPrototypeOf(Int8Array.prototype).includes.call(obj, elm)
+          } else if (obj instanceof Set) {
+            return Set.prototype.has.call(obj, elm)
+          } else if (obj instanceof WeakSet) {
+            return WeakSet.prototype.has.call(obj, elm)
+          } else if (obj instanceof Map) {
+            return Map.prototype.has.call(obj, elm)
+          } else if (obj instanceof WeakMap) {
```
BridgeAR, line 99: Instead of using instanceof, should we use `util.types.isWeakMap()` instead? Those work across realms, instanceof can't handle things across realms.

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,237 @@
...
+            throw new TypeError('Variable is not a string')
+          }
+        })(${args[0]}, ${args[1]})`
+      case 'filter': return `(($dd_var) => {
+          return ${isIterableCollection('$dd_var')}
+            ? Array.from($dd_var).filter(($dd_it) => ${args[1]})
+            : Object.entries($dd_var).filter(([$dd_key, $dd_value]) => ${args[1]}).reduce((acc, [k, v]) => {
+                acc[k] = v
+                return acc
+              }, {})
```
BridgeAR, line 124: Nit: the filter could be done in the reduce. That way there is one less intermediate array allocation.

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,237 @@
...
+function iterateOn (fnName, variable, callbackCode) {
+  return `(($dd_val) => {
+    return ${isIterableCollection('$dd_val')}
+      ? Array.from($dd_val).${fnName}(($dd_it) => ${callbackCode})
+      : Object.entries($dd_val).${fnName}(([$dd_key, $dd_value]) => ${callbackCode})
+  })(${variable})`
+}
+
+function isString (variable) {
+  return `typeof ${variable} === 'string' || ${variable} instanceof String`
```
BridgeAR, line 150: Do we have to handle boxed primitives?

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,237 @@
...
+    } else {
+      throw new TypeError('Cannot get length or size of string/collection')
+    }
+  })(${variable})`
+}
+
+function accessProperty (variable, keyOrIndex, allowMapAccess) {
+  return `((val, key) => {
+    if (val instanceof Set || val instanceof WeakSet) {
+      throw new Error('Accessing a Set or WeakSet is not allowed')
```
BridgeAR, line 176: Nit: I would always put these to the end. That way the positive ones are handled faster.

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,237 @@
...
+}
+
+function guardAgainstCoercionSideEffects (variable) {
+  return `((val) => {
+    if (
+      typeof val === 'object' && val !== null && (
+        ${maybeProxy('val')} ||
+        val[Symbol.toPrimitive] !== undefined ||
+        val.valueOf !== Object.prototype.valueOf ||
+        val.toString !== Object.prototype.toString
```
BridgeAR, line 208: Are arrays allowed? These have another `.toString()` property.

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: Hmm... we don't have access to `require` in the scope that this code runs, so we can't easily load `util`. I made an exception for `isProxy` and added it to `process` so that I could access it here, because there's no other way to check if something is a proxy.

1. So we can either decide we don't care about realm support
2. Or we can add the specific methods we need to `process` just like I did with `isProxy`
3. Or we can just go all in and add all of `util.types` there

Are there any cons to just doing no 3 that you can think of?

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: No it's only executed when a new probe is being added, which happens rarely. Maybe on a really active project it happens a few hundred times a day

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,237 @@
...
+function isString (variable) {
+  return `typeof ${variable} === 'string' || ${variable} instanceof String`
+}
+
+function isIterableCollection (variable) {
+  return `${isArrayOrTypedArray(variable)} || ${variable} instanceof Set || ${variable} instanceof WeakSet`
+}
+
+function isArrayOrTypedArray (variable) {
+  return `Array.isArray(${variable}) || ${variable} instanceof Object.getPrototypeOf(Int8Array)`
```
shatzi, line 158:
```suggestion
  return `(Array.isArray(${variable}) || ${variable} instanceof Object.getPrototypeOf(Int8Array))`
```

shatzi (approved):

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: I agree it's an edge case, but you'll be surprised what kind of code people write in the wild. You could of course argue that we just don't support that type of strings, but I guess it's an easy win with this extra `instanceof` check, so I figured, why not 🤷

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,237 @@
...
+  'implements', 'interface', 'package', 'private', 'protected', 'public',
+
+  // Litterals
+  'NaN'
+])
+
+// TODO: Consider storing some of these functions on `process` so they can be reused across probes
+function compile (node) {
+  if (node === null || typeof node === 'number' || typeof node === 'boolean' || typeof node === 'string') {
+    return JSON.stringify(node)
```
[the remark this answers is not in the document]
watson: Not sure I understand your suggestion?

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: No I don't think there would be any reason to compare an array using `<` or `>`.

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: Fixed 👍

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: Based on IRL conversation, I've added `util.types` to the `process` symbol so we have access to all these and I've updated the code accordingly 👍

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -0,0 +1,244 @@
...
+
+  throw new TypeError(`Unknown AST node type: ${type}`)
+}
+
+function iterateOn (fnName, variable, callbackCode) {
+  return `(($dd_val) => {
+    return ${isIterableCollection('$dd_val')}
+      ? Array.from($dd_val).${fnName}(($dd_it) => ${callbackCode})
+      : Object.entries($dd_val).${fnName}(([$dd_key, $dd_value]) => ${callbackCode})
+  })(${variable})`
```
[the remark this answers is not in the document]
watson: The `ref` value is sanitized (see `assertIdentifier`), so I believe this issue has been mitigated.

packages/dd-trace/src/debugger/devtools_client/condition.js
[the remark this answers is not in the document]
BridgeAR: Instead of calling `JSON.stringify(number)` it could be `"${String(number)}"` (in backticks)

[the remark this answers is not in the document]
watson: Conclusion after IRL conversation: This optimization would make the code more verbose / harder to maintain, so let's just leave it as is since this isn't a hot code-path.

integration-tests/debugger/basic.spec.js
```diff
@@ -391,6 +391,54 @@ describe('Dynamic Instrumentation', function () {
...
+          assert.fail('Should not trigger when condition is not met')
+        })
+
+        t.agent.addRemoteConfig(t.generateRemoteConfig({
+          when: { json: { eq: [{ getmember: [{ getmember: [{ ref: 'request' }, 'params'] }, 'name'] }, 'invalid'] } }
+        }))
+      })
+
+      it('should report error if condition is invalid', function (done) {
+        t.agent.on('debugger-diagnostics', ({ payload }) => {
```
BridgeAR, line 427: For future reference: using EventEmitter's `once()` would allow the tests to likely be nicer to read :)

shatzi (approved):

PR #5492, opened 2025-03-27
centralize node version numbers for ci workflows

rochdev:
### What does this PR do?

Centralize Node version numbers for CI workflows.

### Motivation

Previous approach was repeating version numbers explicitly in the workflows, which is more difficult to maintain when a new version is released and the semantics change.


.github/actions/node/oldest-maintenance-lts/action.yml
```diff
@@ -3,6 +3,6 @@ description: Install the oldest Maintenance LTS version of Node.js
 runs:
   using: composite
   steps:
-    - uses: actions/setup-node@v4
```
BridgeAR, line 6: This might break in-between versions. I would keep that as is

PR #5494, opened 2025-03-27
new wasm rewriter w/ orchestrion

bengl:
### What does this PR do?
Adds support for `langchain` via Orchestrion-JS, by way of the new `wasm-js-rewriter`, which replaces the old `native-iast-rewriter`. The new rewriter contains both Orchestrion-JS and IAST rewriting for taint tracking.

### Motivation
Prior to Orchestrion-JS, we were unable to support `langchain` when used in an ESM app.

### Additional Notes

Right now this PR depends on a branch of the rewriter. Once the branch is merged on the rewriter, a cleanup commit will be added that removes things like `yarn link` and git dependencies.


packages/datadog-instrumentations/src/langchain.js
```diff
@@ -46,51 +16,14 @@ for (const extension of extensions) {
...
-    shimmer.wrap(BaseLLM.prototype, 'generate', generate => wrapLangChainPromise(generate, 'llm'))
     return exports
   })
 
   addHook({ name: '@langchain/core', file: `dist/embeddings.${extension}`, versions: ['>=0.1'] }, exports => {
-    // we cannot patch the prototype of the Embeddings class directly
-    // this is because the "abstract class Embeddings" is transpiled from TypeScript to not include abstract functions
-    // thus, we patch the exported class directly instead instead.
-
-    shimmer.wrap(exports, 'Embeddings', Embeddings => {
```
rochdev, line 69: Removing the JS implementation will make this stop working in Node <20.6

packages/dd-trace/src/llmobs/plugins/langchain/index.js
```diff
@@ -20,7 +24,8 @@ const ChatModelHandler = require('./handlers/chat_model')
 const LlmHandler = require('./handlers/llm')
 const EmbeddingHandler = require('./handlers/embedding')
 
-class LangChainLLMObsPlugin extends LLMObsPlugin {
+class BaseLangChainLLMObsPlugin extends LLMObsPlugin {
+  static get integration () { return 'langchain' }
```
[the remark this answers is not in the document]
bengl: @sabrenner Can you answer this?

packages/dd-trace/src/proxy.js
```diff
@@ -247,10 +248,13 @@ class Tracer extends NoopProxy {
       if (config.iast.enabled) {
         this._modules.iast.enable(config, this._tracer)
       }
+      // This needs to be after the IAST module is enabled
+      this._modules.rewriter.enable(config)
```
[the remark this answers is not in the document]
bengl: Done

packages/dd-trace/src/appsec/iast/taint-tracking/filter.js
```diff
@@ -6,10 +6,24 @@ const isPrivateModule = function (file) {
   return file && file.indexOf(NODE_MODULES) === -1
 }
 
-const isNotLibraryFile = function (file) {
+let isNotLibraryFile = function (file) {
   return file && file.indexOf('dd-trace-js') === -1 && file.indexOf('dd-trace') === -1
 }
 
+if (process.env.PLUGINS) {
```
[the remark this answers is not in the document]
bengl: Done

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter-esm.mjs
```diff
@@ -35,15 +40,26 @@ export async function load (url, context, nextLoad) {
   if (!result.source) return result
   if (url.includes(ddTraceDir) || url.includes('iitm=true')) return result
 
+  let passes
   try {
+    if (!isNotLibraryFile(url)) {
```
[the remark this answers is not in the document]
bengl: Done.

packages/dd-trace/src/proxy.js
```diff
@@ -247,10 +248,13 @@ class Tracer extends NoopProxy {
...
     } else if (this._tracingInitialized) {
       this._modules.appsec.disable()
       this._modules.iast.disable()
       this._modules.llmobs.disable()
+      this._modules.rewriter.disable()
```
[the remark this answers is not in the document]
bengl: Done.

packages/datadog-plugin-langchain/src/tracing.js
```diff
@@ -85,4 +92,98 @@ class LangChainTracingPlugin extends TracingPlugin {
...
+  end (ctx) {
+    const { self } = ctx
+    const namespace = ['langchain', 'embeddings']
+
+    if (self.constructor.name === 'OpenAIEmbeddings') {
+      namespace.push('openai')
+    }
+
+    const queryChannel = tracingChannel('apm:@langchain/core:Embeddings_embedQuery')
+    shimmer.wrap(self, 'embedQuery', embedQuery => wrapLangChainPromise(embedQuery, namespace, queryChannel))
```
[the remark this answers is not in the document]
bengl: Done

packages/dd-trace/src/llmobs/plugins/langchain/index.js
```diff
@@ -129,4 +137,101 @@ class LangChainLLMObsPlugin extends LLMObsPlugin {
...
+  end (ctx) {
+    const { self } = ctx
+    const namespace = ['langchain', 'embeddings']
+
+    if (self.constructor.name === 'OpenAIEmbeddings') {
+      namespace.push('openai')
+    }
+
+    const queryChannel = tracingChannel('apm:@langchain/core:Embeddings_embedQuery')
+    shimmer.wrap(self, 'embedQuery', embedQuery => wrapLangChainPromise(embedQuery, namespace, queryChannel))
```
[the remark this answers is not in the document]
bengl: Done.

packages/datadog-instrumentations/src/langchain.js
bengl: Done.

packages/dd-trace/src/llmobs/plugins/langchain/index.js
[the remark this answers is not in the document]
rochdev: I guess my thought was that if we can use `component` then it's 2 birds in 1 stone since it also has a meaning for the plugin base class. Not that it matters much and definitely not a blocker.

packages/dd-trace/src/llmobs/plugins/langchain/index.js
```diff
@@ -129,4 +137,101 @@ class LangChainLLMObsPlugin extends LLMObsPlugin {
   }
 }
 
-module.exports = LangChainLLMObsPlugin
+class RunnableSequenceInvokePlugin extends BaseLangChainLLMObsPlugin {
```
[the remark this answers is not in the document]
rochdev: I guess each line differs just enough and without a pattern to make this difficult. Oh well, I still think we should improve this but it's not a blocker.

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
```diff
@@ -1,72 +1,72 @@
...
 
       const chainSourceMap = isFlagPresent('--enable-source-maps')
-      const getOriginalPathAndLineFromSourceMap = iastRewriter.getOriginalPathAndLineFromSourceMap
-      if (getOriginalPathAndLineFromSourceMap) {
-        getRewriterOriginalPathAndLineFromSourceMap =
-          getGetOriginalPathAndLineFromSourceMapFunction(chainSourceMap, getOriginalPathAndLineFromSourceMap)
-      }
+      setGetOriginalPathAndLineFromSourceMapFunction(chainSourceMap, iastRewriter)
 
       rewriter = new Rewriter({
```
IlyasShabi, line 62: Are you planning to move this outside IAST ?

packages/datadog-instrumentations/orchestrion.yml
IlyasShabi, line 1: Are we supposed to have one single `orchestrion.yml` file or one for each library?

packages/dd-trace/src/appsec/iast/taint-tracking/filter.js
```diff
@@ -6,11 +6,11 @@ const isPrivateModule = function (file) {
   return file && file.indexOf(NODE_MODULES) === -1
 }
 
-const isNotLibraryFile = function (file) {
-  return file && file.indexOf('dd-trace-js') === -1 && file.indexOf('dd-trace') === -1
+let isLibraryFile = function (file) {
```
dubloom, line 9: Nit: I would argue 'isDDTrace' or something like that is a better name for this function. Because 'library' could be anything.

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter-esm.mjs
```diff
@@ -3,29 +3,34 @@
...
-let port, rewriter
+let port, rewriter, iastEnabled
+
+let orchestrionConfig
 
 export async function initialize (data) {
   if (rewriter) return Promise.reject(new Error('ALREADY INITIALIZED'))
 
-  const { csiMethods, telemetryVerbosity, chainSourceMap } = data
+  const { csiMethods, telemetryVerbosity, chainSourceMap, orchestrion } = data
```
dubloom, line 19: From what I understood, orchestrion is a list of module_name and module_version. Therefore, this name is not very clear. Orchestrion could be a lot of things which make the code not really understandable.

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter-telemetry.js
```diff
@@ -4,39 +4,10 @@ const iastTelemetry = require('../telemetry')
 const { Verbosity } = require('../telemetry/verbosity')
 const { INSTRUMENTED_PROPAGATION } = require('../telemetry/iast-metric')
 
-const telemetryRewriter = {
```
dubloom, line 7: Is there a reason you deleted the whole telemetry part ?

dubloom (commented):
I only reviewed the part related to the rewriter, few comments:
- the rewriter part should be extract as soon as possible from the taint-tracking (I get this is like that to release it quickly). I can help if needed
-  Is the rewriter.js the rewriter for common js ? If yes I would argue it would make sense to rename it rewriter-cjs.

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter-esm.mjs
```diff
@@ -3,29 +3,34 @@
...
   const { NonCacheRewriter } = iastRewriter.default
 
   rewriter = new NonCacheRewriter({
     csiMethods,
     telemetryVerbosity: getName(telemetryVerbosity),
-    chainSourceMap
+    chainSourceMap,
+    orchestrion
   })
+  orchestrionConfig = orchestrion
```
uurien, line 33: why? looks like it is not used

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
```diff
@@ -175,29 +187,28 @@ function enableEsmRewriter (telemetryVerbosity) {
...
-    const chainSourceMap = isFlagPresent('--enable-source-maps')
-    const data = {
-      port: port2,
-      csiMethods,
-      telemetryVerbosity,
-      chainSourceMap
-    }
-
     try {
       Module.register('./rewriter-esm.mjs', {
```
uurien, line 191: as this is not going to be specific thing for iast anymore, should we move this or at least import this from `initialize.mjs` or `loader-hooks.mjs`?

For iast we can assume that the rewriter is not working if node version is `<v20.6.0`, but can whole orchestrion assume that? Asking because `module.register` was added in 20.6.0.

Keep in mind that this code is not executed if there is no loader configured before, we added it for iast because there isn't any need to rewrite esm customer code when esm is not properly configured.

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
```diff
@@ -89,22 +89,34 @@ function getPrepareStackTraceAccessor () {
...
+        return compileMethod.apply(this, [content, filename])
+      }
+      if (isPrivateModule(filename)) {
+        // TODO error tracking needs to be added based on config
+        passes = ['error_tracking']
+        if (config.iast?.enabled) {
+          passes.push('iast')
         }
+      } else {
+        passes = ['orchestrion']
```
uurien, line 105: Before loading each file in `node_module` (could be thousands) we are calling to a wasm module, with the cost of converting all the content from js object to webassembly objects, this can be executed thousands of times just to return a false in the moment that the rewriter checks that the file shouldn't be rewritten (becaues it is not langchain), it can add startup overhead even when the libraries that we want to rewrite are not part of the project (I think that 99% of the apps that we are instrumenting do not use langchain).

Should we parse the orchestoin.yml in node to prevent the extra calls to rewriter? If parsing a yml is a problem... can we return a json object with the parsed yml when the rewriter is initialized, or at least the list of the files that the rewriter should rewrite. imo this will reduce the startup time...

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
bengl: Yes, but not in this PR.

packages/datadog-instrumentations/orchestrion.yml
bengl: One single one.

That said, I can have it generated by having a separate one for each library and just combining them. That's relatively easy. Would you prefer that?

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter-esm.mjs
bengl: It's the config yaml for orchestrion. Would `orchestrionConfig` be preferable?

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter-telemetry.js
bengl: This _should_ be restored elsewhere. I'll double-check that.

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter-esm.mjs
dubloom: I would say yes. At least we know it is a config !

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
bengl: Yes, I think it's an appropriate optimization. Currently, I think it ought to be done in the JS side of the rewriter, rather than here. Does that SGTY?

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
bengl: I was talking about this with @rochdev and the biggest issue with doing this early on in the loading process is that we don't have config. We need to see if `config.iast.enabled` before initializing the rewriter. That might be a breaking change. We can probably figure out a way of initializing at least _that_ config early on, but changing how all that works is still a bigger change, and I'm not sure yet if it's breaking. What I've done instead is ensure that the existing langchain integration (i.e. by traditional RITM instrumentation) still works, so we're not reliant on Orchestrion for that. That way this PR is not a breaking change, and we can figure out how to enable Orchestrion for older Node.js at a later time, maybe in v6. How does that sound?

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
uurien: yeah, SGTM, because is the rewriter who knows about the `orchestrion`, we can rely on them.

Not sure if we should also rely in rewriter to decide about `isPrivateModule()`, but it could be done in the future.

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
uurien: > We need to see if config.iast.enabled before initializing the rewriter.

oh, I'm stupid, yes. It was also one of the reasons I wrote this here originally.

> Keep in mind that this code is not executed if there is no loader configured before, we added it for iast because there isn't any need to rewrite esm customer code when esm is not properly configured.

I emphasize on this, this code is not executed if customer has not `--loader` configured or `initialize.mjs` loaded. 
Is something that you can assume for now?

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter-telemetry.js
bengl: Yeah, so this entire file was just to turn telemetry on and off based on verbosity. I found it very confusing and redundant, so I replaced most of it with just line 97 in `rewriter.js`. I _think_ that preserves functionality. The tests needed to be removed because they were just for the `on/off` mechanism.

I needed to ensure appropriate args were being passed along and this wrapper was making it all confusing.

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
bengl: > Is something that you can assume for now?

That's fine for now. We can adjust this later if needed, but for now we just assume this way of enabling ESM support was used.

packages/dd-trace/src/appsec/iast/taint-tracking/filter.js
bengl: Done.

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter-esm.mjs
bengl: Done.

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter-esm.mjs
bengl: Done.

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
bengl: Alright, made this optimization in the rewriter PR, so closing this one.

packages/datadog-instrumentations/orchestrion.yml
tlhunter: It would probably be easier to maintain if we had a YAML file local to each instrumentation?

uurien (dismissed):
approving, but I haven't reviewed the code related with langchain, only the rewriter related code.

sabrenner (dismissed):
lgtm for all the langchain stuff! we did a lot of verifying in our rubber duck/debugging sessions, and i verified manually as well.

just left some cleanup comments that the linter is probably complaining about anyways, and one small question that doesn't need action on this PR.

dubloom (dismissed):

packages/datadog-instrumentations/src/langchain.js
```diff
@@ -46,51 +40,54 @@ for (const extension of extensions) {
...
+
+            wrap(this, 'embedQuery', 'apm:@langchain/core:Embeddings_embedQuery', namespace)
+            wrap(this, 'embedDocuments', 'apm:@langchain/core:Embeddings_embedDocuments', namespace)
+          }
+        }
+      })
+    } else {
+      const channel = tracingChannel('orchestrion:@langchain/core:Embeddings_constructor')
+      channel.subscribe({
+        end (ctx) {
```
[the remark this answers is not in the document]
bengl: :shrug: We can deal with this later once we decide what we're going to do with the instrumentations directory.

.npmignore
```diff
@@ -3,6 +3,7 @@
 !packages/*/lib/**/*
 !packages/*/src/**/*
 !packages/*/index.js
+!packages/datadog-instrumentations/orchestrion.yml
```
rochdev, line 6: Nit, but since it's just a config file could we just move it to the root?

integration-tests/helpers/index.js
```diff
@@ -25,10 +25,13 @@ async function runAndCheckOutput (filename, cwd, expectedOut) {
       out = Buffer.concat([out, data])
     })
     proc.stderr.pipe(process.stdout)
-    proc.on('exit', () => resolve(out.toString('utf8')))
-    setTimeout(() => {
+    const exitTimeout = setTimeout(() => {
       if (proc.exitCode === null) proc.kill()
-    }, 1000) // TODO this introduces flakiness. find a better way to end the process.
+    }, 10000) // TODO this introduces flakiness. find a better way to end the process.
```
rochdev, line 30: Why did this have to be increased?

packages/datadog-instrumentations/src/langchain.js
```diff
@@ -35,9 +28,10 @@ const extensions = ['js', 'cjs']
 
 for (const extension of extensions) {
   addHook({ name: '@langchain/core', file: `dist/runnables/base.${extension}`, versions: ['>=0.1'] }, exports => {
-    const RunnableSequence = exports.RunnableSequence
-    shimmer.wrap(RunnableSequence.prototype, 'invoke', invoke => wrapLangChainPromise(invoke, 'chain'))
-    shimmer.wrap(RunnableSequence.prototype, 'batch', batch => wrapLangChainPromise(batch, 'chain'))
+    if (extension === 'cjs') {
+      wrap(exports.RunnableSequence.prototype, 'invoke', 'orchestrion:@langchain/core:RunnableSequence_invoke')
```
rochdev, line 32: Any way to configure these names? It would be nice if we could update the publisher to the new approach without the subscriber having to know about it.

packages/datadog-instrumentations/src/orchestrion-config/index.js
```diff
@@ -0,0 +1,5 @@
+const path = require('path')
+const fs = require('fs')
+
+// TODO this needs to be inlined to prevent issues in bundling
```
rochdev, line 4: Doesn't the native addon itself have the same issue?

packages/dd-trace/test/llmobs/plugins/langchain/index.spec.js
```diff
@@ -59,6 +63,13 @@ describe('integrations', () => {
 
   describe('langchain', () => {
     before(async () => {
+      iastFilter.isDdTrace = file => {
```
rochdev, line 66: Could this somehow be done outside of individual tests? Maybe in setup files?

rochdev: The more I think about it the more I wonder if this could be done another way entirely instead too.

.npmignore
bengl: Since it's effectively a list of definitions of instrumentations, it made sense to me to keep it close to existing instrumentations. I'd prefer that it's in here.

integration-tests/helpers/index.js
bengl: It doesn't. Latest rev doesn't touch this number.

packages/datadog-instrumentations/src/langchain.js
bengl: The prefix of `orchestrion:${moduleName}:` is by design so we always know where this is coming from.

Not only that, the assumption that the subscriber wouldn't have to know about it is incorrect, because orchestrion doesn't do some of the processing previously done in the publisher, and uses different property names on the context object.

packages/datadog-instrumentations/src/orchestrion-config/index.js
bengl: The rewriter is entirely wasm, not an addon, but yeah, maybe. This comment is more a reminder to think about this sort of thing, and not necessarily a future requirement. Would you like me to reword it?

packages/dd-trace/test/llmobs/plugins/langchain/index.spec.js
bengl: We can (and probably should) encapsulate this in a test helper, but I think that can probably be done in a future PR once it becomes necessary, i.e. when we have multiple instrumentations. WDYT?

rochdev (approved):

PR #5498, opened 2025-03-28
[DI] Temporarily disable the Node.js 18 Debugger tests

watson:
### What does this PR do?

As the title says :)

### Motivation

There's a race condition in Node.js 18 that these tests sometimes trigger when requiring the `async_hooks` module.
    
Once https://github.com/DataDog/dd-trace-js/pull/5357 is merged, this PR can be reverted.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/.github/workflows/debugger.yml
+++ b/.github/workflows/debugger.yml
@@ -17,7 +17,7 @@ jobs:
   ubuntu:
     strategy:
       matrix:
-        version: [18, 20, 22, latest]
+        version: [20, 22, latest]
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


BridgeAR (changes requested):
The latest Node.js release includes a fix. As soon as the CI installs the latest version, it's going to also be resolved. I don't think we need this therefore.
Right now, the cache is still hit. Seems like we do not automatically check for latest versions.

PR #5448, opened 2025-03-20
Pin composite actions SHA

TonyCTHsu:
### Motivation

Followed up on https://github.com/DataDog/dd-trace-js/pull/5193

- Composite actions under `.github/actions/*` are not pinned
- Dependabot is only configured to scanned `.github/workflows` 

### What does this PR do?

- Pin actions under `.github/actions/*`
- Configure Dependabot to scan composite actions

PR #5504, opened 2025-03-31
Always check for the latest Node.js version being installed

BridgeAR:
Make sure all actions use the one node installer action

This aligns the behavior, allows to update the action easier and it makes sure we always use consistent versions.

### DIFF
--- a/.github/actions/node/action.yml
+++ b/.github/actions/node/action.yml
@@ -8,6 +8,8 @@ inputs:
 runs:
   using: composite
   steps:
-    - uses: actions/setup-node@v4
+    - uses: actions/setup-node@1d0ff469b7ec7b3cb9d8673fde0c81c44821de2a # v4.2.0
       with:
         node-version: ${{ inputs.version == 'eol' && '16' || inputs.version == 'oldest' && '18' || inputs.version == 'maintenance' && '20' || inputs.version == 'active' && '22' || inputs.version }}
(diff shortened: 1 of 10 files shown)


.github/actions/node/action.yml
```diff
@@ -8,6 +8,8 @@ inputs:
 runs:
   using: composite
   steps:
-    - uses: actions/setup-node@v4
+    - uses: actions/setup-node@1d0ff469b7ec7b3cb9d8673fde0c81c44821de2a # v4.2.0
```
watson, line 11: I don't think we should do this until that the automated process for updating these hashes are expanded to also check the `.github/actions` directory. I think right now it's just checking the `.github/workflows` repo. I've been meaning to raise this as an issue. Until then, this will be stuck at this hardcoded hash which is probably not what we want

watson, line 11: I don't think we should do this until that the automated process for updating these hashes are expanded to also check the `.github/actions` directory. I think right now it's just checking the `.github/workflows` repo. I've been meaning to raise this as an issue. Until then, this will be stuck at this hardcoded hash which is probably not what we want

watson: Nevermind... https://github.com/DataDog/dd-trace-js/pull/5448 will fix it

watson (approved):

PR #5505, opened 2025-03-31
Profiler needn't use async_hooks when tracing is off

szegedi:
### What does this PR do?
We can delay profiler’s initialization of `async_hooks` until it's needed for either code hotspots or endpoint profiling. We can also make sure that these features are not enabled when tracing is off.

### Motivation
This should reduce the overhead of profiling when tracing is disabled as it avoids initializing and using `async_hooks`. `async_hooks` is used for features that collect data from tracing spans, so when tracing is disabled we can avoid hook initialization.

### DIFF
--- a/packages/dd-trace/src/profiler.js
+++ b/packages/dd-trace/src/profiler.js
@@ -8,7 +8,19 @@ process.once('beforeExit', () => { profiler.stop() })
 
 module.exports = {
   start: config => {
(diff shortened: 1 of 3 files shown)


rochdev (commented):
What happens if tracing is enabled at runtime by remote config?

szegedi:
> What happens if tracing is enabled at runtime by remote config?

Yeah, these features would remain disabled. I guess if I wanted to do this properly, I'd need a diagnostic channel to be able to observe remote configuration changes, right? Would it be okay to create one?

PR #5508, opened 2025-03-31
add support for cgroupv2 inode fallback

rochdev:
### What does this PR do?

Add support for cgroupv2 inode fallback.

### Motivation

When using cgroupv2 the container ID is not always set, but the inode can be used as a fallback for correlation.


packages/dd-trace/src/exporters/common/docker.js
```diff
@@ -2,34 +2,64 @@
...
   }
 }
 
 module.exports = {
-  // can be the container ID but not always depending on the orchestrator
-  id () {
-    return entityId
+  inject (carrier) {
+    if (entityId) {
+      carrier['Datadog-Container-Id'] = entityId
```
tlhunter, line 55: Is the `Id` case inconsistency intentional?

rochdev: @tlhunter Do you want me to make the change or it's fine as it is? And if I should make the change, what casing would you like to see?

tlhunter: I suppose it's fine as is, especially if Python is doing it this way and with the agent being robust with its parsing.

tlhunter (approved):

PR #5515, opened 2025-04-01
tracing: fix instrumentation register code

### LINKED ISSUE #5092: [BUG]: Express integration - v5.29.0+ reports incompatible version found when using Express v4
### Tracer Version(s)

5.29.0

### Node.js Version(s)

22.11.0

### Bug Report

When upgrading our dd-trace module, when we update to v5.29.0 or higher, we get the following output when debug mode is enabled.

```
$ DD_TRACE_DEBUG=true DD_TRACE_LOG_FILE=ddtrace_logs.log NODE_ENV=production node --import ./src/node-instrumentation.mjs src/server/index.mjs
Found incompatible integration version: express@4.21.2
```

I know that 5.29.0 is the same version that released support for express v5, but I would assume that express v4 is still supported based on the README (as it's based on node version).

### Reproduction Code

```js
// ./src/server/node-instrumentation.mjs 

tracer.init({
  hostname: process.env.DD_HOST,
  service: 'app',
  debug: process.env.DATADOG_DEBUG_ENABLED === 'true',
});

tracer.use('express')
```

```js
// ./src/server/index.mjs

import express from 'express';

const app = express()
const port = 3000

app.get('/', (req, res) => {
  res.send('Hello World!')
})

app.listen(port, () => {
  console.log(`Example app listening on port ${port}`)
})
```

Start command

```
$ DD_TRACE_DEBUG=true DD_TRACE_LOG_FILE=ddtrace_logs.log NODE_ENV=production node --import ./src/node-instrumentation.mjs src/server/index.mjs
```

### Error Logs

N/A

### Tracer Config

```js
tracer.init({
  hostname: process.env.DD_HOST,
  service: 'app',
  debug: process.env.DATADOG_DEBUG_ENABLED === 'true',
});
```

### Operating System

Darwin XXX 24.2.0 Darwin Kernel Version 24.2.0

### Bundling

No Bundling

wconti27:
### What does this PR do?

Addresses an issue where "incompatible integration version" was incorrectly logged for packages like Redis v4+ and Express v4. This occurred because the compatibility check didn't recognize when instrumentation succeeded by patching internal dependencies (e.g., `@redis/client` for redis) rather than the main package directly, or when only a partial patch was applied (for Express, we have different hooking code between v4 and v5).

This PR introduces a package linking mechanism via registerPackageLink in `helpers/instrument.js`. The registration logic in helpers/register.js now uses this metadata and a global success tracking set (`successfullyPatchedCombos`). Before logging incompatibility for a package version, it checks if any explicitly linked dependent package has already been successfully instrumented (using a prefix check against the global set). This prevents the incompatability log when the core functionality is covered via a related module. The specific Redis link (`redis -> @redis/client, @node-redis/client`) has been added. Additionally, it uses `successfullyPatchedCombos` to check whether some portion of the package was successfully instrumented, since express was previously throwing the incompatability log since only either v4 or v5 patching hooks were being matched.

Fixes #5092 

### Motivation
This impacts SSI injection since the instrumentation was logging that it was being aborted for v4 express, which is not correct.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/datadog-instrumentations/src/helpers/instrument.js
+++ b/packages/datadog-instrumentations/src/helpers/instrument.js
@@ -14,6 +14,28 @@ exports.channel = function (name) {
   return ch
 }
 
+const packageLinks = new Map()
+
+/**
+ * Registers a link between a parent package and its internally instrumented children.
+ * Helps determine compatibility if the parent lacks direct hooks.
+ * @param {string} parentName The name of the parent package (e.g., 'redis')
+ * @param {string[]} childNames An array of instrumented child package names (e.g., ['@redis/client'])
+ */
+exports.registerPackageLink = function registerPackageLink (parentName, childNames) {
+  if (!packageLinks.has(parentName)) {
+    packageLinks.set(parentName, new Set())
+  }
+  const links = packageLinks.get(parentName)
+  for (const childName of childNames) {
+    links.add(childName)
+  }
+}
+
+exports.getPackageLinks = function getPackageLinks () {
+  return packageLinks
+}
+
 /**
  * @param {string} args.name module name
  * @param {string[]} args.versions array of semver range strings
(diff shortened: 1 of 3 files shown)


rochdev (changes requested):
Wouldn't just removing the `namesAndSuccesses['${name}@${version}'] = false` assignment in `register.js` fix the issue? Why does that assignment exist to begin with? How is loading a file an indication that a version is not supported?

PR #5516, opened 2025-04-01
fix incompatible version detection false positives

rochdev:
### What does this PR do?

Fix incompatible version detection false positives.

### Motivation

See #5515 as it solves the same issue but in a different way that instead ignores file matches entirely and focuses only on the version since that is what we're checking for support.


BridgeAR:
@rochdev is this something that is still worked on? I am not sure if this got resolved differently in the meanwhile.

rochdev:
Closing as this was an attempt to simplify another approach for what the title says but it ended up being incomplete.

PR #5519, opened 2025-04-02
Filter out log stack frames from error log messages

szegedi:
### What does this PR do?
Ensures that when log writer writes a stack trace for an error message it will omit the logging code frames from it.

### Motivation
Log writer will transform a message into an error with a stack trace before logging it. However then it always logs the frames for its own code as well. This causes the frames:
```
at /path/to/my/app/node_modules/dd-trace/packages/dd-trace/src/log/writer.js:69:46
at withNoop (/path/to/my/app/node_modules/dd-trace/packages/dd-trace/src/log/writer.js:21:3)
at onError (/path/to/my/app/node_modules/dd-trace/packages/dd-trace/src/log/writer.js:69:18)
at Channel.publish (node:diagnostics_channel:143:9)
at Object.error (/path/to/my/app/node_modules/dd-trace/packages/dd-trace/src/log/index.js:82:20)
```
to be present in all logged errors. This can lead to a confusion as people looking to diagnose problems will Google for these frames, find unrelated issues, and presume they're related. I'm submitting this because just yesterday I had a TEE ask if maybe https://github.com/DataDog/dd-trace-js/issues/5111 is a related issue to something completely different based on the match on these frames.

### Additional Notes
Jira: [PROF-11569]


packages/dd-trace/src/log/writer.js
```diff
@@ -61,12 +62,28 @@ function getErrorLog (err) {
...
   // calling twice logger.error() because Error cause is only available in nodejs v16.9.0
   // TODO: replace it with Error(message, { cause }) when cause has broad support
-  if (formatted) withNoop(() => logger.error(new Error(formatted)))
+  if (formatted) {
+    withNoop(() => {
+      const l = Error.stackTraceLimit
+      Error.stackTraceLimit = 0
+      const e = new Error(formatted)
+      Error.stackTraceLimit = l
+      Error.captureStackTrace(e, stackTraceLimitFunction)
```
watson, line 83: Would it make sense to fall back to `onError` in case there's no `stackTraceLimitFunction`?

```suggestion
      Error.captureStackTrace(e, stackTraceLimitFunction ?? onError)
```

tlhunter (approved):
It's worth noting that this may blast us with "new" errors as far as the error logging product is concerned since the stack traces will change.

PR #5524, opened 2025-04-02
Start collecting async IDs in profiles

szegedi:
### What does this PR do?
* Upgrades `@datadog/pprof` to 5.7.0
* Enables its [feature](https://github.com/DataDog/pprof-nodejs/pull/197) for collecting Node's execution async IDs in profiles.
* Adds a `DD_PROFILING_ASYNC_ID_ENABLED` env var to regulate the feature. It defaults to true when timelines are enabled. We have this env var mostly so we have a lever to give to customers to turn the feature off if it causes problems. We do not anticipate problems. (An earlier implementation of this feature did not guard against trying to retrieve the async ID while V8 was doing garbage collection. That caused crashes, and it has been fixed in pprof 5.7.0.)

### Motivation
We have an insight in the profiling backend for detecting long-running event callbacks that block other callbacks from running on the event loop thread. An initial version of this insight uses span IDs associated with profiling samples to identify streaks of adjacent samples that belong to the same invocation. We have since upgraded the backend algorithm to also be able to use execution async IDs when available. Execution async ID is a better datum for this for two reasons:
* it is less likely to cause false positives if two callbacks working on behalf of the same span run immediately one after the other, and
* async IDs are available to the profiler even when tracing is turned off.

### Additional information
* The ability to handle async IDs was already added in #5316, this PR merely turns the feature on in the underlying profiler.
* We released an earlier version of profiler with #5153 that didn't have the garbage collection safeguards then reverted it with #5174. This now re-enables the feature with the profiler having the necessary guards to only read the async ID when it can be safely done.

### Additional Notes
Jira: [PROF-11107]

[PROF-11107]: https://datadoghq.atlassian.net/browse/PROF-11107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integration-tests/profiler/profiler.spec.js
+++ b/integration-tests/profiler/profiler.spec.js
@@ -355,7 +355,12 @@ describe('profiler', () => {
             case threadNameKey: threadName = label.str; break
             case threadIdKey: threadId = label.str; break
             case osThreadIdKey: osThreadId = label.str; break
-            case asyncIdKey: asyncId = label.num; break
+            case asyncIdKey:
+              asyncId = label.num
+              if (asyncId === 0) {
+                asyncId = undefined
+              }
+              break
             default: assert.fail(`Unexpected label key ${strings.dedup(label.key)} ${encoded}`)
           }
         }
--- a/integration-tests/profiler/profiler.spec.js
+++ b/integration-tests/profiler/profiler.spec.js
@@ -389,16 +389,27 @@ describe('profiler', () => {
             continue
           }
           const spanData = { rootSpanId, endpoint, asyncId }
+          // Record async ID so we can verify we encountered 9 different values.
+          // Async ID can be sporadically missing if sampling hits an intrinsified
+          // function.
+          if (asyncId !== undefined) {
+            asyncIds.add(asyncId)
+          }
           const existingSpanData = spans.get(spanId)
           if (existingSpanData) {
             // Span's root span, endpoint, and async ID must be consistent
             // across samples.
-            assert.deepEqual(spanData, existingSpanData, encoded)
+            assert.equal(existingSpanData.rootSpanId, rootSpanId, encoded)
+            assert.equal(existingSpanData.endpoint, endpoint, encoded)
+            // Account for asyncID sporadically missing
+            if (existingSpanData.asyncId === undefined) {
+              existingSpanData.asyncId = asyncId
+            } else if (asyncId !== undefined) {
+              assert.equal(existingSpanData.asyncId, asyncId, encoded)
+            }
           } else {
             // New span id, store span data
             spans.set(spanId, spanData)
-            // Record async ID so we can verify we encountered 9 different values
-            asyncIds.add(asyncId)
             // Verify endpoint value
             const endpointVal = strings.strings[endpoint]
             switch (endpointVal) {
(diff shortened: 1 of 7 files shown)


BridgeAR (approved):
LGTM

BridgeAR (changes requested):
Tests are actually failing. It seems like the wall.js options have to be changed to know about the option.

integration-tests/profiler/profiler.spec.js
```diff
@@ -407,9 +424,10 @@ describe('profiler', () => {
           }
         }
       }
-      // Need to have a total of 9 different spans, with 3 different root spans
-      // and 3 different endpoints.
+      // Need to have a total of 9 different spans, with 9 different async IDs,
+      // 3 different root spans, and 3 different endpoints.
       assert.equal(spans.size, 9, encoded)
+      assert.equal(asyncIds.size, 9, encoded)
```
BridgeAR, line 430: I guess this could even be zero due to the IDs sometimes missing?

szegedi: We collect multiple samples for each span – we'll have a total of 9 spans, each span runs for ~100ms, and we sample at 99Hz, so statistically each span will have ~9.9 samples (I might increase the span execution time to 102ms which'd bring the expected sample value a tad over 10.) So we'd need to fail to capture the async ID in all 9-10 samples for a span in order for a particular async ID to be completely absent from this set. 

Unfortunately, this is [exactly what we see on Ubuntu profiler tests now](https://github.com/DataDog/dd-trace-js/actions/runs/14227845270/job/39872387872?pr=5524): it observes 2-3 async IDs. Not sure what to make of it, but I'll be inspecting this by running some tests locally on an arm Ubuntu VM to try to reproduce.

szegedi: On a closer look at the [profiling GH workflow](https://github.com/DataDog/dd-trace-js/blob/master/.github/workflows/profiling.yml) though it turns out that we only run profiling tests with `active-lts` version of Node on macOS, but additionally run them with `oldest-maintenance-lts` and `newest-maintenance-lts` on Linux.

Oldest maintenance is 18.20.8 and for it the test indeed fails even on my local macOS. It also fails for 20.19.0, but it passes for 22.14.0. I guess one way to fix the issue is to only enable the feature only on Node 22 and above. It's an interesting question though why does `node::AsyncHooksGetExecutionAsyncId()` work reliably on Node 22 but not on earlier ones. Do you maybe have some information on this?

szegedi: Hm, I bisected on minor node versions and 22.10.0 is the first one where the test reliably passes, and 22.9.0 the last one where it reliably fails. The functions that execute here: `AsyncHooksGetExecutionAsyncId` and its callees  `Environment::GetCurrent` and `Environment::execution_async_id` don't seem like they had a code change in years. Except for https://github.com/nodejs/node/commit/d6bcc448296af0d57e87eff3762a8cea22602cb8 which did indeed go into 22.10.0 but I'd be surprised if that was the one that made the difference.

BridgeAR (approved):

PR #5525, opened 2025-04-02
Adding Support for `TRACE_PROPAGATION_BEHAVIOR_EXTRACT` 

mhlidd:
### What does this PR do?
Introduces support for a new config `TRACE_PROPAGATION_BEHAVIOR_EXTRACT`. [RFC](https://docs.google.com/document/d/1KWx5F-XEzJ55XZPwY30t0otVIyOUYHWsDhEt3YPDnso/edit?pli=1&tab=t.0#heading=h.g0jmfjh7mhtk)

Accepted Values:
- `continue`: The tracing library continues the trace from the incoming headers, if present. Also, incoming baggage is propagated. (Default Value)
- `restart`: The tracing library always starts a new trace with a new trace-id. Context extraction will occur as otherwise configured, except that the local span will no longer have a parent-child span relationship with the incoming trace context, instead it will reference the incoming trace context via a span link. Also, incoming baggage is propagated.
- `ignore`: The tracing library always starts a new trace with a new trace-id without creating any span links. Also, incoming baggage is dropped.
### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/test/opentracing/span.spec.js
+++ b/packages/dd-trace/test/opentracing/span.spec.js
@@ -3,7 +3,6 @@
 require('../setup/tap')
 
 const Config = require('../../src/config')
-const DatadogTracer = require('../../src/tracer')
 const TextMapPropagator = require('../../src/opentracing/propagation/text_map')
 
 const { channel } = require('dc-polyfill')
@@ -491,17 +490,21 @@ describe('Span', () => {
       })
 
       it('should not propagate baggage items when Trace_Propagation_Behavior_Extract is set to ignore', () => {
-        process.env.DD_TRACE_PROPAGATION_BEHAVIOR_EXTRACT = 'ignore'
-        const config = new Config()
-        const tracers = new DatadogTracer(config, undefined)
-        span = new Span(tracers, processor, prioritySampler, { operationName: 'operation', parent })
+        tracer = {
+          _config: {
+            tracePropagationBehaviorExtract: 'ignore'
+          }
+        }
+        span = new Span(tracer, processor, prioritySampler, { operationName: 'operation', parent })
         expect(span._spanContext._baggageItems).to.deep.equal({})
       })
 
       it('should propagate baggage items when Trace_Propagation_Behavior_Extract is set to restart', () => {
-        process.env.DD_TRACE_PROPAGATION_BEHAVIOR_EXTRACT = 'restart'
-        const config = new Config()
-        const tracer = new DatadogTracer(config, undefined)
+        tracer = {
+          _config: {
+            tracePropagationBehaviorExtract: 'restart'
+          }
+        }
         span = new Span(tracer, processor, prioritySampler, { operationName: 'operation', parent })
         expect(span._spanContext._baggageItems).to.deep.equal({ foo: 'bar' })
       })
--- a/packages/dd-trace/test/opentracing/propagation/text_map.spec.js
+++ b/packages/dd-trace/test/opentracing/propagation/text_map.spec.js
@@ -1181,7 +1181,6 @@ describe('TextMapPropagator', () => {
     describe('tracePropagationBehaviorExtract', () => {
       let traceId
       let spanId
-      let extractedContext
 
       beforeEach(() => {
         traceId = '1111aaaa2222bbbb3333cccc4444dddd'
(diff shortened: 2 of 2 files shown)


packages/dd-trace/src/config.js
```diff
@@ -1114,6 +1123,13 @@ class Config {
     this._setTags(opts, 'tags', tags)
     this._setBoolean(opts, 'traceId128BitGenerationEnabled', options.traceId128BitGenerationEnabled)
     this._setBoolean(opts, 'traceId128BitLoggingEnabled', options.traceId128BitLoggingEnabled)
+    if (options.tracePropagationBehaviorExtract) {
```
khanayan123, line 1126: Do we want to support setting this programmatically? Generally I think we should move towards only allowing configs to be set through environment variables/stable config

packages/dd-trace/src/config.js
```diff
@@ -1114,6 +1123,13 @@ class Config {
...
+      const stringPropagationBehaviorExtract = String(options.tracePropagationBehaviorExtract)
```
khanayan123, line 1127: This code is also duplicated with the above

packages/dd-trace/src/opentracing/propagation/text_map.js
```diff
@@ -354,6 +355,19 @@ class TextMapPropagator {
 
     this._extractBaggageItems(carrier, context)
 
+    if (this._config.tracePropagationBehaviorExtract !== 'continue') {
+      context._links = []
```
khanayan123, line 359: should we be instantiating a new links array?

packages/dd-trace/src/opentracing/propagation/text_map.js
```diff
@@ -354,6 +355,19 @@ class TextMapPropagator {
...
+      if (this._config.tracePropagationBehaviorExtract === 'restart') {
```
khanayan123, line 360: Is the nested if statement needed? can we just directly check for 'restart' instead?

packages/dd-trace/test/opentracing/span.spec.js
```diff
@@ -411,6 +412,42 @@ describe('Span', () => {
     })
   })
 
+  describe('tracePropagationBehaviorExtract', () => {
```
khanayan123, line 415: These tests belong in tracer.spec.js

khanayan123 (commented):
Just some minor comments, mainly LGTM

packages/dd-trace/src/opentracing/propagation/text_map.js
mhlidd: We want to "reset" the links from extraction when behavior is not continue. ([RFC](https://docs.google.com/document/d/1KWx5F-XEzJ55XZPwY30t0otVIyOUYHWsDhEt3YPDnso/edit?pli=1&tab=t.0#heading=h.be4r8fl6z8mi))

packages/dd-trace/src/opentracing/propagation/text_map.js
mhlidd: I can separate this into two distinct checks, one for `ignore` and one for `restart`. This way if future values are added to this config, we would avoid any unexpected behavior.

packages/dd-trace/src/config.js
mhlidd: This one is for `applyOptions` and the one above is for `applyEnv`

packages/dd-trace/test/opentracing/span.spec.js
```diff
@@ -470,5 +471,40 @@ describe('Span', () => {
...
+          },
+          _isRemote: true
+        }
+      })
+
+      it('should not propagate baggage items when Trace_Propagation_Behavior_Extract is set to ignore', () => {
+        process.env.DD_TRACE_PROPAGATION_BEHAVIOR_EXTRACT = 'ignore'
+        const config = new Config()
+        const tracers = new DatadogTracer(config, undefined)
+        span = new Span(tracers, processor, prioritySampler, { operationName: 'operation', parent })
```
khanayan123, line 497: what if instead of tracers you pass in parent? that way you won't need to import DatadogTracer

packages/dd-trace/test/opentracing/span.spec.js
```diff
@@ -470,5 +471,40 @@ describe('Span', () => {
 
       expect(processor.process).to.have.been.calledOnce
     })
+
+    describe('tracePropagationBehaviorExtract and Baggage', () => {
+      let parent
+
+      beforeEach(() => {
+        parent = {
```
khanayan123, line 479: I would update this to add:

```
parent = {
_config: {
tracePropagationBehaviorExtract: "ignore"
}
...
}
```

khanayan123 (approved):

PR #5526, opened 2025-04-02
add feature to disable mongo heartbeats spans

wconti27:
### What does this PR do?
This PR adds a plugin configuration `mongoHeartbeatDisabled` for the `mongodb-core`, along with an environment variable configuration, `DD_TRACE_MONGO_HEARTBEAT_DISABLED` that can be set to `true` to disable tracing of heartbeat spans.

### Motivation
Escalation

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/datadog-plugin-mongodb-core/src/index.js
+++ b/packages/datadog-plugin-mongodb-core/src/index.js
@@ -12,16 +12,16 @@ class MongodbCorePlugin extends DatabasePlugin {
 
   configure (config) {
     super.configure(config)
-    this.config.mongoHeartbeatDisabled = coalesce(
-      config.mongoHeartbeatDisabled,
-      process.env.DD_TRACE_MONGO_HEARTBEAT_DISABLED,
-      false
+    this.config.heartbeatEnabled = coalesce(
+      config.heartbeatEnabled,
+      process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED,
+      true
     )
   }
 
   start ({ ns, ops, options = {}, name }) {
-    // heartbeat commands can be disabled if this.config.mongoHeartbeatDisabled is true
-    if (isHeartbeatDisabled(ops, this.config)) {
+    // heartbeat commands can be disabled if this.config.heartbeatEnabled is false
+    if (!this.config.heartbeatEnabled && isHeartbeat(ops, this.config)) {
       return
     }
     const query = getQuery(ops)
@@ -168,12 +168,7 @@ function isBinary (val) {
   return val && val._bsontype === 'Binary'
 }
 
-function isHeartbeatDisabled (ops, config) {
-  // do nothing if mongoHeartbeatDisabled is not set to true
-  if (config.mongoHeartbeatDisabled !== true) {
-    return false
-  }
-
+function isHeartbeat (ops, config) {
   // Check if it's a heartbeat command hello: 1 or helloOk: 1
   return ops && typeof ops === 'object' && (ops.hello === 1 || ops.helloOk === true)
 }
--- a/packages/datadog-plugin-mongodb-core/src/index.js
+++ b/packages/datadog-plugin-mongodb-core/src/index.js
@@ -9,6 +9,10 @@ class MongodbCorePlugin extends DatabasePlugin {
   // should be removed if one day this will be fixed
   static get peerServicePrecursors () { return [] }
   start ({ ns, ops, options = {}, name }) {
+    // heartbeat commands can be disabled if this.config.disableMongoHeartbeat is true
+    if (isHeartbeatDisabled(ops, this.config)) {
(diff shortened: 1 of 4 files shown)


packages/dd-trace/src/config.js
```diff
@@ -767,6 +768,7 @@ class Config {
       DD_VERTEXAI_SPAN_CHAR_LIMIT,
       DD_TRACE_INFERRED_PROXY_SERVICES_ENABLED,
       DD_TRACE_NATIVE_SPAN_EVENTS,
+      DD_TRACE_DISABLE_MONGO_HEARTBEAT,
```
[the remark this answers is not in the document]
rochdev: Also, is there a use case for the health check? Should we just remove it entirely or disable it by default in the next major?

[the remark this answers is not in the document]
wconti27: Will fix the plugin configuration 👍 . As for disabling these spans, I can't see a valid use case for keeping them tbh. They just add noise. We can add the plugin configuration as enabled for now, then  switch it to disabled in the next major. Side note, is there a way we can track configurations that we aim to change in the next major? If not, that could be something helpful to add since something like this can easily be forgotten by the next major.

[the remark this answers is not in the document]
wconti27: Honestly surprised that we haven't received other requests to disable them

packages/datadog-plugin-mongodb-core/src/index.js
```diff
@@ -1,14 +1,29 @@
...
   static get id () { return 'mongodb-core' }
   static get component () { return 'mongodb' }
   // avoid using db.name for peer.service since it includes the collection name
   // should be removed if one day this will be fixed
   static get peerServicePrecursors () { return [] }
+
+  configure (config) {
+    super.configure(config)
+    this.config.mongoHeartbeatDisabled = coalesce(
+      config.mongoHeartbeatDisabled,
```
rochdev, line 16: `mongo` is redundant here

packages/datadog-plugin-mongodb-core/src/index.js
```diff
@@ -1,14 +1,29 @@
...
+      process.env.DD_TRACE_MONGO_HEARTBEAT_DISABLED,
```
rochdev, line 17: `DD_TRACE_MONGO_HEARTBEAT_DISABLED` should be `DD_TRACE_MONGODB_HEARTBEAT_DISABLED` to match the integration name. While this is not important for now, we might be able to automate this in the future if a standard is respected.

packages/datadog-plugin-mongodb-core/src/index.js
rochdev: Also for most other things we'd say `enabled` and set it to true by default, not the other way around.

packages/datadog-plugin-mongodb-core/src/index.js
```diff
@@ -1,14 +1,29 @@
...
+      false
+    )
+  }
+
   start ({ ns, ops, options = {}, name }) {
+    // heartbeat commands can be disabled if this.config.mongoHeartbeatDisabled is true
+    if (isHeartbeatDisabled(ops, this.config)) {
```
rochdev, line 24: The name is confusing as it returns if it's a heartbeat, not if it's disabled. Or I guess it checks both which makes naming more difficult. Maybe something like `!config.heartbeatEnabled && !isHeartbeat(ops, this.config` or combine them in some sort of `shouldTrace(ops, this.config)`.

packages/dd-trace/src/config.js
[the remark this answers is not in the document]
rochdev: > Side note, is there a way we can track configurations that we aim to change in the next major?

I don't think we have anything like that right now, at least not in code. Historically when close to releases we've been putting ideas as a checklist in a Jira card.

> Honestly surprised that we haven't received other requests to disable them

It's pretty easy to ignore in the UI so my guess is most users just do that.

packages/datadog-plugin-mongodb-core/src/index.js
wconti27: done

wconti27: and makes sense

packages/datadog-plugin-mongodb-core/src/index.js
wconti27: fair enough, fixed

packages/datadog-plugin-mongodb-core/src/index.js
wconti27: understood, fixed the naming

rochdev (approved):

PR #5535, opened 2025-04-04
[DI] Support multiple probes in the same location

watson:
### What does this PR do?

Allow the user to set multiple probes on the same location (file/line/column).

The Chrome DevTools Protocol API that we use (`Debugger.setBreakpoint`) does not allow more than one breakpoint per location. To get around this, we keep a reference to which probes should be at each physical breakpoint and "execute" those when the shared breakpoint is hit.

### Motivation

Align on feature parity with the other tracers

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
+++ b/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
@@ -52,15 +52,15 @@ async function addBreakpoint (probe) {
 
   const release = await lock()
 
-  log.debug(
-    '[debugger:devtools_client] Adding breakpoint at %s:%d:%d (probe: %s, version: %d)',
-    url, lineNumber, columnNumber, probe.id, probe.version
-  )
-
-  const locationKey = generateLocationKey(scriptId, lineNumber, columnNumber)
-  const breakpoint = locationToBreakpoint.get(locationKey)
-
   try {
+    log.debug(
+      '[debugger:devtools_client] Adding breakpoint at %s:%d:%d (probe: %s, version: %d)',
+      url, lineNumber, columnNumber, probe.id, probe.version
+    )
+
+    const locationKey = generateLocationKey(scriptId, lineNumber, columnNumber)
+    const breakpoint = locationToBreakpoint.get(locationKey)
+
     if (breakpoint) {
       // A breakpoint already exists at this location, so we need to add the probe to the existing breakpoint
       await updateBreakpoint(breakpoint, probe)
--- a/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
+++ b/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
@@ -1,10 +1,11 @@
 'use strict'
 
 const { getGeneratedPosition } = require('./source-maps')
+const lock = require('./lock')()
 const session = require('./session')
 const compileCondition = require('./condition')
 const { MAX_SNAPSHOTS_PER_SECOND_PER_PROBE, MAX_NON_SNAPSHOTS_PER_SECOND_PER_PROBE } = require('./defaults')
-const { findScriptFromPartialPath, probes, breakpoints } = require('./state')
+const { findScriptFromPartialPath, locationToBreakpoint, breakpointToProbes, probeToLocation } = require('./state')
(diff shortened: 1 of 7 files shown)


watson:
* **#5535** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5535?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5535?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/dd-trace/src/debugger/devtools_client/breakpoints.js
```diff
@@ -43,54 +44,124 @@ async function addBreakpoint (probe) {
     ({ line: lineNumber, column: columnNumber } = await getGeneratedPosition(url, source, lineNumber, sourceMapURL))
   }
 
+  try {
+    probe.condition = probe.when?.json && compileCondition(probe.when.json)
+  } catch (err) {
+    throw new Error(`Cannot compile expression: ${probe.when.dsl}`, { cause: err })
+  }
+
+  const release = await lock()
```
tylfin, line 53: Should `try` be moved to L54 to ensure `release()` is always called in the `finally` block? Everything here looks safe, but it could protect against future mistakes.

watson: Nice catch (no pun intended)... it is fixed now 👍 It was just before the `try` block in an older version of this PR, but after some refactoring it got moved 😅

packages/dd-trace/src/debugger/devtools_client/lock.js
```diff
@@ -0,0 +1,8 @@
+'use strict'
+
+module.exports = () => async function lock () {
+  if (lock.p) await lock.p
+  let resolve
+  lock.p = new Promise((_resolve) => { resolve = _resolve }).then(() => { lock.p = null })
```
szegedi, line 6: So sad we can't use `Promise.withResolvers` yet.

packages/dd-trace/src/debugger/devtools_client/index.js
```diff
@@ -25,57 +26,92 @@ const expression = `
...
     }
 
-    if (probe.captureSnapshot === true) {
-      // This algorithm to calculate number of sampled snapshots within the last second is not perfect, as it's not a
-      // sliding window. But it's quick and easy :)
-      if (i === 0 && start - globalSnapshotSamplingRateWindowStart > oneSecondNs) {
-        snapshotsSampledWithinTheLastSecond = 1
-        globalSnapshotSamplingRateWindowStart = start
-      } else if (snapshotsSampledWithinTheLastSecond >= MAX_SNAPSHOTS_PER_SECOND_GLOBALLY) {
+    // If all the probes has a condition, we know that it triggered. If at least one probe doesn't have a condition, we
```
szegedi, line 69:
```suggestion
    // If all the probes have a condition, we know that it triggered. If at least one probe doesn't have a condition, we
```

packages/dd-trace/src/debugger/devtools_client/index.js
```diff
@@ -25,57 +26,92 @@ const expression = `
...
+    // need to verify which conditions are met.
+    const shouldVerifyConditions = !(
+      SUPPORT_ITERATOR_METHODS
+        ? probesAtLocation.values()
+        : Array.from(probesAtLocation.values())
+    ).every((probe) => probe.condition !== undefined)
```
szegedi, line 75: `!x.every(p => !f(p))` is same as `x.some(f(p))`, so "not every probe condition is not undefined" could read "some probe condition is undefined" instead.

```suggestion
    const shouldVerifyConditions = (
      SUPPORT_ITERATOR_METHODS
        ? probesAtLocation.values()
        : Array.from(probesAtLocation.values())
    ).some((probe) => probe.condition === undefined)
```

szegedi (changes requested):

szegedi (approved):

PR #5537, opened 2025-04-04
centralize ci action versions

rochdev:
### What does this PR do?

Centralize CI action versions.

### Motivation

These hashes are otherwise repeated dozens of times in many files and might deviate over time, when in reality what we really want is to use latest. Having the version in a single place makes it a lot easier to update everything at once and ensure consistency.

### DIFF
--- a/.github/actions/checkout/action.yml
+++ b/.github/actions/checkout/action.yml
@@ -1,0 +1,6 @@
+name: Checkout
+description: Checkout
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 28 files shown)


BridgeAR:
This needs a rebase

BridgeAR:
Ping @rochdev

rochdev:
Going to close this as there is no way to handle `actions/checkout` and it's one of the biggest offenders.

PR #5546, opened 2025-04-07
Fix span pointers env var & tests

nhulston:
### What does this PR do?

Fixes dynamodb span pointer tests which were not running correctly. 

Also fixes the `trace.dynamoDb.tablePrimaryKeys` env var.

Also, here I remove a duplicate log
Screenshot 2025-04-07 at 4 08 25 PM

### Motivation
It looks someone accidentally renamed the env var `trace.dynamoDb.tablePrimaryKeys` to `aws.dynamoDb.tablePrimaryKeys`. This was not caught because the tests were not failing when expected.
(Originally, it was `aws.dynamoDb` but I changed it to `trace.dynamoDb`. Looks like later, someone must've rebased incorrectly and reverted my change in https://github.com/DataDog/dd-trace-js/pull/5266)

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


rochdev (changes requested):
This should be tested to avoid the same situation happening again.

nhulston:
> This should be tested to avoid the same situation happening again.

@rochdev 
I have unit tests here, I'm not sure how they passed when the env var was initially changed. WDYT? https://github.com/DataDog/dd-trace-js/blob/nicholas.hulston/fix-dynamodb-span-ptr-env-var/packages/datadog-plugin-aws-sdk/test/dynamodb.spec.js#L325-L340

I'm guessing the config does not require defaults to be set for unit tests to pass? But in practice (from manual testing on Lambda), the default does need to be set.

This seems like a bigger issue with testing `config.js`.


I'm currently working on functional tests for Serverless; that's how I discovered this issue:
https://github.com/DataDog/serverless-e2e-tests/pull/28
This will notify the Serverless team of any future regression before we release, and the APM-serverless team has an OKR item to run these tests on every tracer repo

BridgeAR (changes requested):
The faulty change is about the default value. It was originally just undefined. Changing it to the correct name should still result in the same behavior. It is still a falsy value and that did not change by renaming the default.

AFAIC the issue is somewhere else.

nhulston:
> The faulty change is about the default value. It was originally just undefined. Changing it to the correct name should still result in the same behavior. It is still a falsy value and that did not change by renaming the default.
> 
> AFAIC the issue is somewhere else.

@BridgeAR 
It definitely looks that way and I see where you're coming from, but the default value is actually required, see https://github.com/DataDog/dd-trace-js/blob/nicholas.hulston/fix-dynamodb-span-ptr-env-var/packages/dd-trace/src/config.js#L444-L445

I've tested manually and this one-liner fixes our regression

BridgeAR (approved):
I checked again and we seem to iterate over the default keys in the config to merge just those values. That is something we should probably fix. It is weird that we need a default value for any value to be defined later.
This will therefore indeed fix the issue.

After looking into the tests, it seems that those do not get executed properly. If I read the code correct, the test calls testSpanPointers() which returns a method that is never called and the test passes. Please always make sure that the test fails initially before letting it pass. That way we guarantee that our test actually tests the changed code. A good coding pattern for this is Test Driven Development.

I am going to approve this due to indeed fixing the issue but please fix the tests right afterwards. I will have a look at the config sometime during the next weeks.

rochdev (changes requested):
I don't see having tests elsewhere that don't fail on PRs and that might manually be checked before releasing a dependant library after releasing a broken dd-trace a satisfactory testing strategy. If there is a missing test for this feature, let's add it in this repo.

nhulston:
> I don't see having tests elsewhere that don't fail on PRs and that might manually be checked before releasing a dependant library after releasing a broken dd-trace a satisfactory testing strategy. If there is a missing test for this feature, let's add it in this repo.

The issue was that the dynamodb span pointer tests were not failing when they should have failed. I've fixed that issue in this PR. Thanks :)

packages/datadog-plugin-aws-sdk/test/dynamodb.spec.js
```diff
@@ -287,247 +287,291 @@ describe('Plugin', () => {
...
-                  process.env.DD_TRACE_DYNAMODB_TABLE_PRIMARY_KEYS =
-                    '{"OneKeyTable": ["name"]}'
-                  dynamo.putItem({
-                    TableName: oneKeyTableName,
-                    Item: {
-                      name: { S: 'test1' },
-                      foo: { S: 'bar1' }
+            it('should add span pointer for putItem when config is valid', function (done) {
+              setupTest('{"OneKeyTable": ["name"]}')
+                .then(() => {
```
BridgeAR, line 338: Using async await would probably make these tests more readable.

If you like, we could also briefly pair on this.

BridgeAR (commented):
Thanks for looking into the tests! Now that they run, they seem to cause a lot of tests to fail 😅

packages/datadog-plugin-aws-sdk/test/dynamodb.spec.js
```diff
@@ -382,37 +375,38 @@ describe('Plugin', () => {
...
 
-            it('should add span pointer for deleteItem', () => {
-              testSpanPointers({
+            it('should add span pointer for deleteItem', async function () {
+              return testSpanPointers({
                 expectedHashes: '27f424c8202ab35efbf8b0b444b1928f',
-                operation: (callback) => {
-                  dynamo.deleteItem({
+                operation () {
+                  return util.promisify(dynamo.deleteItem).bind(dynamo)({
```
BridgeAR, line 387: Could you please use a helper method that checks if the method already supports promises and to use that instead in those cases? That way we prevent the deprecation warning from popping up and having to fix this later.

BridgeAR (commented):
LGTM with the deprecation warnings gone :)

BridgeAR (dismissed):

packages/datadog-plugin-aws-sdk/test/dynamodb.spec.js
```diff
@@ -274,106 +271,116 @@ describe('Plugin', () => {
         })
 
         describe('span pointers', () => {
-          beforeEach(() => {
-            DynamoDb.dynamoPrimaryKeyConfig = null
-            delete process.env.DD_TRACE_DYNAMODB_TABLE_PRIMARY_KEYS
-          })
+          async function testSpanPointers ({ env, expectedHashes, operation }) {
+            await agent.close({ ritmReset: false, wipe: true })
```
rochdev, line 275: This should be done in `after` or `afterEach`, never in tests directly to avoid dependencies between tests.

nhulston: Thanks. I moved the `agent.close` call into `beforeEach`. However, we still have to load the agent inside this method because we have to update the env var before loading the agent.

packages/datadog-plugin-aws-sdk/test/dynamodb.spec.js
```diff
@@ -33,14 +35,10 @@ describe('Plugin', () => {
...
-        before(() => {
+        before(async () => {
           tracer = require('../../dd-trace')
           tracer.init()
-          return agent.load('aws-sdk')
-        })
-
-        before(async () => {
-          await agent.load('aws-sdk')
+          await agent.load()
```
rochdev, line 41: Why the empty `agent.load()`?

packages/datadog-plugin-aws-sdk/test/dynamodb.spec.js
rochdev: Why would it be needed in `beforeEach` if `afterEach` cleans up properly?

packages/datadog-plugin-aws-sdk/test/dynamodb.spec.js
nhulston: I am just trying to fix a regression; these tests are unrelated

packages/datadog-plugin-aws-sdk/test/dynamodb.spec.js
nhulston: If we try to close the agent but it's already been closed, the tests fail. So the first test in this describe block will fail if we put it in afterEach.

Seems like an issue with the mock agent, but that's outside the scope of this PR.

I'm not a member of the APM team; just trying to fix a feature for Serverless

BridgeAR (approved):
Still LGTM. This is definitely a big improvement. It fixes the reported issue and the tests are now run. So no matter that there are a few more theoretical improvements about how these tests could be implemented, it's a big improvement.

rochdev (approved):

PR #5548, opened 2025-04-08
[DI] Process log template

watson:
### What does this PR do?

Evaluate the log message template in the context of the line the probe is instrumenting.

#### Data structure

The message template for a log probe is provided with the rest of the probe over Remote Config, and is found in the two probe fields `message` and `segments`. Previously, we just used `message` as-is, which is the raw string that the user entered in the UI. However, the backend parses this string and turns it into an AST which is stored in the `segments` array. This PR switches to use this AST if any of the segments requires evaluation. If not, the raw `message` is still just used as-is. The `message` property and `segments` array has the following format:

```js
{
  message: 'Value of foo: {foo}!',
  segments: [
    { str: 'Value of foo: ' }, // str: this segment should not be evaluated, but is just a string
    { dsl: '{foo}', json: { ...ast... } }, // dsl: raw string user entered, json: the parsed AST
    { str: '!' } // another string
  ]
}
```

The AST in the `json` property has the same format as the DI probe conditions already supported by this tracer, so the compiler used to compile probe conditions is reused to compile these segments in this PR.

#### Performance optimizations

If there's more than one probe on a specific file/line/column, all evaluations happen in a single `Debugger.evaluateOnCallFrame` call, as to minimise the number of roundtrips over the Chrome DevTools Protocol, and hence, minimise the time the inspected thread is paused.

#### Stringification method

The `util.inspect` method is used to stringify the results of the code evaluations (expect for strings which are just returned as is). The following options are passed to `util.inspect` to keep the returned string small, and to align as much as possible with the same implementation in the other tracers.

```js
{
  depth: 0, // Only traverse the root level of objects
  customInspect: false, // Do not execute custom [util.inspect.custom](depth, opts, inspect) functions
  maxArrayLength: 3, // Only show the first 3 elements in arrays, sets, maps, etc
  maxStringLength: 8 * 1024, // Truncate strings longer than 8KiB
  breakLength: Infinity // Do not add linebreaks to the returned string
}
```

The total length of the full message is also truncated at 8KiB.

### Motivation

Feature parity with the other tracers.

### Additional Notes

The only two major differences between this implementation and the implementations in the other tracers are;
- If any of the objects traversed by `util.inspect` contains the magic `[Symbol.toStringTag]` getter, it will get executed. This is technically a side-effect, which we don't like, but in this specific case, it has been considered tolerable, as anyone including such a function, will do so with the express intent of modifying the string representation of instances of a given class.
- The RFC for how message templates are evaluated, states that any objects with fields, should have their field count capped at 5 in when stringified. This is not possible to do with `util.inspect`, so for now all fields are included in the string representation. This is an ok compromise for an MVP.


watson:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5548?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5548** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5548?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5548?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5535** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5535?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

packages/dd-trace/src/debugger/devtools_client/index.js
```diff
@@ -155,6 +166,7 @@ session.on('Debugger.paused', async ({ params }) => {
   }
 
   const stack = getStackFromCallFrames(params.callFrames)
+  let i = 0
```
tylfin, line 169: Did this end up getting used?

tylfin (approved):

packages/dd-trace/src/debugger/devtools_client/condition.js
```diff
@@ -35,6 +39,20 @@ const reservedWords = new Set([
...
+  if (segments === undefined) return false // There should always be segments, but just in case
+  for (const { str } of segments) {
+    if (str === undefined) return true
+  }
+  return false
+}
+
+function compileSegments (segments) {
+  return segments
+    .map(({ str, json }) => str !== undefined ? str : `\${${compile(json)}}`)
```
shatzi, line 52: we have different rules for serialization of object in templates. for example, objects here would be [object object]. but we would like to have `{ prop:value, probe:value, ...}` for the first level props.

packages/dd-trace/src/debugger/devtools_client/index.js
```diff
@@ -45,6 +45,7 @@ session.on('Debugger.paused', async ({ params }) => {
   let sampled = false
   let numberOfProbesWithSnapshots = 0
   const probes = []
```
shatzi, line 47: suggest to have a single array of objects to avoid out of sync errors between probes and errors. also enable to avoid two push calls.

`const probesEval = {probe:, eval: }[]`

shatzi (commented):
Awesome progress.

I suggest we fix the spec on how errors are handled before commit.

integration-tests/debugger/basic.spec.js
```diff
@@ -447,6 +447,58 @@ describe('Dynamic Instrumentation', function () {
...
+
+        it('should report evaluation error if template cannot be evaluated', function (done) {
+          t.agent.on('debugger-input', ({ payload: [payload] }) => {
+            assert.strictEqual(payload.message, '')
+
+            const evaluationErrors = payload.debugger.snapshot.evaluationErrors
+
+            assert.isArray(evaluationErrors)
+            assert.strictEqual(evaluationErrors.length, 1)
+            assert.strictEqual(evaluationErrors[0].expr, 'Hello {request.invalid.name}!')
```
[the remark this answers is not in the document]
watson: I've updated the output now to match what the Java tracer does:

Given this input:
```js
{
  template: 'Hello {request.invalid.name}!',
  segments: [
    { str: 'This should fail: ' },
    {
      dsl: 'request.invalid.name',
      json: { getmember: [{ getmember: [{ ref: 'request' }, 'invalid'] }, 'name'] }
    },
    { str: ', this should work: ' },
    {
      dsl: 'request.params.name',
      json: { getmember: [{ getmember: [{ ref: 'request' }, 'params'] }, 'name'] }
    },
    { str: ', and this should fail: ' },
    {
      dsl: 'invalid',
      json: { ref: 'invalid' }
    }
  ]
}
```

It generates this payload
```
{
  message: 'This should fail: {TypeError: Cannot convert undefined or null to object}, this should work: bar, and this should fail: {ReferenceError: invalid is not defined}'
  debugger: {
    snapshot: {
      evaluationErrors: [
        { expr: 'request.invalid.name', message: 'TypeError: Cannot convert undefined or null to object' },
        { expr: 'invalid', message: 'ReferenceError: invalid is not defined' }
      ]
    }
  }
}
```

integration-tests/debugger/basic.spec.js
```diff
@@ -447,6 +447,58 @@ describe('Dynamic Instrumentation', function () {
 
         t.agent.addRemoteConfig(t.rcConfig)
       })
+
+      describe('template evaluation', function () {
```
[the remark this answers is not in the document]
watson: Done 👍

integration-tests/debugger/basic.spec.js
```diff
@@ -447,6 +447,58 @@ describe('Dynamic Instrumentation', function () {
...
+              { str: '!' }
+            ]
+          }))
+
+          t.triggerBreakpoint()
+        })
+
+        it('should report evaluation error if template cannot be evaluated', function (done) {
+          t.agent.on('debugger-input', ({ payload: [payload] }) => {
+            assert.strictEqual(payload.message, '')
```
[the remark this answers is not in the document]
watson: Fixed 👍

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: Not sure I understand? I assume `str` will always be a string and I assume `json` will always be the EL AST?

packages/dd-trace/src/debugger/devtools_client/index.js
```diff
@@ -107,9 +99,22 @@ session.on('Debugger.paused', async ({ params }) => {
         maxLength = highestOrUndefined(probe.capture.maxLength, maxLength)
       }
 
+      if (shouldVerifyConditions && probe.condition !== undefined) {
+        const { result } = await session.post('Debugger.evaluateOnCallFrame', {
+          callFrameId: params.callFrames[0].callFrameId,
+          expression: probe.condition,
+          returnByValue: true
+        })
+        if (result.value !== true) continue
```
shatzi, line 108: can we have a condition that fails with exception?

the spec should generate an evaluation errors every X minutes - so the user can fix those errors.
I think at the current state failed condition simply stop all probes from been trigger, is that right?

packages/dd-trace/src/debugger/devtools_client/index.js
shatzi: yeah. I think it is ok and now that we also have getDDTagsExpression in there my suggestion will not work.

it is added complexity but I think it is handleable.

shatzi (dismissed):

packages/dd-trace/src/debugger/devtools_client/index.js
watson: Yes, that's an oversight from me when I created the condition logic. We should report any errors that might occur here. I'll do that in a follow up PR since it's nothing to do with log templates

packages/dd-trace/src/debugger/devtools_client/condition.js
watson: Update: This was discussed on Slack and the issue is as follows: If someone writes a template where the evaluation of a segment doesn't returns an object instead for example a string or a number, that object with be stringified and the result in the log message will just be `[object Object]` (which is what you get if you evaluate `${{}}` inside of a template string).

We have decided not to deal with that issue in this PR, but make a follow up PR that properly stringifies these objects. Most likely using `util.inspect`.

watson: Update: I've updated the PR with support for this anyway. It's all pushed now 👍

shatzi (approved):

packages/dd-trace/src/debugger/devtools_client/index.js
```diff
@@ -179,32 +214,40 @@ session.on('Debugger.paused', async ({ params }) => {
...
+          if (typeof result === 'string') {
+            message += result
+          } else {
+            // If `result` isn't a string, it's an evaluation error object
+            if (snapshot.evaluationErrors === undefined) {
+              snapshot.evaluationErrors = [result]
+            } else {
+              snapshot.evaluationErrors.push(result)
+            }
+            message += `{${result.message}}`
```
BridgeAR, line 233: Should that maybe use the stack information instead?

integration-tests/debugger/basic.spec.js
```diff
@@ -577,6 +577,8 @@ describe('Dynamic Instrumentation', function () {
             if (diagnostics.status === 'ERROR') {
               assert.strictEqual(diagnostics.exception.message, 'Cannot compile expression: original dsl')
               done()
+            } else if (diagnostics.status === 'INSTALLED') {
+              assert.fail('Should not install when condition cannot be compiled')
```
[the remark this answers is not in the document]
watson: `assert.fail` will immediately end the test without having to call `done` or waiting for a timeout. I added it here as an escape hatch in case somehow `describe.status` was unexpectedly `INSTALLED` which should never happen in this test.

It has technically nothing to do with this PR, but while I was working on other tests during the development in this PR I noticed I could make this improvement (those other tests was a copy of this test, but has since been refactored, so that's why this single change remains in this PR)

packages/dd-trace/src/debugger/devtools_client/index.js
```diff
@@ -179,32 +214,40 @@ session.on('Debugger.paused', async ({ params }) => {
...
+        for (const result of results) {
+          if (typeof result === 'string') {
+            message += result
+          } else {
+            // If `result` isn't a string, it's an evaluation error object
+            if (snapshot.evaluationErrors === undefined) {
+              snapshot.evaluationErrors = [result]
+            } else {
+              snapshot.evaluationErrors.push(result)
+            }
```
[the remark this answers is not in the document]
watson: The `evaluationErrors` are shown in the UI to the user who wrote the message template and is there to allow the user to debug it.

packages/dd-trace/src/debugger/devtools_client/index.js
watson: The stack doesn't provide any meaningful information to the end user (it's just internal frames). Showing the error message only in curly brases is the standard we have used across all tracing languages, so here I'm just aligning with the rest of the tracers.

packages/dd-trace/src/debugger/devtools_client/index.js
```diff
@@ -107,9 +109,24 @@ session.on('Debugger.paused', async ({ params }) => {
...
       }
 
+      if (shouldVerifyConditions && probe.condition !== undefined) {
+        // TODO: Bundle all conditions and evaluate them in a single call
+        // TODO: Handle errors
+        const { result } = await session.post('Debugger.evaluateOnCallFrame', {
+          callFrameId: params.callFrames[0].callFrameId,
+          expression: probe.condition,
+          returnByValue: true
+        })
```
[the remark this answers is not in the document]
watson: Good idea. I'll do that in a later PR as the condition evaluation has nothing to do with this PR. This diff just appeared in this PR because I moved the code a few lines down from where it was before and added the TODO's so that I could remember to deal with this later.

integration-tests/debugger/basic.spec.js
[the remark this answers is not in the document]
BridgeAR: I guess the test runner does some magic in this case to detect it? Or is this code sync?

BridgeAR (approved):
LGTM

integration-tests/debugger/basic.spec.js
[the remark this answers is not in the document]
watson: No the code is async. Yeah, I assume it does so as well. I just tested it to make sure.

PR #5549, opened 2025-04-08
Fix profiler test flakiness

szegedi:
### What does this PR do?
Adds retries for execution of the prop program for the profiler integration test for code hotspots if its execution gathered less than 60% of expected samples. Adaptively adjusts the run duration of the prop to achieve the desired sample count.

### Motivation
It's flaky. 

### Additional Notes
The profiler integration test for code hotspots will execute a test program `codehotspots.js` for about 900ms (the test program creates 9 spans, each then runs a busy loop for about 101ms) while profiling it. Under normal conditions, the profiler would capture 90 samples during 900ms at its normal sampling rate of 99Hz. However various factors – most likely, CPU saturation – can cause the V8 thread that triggers PROF signals to be CPU starved, leading to a profile with too few samples (on flaky failures, we observed as little as 18 samples gathered.) We'll retry execution, adjusting the duration of the test program run until we get a profile with at least 60% of the ideal count, which should be enough for the rest of the test to succeed.

Jira: [PROF-11107]

[PROF-11107]: https://datadoghq.atlassian.net/browse/PROF-11107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integration-tests/profiler/codehotspots.js
+++ b/integration-tests/profiler/codehotspots.js
@@ -1,16 +1,21 @@
 'use strict'
 
 const DDTrace = require('dd-trace')
-
 const tracer = DDTrace.init()
+
+// Busy cycle duration is communicated in nanoseconds through the environment
+// variable by the test. On first execution, it'll be 10 * the sampling period
+// at 99Hz (so, 101010101ns). If subsequent executions are needed, it will be
+// prolonged.
+const busyCycleTime = BigInt(process.env.BUSY_CYCLE_TIME)
 
 function busyLoop () {
   const start = process.hrtime.bigint()
   let x = 0
   for (;;) {
     const now = process.hrtime.bigint()
-    // Busy cycle for 100ms
-    if (now - start > 100000000n) {
+    // Busy cycle
+    if (now - start > busyCycleTime) {
       break
     }
     // Do something in addition to invoking hrtime
--- a/integration-tests/profiler/profiler.spec.js
+++ b/integration-tests/profiler/profiler.spec.js
@@ -296,7 +296,6 @@ describe('profiler', () => {
   // Computed values for the code hotspots test. busyCycleTimeNs is adaptively adjusted by the test
   // when it needs to be repeated.
   const idealSampleCount = idealSamplesPerSpan * expectedSpans // we'd like 10 samples per span, ideally
-  const acceptedSampleCount = (idealSampleCount * 0.6) | 0 // we'll settle for 60% of that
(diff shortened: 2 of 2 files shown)


BridgeAR (commented):
Oof... a 60% in 1 out of 10 is a pretty bad expected output to reach 99.n % passing.
I get that this is about making sure the test passes but I just wonder if the feature is really working as intended, if we have such a big data loss in average.

rochdev:
My main issue with retries is that it does not fix the flakiness, it only reduces it. Also, in my experience, when you get a really bad machine, it will be bad for the entire duration of the workflow, so it's possible all retries will still be super slow. We have definitely experienced this in the past where something that usually takes 20-30 seconds times out at 2 minutes for example. Is there any way to make do with less samples instead?

BridgeAR (approved):
I just discussed it briefly with @szegedi and since there is no ideal solution, I suggest to give this a try for a bit. If it does show up as flaky anytime soon, @szegedi promised to have another look and improve it further.

szegedi:
> Oof... a 60% in 1 out of 10 is a pretty bad expected output to reach 99.n % passing.

> Is there any way to make do with less samples instead?

You're both right! My original thinking was that on my laptop I'll usually get about 75 samples in 1s which is 83% of the maximum number of samples collectable (90.) With 60% I basically say "the test should still pass even if we only collect 60% of the theoretically maximum number of samples".

But there's no need for us to establish a minimum! In ideal case, as few as 9 samples can be enough if they're the right 9 samples – one hitting each span's execution. It certainly seems that if we only collect ~20 samples (saw this previously) then the test fails, as the profiler won't adequately hit every span. But we needn't decide that upfront – just run the entirety of the test, and if it fails, re-run the test program with the recomputed longer duration.

BRB, I'll implement this.

> My main issue with retries is that it does not fix the flakiness, it only reduces it. 

This is a fair point, but at the end of the day, the profiler is statistically collecting samples and there will always be a very low but nonzero possibility that an execution won't collect the samples we wanted to – it can always hit in GC, or while a V8 intrinsified operation is running (those are invoked with no associated `v8::Context`). We can't even direct it _when_ to sample, but even if we could, we wouldn't want to as then we'd be biasing it. That's why the prop program has longish running busy loops – to give the profiler more chances to collect relevant samples. What I'm saying is that from the very nature of the profiler I don't see a way to make this test entirely deterministic.

BridgeAR (approved):

PR #5550, opened 2025-04-08
[test-optimization] [SDTEST-1830] Add @test.retry_reason to the retries of different frameworks

Mariovido:
### What does this PR do?

This PR adds `@test.retry_reason` tag to all retries. The new type of retries are the following ones:
- `early_flake_detection` previously (`efd`): The retry is from Early Flake Detection
- `auto_test_retry`: The retry is from Auto Test Retries
- `attempt_to_fix`: The retry is from Test Management (Attempt to Fix)
- `external`: The retry is from an unknown source

The frameworks which tags are changed are the following ones:
- **Jest**
- **Mocha**
- **Cucumber**
- **Playwright**
- **Cypress**
- **Vitest**

### Motivation

Previously we were missing the tags in the retries and it was hard to identify which type of retry it was. Specially for the cases where the retry was done by the testing framework.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/datadog-plugin-cucumber/src/index.js
+++ b/packages/datadog-plugin-cucumber/src/index.js
@@ -260,7 +260,7 @@ class CucumberPlugin extends CiPlugin {
         if (isAtrRetry) {
           span.setTag(TEST_RETRY_REASON, 'auto_test_retry')
         } else {
-          span.setTag(TEST_RETRY_REASON, 'unknown')
+          span.setTag(TEST_RETRY_REASON, 'external')
         }
       }
       span.setTag('error', error)
--- a/packages/datadog-plugin-cucumber/src/index.js
+++ b/packages/datadog-plugin-cucumber/src/index.js
@@ -260,7 +260,7 @@ class CucumberPlugin extends CiPlugin {
         if (isAtrRetry) {
           span.setTag(TEST_RETRY_REASON, 'atr')
         } else {
-          span.setTag(TEST_RETRY_REASON, 'native_retry')
+          span.setTag(TEST_RETRY_REASON, 'unknown')
         }
       }
       span.setTag('error', error)
--- a/packages/datadog-plugin-cucumber/src/index.js
+++ b/packages/datadog-plugin-cucumber/src/index.js
@@ -252,11 +252,16 @@ class CucumberPlugin extends CiPlugin {
       }
     })
 
-    this.addSub('ci:cucumber:test:retry', ({ isFirstAttempt, error }) => {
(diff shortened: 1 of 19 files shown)


packages/datadog-plugin-cucumber/src/index.js
```diff
@@ -252,11 +252,16 @@ class CucumberPlugin extends CiPlugin {
...
-    this.addSub('ci:cucumber:test:retry', ({ isFirstAttempt, error }) => {
+    this.addSub('ci:cucumber:test:retry', ({ isFirstAttempt, error, isAtrRetry }) => {
       const store = storage('legacy').getStore()
       const span = store.span
       if (!isFirstAttempt) {
         span.setTag(TEST_IS_RETRY, 'true')
+        if (isAtrRetry) {
+          span.setTag(TEST_RETRY_REASON, 'auto_test_retry')
+        } else {
+          span.setTag(TEST_RETRY_REASON, 'unknown')
```
anmarchenko, line 263: on the meeting yesterday we agreed on "external" to make it sound better :) sorry, our naming discussions are always extremely confusing

packages/datadog-plugin-cypress/src/cypress-plugin.js
```diff
@@ -654,10 +656,18 @@ class CypressPlugin {
...
+            !finishedTest.isEfdRetry
           if (attemptIndex > 0) {
             finishedTest.testSpan.setTag(TEST_IS_RETRY, 'true')
             if (finishedTest.isEfdRetry) {
-              finishedTest.testSpan.setTag(TEST_RETRY_REASON, 'efd')
+              finishedTest.testSpan.setTag(TEST_RETRY_REASON, 'early_flake_detection')
+            } else if (isAtrRetry) {
+              finishedTest.testSpan.setTag(TEST_RETRY_REASON, 'auto_test_retry')
+            } else {
+              finishedTest.testSpan.setTag(TEST_RETRY_REASON, 'unknown')
```
anmarchenko, line 670: also 'external' here (maybe better to use global constants for all reasons?)

packages/datadog-plugin-cucumber/src/index.js
Mariovido: You're right I didn't pushed yesterday my latest commit doing that. Good catch :)

packages/datadog-plugin-cypress/src/cypress-plugin.js
Mariovido: I changed to use a global constant as probably it will be more suitable in case of future renamings, or additions.

anmarchenko (approved):

PR #5555, opened 2025-04-09
instrument the iovalkey package

tlhunter:
### What does this PR do?
- blesses #5484, closes #5484

### Plugin Checklist

- [x] Unit tests.
- [x] TypeScript [tests][2].
- [x] API [documentation][3].


packages/datadog-instrumentations/src/iovalkey.js
```diff
@@ -0,0 +1,51 @@
...
+    const options = this.options || {}
+    const connectionName = options.connectionName
+    const db = options.db
+    const connectionOptions = { host: options.host, port: options.port }
+
+    const asyncResource = new AsyncResource('bound-anonymous-fn')
+    return asyncResource.runInAsyncScope(() => {
+      startCh.publish({ db, command: command.name, args: command.args, connectionOptions, connectionName })
+
+      const onResolve = asyncResource.bind(() => finish(finishCh, errorCh))
```
BridgeAR, line 29:
```suggestion
      const onResolve = asyncResource.bind(() => finishCh.publish())
```

packages/datadog-plugin-iovalkey/test/index.spec.js
```diff
@@ -0,0 +1,184 @@
...
+              expect(traces[0][0].meta).to.have.property('db.type', 'redis')
+              expect(traces[0][0].meta).to.have.property('span.kind', 'client')
+              expect(traces[0][0].meta).to.have.property('out.host', 'localhost')
+              expect(traces[0][0].meta).to.have.property('valkey.raw_command', 'GET foo')
+              expect(traces[0][0].metrics).to.have.property('network.destination.port', 6379)
+            })
+            .then(done)
+            .catch(done)
+
+          redis.get('foo').catch(done)
```
BridgeAR, line 51: Opinionated nit: I personally find promises easier to read.
Due to the extra work, I'd probably stick to as it is, but it's a bit of a chicken / egg problem: new tests orient themselves upon existing ones.

```suggestion
        it('should do automatic instrumentation when using callbacks', async () => {
          agent.use(() => {}) // wait for initial info command
          const promise = agent
            .use(traces => {
              expect(traces[0][0]).to.have.property('name', expectedSchema.outbound.opName)
              expect(traces[0][0]).to.have.property('service', expectedSchema.outbound.serviceName)
              expect(traces[0][0]).to.have.property('resource', 'get')
              expect(traces[0][0]).to.have.property('type', 'redis')
              expect(traces[0][0].meta).to.have.property('component', 'iovalkey')
              expect(traces[0][0].meta).to.have.property('db.name', '0')
              expect(traces[0][0].meta).to.have.property('db.type', 'redis')
              expect(traces[0][0].meta).to.have.property('span.kind', 'client')
              expect(traces[0][0].meta).to.have.property('out.host', 'localhost')
              expect(traces[0][0].meta).to.have.property('valkey.raw_command', 'GET foo')
              expect(traces[0][0].metrics).to.have.property('network.destination.port', 6379)
            })

          return Promise.all([
            redis.get('foo'),
            promise
          ])
```

BridgeAR (approved):
LGTM!

packages/datadog-instrumentations/src/iovalkey.js
```diff
@@ -0,0 +1,51 @@
...
+  addHook,
+  AsyncResource
+} = require('./helpers/instrument')
+const shimmer = require('../../datadog-shimmer')
+
+const startCh = channel('apm:iovalkey:command:start')
+const finishCh = channel('apm:iovalkey:command:finish')
+const errorCh = channel('apm:iovalkey:command:error')
+
+addHook({ name: 'iovalkey', versions: ['>=0'] }, Redis => {
```
BridgeAR, line 14:
```suggestion
addHook({ name: 'iovalkey', versions: ['>=0.0.1'] }, Redis => {
```

I believe that makes the tests pass as well?

packages/datadog-instrumentations/src/helpers/hooks.js
```diff
@@ -61,6 +61,7 @@ module.exports = {
   http2: () => require('../http2'),
   https: () => require('../http'),
   ioredis: () => require('../ioredis'),
+  iovalkey: () => require('../valkey'),
```
BridgeAR, line 64:
```suggestion
  iovalkey: () => require('../iovalkey'),
```

BridgeAR (commented):
The tests don't complain anymore that there is no version 0.0.0 but files can't be found now. I did not check if there are any other issues in addition.

packages/datadog-plugin-iovalkey/test/integration-test/server.mjs
```diff
@@ -0,0 +1,6 @@
+import 'dd-trace/init.js'
+import Redis from 'iovalkey'
+
+const client = new Redis({ connectionName: 'test' })
```
BridgeAR, line 4:
```suggestion
import Valkey from 'iovalkey'

const client = new Valkey({ connectionName: 'test' })
```

packages/datadog-plugin-iovalkey/test/index.spec.js
```diff
@@ -0,0 +1,184 @@
...
+  let Redis
+  let redis
+  let tracer
+
+  describe('iovalkey', () => {
+    withVersions('iovalkey', 'iovalkey', version => {
+      beforeEach(() => {
+        tracer = require('../../dd-trace')
+        Redis = require(`../../../versions/iovalkey@${version}`).get()
+        redis = new Redis({ connectionName: 'test' })
```
BridgeAR, line 19:
```suggestion
  let Valkey
  let redis
  let tracer

  describe('iovalkey', () => {
    withVersions('iovalkey', 'iovalkey', version => {
      beforeEach(() => {
        tracer = require('../../dd-trace')
        Valkey = require(`../../../versions/iovalkey@${version}`).get()
        redis = new Valkey({ connectionName: 'test' })
```

Ideally the client instance is also renamed.

packages/dd-trace/src/service-naming/schemas/v1/storage.js
```diff
@@ -38,6 +38,7 @@ const storage = {
       serviceName: configWithFallback
     },
     ioredis: redisNaming,
+    iovalkey: redisNaming,
```
BridgeAR, line 41:
```suggestion
    iovalkey: {
      opName: () => 'valkey.command',
      serviceName: configWithFallback
    },
```

packages/dd-trace/src/service-naming/schemas/v0/storage.js
```diff
@@ -57,6 +57,7 @@ const storage = {
         pluginConfig.service || `${tracerService}-elasticsearch`
     },
     ioredis: redisConfig,
+    iovalkey: redisConfig,
```
BridgeAR, line 60: This has to also be replaced. The command is not correct.
We could just make it a method and accept a name as parameter that defaults to redis.

packages/datadog-plugin-iovalkey/test/naming.js
```diff
@@ -0,0 +1,19 @@
+const { resolveNaming } = require('../../dd-trace/test/plugins/helpers')
+
+const rawExpectedSchema = {
+  outbound: {
+    v0: {
+      opName: 'valkey.command',
+      serviceName: 'test-redis'
```
BridgeAR, line 7:
```suggestion
      serviceName: 'test-valkey'
```

packages/datadog-instrumentations/src/iovalkey.js
```diff
@@ -0,0 +1,51 @@
...
+  AsyncResource
+} = require('./helpers/instrument')
+const shimmer = require('../../datadog-shimmer')
+
+const startCh = channel('apm:iovalkey:command:start')
+const finishCh = channel('apm:iovalkey:command:finish')
+const errorCh = channel('apm:iovalkey:command:error')
+
+addHook({ name: 'iovalkey', versions: ['>=0.0.1'] }, Redis => {
+  shimmer.wrap(Redis.prototype, 'sendCommand', sendCommand => function (command, stream) {
```
BridgeAR, line 15:
```suggestion
addHook({ name: 'iovalkey', versions: ['>=0.0.1'] }, Valkey => {
  shimmer.wrap(Valkey.prototype, 'sendCommand', sendCommand => function (command, stream) {
```

packages/datadog-instrumentations/src/iovalkey.js
```diff
@@ -0,0 +1,51 @@
...
+    if (!startCh.hasSubscribers) return sendCommand.apply(this, arguments)
+
+    if (!command || !command.promise) return sendCommand.apply(this, arguments)
```
BridgeAR, line 18:
```suggestion
    if (!command?.promise) return sendCommand.apply(this, arguments)
```

packages/datadog-instrumentations/src/iovalkey.js
```diff
@@ -0,0 +1,51 @@
...
+      try {
+        return sendCommand.apply(this, arguments)
+      } catch (err) {
+        errorCh.publish(err)
+
+        throw err
+      }
+    })
+  })
+  return Redis
```
BridgeAR, line 43:
```suggestion
  return Valkey
```

packages/datadog-plugin-iovalkey/src/index.js
```diff
@@ -0,0 +1,11 @@
+'use strict'
+
+const RedisPlugin = require('../../datadog-plugin-redis/src')
+
+class IOValkeyPlugin extends RedisPlugin {
```
BridgeAR, line 5: By extending from the Redis plugin, a couple of things are hardcoded to `redis`. This is probably not ideal or should we just accept it being reported as redis? That might be confusing long term?

I guess we could start with it and change it later on as well.

packages/datadog-plugin-iovalkey/test/index.spec.js
```diff
@@ -0,0 +1,184 @@
...
+        })
+
+        it('should work with userland promises', done => {
+          agent.use(() => {}) // wait for initial info command
+          agent
+            .use(traces => {
+              expect(traces[0][0]).to.have.property('name', expectedSchema.outbound.opName)
+              expect(traces[0][0]).to.have.property('service', expectedSchema.outbound.serviceName)
+              expect(traces[0][0]).to.have.property('resource', 'get')
+              expect(traces[0][0]).to.have.property('type', 'redis')
```
BridgeAR, line 93: This is a hardcoded value in the redis plugin.

packages/datadog-plugin-iovalkey/test/index.spec.js
```diff
@@ -0,0 +1,184 @@
...
+              expect(traces[0][0].meta).to.have.property('db.name', '0')
+              expect(traces[0][0].meta).to.have.property('db.type', 'redis')
```
BridgeAR, line 95: This is a hardcoded value in the redis plugin.

BridgeAR (commented):
Now that the tests run regularly, I had another look and the main issue is around Valkey vs Redis. I think it would be good switch these to become valkey ones.

BridgeAR (commented):
This is already looking very good and I think it's almost done! Just some tests are still failing and they need a closer look.

packages/datadog-plugin-iovalkey/test/index.spec.js
```diff
@@ -0,0 +1,194 @@
...
+          let run = 0
+          agent.use(traces => {
+            const expectedResource = runs[run++]
+            expect(traces[0][0]).to.have.property('resource', expectedResource)
+            if (expectedResource !== 'client') return
+
+            expect(traces[0][0]).to.have.property('name', expectedSchema.outbound.opName)
+            expect(traces[0][0]).to.have.property('service', expectedSchema.outbound.serviceName)
+            expect(traces[0][0]).to.have.property('resource', 'get')
+            expect(traces[0][0]).to.have.property('type', 'redis')
```
tlhunter, line 44: I think the last important change is that type and db.type should be set to valkey instead of redis.

packages/datadog-plugin-iovalkey/test/index.spec.js
```diff
@@ -0,0 +1,185 @@
...
+          agent
+            .use(traces => {
+              expect(traces[0][0]).to.have.property('error', 1)
+              expect(traces[0][0].meta).to.have.property(ERROR_TYPE, error.name)
+              expect(traces[0][0].meta).to.have.property(ERROR_MESSAGE, error.message)
+              expect(traces[0][0].meta).to.have.property(ERROR_STACK, error.stack)
+              expect(traces[0][0].meta).to.have.property('component', 'iovalkey')
+            })
+            .then(done)
+            .catch(done)
```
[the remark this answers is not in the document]
tlhunter: 90% of tests are copypasta from other tests. If we can mass-modify the existing tests to consistently use async/await then we almost guarantee all future tests will do the same.

tlhunter:
Getting some new failures in CI that I don't get locally:

```
      with @redis/client >=1.1 (5.1.0)
        without configuration
          1) "before each" hook for "should do automatic instrumentation when using callbacks"
          2) "after each" hook for "should do automatic instrumentation when using callbacks"
        with configuration
          3) "before each" hook for "should be configured with the correct values"
          4) "after each" hook for "should be configured with the correct values"
        with blocklist
          5) "before each" hook for "should be able to filter commands on a case-insensitive basis"
          6) "after each" hook for "should be able to filter commands on a case-insensitive basis"
```

Might be related to running multiple redis containers on the same port? And a Redis library sending what's now the wrong command?

tlhunter:
Maybe it's a real compat bug:
https://www.npmjs.com/package/redis?activeTab=versions
```
5.1.0	0	33 minutes ago
```

.github/workflows/plugins.yml
```diff
@@ -1023,16 +1023,44 @@ jobs:
...
+  ioredis:
+    runs-on: ubuntu-latest
+    services:
+      redis:
+        image: redis:6.2-alpine
+        ports:
+          - 6379:6379
+    env:
+      PLUGINS: ioredis
+      SERVICES: ioredis
```
IlyasShabi, line 1045: I think this should be 
```suggestion
      SERVICES: redis
```

BridgeAR (approved):
LGTM! Thanks for the great addition :)

.github/workflows/plugins.yml
BridgeAR: I think it's actually nice to have it as separate CI job. That way we can see if a single implementation fails or not :)

PR #5558, opened 2025-04-10
Fix MongoDB heartbeat filtering when DD_TRACE_MONGODB_HEARTBEAT_ENABL…

aitrics-ray:
…ED is set to 'false'

### What does this PR do?

This PR fixes an issue where MongoDB heartbeat traces were still being collected even when the environment variable `DD_TRACE_MONGODB_HEARTBEAT_ENABLED` was set to 'false'. The fix properly converts the string 'false' to a boolean value to ensure the heartbeat filtering works as expected.

### Motivation

While testing MongoDB tracing, we discovered that setting `DD_TRACE_MONGODB_HEARTBEAT_ENABLED=false` didn't actually disable heartbeat traces. This was because the string value 'false' is truthy in JavaScript, causing the heartbeat filtering condition to fail. This fix ensures that MongoDB monitoring heartbeats can be properly excluded from traces when users want to reduce trace volume.
image

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/datadog-plugin-mongodb-core/src/index.js
+++ b/packages/datadog-plugin-mongodb-core/src/index.js
@@ -1,8 +1,8 @@
 'use strict'
 
+const { isTrue } = require('../../dd-trace/src/util')
 const DatabasePlugin = require('../../dd-trace/src/plugins/database')
 const coalesce = require('koalas')
-const { isTrue } = require('../../dd-trace/src/util')
 
 class MongodbCorePlugin extends DatabasePlugin {
   static get id () { return 'mongodb-core' }
--- a/packages/datadog-plugin-mongodb-core/src/index.js
+++ b/packages/datadog-plugin-mongodb-core/src/index.js
@@ -14,7 +14,7 @@ class MongodbCorePlugin extends DatabasePlugin {
     super.configure(config)
     this.config.heartbeatEnabled = coalesce(
       config.heartbeatEnabled,
-      process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED,
+      process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED === 'false' ? false : null,
       true
     )
   }
--- a/packages/datadog-plugin-mongodb-core/src/index.js
+++ b/packages/datadog-plugin-mongodb-core/src/index.js
@@ -2,6 +2,7 @@
 
 const DatabasePlugin = require('../../dd-trace/src/plugins/database')
 const coalesce = require('koalas')
+const { isTrue } = require('../../dd-trace/src/util')
 
 class MongodbCorePlugin extends DatabasePlugin {
   static get id () { return 'mongodb-core' }
@@ -14,7 +15,7 @@ class MongodbCorePlugin extends DatabasePlugin {
     super.configure(config)
     this.config.heartbeatEnabled = coalesce(
       config.heartbeatEnabled,
-      process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED === 'false' ? false : null,
+      isTrue(process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED),
       true
     )
   }
--- a/packages/datadog-plugin-mongodb-core/src/index.js
+++ b/packages/datadog-plugin-mongodb-core/src/index.js
@@ -13,9 +13,12 @@ class MongodbCorePlugin extends DatabasePlugin {
 
   configure (config) {
     super.configure(config)
+
+    const heartbeatFromEnv = process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED
+
     this.config.heartbeatEnabled = coalesce(
       config.heartbeatEnabled,
-      isTrue(process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED),
+      heartbeatFromEnv && isTrue(heartbeatFromEnv),
       true
     )
   }


aitrics-ray:
This is a bug fix. Could someone with write access please add the semver-patch label to this PR? Thank you!

packages/datadog-plugin-mongodb-core/src/index.js
```diff
@@ -14,7 +14,7 @@ class MongodbCorePlugin extends DatabasePlugin {
     super.configure(config)
     this.config.heartbeatEnabled = coalesce(
       config.heartbeatEnabled,
-      process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED,
+      process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED === 'false' ? false : null,
       true
```
simon-id, line 18: you'll also need to require at the top of the file: ``const { isTrue } = require('../../dd-trace/src/util)
```suggestion
      isTrue(process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED)
```

simon-id:
Thank you for the PR, it indeed looks like we've made a mistake here. We'll do a first draft here, and once we agree on the changes, our team will recreate this PR, as the CI is not allowed to run on external contributions.

packages/datadog-plugin-mongodb-core/src/index.js
aitrics-ray: https://github.com/DataDog/dd-trace-js/pull/5558/commits/ee518c73f1e442163585f3983b25611e615b6a2a

packages/datadog-plugin-mongodb-core/src/index.js
```diff
@@ -14,7 +15,7 @@ class MongodbCorePlugin extends DatabasePlugin {
     super.configure(config)
     this.config.heartbeatEnabled = coalesce(
       config.heartbeatEnabled,
-      process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED,
+      isTrue(process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED),
```
simon-id, line 18:
```suggestion
      process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED && isTrue(process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED),
```

simon-id: sorry forgot something

simon-id:
I'd say we should also write a test for this configuration, as we're clearly missing coverage if we didn't realize the feature was not working correctly. @tlhunter

packages/datadog-plugin-mongodb-core/src/index.js
BridgeAR: I don't believe that is necessary. If it's undefined, it is just false. So that's fine, even without the extra check, no?

aitrics-ray: The expression
```
process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED && isTrue(process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED)
```
hurts readability a bit. What do you think about pulling it out into a separate variable instead?

For example:
```
const heartbeatFromEnv = process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED !== undefined 
  ? isTrue(process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED) 
  : undefined;

this.config.heartbeatEnabled = coalesce(
  config.heartbeatEnabled,
  heartbeatFromEnv,
  true
);
```

It feels cleaner and easier to follow.

aitrics-ray: @BridgeAR If we go with `isTrue(process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED)`, the final fallback value (true) in the coalesce call will never be used, since isTrue treats even undefined as false.

I’m not sure the original author intended for it to behave this way.

simon-id: @BridgeAR If it's undefined, isTrue() will return false, and coalesce() will thus pick it instead of the default value "true". So if the env var is not defined, it will default to false instead of true. This is why you need the "not empty" check first.

simon-id: @aitrics-ray I'm totally fine with this, but please change the condition from `process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED !== undefined ? ...` to `process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED ? ...`. This is to handle a potentially empty env var, which we consider as undefined, but is sent by node as an empty string.

simon-id: Another alternative that would reflect the way we write those in the rest of our codebase:
```javascript
const heartbeatFromEnv = process.env.DD_TRACE_MONGODB_HEARTBEAT_ENABLED

this.config.heartbeatEnabled = coalesce(
  config.heartbeatEnabled,
  heartbeatFromEnv && isTrue(heartbeatFromEnv),
  true
)
```

aitrics-ray: https://github.com/DataDog/dd-trace-js/pull/5558/commits/4de4cfab708393d93e844ef7346a766d55933638

simon-id:
ok fix is LGTM now, we should take over the rest and probably add a test

simon-id:
Cherrypicked your commits in https://github.com/DataDog/dd-trace-js/pull/5562, let's move the discussion there

PR #5572, opened 2025-04-12
[DI] Add debug log when translating breakpoint location using source map

watson:
### What does this PR do?

This PR adds a debug statement just before translating a probe location using source maps. It logs the original file and line number before the translation.

This PR also reduces overhead slightly when translating the location using source map, by not awaiting the result of the translation, but simply returning the promise. The caller will await anyway, so the extra await is just creating extra promises in memory which we don't need.

### Motivation

When debugging an issue with a probe, it's useful to know what the original file/line was before being translated by the source map. It's also the only indication we have in the logs that a source map was ever involved, which is even more important in a debugging scenario (in case we want to know if source maps was deployed together with the application or not).

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
+++ b/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
@@ -48,6 +48,10 @@ async function addBreakpoint (probe) {
   const { url, scriptId, sourceMapURL, source } = script
 
   if (sourceMapURL) {
+    log.debug(
+      '[debugger:devtools_client] Translating location using source map for %s:%d:%d (probe: %s, version: %d)',
+      file, lineNumber, columnNumber, probe.id, probe.version
+    );
     ({ line: lineNumber, column: columnNumber } = await getGeneratedPosition(url, source, lineNumber, sourceMapURL))
   }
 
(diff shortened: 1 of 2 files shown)


watson:
* **#5572** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5572?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5572?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/dd-trace/src/debugger/devtools_client/breakpoints.js
```diff
@@ -48,6 +48,10 @@ async function addBreakpoint (probe) {
   const { url, scriptId, sourceMapURL, source } = script
 
   if (sourceMapURL) {
+    log.debug(
+      '[debugger:devtools_client] Translating location using source map for %s:%d:%d (probe: %s, version: %d)',
+      file, lineNumber, columnNumber, probe.id, probe.version
+    );
```
simon-id, line 54: oopsie ?
```suggestion
    )
```

simon-id: or wait, is it just required because next line is a parenthesis ? so the ASI fails ?

watson: Correct. I was contemplating moving to the beginning of the next line to make it more clear though. Do you want me to do that?

BridgeAR (approved):

PR #5578, opened 2025-04-15
switch from default token to app token for release proposal

rochdev:
### What does this PR do?

Switch from default token to app token for release proposal.

### Motivation

This is required for a workflow triggered by the `workflow_dispatch` event to trigger other types of workflow, otherwise no workflow will run, for example `on: push` or `on: pull_request`.

Reference: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#triggering-a-workflow-from-a-workflow

### Additional Notes

I kept the default token with read-only permissions for `branch-diff` since it doesn't need to write.

### DIFF
--- a/.github/workflows/release-proposal.yml
+++ b/.github/workflows/release-proposal.yml
@@ -24,20 +24,26 @@ jobs:
   create-proposal:
     runs-on: ubuntu-latest
     permissions:
-      contents: write
-      pull-requests: write
-    env:
-      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      contents: read
+      pull-requests: read
     steps:
+      - uses: actions/create-github-app-token@3ff1caaa28b64c9cc276ce0a02e2ff584f3900c5 # v2.0.2
+        id: app-token
+        with:
+          app-id: ${{ vars.APP_ID }}
+          private-key: ${{ secrets.PRIVATE_KEY }}
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
+          token: ${{ steps.app-token.outputs.token }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-proposal.yml
```diff
@@ -24,20 +24,26 @@ jobs:
...
-          echo "{\"token\":\"${{secrets.GITHUB_TOKEN}}\",\"user\":\"${{github.actor}}\"}" > ~/.config/changelog-maker/config.json
+          echo "{\"token\":\"${{ secrets.GITHUB_TOKEN }}\",\"user\":\"${{ github.actor }}\"}" > ~/.config/changelog-maker/config.json
           git config user.name ${{ github.actor }}
           git config user.email ${{ github.actor }}@users.noreply.github.com
-      - run: node scripts/release/proposal 5 -y --{{ inputs.increment }}
-        if: inputs.release-line == 'all' || inputs.release-line == '5'
+      - if: inputs.release-line == 'all' || inputs.release-line == '5'
+        run: node scripts/release/proposal 5 -y --{{ inputs.increment }}
+        env:
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
```
uurien, line 49: The environment variable before this PR was `GH_TOKEN` and now you are setting `GITHUB_TOKEN`, is expected?

https://github.com/DataDog/dd-trace-js/pull/5578/files#diff-e9ea1b20155b0c09b80c4162da05df3b2ef18a2225da9c05a2dce07696823295L30

rochdev: Yes. This environment variable is almost always often referred to as `GITHUB_TOKEN` and both are supported by the script. The only reason that `GH_TOKEN` was previously used is because that's what's used in the example for passing the secret as an environment variable.

Issue #5528, opened 2025-04-03
[BUG]: dd-trace with log injection removes stack traces in Node 22


JavaScriptBach:
### Tracer Version(s)

5.45.0

### Node.js Version(s)

22.14.0

### Bug Report

On Node 22, using dd-trace 5.45.0 with log injection breaks winston logs by removing stack traces.

### Reproduction Code

https://github.com/JavaScriptBach/dd-trace-winston-node-22

```javascript
const tracer = require("dd-trace");
tracer.init({
  logInjection: true,
});

const { createLogger, format, transports } = require("winston");

const Logger = createLogger({
  level: "info",
  transports: [new transports.Console()],
  format: format.combine(format.errors({ stack: true }), format.prettyPrint()),
});

Logger.error(new Error("test error with stack"));
```

### Error Logs

On Node 22, this gives:
```
{
  dd: { service: 'dd-trace-winston-node-22', version: '1.0.0' },
  level: 'error',
  message: 'test error with stack',
  stack: undefined
}
```

On Node 20:
```
{
  dd: { service: 'dd-trace-winston-node-22', version: '1.0.0' },
  level: 'error',
  message: 'test error with stack',
  stack: 'Error: test error with stack\n' +
    '    at Object.<anonymous> (/Users/philliphuang/dd-trace-winston-node-22/index.js:17:14)\n' +
    '    at Module._compile (node:internal/modules/cjs/loader:1469:14)\n' +
    '    at Module._extensions..js (node:internal/modules/cjs/loader:1548:10)\n' +
    '    at Module.load (node:internal/modules/cjs/loader:1288:32)\n' +
    '    at Module._load (node:internal/modules/cjs/loader:1104:12)\n' +
    '    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:174:12)\n' +
    '    at node:internal/main/run_main_module:28:49'
}
```

### Bundling

No Bundling

khanayan123:
Hi @JavaScriptBach upon further inspection this seems to be a bug in v8 itself, but in the meantime we have merged in a work around which should be included in the next release.

PR #5580, opened 2025-04-15
add explicit handling of stack errors in logger

khanayan123:
### What does this PR do?
Fixes Winston error stack traces in Node.js 22 by modifying the LogPlugin's message proxy to preserve stack properties when log injection is enabled. Also adds a unit test for the bug.

### Motivation
Add a fix for bug https://github.com/DataDog/dd-trace-js/issues/5528

### DIFF
--- a/packages/dd-trace/src/plugins/log_plugin.js
+++ b/packages/dd-trace/src/plugins/log_plugin.js
@@ -7,6 +7,11 @@ const { storage } = require('../../../datadog-core')
 const hasOwn = (obj, prop) => Object.prototype.hasOwnProperty.call(obj, prop)
 
 function messageProxy (message, holder) {
+  const originalStack = message && typeof message === 'object' &&
+    (message instanceof Error || message.stack)
+    ? message.stack
+    : undefined
+
   return new Proxy(message, {
     get (target, p, receiver) {
       if (p === Symbol.toStringTag) {
@@ -15,6 +20,11 @@ function messageProxy (message, holder) {
 
       if (shouldOverride(target, p)) {
         return holder.dd
+      }
+
+      // Special handling for Error stack property in Node.js 22
+      if (p === 'stack' && originalStack !== undefined) {
+        return originalStack
       }
 
       return Reflect.get(target, p, receiver)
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/plugins/log_plugin.js
```diff
@@ -6,7 +6,14 @@ const { storage } = require('../../../datadog-core')
 
 const hasOwn = (obj, prop) => Object.prototype.hasOwnProperty.call(obj, prop)
 
-function messageProxy (message, holder) {
+function messageProxy (message, holder, constructorId) {
+  // Only apply special stack handling for Winston to avoid breaking other plugins
+  const isWinston = constructorId === 'winston'
```
rochdev, line 11: Could any Winston specific handling be moved to the Winston plugin?

khanayan123: We would need to copy the whole messageProxy function which would be alot of duplicate code

rochdev: Why? The `isWinston` variable is only used once on the next line. Can't this be done before calling the function?

packages/dd-trace/src/plugins/log_plugin.js
```diff
@@ -17,6 +24,11 @@ function messageProxy (message, holder) {
         return holder.dd
       }
 
+      // Special handling for Error stack property in Node.js 22
+      if (p === 'stack' && originalStack !== undefined) {
+        return originalStack
+      }
```
BridgeAR, line 30:
```suggestion
      // This is a workaround for a V8 bug that surfaced in Node.js 22
      if (p === 'stack') {
        return target.stack
      }
```

This seems to be the only change we should do here. The issue seems to be in V8 itself that the stack is not calculated when being called with Reflect.get inside of a proxy.

BridgeAR (changes requested):
This does not seem to be an issue with winston but with V8. It stopped returning the stack if Reflect.get is called from inside of a proxy on the stack property.

I will check the spec again if this is intentional or not but I am quite confident it's just a recent regression.

This is a great opportunity for making a V8 contribution by fixing the issue there, if we determine that it is indeed a V8 bug.

packages/dd-trace/src/plugins/log_plugin.js
khanayan123: with Ruben's suggestion it's no longer an issue

packages/dd-trace/src/plugins/log_plugin.js
khanayan123: Awesome!

BridgeAR: Note: this is not completely safe to do but I think it's a good short term solution.

A safer way would probably be to check if Reflect.get returns undefined and if that's the case, check if the property is `'stack'` and return that. That way the access would use Reflect.get for all cases where we actually do get a value.

PR #5586, opened 2025-04-16
inline orchestrion.yml to avoid bundling issues

bengl:
### What does this PR do?
see title
### Motivation
bundlers won't inline the yaml by default, so this prevents those errors.


tlhunter (approved):

bouwkast (approved):
Nice thanks!

Issue #5552, opened 2025-04-09
[BUG]: Esbuild plugin does not support @koa/router tracing


maciekbb-gyg:
### Tracer Version(s)

5.45.0

### Node.js Version(s)

22

### Bug Report

It seems `packages/datadog-esbuild/index.js` treats `@koa/router` as a local file and so it opts out from instrumentation.

I tracked it down to the following code

```
    // TODO: Should this also check for namespace === 'file'?
    if (args.path.startsWith("@") && !args.importer.includes("node_modules/")) {
      // This is the Next.js convention for loading local files
      if (DEBUG) {
        console.log(`@LOCAL: ${args.path}`);
      }
      return;
    }
```

When running build in debug mode I can see

```
LOCAL: ./index.ts
RESOLVE: koa@2.16.1
@LOCAL: @koa/router
```

### Reproduction Code

```
import Koa from "koa";
import Router from "@koa/router";

const app = new Koa();

const fooRouter = new Router();
fooRouter.get("/fooo", (ctx) => {
  ctx.body = "fooo";
});

app.use(fooRouter.routes());

app.listen(3000);

```

and then build it with

```
const ddPlugin = require("dd-trace/esbuild");
const esbuild = require("esbuild");

esbuild
  .build({
    entryPoints: ["index.ts"],
    bundle: true,
    outdir: "dist",
    plugins: [ddPlugin],
    sourcemap: true,
    platform: "node",
    external: ["graphql/language/visitor", "graphql/language/printer", "graphql/utilities"],
  })
  .then(() => {
    console.log("Build completed successfully.");
  })
  .catch((err) => {
    console.error(err);
    process.exit(1);
  });
```

### Bundling

ESBuild

maciekbb-gyg:
I suppose an easy fix could be to fix the condition like this

```
if (
      args.path.startsWith("@") &&
      !args.importer.includes("node_modules/") &&
      !modulesOfInterest.has(args.path)
    )
```

BridgeAR:
@maciekbb-gyg thank you for the great bug report. Your suggestion also seems to be the right fix!

I opened a PR to implement this and it should be resolved soon.

PR #5588, opened 2025-04-17
Fix esbuild not instrumenting namespaced modules

BridgeAR:
As drive-by it simplifies the tests and moved the test to the dedicated esbuild part.

It also cleans up the temporary files better than before.

Fixes https://github.com/DataDog/dd-trace-js/issues/5552


LICENSE-3rdparty.csv
```diff
@@ -63,6 +64,7 @@ dev,globals,MIT,Copyright (c) Sindre Sorhus <sindresorhus@gmail.com> (https://si
 dev,graphql,MIT,Copyright 2015 Facebook Inc.
 dev,jszip,MIT,Copyright 2015-2016 Stuart Knightley and contributors
 dev,knex,MIT,Copyright (c) 2013-present Tim Griesser
+dev,koa,MIT,Copyright (c) 2019 Koa Contributors
```
rochdev, line 67: Any way to test `koa` support without having to explicitly depend on it in our devDependencies?

BridgeAR: It would require more substantial changes.

BridgeAR: Or we could use a own mocked module that we instrument. I think it just makes sense to rely on the real modules in that case.

tlhunter (dismissed):

LICENSE-3rdparty.csv
rochdev: Definitely makes sense to rely on the real module. However, if we were to do that, we'd end up with dependencies on every single module we instrument. That also makes it impossible to test multiple versions easily. That's why I wonder if something can be done quickly to avoid this, and if not, we should definitely come up with a plan to address this as soon as possible.

rochdev (approved):

PR #5590, opened 2025-04-17
[LANGPLAT-380] Fix OOM tests flakiness

nsavoire:
### What does this PR do?

- Use `tracer.profilerStarted()` to ensure the profiler is started before running the test
- Remove export tests of strategies that are not used in the profiler and are not guaranteed to succeed (async)

### Motivation

Reduce OOM tests flakyness.

### DIFF
--- a/integration-tests/profiler/oom.js
+++ b/integration-tests/profiler/oom.js
@@ -2,7 +2,7 @@
 
 /* eslint-disable no-console */
 
-require('dd-trace').init()
+const tracer = require('dd-trace').init()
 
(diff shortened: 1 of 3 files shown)


rochdev (dismissed):

BridgeAR (commented):
Instead of removing the tests, could we just make them less flaky?

nsavoire:
> Instead of removing the tests, could we just make them less flaky?

The removed tests test unreliable export strategies that are not used by default. I re-added them with 3 retries and will remove them in the future if the test remains flaky.

szegedi (approved):

PR #5591, opened 2025-04-17
accept multiple values for tags in dogstatsd

patrick-vandusen-creditgenie:
### What does this PR do?
Accepts arrays of tag values in dogstatsd.

### Motivation
We need it, Datadog supports it, datadog-lambda-js can already do it, there's no reason dogstatsd shouldn't be able to do it too.


patrick-vandusen-creditgenie:
@BridgeAR 

> @patrick-vandusen-creditgenie adding this is going to cause quite some overhead for everyone. 

I haven't noted significant overhead from the use of Array.isArray, can you elaborate on why you believe that will be the case?

> Would it not be fine to just use a small helper method that calls the API after transforming it to the regular format?

I tried to find a way to do this within the existing interface of dd-trace but wasn't able to, could you explain how you would approach this?

rochdev:
@tlhunter What was the reason the original API didn't accept `['foo:bar', 'foo:baz']` since that's the format of the underlying implementation?

rochdev:
I opened #5748 to support this in the same way as the underlying protocol/implementation instead.

PR #5593, opened 2025-04-18
Improve shimmer instrumentation performance

BridgeAR:
The following improvements are implemented:

1. The prototype has to only be looked up in case the constructor is not the global function. That is a rare case, so it'll be faster in average.
2. The assertions are now faster by preventing duplicated lookups in the positive case.
3. The extra descriptor creation is gone by reusing the received one.
4. Setting descriptors does not fail anymore and the catch may be removed. It got improved significantly in general and uses fast paths


packages/datadog-shimmer/src/shimmer.js
```diff
@@ -3,20 +3,24 @@
...
-  if (proto !== Function.prototype) {
-    Object.setPrototypeOf(wrapped, proto)
+  if (original.constructor !== Function) {
+    const proto = Object.getPrototypeOf(original)
+    if (proto !== Function.prototype) {
+      Object.setPrototypeOf(wrapped, proto)
+    }
   }
-  const props = Object.getOwnPropertyDescriptors(original)
-  const keys = Reflect.ownKeys(props)
```
simon-id, line 13: since you don't iterate on `ownKeys` anymore, is there absolutely no difference between the list given by `ownKeys()`, and the one given by `getOwnPropDes...()` ?
Just giving a made up example but maybe `ownKeys()` doesn't return non-enumerable props, but `getOwnPropDesc()` does ?

simon-id: Even if that's the case, it might not be a problem, to also copy non-enumerable props for example, but it might still  be something to double check

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -142,14 +123,24 @@ function safeWrapper (original, wrapper) {
...
-    callState.startCall()
+    callState[0] = 1
     const retVal = original.apply(this, args)
     if (isPromise(retVal)) {
-      retVal.then(callState.endCall.bind(callState))
+      // TODO: should this not be consistently be either a promise or not? I
+      // would expect we only have to check once and can safe that state.
+      retVal.then((value) => {
+        callState[0] = 2
+        callState[1] = value
```
simon-id, line 139: so from what I understand:
`callState[0] = 0` -> not called
`callState[0] = 1` -> called
`callState[0] = 2` -> completed
`callState[1]` -> return value
is that correct ?
It doesn't look like this is explained anywhere? I had to squint my eyes and use my brain to reverse engineer it.

Two suggestions:
1. something like: `callState[STATE] = COMPLETED`
2. a good comment block to explain all that

that shouldn't decrease the optimization right ?

simon-id (commented):
Lastly, did you measure any real performance gain with this ?
Otherwise LGTM

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -3,20 +3,24 @@
...
+
+  try {
+    // Fast path
+    Object.defineProperties(wrapped, descriptors)
+  } catch {
+    // Fallback for non-configurable properties
+    for (const [key, value] of Object.entries(descriptors)) {
+      if (value.configurable) {
+        Object.defineProperty(wrapped, key, value)
+      }
```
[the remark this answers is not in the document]
BridgeAR: The likelihood for non-configurable options is super rare (almost no code those that). I am happy to pay the price in case someone works with those.

packages/datadog-shimmer/src/shimmer.js
BridgeAR: `Reflect.ownKeys()` returns all keys, no matter if they are enumerable or not. That is however not important in this case due to only getting the keys of the object returned by `getOwnPropertyDescriptors()` and that object only contains enumerable properties.

Your comment made me read the code again and it did indeed contain an issue! So thank you for that! I accidentally used Object.entries() in the fallback and that is wrong (it would skip the symbol properties). I'll also add a test case to address that.

For questions around native APIs, I generally recommend to check [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/ownKeys) (it is easier to read than the spec) :-)

simon-id: Ah I'm glad because I also was looking at the `Object.entries()` with suspicious eyes!

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -33,79 +52,55 @@ function wrapFunction (original, wrapper) {
...
 }
 
-function wrapMethod (target, name, wrapper, noAssert) {
+function wrap (target, name, wrapper, noAssert) {
+  assert(typeof name !== 'function', 'Implementor error: wrap() name should not be a function')
   if (!noAssert) {
     assertMethod(target, name)
-    assertFunction(wrapper)
+    if (typeof wrapper !== 'function') {
+      throw new Error(wrapper ? 'Target is not a function.' : 'No function provided.')
```
watson, line 69:
```suggestion
      throw new Error(wrapper ? 'Target is not a function' : 'No function provided')
```

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -141,52 +136,111 @@ function safeWrapper (original, wrapper) {
...
+  }
 
   // This is the crux of what we're doing in safe mode. It handles errors
   // that _we_ cause, by logging them, and transparently providing results
   // as if no wrapping was done at all. That means detecting (via callState)
   // whether the function has already run or not, and if it has, returning
   // the result, and otherwise calling the original function unwrapped.
-  const handleError = function (args, callState, e) {
-    if (callState.completed) {
+  const handleError = function (args, callState, error) {
```
watson, line 222: If `original` was bound as well to this function, the entire function definition could be moved outside the body of `safeWrapper` so it was only defined once, at load time. But I'm not sure if it actually improves any performance 🤷

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -131,7 +126,7 @@ function safeWrapper (original, wrapper) {
 
   // We're going to hold on to current callState in this variable in this scope,
   // which is fine because any time we reference it, we're referencing it synchronously.
-  // We'll use it in the our wrapper (which, again, is called syncrhonously), and in the
+  // We'll use it in the our wrapper (which, again, is called synchronously), and in the
```
watson, line 129:
```suggestion
  // We'll use it in our wrapper (which, again, is called synchronously), and in the
```

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -141,52 +136,111 @@ function safeWrapper (original, wrapper) {
...
+    // Fast path for async methods. Those are always returning promises, so need
+    // to check for that. That also prevents the need to change the prototype later.
+    // TODO(BridgeAR): Check if the overhead with async hooks and the async methods
+    // is higher than the overhead of the prototype change. That way it's possible
+    // to check for the async methods in copyProperties and skip the prototype change.
+    return function (...args) {
+      currentCallState = [NOT_STARTED, undefined]
+      const callState = currentCallState
+
+      return innerWrapped.apply(this, args).catch((error) => {
```
watson, line 194: I would add a comment about this error handler being a copy of `handleError` and that they probably should be kept in sync. If you ever edit one, you should probably edit the other.

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -1,22 +1,41 @@
...
+  }
+  if (original.name !== wrapped.name) {
+    Object.defineProperty(wrapped, 'name', { value: original.name, configurable: true })
+  }
+  if (ownKeys.length === 2) {
+    return
+  }
+  for (const key of ownKeys) {
+    if (skipMethods.has(key)) continue
+    const descriptor = Object.getOwnPropertyDescriptor(original, key)
```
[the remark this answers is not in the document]
BridgeAR: `Object.getOwnPropertyDescriptors()` returns an object with the properties as keys, so we still have to do the same work afterwards. I did think about having a special case in case a method has many methods, that is just likely a rare case, so I decided for a simpler implementation first. This already improves the simple method case by a factor of 100x in my local benchmark.

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -1,22 +1,41 @@
...
+    if (descriptor?.writable || descriptor?.configurable || !Object.prototype.hasOwnProperty.call(wrapped, key)) {
```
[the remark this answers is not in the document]
BridgeAR: In a recent benchmark, I noticed that the prototype one somehow performs better than `Object.hasOwn()`. That was my reason to use it, while this should not really matter here (it's a fallback for a rare case anyway). I'll update it accordingly.

packages/datadog-shimmer/src/shimmer.js
BridgeAR: I did not want to change the errors so far. So I copied it 1-to-1 as it was. I am fine either way.

packages/datadog-shimmer/src/shimmer.js
BridgeAR: Safe mode had to be removed, so it's gone by now.

packages/datadog-shimmer/src/shimmer.js
BridgeAR: Obsolete due to the removal of safe mode :)

packages/datadog-shimmer/src/shimmer.js
BridgeAR: Obsolete due to the removal of safe mode

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -1,85 +1,88 @@
...
 }
 
 function wrapFunction (original, wrapper) {
-  if (typeof original === 'function') assertNotClass(original)
-
   const wrapped = wrapper(original)
 
-  if (typeof original === 'function') copyProperties(original, wrapped)
+  if (typeof original === 'function') {
+    assertNotClass(original)
```
bengl, line 38: This is a slight change in behaviour, because the wrapper function still gets called even if it's a class, whereas before it wouldn't (it would throw).

In practice, this shouldn't be a problem because we'll throw either way, and any side effects of wrapping likely won't matter much since we'll be aborting instrumentation anyway.

This is probably fine, since tests are currently passing, and functions don't _typically_ become classes in new versions anymore (this used to be common when ES5 syntax was new, but it's been a decade now 🤣).

Non-blocker, just notable.

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -1,85 +1,88 @@
...
+    Object.defineProperty(wrapped, 'length', { value: original.length, configurable: true })
+  }
+  if (original.name !== wrapped.name) {
+    Object.defineProperty(wrapped, 'name', { value: original.name, configurable: true })
+  }
+  if (ownKeys.length !== 2) {
+    for (const key of ownKeys) {
+      if (skipMethods.has(key)) continue
+      const descriptor = Object.getOwnPropertyDescriptor(original, key)
+      if (descriptor.writable || descriptor.configurable || !Object.hasOwn(wrapped, key)) {
```
bengl, line 27: This line makes the assumption that any property name that exists both on the wrapped and the original has roughly the same descriptor, modulo the value. This is because configurability and writability are checked on the original, but not on the wrapped. If we're already making that assumption, maybe it's safe to make a fast path here and just set the value directly if it's writable?

Or maybe that's a rare enough case that it's not worth bothering with here. WDYT?

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -1,85 +1,88 @@
...
   if (typeof original === 'function') copyProperties(original, wrapped)
 
+  const descriptor = Object.getOwnPropertyDescriptor(target, name)
+
   if (descriptor) {
     if (descriptor.get || descriptor.set) {
-      attributes.get = () => wrapped
+      // TODO(BridgeAR): What happens in case there is a setter? This seems wrong?
+      // What happens in case the user does indeed set this to a different value?
+      // In that case the getter would potentially return the wrong value?
```
bengl, line 64: Any getter at all is sufficiently a problem here, since a setter doesn't need to exist for its return value to be different each time.

A bailout here would save us from causing damage, but I also think there's a correct approach here where we ensure that the wrapper always calls the original getter on every invocation.

That can be done in a separate PR, since that's a pre-existing issue here. One could also argue that because tests are passing, the bug/shortcut is fine. I'd disagree with that argument, because we don't exercise in tests the full capabilities of libraries we instrument.

BridgeAR: I would like to address it in a separate PR while agreeing to your perspective.

packages/datadog-shimmer/src/shimmer.js
BridgeAR: I am not sure I follow the comment.

The assumption is indeed that any property on a method that is either writable or configurable on one side would also be that on the other side, if it exists. If it does not exist, it does not matter, than we have to set it anyway.
I always need the descriptor of the original while this prevents looking up the descriptor on the wrapper.

packages/datadog-shimmer/src/shimmer.js
BridgeAR: The reasoning for the change is: the safe guard is there for our tests, so that we detect this early on. It would definitely fail the same as before, even if a class is passed through, just after passing the original to the wrapper. That could theoretically fail earlier but since both cases end up with an error, the behavior does not really change.

bengl (approved):

PR #5596, opened 2025-04-21
refactor(test-optimization)!: raise Playwright minimum version to 1.38.0

Mariovido:
### What does this PR do?

With this we are incresing the minimum supported version of playwright to the version `1.38.0` for v6 and posterior.

This should be a noop for v5.

### Motivation
We want to stop having features depend on the test framework's version. `v1.38.0` is ~2 years old and we'll maintain compatibility in v5 for a year, so I think that's reasonable. 

### Plugin Checklist

- [x] Unit tests.


juan-fernandez (approved):
@Mariovido I've simplified the logic in the PR a bit. Have a look and let me know if it makes sense to you

PR #5597, opened 2025-04-21
Add shimmer shimmer benchmarks

bengl:
The "iterations" is currently set to 1 in the config. Both this and the ITERATIONS constant in each test will likely require some tweaking before this lands.

### DIFF
--- a/benchmark/sirun/means.js
+++ b/benchmark/sirun/means.js
@@ -1,0 +1,29 @@
+/* eslint-disable no-console */
+
+const chunks = []
+process.stdin.on('data', (data) => {
+  chunks.push(data)
+})
+process.stdin.on('end', () => {
+  const data = Buffer.concat(chunks).toString()
+  const json = JSON.parse(data)
+  for (const testSuite in json) {
+    console.log('Test Suite: ', testSuite)
+    const variants = Object.keys(json[testSuite])
+    const results = {}
+    for (const variant of variants) {
+      const { summary } = json[testSuite][variant]
+      results[variant] = {
+        wallTime: summary['wall.time'].mean,
+        userTime: summary['user.time'].mean,
+        sysTime: summary['system.time'].mean,
+        maxResSize: summary['max.res.size'].mean
+      }
+    }
+    const newResults = []
+    Object.keys(results).sort().forEach((name) => {
+      newResults.push({ name, ...results[name] })
+    })
+    console.table(newResults)
+  }
+})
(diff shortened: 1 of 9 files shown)


benchmark/sirun/shimmer-runtime/index.js
```diff
@@ -0,0 +1,66 @@
...
+
+if (ENABLED !== 'true') {
+  for (let i = 0; i < ITERATIONS; i++) {
+    testedFn()
+  }
+} else {
+  if (WRAP_FUNCTION === 'true') {
+    const wrapped = shimmer.wrapFunction(testedFn, (original) => {
+      return function () {
+        return original.apply(this, arguments)
```
BridgeAR, line 46: Please add some dc.channels and some minimal if argument checks to pretend there is some work? That way we can measure some actual overhead. Right now I expect the in-between method to pretty much be zero overhead.

benchmark/sirun/means.js
```diff
@@ -0,0 +1,25 @@
...
+  const json = JSON.parse(data)
+  for (const testSuite in json) {
+    console.log('Test Suite: ', testSuite)
+    const variants = Object.keys(json[testSuite])
+    let results = {}
+    for (const variant of variants) {
+      const wallTime = json[testSuite][variant].summary['wall.time'].mean
+      const userTime = json[testSuite][variant].summary['user.time'].mean
+      const sysTime = json[testSuite][variant].summary['system.time'].mean
+      const maxResSize = json[testSuite][variant].summary['max.res.size'].mean
```
BridgeAR, line 16:
```suggestion
      const { summary } = json[testSuite][variant]
      const wallTime = summary['wall.time'].mean
      const userTime = summary['user.time'].mean
      const sysTime = summary['system.time'].mean
      const maxResSize = summary['max.res.size'].mean
```

benchmark/sirun/shimmer-runtime/index.js
BridgeAR: Ok, makes sense. I would just not fully rely on that. Inlining could for example work magic for this case which it could potentially not do in other cases.

BridgeAR (commented):
It is rather confusing that the benchmark addition changes the performance for existing benchmarks. I can't find an explanation besides unreliable benchmarks.

benchmark/sirun/shimmer-startup/index.js
```diff
@@ -0,0 +1,61 @@
...
+
+const shimmer = require('../../../packages/datadog-shimmer')
+
+const {
+  ENABLED,
+  WRAP_FUNCTION,
+  FUNCTION_NAME
+} = process.env
+
+const ITERATIONS = 1e7
```
BridgeAR, line 11:
```suggestion
const ITERATIONS = 2e6
```

BridgeAR (approved):

PR #5598, opened 2025-04-21
remove safeWrap from shimmer

bengl:
### What does this PR do?
See title

### Motivation
It was never used (due to a typo), and when we attempted to fix and benchmark it, it cause OOM errors we couldn't easily fix. For now, we're better off removing it and waiting until we can implement it more correctly using Orchestrion-JS.


packages/datadog-shimmer/src/shimmer.js
```diff
@@ -22,13 +20,12 @@ function copyProperties (original, wrapped) {
 }
 
 function wrapFunction (original, wrapper) {
-  if (typeof original === 'function') assertNotClass(original)
+  assertFunction(original)
```
BridgeAR, line 23: This fails in our tests. Seems like sometimes we pass in undefined (mariadb in case there is no callback. That could be optimized so that we just skip the check in our instrumentation in that case) and other spots. Sometimes we also just pass through `null` directly. All of these should not use shimmer for these cases in the first place. So it's good to detect these cases now, we just have to fix the code accordingly. Question is: separate the behavior / validation from this PR? I guess that is best.

bengl: Fixed in latest rev

BridgeAR (approved):

PR #5601, opened 2025-04-23
Add minimal fast path for esm being enabled

BridgeAR:
The main change is to check for the top level presence of import- in-the-middle. That way there is no need for extra checks.

It won't change much, but it's definitely less work in case the cache has lots and lots of entries.

### DIFF
--- a/packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
+++ b/packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
@@ -19,9 +19,11 @@ let config
 const hardcodedSecretCh = dc.channel('datadog:secrets:result')
 let rewriter
 let unwrapCompile = () => {}
-let getPrepareStackTrace, cacheRewrittenSourceMap
+let getPrepareStackTrace
+let cacheRewrittenSourceMap
 let kSymbolPrepareStackTrace
-let esmRewriterEnabled = false
+
+function noop () {}
 
 function isFlagPresent (flag) {
   return process.env.NODE_OPTIONS?.includes(flag) ||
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
```diff
@@ -181,16 +183,17 @@ function isEsmConfigured () {
   const hasLoaderArg = isFlagPresent('--loader') || isFlagPresent('--experimental-loader')
   if (hasLoaderArg) return true
 
-  const initializeLoaded = Object.keys(require.cache).find(file => file.includes('import-in-the-middle/hook.js'))
-  return !!initializeLoaded
+  // Fast path for common case when enabled
+  if (require.cache[`${process.cwd()}/node_modules/import-in-the-middle/hook.js`]) {
```
rochdev, line 187: This would be more useful for serverless, but in serverless the tracer is installed elsewhere. On servers, I don't think the tiny gain justifies the added code.

uurien: This would be also problematic for us, in development, when I am trying the tracer locally in a custom app, the tracer is in other directory with their own `node_modules`, not in `process.cwd()`.

rochdev (approved):

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
rochdev: @duncanista Do you know what the path would be in Lambda so that it can also benefit from this change?

PR #5603, opened 2025-04-23
Initial central configuration inversion draft

BridgeAR:


packages/dd-trace/src/config-helper.js
```diff
@@ -0,0 +1,34 @@
...
+// 3. Add a file that defines the supported configurations and their aliases
+// 4. Simplify config.js
+// 5. Make sure config.js is loaded first, right after calling init. The order matters
+
+const { supportedConfigurations, aliases } = require('./supported-configurations')
+
+const configs = {}
+// Round 1: assign all valid configs and backup to aliases if needed
+for (const name of Object.keys(supportedConfigurations)) {
+  if (name in process.env) {
```
BridgeAR, line 15: `in` in JS is something weird. It checks the prototype as well, so we should use `Object.hasOwn()` instead :)

```suggestion
  if (Object.hasOwn(process.env, name)) {
```

We could also check for truthy values instead of accepting empty strings, but that might change our behavior.

packages/dd-trace/src/config-helper.js
```diff
@@ -0,0 +1,34 @@
...
+    configs[name] = process.env[name]
+  } else {
+    for (const alias in aliases[name]) {
+      if (alias in process.env) {
```
BridgeAR, line 19:
```suggestion
    for (const alias of aliases[name]) {
```

The alias validation should ideally happen in the CI. That way we don't have to do that during runtime.

packages/dd-trace/src/config-helper.js
```diff
@@ -0,0 +1,75 @@
...
+    return configs
+  },
+  getConfiguration (name) {
+    const config = process.env[name]
+    if ((name.startsWith('DD_') || name.startsWith('OTEL_')) &&
+        !supportedConfigurations[name] &&
+        !aliasObject[name]) {
+      throw new Error(`Missing ${name} configuration in supported-configurations file.`)
+    }
+    if (config === undefined && aliasObject[name]) {
```
mhlidd, line 66:
```suggestion
    if (config === undefined) {
```
`aliasObject` would not have a primary env, which is the case we are looking at here.

.gitlab-ci.yml
```diff
@@ -5,11 +5,15 @@ stages:
...
+    - when: on_success
+  variables:
+    LOCAL_JSON_PATH: "packages/dd-trace/src/supported-configurations.json"
+
 include:
   # See: https://gitlab.ddbuild.io/DataDog/apm-reliability/libdatadog-build
-  - project: DataDog/apm-reliability/libdatadog-build
-    ref: 0f677257308e1c379af490b754febfb40fa2c06d
-    file: templates/one-pipeline.yml
+  - remote: https://gitlab-templates.ddbuild.io/libdatadog/include-wip/anna/config-inversion-add-step/one-pipeline.yml
```
BridgeAR, line 16: The CI pipeline PR has to be merged before and this has to be updated before landing

packages/dd-trace/src/config-helper.js
```diff
@@ -0,0 +1,90 @@
...
+  }
+}
+
+// TODO: Consider to join deprecations with aliases by just making those entries an object.
+const deprecationMethods = {}
+for (const deprecation of Object.keys(deprecations)) {
+  deprecationMethods[deprecation] = deprecate(
+    () => {},
+    `The environment variable ${deprecation} is deprecated. Please use ${aliasToCanonical[deprecation]} instead.`,
+    `DATADOG_${deprecation}`
```
watson, line 24: I wonder if this can easily be confused with an actual environment variable..... could this somehow include the word `DEP` so it's similar to the Node.js deprecations? Technically there's no rule that I know of saying they have the use the same `code` between reboots, so it could just be a sequential number also... e.g. `DD_DEP0001` etc (but I'd prefer if they never changed, which would make googling easier)

packages/dd-trace/src/supported-configurations.json
```diff
@@ -0,0 +1,432 @@
+{
+  "supportedConfigurations": {
```
watson, line 2: Why isn't this an array if strings? As far as I can see, we're just using each element to verify the if they are present or not... or have I misunderstood something?

eslint-rules/eslint-process-env.mjs
```diff
@@ -0,0 +1,61 @@
...
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Disallow usage of process.env outside config.js'
+    },
+    schema: []
+  },
+  create (context) {
+    // eslint-disable-next-line @stylistic/js/max-len
+    const allowedFile = /[/\\]packages[/\\]dd-trace[/\\]src[/\\](config-helper|guardrails[/\\](index|log|telemetry))\.js$/
```
watson, line 11: Why not just add `/* eslint-disable eslint-rules/eslint-process-env */` at the top of these files?

packages/dd-trace/src/config-helper.js
```diff
@@ -0,0 +1,90 @@
...
+            }
+          }
+          deprecationMethods[env]?.()
+        } else {
+          debug(
+            `Missing configuration ${env} in supported-configurations file. The environment variable is ignored.`
+          )
+        }
+      } else {
+        configs[env] = value
```
watson, line 60: This will also return environment variables not starting with `DD_` or `OTEL_`. That goes against what's stated in the JSDoc comment. Either that or this logic has to change

packages/dd-trace/src/config-helper.js
```diff
@@ -0,0 +1,90 @@
...
+const deprecationMethods = {}
+for (const deprecation of Object.keys(deprecations)) {
+  deprecationMethods[deprecation] = deprecate(
+    () => {},
+    `The environment variable ${deprecation} is deprecated. Please use ${aliasToCanonical[deprecation]} instead.`,
+    `DATADOG_${deprecation}`
+  )
+}
+
+const debug = debuglog('dd:debug')
```
watson, line 28: Why use `debuglog` and not our own logger?

packages/dd-trace/src/config-helper.js
```diff
@@ -0,0 +1,90 @@
+'use strict'
+
+// TODO: Add jira tickets
```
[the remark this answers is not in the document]
BridgeAR: I'll add those and remove it.

packages/dd-trace/src/config-helper.js
BridgeAR: We can use what ever prefix we like. I used this, since it seemed most intuitive for me and the sequential one change over time. This implementation keeps it consistent. I would like that.

eslint-rules/eslint-process-env.mjs
BridgeAR: No strong opinion. I believe we should not lint the guardrails in general. We can also overcome that differently though.

packages/dd-trace/src/config-helper.js
BridgeAR: It would cause a circular structure that is not possible to be resolved. We'd need the logger to be changed significantly.

packages/dd-trace/src/config-helper.js
BridgeAR: The comment is actually somewhat correct (environment variables that don't start with DD_ or OTEL_ are generally supported). Do you have a suggestion to make the comment clearer? :)

packages/dd-trace/src/config-helper.js
watson: Yeah I agree... just don't like how much these actually look like environment variables as it might confuse users. But it's not a deal breaker to me, so feel free to keep it as is.

packages/dd-trace/src/supported-configurations.json
watson: But am I correct that it's not used currently for anything? Maybe an example of how it is intended to be used would be nice

packages/dd-trace/src/config-helper.js
watson: Are you sure? If I recall correctly, I removed the tight coupling between the config and logger in https://github.com/DataDog/dd-trace-js/pull/4515

Also, just tried this in the console which works:

image

watson: Ah sorry, I might have misunderstood. You specifically mention circular references, which I haven't checked. So I assume you've verified the circular reference hasn't been resolved by the time the logger is needed?

packages/dd-trace/src/supported-configurations.json
BridgeAR: The versions are needed. They are not used here in the repository but in the backend to verify the correct support in each library. It would not work without that.

The version declares the behavior of the individual configuration. That could be identical in another platform, it could however also be another default value, different behavior or similar and then have another version.

packages/dd-trace/src/config-helper.js
BridgeAR: They are logged as error code, so it should at least be clear as soon as we receive the deprecation warning :)
image

packages/dd-trace/src/config-helper.js
BridgeAR: Yes, that is the issue. I can have another look at it while it did not seem straight forward to resolve it last time I had a look.

packages/dd-trace/src/config-helper.js
```diff
@@ -0,0 +1,90 @@
...
+
+  /**
+   * Returns the environment variable, if it's supported or a non Datadog
+   * configuration. Otherwise, it throws an error.
+   *
+   * @param {string} name Environment variable name
+   * @returns {string|undefined}
+   * @throws {Error} if the configuration is not supported
+   */
+  getConfiguration (name) {
```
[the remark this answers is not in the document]
BridgeAR: The benchmarks are sadly misleading. The change comes from reading the JSON file and we don't get around that.
The change shows up not due to actual runtime overhead but due to the load time which is also taken into account in our benchmarks. I have already spoken with @bengl about that problem and I want to find a way with @ddyurchenko to improve the situation (there is a draft PR by @bengl to improve. It would however likely still show up in some metrics. They are just somewhat flawed). The actual overhead is not noticeable during runtime.

Let me think a bit more about finding a better name. I agree that it's likely better to rename it.

packages/dd-trace/src/config-helper.js
watson: It would be prudent to make the logger work in situations like this. Having two different loggers is not ideal. That being said, the way the current logger works is great. So maybe it's not worth it trying to fix it. For example:
- You either enable all or nothing (I'd like to be able to enable logging just for a specific instrumentation or just for profiling for example)
- The environment variable to enable logging is called `DD_TRACE_DEBUG`... it's used outside of just traces and it's has a log level in the name, when in fact you you can change the level independently

In short, it's a bit of a mess.

BridgeAR: I agree to all of these points. The question is if we want to address it right away? I could remove the logging for now, while I am not sure if that's the best solution.

packages/dd-trace/test/profiling/config.spec.js
```diff
@@ -244,13 +244,13 @@ describe('config', () => {
...
-    expect(warnings[1]).to.equal(
-      'DD_PROFILING_EXPERIMENTAL_CODEHOTSPOTS_ENABLED is deprecated. ' +
-      'Use DD_PROFILING_CODEHOTSPOTS_ENABLED instead.')
+    // expect(warnings.length).to.equal(2)
+    // expect(warnings[0]).to.equal(
+    //   'DD_PROFILING_EXPERIMENTAL_ENDPOINT_COLLECTION_ENABLED is deprecated. ' +
+    //   'Use DD_PROFILING_ENDPOINT_COLLECTION_ENABLED instead.')
+    // expect(warnings[1]).to.equal(
+    //   'DD_PROFILING_EXPERIMENTAL_CODEHOTSPOTS_ENABLED is deprecated. ' +
+    //   'Use DD_PROFILING_CODEHOTSPOTS_ENABLED instead.')
```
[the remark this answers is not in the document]
szegedi: I think it'd be perfectly fine, yes. I'm not actually entirely sure what that is, but if it's the way Node.js reports deprecations, then I'd be in favor of it, because that's then consistent in appearance to Node.js users.

packages/dd-trace/src/config-helper.js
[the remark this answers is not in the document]
BridgeAR: I updated it with `getEnvironmentVariable`

packages/dd-trace/src/config-helper.js
BridgeAR: I removed the logging for now and added a TODO

packages/dd-trace/src/config-helper.js
```diff
@@ -0,0 +1,90 @@
...
+   *
+   * @param {string} name Environment variable name
+   * @returns {string|undefined}
+   * @throws {Error} if the configuration is not supported
+   */
+  getEnvironmentVariable (name) {
+    const config = process.env[name]
+    if ((name.startsWith('DD_') || name.startsWith('OTEL_')) &&
+        !supportedConfigurations[name] &&
+        !aliasToCanonical[name]) {
```
watson: I would prefer, just as our code is not allowed to reference any of the deprecated environment variables, that it shouldn't reference any of the aliases either - only the canonical version. This will make searching for it much simpler.

```suggestion
        !supportedConfigurations[name]) {
```

packages/dd-trace/src/config-helper.js
```diff
@@ -0,0 +1,90 @@
...
+      throw new Error(`Missing ${name} configuration in supported-configurations file.`)
```
watson: I prefer to have the full filename in the error message. Makes it much easier to find the file if you're not familiar with it:

```suggestion
      throw new Error(`Missing ${name} configuration in "supported-configurations.json" file.`)
```

packages/dd-trace/src/config-helper.js
```diff
@@ -0,0 +1,90 @@
...
+    }
```
watson: Just moving initializing of the `config` const down below the initial guard:

```suggestion
    if ((name.startsWith('DD_') || name.startsWith('OTEL_')) &&
        !supportedConfigurations[name] &&
        !aliasToCanonical[name]) {
      throw new Error(`Missing ${name} configuration in supported-configurations file.`)
    }
    const config = process.env[name]
```

packages/dd-trace/src/config-helper.js
```diff
@@ -0,0 +1,90 @@
...
+          if (configs[aliasToCanonical[env]] === undefined) {
+            // In case that more than a single alias exist, use the one defined first in our own order
+            for (const alias of aliases[aliasToCanonical[env]]) {
+              if (process.env[alias] !== undefined) {
+                configs[aliasToCanonical[env]] = value
+                break
+              }
+            }
+          }
+          deprecationMethods[env]?.()
```
watson: It's currently the practise, that a deprecated environment variable is also an alias for a canonical version. But is this a constant truth, or should this code support that the `deprecations` object might contain keys not aliased in the `aliases` object? If so, the call to this function should be moved outside this if-block.

BridgeAR: It is currently implemented as subset. Also the deprecation warning itself. Since we do not have any other cases so far, I'd keep it like that for now.
Let's change it when needed :)

watson: My point is that it's easy to add a deprecation key to the `deprecations` object in the JSON file, not knowing that it will not do anything unless it's also in one of the arrays in `aliases`. So it's an easy mistake to make in the future if you don't know that these two a tightly linked.

BridgeAR: I implemented it in a generic way. It's now possible to either provide another configuration (for aliases) or a specific deprecation message that will be used if there is no alias.

watson (approved):

PR #5605, opened 2025-04-23
Support for Prisma client library

### LINKED ISSUE #1244: Request: Add support for Prisma query tracing
I'm using [Prisma](prisma.io) to manage database connections / queries. Really missing being able to track down slow MySQL queries in DD. Would be great to see support for prisma query tracing added

bojbrook:
### What does this PR do?
Adds tracing support for the Prisma ORM library for v6.1.0+.

Fixes #1244 

**Full Example trace**
image

**Example of generated SQL span**
image

**Transaction query**
image

### Motivation
Multiple submitted FR for this library.

### Plugin Checklist

- [x] Unit tests.
- [x] TypeScript [tests][2].
- [x] API [documentation][3].


packages/dd-trace/src/tracer.js
```diff
@@ -77,7 +77,7 @@ class DatadogTracer extends Tracer {
           }
         )
       } else {
-        span.finish()
+        span.finish(options.finishTime)
```
tlhunter, line 80: This is interesting! What's happening here?

bojbrook: `span.finish()` takes a finishTime argument but it wasn't getting passed down in the `tracer.startSpan` function call.  This is needed for the `prisma.engine` spans as the engine binray returns an array of span attributes with a start and stop time that are then built into spans.

packages/datadog-plugin-prisma/test/index.spec.js
```diff
@@ -0,0 +1,262 @@
...
+
+describe('Plugin', () => {
+  let prisma
+  let prismaClient
+  let tracingHelper
+
+  describe('prisma', () => {
+    withVersions('prisma', ['@prisma/client'], version => {
+      before(async () => {
+        require('../../dd-trace')
```
tlhunter, line 55: sus

packages/datadog-instrumentations/src/prisma.js
```diff
@@ -0,0 +1,100 @@
...
+// https://github.com/prisma/prisma/releases/tag/6.1.0
+addHook({ name: '@prisma/client', versions: ['>=6.1.0'] }, prisma => {
+  const tracingHelper = new TracingHelper()
+  global.PRISMA_INSTRUMENTATION = {
+    helper: tracingHelper
+  }
+
+  global.V0_PRISMA_INSTRUMENTATION = {
+    helper: tracingHelper
+  }
```
tlhunter, line 97: Def gonna need comments

packages/datadog-plugin-prisma/src/engine.js
```diff
@@ -0,0 +1,98 @@
...
+          startTime: unixStartTime,
+          type: type || engineSpan.attributes['db.system'],
+          meta: {
+            ...options.meta,
+            'db.type': dbType || engineSpan.attributes['db.system']
+          },
+          childOf
+        }
+      }
+        break
```
tlhunter, line 75:
```suggestion
      break
```

packages/datadog-instrumentations/src/prisma.js
```diff
@@ -0,0 +1,100 @@
...
+  dispatchEngineSpans (spans) {
+    const roots = spans.filter((span) => span.parentId === null)
+    for (const root of roots) {
+      prismaEngineStart.publish({ engineSpan: root, allEngineSpans: spans })
+    }
+  }
+
+  getActiveContext () {}
+
+  runInChildSpan (options, callback) {
```
tlhunter, line 40: console.log(this, arguments)

tlhunter:
Main concern: what exactly should the spans look like? Can you generate a unrelated pg query in your test app talking to the same DB to compare results / attributes? Might be worth meeting with a customer next week.

packages/datadog-plugin-prisma/test/index.spec.js
```diff
@@ -0,0 +1,262 @@
...
+  unlink(generatedSchemaPath, () => { })
+}
+
+describe('Plugin', () => {
+  let prisma
+  let prismaClient
+  let tracingHelper
+
+  describe('prisma', () => {
+    withVersions('prisma', ['@prisma/client'], version => {
```
tlhunter, line 53:
```suggestion
    withVersions('prisma', ['@prisma/client'], version => {
      if (version === '6.1.0') return; // temp test
```

tlhunter: (this won't actually get merged, just a test run)

basically run it like this in CI to test if the issue really is with 6.1.0 or if it's something to do with leaky state between test runs

packages/dd-trace/src/tracer.js
bojbrook: Removed this field as it was not needed

packages/datadog-plugin-prisma/test/index.spec.js
bojbrook: This issue has been fixed

packages/datadog-plugin-prisma/test/index.spec.js
bojbrook: I removed this as it wasn't needed

packages/datadog-instrumentations/src/prisma.js
bojbrook: Comments and links have been added to explain the process

packages/datadog-instrumentations/src/prisma.js
```diff
@@ -0,0 +1,141 @@
...
+  // needs a sampled tracecontext to generate engine spans
+  getTraceParent (context) {
+    return '00-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-bbbbbbbbbbbbbbbb-01' // valid sampled traceparent
+  }
+
+  dispatchEngineSpans (spans) {
+    const roots = spans.filter((span) => span.parentId === null)
+    for (const root of roots) {
+      prismaEngineStart.publish({ engineSpan: root, allEngineSpans: spans, dbConfig: this.dbConfig })
+    }
```
BridgeAR, line 36: Nit: this prevents the extra array allocation:

```suggestion
    for (const span of spans) {
      if (span.parentId === null) {
        prismaEngineStart.publish({ engineSpan: span, allEngineSpans: spans, dbConfig: this.dbConfig })
      }
    }
```

packages/datadog-instrumentations/src/prisma.js
```diff
@@ -0,0 +1,141 @@
...
+      }
+
+      return asyncResource.runInAsyncScope(() => {
+        if (!primsaClientStartCH.hasSubscribers) {
+          return callback.apply(this, ctx)
+        }
+        primsaClientStartCH.publish(ctx)
+        try {
+          const result = callback.apply(this, ctx)
+          if (isPromise(result)) {
```
BridgeAR, line 62: Are we certain this is always a native promise and not potentially a thenable?

packages/datadog-plugin-prisma/src/client.js
```diff
@@ -0,0 +1,38 @@
...
+    this.startSpan(this.operationName({ operation: this.operation }), options)
+  }
+}
+
+function formatResourceName (resource, attributes) {
+  if (attributes && attributes.name) {
+    return `${attributes.name || ''}`.trim()
+  }
+  if (attributes && attributes.model && attributes.method) {
+    return `${attributes.model || ''}.${attributes.method || ''}`.trim()
```
BridgeAR, line 32:
```suggestion
  if (attributes?.name) {
    return `${attributes.name}`.trim()
  }
  if (attributes?.model && attributes.method) {
    return `${attributes.model}.${attributes.method}`.trim()
```

These are impossible to be empty strings due to the truthy checks in the if statements.

packages/datadog-plugin-prisma/src/engine.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+}
+
+class PrismaEngine extends DatabasePlugin {
+  static get id () { return 'prisma' }
+  static get operation () { return 'engine' }
+  static get system () { return 'prisma' }
+
+  start (ctx) {
+    // console.log('PrismaEngine start called', ctx.engineSpan)
```
BridgeAR, line 31:
```suggestion
```

packages/datadog-plugin-prisma/src/engine.js
```diff
@@ -0,0 +1,103 @@
...
+    const { engineSpan, allEngineSpans, childOf, dbConfig } = ctx
+
+    const service = this.serviceName({ pluginConfig: this.config, system: this.system })
+
+    const unixStartTime = hrTimeToUnixTimeMs(engineSpan.startTime)
+
+    const spanName = engineSpan.name.substring(14) // remove 'prisma:engine:' prefix
```
BridgeAR, line 38:
```suggestion
    const spanName = engineSpan.name.slice(14) // remove 'prisma:engine:' prefix
```

Substring has some behavior that is more difficult to follow in JS

packages/datadog-plugin-prisma/src/engine.js
```diff
@@ -0,0 +1,103 @@
...
+    switch (spanName) {
+      case 'db_query':{
+        const query = engineSpan.attributes['db.query.text']
+        const originalStatement = this.maybeTruncate(query)
+        let type
+        let dbType
+        if (databaseDriverMapper[engineSpan.attributes['db.system']]) {
+          type = databaseDriverMapper[engineSpan.attributes['db.system']].type
+          dbType = databaseDriverMapper[engineSpan.attributes['db.system']]['db.type']
+        }
```
BridgeAR, line 60:
```suggestion
        const type = databaseDriverMapper[engineSpan.attributes['db.system']]?.type
        const dbType = databaseDriverMapper[engineSpan.attributes['db.system']]?.['db.type']
```

packages/datadog-plugin-prisma/src/engine.js
```diff
@@ -0,0 +1,103 @@
...
+          meta: {
+            ...options.meta,
+            'db.type': dbType || engineSpan.attributes['db.system'],
+            'db.instance': dbConfig?.database,
+            'db.name': dbConfig?.user,
+            'out.host': dbConfig?.host,
+            [CLIENT_PORT_KEY]: dbConfig?.port
+          },
+          childOf
+        }
```
BridgeAR, line 78: While it's convenient to spread the object here for readability reasons, I'd actually just replace the individual properties on the object. That way the object is not copied and just changed, which is slightly faster.

It will likely not have a noticeable impact here, so it's fine for me to keep it.

The same with the options copy in the default case.

packages/datadog-plugin-prisma/src/engine.js
```diff
@@ -0,0 +1,103 @@
...
+        break
+      }
+    }
+
+    const activeSpan = this.startSpan(this.operationName({ operation: this.operation }), options)
+    const children = allEngineSpans.filter((span) => span.parentId === engineSpan.id)
+    for (const child of children) {
+      const startCtx = { engineSpan: child, allEngineSpans, childOf: activeSpan, dbConfig }
+      this.start(startCtx)
+    }
```
BridgeAR, line 93: Preventing the array allocation is actually quite good for performance.

```suggestion
    for (const span of allEngineSpans) {
      if (span.parentId === engineSpan.id) {
        const startCtx = { engineSpan: span, allEngineSpans, childOf: activeSpan, dbConfig }
        this.start(startCtx)
      }
    }
```

packages/datadog-plugin-prisma/test/index.spec.js
```diff
@@ -0,0 +1,310 @@
...
+            // grabbing actual db query span
+            const engineDBSpan = traces[0].find(span => span.meta['prisma.name'] === 'db_query')
+            expect(engineDBSpan).to.have.property('resource', 'SELECT 1')
+            expect(engineDBSpan).to.have.property('type', 'sql')
+            expect(engineDBSpan.meta).to.have.property('span.kind', 'client')
+            expect(engineDBSpan).to.have.property('name', expectedSchema.engine.opName)
+            expect(engineDBSpan).to.have.property('service', expectedSchema.engine.serviceName)
+          }).then(done).catch(done)
+
+          prismaClient.$queryRaw`SELECT 1`.catch(done)
```
BridgeAR, line 141: I am currently trying to make use of more async/await in the tests for readability reasons and because we often see "done called more than once", which would be fixed with this:

```suggestion
          })

          await Promise.all([
            prismaClient.$queryRaw`SELECT 1`,
            tracingPromise
          ])
```

It does need the method to be changed to an async one and the `tracingPromise` must be defined.

I added this comment as example for the rest of the tests.

packages/dd-trace/src/plugins/tracing.js
```diff
@@ -101,14 +101,15 @@ class TracingPlugin extends Plugin {
...
-  startSpan (name, { childOf, kind, meta, metrics, service, resource, type } = {}, enter = true) {
+  startSpan (name, { childOf, kind, meta, metrics, service, resource, type, startTime } = {}, enter = true) {
     const store = storage('legacy').getStore()
     if (store && childOf === undefined) {
       childOf = store.span
     }
 
     const span = this.tracer.startSpan(name, {
       childOf,
+      startTime, // allow custom start time
```
BridgeAR, line 112: I don't think we should generally add the startTime, since this is only used in Prisma.

Could the startTime also not just be a tag instead?

BridgeAR (commented):
This is definitely great work!

I left a few minor comments about making the code a tad faster and tests a tad more readable.

The only concern I have is the added `startTime` property added to the span. I don't think it's a good idea to have that only for Prisma.

tlhunter:
Also we would like to release this disabled by default, get some user feedback, before enabling by default.

packages/dd-trace/src/plugins/tracing.js
bojbrook: I don't believe adding startTime as a tag would work here. When we create a new span we either use the provided startTime or get the current time `this._startTime = fields.startTime || this._getTime()`. 

The other option I can think of if we don't want to add a startTime to the tracing plugin is to call the downstream `tracer.startSpan` directly from the prisma plugin since it already accepts a startTime field - https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/opentracing/tracer.js#L50.

packages/datadog-plugin-prisma/src/client.js
bojbrook: Fixed 👍

packages/datadog-plugin-prisma/src/engine.js
bojbrook: 👍

packages/datadog-plugin-prisma/src/engine.js
bojbrook: This makes sense. I changed it to where the options object is modified instead of using the spread operation

packages/datadog-instrumentations/src/prisma.js
```diff
@@ -0,0 +1,142 @@
...
+
+const prismaEngineStart = channel('apm:prisma:engine:start')
+
+const allowedClientSpanOperations = [
+  'operation',
+  'serialize',
+  'transaction'
+]
+
+class TracingHelper {
```
rochdev, line 21: Anything related to tracing should be on the subscriber side.

packages/datadog-instrumentations/src/prisma.js
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    if (allowedClientSpanOperations.includes(options.name)) {
+      const asyncResource = new AsyncResource('bound-anonymous-fn')
+      const ctx = {
+        resourceName: options.name,
+        attributes: options.attributes || {}
+      }
+
+      return asyncResource.runInAsyncScope(() => {
```
rochdev, line 56: `TracingChannel` should be used instead of `AsyncResource`.

rochdev (changes requested):

packages/datadog-instrumentations/src/prisma.js
BridgeAR: @rochdev that is definitely true. Should we nevertheless already land this with AsyncResource for now? It would likely otherwise take even more time to get it in? It would also be deactivated by default.

rochdev: The main issue is that there is a lot of tracing code in the publisher. This means that a lot of the code will have to move around anyway. At that point, switching to `TracingChannel` at the same time won't be a ton of effort.

packages/datadog-instrumentations/src/prisma.js
```diff
@@ -0,0 +1,141 @@
...
+
+  prisma.PrismaClient = PrismaClient
+  /*
+    * This is taking advantage of the built in tracing support in Prisma.
+    * This is done by setting a global tacing helper that Prisma uses
+    * throuthout the Prisma codebase
+  */
+  // eslint-disable-next-line @stylistic/js/max-len
+  // https://github.com/prisma/prisma/blob/478293bbfce91e41ceff02f2a0b03bb8acbca03e/packages/instrumentation/src/PrismaInstrumentation.ts#L42
+  global.PRISMA_INSTRUMENTATION = {
```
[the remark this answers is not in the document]
tlhunter: I believe it will but I think it's a non issue? Don't we always conflict with other APM tools?

This is the only way Prisma allows tools like us to hook into it. 

There might be a hacky solution where we use a getter/setter on the global and then fall through to another instrumentation library if present? Not sure if it's worth the pain though.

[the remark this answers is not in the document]
rochdev: Definitely not a blocker, just wanted to make sure it was considered.

packages/datadog-instrumentations/src/prisma.js
rochdev: This was a misread on my part, this is only there as this is what is used to instrument Prisma.

packages/datadog-instrumentations/src/prisma.js
bojbrook: I just pushed the change for `TracingChannels` so we should be good here.

packages/datadog-instrumentations/src/prisma.js
```diff
@@ -0,0 +1,114 @@
...
+    * This is taking advantage of the built in tracing support in Prisma.
+    * This is done by setting a global tacing helper that Prisma uses
+    * throuthout the Prisma codebase
+  */
+  // https://github.com/prisma/prisma/blob/478293bbfce91e41ceff02f2a0b03bb8acbca03e/packages/instrumentation/src/PrismaInstrumentation.ts#L42
+  global.PRISMA_INSTRUMENTATION = {
+    helper: tracingHelper
+  }
+
+  global.V0_PRISMA_INSTRUMENTATION = {
```
BridgeAR: So this is an interesting one. They dynamically define this variable during built time to be current major.
It would be `V6_PRISMA_INSTRUMENTATION` right now, since the current latest major is 6.

They intentionally version it for instrumentation libraries such as ours to guarantee it would not break us in case their structure changes. I feel that is a confusing expectation, since it would definitely break any major if we do not automatically read this as well. We should probably reach out to them and discuss this topic.

As long as we're confident that we just continue to handle things gracefully, even if they change their internal format, we could read the information dynamically as well:

```js
const version = require('prisma/package.json').version
```

This is pretty much the fastest way possible to know what version the user has installed.

packages/datadog-instrumentations/src/prisma.js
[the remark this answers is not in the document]
BridgeAR: Reading the comment in their source, we should actually not use this variable at all.
We should only use the one below.

packages/datadog-plugin-prisma/src/client.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+    const operationName = this.operationName({ operation: this.operation })
+    this.startSpan(operationName, options, ctx)
+
+    return ctx.currentStore
+  }
+
+  end (ctx) {
+    // Only synchronous operations would have `result` on `end`.
+    if (!ctx.hasOwnProperty('result')) return
```
BridgeAR: Please do not use `.hasOwnProperty()`. Instead, please always use

```suggestion
    if (!Object.hasOwn(ctx, 'result')) return
```

packages/datadog-plugin-prisma/test/index.spec.js
```diff
@@ -0,0 +1,328 @@
...
+          ])
+        })
+
+        it('should handle errors', async () => {
+          let error
+          const tracingPromise = agent.assertSomeTraces(traces => {
+            expect(traces[0][0].meta).to.have.property(ERROR_TYPE, error.name)
+            expect(traces[0][0].meta).to.have.property(ERROR_MESSAGE, error.message)
+            expect(traces[0][0].meta).to.have.property(ERROR_STACK, error.stack)
+          })
```
BridgeAR: We have a new nicer helper method that is easier to read, use, and it produces better error messages by showing all information at once.

Please also apply this pattern to **all** cases where applicable. That is almost of all these calls.

```suggestion
          const tracingPromise = agent.assertFirstTraceSpan({
            meta: {
              [ERROR_TYPE]: error.name,
              [ERROR_MESSAGE]: error.message,
              [ERROR_STACK]: error.stack
            }
          })
```

BridgeAR (changes requested):
Definitely great work!

I am looking forward to soon releasing it :)

One thing should be addressed before we land it.

packages/datadog-instrumentations/src/prisma.js
crysmags: We can remove the first global since the versioned one takes precedence and I will add a line to check the prisma version.

crysmags: Question, do we want to hardcode the prisma version for the variable? We could return the highest supported major version of prisma. Though this won't work to support different version ranges, such as when v7 is released for prisma.

BridgeAR: As suggested above: I would not hardcode it. Even if that is what the Prisma authors currently suggest. It just means we will not support anything in case they have a new major vs we might not support things when they release a new version. I am therefore surprised they decided this is the way to go.
I think we should reach out and discuss that with the maintainers.

crysmags: Got it, so we would need to grab the version the customer is instrumenting. There is logic currently to check the package and the module path, `extractPackageAndModulePath` returns the package name and module path. I could use this or similar logic to make a helper and then grab the version using your suggestion with `require` and `.version`. 
cc @tlhunter and @rochdev in case you guys know a better way to do this that I am not aware of.

tlhunter: I think we can do this pretty cleanly like so:

```js
addHook({ name: '@prisma/client', versions: ['>=6.1.0'] }, (prisma, version) => {
  // existing stuff
  const major = version.split('.')[0]
  global[`V${major}_PRISMA_INSTRUMENTATION`] = {
    helper: tracingHelper
  }
  return prisma
})
```

packages/datadog-plugin-prisma/src/engine.js
```diff
@@ -0,0 +1,88 @@
...
+      meta: {
+        prisma: {
+          name: spanName,
+          type: 'engine'
+        }
+      }
+    }
+
+    switch (spanName) {
+      case 'db_query':{
```
rochdev: A switch case with only 1 branch should probably be an `if`.

rochdev (commented):
Approving as I only had one small nit that I commented on. However, before merging this I think

packages/datadog-instrumentations/src/prisma.js
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// Tracing support GA in  v6.1.0+
+// https://github.com/prisma/prisma/releases/tag/6.1.0
+let majorVersion
+
+addHook({ name: '@prisma/client', versions: ['>=6.1.0'] }, (prisma, version) => {
+  const tracingHelper = new TracingHelper()
+
+  majorVersion = [...version][0]
```
BridgeAR:
```suggestion
addHook({ name: '@prisma/client', versions: ['>=6.1.0'] }, (prisma, version) => {
  const tracingHelper = new TracingHelper()
```

And we read the major version only where it's used. That way the locality of the variable is improved :)

packages/datadog-instrumentations/src/prisma.js
```diff
@@ -0,0 +1,115 @@
...
+
+  prisma.PrismaClient = PrismaClient
+  /*
+    * This is taking advantage of the built in tracing support from Prisma.
+    * The below variable is setting a global tracing helper that Prisma uses
+    * to enable OpenTelemetry.
+  */
+  // https://github.com/prisma/prisma/blob/478293bbfce91e41ceff02f2a0b03bb8acbca03e/packages/instrumentation/src/PrismaInstrumentation.ts#L42
+
+  global[`V${majorVersion}_PRISMA_INSTRUMENTATION`] = {
```
BridgeAR:
```suggestion
  const majorVersion = version.slice(0, version.indexOf('.'))
  global[`V${majorVersion}_PRISMA_INSTRUMENTATION`] = {
```

The former version splitting would not have worked with majors > 9 as it would have only picked up the first character, not the whole major number.

packages/datadog-plugin-prisma/src/client.js
```diff
@@ -0,0 +1,61 @@
...
+    const operationName = this.operationName({ operation: this.operation })
+    this.startSpan(operationName, options, ctx)
+
+    return ctx.currentStore
+  }
+
+  end (ctx) {
+    // Only synchronous operations would have `result` on `end`.
+    if (!Object.hasOwn(ctx, 'result')) return
+    this.finish(ctx)
```
BridgeAR: Personal style nit (I feel like this is easier to read)

```suggestion
    if (Object.hasOwn(ctx, 'result')) {
      this.finish(ctx)
    }
```

BridgeAR:
Some tests are still failing :)

packages/datadog-plugin-prisma/test/index.spec.js
```diff
@@ -118,7 +118,8 @@ describe('Plugin', () => {
           prismaClient = new prisma.PrismaClient()
           const matched = version.match(/(\d+)\.\d+\.\d+$/)
           const majorVersion = matched[1]
-          tracingHelper = global[`V${majorVersion}_PRISMA_INSTRUMENTATION`].helper
+          tracingHelper = global.PRISMA_INSTRUMENTATION.helper ||
```
BridgeAR: I am surprised this does not fail right now. Does this not need to be the following?

```suggestion
          tracingHelper = global.PRISMA_INSTRUMENTATION?.helper ||
```

BridgeAR (commented):
LGTM. Just not yet approving since v6.9 breaks our test setup

packages/datadog-plugin-prisma/test/index.spec.js
crysmags: I added and updated the next variable which was missing `global`

scripts/install_plugin_modules.js
```diff
@@ -145,22 +145,25 @@ async function assertPackage (name, version, dependencyVersionRange, external) {
...
-async function addDependencies (dependencies, name, versionRange) {
-  const versionList = await getVersionList(name)
+async function addDependencies (dependencies, externalName, versionRange) {
+  const versionList = await getVersionList(externalName)
   const version = semver.maxSatisfying(versionList, versionRange)
-  const pkgJson = await npmView(`${name}@${version}`)
-  for (const dep of externalDeps.get(name)) {
+  const pkgJson = await npmView(`${externalName}@${version}`)
+  for (const { dep, name } of externalDeps.get(externalName)) {
+    // do stuff with dep
```
BridgeAR, line 156:
```suggestion
```

BridgeAR (approved):
Still LGTM

rochdev (approved):

packages/dd-trace/src/supported-configurations.json
```diff
@@ -351,6 +351,7 @@
     "DD_TRACE_PLAYWRIGHT_CORE_ENABLED": ["A"],
     "DD_TRACE_PLAYWRIGHT_ENABLED": ["A"],
     "DD_TRACE_PLAYWRIGHT_TEST_ENABLED": ["A"],
+    "DD_TRACE_PRISMA_ENABLED":["A"],
```
watson: Nit:

```suggestion
    "DD_TRACE_PRISMA_ENABLED": ["A"],
```

BridgeAR (approved):
I just signed all commits and fixed the mentioned typo.

Still LG from my side

swordfish444:
This is amazing! When do we expect the release?

PR #5606, opened 2025-04-24
simple benchmark helper for adding operations per second

bengl:
Adds a benchmark helper for calculating the operations per second in a given benchmark, then sending that off to Sirun via statsd, for inclusion in the output.

### DIFF
--- a/benchmark/sirun/helpers.js
+++ b/benchmark/sirun/helpers.js
@@ -1,0 +1,23 @@
+const StatsD = require('./statsd')
+
+let statsd
+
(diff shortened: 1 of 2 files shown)


BridgeAR (dismissed):

BridgeAR:
The benchmark suite does not yet seem to accept this as is

BridgeAR:
Closing as superseded. The feature landed in benchmarking platform and I am integrating it into our startup guard to automatically report this metric as well.

PR #5609, opened 2025-04-24
add disabled by default integration concept

wconti27:
### What does this PR do?
Adds ability to disable an integration by default. To be used by integrations team to disable newly added integrations until considered stable.

### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].


packages/datadog-instrumentations/src/helpers/hooks.js
```diff
@@ -1,135 +1,144 @@
...
+    tedious: () => require('../tedious'),
+    undici: () => require('../undici'),
+    url: () => require('../url'),
+    vitest: { esmFirst: true, fn: () => require('../vitest') },
+    vm: () => require('../vm'),
+    when: () => require('../when'),
+    winston: () => require('../winston'),
+    workerpool: () => require('../mocha')
+  },
+  DISABLED_BY_DEFAULT_INTEGRATIONS
```
tlhunter, line 143: I don't think we'd need to make this object property all caps by default since it's no more of a constant than the `.hooks` object property:

```suggestion
  disabledByDefaultIntegrations: DISABLED_BY_DEFAULT_INTEGRATIONS
```

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -24,14 +25,27 @@ const disabledInstrumentations = new Set(
 )
 
 // Check for DD_TRACE_<INTEGRATION>_ENABLED environment variables
+const enabledIntegrationsOverrides = new Set()
 for (const [key, value] of Object.entries(process.env)) {
   const match = key.match(/^DD_TRACE_(.+)_ENABLED$/)
```
tlhunter, line 30: This pattern is a little unfortunate. For example, to enable the `@prisma/client` library, we'd need an env var named `DD_TRACE_@PRISMA/CLIENT_ENABLED`.

wconti27: The configuration uses the integration name. Assuming the prisma integration name used is `prisma`, the `disabledByDefaultIntegrations `(mapping of integration name to package name) would equal `{ prisma: @prisma/client }`. So the configuration to enable it would be `DD_TRACE_PRISMA_ENABLED`

tlhunter: Oh, I misunderstood. The disabled by default list is essentially a list of aliases. Got it.

tlhunter (changes requested):
needs tests

packages/datadog-instrumentations/src/helpers/hooks.js
rochdev: Why not just use a property instead similar to `serverless` and `esmFirst`?

tlhunter: If the `packages/datadog-instrumentations/src/*.js` files exported something instead of executing immediately we could export a `disabledByDefault=true` property (and `esmFirst` and `serverless`). This might be nicer?

rochdev: That would require loading the files which would significantly impact startup time.

wconti27: Ah I never caught that we can define properties in the hooks object, I like that idea @rochdev, easy to implement as well

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -19,16 +19,27 @@ const hooks = require('./hooks')
...
+  return name.replaceAll('@', '').replaceAll('-', '').replaceAll('/', '').replaceAll('_', '').toLowerCase()
+}
+
 const disabledInstrumentations = new Set(
-  DD_TRACE_DISABLED_INSTRUMENTATIONS ? DD_TRACE_DISABLED_INSTRUMENTATIONS.split(',') : []
+  DD_TRACE_DISABLED_INSTRUMENTATIONS
+    ? DD_TRACE_DISABLED_INSTRUMENTATIONS.split(',').map(convertInstrumentationName)
+    : []
 )
+const enabledInstrumentations = new Set()
```
tlhunter: I think this might be more clear:

```suggestion
const reenabledInstrumentations = new Set()
```

Specifically, it's not really a list of enabled instrumentations. There are many instrumentations that could be enabled that are outside of the list. Say, mysql or express. Instead it's a list of instrumentations that we force to be enabled in an attempt to modify their default behavior of being disabled. It's like a disdisable or undisable or reenable.

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -19,16 +19,27 @@ const hooks = require('./hooks')
...
 // Check for DD_TRACE_<INTEGRATION>_ENABLED environment variables
 for (const [key, value] of Object.entries(process.env)) {
   const match = key.match(/^DD_TRACE_(.+)_ENABLED$/)
   if (match && (value?.toLowerCase() === 'false' || value === '0')) {
-    const integration = match[1].toLowerCase()
+    const integration = convertInstrumentationName(match[1])
     disabledInstrumentations.add(integration)
+  } else if (match && (value?.toLowerCase() === 'true' || value === '1')) {
+    const integration = convertInstrumentationName(match[1])
+    enabledInstrumentations.add(integration)
```
tlhunter:
```suggestion
    reenabledInstrumentations.add(integration)
```

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -65,6 +77,13 @@ for (const packageName of names) {
     if (hook.serverless === false && isInServerlessEnvironment()) continue
 
     hookOptions.internals = hook.esmFirst
+
+    // some integrations are disabled by default, but can be enabled by setting
+    // the DD_TRACE_<INTEGRATION>_ENABLED environment variable to true
+    if (hook.disabled === true && !enabledInstrumentations.has(normalizedPackageName)) {
```
tlhunter:
```suggestion
    if (hook.disabled === true && !reenabledInstrumentations.has(normalizedPackageName)) {
```

packages/datadog-instrumentations/src/helpers/register.js
tlhunter: (optional nit)

tlhunter (approved):

tlhunter (approved):

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -19,16 +19,27 @@ const hooks = require('./hooks')
 const instrumentations = require('./instrumentations')
 const names = Object.keys(hooks)
 const pathSepExpr = new RegExp(`\\${path.sep}`, 'g')
+
+const convertInstrumentationName = (name) => {
+  return name.replaceAll('@', '').replaceAll('-', '').replaceAll('/', '').replaceAll('_', '').toLowerCase()
```
BridgeAR:
```suggestion
  return name.replaceAll(/[-@_/]/g, '').toLowerCase()
```

More important though: we have a different handling in our plugin manager. I believe we should reuse the same regular expression, otherwise we might end up detecting things slightly different over time.

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -19,16 +19,27 @@ const hooks = require('./hooks')
...
 for (const [key, value] of Object.entries(process.env)) {
   const match = key.match(/^DD_TRACE_(.+)_ENABLED$/)
   if (match && (value?.toLowerCase() === 'false' || value === '0')) {
-    const integration = match[1].toLowerCase()
+    const integration = convertInstrumentationName(match[1])
     disabledInstrumentations.add(integration)
+  } else if (match && (value?.toLowerCase() === 'true' || value === '1')) {
+    const integration = convertInstrumentationName(match[1])
+    reenabledInstrumentations.add(integration)
   }
```
BridgeAR:
```suggestion
if (match && value) {
  const integration = convertInstrumentationName(match[1])
  if (isFalse(value)) {
    disabledInstrumentations.add(integration)
  } else if (isTrue(value)) {
    reenabledInstrumentations.add(integration)
  }
}
````

In case the env is set, we can be quite confident that the user has it actively enabled or disabled. That is why I pulled it to the top. We also have the helper methods for isTrue and isFalse that we could use here.

**Review by BridgeAR — CHANGES_REQUESTED** [review]
This is almost LGTM! I do think it is important to use the same name as we define our environment variable for the plugins, so that should be addressed in my opinion before landing.

File: packages/datadog-instrumentations/src/helpers/register.js (hunk 7 above)
Comment (wconti27) [reply packages/datadog-instrumentations/src/helpers/register.js]: makes sense!

**Review by rochdev — APPROVED** [review]

File: packages/datadog-instrumentations/src/helpers/register.js [hunk 8]
```diff
@@ -22,16 +23,28 @@ const hooks = require('./hooks')
 const instrumentations = require('./instrumentations')
 const names = Object.keys(hooks)
 const pathSepExpr = new RegExp(`\\${path.sep}`, 'g')
+
+const convertInstrumentationName = (name) => {
+  return name.replaceAll(/[-@_/]/g, '').toLowerCase()
```
BridgeAR: Right now, the name used for the environment variable to activate or deactivate an instrumentation is in https://github.com/DataDog/dd-trace-js/blob/f8f0a9594401e04b485d68b27693a8eaba372558/packages/dd-trace/src/plugin_manager.js#L36

It works with the Plugin.id, which is defined in each Plugin with a getter. That is normally a 'simple' name for the instrumentation.
The actual names used here in line 24 https://github.com/DataDog/dd-trace-js/pull/5609/files#diff-0f4756798ec0b2564244375056c78d8ceca75a90a6b6dbc152d5468eb754c618R24 are actually the 'real' module names. So we effectively have a bug that most instrumentations that deviate with their simple ID abstraction (after substitution) won't be able to be deactivated using the `DD_TRACE_${instrumentation}_ENABLED` schema. It would work for the DD_TRACE_DISABLED_INSTRUMENTATIONS environment variable, since this does _not_ substitute anything of the name.

Where does this bring us?

A) We have to support both, the actual module name and the substituted ID name. Why can't we not unify it? The names could diverge and the substitution regular expression could produce a different result on the actual dependency name because the ID is manually defined.

B) The regular expression used here should be replaced with the one from the plugin_manager by placing it so, that it would be loaded in both places (in case it would ever be updated, it would be updated in both spots). One idea is to put it into the main util.js file. I am not sure if that is the right spot.

C) To check for both (Plugin.id and module name) we have to get access to that ID in here. I did not check how that is possible. We'd then have to build a new mapping `moduleName -> substituted(Plugin.id)` that is used to normalize the names before looking them up in the set and while reading the DD_TRACE_DISABLED_INSTRUMENTATIONS. Ideally, we just do that normalization in addition to another one that adds `substituted(readName)` as well. That way we are safe.

So yeah... a bit of a mess with the details...

Does this actually hurt if we don't do it? Most users will currently effectively not even know how to use those environment variables and will never do so. But I still believe we should make sure things work fully as expected.

BridgeAR (commented):
After looking into it a bit closer, I believe to have surfaced a couple of minor bugs that I think are worth fixing along the way. See my long comment :D 
Please also correct me, if I might have misread some of the code!

packages/datadog-instrumentations/src/helpers/register.js
wconti27: @BridgeAR We can enforce that plugin ids do not diverge from our file names to ensure this never becomes a problem! I added a new test suite that ensures that the `plugin id == plugin directory name == instrumentation file name`. This should alleviate any possible concerns of divergences, then we can just do the regex substitution and be ✅ .

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -68,6 +78,13 @@ for (const packageName of names) {
...
 
     hookOptions.internals = hook.esmFirst
+
+    // some integrations are disabled by default, but can be enabled by setting
+    // the DD_TRACE_<INTEGRATION>_ENABLED environment variable to true
+    if (hook.disabled === true && !reenabledInstrumentations.has(normalizedPackageName)) {
+      continue
+    }
+
     hook = hook.fn
```
BridgeAR:
```suggestion
    if (!hook.disabled || reenabledInstrumentations.has(normalizedPackageName)) {
      hook = hook.fn
    }
```

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -22,16 +23,24 @@ const hooks = require('./hooks')
 const instrumentations = require('./instrumentations')
 const names = Object.keys(hooks)
 const pathSepExpr = new RegExp(`\\${path.sep}`, 'g')
+
 const disabledInstrumentations = new Set(
-  DD_TRACE_DISABLED_INSTRUMENTATIONS ? DD_TRACE_DISABLED_INSTRUMENTATIONS.split(',') : []
+  DD_TRACE_DISABLED_INSTRUMENTATIONS
+    ? DD_TRACE_DISABLED_INSTRUMENTATIONS.split(',').map(name => normalizePluginEnvName(name, true))
+    : []
```
BridgeAR:
```suggestion
  DD_TRACE_DISABLED_INSTRUMENTATIONS?.split(',').map(name => normalizePluginEnvName(name, true))
```

packages/dd-trace/test/plugins/plugin-structure.spec.js
```diff
@@ -0,0 +1,104 @@
...
+
+        expect(instrumentationFiles.has(pluginId))
+          .to.equal(true, `Missing instrumentation file: ${pluginId}.js`)
+      })
+
+      it('should export a class with a static id getter', () => {
+        expect(Plugin).to.be.a('function')
+        expect(Plugin.id).to.be.a('string')
+        expect(Plugin.id).to.not.be.empty
+      })
```
BridgeAR:
```suggestion
```

This would not be reached if that would not be the case :)

packages/dd-trace/test/plugins/plugin-structure.spec.js
```diff
@@ -0,0 +1,104 @@
...
+  const instrumentationsDir = path.join(packagesDir, 'datadog-instrumentations', 'src')
+
+  let pluginDirs
+  let allPluginIds
+  let instrumentationFiles
+
+  before(() => {
+    pluginDirs = fs.readdirSync(packagesDir)
+      .filter(dir => dir.startsWith('datadog-plugin-'))
+      .filter(dir => {
```
BridgeAR: This is just a test file, so it's not important here while the functional style like that always creates a new array on each step. Especially in our production code we always have to focus on preventing overhead, thus, I recommend to not use such pattern in general to get used to preventing it :)

Here, a single filter would be the right way while it's often even better to use a simple loop. Below, where a filter is combined with a map would be an example where a simple loop would be better (while the loop would just add the entry directly to the set, skipping creating any intermediate array).

packages/dd-trace/test/plugins/plugin-structure.spec.js
```diff
@@ -0,0 +1,104 @@
...
+
+      if (seenIds.has(id)) {
+        duplicates.push(id)
+      } else {
+        seenIds.add(id)
+      }
+    })
+
+    expect(duplicates).to.be.empty
+  })
```
BridgeAR: The ID is checked above against the plugin directory name. That way it's already guaranteed to be unique :)

```suggestion
```

packages/dd-trace/test/plugins/plugin-structure.spec.js
```diff
@@ -0,0 +1,104 @@
...
+  it('should have all plugins accounted for in instrumentations', () => {
+    const missingInstrumentations = []
+
+    allPluginIds.forEach(pluginId => {
+      if (!instrumentationFiles.has(pluginId) && !abstractPlugins.includes(pluginId)) {
+        missingInstrumentations.push(pluginId)
+      }
+    })
+
+    expect(missingInstrumentations).to.be.empty
```
BridgeAR: Nit (not important, just simpler doing something like that)

```suggestion
    allPluginIds.forEach(pluginId => {
      assert.ok(!instrumentationFiles.has(pluginId) && !abstractPlugins.includes(pluginId), `Missing id ${pluginId}`)
    })
```

Would it actually be possible to test it the other way around? That all instrumentations also have a plugin? I am not sure if that's the case.

packages/dd-trace/test/plugins/plugin-structure.spec.js
```diff
@@ -0,0 +1,104 @@
...
+  let pluginDirs
+  let allPluginIds
+  let instrumentationFiles
+
+  before(() => {
+    pluginDirs = fs.readdirSync(packagesDir)
+      .filter(dir => dir.startsWith('datadog-plugin-'))
+      .filter(dir => {
+        const pluginPath = path.join(packagesDir, dir, 'src', 'index.js')
+        return fs.existsSync(pluginPath)
```
BridgeAR: How do we handle the case when a plugin has no such file? There is no id as such, if I am not mistaken but how do we get that instead?

packages/dd-trace/test/plugins/plugin-structure.spec.js
```diff
@@ -0,0 +1,104 @@
...
+      })
+
+    allPluginIds = new Set()
+    instrumentationFiles = new Set(
+      fs.readdirSync(instrumentationsDir)
+        .filter(file => file.endsWith('.js'))
+        .map(file => file.replace('.js', ''))
+    )
```
BridgeAR, line 45: I think it would be best to use `Object.keys(require('packages/datadog-instrumentations/src/helpers/hooks.js'))`

That way we make sure the name of these is actually aligned, since we also use those names later on and they are currently defined manually.

BridgeAR (commented):
This is mostly LGTM. Can you please also remove `DD_TRACE__CONFLUENTINC_KAFKA_JAVASCRIPT_ENABLED` from our `supported-configurations.json` file? :)

It won't be used anymore after the renaming here. I think it's more of a bug fix with the renaming and not a breaking change as such.

I did leave a few comments to simplify the tests and some code.

packages/dd-trace/test/plugins/plugin-structure.spec.js
wconti27: Done.

packages/dd-trace/test/plugins/plugin-structure.spec.js
wconti27: Testing this the other way doesn't really work, since tracing plugins are a subset of instrumentations. Some instrumentations are purely for appsec / other products, and they don't have an associated tracing plugin.

packages/dd-trace/test/plugins/plugin-structure.spec.js
wconti27: Makes sense, I've removed this test

packages/dd-trace/test/plugins/plugin-structure.spec.js
wconti27: Okay, I've made this change. good to know.

packages/datadog-instrumentations/src/helpers/register.js
wconti27: Fixed.

packages/datadog-instrumentations/src/helpers/register.js
wconti27: Fixed.

packages/dd-trace/test/plugins/plugin-structure.spec.js
wconti27: I've removed this test case.

packages/dd-trace/test/plugins/plugin-structure.spec.js
wconti27: We can't do exactly this change, since the hook key corresponds to the module, and our plugin name is a different pattern. Instead, I have added a new test case, that compares the normalized name for the hook key against the normalized name for the plugin name. This should ensure consistency in that same way that we normalize when registering the hooks.

packages/dd-trace/test/plugins/plugin-structure.spec.js
wconti27: I have made the test more explicit, and added an allowlist of a few plugin directories which exist (`axios`, `limitd-client`, `mongoose`, which we have plugin dirs for but no actual plugin implemented). I also added comments explaining why they exist. These are special cases and should not be the norm.

packages/dd-trace/test/plugins/plugin-structure.spec.js
```diff
@@ -0,0 +1,97 @@
...
+    // since module names can have characters like @, /, we need to normalize them to compare to the plugin names
+
+    const missingHooks = []
+
+    allPluginIds.forEach(pluginId => {
+      const normalizedPluginId = normalizePluginEnvName(pluginId)
+      if (!normalizedInstrumentationHooks.has(normalizedPluginId) && !abstractPlugins.includes(normalizedPluginId)) {
+        missingHooks.push(normalizedPluginId)
+      }
+    })
```
BridgeAR, line 106: This is likely meant to still have an assertion at the end? :)

wconti27: fixed.

BridgeAR (approved):

packages/dd-trace/test/plugins/plugin-structure.spec.js
wconti27: fixed.

packages/dd-trace/test/plugins/plugin-structure.spec.js
```diff
@@ -0,0 +1,114 @@
...
+
+    // some hooks are for submodules, so we should see if the pluginId is a submodule of the hook
+    // e.g: '@jest/core' is a submodule of 'jest'. 'jest' is our plugin id
+    normalizedInstrumentationHooks.forEach(hook => {
+      missingHooks.forEach(pluginId => {
+        if (hook.includes(pluginId)) {
+          missingHooks.splice(missingHooks.indexOf(pluginId), 1)
+        }
+      })
+    })
```
BridgeAR: This is not a super safe test, since the name could theoretically conflict.
It should be very rare though.

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -68,6 +76,13 @@ for (const packageName of names) {
...
 
     hookOptions.internals = hook.esmFirst
+
+    // some integrations are disabled by default, but can be enabled by setting
+    // the DD_TRACE_<INTEGRATION>_ENABLED environment variable to true
+    if (hook.disabled && !reenabledInstrumentations.has(normalizedPackageName)) {
+      continue
+    }
+
     hook = hook.fn
```
BridgeAR:
```suggestion
    if (!hook.disabled || reenabledInstrumentations.has(normalizedPackageName)) {
      hook = hook.fn
    }
```

BridgeAR (dismissed):
This improves the status quo a lot, so I think we should merge it as is. @wconti27 thanks for sticking to this!
We are still not going to match everything as I would expect it to match though, so I think we need to further refine some details later on.

packages/dd-trace/test/plugins/plugin-structure.spec.js
wconti27: test has been fixed.

packages/datadog-instrumentations/src/helpers/register.js
wconti27: we can't make this change since we need to continue if the instrumentation is disabled.

PR #5613, opened 2025-04-25
cut down shimmer-startup benchmark so it can run in benchmarking platform

bengl:
These benchmarks were taking too long and causing builds to be flaky.

### DIFF
--- a/benchmark/sirun/shimmer-startup/index.js
+++ b/benchmark/sirun/shimmer-startup/index.js
@@ -8,7 +8,7 @@ const {
   FUNCTION_NAME
 } = process.env
 
-const ITERATIONS = 1e6
+const ITERATIONS = 1e5
 
 let counter = 0
 function declared () {
--- a/benchmark/sirun/shimmer-startup/index.js
+++ b/benchmark/sirun/shimmer-startup/index.js
@@ -8,7 +8,7 @@ const {
   FUNCTION_NAME
 } = process.env
 
-const ITERATIONS = 2e6
+const ITERATIONS = 1e6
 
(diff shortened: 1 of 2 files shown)


watson (approved):

benchmark/sirun/shimmer-startup/index.js
```diff
@@ -8,7 +8,7 @@ const {
   FUNCTION_NAME
 } = process.env
 
-const ITERATIONS = 2e6
+const ITERATIONS = 1e6
```
BridgeAR, line 11:
```suggestion
const ITERATIONS = 1e5
```

benchmark/sirun/shimmer-startup/meta.json
```diff
@@ -3,7 +3,7 @@
   "run": "node index.js",
   "run_with_affinity": "bash -c \"taskset -c $CPU_AFFINITY node index.js\"",
   "cachegrind": false,
-  "iterations": 10,
+  "iterations": 5,
```
BridgeAR, line 6: Iterations should actually normally be much higher. Otherwise the variance is going to be high. I would normally go for something like 100 or above.

```suggestion
  "iterations": 20,
```

ddyurchenko: +1, I would set iterations closer to 20, as 5 is definitely not enough.

> I would normally go for something like 100 or above.
Same. But given how long execution takes, we have to be a bit more laxed (sadly).

watson (approved):

BridgeAR (approved):

PR #5614, opened 2025-04-25
Address cookie vulnerability cardinality

uurien:
### What does this PR do?

Stop using cookie name as hash key, and calculate cookie hashes in the same way that the others are calculated.

### Motivation

In some applications, cookie names could have random values or timestamps, generating infinite vulnerabilities.

### Additional Notes

[APPSEC-56386]

[APPSEC-56386]: https://datadoghq.atlassian.net/browse/APPSEC-56386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dd-trace/src/appsec/iast/analyzers/cookie-analyzer.js
CarlesDD, line 1: All this code could be deleted now, since there is no filtering in cookies.

https://github.com/DataDog/dd-trace-js/pull/5614/files#diff-c4757f0d561f029a40893d548f6778f1801b7f27fabce9d8bebf7eefa2672d8bR16-R21

It also affects to the config. `iast.cookieFilterPattern` `DD_IAST_COOKIE_FILTER_PATTERN`are useless now.

CarlesDD (changes requested):

packages/dd-trace/src/appsec/iast/analyzers/cookie-analyzer.js
CarlesDD: Moreover, `CookieAnalyzer` was the only `Analyzer` that needed the tracer configuration in the `onConfigure` method, so I think we can get rid of this argument from [IastPlugin](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/appsec/iast/iast-plugin.js#L117) to all its subclasses.

Issue #5612, opened 2025-04-25
[BUG]: Next.js 15 app becomes unresponsive after a while


mintuhouse:
### Tracer Version(s)

5.49

### Node.js Version(s)

22.15.0

### Bug Report

We are deploying a standalone build of nextjs 15 with app router
It was working fine until we are on `5.37.1`
Last month we upgraded it to `5.41.1` (then latest)
But we observed it caused memory leak in our application ([issue1](https://github.com/DataDog/dd-trace-js/issues/5554), [issue2](https://github.com/DataDog/dd-trace-js/issues/5394))
A few days ago, we upgraded to `5.48.1` to fix this leak
But the entire app gets unresponsive after an hour or so (reproducible on `5.49.0` too)

### Reproduction Code

next.js ( reproducible on both 15.2 & 15.3) app router standalone output build

src/lib/tracer.ts
```ts
import tracer from "dd-trace"

// refer: https://docs.datadoghq.com/tracing/trace_collection/automatic_instrumentation/dd_libraries/nodejs/#typescript-and-bundlers

const ddtracer = tracer.init() // initialized in a different file to avoid hoisting
const provider = new ddtracer.TracerProvider()
provider.register()

export { default } from "dd-trace"
````

src/instrumenation.ts
```ts
// https://nextjs.org/docs/app/building-your-application/optimizing/instrumentation
// https://vercel.com/docs/observability/otel-overview
export async function register() {
  // To avoid errors with built-in modules not being defined https://github.com/vercel/next.js/discussions/16600#discussioncomment-7204740
  if (process.env.NEXT_RUNTIME === "nodejs") {
    if (process.env.DD_API_KEY) {
      await import("./lib/tracer") // must come before importing any instrumented module
    }
  }
}

```

### Tracer Config

DD_SERVICE=webui DD_VERSION=$REVISION DD_SITE="datadoghq.com" 
DD_LOGS_INJECTION=true DD_LOGS_ENABLED=true DD_TRACE_ENABLED=true DD_TRACE_SAMPLE_RATE=1.0 DD_PROFILING_ENABLED=true DD_IAST_ENABLED=true DD_APPSEC_ENABLED=true DD_DYNAMIC_INSTRUMENTATION_ENABLED=true DD_RUNTIME_METRICS_ENABLED=true DD_EXCEPTION_DEBUGGING_ENABLED=true

### Operating System

Linux x86_64 GNU/Linux

### Bundling

Next.js

[...]

Issue #5565, opened 2025-04-10
[BUG]: Instance Failures in 5.46.0


MMShep97:
### Tracer Version(s)

5.46.0

### Node.js Version(s)

22.14.0

### Bug Report

Posting for visibility -- this library is causing us grief again with our automerges; this time with failed instances. Going to be setting a special rule for `dd-trace` for the time being.

![Image](https://github.com/user-attachments/assets/d9d440a3-fa37-409c-a4bd-8f89f1ddd4e9)

Disregard the lack of legend -- You can see the pattern of before 5.46.0 going in, during, and after reverting. The memory min going to 0 correlates to our ECS tasks failing.

This occurred when our automerger updated the lib from `5.45.0` to `5.46.0`

### Tracer Config

```typescript
import tracer from "dd-trace";

const ddTrace = tracer.init();
ddTrace.use('aws-sdk');
ddTrace.use('hapi');
ddTrace.use('http', {
    blocklist: [
        /^\/health(.*)/i
    ]
});
ddTrace.use('redis');
```

Here's our env variables inside of our task definition:

```json
{
      {
        "name": "DD_ENV",
        "value": "${environment_name}"
      },
      {
        "name": "DD_SERVICE",
        "value": "${project_name}"
      },
      {
        "name": "DD_VERSION",
        "value": "git:${git_commit}"
      },
      {
        "name": "DD_RUNTIME_METRICS_ENABLED",
        "value": "true"
      },
      {
        "name": "DD_IAST_ENABLED",
        "value": "true"
      },
      {
        "name": "DD_APPSEC_ENABLED",
        "value": "true"
      },
      {
        "name": "DD_DATA_STREAMS_ENABLED",
        "value": "true"
      },
      {
        "name": "DD_APM_ENABLED",
        "value": "true"
      },
      {
        "name": "DD_TRACE_ENABLED",
        "value": "true"
      },
      {
        "name": "DD_RUNTIME_METRICS_ENABLED",
        "value": "true"
      },
      {
        "name": "DD_GIT_REPOSITORY_URL",
        "value": "${github_repo_url}"
      },
      {
        "name": "DD_PROFILING_ENABLED",
        "value": "true"
      },
      {
          "name": "DD_DBM_PROPAGATION_MODE",
          "value": "full"
      },
      {
          "name": "DD_CRASHTRACKING_ENABLED",
          "value": "false"
      },
      {
          "name": "DD_TRACE_DEBUG",
          "value": "true"
[...]

PR #5621, opened 2025-04-28
Turn async ID gathering off by default

szegedi:
### What does this PR do?
Sets `DD_PROFILING_ASYNC_ID_ENABLED` to disabled by default.

### Motivation
Customers have problems with it enabled by default. See the Jira issue for more details. It mostly hinges on non-safety of calling `node::AsyncHooksGetExecutionAsyncId(v8::Isolate*)` under certain circumstances. We should re-enable this after we can call `node::AsyncHooksGetExecutionAsyncId(v8::Local<v8::Context>)` from https://github.com/nodejs/node/pull/57820 with a safely pre-obtained `v8::Context` object.

### Additional Notes
Jira: [PROF-11628]

[PROF-11628]: https://datadoghq.atlassian.net/browse/PROF-11628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integration-tests/profiler/codehotspots.js
+++ b/integration-tests/profiler/codehotspots.js
@@ -11,19 +11,13 @@ const busyCycleTime = BigInt(process.env.BUSY_CYCLE_TIME)
 
 function busyLoop () {
   const start = process.hrtime.bigint()
-  let x = 0
   for (;;) {
     const now = process.hrtime.bigint()
     // Busy cycle
     if (now - start > busyCycleTime) {
       break
     }
-    // Do something in addition to invoking hrtime
-    for (let i = 0; i < 1000; i++) {
-      x += Math.sqrt(Math.random() * 2 - 1)
-    }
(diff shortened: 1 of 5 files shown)


BridgeAR (changes requested):
If the feature is actually unstable in all current Node.js versions, let's just plain out prevent usage for now. I don't think we should enable this.

After looking at recent github issues, I fear this might be related to these reports:

https://github.com/DataDog/dd-trace-js/issues/5565
https://github.com/DataDog/dd-trace-js/issues/5612

nsavoire (dismissed):

BridgeAR (approved):
Thanks for the quick follow-up :)
Reverting does indeed seem like the cleanest way!

PR #5631, opened 2025-04-30
Improve log safety

BridgeAR:
The logger instrumentation is using a proxy. Calling Reflect.get with a receiver might actually cause issues at times. Instead, just access the target directly. That is easier and safer in this situation.
On top, remove the special handling for Symbol.toStringTag. That was added for the testing library and should not have been there in the first place.
As a drive-by fix, it also improves the performance / cleans up the code.

### DIFF
--- a/packages/datadog-plugin-winston/test/index.spec.js
+++ b/packages/datadog-plugin-winston/test/index.spec.js
@@ -5,6 +5,7 @@ const agent = require('../../dd-trace/test/plugins/agent')
 const http = require('http')
 const { expect } = require('chai')
 const proxyquire = require('proxyquire').noPreserveCache()
+const { inspect } = require('util')
 
 function createLogServer () {
   return new Promise((resolve, reject) => {
@@ -194,7 +195,6 @@ describe('Plugin', () => {
                 span_id: span.context().toSpanId()
               }
             }
-
             const error = new Error('boom')
 
             tracer.scope().activate(span, () => {
@@ -203,7 +203,7 @@ describe('Plugin', () => {
               const index = semver.intersects(version, '>=3') ? 0 : 2
               const record = log.firstCall.args[index]
 
-              expect(record).to.be.an('error')
+              expect(record).to.be.an.instanceof(Error)
               expect(error).to.not.have.property('dd')
               expect(spy).to.have.been.calledWithMatch(meta.dd)
             })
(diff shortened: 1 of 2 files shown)


packages/datadog-plugin-winston/test/index.spec.js
```diff
@@ -203,14 +204,42 @@ describe('Plugin', () => {
               const index = semver.intersects(version, '>=3') ? 0 : 2
               const record = log.firstCall.args[index]
 
-              expect(record).to.be.an('error')
+              assert(record instanceof Error)
```
watson, line 207: Why not

```suggestion
              expect(record).to.be.an.instanceof(Error)
```

packages/datadog-plugin-winston/test/index.spec.js
```diff
@@ -203,14 +204,42 @@ describe('Plugin', () => {
...
               expect(error).to.not.have.property('dd')
               expect(spy).to.have.been.calledWithMatch(meta.dd)
             })
             expect(await logServer.logPromise).to.include(meta.dd)
           })
 
           if (semver.intersects(version, '>=3')) {
+            it('should support sets and getters', async () => {
```
watson, line 215:
```suggestion
            it('should support sets with getters', async () => {
```

watson (commented):
I thought you wanted to get rid of the `Proxy` entirely? Does this change fully address your original concerns?

BridgeAR:
@watson 
> I thought you wanted to get rid of the Proxy entirely? Does this change fully address your original concerns?

That would be ideal. When looking into why this was done, it did seem like the best short term trade-off to keep the proxy but to not access the receiver anymore. That addresses at least the issues that we recently faced.

packages/datadog-plugin-winston/test/index.spec.js
BridgeAR: That should also work. I try to just prevent using chai in general. It often does not work the way I would expect things to work. Using plain checks always work.

watson: I agree our tests are currently a mess. But isn't it less of a mess if we at least keep to using the same assertion library in the same file, or at least within the same test?

BridgeAR: It is a bit messy, I agree. It was more the way of introducing the assert library more. I think we should make more use of it instead of chai.

packages/datadog-plugin-winston/test/index.spec.js
```diff
@@ -5,6 +5,8 @@ const agent = require('../../dd-trace/test/plugins/agent')
 const http = require('http')
 const { expect } = require('chai')
 const proxyquire = require('proxyquire').noPreserveCache()
+const assert = require('assert/strict')
```
BridgeAR, line 8:
```suggestion
```

packages/datadog-plugin-winston/test/index.spec.js
watson: Ah, ok. I misunderstood. I thought it was specifically getters on a set. Carry on 👍

rochdev (approved):

PR #5632, opened 2025-04-30
Safer shimmer implementation

BridgeAR:
This improves the shimmer by replacing getters safely from now on.
The getters were always replaced before. Now, they are hooked into
by default and only replaced, if a developer opts into it.

This also simplifies the usage due to not needed workarounds to
instrument getters regularly otherwise.

### DIFF
--- a/packages/datadog-instrumentations/src/helpers/register.js
+++ b/packages/datadog-instrumentations/src/helpers/register.js
@@ -43,7 +43,7 @@ if (!disabledInstrumentations.has('process')) {
   require('../process')
 }
 
-const HOOK_SYMBOL = Symbol('hookExportsMap')
+const HOOK_SYMBOL = Symbol('hookExportsSet')
 
 if (DD_TRACE_DEBUG && DD_TRACE_DEBUG.toLowerCase() !== 'false') {
   checkRequireCache.checkForRequiredModules()
@@ -89,11 +89,14 @@ for (const packageName of names) {
         fullFilePattern = filename(name, fullFilePattern)
       }
 
-      // Create a WeakMap associated with the hook function so that patches on the same moduleExport only happens once
+      // Create a WeakSet associated with the hook function so that patches on the same moduleExport only happens once
       // for example by instrumenting both dns and node:dns double the spans would be created
-      // since they both patch the same moduleExport, this WeakMap is used to mitigate that
+      // since they both patch the same moduleExport, this WeakSet is used to mitigate that
+      // TODO(BridgeAR): Instead of using a WeakSet here, why not just use aliases for the hook in register?
+      // That way it would also not be duplicated. The actual name being used has to be identified else wise.
+      // Maybe it is also not important to know what name was actually used?
       if (!hook[HOOK_SYMBOL]) {
-        hook[HOOK_SYMBOL] = new WeakMap()
+        hook[HOOK_SYMBOL] = new WeakSet()
       }
       let matchesFile = false
 
@@ -114,7 +117,7 @@ for (const packageName of names) {
           log.error('Error getting version for "%s": %s', name, e.message, e)
           continue
         }
-        if (typeof namesAndSuccesses[`${name}@${version}`] === 'undefined') {
+        if (namesAndSuccesses[`${name}@${version}`] === undefined) {
           // TODO If `file` is present, we might elsewhere instrument the result of the module
           // for a version range that actually matches, so we can't assume that we're _not_
           // going to instrument that. However, the way the data model around instrumentation
@@ -140,12 +143,19 @@ for (const packageName of names) {
             loadChannel.publish({ name, version, file })
             // Send the name and version of the module back to the callback because now addHook
             // takes in an array of names so by passing the name the callback will know which module name is being used
-            moduleExports = hook(moduleExports, version, name)
-            // Set the moduleExports in the hooks weakmap
-            hook[HOOK_SYMBOL].set(moduleExports, name)
+            // TODO(BridgeAR): This is only true in case the name is identical
+            // in all loads. If they deviate, the deviating name would not be
+            // picked up due to the unification. Check what modules actually use the name.
+            // TODO(BridgeAR): Only replace moduleExports if the hook returns a new value.
+            // This allows to reduce the instrumentation code (no return needed).
+            const newModuleExports = hook(moduleExports, version, name)
+            if (newModuleExports) {
+              moduleExports = newModuleExports
+            }
+            // Set the moduleExports in the hooks WeakSet
+            hook[HOOK_SYMBOL].add(moduleExports)
           } catch (e) {
-            log.info('Error during ddtrace instrumentation of application, aborting.')
-            log.info(e)
+            log.info('Error during ddtrace instrumentation of application, aborting.', e)
             telemetry('error', [
               `error_type:${e.constructor.name}`,
               `integration:${name}`,
--- a/packages/datadog-instrumentations/src/helpers/register.js
+++ b/packages/datadog-instrumentations/src/helpers/register.js
@@ -95,9 +95,7 @@ for (const packageName of names) {
       // TODO(BridgeAR): Instead of using a WeakSet here, why not just use aliases for the hook in register?
       // That way it would also not be duplicated. The actual name being used has to be identified else wise.
       // Maybe it is also not important to know what name was actually used?
-      if (!hook[HOOK_SYMBOL]) {
-        hook[HOOK_SYMBOL] = new WeakSet()
-      }
+      hook[HOOK_SYMBOL] ??= new WeakSet()
       let matchesFile = false
 
       matchesFile = moduleName === fullFilename
@@ -148,10 +146,7 @@ for (const packageName of names) {
             // picked up due to the unification. Check what modules actually use the name.
             // TODO(BridgeAR): Only replace moduleExports if the hook returns a new value.
             // This allows to reduce the instrumentation code (no return needed).
-            const newModuleExports = hook(moduleExports, version, name)
-            if (newModuleExports) {
-              moduleExports = newModuleExports
-            }
+            moduleExports = hook(moduleExports, version, name) ?? moduleExports
             // Set the moduleExports in the hooks WeakSet
             hook[HOOK_SYMBOL].add(moduleExports)
           } catch (e) {
--- a/packages/datadog-instrumentations/src/jest.js
+++ b/packages/datadog-instrumentations/src/jest.js
@@ -635,7 +635,7 @@ addHook({
 })
 
 function cliWrapper (cli, jestVersion) {
-  const wrapped = shimmer.wrap(cli, 'runCLI', runCLI => async function () {
+  shimmer.wrap(cli, 'runCLI', runCLI => async function () {
     let onDone
     const configurationPromise = new Promise((resolve) => {
       onDone = resolve
@@ -873,8 +873,6 @@ function cliWrapper (cli, jestVersion) {
 
     return result
   })
-
-  cli.runCLI = wrapped.runCLI
 
   return cli
 }
--- a/packages/datadog-instrumentations/src/url.js
+++ b/packages/datadog-instrumentations/src/url.js
@@ -77,6 +77,4 @@ addHook({ name: names }, function (url) {
       }
     })
   }
-
-  return url
 })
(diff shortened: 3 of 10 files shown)


packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -131,12 +132,19 @@ for (const packageName of names) {
...
-            hook[HOOK_SYMBOL].set(moduleExports, name)
+            // TODO(BridgeAR): This is only true in case the name is identical
+            // in all loads. If they deviate, the deviating name would not be
+            // picked up due to the unification. Check what modules actually use the name.
+            // TODO(BridgeAR): Only replace moduleExports if the hook returns a new value.
+            // This allows to reduce the instrumentation code (no return needed).
+            const newModuleExports = hook(moduleExports, version, name)
+            if (newModuleExports) {
+              moduleExports = newModuleExports
+            }
```
watson, line 143: Slight readability improvement:

```suggestion
            moduleExports = hook(moduleExports, version, name) || moduleExports
```

Of if you actually only care about `null` and `undefined`, it could be:

```suggestion
            moduleExports = hook(moduleExports, version, name) ?? moduleExports
```

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -73,25 +89,61 @@ function wrap (target, name, wrapper) {
...
+      // replaceGetter may only be used when the getter has no side effect.
+      if (replaceGetter === 'REPLACE_GETTER') {
+        if (descriptor.set) {
+          throw new Error('Cannot replace getter due to potential side effects with the setter')
+        }
+        descriptor.get = () => wrapped
+      } else {
+        descriptor.get = wrapped
+      }
+    } else if (descriptor.set) {
```
IlyasShabi, line 102: If the descriptor only has a setter, shouldn’t we replace it with wrapped?

packages/datadog-instrumentations/src/helpers/register.js
BridgeAR: Good point!

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -44,18 +59,19 @@ function wrapFunction (original, wrapper) {
   return wrapped
 }
 
-function wrap (target, name, wrapper) {
-  assertMethod(target, name)
+function wrap (target, name, wrapper, replaceGetter) {
```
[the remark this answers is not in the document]
BridgeAR: I'll update the PR with the suggestion :)

packages/datadog-shimmer/src/shimmer.js
BridgeAR: This made me think about fixing the replace case with setters as well.

It would work as long as the setter is never called, so we could just automatically undo the wrapping in case the setter is called for all following calls. We could even reinstrument the newly set part in that case and the unwrapping would not be needed in that case.

I wonder if we want to do that. Looping in @rochdev @bengl and @watson as well.

packages/datadog-instrumentations/src/jest.js
```diff
@@ -849,8 +849,6 @@ function cliWrapper (cli, jestVersion) {
     return result
   })
 
-  cli.runCLI = wrapped.runCLI
```
watson, line 852: I can't understand why this was here previously. Do you know?

packages/datadog-instrumentations/src/url.js
```diff
@@ -83,6 +77,4 @@ addHook({ name: names }, function (url) {
       }
     })
   }
-
-  return url
```
watson, line 87: Why remove this return value?

packages/datadog-shimmer/src/shimmer.js
watson: @BridgeAR I assume the 2nd of your two comments is about the case there's also a getter?

packages/datadog-instrumentations/src/url.js
BridgeAR: It was pretty much my test case for simplifying the implementation code.

packages/datadog-instrumentations/src/jest.js
BridgeAR: No, I think it was just a mistake. It didn't change anything anyway.

packages/datadog-shimmer/src/shimmer.js
BridgeAR: Yes, correct.

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -73,25 +89,67 @@ function wrap (target, name, wrapper) {
...
+          // either undos the original get replacement (that should be safe to
+          // do, we just loose the instrumentation afterwards) or that
+          // reinstruments the passed through value after calling get. The
+          // latter might however have side effects due to maybe not replacing
+          // the value that the getter instrumented.
+          throw new Error('Replacing getter with sets is not supported. Implement if required.')
+        }
+        descriptor.get = () => wrapped
+      } else {
+        descriptor.get = wrapped
```
watson, line 106: Just a mental note to help me understand the flow (maybe it warrants a code comment): We can only get here if `replaceGetter: false` and the property has a getter. There might also be a setter. In this case `wrapped` is a product of `wrapper(descriptor.get)`.

If you want to add a comment, maybe this one:

```suggestion
        // State: Property has a getter and `replaceGetter: false`.
        // The `wrapped` property is a product of `wrapper(descriptor.get)`
        descriptor.get = wrapped
```

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -44,18 +59,19 @@ function wrapFunction (original, wrapper) {
   return wrapped
 }
 
-function wrap (target, name, wrapper) {
-  assertMethod(target, name)
+function wrap (target, name, wrapper, options) {
```
watson, line 62: I would prefer that we add a JSDoc describing what `replaceGetter` does

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -73,25 +89,67 @@ function wrap (target, name, wrapper) {
...
+        if (descriptor.set) {
+          // This case is possible by replacing the setter with a wrapper that
+          // either undos the original get replacement (that should be safe to
+          // do, we just loose the instrumentation afterwards) or that
+          // reinstruments the passed through value after calling get. The
+          // latter might however have side effects due to maybe not replacing
+          // the value that the getter instrumented.
+          throw new Error('Replacing getter with sets is not supported. Implement if required.')
+        }
+        descriptor.get = () => wrapped
```
watson, line 104: Just a mental note to help me understand the flow (maybe it warrants a code comment): We can only get here if `replaceGetter: true` and the property has only a getter but not a setter. In this case `wrapped` is the product of `wrapper(target[name])`, which means that the getter was triggered and its return value was wrapped by `wrapper`.

If you want to add a comment, maybe this one:

```suggestion
        // State: Property has a getter (but no setter) and `replaceGetter: true`.
        // The `wrapped` property is a product of `wrapper(target[name])`. Original getter is triggered
        descriptor.get = () => wrapped
```

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -44,18 +59,19 @@ function wrapFunction (original, wrapper) {
...
-  assertMethod(target, name)
+function wrap (target, name, wrapper, options) {
   if (typeof wrapper !== 'function') {
     throw new Error(wrapper ? 'Target is not a function' : 'No function provided')
   }
 
-  const original = target[name]
-  const wrapped = wrapper(original)
+  let descriptor = Object.getOwnPropertyDescriptor(target, name)
+  const original = descriptor?.get && (!options?.replaceGetter || descriptor.set) ? descriptor.get : target[name]
```
watson, line 68: I think the logic here is hard to follow and it would benefit with a code comment:

```suggestion
  // Replacing the getter is not supported if the property also has a setter
  const original = descriptor?.get && (!options?.replaceGetter || descriptor.set) ? descriptor.get : target[name]
```

packages/datadog-shimmer/src/shimmer.js
watson: On one hand I agree that there's no reason to implement support for a feature we currently don't need, and for now `replaceGetter` is only set to `true` in `packages/datadog-instrumentations/src/next.js`. On the other hand, the logic flow in this code is kinda tricky and implementing this in the future when all the context has been offloaded from our brains, will take extra time and there's a risk we'll do it wrong because of that. So if it's not too much trouble and if you don't mind, I'd say go for it. But I can also support leaving it as is.

packages/datadog-shimmer/src/shimmer.js
BridgeAR: I think the current code should be fine due to throwing the error with a similar message right above.

packages/datadog-shimmer/src/shimmer.js
BridgeAR: I thought about it a few times and decided not to implement it for now due to A) we have no actual use case and B) ideally, we soon do not use shimmer anymore and instead rely on Orchestrion-JS to just rewrite the method directly.

packages/datadog-shimmer/src/shimmer.js
BridgeAR: I restructured the code a bit and I believe it is now easier to follow in general.

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -44,54 +82,112 @@ function wrapFunction (original, wrapper) {
...
+      // Just reuse the already created object.
+      let moduleExports = nonConfigurableModuleExports.get(target)
+      if (!moduleExports) {
+        if (typeof target === 'function') {
+          const original = target
+          moduleExports = function (...args) { return original.apply(original, args) }
+          // This is a rare case. Accept the slight performance hit.
+          skipMethods.add(name)
+          copyProperties(target, moduleExports)
+          skipMethods.delete(name)
```
watson, line 183: If `name` is already in `skipMethods`, it will wrongly be deleted. Only if it's a new name should we perform the add/delete action:

```suggestion
          const notAlreadySkipped = !skipMethods.has(name)
          if (notAlreadySkipped) skipMethods.add(name)
          copyProperties(target, moduleExports)
          if (notAlreadySkipped) skipMethods.delete(name)
```

You could argue that we'll never call `wrap` with `caller`, `arguments`, `name`, or `length`, in which case this will not be an issue. But it feels unstable 🤷

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -111,26 +217,48 @@ function massWrap (targets, names, wrapper) {
...
+function assertMethod (target, name, method) {
+  if (typeof method !== 'function') {
     let message = 'No target object provided'
 
     if (target) {
       if (typeof target !== 'object' && typeof target !== 'function') {
         message = 'Invalid target'
       } else {
-        message = target[name] ? `Original method ${name} is not a function` : `No original method ${name}`
+        name = String(name)
```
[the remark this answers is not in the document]
BridgeAR: I just addressed symbols being accessed. This would now log the error properly instead. I just thought I'll address everything highlighted by TS.

packages/datadog-shimmer/src/shimmer.js
BridgeAR: That is indeed correct. This is a case that is specifically designed for functions though, so using these properties on a function would shadow the original one.
I guess we can go for the super safe implementation. It's a rare case anyway.

watson: 👍 Normally in cases like this, where there's no reason to guard against a specific edge case because you're sure it's never going to happen, I like to put a code comment in there to future me/others that this is intentional.

BridgeAR: I implemented a fix instead of a comment. I would just skip writing the test case for it.

watson (approved):
Just a small question (see comments), but if there's a good reason for it, feel free to merge

packages/datadog-shimmer/src/shimmer.js
```diff
@@ -23,7 +35,7 @@ function copyProperties (original, wrapped) {
   if (ownKeys.length !== 2) {
     for (const key of ownKeys) {
       if (skipMethods.has(key)) continue
-      const descriptor = Object.getOwnPropertyDescriptor(original, key)
+      const descriptor = Object.getOwnPropertyDescriptor(original, key) ?? {}
```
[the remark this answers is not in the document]
BridgeAR: Sadly yes. I could not immediately see a good way of addressing it otherwise. Thinking about it again, it's probably not ideal to add this code, since it is never executed and other devs could anticipate that it could be reached. I'll remove that again.

watson (approved):

PR #5644, opened 2025-05-01
prevent exclusion of commit based on the release line

rochdev:
### What does this PR do?

Prevent exclusion of commit based on the release line.

### Motivation

This is a continuation of the work to prevent merge conflicts and release line branches deviating. If this ever ends up being needed and a blocker, we can re-evaluate for the specific commit, but conditions that can land everywhere cleanly should be prioritized to enable automation.

### DIFF
--- a/scripts/release/proposal.js
+++ b/scripts/release/proposal.js
@@ -68,14 +68,18 @@ try {
 
   pass(`v${releaseLine}.x`)
 
-  const diffCmd = [
-    'branch-diff',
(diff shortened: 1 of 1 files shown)


bengl (approved):
This is fine, since `semver-major` is still filtered out. Any time we need to block a commit from landing somehwere, it _should_ be because it's semver major. The only other thing I can think of is CODEOWNERS, but we should deal with that separately.

PR #5646, opened 2025-05-01
fix: Support for inode when /proc/self/cgroup is a multiline string

AndreLouisCaron:
### What does this PR do?
Fix inode detection when `/proc/self/cgroup` contains a multi-line string.

### Motivation
We're porting some Kubernetes hosts from Amazon Linux 2023 to Bottlerocket OS and the contents of `/proc/self/cgroup` have changed from `0::/` to `1:name=systemd:/\n0::/`, which caused the inode detection to break.  As a result, the `namespaceLabelsAsTags` directed in our Datadog agent no longer works.

### Additional Notes
We also run Java applications with [Datadog/dd-trace-java](https://github.com/DataDog/dd-trace-java) and did not encounter this problem for our Java apps.


BridgeAR:
@AndreLouisCaron thank you for the PR! I am soon going to port this to a own PR (we can't run our CI on external PRs).

PR #5649, opened 2025-05-02
manual keep trace on asm standalone and api security

IlyasShabi:
### Plugin Checklist

- [ ] Unit tests.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].

### DIFF
--- a/integration-tests/standalone-asm.spec.js
+++ b/integration-tests/standalone-asm.spec.js
@@ -10,7 +10,7 @@ const {
   curlAndAssertMessage,
   curl
 } = require('./helpers')
-const { USER_KEEP, AUTO_REJECT, AUTO_KEEP } = require('../ext/priority')
+const { USER_KEEP, AUTO_REJECT } = require('../ext/priority')
 
 describe('Standalone ASM', () => {
   let sandbox, cwd, startupTestFile, agent, proc, env
@@ -31,7 +31,8 @@ describe('Standalone ASM', () => {
 
       env = {
         AGENT_PORT: agent.port,
-        DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED: 'true'
+        DD_APM_TRACING_ENABLED: 'false',
+        DD_APPSEC_ENABLED: 'true'
       }
 
       const execArgv = []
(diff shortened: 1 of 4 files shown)


packages/dd-trace/src/appsec/api_security_sampler.js
```diff
@@ -41,6 +45,11 @@ function sampleRequest (req, res, force = false) {
   const rootSpan = web.root(req)
   if (!rootSpan) return false
 
+  if (asmStandaloneEnabled) {
```
uurien, line 48: Is priority `AUTO_REJECT` when `asmStandaloneEnabled`? If it is, don't we want to keep in mind the `priority === USER_REJECT`?

packages/dd-trace/test/appsec/api_security_sampler.spec.js
```diff
@@ -197,4 +197,41 @@ describe('API Security Sampler', () => {
       assert.isTrue(apiSecuritySampler.sampleRequest(req, res, true))
     })
   })
+
+  describe('ASM Standalone', () => {
```
uurien, line 201: Can we add few tests to check that everything is working as expected in standalone mode? At least something like:

```javascript
      assert.isTrue(apiSecuritySampler.sampleRequest(req, res, true))
      assert.isFalse(apiSecuritySampler.sampleRequest(req, res, true))
```

integration-tests/standalone-asm.spec.js
```diff
@@ -98,11 +97,7 @@ describe('Standalone ASM', () => {
           assert.strictEqual(fifthReq.length, 5)
 
           const { meta, metrics } = fifthReq[0]
-          assert.notProperty(meta, 'manual.keep')
-          assert.notProperty(meta, '_dd.p.ts')
-
-          assert.propertyVal(metrics, '_sampling_priority_v1', AUTO_KEEP)
-          assert.propertyVal(metrics, '_dd.apm.enabled', 0)
+          assertKeep({ meta, metrics })
```
uurien, line 100: This test was checking `AUTO_KEEP` in `_sampling_priority_v1` and now is checking `USER_KEEP`, is the change expected?

IlyasShabi: I updated to test to be working as expected without API sec sampler impact, I also created an integration test wit hAPI sec enabled

packages/dd-trace/test/appsec/api_security_sampler.spec.js
IlyasShabi: Done

packages/dd-trace/src/appsec/api_security_sampler.js
IlyasShabi: I think that we should not consider the current prio when `asmStandaloneEnabled` since the goal is to make change it to manual keep

packages/dd-trace/src/standalone/tracesource_priority_sampler.js
```diff
@@ -33,7 +33,7 @@ class TraceSourcePrioritySampler extends PrioritySampler {
       return USER_KEEP
     }
 
-    return this._isSampledByRateLimit(context) ? AUTO_KEEP : AUTO_REJECT
+    return this._isSampledByRateLimit(context) ? AUTO_KEEP : USER_REJECT
```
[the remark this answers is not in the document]
simon-id: definitely to be confirmed, because it looks very weird

uurien (approved):

PR #5651, opened 2025-05-04
remove release proposal dont-land-on checks

rochdev:
### What does this PR do?

Remove release proposal `dont-land-on` checks.

### Motivation

I added old commits with this label to #5633 which resulted in empty commits because of a prior fix, and it looks like empty commits are just discarded by GitHub so they are still detected as missing on the release branch. I'll have to merge them in a different way to make it work and then re-enable the check.

### DIFF
--- a/scripts/release/proposal.js
+++ b/scripts/release/proposal.js
@@ -73,10 +73,11 @@ try {
   const legacyDiff = capture(`${diffCmd} --require-label=dont-land-on-v${releaseLine}.x v${releaseLine}.x ${main}`)
 
   if (legacyDiff) {
-    fatal(
-      `The "dont-land-on-v${releaseLine}.x" label is no longer supported.`,
-      'Please remove the label from any offending PR to continue.'
-    )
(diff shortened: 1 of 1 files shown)


szegedi (approved):

BridgeAR:
I am fine with landing this, I just wonder if we can't land the commits that this PR mentions in the comment directly to resolve the problem.

rochdev:
> I am fine with landing this, I just wonder if we can't land the commits that this PR mentions in the comment directly to resolve the problem.

They are already on the main branch, just not on the release branch. The problem is that I pushed a new commit on the release branch to fix the CODEOWNERS instead of using the old commits. I think merging with a different strategy could work to make sure the commits at least apply and have the same net result in the end. I may simply revert my previous commit and then reapply those which should work, but that needs to happen on the release branch meaning it has to happen during a release (since releases are triggered automatically from merging).

This is only an issue because we were not enforcing consistency before so release branches could deviate. This will no longer be possible once the branches are consistent and this is enabled.

PR #5652, opened 2025-05-05
Add _dd.integration to spans to calculate internal stats

amarziali:
### What does this PR do?

This PR adds `_dd.integration` tag to all the spans containing the integration name. While the `component` tag could have the same information, there is no guarantee about the content of the component tag since it can be modified by the user. 
This one is internal and the user cannot change it.

We are changing all the tracers to infer this information in order to calculate stats upstream around integration usage

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/opentracing/span.js
+++ b/packages/dd-trace/src/opentracing/span.js
@@ -242,6 +242,7 @@ class DatadogSpan {
     }
 
     getIntegrationCounter('spans_finished', this._integrationName).inc()
+    this.setTag('_dd.integration', this._integrationName)
 
     if (DD_TRACE_EXPERIMENTAL_SPAN_COUNTS && finishedRegistry) {
       runtimeMetrics.decrement('runtime.node.spans.unfinished')
--- a/packages/dd-trace/test/opentracing/span.spec.js
+++ b/packages/dd-trace/test/opentracing/span.spec.js
@@ -471,6 +471,17 @@ describe('Span', () => {
       expect(processor.process).to.have.been.calledOnce
     })
 
+    it('should add _dd.integration', () => {
+      processor.process.returns(Promise.resolve())
+
+      span = new Span(tracer, processor, prioritySampler, { operationName: 'operation' })
+      span.setTag('component', 'custom')
+      span.finish()
+
+      expect(tagger.add).to.have.been.calledWith(span.context()._tags, { component: 'custom' })
+      expect(tagger.add).to.have.been.calledWith(span.context()._tags, { '_dd.integration': 'opentracing' })
+    })
+
     describe('tracePropagationBehaviorExtract and Baggage', () => {
       let parent
 


vandonr (approved):
approving the intent, but I never touched the JS tracer so I can't really tell about the code

wconti27 (changes requested):
`integrationName` isn't a very reliable attribute to be using within the Opentracing `startSpan`, for the most part it is not being used across the board for all integrations. If you look at Org2 metrics for `spans_created`, you will see only about 10 integrations are setting it properly, and the vast majority are `opentracing`. If we want to go this route, we will need to update the integrations to properly set this attribute when creating a span, which will be quite a few changes. It would also be helpful to add some assertions in the plugin tests directly that the `_dd.integration` is being set correctly, such as for `express` or `mysql`. But in the current state, we are going to end up mainly with `opentracing` being the value sent.

BridgeAR:
What's the status here? :)
Should we address this differently or table it for now?

packages/datadog-plugin-opensearch/src/index.js
```diff
@@ -6,6 +6,10 @@ class OpenSearchPlugin extends ElasticsearchPlugin {
   static get id () {
     return 'opensearch'
   }
+
+  startSpan (...args) {
+    super.startSpan(...args)
+  }
```
wconti27: This change seems unnecessary.

packages/dd-trace/src/opentracing/span.js
```diff
@@ -240,6 +240,7 @@ class DatadogSpan {
     }
 
     getIntegrationCounter('spans_finished', this._integrationName).inc()
+    this.setTag('_dd.integration', this._integrationName)
```
wconti27: Can we put this in the constructor instead? And just place the attribute directly into the span context `tags` instead of calling `setTag`

wconti27 (commented):
LGTM other than the few comments

packages/dd-trace/src/opentracing/span.js
amarziali: done ty

packages/dd-trace/src/plugins/apollo.js
```diff
@@ -18,6 +18,8 @@ class ApolloBasePlugin extends TracingPlugin {
       meta: {}
     }, false)
 
+    span._integrationName = 'apollo'
```
wconti27: I think this call can be moved into the start span call as well.

packages/dd-trace/src/opentracing/span.js
```diff
@@ -240,6 +240,7 @@ class DatadogSpan {
     }
 
     getIntegrationCounter('spans_finished', this._integrationName).inc()
+    this._spanContext._tags['_dd.integration'] = this._integrationName
```
wconti27, line 243: Can we move this call up to the constructor?

wconti27 (commented):
All looks good to me except for the nits + failing tests currently.

packages/dd-trace/src/opentracing/span.js
amarziali: This should be done at the very end when the span finishes.

wconti27: This is fine. The reason I wanted it set in the constructor was to ensure instrumentation telemetry metrics have the correct name, but I see we are already using the integration name in the metrics.

PR #5655, opened 2025-05-05
Report waf results

IlyasShabi:
### What does this PR do?
Returns all WAF results instead of only actions and removes `ruleTriggered` from the request store.

### Motivation

We may call the WAF multiple times during a single request, when a rule is triggered on the first call, we set: `store[DD_TELEMETRY_REQUEST_METRICS].ruleTriggered = true` this value is reset to null after calling `trackRaspRuleMatch`

However, if `ruleTriggered = true` on the first WAF call, but returns false on a second one, we might execute `trackRaspRuleMatch` for the second WAF call before the first one. This race condition will incorrectly increment the `rasp.rule.match` metric.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


BridgeAR (approved):
Code is LGTM. I guess it would still be good to get a second review though from appsec.

I do wonder if the result is not always truthy or if we could rewrite the code that it is in case it currently is not (I would guess it should not reach those spots in that case).

CarlesDD:
LGTM but I miss a test to check the case that motivates this PR (incorrect increment the `rasp.rule.match` metric when there are multiple WAF calls in the same request)

CarlesDD:
LGTM but I miss a test to check the case that motivates this PR - incorrect increment the `rasp.rule.match` metric when multiple WAF calls in the same request.

CarlesDD (approved):

PR #5659, opened 2025-05-05
fix rasp ssrf flaky test

IlyasShabi:
### What does this PR do?
fix rasp ssrf flaky test

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/test/appsec/rasp/ssrf.express.plugin.spec.js
+++ b/packages/dd-trace/test/appsec/rasp/ssrf.express.plugin.spec.js
@@ -121,11 +121,11 @@ describe('RASP - ssrf', () => {
 
           it('Should not detect threat', async () => {
             app = (req, res) => {
-              axiosToTest.get(`https://${req.query.host}`)
+              axiosToTest.get(`https://${req.query.host}`).catch(noop) // swallow network error
               res.end('end')
             }
 
-            axios.get('/?host=www.datadoghq.com')
+            await axios.get('/?host=www.datadoghq.com')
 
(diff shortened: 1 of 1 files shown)


packages/dd-trace/test/appsec/rasp/ssrf.express.plugin.spec.js
```diff
@@ -121,11 +121,11 @@ describe('RASP - ssrf', () => {
 
           it('Should not detect threat', async () => {
             app = (req, res) => {
-              axiosToTest.get(`https://${req.query.host}`)
+              axiosToTest.get(`https://${req.query.host}`).catch(noop) // swallow network error
```
IlyasShabi, line 124: we could have a network error calling `www.datadoghq.com` which is not interesting in this test

CarlesDD: I've tried forcing an `ENOTFOUND` without the `.catch(noop)` and the test keeps passing. I'm not sure if this is the root cause of the flakiness.

IlyasShabi: I don't think either that the flaky test is related to this, this is just an additional helper.
The solution is to add an `await` to wait for request completion before `checkRaspExecutedAndNotThreat`

BridgeAR (approved):

PR #5667, opened 2025-05-06
Add a fake agent for all profiler integration tests

szegedi:
### What does this PR do?
Ensures FakeAgent instance exists and is used by all profiler integration tests that fork.

### Motivation
Few tests run by `yarn test:integration:profiler` used to work fine without a FakeAgent, but remote config will now prevent forked test programs from terminating before timeout as it's doing futile attempts to connect to an agent. This fortunately seems to not be a problem in CI for whatever reason, but it definitely became a problem on this developer's laptop.

Jira: [PROF-11765]

[PROF-11765]: https://datadoghq.atlassian.net/browse/PROF-11765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integration-tests/profiler/profiler.spec.js
+++ b/integration-tests/profiler/profiler.spec.js
@@ -150,8 +150,8 @@ class NetworkEventProcessor extends TimelineEventProcessor {
   }
 }
 
-async function gatherNetworkTimelineEvents (cwd, scriptFilePath, eventType, args) {
-  return gatherTimelineEvents(cwd, scriptFilePath, eventType, args, NetworkEventProcessor)
+async function gatherNetworkTimelineEvents (cwd, scriptFilePath, agentPort, eventType, args) {
+  return gatherTimelineEvents(cwd, scriptFilePath, agentPort, eventType, args, NetworkEventProcessor)
 }
 
 class FilesystemEventProcessor extends TimelineEventProcessor {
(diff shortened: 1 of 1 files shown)


integration-tests/profiler/profiler.spec.js
```diff
@@ -201,18 +201,19 @@ class FilesystemEventProcessor extends TimelineEventProcessor {
...
-async function gatherTimelineEvents (cwd, scriptFilePath, eventType, args, Processor) {
+async function gatherTimelineEvents (cwd, scriptFilePath, agentPort, eventType, args, Processor) {
   const procStart = BigInt(Date.now() * 1000000)
   const proc = fork(path.join(cwd, scriptFilePath), args, {
     cwd,
     env: {
       DD_PROFILING_EXPORTERS: 'file',
       DD_PROFILING_ENABLED: 1,
-      DD_INTERNAL_PROFILING_TIMELINE_SAMPLING_ENABLED: 0 // capture all events
+      DD_INTERNAL_PROFILING_TIMELINE_SAMPLING_ENABLED: 0, // capture all events
```
nsavoire, line 215: You could also disable remote config with `DD_REMOTE_CONFIGURATION_ENABLED: 0`

BridgeAR (approved):

integration-tests/profiler/profiler.spec.js
szegedi: I did, and that alone didn't help. Something else (maybe trace result posting) is still keeping the process alive for about 25-30 seconds.

PR #5672, opened 2025-05-06
Remove code targeting unsupported Node.js versions

BridgeAR:


watson (dismissed):
Primarily reviewed based on changes to the Debugger code which is 👍 Thanks!

I did also look over the other files as well to see if there were any obvious issues. However, I can't determine if the removed support in those files are what we want. I'll leave that up to the responsible teams

watson:
Shouldn't the label be `semver-major` instead of `semver-patch`?

BridgeAR:
@watson I guess yes, while this only removes special handling for Node.js versions up to 16 and that's our v4 range.

watson:
@BridgeAR If it has the label `breaking-change` it should be a major bump no? If it doesn't contain any breaking changes for 5.x it shouldn't have the `breaking-change` label either. But you're removing support for at least one instrumentation and reducing the supported version range of others, so to me it sounds like a breaking change and hence not something we can land in 5.x... have I misunderstood something?

szegedi (dismissed):

IlyasShabi (approved):
LGTM

.github/workflows/project.yml
```diff
@@ -124,12 +124,9 @@ jobs:
...
     strategy:
       matrix:
-        # Important: This is outside the minimum supported version of dd-trace-js
-        # Node > 16 does not work with Cypress@6.7.0 (not even without our plugin)
-        # TODO: figure out what to do with this: we might have to deprecate support for cypress@6.7.0
         version: [eol, latest]
-        # 6.7.0 is the minimum version we support
-        cypress-version: [6.7.0, latest]
+        # 10.2.0 is the minimum version we support
+        cypress-version: [10.2.0, latest]
```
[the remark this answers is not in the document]
Mariovido: I verify that's what we want. According to Juan notes, that's the lowest version with the new plugin system that seems stable (manually tested). And that version defaults to `nodeVersion=system`.

packages/datadog-plugin-fastify/test/code_origin.spec.js
```diff
@@ -4,16 +4,13 @@ const axios = require('axios')
...
 const { getNextLineNumber } = require('../../dd-trace/test/plugins/helpers')
-const { NODE_MAJOR } = require('../../../version')
 
 describe('Plugin', () => {
   let fastify
   let app
 
   describe('fastify', () => {
     withVersions('fastify', 'fastify', (version, _, specificVersion) => {
-      if (NODE_MAJOR <= 18 && semver.satisfies(specificVersion, '>=5')) return
```
juan-fernandez, line 15: aren't we still supporting node18 in v5? I might've misunderstood but I thought we wanted `master` to look as closely as possible as release branches so as to avoid merge conflicts

packages/dd-trace/src/debugger/devtools_client/source-maps.js
```diff
@@ -35,14 +34,7 @@ const self = module.exports = {
...
 }
 
-// The version check inside this function is to guard against a bug Node.js version 18, in which calls to `setTimeout`
-// might throw an uncatchable error from within `AsyncLocalStorage._propagate` with the following error message:
-//
-//     TypeError: Cannot read properties of undefined (reading 'Symbol(kResourceStore)')
-//
-// Source: https://github.com/nodejs/node/blob/v18.20.6/lib/async_hooks.js#L312
 function cacheIt (key, value) {
-  if (NODE_MAJOR < 20) return value
```
juan-fernandez, line 45: same comment: if v5 still supports node18, we either still need this check or the code from master and v5 will diverge, making release harder. Let me know if I'm missing something

packages/datadog-plugin-fastify/test/code_origin.spec.js
BridgeAR: Yes, I guess that should stay until we stop supporting v5

packages/dd-trace/src/debugger/devtools_client/source-maps.js
BridgeAR: This one is fine as is: the cache was deactivated because of a Node.js issue that was actually independent. It is now fixed in Node.js directly and deactivating did not fix the issue in the end. Thus, activating this is actually good.

BridgeAR:
I changed the label back to a non-breaking change due to only removing support for release lines that we do not support anymore and due to Cypress not being an issue anymore (@juan-fernandez confirmed this).

I believe this is good to land.

rochdev (approved):

PR #5673, opened 2025-05-06
remove async resource usage from net integration

rochdev:
### What does this PR do?

Remove `AsyncResource` usage from `net` integration.

### Motivation

Using `AsyncResource` for this is incorrect as it will change the underlying resource for all stores and not just ours. There is also a bug in Node 24 when using `AsyncResource` that isn't triggered when using `runStores` instead.

### DIFF
--- a/packages/datadog-instrumentations/src/net.js
+++ b/packages/datadog-instrumentations/src/net.js
@@ -1,10 +1,6 @@
 'use strict'
 
-const {
-  channel,
-  addHook,
-  AsyncResource
-} = require('./helpers/instrument')
+const { channel, addHook } = require('./helpers/instrument')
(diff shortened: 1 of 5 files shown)


BridgeAR (commented):
This looks really great!

Just two profiler tests are still failing and it still needs some additional tests to cover the current changed code :)

bengl:
I'm going to pull this into `feature/node-24` and make sure it passed on Node.js 24.

rochdev:
Just two profiler tests are still failing and it still needs some additional tests to cover the current changed code :)

This fixes an issue in Node 24. It will become tested once we enable the tests for Node 24. Meanwhile, @bengl has validated above that it fixes the issue.

bengl (approved):

PR #5674, opened 2025-05-06
add release status checks

rochdev:
### What does this PR do?

Add release status checks.

### Motivation

Since we release from a branch and not from a PR, the status checks are not reported in the same way and it's not obvious when the release isn't successful because even through the jobs in GitHub are green, it's not necessarily true for jobs outside of GitHub like any GitLab jobs. This adds a job in GitHub that reports the status of any important external checks to make sure the release workflow is marked as failed if there is any failures.

### DIFF
--- a/scripts/release/status.js
+++ b/scripts/release/status.js
@@ -1,0 +1,62 @@
+'use strict'
+
+/* eslint-disable no-console */
+
+const { GITHUB_REF, GITHUB_TOKEN } = process.env
+
+const TIMEOUT = 10 * 1000
+const MAX_ATTEMPTS = 30
+
+let attempts = 0
+
+async function checkStatuses (contexts) {
+  const url = `https://api.github.com/repos/DataDog/dd-trace-js/commits/${GITHUB_REF}/status`
+  const response = await fetch(url, {
+    headers: {
+      Accept: 'application/vnd.github+json',
+      Authorization: `Bearer ${GITHUB_TOKEN}`,
+      'X-GitHub-Api-Version': '2022-11-28'
+    }
+  })
+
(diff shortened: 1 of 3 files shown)


scripts/release/status.js
```diff
@@ -0,0 +1,62 @@
...
+          case 'failure':
+          case 'stale':
+          case 'timed_out':
+            throw new Error(`Job was not successful: ${context}.`)
+        }
+      }
+    }
+  }
+
+  if (contexts.length === 0) return
```
BridgeAR, line 48: An optimistic approach about each status containing a context and the state as such having to succeed without looking it up.

```suggestion
  const contextSet = new Set(contexts)

  for (const status of statuses) {
    const context = contextSet.get(status.context)
    if (status.state !== 'success') {
      throw new Error(`Job was not successful: ${context}.`)
    }
    contextSet.delete(context)
  }
  
  if (contextSet.size === 0) return
```

scripts/release/status.js
```diff
@@ -0,0 +1,62 @@
...
+      Authorization: `Bearer ${GITHUB_TOKEN}`,
+      'X-GitHub-Api-Version': '2022-11-28'
+    }
+  })
+
+  if (response.status !== 200) {
+    console.log(response)
+    console.log(response.text())
+
+    throw new Error('Could not get status from GitHub.')
```
BridgeAR, line 26: I guess having it all in the error message might be nice?

```suggestion
    throw new Error(
      `Could not get status from GitHub.\n\n${util.inspect(response)}\n${util.inspect(response.text?.())}`
    )
```

BridgeAR (approved):
LGTM!

PR #5682, opened 2025-05-08
Use errorMonitor symbol instead of listening to the error event

BridgeAR:
This is important to guarantee user code behaves as it did without
the instrumentation. So user crashes should happen as before.

I did not verify if we handle this else-wise and I mainly wanted to open this to look into this closer to make sure this is indeed what we want or not.

### DIFF
--- a/packages/datadog-instrumentations/src/couchbase.js
+++ b/packages/datadog-instrumentations/src/couchbase.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const { errorMonitor } = require('events')
 const {
   channel,
   addHook,
@@ -186,7 +187,7 @@ addHook({ name: 'couchbase', file: 'lib/bucket.js', versions: ['^2.6.12'] }, Buc
         finishCh.publish(undefined)
       }))
 
-      emitter.once('error', asyncResource.bind((error) => {
+      emitter.once(errorMonitor, asyncResource.bind((error) => {
         errorCh.publish(error)
         finishCh.publish(undefined)
       }))
--- a/packages/datadog-instrumentations/src/pg.js
+++ b/packages/datadog-instrumentations/src/pg.js
@@ -132,6 +132,9 @@ function wrapQuery (query) {
           .once(errorMonitor, finish)
           .once('end', (res) => finish(null, res))
       } else {
+        // TODO: This code is never reached in our tests.
+        // Internally, pg always uses callbacks or streams, even for promise based queries.
+        // Investigate if this code should just be removed.
         newQuery.then((res) => finish(null, res), finish)
       }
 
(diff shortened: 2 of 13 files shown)


packages/datadog-instrumentations/src/pg.js
```diff
@@ -129,21 +129,12 @@ function wrapQuery (query) {
...
 
-      if (stream) {
-        newQuery.on('end', () => {
-          finish(null, [])
-        })
-        newQuery.on(errorMonitor, (err) => {
-          finish(err)
-        })
-      }
-
```
BridgeAR, line 146: This was actually adding a second listener. Now, it's only a single one and only the monitor one.

packages/datadog-instrumentations/test/mysql2.spec.js
```diff
@@ -435,16 +436,18 @@ describe('mysql2 instrumentation', () => {
...
+              let error
+              query.on('error', err => { error = err })
+              expect(query.listenerCount('error')).to.equal(1)
 
-                done()
-              })
+              await once(query, 'end')
+              expect(query.listenerCount('error')).to.equal(1)
+              sinon.assert.called(apmQueryStart)
+              expect(error).to.be.undefined
```
watson, line 450: This check, and the `error` listener you add above shouldn't be necessary, as the `await once(query, 'end')` will be rejected if `query` emits and error, and hence the test will fail and never reach this point.

packages/datadog-instrumentations/src/pg.js
BridgeAR: The time this was added, tests were added that would fail if the above listener is removed. All callback cases are event emitters but it's fine, since we wait for that to finish. It was only important if there was no callback.

I did however do a closer check and added a test for failing cases as well.
Interestingly, the .then part is never going to be reached in any of our cases, even the promise related tests. The pg internals just all work with callbacks. I wonder if we want to remove that condition.

packages/datadog-instrumentations/test/mysql2.spec.js
BridgeAR: True, while it likely depends on the Node.js version to prevent an error from being emitted after 'end'. I'll simplify it.

packages/datadog-instrumentations/src/couchbase.js
```diff
@@ -1,5 +1,6 @@
 'use strict'
 
+const { errorMonitor } = require('events')
```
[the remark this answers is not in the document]
BridgeAR: Yeah... I am a bit indecisive to use them.
I'll keep it for now due to otherwise running an extra CI and I don't think it should matter overall.

packages/datadog-instrumentations/src/pg.js
watson: I think for now leave it as is, as it's a bit outside the scope of this PR. Maybe just add a TODO comment?

watson (dismissed):
Looks good. Just my one nit about the TODO comment. Feel free to merge without it or add it and I'll re-approve

watson (approved):

PR #5683, opened 2025-05-08
fix: Support for inode when /proc/self/cgroup is a multiline string

BridgeAR:
There could be more than one subsystem and we only use one. I guess that's fine? Or should we look into this closer?

I did a small refactoring of the code since I had to stare at it for a bit too see what part did what. I feel the less lines are nicer to read while I am also fine to keep the old code.

Closes #5646

### DIFF
--- a/packages/dd-trace/src/exporters/common/docker.js
+++ b/packages/dd-trace/src/exporters/common/docker.js
@@ -13,40 +13,22 @@ const taskSource = '[0-9a-f]{32}-\\d+'
 const lineReg = /^(\d+):([^:]*):(.+)$/m
 const entityReg = new RegExp(`.*(${uuidSource}|${containerSource}|${taskSource})(?:\\.scope)?$`, 'm')
 
-const cgroup = readControlGroup()
-const entityId = getEntityId()
-const inode = getInode()
+let inode = 0
+let cgroup = ''
+let entityId
 
-function getEntityId () {
-  const match = cgroup.match(entityReg) || []
+try {
+  cgroup = fs.readFileSync('/proc/self/cgroup', 'utf8').trim()
+  entityId = cgroup.match(entityReg)?.[1]
+} catch { /* Ignore error */ }
 
-  return match[1]
-}
-
-function getInode () {
-  const match = cgroup.match(lineReg) || []
-
-  return readInode(match[3])
-}
-
-function readControlGroup () {
-  try {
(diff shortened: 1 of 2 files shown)


packages/dd-trace/test/exporters/common/docker.spec.js
```diff
@@ -35,7 +35,7 @@ describe('docker', () => {
       `1:name=systemd:/docker/${id}`
     ].join('\n')
 
-    fs.readFileSync.withArgs('/proc/self/cgroup').returns(Buffer.from(cgroup))
+    fs.readFileSync.withArgs('/proc/self/cgroup').returns(cgroup)
```
rochdev, line 38: `readFileSync` only returns a string with the encoding specified, so let's add the encoding to the test too.

packages/dd-trace/src/exporters/common/docker.js
```diff
@@ -10,43 +10,25 @@ const uuidSource =
...
 const containerSource = '[0-9a-f]{64}'
 const taskSource = '[0-9a-f]{32}-\\d+'
-const lineReg = /^(\d+):([^:]*):(.+)$/
+const lineReg = /^(\d+):([^:]*):(.+)$/m
 const entityReg = new RegExp(`.*(${uuidSource}|${containerSource}|${taskSource})(?:\\.scope)?$`, 'm')
 
-const cgroup = readControlGroup()
-const entityId = getEntityId()
-const inode = getInode()
+let inode = 0
```
rochdev, line 16: Why this unrelated change? In the PR description you mentioned that you think this is more readable, but I'm not sure I understand why. I definitely prefer being able to be able to quickly understand where a value comes from by isolating its provenance to a function call and knowing it won't be changed afterwards thanks to `const` instead of having to read through the whole file to find any code that could be changing the value.

rochdev (changes requested):

packages/dd-trace/test/exporters/common/docker.spec.js
BridgeAR: Good catch! Thank you for pointing that out!

packages/dd-trace/src/exporters/common/docker.js
BridgeAR: The former code was more than twice as long and abstracting it in methods meant I had to check if these were called more than once and from where they were called. Now it's just sequential simple code :)

rochdev (approved):

PR #5691, opened 2025-05-09
Skip app-dir app in nextjs 3.2 test in node 24

uurien:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/test/appsec/index.next.plugin.spec.js
+++ b/packages/dd-trace/test/appsec/index.next.plugin.spec.js
@@ -9,6 +9,8 @@ const { satisfies } = require('semver')
 const path = require('path')
 
 const agent = require('../plugins/agent')
+const { NODE_MAJOR } = require('../../../../version')
+
 
 describe('test suite', () => {
   let server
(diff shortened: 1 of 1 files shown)


packages/dd-trace/test/appsec/index.next.plugin.spec.js
```diff
@@ -146,7 +152,7 @@ describe('test suite', () => {
       }
     ]
 
-    if (satisfies(realVersion, '>=13.2')) {
+    if (satisfies(realVersion, '>=13.2') && !(NODE_MAJOR >= 24 && satisfies(realVersion, '=13.2'))) {
```
BridgeAR, line 155: I personally find negations always more tricky to read than the other way around. So I would have written it as:

```suggestion
    if (satisfies(realVersion, '>=13.2') && (NODE_MAJOR < 24 || satisfies(realVersion, '!=13.2'))) {
```

BridgeAR (approved):
LGTM, just left a suggestion. While landing, we should update the commit message to match Next.js 13.2 instead of 3.2 :)

packages/dd-trace/test/appsec/index.next.plugin.spec.js
uurien: I have no strong opinion about one or other... suggestion addressed.

uurien:
> While landing, we should update the commit message to match Next.js 13.2 instead of 3.2 :)

Yes, you're right. Anyway, this commit is pointing to `feature/node-24` and the PRs are squashed & merged so this commit message will disappear 😅

PR #5700, opened 2025-05-12
[DI] Clear state when last probe is removed

watson:
### What does this PR do?

When the last probe is removed (or disabled), the debugger session is stopped. However, some of the internal state keeping track of the loaded files in the process isn't cleared. If a new probe is then later added, or an existing probe enabled, a new debugger session is started, and the internal state is re-generated. But instead of starting from an empty state, it's essentially duplicated. The more times this happens, the more memory is used, causing a memory leak.
    
To fix this, the internal state is now cleared when the debugger session is stopped.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/test/debugger/devtools_client/breakpoints.spec.js
+++ b/packages/dd-trace/test/debugger/devtools_client/breakpoints.spec.js
@@ -30,6 +30,7 @@ describe('breakpoints', function () {
         sourceMapURL: null,
         source: null
       }),
+      clearState: sinon.stub(),
       locationToBreakpoint: new Map(),
       breakpointToProbes: new Map(),
       probeToLocation: new Map(),
@@ -88,8 +89,7 @@ describe('breakpoints', function () {
         }
       })
 
-      expect(sessionMock.post.callCount).to.equal(1)
-      expect(sessionMock.post.firstCall).to.have.been.calledWith('Debugger.setBreakpoint')
+      expect(sessionMock.post).to.have.been.calledOnceWith('Debugger.setBreakpoint')
     })
 
     describe('add multiple probes to the same location', function () {
@@ -116,7 +116,7 @@ describe('breakpoints', function () {
           }
         })
 
-        expect(sessionMock.post.callCount).to.equal(0)
+        expect(sessionMock.post).to.not.have.been.called
       })
 
       it('mixed: 2nd probe no condition', async function () {
@@ -190,7 +190,7 @@ describe('breakpoints', function () {
           }
         })
 
-        expect(sessionMock.post.callCount).to.equal(0)
+        expect(sessionMock.post).to.not.have.been.called
       })
 
       it('all conditions', async function () {
@@ -316,8 +316,8 @@ describe('breakpoints', function () {
 
       await breakpoints.removeBreakpoint({ id: 'probe-1' })
 
-      expect(sessionMock.post.callCount).to.equal(1)
-      expect(sessionMock.post).to.have.been.calledWith('Debugger.disable')
+      expect(sessionMock.post).to.have.been.calledOnceWith('Debugger.disable')
+      expect(stateMock.clearState).to.have.been.calledOnce
     })
 
     it('should not disable debugger when there are other breakpoints', async function () {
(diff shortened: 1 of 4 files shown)


packages/dd-trace/test/debugger/devtools_client/state.spec.js
```diff
@@ -189,6 +189,49 @@ describe('findScriptFromPartialPath', function () {
     it('a Windows drive letter with a backslash', testPathNoMatch('c:\\'))
   })
 
+  describe('state', function () {
+    it('should be cleared when calling clearState', function () {
```
watson, line 193: I don't like this test. It's testing an internal implementation detail which isn't good practise. But I couldn't find a better way to unit test this, besides exposing the external state just for testing purposes. But I don't like doing that either. I think the conclusion is that this could all do with some refactoring. There's currently a lot of strong dependences between `state.js` and `breakpoint.js`, which in itself is a bit of a code smell.

packages/dd-trace/test/debugger/devtools_client/breakpoints.spec.js
```diff
@@ -376,7 +379,7 @@ describe('breakpoints', function () {
 
         await breakpoints.removeBreakpoint({ id: 'probe-1' })
 
-        expect(sessionMock.post.callCount).to.equal(0)
+        expect(sessionMock.post).to.not.have.been.called
```
BridgeAR, line 382: I am not sure if a proxy is used here, but if not, having a typo would make the test pass. I therefore always prefer a function to be called at the end. That way it's guaranteed to be detected in case there is a typo.

BridgeAR (commented):
Code LGTM, while I would keep the function call at the end. The comment about the internals would of course be nice to be addressed. Would that take too long to improve in the same PR?

packages/dd-trace/test/debugger/devtools_client/breakpoints.spec.js
watson: If you misspell it, it does error out:
image

watson: The reason why I made these changes in this PR is that it provides better error messages in case of an assertion error and is less verbose

packages/dd-trace/test/debugger/devtools_client/state.spec.js
watson: > 1. The state clearing should be called during stop and that clearing would have an effect on future calls. So we could verify that starting the debugger for a specific script, stopping it, and now starting the debugger on another one that would theoretically be impacted by the former state to know that the state is indeed cleared.

Unfortunately, the only side-effect of not clearing the state is a memory leak. It's not going to affect the functionality of the product. So from the outside looking in, you can only see that the memory keeps going up the more times `stop` and `start` is called.

> 2. I would probably handle the state from inside of breakpoint with dependency injection.

Do you mean that `breakpoint.js` initialises all the state in `state.js` during the call to `start`? That essentially has the same effect as exposing the internal state for testing purposes. But maybe that's the lesser of two evils?

watson: Update after Slack discussion: I've refactored the test so it no longer tests the internals of the `findScriptFromPartialPath` function, but instead exposes the state for testing purposes. This is still not ideal, but better than before. In reality this is not really a good candidate for a unit test.

There were a discussion to call an `init` function from the `start` function to initialize the state. I like this, but it doesn't really have anything to do with clearing the state in when stopping (which I'd like to do instead of waiting for `start` to be called, to mark the memory as clearable.

BridgeAR (commented):
It's fine for me to expose it as such, since I don't have a great idea around it otherwise either.

I would still implement it so that calling `start()` would just define its own state. That way, there should also not be a leak, due to the variables not being referenced by anything anymore after usage.

BridgeAR (approved):

PR #5706, opened 2025-05-12
use no async context frame option for now

bengl:
### What does this PR do?
Adds `--no-async-context-frame` on tests whenever child processes are used. This mainly applies to integration tests.

The debugger test is also affected, but since it's been updated to work with `AsyncContextFrame`, there's an override option for that one.

This makes all integration tests pass on Node.js 24, since the flag reverts to Node.js <24 behaviour.

### Additional Notes

We'll back all of this out once we can fully support Node.js 24 and not use the flag.

### DIFF
--- a/packages/dd-trace/test/setup/core.js
+++ b/packages/dd-trace/test/setup/core.js
@@ -4,6 +4,7 @@ const sinon = require('sinon')
 const chai = require('chai')
 const sinonChai = require('sinon-chai')
 const proxyquire = require('../proxyquire')
+const { NODE_MAJOR } = require('../../../../version')
 
 {
   // get-port can often return a port that is already in use, thanks to a race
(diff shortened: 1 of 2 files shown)


rochdev (approved):

packages/dd-trace/test/setup/core.js
```diff
@@ -48,3 +49,35 @@ if (/^v\d+\.x$/.test(process.env.GITHUB_BASE_REF || '')) {
...
+        opts.env.NODE_OPTIONS = '--no-async-context-frame'
+      }
+    }
+    return exec.apply(this, args)
+  }
+
+  childProcess.fork = function (...args) {
+    const opts = args[1]
+    if (opts) {
+      if (opts?.env?.NODE_OPTIONS && !opts.env.NODE_OPTIONS.includes('--no-async-context-frame')) {
```
BridgeAR, line 74: I believe the second condition should actually be an inner if and it would just skip the addition. Otherwise the presence of the flag would remove all other NODE_OPTIONS potentially also being set.

bengl: I think you're right. While this is only in test code where we control NODE_OPTIONS, we don't necessarily know _where_ we might want to add other things, so this is worth fixing. I'll do that in a follow-up PR in a few min. 👍

PR #5707, opened 2025-05-12
convert aws-sdk to runstores instead of asyncresource

rochdev:
### What does this PR do?

Convert `aws-sdk` to `runStores` instead of `AsyncResource`.

### Motivation

Otherwise it's broken on Node 24 as `AsyncResource` is broken.

### DIFF
--- a/packages/datadog-plugin-aws-sdk/src/base.js
+++ b/packages/datadog-plugin-aws-sdk/src/base.js
@@ -33,17 +33,21 @@ class BaseAwsSdkPlugin extends ClientPlugin {
   constructor (...args) {
     super(...args)
 
-    this.addSub(`apm:aws:request:start:${this.serviceIdentifier}`, ({
-      request,
-      operation,
-      awsRegion,
-      awsService
-    }) => {
-      if (!this.isEnabled(request)) {
-        return
-      }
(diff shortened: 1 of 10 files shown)


packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -114,24 +127,31 @@ class BaseAwsSdkPlugin extends ClientPlugin {
...
-        const params = response.request.params
-        const operation = response.request.operation
-        this.responseExtractDSMContext(operation, params, response.data ?? response, span)
-      }
-      this.addResponseTags(span, response)
 
-      if (this._tracerConfig?.trace?.aws?.addSpanPointers) {
-        this.addSpanPointers(span, response)
-      }
+      storage('legacy').run(currentStore, () => {
```
wconti27, line 136: Why is this `store.run` needed? From my understanding, there's nothing context specific here, and all the variables being used are in scope and not coming from the store.

rochdev: DSM methods like `responseExtractDSMContext` access `storage('legacy').getStore` internally.

wconti27: Makes sense 👍

PR #5709, opened 2025-05-13
Extended Request Data Collection

CarlesDD:
### What does this PR do?
Provides a feature to collect additional data:
- Request and response headers on AppSec events, extending the already implemented header collection system.
- Request body on RASP events, reported using `meta_struct`

These collections are ruled by new configurations (see `index.d.ts`)

### Motivation
Meeting the need of more information for the in-depth investigation of high-risk events.

### Additional Notes
[APPSEC-57290](https://datadoghq.atlassian.net/browse/APPSEC-57290)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -696,7 +696,12 @@ declare namespace tracer {
         /** Whether to enable RASP.
          * @default false
          */
-        enabled?: boolean
+        enabled?: boolean,
+
+        /** Whether to enable request body collection on RASP event
+         * @default false
+         */
+        bodyCollection?: boolean
       },
       /**
        * Configuration for stack trace reporting
--- a/index.d.ts
+++ b/index.d.ts
@@ -716,6 +716,25 @@ declare namespace tracer {
          * @default 32
          */
         maxDepth?: number,
+      },
+      /**
+       * Configuration for extended headers collection tied to security events
+       */
+      extendedHeadersCollection?: {
+        /** Whether to enable extended headers collection
+         * @default false
+         */
+        enabled: boolean,
+
+        /** Whether to redact collected headers
+         * @default true
+         */
+        redaction: boolean,
+
+        /** Specifies the maximum number of headers collected.
+         * @default 50
+         */
+        maxHeaders: number,
       }
     }
 
--- a/packages/dd-trace/src/appsec/reporter.js
+++ b/packages/dd-trace/src/appsec/reporter.js
@@ -17,6 +17,7 @@ const {
 const zlib = require('zlib')
 const { keepTrace } = require('../priority_sampler')
 const { ASM } = require('../standalone/product')
+const { HTTP_INCOMING_BODY } = require('./addresses')
 
 const REQUEST_HEADER_TAG_PREFIX = 'http.request.headers.'
 const RESPONSE_HEADER_TAG_PREFIX = 'http.response.headers.'
@@ -253,10 +254,11 @@ function reportAttack (attackData) {
   const currentJson = currentTags['_dd.appsec.json']
 
   // merge JSON arrays without parsing them
+  const attackDataStr = JSON.stringify(attackData)
   if (currentJson) {
-    newTags['_dd.appsec.json'] = currentJson.slice(0, -2) + ',' + attackData.slice(1) + '}'
+    newTags['_dd.appsec.json'] = currentJson.slice(0, -2) + ',' + attackDataStr.slice(1) + '}'
   } else {
-    newTags['_dd.appsec.json'] = '{"triggers":' + attackData + '}'
+    newTags['_dd.appsec.json'] = '{"triggers":' + attackDataStr + '}'
   }
 
   if (req.socket) {
@@ -264,6 +266,27 @@ function reportAttack (attackData) {
   }
 
   rootSpan.addTags(newTags)
+
+  if (isRaspAttack(attackData)) {
+    reportRequestBody(rootSpan, req.body)
+  }
+}
+
+function reportRequestBody (rootSpan, requestBody) {
+  if (!requestBody) return
+
+  if (!rootSpan.meta_struct) {
+    rootSpan.meta_struct = {}
+  }
+
+  if (!rootSpan.meta_struct['http.request.body']) {
+    // TODO truncate requestBody
+    rootSpan.meta_struct['http.request.body'] = requestBody
+  }
+}
+
+function isRaspAttack(events) {
+  return events.some(e => e.rule?.tags?.module === 'rasp')
 }
 
 function isFingerprintDerivative (derivative) {
--- a/packages/dd-trace/src/appsec/reporter.js
+++ b/packages/dd-trace/src/appsec/reporter.js
@@ -18,8 +18,13 @@ const zlib = require('zlib')
 const { keepTrace } = require('../priority_sampler')
 const { ASM } = require('../standalone/product')
 
+const REQUEST_HEADER_TAG_PREFIX = 'http.request.headers.'
+const RESPONSE_HEADER_TAG_PREFIX = 'http.response.headers.'
+
 // default limiter, configurable with setRateLimit()
 let limiter = new Limiter(100)
+
+let extendedCollection = null
 
 const metricsQueue = new Map()
 
@@ -40,7 +45,7 @@ const EVENT_HEADERS_MAP = mapHeaderAndTags([
   'host',
   'accept-encoding',
   'accept-language'
-], 'http.request.headers.')
+], REQUEST_HEADER_TAG_PREFIX)
 
 const identificationHeaders = [
   'x-amzn-trace-id',
@@ -59,12 +64,24 @@ const REQUEST_HEADERS_MAP = mapHeaderAndTags([
   'user-agent',
   'accept',
   ...identificationHeaders
-], 'http.request.headers.')
+], REQUEST_HEADER_TAG_PREFIX)
 
-const RESPONSE_HEADERS_MAP = mapHeaderAndTags(contentHeaderList, 'http.response.headers.')
+const RESPONSE_HEADERS_MAP = mapHeaderAndTags(contentHeaderList, RESPONSE_HEADER_TAG_PREFIX)
+
+function formatHeaderName (name) {
+  return name
+    .trim()
+    .slice(0, 200)
+    .replace(/[^a-zA-Z0-9_\-:/]/g, '_')
+    .toLowerCase()
+}
+
+function getHeaderTag (tagPrefix, headerName) {
+  return `${tagPrefix}${formatHeaderName(headerName)}`
+}
 
 function mapHeaderAndTags (headerList, tagPrefix) {
-  return new Map(headerList.map(headerName => [headerName, `${tagPrefix}${formatHeaderName(headerName)}`]))
+  return new Map(headerList.map(headerName => [headerName, getHeaderTag(tagPrefix, headerName)]))
 }
 
 function filterHeaders (headers, map) {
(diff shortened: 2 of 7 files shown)


packages/dd-trace/src/appsec/reporter.js
```diff
@@ -163,17 +261,66 @@ function reportAttack (attackData) {
...
 
   rootSpan.addTags(newTags)
+
+  if (extendedCollection?.raspBodyCollection && isRaspAttack(attackData)) {
+    reportRequestBody(rootSpan, req.body)
+  }
+}
+
+function truncateRequestBody (target, depth = 0) {
+  switch (typeof target) {
```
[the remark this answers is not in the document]
CarlesDD: Yes. This point has been discussed previously in order to have the truncation logic once, as other tracers do. But, since it will require a non despreciable effort and this feature is P0, it will be done in a separate task.

packages/dd-trace/src/appsec/index.js
```diff
@@ -58,6 +58,10 @@ function enable (_config) {
     remoteConfig.enableWafUpdate(_config.appsec)
 
     Reporter.setRateLimit(_config.appsec.rateLimit)
+    Reporter.setExtendedCollection({
```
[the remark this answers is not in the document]
CarlesDD: Much better, yes

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -163,17 +261,66 @@ function reportAttack (attackData) {
...
+function reportRequestBody (rootSpan, requestBody) {
+  if (!requestBody) return
+
+  if (!rootSpan.meta_struct) {
+    rootSpan.meta_struct = {}
+  }
+
+  if (!rootSpan.meta_struct['http.request.body']) {
+    // TODO truncate requestBody
+    rootSpan.meta_struct['http.request.body'] = truncateRequestBody(requestBody)
```
IlyasShabi, line 318: We also should report `_dd.appsec.rasp.request_body_size.exceeded` if at least one limit is exceeded no?

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -82,12 +103,89 @@ function filterHeaders (headers, map) {
...
+  // Extended collection
+  if (!(extendedCollection?.enabled && !extendedCollection?.redaction)) {
+    return Object.assign(
+      mandatoryCollectedHeaders,
+      requestEventCollectedHeaders,
+      responseEventCollectedHeaders
+    )
+  }
+
+  const requestExtendedHeadersAvailableCount =
```
uurien, line 142: The comment is making me confuse. The extended collection starts on line 142 and not here, the code into the `if` is the "standard" collection. 
```suggestion
  if (!(extendedCollection?.enabled && !extendedCollection?.redaction)) {
    // standard collection
    return Object.assign(
      mandatoryCollectedHeaders,
      requestEventCollectedHeaders,
      responseEventCollectedHeaders
    )
  }

  // Extended collection
  const requestExtendedHeadersAvailableCount =
```

index.d.ts
```diff
@@ -696,7 +696,12 @@ declare namespace tracer {
         /** Whether to enable RASP.
          * @default false
          */
-        enabled?: boolean
+        enabled?: boolean,
```
uurien, line 699: You should update `docs/test.ts` file with these options.

uurien (changes requested):
I miss integration or plugin style tests testing the whole feature.

packages/dd-trace/src/appsec/reporter.js
CarlesDD: Done!

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -163,17 +261,66 @@ function reportAttack (attackData) {
...
   }
 
   rootSpan.addTags(newTags)
+
+  if (extendedCollection?.raspBodyCollection && isRaspAttack(attackData)) {
+    reportRequestBody(rootSpan, req.body)
+  }
+}
+
+function truncateRequestBody (target, depth = 0) {
```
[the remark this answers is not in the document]
CarlesDD: I've implemented the replacing of toJSON fn for the resulting object.

packages/dd-trace/src/appsec/index.js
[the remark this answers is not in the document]
CarlesDD: Improved

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -252,15 +447,9 @@ function finishRequest (req, res) {
...
-  rootSpan.addTags(mandatoryTags)
-
   const tags = rootSpan.context()._tags
-  if (!shouldCollectEventHeaders(tags)) return
 
-  const newTags = filterHeaders(res.getHeaders(), RESPONSE_HEADERS_MAP)
-  Object.assign(newTags, filterHeaders(req.headers, EVENT_HEADERS_MAP))
+  const newTags = getCollectedHeaders(req, res, shouldCollectEventHeaders(tags))
 
   if (tags['appsec.event'] === 'true' && typeof req.route?.path === 'string') {
```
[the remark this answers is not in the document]
CarlesDD: Yes, I grouped all headers collection logic in this fn, where `shouldCollectEventHeaders` is checked to decide which headers are collected.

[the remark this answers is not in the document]
simon-id: no but I mean this line `  if (tags['appsec.event'] === 'true' && typeof req.route?.path === 'string') {` is now always executed. That's a change in spec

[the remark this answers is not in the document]
CarlesDD: I see what you mean, but I can't find in which case the spec changes.

In the previous code the execution halts after reporting the mandatory headers if `shouldCollectEventHeaders() === false`, so the code to report the event req headers, the response headers and the path is not executed.

With the new code, if `shouldCollectEventHeaders() === false` the premature halt does not occur. In this case the mandatory headers are reported (as in the previous code) and the condition for reporting the path is checked. If `shouldCollectEventHeaders() === false`, it means that `tags['appsec.event'] !== 'true'`, so the path will not be reported under these conditions.

Have you found any case where this result changes?

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -163,17 +278,97 @@ function reportAttack (attackData) {
...
+      return { value: target, truncated: false }
+    case 'object':
+      if (target === null) {
+        return { value: target, truncated: false }
+      }
+
+      if (depth < COLLECTED_REQUEST_BODY_MAX_DEPTH) {
+        if (Array.isArray(target)) {
+          const sliced = target.slice(0, COLLECTED_REQUEST_BODY_MAX_ELEMENTS_PER_NODE)
+          wasTruncated = target.length > COLLECTED_REQUEST_BODY_MAX_ELEMENTS_PER_NODE
```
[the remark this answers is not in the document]
CarlesDD: Slightly more performant with for loop. I'll apply the change.

packages/dd-trace/src/appsec/reporter.js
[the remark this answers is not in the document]
simon-id: Indeed I was reading it wrong sorry about that :+1:

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -82,12 +116,90 @@ function filterHeaders (headers, map) {
...
+  // Extended collection
+  const requestExtendedHeadersAvailableCount =
+    config.maxHeadersCollected -
+    Object.keys(mandatoryCollectedHeaders).length -
+    Object.keys(requestEventCollectedHeaders).length
+
+  const requestEventExtendedCollectedHeaders =
+    filterExtendedHeaders(
+      req.headers,
+      [...REQUEST_HEADERS_MAP.keys()].concat([...EVENT_HEADERS_MAP.keys()]),
```
simon-id, line 164: would be better to precomputer those two lists, keys() + spread operator + concat, is sloooow

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -82,12 +116,90 @@ function filterHeaders (headers, map) {
...
+  const headersTags = Object.assign(
+    mandatoryCollectedHeaders,
+    requestEventCollectedHeaders,
+    requestEventExtendedCollectedHeaders,
+    responseEventCollectedHeaders,
+    responseEventExtendedCollectedHeaders
+  )
+
+  // Check discarded headers
+  const requestHeadersCount = Object.keys(req.headers).length
```
simon-id, line 190: aw man, we really need what Ruben is working on: https://github.com/tc39/proposal-object-property-count

packages/dd-trace/test/config.spec.js
```diff
@@ -247,6 +247,7 @@ describe('Config', () => {
     expect(config).to.have.nested.property('appsec.enabled', undefined)
     expect(config).to.have.nested.property('appsec.rules', undefined)
     expect(config).to.have.nested.property('appsec.rasp.enabled', true)
+    expect(config).to.have.nested.property('appsec.rasp.bodyCollection', false)
```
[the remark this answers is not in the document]
CarlesDD: 3 new config entries in the default test were added. Besides that, the `maxCollectionHeaders` config is parsed to int.

packages/dd-trace/src/appsec/reporter.js
CarlesDD: Also, I've refactored a bit the header list declarations trying to improve the readability.

simon-id: ah yes true, 2. was wrong

packages/dd-trace/test/config.spec.js
```diff
@@ -297,6 +301,7 @@ describe('Config', () => {
         origin: 'default'
       },
       { name: 'appsec.rasp.enabled', value: true, origin: 'default' },
+      { name: 'appsec.rasp.bodyCollection', value: false, origin: 'default' },
```
simon-id, line 304: still missing the 3 other vars here ;)

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -82,12 +116,90 @@ function filterHeaders (headers, map) {
...
+  if (!headers) return result
+
+  let counter = 0
+  for (const [headerName, headerValue] of Object.entries(headers)) {
+    if (counter >= limit) break
+    if (!excludedHeaderNames.includes(headerName)) {
+      result[getHeaderTag(tagPrefix, headerName)] = '' + headerValue
+      counter++
+    }
+  }
```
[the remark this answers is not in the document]
simon-id: ah you're right

simon-id (dismissed):

integration-tests/appsec/data-collection.spec.js
```diff
@@ -0,0 +1,125 @@
...
+
+    afterEach(async () => {
+      proc.kill()
+      await agent.stop()
+    })
+  }
+
+  async function assertHeadersReported (requestHeaders, responseHeaders) {
+    await agent.assertMessageReceived(({ headers, payload }) => {
+      // Request headers
```
IlyasShabi, line 59: NIT: I would like to check `_dd.appsec.rasp.request_body_size.exceeded` tag exist in this integration test but it's not a blocker

IlyasShabi (dismissed):

integration-tests/appsec/data-collection.spec.js
CarlesDD: Added a test case to check the tag and the reported truncated req body

IlyasShabi (approved):
LGTM

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -163,17 +284,99 @@ function reportAttack (attackData) {
...
+function truncateRequestBody (target, depth = 0) {
+  let wasTruncated = false
+
+  switch (typeof target) {
+    case 'string':
+      if (target.length > COLLECTED_REQUEST_BODY_MAX_STRING_LENGTH) {
+        return { value: target.slice(0, COLLECTED_REQUEST_BODY_MAX_STRING_LENGTH), truncated: true }
+      }
+      return { value: target, truncated: false }
+    case 'object': {
```
uurien, line 314: why this code block? `{`

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -163,17 +284,99 @@ function reportAttack (attackData) {
...
+        for (let i = 0; i < maxArrayLength; i++) {
+          const { value, truncated } = truncateRequestBody(target[i], depth + 1)
+          if (truncated) wasTruncated = true
+          truncatedArray[i] = value
+        }
+
+        return { value: truncatedArray, truncated: wasTruncated }
+      }
+
+      if (typeof target.toJSON === 'function') {
```
uurien, line 336: Shouldn't this be before the `Array.isArray`?

PR #5712, opened 2025-05-13
feat: add deterministic sampling with knuth based on TraceId

genesor:
### What does this PR do?

The PR changes the sampling logic from a random one to a deterministic method to align the JS tracer with others.
This new sampling is based on the TraceID as opposed to a random based one. The current one has been backported to `RandomSampler` to use whenever a Trace/Span is not available (or the item to be sampled is not based on one)

Some tags applied to ML Obs plugins are dependent on sampling, the are now applied **after** span creation as the span is now needed to get a sampling decision.

### Motivation

We want to have a consistent sampling logic amongst all tracers in order to control better what is kept or dropped. This will also ensure that logs related a trace are kept and help customer down the line.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/id.js
+++ b/packages/dd-trace/src/id.js
@@ -26,6 +26,14 @@ class Identifier {
       : toNumberString(this._buffer, radix)
   }
 
+  toBigInt () {
+    let result = BigInt(0)
+    for (let i = 0; i < this._buffer.length; i++) {
+      result = (result << BigInt(8)) | BigInt(this._buffer[i])
+    }
+    return result
+  }
+
   toBuffer () {
     return this._buffer
   }
--- a/packages/dd-trace/src/id.js
+++ b/packages/dd-trace/src/id.js
@@ -27,11 +27,7 @@ class Identifier {
   }
 
   toBigInt () {
-    let result = BigInt(0)
-    for (let i = 0; i < this._buffer.length; i++) {
-      result = (result << BigInt(8)) | BigInt(this._buffer[i])
-    }
-    return result
+    return Buffer.from(this._buffer).readBigUInt64BE(0)
   }
 
   toBuffer () {
--- a/packages/dd-trace/src/opentracing/span_context.js
+++ b/packages/dd-trace/src/opentracing/span_context.js
@@ -43,6 +43,12 @@ class DatadogSpanContext {
     }
   }
 
+  // toTraceIdBigInt returns the traceId as a BigInt
+  // in case of a 128 bit traceId, it returns the lower 64 bits
+  toTraceIdBigInt () {
+    return this._traceId.toBigInt()
+  }
+
   toTraceId (get128bitId = false) {
     if (get128bitId) {
       return this._traceId.toBuffer().length <= 8 && this._trace.tags[TRACE_ID_128]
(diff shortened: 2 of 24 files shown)


BridgeAR (commented):
This is really great work!
As far as I can tell, this does implement the spec as defined. LGTM with my comment being addressed.

packages/dd-trace/src/opentracing/span_context.js
```diff
@@ -43,6 +43,12 @@ class DatadogSpanContext {
     }
   }
 
+  // toTraceIdBigInt returns the traceId as a BigInt
+  // in case of a 128 bit traceId, it returns the lower 64 bits
+  toTraceIdBigInt () {
```
rochdev, line 48: This adds a public API. It should either be documented if we expect users to use it, or we should use the ID directly if we prefer to keep this private.

packages/dd-trace/src/sampler.js
```diff
@@ -16,10 +32,21 @@ class Sampler {
...
   /**
-   * @returns {boolean}
+   * Determines whether a trace/span should be sampled based on the configured sampling rate.
+   *
+   * @param {SpanContext} context - The span context to evaluate.
+   * @returns {boolean} `true` if the trace/span should be sampled, otherwise `false`.
    */
-  isSampled () {
-    return this._rate === 1 || Math.random() < this._rate
+  isSampled (context) {
```
rochdev, line 40: This could accept a span as well as a context to avoid having to call `.context()` everywhere.

packages/dd-trace/src/id.js
```diff
@@ -26,6 +26,14 @@ class Identifier {
       : toNumberString(this._buffer, radix)
   }
 
+  toBigInt () {
+    let result = BigInt(0)
+    for (let i = 0; i < this._buffer.length; i++) {
+      result = (result << BigInt(8)) | BigInt(this._buffer[i])
+    }
```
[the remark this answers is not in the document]
rochdev: It should even be possible at this point to just always use a BigInt internally instead of a byte array which will remove the need for a conversion and could potentially be faster than the current approach.

rochdev (changes requested):

packages/dd-trace/src/sampler.js
genesor: > This could accept a span as well as a context to avoid having to call `.context()` everywhere.

Indeed @rochdev this would be better on the caller side. 

Should we accept both Span and SpanContext or just Span ?

rochdev: I'd say to accept both so that the caller can pass what they have.

packages/dd-trace/src/id.js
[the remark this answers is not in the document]
genesor: @BridgeAR @rochdev I changed the code to `Buffer.readBigUInt64BE` instead of manually shifting bits. I should be safer. 

About the BigInt instead of byte array I'd agree (especially for clarity purpose) but I'm afraid of the impact that would change the scope of this PR.

rochdev (approved):

BridgeAR (approved):

PR #5713, opened 2025-05-13
Pin shared pipeline

TonyCTHsu:
### What does this PR do?

Pin a version to shared pipeline to provide consistent result.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,7 +6,9 @@ stages:
   - macrobenchmarks
 
 include:
-  - remote: https://gitlab-templates.ddbuild.io/libdatadog/include/one-pipeline.yml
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -6,7 +6,9 @@ stages:
   - macrobenchmarks
 
 include:
-  - remote: https://gitlab-templates.ddbuild.io/libdatadog/include/one-pipeline.yml
+  - project: DataDog/apm-reliability/libdatadog-build
```
bengl, line 9: This is as good a time as any to put a comment in here linking to the actual repo. It's nearly impossible to find based on the information here alone.

bengl (approved):

PR #5716, opened 2025-05-13
Cache tainted value of query params in express v5

IlyasShabi:
### What does this PR do?

Cache tainted value of query param in express v5 in order to reuse it again if another req.query is called

### Motivation

[APPSEC-57134](https://datadoghq.atlassian.net/browse/APPSEC-57134)

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes

[APPSEC-57134]: https://datadoghq.atlassian.net/browse/APPSEC-57134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/test-optimization.yml
+++ b/.github/workflows/test-optimization.yml
@@ -63,7 +63,7 @@ jobs:
     strategy:
       matrix:
         version: [oldest, latest]
-        framework: [cucumber, selenium, jest, mocha]
+        framework: [cucumber, selenium, mocha]
     runs-on: ubuntu-latest
     env:
       DD_SERVICE: dd-trace-js-integration-tests
@@ -173,21 +173,21 @@ jobs:
       - uses: codecov/codecov-action@ad3126e916f78f00edff4ed0317cf185271ccc2d # v5.4.2
 
   # TODO: fix performance issues and test more Node versions
-  plugin-jest:
-    runs-on: ubuntu-latest
-    env:
-      PLUGINS: jest
-    steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-      - uses: ./.github/actions/testagent/start
-      - uses: ./.github/actions/node/active-lts
-      - uses: ./.github/actions/install
-      - run: yarn test:plugins:ci
-      - if: always()
-        uses: ./.github/actions/testagent/logs
-        with:
-          suffix: plugins-${{ github.job }}
-      - uses: codecov/codecov-action@ad3126e916f78f00edff4ed0317cf185271ccc2d # v5.4.2
+  # plugin-jest:
+  #   runs-on: ubuntu-latest
+  #   env:
+  #     PLUGINS: jest
+  #   steps:
+  #     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+  #     - uses: ./.github/actions/testagent/start
+  #     - uses: ./.github/actions/node/active-lts
+  #     - uses: ./.github/actions/install
+  #     - run: yarn test:plugins:ci
+  #     - if: always()
+  #       uses: ./.github/actions/testagent/logs
+  #       with:
+  #         suffix: plugins-${{ github.job }}
+  #     - uses: codecov/codecov-action@ad3126e916f78f00edff4ed0317cf185271ccc2d # v5.4.2
 
   plugin-mocha:
     runs-on: ubuntu-latest
--- a/scripts/verify-ci-config.js
+++ b/scripts/verify-ci-config.js
@@ -148,7 +148,7 @@ checkPlugins(path.join(__dirname, '..', '.github', 'workflows', 'test-optimizati
     .filter(file => fs.existsSync(path.join(__dirname, '..', 'packages', file, 'test')))
     .map(file => file.replace('datadog-plugin-', ''))
   for (const plugin of allPlugins) {
-    if (!allTestedPlugins.has(plugin)) {
+    if (!allTestedPlugins.has(plugin) && plugin !== 'jest') {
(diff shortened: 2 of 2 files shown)


packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
```diff
@@ -63,7 +63,33 @@ class TaintTrackingPlugin extends SourceIastPlugin {
...
+          if (!currentValue) continue
+
+          const cachedValue = iastContext.queryCache.get(key)
+
+          if (cachedValue === currentValue) {
+            // Reuse cached tainted value
+            query[key] = cachedValue
+          } else {
+            // Taint new value and cache it
+            const taintedValue = typeof currentValue === 'string'
```
uurien, line 85: if `currentValue` could be an object, the `if` of L80 will be `false` because the object instances are going to be diferent.

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
```diff
@@ -63,7 +63,18 @@ class TaintTrackingPlugin extends SourceIastPlugin {
...
+      ({ query }) => {
+        const iastContext = getIastContext(storage('legacy').getStore())
+        if (!iastContext || !query) return
+
+        if (!iastContext.queryCache) {
+          iastContext.queryCache = taintObject(iastContext, query, HTTP_REQUEST_PARAMETER)
+        }
+
+        for (const key in query) {
+          query[key] = iastContext.queryCache[key]
```
uurien, line 75: if we are not checking the previous value of the property, aren't we breaking the expected behaviour of express doing this?
I don't know what is the point of express parsing the whole query each time that the getter is called, but if different values are expected, we are breaking the behaviour.
We should iterate in the query object and compare each string with the original, if the value is the same, reuse, if it is not, create new taintObject.

uurien: And if multiple values are possible for the same key... we should store a string in the cache with possible values, not only one value.

IlyasShabi: @uurien I pushed a new commit with what we talked about yesterday, could you please check again

packages/dd-trace/src/appsec/iast/taint-tracking/operations-taint-object.js
```diff
@@ -40,6 +40,55 @@ function taintObject (iastContext, object, type) {
...
+
+  return traverseAndTaint(query, iastContext.queryCache, '', type, transactionId)
+}
+
+function traverseAndTaint (value, cache, path, type, transactionId) {
+  if (value == null) return value
+
+  if (typeof value === 'string') {
+    // If we already have a tainted version of this exact string in the cache
+    if (typeof cache === 'string') {
```
uurien, line 59:
```suggestion
    if (cache === value) {
```

packages/dd-trace/src/appsec/iast/taint-tracking/operations-taint-object.js
```diff
@@ -40,6 +40,55 @@ function taintObject (iastContext, object, type) {
...
+    // If we already have a tainted version of this exact string in the cache
+    if (value === cache) {
+      return cache
+    }
+
+    return TaintedUtils.newTaintedString(transactionId, value, path, type)
+  }
+
+  if (typeof value === 'object') {
+    const isArray = Array.isArray(value)
```
uurien, line 67: I don't like the name `isArray` which is like the function name. Maybe something like `valueIsArray` or `isValueArray`?

packages/dd-trace/src/appsec/iast/taint-tracking/operations-taint-object.js
```diff
@@ -40,6 +40,55 @@ function taintObject (iastContext, object, type) {
...
+
+      value[key] = traverseAndTaint(
+        childValue,
+        cache[key],
+        childPath,
+        type,
+        transactionId
+      )
+
+      cache[key] = value[key]
```
uurien, line 84: If an application modifies `req.query` object, the cache will be also updated:

```
(req, res) => {
  req.query.paramThatNotExists = 'AAAA'
// or also
  req.query.paramThatExists = 'aaaa' // => our cached object will be updated
}
```

our object in cache must be different instances of objects that are in the original query object, only the string instances can be the same instances.

packages/dd-trace/src/appsec/iast/taint-tracking/operations-taint-object.js
```diff
@@ -40,6 +43,43 @@ function taintObject (iastContext, object, type) {
...
+  return query
+}
+
+function traverseAndTaint (node, path, cache, transactionId) {
+  if (node == null) return node
+
+  if (typeof node === 'string') {
+    if (cache.get(path) === node) {
+      return cache.get(path)
+    }
```
uurien: Prevent double map access:

```suggestion
    const cachedValue = cache.get(path)
    if (cachedValue === node) {
      return cachedValue
    }
```

packages/dd-trace/src/appsec/iast/taint-tracking/operations-taint-object.js
```diff
@@ -40,6 +43,43 @@ function taintObject (iastContext, object, type) {
...
+
+    const tainted = TaintedUtils.newTaintedString(transactionId, node, path, HTTP_REQUEST_PARAMETER)
+    cache.set(path, tainted)
+    return tainted
```
uurien, line 69:
```suggestion

    return tainted
```

CarlesDD (approved):
LGTM

PR #5717, opened 2025-05-13
simpler adding --no-async-context-frame

bengl:
And more correct.


BridgeAR:
The CI failure seems unrelated as it also fails on master: https://github.com/DataDog/dd-trace-js/actions/runs/15000186690/job/42144791850

PR #5718, opened 2025-05-13
tracing: ensure router next function isn't double wrapped leading to infinite loop

wconti27:
### What does this PR do?
The router wrapNext function wraps `next`. A customer was seeing an exception from maximum call stack being hit, and the issue appears to be that next was being wrapped and called multiple times. 

```
unhandledException Maximum call stack size exceeded
RangeError: Maximum call stack size exceeded
at trim_prefix (/app/node_modules/express/lib/router/index.js:326:13)
at /app/node_modules/express/lib/router/index.js:286:9
at Function.process_params (/app/node_modules/express/lib/router/index.js:346:12)
at Function.process_params (/app/node_modules/dd-trace/packages/datadog-instrumentations/src/express.js:134:23)
at next (/app/node_modules/express/lib/router/index.js:280:10)
at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:97:5)
at trim_prefix (/app/node_modules/express/lib/router/index.js:328:13)
at /app/node_modules/express/lib/router/index.js:286:9
at Function.process_params (/app/node_modules/express/lib/router/index.js:346:12)
at Function.process_params (/app/node_modules/dd-trace/packages/datadog-instrumentations/src/express.js:134:23)
at next (/app/node_modules/express/lib/router/index.js:280:10)
at next (/app/node_modules/dd-trace/packages/datadog-instrumentations/src/router.js:100:12)
at next (/app/node_modules/dd-trace/packages/datadog-shimmer/src/shimmer.js:35:21)
at /app/node_modules/express/lib/router/index.js:646:15
at next (/app/node_modules/express/lib/router/index.js:265:14)
at next (/app/node_modules/dd-trace/packages/datadog-instrumentations/src/router.js:100:12)
```

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/router.js
+++ b/packages/datadog-instrumentations/src/router.js
@@ -13,6 +13,8 @@ function createWrapRouterMethod (name) {
   const errorChannel = channel(`apm:${name}:middleware:error`)
   const nextChannel = channel(`apm:${name}:middleware:next`)
 
+  const DD_WRAPPED_SYMBOL = Symbol('_datadog_instrumentation_wrapped')
+
   const layerMatchers = new WeakMap()
   const regexpCache = Object.create(null)
 
@@ -84,10 +86,15 @@ function createWrapRouterMethod (name) {
   }
 
   function wrapNext (req, next) {
+    if (next[DD_WRAPPED_SYMBOL]) {
+      return next
+    }
     return shimmer.wrapFunction(next, next => function (error) {
       if (error && error !== 'route' && error !== 'router') {
         errorChannel.publish({ req, error })
       }
+      // ensure we don't wrap the same next function twice
+      next[DD_WRAPPED_SYMBOL] = true
 
       nextChannel.publish({ req })
       finishChannel.publish({ req })


packages/datadog-instrumentations/src/router.js
```diff
@@ -84,10 +86,15 @@ function createWrapRouterMethod (name) {
...
   function wrapNext (req, next) {
+    if (next[DD_WRAPPED_SYMBOL]) {
+      return next
+    }
     return shimmer.wrapFunction(next, next => function (error) {
       if (error && error !== 'route' && error !== 'router') {
         errorChannel.publish({ req, error })
       }
+      // ensure we don't wrap the same next function twice
+      next[DD_WRAPPED_SYMBOL] = true
```
BridgeAR, line 97: The wrapping happens as soon as `wrapFunction()` is called. So in case the method is executed later, we could have still instrumented it multiple times, if we don't immediately set the symbol to prevent a second instrumentation.

I would also use a WeakSet to prevent manipulating the next code.

```js
const instrumentedNextInstances = new WeakSet()
```

```suggestion
    // Ensure we don't wrap the same next function twice
    if (instrumentedNextInstances.has(next)) {
      return next
    }
    instrumentedNextInstances.add(next)

    return shimmer.wrapFunction(next, next => function (error) {
      if (error && error !== 'route' && error !== 'router') {
        errorChannel.publish({ req, error })
      }
```

BridgeAR (commented):
I don't feel like this is the right spot to address the problem. The reason for my thought is, that the next function instrumentation always returns a new method. If that is not the case, it means a method is reused in multiple places and in a way that it's also used in a kind of child call. I believe that could cause multiple issues, not only here. This makes me wonder if the fix would not be about not even reaching this spot in the first place by doing something else (which I do not have an answer to :D )

PR #5721, opened 2025-05-13
chore(ssi): send injection and instrumentation telemetry

mabdinur:
### What does this PR do?

Submits the value of `DD_INJECT_FORCE` and `DD_INJECTION_ENABLED` (set by the injector) to the instrumentation telemetry platform. 

Compute `DD_INSTRUMENTATION_SOURCE`, this configuration will track whether a processes was successfully instrumented via ssi. This PR adds the values `ssi` and `unknown`. 

A future PR will track whether the ddtrace products were enabled manually (via imports) or set up only using command line arguments + env_vars (via --require dd-trace/init).

### Motivation

We a mechanism to track how services were instrumented, this will allow us to better troubleshoot customer setups and provide meaningful remediation.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/integration-tests/init.spec.js
+++ b/integration-tests/init.spec.js
@@ -36,18 +36,18 @@ function testInjectionScenarios (arg, filename, esmWorks = false) {
 
       if (currentVersionIsSupported) {
         context('without DD_INJECTION_ENABLED', () => {
-          it('should initialize the tracer', () => doTest('init/trace.js', 'true\n'))
-          it('should initialize instrumentation', () => doTest('init/instrument.js', 'true\n'))
+          it('should initialize the tracer', () => doTest('init/trace.js', 'true\n', [], 'ssi'))
+          it('should initialize instrumentation', () => doTest('init/instrument.js', 'true\n', [], 'ssi'))
           it(`should ${esmWorks ? '' : 'not '}initialize ESM instrumentation`, () =>
-            doTest('init/instrument.mjs', `${esmWorks}\n`))
+            doTest('init/instrument.mjs', `${esmWorks}\n`, []))
         })
       }
       context('with DD_INJECTION_ENABLED', () => {
         useEnv({ DD_INJECTION_ENABLED })
 
-        it('should not initialize the tracer', () => doTest('init/trace.js', 'false\n'))
-        it('should not initialize instrumentation', () => doTest('init/instrument.js', 'false\n'))
-        it('should not initialize ESM instrumentation', () => doTest('init/instrument.mjs', 'false\n'))
+        it('should not initialize the tracer', () => doTest('init/trace.js', 'false\n', []))
+        it('should not initialize instrumentation', () => doTest('init/instrument.js', 'false\n', []))
+        it('should not initialize ESM instrumentation', () => doTest('init/instrument.mjs', 'false\n', []))
       })
     })
     context('when dd-trace in the app dir', () => {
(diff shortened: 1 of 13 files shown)


packages/dd-trace/src/config.js
```diff
@@ -889,6 +894,8 @@ class Config {
     this._setString(env, 'iast.telemetryVerbosity', DD_IAST_TELEMETRY_VERBOSITY)
     this._setBoolean(env, 'iast.stackTrace.enabled', DD_IAST_STACK_TRACE_ENABLED)
     this._setArray(env, 'injectionEnabled', DD_INJECTION_ENABLED)
+    this._setString(env, 'instrumentationSource', DD_INSTRUMENTATION_SOURCE)
+    this._setString(defaults, 'injectionForce', DD_INJECT_FORCE)
```
mabdinur:
```suggestion
    this._setString(env, 'injectionForce', DD_INJECT_FORCE)
```

packages/dd-trace/src/config.js
```diff
@@ -594,6 +594,9 @@ class Config {
     this._setValue(defaults, 'trace.aws.addSpanPointers', true)
     this._setValue(defaults, 'trace.dynamoDb.tablePrimaryKeys')
     this._setValue(defaults, 'trace.nativeSpanEvents', false)
+    this._setValue(defaults, 'instrumentationSource', 'unknown')
+    this._setValue(defaults, 'injectionForce', '')
+    this._setValue(defaults, 'DD_INJECTION_ENABLED', '')
```
mabdinur:
```suggestion
    this._setValue(defaults, 'injectionEnabled', '')
```

packages/dd-trace/src/guardrails/index.js
```diff
@@ -57,6 +57,9 @@ function guard (fn) {
     var result = fn()
     telemetry('complete', ['injection_forced:' + (forced && initBailout ? 'true' : 'false')])
     log.info('Application instrumentation bootstrapping complete')
+    // Ensure the instrumentation source is set for the current process and potential 
+    // child processes.
+    process.env.DD_INSTRUMENTATION_SOURCE = 'ssi'
```
BridgeAR: Guardrails should likely not be touched. They just guarantee that nothing breaks with auto instrumentation.

integration-tests/init/instrument.js
```diff
@@ -13,8 +13,9 @@ const server = http.createServer((req, res) => {
     res.on('data', () => {})
     res.on('end', () => {
       server.close()
-      // eslint-disable-next-line no-console
+      /* eslint-disable no-console */
       console.log(gotEvent)
+      console.log('instrumentation source:', global._ddtrace._tracer._config.instrumentationSource)
```
BridgeAR: Nit: I do think it is good to not disable the lint rule completely and we can just replace it with a single call.

If you go with that, it also applies to the other files :)

```suggestion
      // eslint-disable-next-line no-console
      console.log(gotEvent, '\ninstrumentation source:', global._ddtrace._tracer._config.instrumentationSource)
```

packages/dd-trace/src/telemetry/telemetry.js
```diff
@@ -319,7 +319,10 @@ const nameMapping = {
   clientIpHeader: 'DD_TRACE_CLIENT_IP_HEADER',
   'grpc.client.error.statuses': 'DD_GRPC_CLIENT_ERROR_STATUSES',
   'grpc.server.error.statuses': 'DD_GRPC_SERVER_ERROR_STATUSES',
-  traceId128BitLoggingEnabled: 'DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED'
+  traceId128BitLoggingEnabled: 'DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED',
+  instrumentationSource: 'instrumentation_source',
+  injectionEnabled: 'ssi_injection_enabled',
+  injectionForce: 'ssi_forced_injection_enabled'
```
BridgeAR: I guess this is spected and we should keep them as such while I wonder if it's not easier to just map them to the environment variables. That way the connection is very easy.

integration-tests/init.spec.js
```diff
@@ -106,38 +106,38 @@ function testRuntimeVersionChecks (arg, filename) {
...
             it('should initialize the tracer, if DD_INJECT_FORCE', () =>
-              doTestForced('true\n', ...telemetryGood))
+              doTestForced('true\n', telemetryGood, 'ssi'))
           })
           context('with debug', () => {
             useEnv({ DD_TRACE_DEBUG })
 
             it('should initialize the tracer', () =>
-              doTest('Application instrumentation bootstrapping complete\ntrue\n', ...telemetryGood))
+              doTest('Application instrumentation bootstrapping complete\ntrue\n', telemetryGood, 'ssi'))
```
BridgeAR, line 138: I am not sure about these spreading changes here. If I read the code correct, the arguments are now passed through as array as second argument to `runAndCheckWithTelemetry` and the second argument is `expectedOut`. That would change the behavior as far as I can tell. The same applies to the one below.

We later on do `assert.strictEqual(out, expectedOut)` and that would only be true if the array is reference equal. I feel like I somehow overlook something.

BridgeAR (commented):
This is mainly LGTM. I believe I overlook something in the tests and have a question around the array being passed through now.

packages/dd-trace/src/telemetry/telemetry.js
mabdinur: It would've been but the .NET tracer has been using these values for about a year now. I decided to just standardize on this value than deal with supporting a new value.

integration-tests/init.spec.js
mabdinur: I replaced `assert.strictEqual(out, expectedOut)` with a regex, so we don't need to worry about the array reference but I agree I wish I didn't have to make this change in so many places but it makes the test a little easier to read. Every test case that successfully instruments ddtrace via ssi should report instrumentation_source as ssi.

PR #5723, opened 2025-05-13
convert mysql from asyncresource to runstores

rochdev:
### What does this PR do?

Convert `mysql` from AsyncResource to runStores.

### Motivation

AsyncResource is broken in Node 24.

### DIFF
--- a/packages/datadog-instrumentations/src/mysql.js
+++ b/packages/datadog-instrumentations/src/mysql.js
@@ -1,10 +1,6 @@
 'use strict'
 
-const {
-  channel,
-  addHook,
-  AsyncResource
-} = require('./helpers/instrument')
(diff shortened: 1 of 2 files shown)


tlhunter (approved):

packages/datadog-instrumentations/src/mysql.js
```diff
@@ -1,10 +1,6 @@
 'use strict'
 
-const {
-  channel,
-  addHook,
-  AsyncResource
-} = require('./helpers/instrument')
+const { channel, addHook, AsyncResource } = require('./helpers/instrument')
```
tlhunter, line 3: Is the remaining usage of `AsyncResource` in this file intentional?

rochdev: They aren't, see #5840

Issue #5702, opened 2025-05-12
[BUG]: Incorrect Exit Code from Playwright Instrumentation


ToreyStapleton:
### Tracer Version(s)

5.51.0

### Node.js Version(s)

22.15.0

### Bug Report

When adding the `NODE_OPTIONS: -r dd-trace/ci/init` param to my existing Playwright test Github Actions workflow, the test runner is not returning an accurate exit code for a specific runtime error (`import` error on missing package) and causing the Github workflow to report a false positive.

### Reproduction Code

I have a minimum reproducible example here: https://github.com/ToreyStapleton/dd-test that is using only `playwright` and `dd-trace`.

The test `tests/example.spec.ts` is meant to fail, as it's importing from `moment` when that package that is not yet installed:

```
import { test, expect } from '@playwright/test';
import moment from 'moment';

test('has title', async ({ page }) => {
  moment.locale('en');
  await page.goto('https://playwright.dev/');

  // Expect a title "to contain" a substring.
  await expect(page).toHaveTitle(/Playwright/);
});
```

In Github Actions, without `dd-trace` steps included in the workflow, this action correctly fails and exits with error code 1: https://github.com/ToreyStapleton/dd-test/actions/runs/14973509428/job/42059855725

Then, after adding the `NODE_OPTIONS` param to my workflow, this action continues to fail with the same error but returns exit code 0 and "passes" in Github Actions: https://github.com/ToreyStapleton/dd-test/actions/runs/14973839114/job/42060976814

AKA this step correctly fails:
```
    - name: Run Playwright tests
      run: npx playwright test
```

Adding this causes the step to incorrectly report a success, despite failing with the same error:
```
    - name: Run Playwright tests
      run: npx playwright test
      env:
        NODE_OPTIONS: -r dd-trace/ci/init
```

### Bundling

Unsure

Mariovido:
@ToreyStapleton,

Thanks for reaching out and good catch. I think I have found the root cause of this. I have raised a PR to fix this and will be added in the next release. Once is merged and released I'll let you know.

You can keep track of this in this PR: https://github.com/DataDog/dd-trace-js/pull/5733

BridgeAR:
Fixes by https://github.com/DataDog/dd-trace-js/pull/5733

PR #5733, opened 2025-05-14
[test-optimization] [SDTEST-2024] Fix Playwright exit code status when it fails before running the test

Mariovido:
### What does this PR do?

With this change we make sure that we don't change the status of the exit code of the session if it fails before running any test.

### Motivation

We want to solve an issue raised by a user
Fixes https://github.com/DataDog/dd-trace-js/issues/5702

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/integration-tests/ci-visibility/playwright-tests-exit-code/exit-code-test.js
+++ b/integration-tests/ci-visibility/playwright-tests-exit-code/exit-code-test.js
@@ -1,4 +1,5 @@
 const { test, expect } = require('@playwright/test')
+// eslint-disable-next-line no-unused-vars
 const dummy = require('dummy') // This should not exist, so should throw an error
 
 test.beforeEach(async ({ page }) => {


anmarchenko (approved):

packages/datadog-instrumentations/src/playwright.js
```diff
@@ -610,7 +610,9 @@ function runnerHook (runnerExport, playwrightVersion) {
         totalAttemptToFixFailedTestCount += testStatuses.filter(status => status === 'fail').length
       }
 
-      if (totalFailedTestCount === totalAttemptToFixFailedTestCount) {
+      if (totalFailedTestCount > 0 &&
+          totalAttemptToFixFailedTestCount > 0 &&
```
BridgeAR, line 614: Nit: this condition is not needed. It would be caught with the next condition comparing the number of failures.

BridgeAR:
@Mariovido I just looked at our release proposal and saw that this is marked as minor. Being a fix, isn't it actually a patch? I just updated it, but please change it back, in case you think it should be a minor instead.

Mariovido:
@BridgeAR You're right it should be a patch. Thank you for noticing and changing it. :)

packages/datadog-instrumentations/src/playwright.js
Mariovido: Yeah it makes sense. Next time that I touch this file I'll change this :)

PR #5737, opened 2025-05-15
Add an env var for executing integration tests without a sandbox

szegedi:
### What does this PR do?
Adds `DD_NO_INTEGRATION_TESTS_SANDBOX` as a test-only env var, and when set to `1` or `true` then it runs the integration tests without a sandbox.

### Motivation
There was already support for running without sandbox, but it required uncommenting code and executing `yarn link` commands externally. This PR now packages the whole functionality so it can be triggered by an env var. 

### Additional information
When sandbox is used, all the dependencies will be installed in the sandbox with their public releases. On a developer machine some dependencies can be yarn-linked in dd-trace. In this case, the integration tests run in a sandbox won't test against them. This is where the execution without sandbox comes in handy. (They also start up significantly faster as the sandbox doesn't have to be built.)


BridgeAR (approved):
LGTM, since the environment variable is only checked in our tests

PR #5742, opened 2025-05-16
Don't use DD_ prefix on test-only env var

szegedi:
### What does this PR do?
Removes `DD_` prefix from a test-only env var.

### Motivation
It was pointed out we only use `DD_` prefix for customer-exposed env vars.


nsavoire (approved):

BridgeAR (commented):
The DD_ envs should be fine when being used solely in tests. It would just be difficult in case we use them in the source code to trigger test behavior :)

BridgeAR (approved):

PR #5743, opened 2025-05-16
Use AsyncContextFrame for storing profiler sampling context

szegedi:
### What does this PR do?
Implements the dd-trace-js bits for using `AsyncContextFrame` to store the profiler sampling context. The pprof-nodejs bits shipped in https://github.com/DataDog/pprof-nodejs/pull/186 and then again in https://github.com/DataDog/pprof-nodejs/pull/229

### Motivation
With `AsyncContextFrame` available in Node 24 by default, we can use it to store the profiler's sampling context (tracing information we want to incorporate into profiler samples.) This then allows us to no longer use the `async_hooks.createHook` `before` callbacks. These callbacks will be deprecated in a future version of Node.js so it is vital we migrate away from using them. Enabling the callbacks also triggers deoptimizations, and we need to execute a callback on every single context switch so their use is generally detrimental to performance.

### Additional Notes
The integration tests will use the new functionality on Node.js 23 or later. Async context frames are available from Node.js 22.7.0 so we could eventually extend the compatibility downwards, but for that we need pprof-nodejs to also support this mode for versions earlier than 23.

Jira: [PROF-11771]

[PROF-11771]: https://datadoghq.atlassian.net/browse/PROF-11771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dd-trace/src/profiling/config.js
```diff
@@ -209,6 +211,33 @@ class Config {
...
 
+    const that = this
+    function turnOffAsyncContextFrame (msg) {
+      that.logger.warn(
+        `DD_PROFILING_ASYNC_CONTEXT_FRAME_ENABLED was set ${msg}, it will have no effect.`)
+      that.asyncContextFrameEnabled = false
+    }
+
+    this.asyncContextFrameEnabled = isTrue(coalesce(options.useAsyncContextFrame,
+      DD_PROFILING_ASYNC_CONTEXT_FRAME_ENABLED, false))
```
BridgeAR: Nit (I would like to remove the library):

```suggestion
    this.asyncContextFrameEnabled = isTrue(options.useAsyncContextFrame ??
      DD_PROFILING_ASYNC_CONTEXT_FRAME_ENABLED ?? false))
```

packages/dd-trace/src/profiling/config.js
```diff
@@ -209,6 +211,33 @@ class Config {
...
+          turnOffAsyncContextFrame('with --no-async-context-frame')
+        }
+      } else if (satisfies(process.versions.node, '>=23.0.0')) {
+        if (!process.execArgv.includes('--experimental-async-context-frame')) {
+          turnOffAsyncContextFrame('without --experimental-async-context-frame')
+        }
+      } else {
+        // NOTE: technically, this should work starting with 22.7.0 but it would
+        // require a change in pprof-nodejs too.
+        turnOffAsyncContextFrame('but it requires at least Node 23')
```
BridgeAR: Nit
```suggestion
        turnOffAsyncContextFrame('but it requires at least Node.js 23')
```

packages/dd-trace/src/profiling/profilers/wall.js
```diff
@@ -17,29 +17,61 @@ const TRACE_ENDPOINT_LABEL = 'trace endpoint'
...
 let kSampleCount
+let kCPEDContextCount
 
 function getActiveSpan () {
   const store = storage('legacy').getStore()
   return store && store.span
 }
 
+function toBigInt (spanId) {
+  return spanId !== null && typeof spanId === 'object' ? spanId.toBigInt() : spanId
```
BridgeAR, line 36: I believe that would suffice in this case (either it is a ID or it is not defined)

```suggestion
  return spanId?.toBigInt()
```

packages/dd-trace/src/profiling/profilers/wall.js
```diff
@@ -17,29 +17,61 @@ const TRACE_ENDPOINT_LABEL = 'trace endpoint'
...
+  // Converting spanIDs to bigints is not necessary as generateLabels will do it
+  // too. When we don't use async context frame, we can convert them when the
+  // sample is taken though so we amortize the latency of operations. It is an
+  // optimization.
+  if (context.spanId !== null && typeof context.spanId === 'object') {
+    context.spanId = context.spanId.toBigInt()
+  }
+  if (context.rootSpanId !== null && typeof context.rootSpanId === 'object') {
+    context.rootSpanId = context.rootSpanId.toBigInt()
+  }
```
nsavoire:
```suggestion
  context.spanId = toBigInt(context.spanId)
  context.rootSpanId = toBigInt(context.rootSpanId)
```

packages/dd-trace/src/profiling/config.js
```diff
@@ -209,6 +211,33 @@ class Config {
...
+    this.asyncContextFrameEnabled = isTrue(coalesce(options.useAsyncContextFrame,
+      DD_PROFILING_ASYNC_CONTEXT_FRAME_ENABLED, false))
+    if (this.asyncContextFrameEnabled) {
+      if (satisfies(process.versions.node, '>=24.0.0')) {
+        if (process.execArgv.includes('--no-async-context-frame')) {
+          turnOffAsyncContextFrame('with --no-async-context-frame')
+        }
+      } else if (satisfies(process.versions.node, '>=23.0.0')) {
+        if (!process.execArgv.includes('--experimental-async-context-frame')) {
+          turnOffAsyncContextFrame('without --experimental-async-context-frame')
```
nsavoire: What if `--no-async-context-frame` or `--experimental-async-context-frame` are set through `NODE_OPTIONS` env variable ?

packages/dd-trace/src/profiling/profilers/wall.js
```diff
@@ -17,29 +17,61 @@ const TRACE_ENDPOINT_LABEL = 'trace endpoint'
...
 function getActiveSpan () {
   const store = storage('legacy').getStore()
   return store && store.span
 }
 
+function toBigInt (spanId) {
+  return spanId !== null && typeof spanId === 'object' ? spanId.toBigInt() : spanId
+}
+
+function updateContext (context) {
```
nsavoire, line 39: That seems to replace `_updateContext` method that therefore could be removed.

packages/dd-trace/src/profiling/config.js
szegedi: That's fair – Koalas has a minor semantic difference in that it also ignores NaN in addition to null and undefined, which can be interesting when some values are expected to be numbers by the way of `parseInt` but in this particular case it's not an issue.

packages/dd-trace/src/profiling/profilers/wall.js
szegedi: Not necessarily, due to an optimization we can convert them to bigint early in `updateContext` when we aren't using `AsyncContextFrame` and then `_generateLabels` can already see them as bigints. It's an optimization, though (it amortizes the cost of converting ~13k span IDs to bigint over the course of a minute, instead having it concentrated when the profile is prepared for serialization.)

It would simplify the code if we removed this optimization, but it can marginally increase the app latency due to proifile encoding. OTOH, the optimization won't work with ACF anyhow so I expect it'll get phased out eventually anyhow.

packages/dd-trace/src/profiling/profilers/wall.js
szegedi: Fair. The idea here was to spare an assignment when they're undefined, but I guess it's not a big deal as it happens once every 10ms.

nsavoire (approved):
🚢 🚢 🚢

PR #5745, opened 2025-05-19
Ignore appsec startup error in serverless

uurien:
### What does this PR do?

Do not log Error when appsec startup fails in serverless
### Motivation

It is not expected to work in the tracer, it works in different way.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes

[APPSEC-57705]

[APPSEC-57705]: https://datadoghq.atlassian.net/browse/APPSEC-57705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/src/appsec/index.js
+++ b/packages/dd-trace/src/appsec/index.js
@@ -34,6 +34,7 @@ const UserTracking = require('./user_tracking')
 const { storage } = require('../../../datadog-core')
 const graphql = require('./graphql')
 const rasp = require('./rasp')
+const { isInServerlessEnvironment } = require('../serverless')
 
 const responseAnalyzedSet = new WeakSet()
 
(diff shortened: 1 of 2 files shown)


simon-id (approved):

CarlesDD (approved):

BridgeAR (commented):
Would it always fail in serverless?
If no, LGTM. If yes, would it not make sense to not allow activation at all?

uurien:
> Would it always fail in serverless?
If no, LGTM. If yes, would it not make sense to not allow activation at all?

I think that not always necessarily, but the binaries are usually deleted (in the Dockerfile that we provide they are deleted: https://github.com/DataDog/datadog-lambda-js/blob/main/Dockerfile#L38) that is why I only want to mute the error instead of disabling the whole feature.

PR #5747, opened 2025-05-19
Migrating Tedious from Async Resource to runStores

crysmags:
### What does this PR do?
Convert Tedious fromAsync Resource to runStores

### Motivation
Async Resource no longer works as expected in Node 24.

### DIFF
--- a/packages/datadog-instrumentations/src/tedious.js
+++ b/packages/datadog-instrumentations/src/tedious.js
@@ -22,8 +22,7 @@ addHook({ name: 'tedious', versions: ['>=1.0.0'] }, tedious => {
     }
 
     const connectionConfig = this.config
-    const payload = { queryOrProcedure, connectionConfig }
-    const ctx = { payload }
+    const ctx = { queryOrProcedure, connectionConfig }
 
     return startCh.runStores(ctx, () => {
       queryParent[queryField] = ctx.sql
--- a/packages/datadog-instrumentations/src/tedious.js
+++ b/packages/datadog-instrumentations/src/tedious.js
@@ -34,15 +34,8 @@ addHook({ name: 'tedious', versions: ['>=1.0.0'] }, tedious => {
           ctx.error = error
           errorCh.publish(ctx)
         }
-<<<<<<< HEAD
-        finishCh.publish()
-
-        return cb.apply(this, arguments)
-      }, null, request)
-=======
         return finishCh.runStores(ctx, cb, this, error, ...args)
       }
->>>>>>> 4d8d59fe1 (migrating tedious to use runStores)
 
       try {
         return makeRequest.apply(this, arguments)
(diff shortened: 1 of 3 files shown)


packages/datadog-plugin-tedious/src/index.js
```diff
@@ -8,7 +8,7 @@ class TediousPlugin extends DatabasePlugin {
   static get operation () { return 'request' } // TODO: change to match other database plugins
   static get system () { return 'mssql' }
 
-  start (payload) {
+  bindStart (payload) {
```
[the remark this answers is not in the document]
crysmags: @rochdev done!

packages/datadog-instrumentations/src/tedious.js
```diff
@@ -22,30 +21,27 @@ addHook({ name: 'tedious', versions: ['>=1.0.0'] }, tedious => {
       return makeRequest.apply(this, arguments)
     }
 
-    const callbackResource = new AsyncResource('bound-anonymous-fn')
-    const asyncResource = new AsyncResource('bound-anonymous-fn')
-
     const connectionConfig = this.config
+    const payload = { queryOrProcedure, connectionConfig }
+    const ctx = { payload }
```
rochdev:
```suggestion
    const ctx = { queryOrProcedure, connectionConfig }
```

Assigning to `ctx` directly will avoid requiring a change to any other subscribers.

packages/datadog-plugin-tedious/src/index.js
```diff
@@ -8,7 +8,8 @@ class TediousPlugin extends DatabasePlugin {
   static get operation () { return 'request' } // TODO: change to match other database plugins
   static get system () { return 'mssql' }
 
-  start (payload) {
+  bindStart (ctx) {
```
rochdev, line 11: `ctx.currentStore` needs to be returned from this function.

packages/datadog-plugin-tedious/src/index.js
```diff
@@ -8,7 +8,8 @@ class TediousPlugin extends DatabasePlugin {
...
+    const { payload } = ctx
```
rochdev:
```suggestion
```

This line is no longer needed if the payload is the context.

packages/datadog-instrumentations/src/tedious.js
```diff
@@ -22,30 +21,26 @@ addHook({ name: 'tedious', versions: ['>=1.0.0'] }, tedious => {
...
-    const asyncResource = new AsyncResource('bound-anonymous-fn')
-
     const connectionConfig = this.config
+    const ctx = { queryOrProcedure, connectionConfig }
 
-    return asyncResource.runInAsyncScope(() => {
-      const payload = { queryOrProcedure, connectionConfig }
-      startCh.publish(payload)
-      queryParent[queryField] = payload.sql
+    return startCh.runStores(ctx, () => {
```
crysmags, line 27: @rochdev, pairing with Ruben on this PR, do we need `runStores` for both `startCh` and `finishCh`, the latter is handling the callback?

rochdev: Yes, when there is a callback involved then `runStores` is needed for `finish` so that the parent context can be restored for the callback.

PR #5752, opened 2025-05-21
change ioredis service name

IlyasShabi:
### What does this PR do?
Fix ioredis service name to `redis` as defined in [L1039](https://github.com/DataDog/dd-trace-js/compare/ishabi/ioredis-service-name?expand=1#diff-660fbacbcd30e79ad55c4c69af9a034da3af7f8a37f8704d72b79617a93ee864R1039)

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -1042,7 +1042,7 @@ jobs:
           - 6379:6379
     env:
       PLUGINS: ioredis
-      SERVICES: ioredis
+      SERVICES: redis
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - uses: ./.github/actions/plugins/test


tlhunter:
IIRC the SERVICES env var is used to ensure the right backing service is running (e.g. the redis server is running for the redis package, the sql server is running for the tedious package, etc). So I think it should be correct that SERVICES=redis for the ioredis package.

However, I am now confused about my valkey PR, since there is no `packages/dd-trace/test/setup/services/valkey.js` file. I would expect it to give errors like in this PR ([output here](https://github.com/DataDog/dd-trace-js/actions/runs/15158336355/job/42618370593?pr=5752)).

BridgeAR:
Since it seems to work and break with this, let's keep it?
If we refactor the test instrumentation, that could also potentially be looked at. I just don't think it's crucial right now.

PR #5759, opened 2025-05-21
test: remove manual promise handling from test

tlhunter:
### What does this PR do?
- proposing an idealized form of basic asynchronous tests

### Motivation
- feed the llms with happy patterns

### DIFF
--- a/packages/datadog-plugin-ioredis/test/index.spec.js
+++ b/packages/datadog-plugin-ioredis/test/index.spec.js
@@ -29,25 +29,22 @@ describe('Plugin', () => {
 
         afterEach(() => agent.close({ ritmReset: false }))
 
-        it('should do automatic instrumentation when using callbacks', done => {
-          agent.assertSomeTraces(() => {}) // wait for initial info command
-          agent.assertSomeTraces(traces => {
-            expect(traces[0][0]).to.have.property('name', expectedSchema.outbound.opName)
-            expect(traces[0][0]).to.have.property('service', expectedSchema.outbound.serviceName)
-            expect(traces[0][0]).to.have.property('resource', 'get')
-            expect(traces[0][0]).to.have.property('type', 'redis')
-            expect(traces[0][0].meta).to.have.property('component', 'ioredis')
-            expect(traces[0][0].meta).to.have.property('db.name', '0')
-            expect(traces[0][0].meta).to.have.property('db.type', 'redis')
-            expect(traces[0][0].meta).to.have.property('span.kind', 'client')
-            expect(traces[0][0].meta).to.have.property('out.host', 'localhost')
-            expect(traces[0][0].meta).to.have.property('redis.raw_command', 'GET foo')
-            expect(traces[0][0].metrics).to.have.property('network.destination.port', 6379)
-          })
-            .then(done)
-            .catch(done)
+        it('should do automatic instrumentation when using callbacks', async () => {
+          await redis.get('foo')
 
-          redis.get('foo').catch(done)
+          await agent.assertFirstTrace(trace => {
+            expect(trace).to.have.property('name', expectedSchema.outbound.opName)
+            expect(trace).to.have.property('service', expectedSchema.outbound.serviceName)
+            expect(trace).to.have.property('resource', 'get')
+            expect(trace).to.have.property('type', 'redis')
+            expect(trace.meta).to.have.property('component', 'ioredis')
+            expect(trace.meta).to.have.property('db.name', '0')
+            expect(trace.meta).to.have.property('db.type', 'redis')
+            expect(trace.meta).to.have.property('span.kind', 'client')
+            expect(trace.meta).to.have.property('out.host', 'localhost')
+            expect(trace.meta).to.have.property('redis.raw_command', 'GET foo')
+            expect(trace.metrics).to.have.property('network.destination.port', 6379)
+          })
         })
 
         it('should run the callback in the parent context', () => {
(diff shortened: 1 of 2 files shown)


packages/dd-trace/test/plugins/agent.js
```diff
@@ -234,36 +234,37 @@ let availableEndpoints = DEFAULT_AVAILABLE_ENDPOINTS
...
     } catch (e) {
-      if (options && options.rejectFirst) {
-        clearTimeout(timeout)
-        deferred.reject(e)
+      if (options?.rejectFirst) {
+        clearTimeout(rejectionTimeout)
+        reject(e)
       } else {
-        error = error || e
+        error = error || e // if no spans match we report exactly the first mismatch error (which is unintuitive)
```
BridgeAR, line 267: We can use an AggregateError() instead

tlhunter: I'm not even sure that's a good solution...

Basically the problem is that if there are three spans we're examining, A + B + C, and we're writing a test against span B, but we expect some field that isn't perfectly equal (so really we're looking for B' instead of B), the problem is that we say we don't match A.

Really what we want to know to help debug is information about the closest matching span, B in this case. An aggregate error shows us mismatched information about A and B and C which is a little better but still not perfect.

If we do switch to an approach like you're suggesting where we declaratively compare deep objects then we could calculate which is the "closest span" and display that specific span's callback error.

BridgeAR (commented):
What about doing

```js
return agent.assertSpanShape(async () => {
  return redis.get('foo')
}, {
  name: 'foo'
  ...
})
```

```js
return agent.assertSpanShape(async () => {
  return redis.get('foo')
}, (trace) => {
  assert...
  return agent.passed
}, {
  traceNumber: 3
})
```

```js
return agent.assertSpanShape(async () => {
  return redis.get('foo')
}, [{ firstTrace }, { secondTrace }])
```

packages/dd-trace/test/plugins/agent.js
```diff
@@ -381,37 +389,85 @@ module.exports = {
...
   reset () {
     traceHandlers.clear()
     statsHandlers.clear()
   },
 
-  // Stop the mock agent, reset all expectations and wipe the require cache.
+  /**
+   * Stop the mock agent, reset all expectations and wipe the require cache.
+   * @param {Object} opts
+   * @param {boolean} opts.ritmReset - If true, resets the Require In The MIddle cache, useful for TODO.
```
tlhunter, line 467: @rochdev can you explain situations where a test author would want to use these two flags? I'd like to document them here so we get the reasons when autocompleting.

rochdev: Same for `wipe`. They're both things we shouldn't be doing.

packages/dd-trace/test/plugins/agent.js
```diff
@@ -234,36 +234,37 @@ let availableEndpoints = DEFAULT_AVAILABLE_ENDPOINTS
...
+    reject = _reject
+  })
 
-  const timeout = setTimeout(() => {
-    if (error) {
-      deferred.reject(error)
-    }
-  }, timeoutMs)
+  const rejectionTimeout = setTimeout(() => {
+    if (error) reject(error)
```
BridgeAR, line 248: I believe error is always truthy (if not throwing something falsy). Otherwise the timeout is cleared.

```suggestion
    reject(error)
```

packages/dd-trace/test/plugins/agent.js
```diff
@@ -275,7 +276,14 @@ function runCallback (callback, options, handlers) {
 }
 
 module.exports = {
-  // Load the plugin on the tracer with an optional config and start a mock agent.
+  /**
+   * Load the plugin on the tracer with an optional config and start a mock agent.
+   *
+   * @param {String|Array<String>} pluginName - Name or list of names of plugins to load
+   * @param {Object} config
+   * @param {Object} tracerConfig
```
BridgeAR, line 284:
```suggestion
   * @param {Record<string, unknown>} config
   * @param {Record<string, unknown>} tracerConfig
```

Object has weird results as type

packages/dd-trace/test/plugins/agent.js
```diff
@@ -381,37 +389,85 @@ module.exports = {
     }
   },
 
-  // Register handler to be executed each agent call, multiple times
+  /**
+   * Register handler to be executed each agent call, multiple times
```
BridgeAR, line 393:
```suggestion
   * Register handler to be executed on each agent call, multiple times
```

packages/dd-trace/test/plugins/agent.js
```diff
@@ -275,7 +276,14 @@ function runCallback (callback, options, handlers) {
 }
 
 module.exports = {
-  // Load the plugin on the tracer with an optional config and start a mock agent.
+  /**
+   * Load the plugin on the tracer with an optional config and start a mock agent.
+   *
+   * @param {String|Array<String>} pluginName - Name or list of names of plugins to load
```
BridgeAR, line 282:
```suggestion
   * @param {String|String[]} pluginName - Name or list of names of plugins to load
```

packages/dd-trace/test/plugins/agent.js
```diff
@@ -381,37 +389,85 @@ module.exports = {
...
+   * @param {span} span - For a given payload, the first span of the first trace.
+   */
+
+  /**
+   * This callback gets executed once for every payload received by the agent.
+   * It calls the callback with a `traces` argument which is an array of traces.
+   * Each of these traces is an array of spans.
+   *
+   * @param {testAssertionTracesCallback} callback - runs once per agent payload
+   * @param {Object} options
```
BridgeAR, line 428: We should name the concrete options here.

packages/dd-trace/test/plugins/agent.js
```diff
@@ -381,37 +389,85 @@ module.exports = {
...
-    return runCallback(callback, options, traceHandlers)
+    return runCallbackAgainstTraces(callback, options, traceHandlers)
+  },
+
+  /**
+   * Same as assertSomeTraces() but only provides the first span (traces[0][0])
+   * This callback gets executed once for every payload received by the agent.
+
+   * @param {testAssertionSpanCallback} callback - runs once per agent payload
+   * @param {Object} options
```
BridgeAR, line 440: Same as above: naming concrete options.

packages/datadog-plugin-ioredis/test/index.spec.js
```diff
@@ -29,25 +29,22 @@ describe('Plugin', () => {
...
+          await agent.assertFirstTrace(trace => {
+            expect(trace).to.have.property('name', expectedSchema.outbound.opName)
+            expect(trace).to.have.property('service', expectedSchema.outbound.serviceName)
+            expect(trace).to.have.property('resource', 'get')
+            expect(trace).to.have.property('type', 'redis')
+            expect(trace.meta).to.have.property('component', 'ioredis')
+            expect(trace.meta).to.have.property('db.name', '0')
+            expect(trace.meta).to.have.property('db.type', 'redis')
+            expect(trace.meta).to.have.property('span.kind', 'client')
+            expect(trace.meta).to.have.property('out.host', 'localhost')
```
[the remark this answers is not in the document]
watson: We could use a polyfill for `partialDeepStrictEqual ` in earlier versions of Node.js. We actually already have something similar here: https://github.com/DataDog/dd-trace-js/blob/68d5534d11b0b9272a4ca71507746974f0961abb/integration-tests/helpers/index.js#L363-L374

packages/dd-trace/test/plugins/agent.js
BridgeAR: Ok, thanks :)

PR #5762, opened 2025-05-22
wip(openai): trying to convert openai to use orchestrion

sabrenner:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-plugin-openai/src/endpoint-hooks/base.js
+++ b/packages/datadog-plugin-openai/src/endpoint-hooks/base.js
@@ -1,0 +1,171 @@
+'use strict'
+
+const TracingPlugin = require('../../../dd-trace/src/plugins/tracing')
+const { storage } = require('../../../datadog-core')
+const Sampler = require('../../../dd-trace/src/sampler')
+const { MEASURED } = require('../../../../ext/tags')
+
+// let normalize
+
+// const { DD_MAJOR } = require('../../../../version')
+
+class OpenAiBaseEndpointHook extends TracingPlugin {
+  static get operation () { return 'request' }
+  static get system () { return 'openai' }
+
+  constructor (services, utilities, ...args) {
+    super(...args)
+
+    const { metrics, logger } = services
+    this.metrics = metrics
+    this.logger = logger
+
+    const { normalize } = utilities
+    this.normalize = normalize
+
+    this.sampler = new Sampler(0.1) // default 10% log sampling
+  }
+
+  bindStart (ctx) {
+    const payloadTags = this.getPayloadTags(ctx)
+    const resource = this.getResource(ctx)
+
+    const span = this.startSpan('openai.request', {
+      service: this.config.service,
+      resource,
+      type: 'openai',
+      kind: 'client',
+      meta: {
+        ...payloadTags,
+        [MEASURED]: 1
+      }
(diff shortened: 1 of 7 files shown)


packages/datadog-plugin-openai/src/endpoint-hooks/completions.js
```diff
@@ -0,0 +1,125 @@
...
+
+function tryRequire (path) {
+  try {
+    return require(path)
+  } catch (e) {
+    return null
+  }
+}
+
+const OPENAI_VERSION = tryRequire('openai/version')
```
sabrenner, line 18: soooo janky and not where we want to do this but i'm not sure where else to do it - besides having a separate plugin for openai >=4.0 that sets this differently 😕 but then we'll have like 46 plugins (9 different functions traced + 5 that are already versioned + 9 by splitting it into v3 as well, all doubled for LLMObs plugins as well)

packages/datadog-plugin-openai/src/endpoint-hooks/completions.js
```diff
@@ -0,0 +1,125 @@
...
+
+    // we need to wrap the stream that the user will consume
+    // the stream is just a generator function, and each chunk could either be a buffer or an object
+
+    // we cannot separately tee and consume the stream, as to accurately represent the execution time
+    // in the users application - we wrap the stream for this reason
+
+    // also this is messy - just copied from the existing instrumentation
+    // it needs to be cleaned up
+    shimmer.wrap(result, 'parse', parse => function () {
```
sabrenner, line 58: comment explains it but this is one pain point so far - how do we properly wrap this stream? we don't want to consume it ourselves because we want to finish the span once the user has finished consuming it (assuming they do some stuff between each chunk). in the current approach in `datadog-instrumentations/src/openai.js` we do what we're doing here and wrap it directly... not sure how it translates to here if we **don't** want to be doing patching here (although, i would maybe argue it's not so much patching as instance wrapping, if that really represents a difference)

packages/datadog-plugin-openai/src/endpoint-hooks/base.js
```diff
@@ -0,0 +1,171 @@
...
+
+  end (ctx) { // sync because the promise types are custom for openai
+    const span = ctx.currentStore?.span
+    if (!span) return
+
+    const { result } = ctx
+    // instead of wrapping the result, queue up a separate promise to handle when the response resolves
+    // since we want the response headers as well, call `withResponse()` to get that
+    // while this makes it easier to manage on our side as opposed to wrapping, it does queue up another promise
+    result.withResponse().then(({ data, response }) => {
```
sabrenner, line 62: as comment explains - not ideal since we'll be queueing up a separate promise, but this way we don't have to do instance patching on the custom promise type.

the custom promise type is also the whole reason we can't do `tracePromise`, since itll return an object of a promise prototype not compatible with the openai node.js sdk api promise ([ref](https://github.com/openai/openai-node/blob/f3c6ba8fe9dc442e1cc33eace38c25825bb1c214/src/core.ts#L121))

PR #5765, opened 2025-05-23
Add support for profiler upload compression methods and levels

szegedi:
### What does this PR do?
Adds a `DD_PROFILING_DEBUG_UPLOAD_COMPRESSION` environment variable that supports setting a compression method for profile uploads. Supported values are "on", "off", "gzip", and "zstd" (which only works for Node.js 23.8.0 or above currently.) Default value is "on" and "on" itself is interpreted as "gzip" (although this can change to "zstd" in the future.) Additionally, "gzip" and "zstd" can accept a level between 0 and 9 after a dash (e.g. "zstd-5") to indicate a compression level.

### Motivation
zstd offers significantly better compression than gzip and often also lower CPU use. A representative CPU profile of about 460 kiB compresses to 124 kiB in 6ms with gzip level 6 while it compresses to the same size with zstd level 3 in 1ms. We don't expect this ever to be a user-configurable option (hence `DEBUG` in the name) but want to experiment with various options ourselves.

### Additional Notes
Jira: [PROF-11828]

[PROF-11828]: https://datadoghq.atlassian.net/browse/PROF-11828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/src/profiling/config.js
+++ b/packages/dd-trace/src/profiling/config.js
@@ -220,8 +220,6 @@ class Config {
         if (level > maxLevel) {
           logger.warn(`Invalid compression level ${level}. Will use ${maxLevel}.`)
           level = maxLevel
-        } else {
-          level |= 0
         }
       }
     }
--- a/packages/dd-trace/src/profiling/config.js
+++ b/packages/dd-trace/src/profiling/config.js
@@ -212,16 +212,17 @@ class Config {
         logger.warn(
           `Invalid compression level "${level0}". Will use default level.`)
         level = undefined
-      } else if (level < 0) {
-        // Best I can tell this can't happen due to splitting the env var at
-        // minus sign, but let's be exhaustive.
-        logger.warn(`Invalid compression level ${level}. Will use 0.`)
-        level = 0
-      } else if (level > 9) {
-        logger.warn(`Invalid compression level ${level}. Will use 9.`)
-        level = 9
+      } else if (level < 1) {
+        logger.warn(`Invalid compression level ${level}. Will use 1.`)
+        level = 1
       } else {
-        level |= 0
+        const maxLevel = { gzip: 9, zstd: 22 }[uploadCompression]
+        if (level > maxLevel) {
+          logger.warn(`Invalid compression level ${level}. Will use ${maxLevel}.`)
+          level = maxLevel
+        } else {
+          level |= 0
+        }
       }
     }
 
--- a/packages/dd-trace/src/profiling/config.js
+++ b/packages/dd-trace/src/profiling/config.js
@@ -15,6 +15,7 @@ const { GIT_REPOSITORY_URL, GIT_COMMIT_SHA } = require('../plugins/util/tags')
 const { tagger } = require('./tagger')
 const { isFalse, isTrue } = require('../util')
 const { getAzureTagsFromMetadata, getAzureAppMetadata } = require('../azure_metadata')
+const satisfies = require('semifies')
 
 class Config {
   constructor (options = {}) {
@@ -25,6 +26,7 @@ class Config {
       DD_PROFILING_CODEHOTSPOTS_ENABLED,
       DD_PROFILING_CPU_ENABLED,
       DD_PROFILING_DEBUG_SOURCE_MAPS,
+      DD_PROFILING_DEBUG_UPLOAD_COMPRESSION,
       DD_PROFILING_ENDPOINT_COLLECTION_ENABLED,
       DD_PROFILING_EXPERIMENTAL_CODEHOTSPOTS_ENABLED,
       DD_PROFILING_EXPERIMENTAL_CPU_ENABLED,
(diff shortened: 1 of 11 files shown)


packages/dd-trace/src/profiling/config.js
```diff
@@ -191,6 +193,44 @@ class Config {
...
+      } else if (level < 0) {
+        // Best I can tell this can't happen due to splitting the env var at
+        // minus sign, but let's be exhaustive.
+        logger.warn(`Invalid compression level ${level}. Will use 0.`)
+        level = 0
+      } else if (level > 9) {
+        logger.warn(`Invalid compression level ${level}. Will use 9.`)
+        level = 9
+      } else {
+        level |= 0
```
nsavoire, line 224: Not sure what is the intent here ?

packages/dd-trace/src/profiling/config.js
```diff
@@ -191,6 +193,44 @@ class Config {
...
+      } else if (isNaN(level)) {
+        logger.warn(
+          `Invalid compression level "${level0}". Will use default level.`)
+        level = undefined
+      } else if (level < 0) {
+        // Best I can tell this can't happen due to splitting the env var at
+        // minus sign, but let's be exhaustive.
+        logger.warn(`Invalid compression level ${level}. Will use 0.`)
+        level = 0
+      } else if (level > 9) {
```
nsavoire, line 220: Seems that zstd supports levels from 1 up to 22

packages/dd-trace/src/profiling/config.js
```diff
@@ -191,6 +193,44 @@ class Config {
...
+    let level = level0 ? parseInt(level0, 10) : undefined
+    if (level !== undefined) {
+      if (['on', 'off'].includes(uploadCompression)) {
+        logger.warn(`Compression levels are not supported for "${uploadCompression}".`)
+        level = undefined
+      } else if (isNaN(level)) {
+        logger.warn(
+          `Invalid compression level "${level0}". Will use default level.`)
+        level = undefined
+      } else if (level < 0) {
```
nsavoire, line 215: level 0 is not supported by zstd and disable compression for zlib, not sure if it should be allowed

packages/dd-trace/src/profiling/config.js
```diff
@@ -191,6 +193,44 @@ class Config {
     logExperimentalVarDeprecation('CPU_ENABLED')
     checkOptionWithSamplingContextAllowed(this.cpuProfilingEnabled, 'CPU profiling')
 
+    const uploadCompression0 = coalesce(options.uploadCompression, DD_PROFILING_DEBUG_UPLOAD_COMPRESSION, 'on')
+    let [uploadCompression, level0] = uploadCompression0.split('-')
+    if (!['on', 'off', 'gzip', 'zstd'].includes(uploadCompression)) {
```
nsavoire, line 198: What would you think of using 'default' instead of 'on' ?

packages/dd-trace/src/profiling/profiler.js
```diff
@@ -86,6 +89,28 @@ class Profiler extends EventEmitter {
             : `Found source maps for following files: [${Array.from(mapper.infoMap.keys()).join(', ')}]`
         })
       }
+
+      const clevel = config.uploadCompression.level
+      switch (config.uploadCompression.method) {
+        case 'gzip':
+          this._compressionFn = promisify(zlib.gzip)
+          if (clevel) {
```
nsavoire, line 97: this condition will be false for `level == 0` which is a valid level for zlib (although it implies no compression).

packages/dd-trace/src/profiling/profiler.js
```diff
@@ -218,7 +243,14 @@ class Profiler extends EventEmitter {
       // encode and export asynchronously
       for (const { profiler, profile } of profiles) {
         try {
-          encodedProfiles[profiler.type] = await profiler.encode(profile)
+          const encoded = await profiler.encode(profile)
+          let compressed
+          if (encoded instanceof Buffer && this._compressionFn !== undefined) {
```
nsavoire, line 248: The test on the type of `encoded` is there for unit tests ?

packages/dd-trace/src/profiling/config.js
szegedi: Levels are supposed to be integers. I figured we can silently turn `zstd-3.14` into `zstd-3`. But we can also warn about it?

packages/dd-trace/src/profiling/config.js
szegedi: There's a notebook attached to the Jira issue that specified the recognized values as these four.

nsavoire: Ok, I missed that for once we tried to standardize across profilers

packages/dd-trace/src/profiling/config.js
nsavoire: hum, `level` is the result of `parseInt(level0, 10)` so it should be an integer here

packages/dd-trace/src/profiling/profiler.js
szegedi: yes :-)

packages/dd-trace/src/profiling/config.js
szegedi: aaah, of course!

packages/dd-trace/src/profiling/config.js
szegedi: Agreed, I disabled them.

nsavoire (approved):

PR #5769, opened 2025-05-26
[DI] Fix race condition if breakpoint is hit while being removed

watson:
### What does this PR do?

The internal state kept by the DI code of which probes are active, might not be 100% in sync with the actual breakpoints configured via the Chrome DevTools Protocol (CDP).
    
This could happen if the probe is in the process of either being updated or removed while at the same time being hit in the instrumented application.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


BridgeAR (approved):
RSLGTM

I wonder if the test would really catch much during a refactoring but at least it's something.

PR #5770, opened 2025-05-26
[test-optimization] [SDTEST-2034] Send tag when there is no branch

Mariovido:
### What does this PR do?

When a GitHub is triggered with a _tag push_, the branch that we are sending is empty, resulting in an error. We want to solve this by sending the tag whenever this happen as the branch.

### Motivation

We want to avoid setting wrongly the library configurations.

### Plugin Checklist

- [x] Unit tests.

### DIFF
--- a/packages/datadog-plugin-cypress/src/cypress-plugin.js
+++ b/packages/datadog-plugin-cypress/src/cypress-plugin.js
@@ -67,7 +67,8 @@ const {
   GIT_BRANCH,
   CI_PROVIDER_NAME,
   CI_WORKSPACE_PATH,
-  GIT_COMMIT_MESSAGE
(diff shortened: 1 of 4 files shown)


juan-fernandez (dismissed):
> There is just one thing I was wondering about: should we maybe change branch to commitish? It could be a hash and other things.

that'd require backend changes and we want to keep this simple, as it's a rare edge case 

About adding a unit test, I think it's a good idea. We can keep it simple by adding a test in `packages/dd-trace/test/ci-visibility/exporters/ci-visibility-exporter.spec.js` in the `getLibraryConfiguration` section, instead of adding a test in every test framework

juan-fernandez (approved):

PR #5782, opened 2025-05-28
feat(tracer): implement process discovery feature

dmehala:
### What does this PR do?
This is PR store the tracer configuration in an memfd file.

Resolves [APMAPI-1071]

### Motivation

[Process Discovery](https://docs.google.com/document/d/1kcW6BLdYxXeTSUz31cBqoqfW1Jjs0IDljfKeUfIRQp4)

### Plugin Checklist

- [x] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes

I am not very familiar with Node.js, Javascript, nor dd-trace-js, so I have NO idea why the CI is mad. Please feel free to take over and make any necessary adjustments to fix the CI. Help would be grealy appreciated :)

image

[APMAPI-1071]: https://datadoghq.atlassian.net/browse/APMAPI-1071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/src/tracer.js
+++ b/packages/dd-trace/src/tracer.js
@@ -26,14 +26,14 @@ class DatadogTracer extends Tracer {
     flushStartupLogs(log)
 
     if (!config._isInServerlessEnvironment()) {
-      try {
-        const storeConfig = require('./tracer_metadata')
-        // Keep a reference to the handle, to keep the memfd alive in memory.
-        // It is read by the service discovery feature.
-        this._inmem_cfg = storeConfig(config)
-      } catch (error) {
-        log.warn('Could not store tracer configuration for service discovery', error)
+      const storeConfig = require('./tracer_metadata')
+      // Keep a reference to the handle, to keep the memfd alive in memory.
+      // It is read by the service discovery feature.
+      const metadata = storeConfig(config)
+      if (metadata === undefined) {
+        log.warn('Could not store tracer configuration for service discovery')
       }
+      this._inmem_cfg = metadata
     }
   }
 
--- a/packages/dd-trace/src/tracer_metadata.js
+++ b/packages/dd-trace/src/tracer_metadata.js
@@ -7,7 +7,7 @@ const tracerVersion = require('../../../version').VERSION
 function storeConfig (config) {
   const processDiscovery = libdatadog.maybeLoad('process-discovery')
   if (processDiscovery === undefined) {
-    throw new Error('Can\'t load process-discovery library')
+    return
   }
 
   const metadata = new processDiscovery.TracerMetadata(


package.json
```diff
@@ -84,7 +84,7 @@
     "node": ">=18"
   },
   "dependencies": {
-    "@datadog/libdatadog": "^0.6.0",
+    "@datadog/libdatadog": "git+https://github.com/DataDog/libdatadog-nodejs.git#a7b36c913f4fa48c19d46a488a4e46c153f5d4b4",
```
dmehala: NOTE: Temporary, until a new version of `libdatadog-nodejs` is released.

packages/dd-trace/src/tracer_metadata.js
```diff
@@ -0,0 +1,25 @@
...
+const tracerVersion = require('../../../package.json').version
+
+function storeConfig (config) {
+  const process_discovery = libdatadog.maybeLoad('process-discovery')
+  if (process_discovery === undefined) {
+    throw new Error('Can\'t load process-discovery library')
+  }
+
+  const metadata = new process_discovery.TracerMetadata(
+    "runtimeid",
```
dmehala: Need help from maintainers, how do I retrieve the `runtimeId`?

packages/dd-trace/src/tracer_metadata.js
```diff
@@ -0,0 +1,25 @@
+'use strict'
+
+// Load binding first to not import other modules if it throws
+const libdatadog = require('@datadog/libdatadog')
+const tracerVersion = require('../../../package.json').version
```
BridgeAR:
```suggestion
const tracerVersion = require('../../../version')
```

That actually exposes the information as well and it's less work to read it than the package.json :)

We could just export the raw version from there as well. That way there is no combining needed either.

yarn.lock
```diff
@@ -206,10 +206,9 @@
     "@babel/helper-string-parser" "^7.27.1"
     "@babel/helper-validator-identifier" "^7.27.1"
 
-"@datadog/libdatadog@^0.6.0":
-  version "0.6.0"
-  resolved "https://registry.yarnpkg.com/@datadog/libdatadog/-/libdatadog-0.6.0.tgz#299a717ca18d1ea643ecf5880095ed864ffa32a8"
-  integrity sha512-Ldu+U59LUnejtd7ceXMKJCAFZeYpNdTEEXr8PISY9HFXMa4DOwepcWMaJAAqCPU7LINJeivVwvSD1Pm14VZy7w==
+"@datadog/libdatadog@git+https://github.com/DataDog/libdatadog-nodejs.git#a7b36c913f4fa48c19d46a488a4e46c153f5d4b4":
```
BridgeAR: This looks like it's a trail run and should probably be changed to the official one? :)

dmehala: Resolved in 3e9761759f245367cae636cd6a8d03a42ed87fe8

packages/dd-trace/src/tracer.js
```diff
@@ -24,6 +25,12 @@ class DatadogTracer extends Tracer {
     this._scope = new Scope()
     setStartupLogConfig(config)
     flushStartupLogs(log)
+
+    try {
+      this._inmem_cfg = storeConfig(config)
+    } catch (error) {
+      log.warn(`Warning: Could not store tracer configuration for service discovery (Reason: ${error})`)
```
watson: Two minor fixes:
- We usually don't prefix warnings logged using `log.warn` with `Warning:`
- If you add the error as a 2nd argument to `log.warn` we don't waste time constructing the string if the log will be silenced by the current log level. The logger will know what to do if an error is passed as the last argument. 

```suggestion
      log.warn('Could not store tracer configuration for service discovery', error)
```

packages/dd-trace/src/tracer.js
```diff
@@ -24,6 +25,12 @@ class DatadogTracer extends Tracer {
     this._scope = new Scope()
     setStartupLogConfig(config)
     flushStartupLogs(log)
+
+    try {
+      this._inmem_cfg = storeConfig(config)
```
watson: What is `this._inmem_cfg` supposed to get used for? I don't see any references to it. It feels like something is missing here?

packages/dd-trace/src/tracer.js
```diff
@@ -24,6 +25,12 @@ class DatadogTracer extends Tracer {
...
+    } catch (error) {
+      log.warn(`Warning: Could not store tracer configuration for service discovery (Reason: ${error})`)
+    }
```
watson: As far as I can see, `new processDiscovery.TracerMetadata()` is not meant to throw if you provide it the required parameters, which you do. In those cases we normally don't surround statements with a try-catch block in JavaScript. So normally I'd write it something like this:

```suggestion
    this._inmem_cfg = storeConfig(config)
    if (this._inmem_cfg === undefined) {
      log.warn('Could not store tracer configuration for service discovery (Can\'t load process-discovery library)')
    }
```

And then update `storeConfig` to just return early in case `libdatadog.maybeLoad('process-discovery')` returns `undefined`

packages/dd-trace/src/tracer.js
BridgeAR: It is used in some C++ code. That's why it is seemingly unused

packages/dd-trace/src/tracer.js
```diff
@@ -9,6 +9,7 @@ const { ERROR_MESSAGE, ERROR_TYPE, ERROR_STACK } = require('../../dd-trace/src/c
 const { DataStreamsCheckpointer, DataStreamsManager, DataStreamsProcessor } = require('./datastreams')
 const { flushStartupLogs } = require('../../datadog-instrumentations/src/helpers/check-require-cache')
 const log = require('./log/writer')
+const storeConfig = require('./tracer_metadata')
```
rochdev: Is this used in serverless environments? If not there should be a conditional to not load it there to avoid the additional startup overhead.

packages/dd-trace/src/tracer_metadata.js
```diff
@@ -0,0 +1,25 @@
+'use strict'
+
+// Load binding first to not import other modules if it throws
+const libdatadog = require('@datadog/libdatadog')
+const tracerVersion = require('../../../version').VERSION
+
+function storeConfig (config) {
+  const processDiscovery = libdatadog.maybeLoad('process-discovery')
+  if (processDiscovery === undefined) {
+    throw new Error('Can\'t load process-discovery library')
```
BridgeAR:
```suggestion
    return
```

That way we do not have to use try / catch at the callsite (as outlined by @watson )

packages/dd-trace/src/tracer.js
```diff
@@ -24,6 +24,17 @@ class DatadogTracer extends Tracer {
...
+
+    if (!config._isInServerlessEnvironment()) {
+      try {
+        const storeConfig = require('./tracer_metadata')
+        // Keep a reference to the handle, to keep the memfd alive in memory.
+        // It is read by the service discovery feature.
+        this._inmem_cfg = storeConfig(config)
+      } catch (error) {
+        log.warn('Could not store tracer configuration for service discovery', error)
+      }
```
BridgeAR: In combination with the comment below:

```suggestion
      const storeConfig = require('./tracer_metadata')
      // Keep a reference to the handle, to keep the memfd alive in memory.
      // It is read by the service discovery feature.
      const metadata = storeConfig(config)
      if (metadata === undefined) {
        log.warn('Could not store tracer configuration for service discovery')
      }
      this._inmem_cfg = metadata
```

BridgeAR (approved):

PR #5792, opened 2025-05-30
Upgrade libddwaf native bindings

CarlesDD:
### What does this PR do?

Introduces changes to use the latest version of native appsec which contains the libddwaf changes up to version 1.24.1. 

This new libddwaf brings a new interface for instance creation as well as instance configuration management, moving the logic of merging configurations received by RC to libddwaf.

Also, diagnostics have been updated and should now be propagated by RC and telemetry. 

### Additional Notes
[ST PR](https://github.com/DataDog/system-tests/pull/4800)
[APPSEC-55565](https://datadoghq.atlassian.net/browse/APPSEC-55565)
RFC-1025

[APPSEC-55565]: https://datadoghq.atlassian.net/browse/APPSEC-55565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dd-trace/src/appsec/reporter.js
```diff
@@ -225,6 +242,53 @@ function reportWafInit (wafVersion, rulesVersion, diagnosticsRules = {}, success
...
+    if (diagnostics[configKey].error) {
+      telemetryLogCh.publish({
+        message: diagnostics[configKey].error,
+        level: 'ERROR',
+        tags: {
+          log_type: `rc::${product.toLowerCase()}::diagnostic`,
+          appsec_config_key: configKey,
+          rc_config_id: rcConfigId
+        }
+      })
```
BridgeAR, line 258: I would abstract this in a small helper with the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))

packages/dd-trace/src/appsec/rule_manager.js
```diff
@@ -26,137 +20,66 @@ function loadRules (config) {
...
-  const newExclusions = new SpyMap(appliedExclusions)
-  const newCustomRules = new SpyMap(appliedCustomRules)
   const newActions = new SpyMap(appliedActions)
 
+  let wafUpdated = false
+  let wafUpdatedSuccess = true
+
   for (const item of toUnapply) {
-    const { product, id } = item
+    if (!['ASM_DD', 'ASM_DATA', 'ASM'].includes(item.product)) continue
```
BridgeAR, line 29: Right now this array is created in each loop iteration. Maybe move it outside of the method and make it a set so that there is only a single lookup happening per iteration? :)

```suggestion
    if (PRODUCTS.has(item.product)) continue
```

The same applies in other places.

packages/dd-trace/src/appsec/rule_manager.js
```diff
@@ -26,137 +20,66 @@ function loadRules (config) {
...
       }
-    } else if (product === 'ASM') {
-      newRulesOverride.delete(id)
-      newExclusions.delete(id)
-      newCustomRules.delete(id)
-      newActions.delete(id)
+    } catch (e) {
+      wafUpdatedSuccess = false
+      item.apply_state = ERROR
+      item.apply_error = e.toString()
```
BridgeAR, line 42: Is it intentional to only care about the error type and message? Calling `.toString()` is pretty bad for Node.js errors such that the stack and other properties are all lost. In case the error should be inspected, I would just use `util.inspect(error)`.

The same for other spots.

packages/dd-trace/src/appsec/rule_manager.js
```diff
@@ -26,137 +20,66 @@ function loadRules (config) {
...
-      }
-
-      if (file?.exclusions?.length) {
-        newExclusions.set(id, file.exclusions)
-      }
-
-      if (file?.custom_rules?.length) {
-        newCustomRules.set(id, file.custom_rules)
+        wafUpdatedSuccess = false
+        item.apply_error = JSON.stringify(extractErrors(updateResult.diagnostics))
```
BridgeAR, line 59: The errors are handled very differently depending on where it happens while the overall logical seems to almost be identical. Should they not be handled all the same? :)

packages/dd-trace/src/appsec/rule_manager.js
```diff
@@ -188,67 +111,28 @@ function concatArrays (files) {
...
-  const copy = { ...rulesData }
-  if (copy.data) {
-    const data = []
-    copy.data.forEach(item => {
-      data.push({ ...item })
-    })
-    copy.data = data
-  }
-  return copy
+  return Object.keys(result).length > 0 ? result : null
```
BridgeAR, line 130: Instead of checking for results keys, I would just use a boolean that I set to true in case the loop happened to match something, since calling `Object.keys()` is quite expensive to just get that information.

In general, is the object really arbitrary that we have to loop over it completely in an unknown way?

BridgeAR (commented):
I can't tell anything about the domain logic in this case, I just had a brief look at the overall code and left a few questions and suggestions :)

packages/dd-trace/src/appsec/rule_manager.js
CarlesDD: Yes, it is. Tracer only reports error type and message in `apply_error` for RC items. 
- https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/remote_config/manager.js#L299
- https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/remote_config/manager.js#L314
- https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/remote_config/manager.js#L323

packages/dd-trace/src/appsec/rule_manager.js
CarlesDD: Suggestion's been applied.

Regarding the arbitrariness of the objetc, no, the object is not really arbitrary, it has a defined schema (at least, for this version of libddwaf).

packages/dd-trace/src/appsec/reporter.js
CarlesDD: New helper has been implemented to publish these diagnostics log messages.

packages/dd-trace/src/appsec/rule_manager.js
CarlesDD: Suggestion has been applied.

packages/dd-trace/src/appsec/waf/waf_manager.js
```diff
@@ -51,19 +54,34 @@ class WAFManager {
...
-      this.ddwaf.update(newRules)
+  update (product, rules, path) {
+    if (product === 'ASM_DD' && this.ddwaf.configPaths.includes(DEFAULT_WAF_CONFIG_PATH)) {
+      this.ddwaf.removeConfig(DEFAULT_WAF_CONFIG_PATH)
+    }
 
-      if (this.ddwaf.diagnostics.ruleset_version) {
-        this.rulesVersion = this.ddwaf.diagnostics.ruleset_version
-      }
+    const success = this.ddwaf.createOrUpdateConfig(rules, path)
```
IlyasShabi: createOrUpdateConfig could throw an error, I think we need a try/catch block

packages/dd-trace/src/appsec/waf/waf_manager.js
```diff
@@ -51,19 +54,34 @@ class WAFManager {
...
-      Reporter.reportWafUpdate(this.ddwafVersion, 'unknown', false)
+    if (diagnostics.ruleset_version) {
+      this.rulesVersion = diagnostics.ruleset_version
+    }
+
+    if (product === 'ASM_DD' && !success && !this.ddwaf.configPaths.some(cp => cp.includes('ASM_DD'))) {
+      this.ddwaf.createOrUpdateConfig(this.defaultRules, DEFAULT_WAF_CONFIG_PATH)
+    }
+
+    return { success, diagnostics }
```
IlyasShabi: I think that we need to update `this.rulesVersion = this.ddwaf.diagnostics.ruleset_version here too`

packages/dd-trace/src/appsec/rule_manager.js
```diff
@@ -26,137 +22,66 @@ function loadRules (config) {
...
-      if (file?.actions?.length) {
-        newActions.set(id, file.actions)
-      }
-
-      batch.add(item)
+    } catch (e) {
+      wafUpdatedSuccess = false
+      item.apply_state = ERROR
+      item.apply_error = e.toString()
+      Reporter.reportWafConfigError(waf.wafManager.ddwafVersion, waf.wafManager.rulesVersion)
```
IlyasShabi: This means that we gonna report probably a wrong rulesVersion if remove fails no?

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -225,6 +242,54 @@ function reportWafInit (wafVersion, rulesVersion, diagnosticsRules = {}, success
   incrementWafInitMetric(wafVersion, rulesVersion, success)
 }
 
+function logWafDiagnosticMessage (product, rcConfigId, configKey, message, level) {
+  telemetryLogCh.publish({
```
uurien, line 233: These logs are sent via telemetry, but aren't shown in the output if the logs are enabled, is the expected behaviour?

CarlesDD: The RFC only establishes the propagation of diagnostics by telemetry and RC. Specifically it states that:

> Note that aside from through telemetry logs, none of the diagnostics presented here should be logged to a regular logger using a level above debug, as it will inevitably result in customer complaints.

packages/dd-trace/src/appsec/rule_manager.js
BridgeAR: Right, so it's just porting this over.
Do we know if the original errors where intentionally defined as such when originally implemented or if that was 'by chance'?

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -225,6 +242,54 @@ function reportWafInit (wafVersion, rulesVersion, diagnosticsRules = {}, success
...
+      for (const [warningMessage, warningIds] of Object.entries(configDiagnostics.warnings)) {
+        logWafDiagnosticMessage(
+          product,
+          rcConfigId,
+          configKey,
+          `"${warningMessage}": ${JSON.stringify(warningIds)}`,
+          'WARN'
+        )
+      }
+    }
```
BridgeAR: Suggestion (I think this is already good as is, so it's up to you): this could be unified with a loop over `['errors', 'warnings']`

packages/dd-trace/src/appsec/reporter.js
BridgeAR: Thanks, the abstraction is easier to read for me and I find it much cleaner :)

packages/dd-trace/src/appsec/rule_manager.js
BridgeAR: Should we just handle the exact structure in that case instead of keeping it arbitrary? I am good both ways, being explicit is just likely faster.

packages/dd-trace/src/appsec/waf/index.js
```diff
@@ -34,18 +33,6 @@ function destroy () {
...
-  if (!waf.wafManager) throw new Error('Cannot update disabled WAF')
-
-  try {
-    waf.wafManager.update(newRules)
-  } catch (err) {
-    log.error('[ASM] Could not apply rules from remote config')
-    throw err
-  }
-}
-
```
simon-id: I'm not sure why you're removing this, the point of this file is to make a abstraction layer for when WAF is enabled or disabled. So basically the rule_manager.js can call update() even when there are no waf currently, this layer handles it.
With your new approach it looks like it you could end up crashing because rule_manager.js is calling waf.wafManager when it could be null no ?

packages/dd-trace/src/appsec/rule_manager.js
simon-id: We have a set list of keys that are supported, so we could iterate on this instead, no sure which one would be faster, but definitely something to do in an optimizing PR instead of here I'd say

packages/dd-trace/src/appsec/waf/waf_manager.js
simon-id: looks like we can close this since it's been added no ?

packages/dd-trace/src/appsec/rule_manager.js
simon-id: maybe it should take the version from the diagnostics instead ? in this case we need to sync with other libs, it's not an implementation detail, it's an architecture concern

packages/dd-trace/src/appsec/reporter.js
simon-id: can be closed then ?

integration-tests/graphql.spec.js
```diff
@@ -112,7 +112,7 @@ describe('graphql', () => {
       assert.property(payload[1][0].metrics, '_dd.appsec.waf.duration')
       assert.propertyVal(payload[1][0].meta, 'appsec.event', 'true')
       assert.property(payload[1][0].meta, '_dd.appsec.json')
-      assert.propertyVal(payload[1][0].meta, '_dd.appsec.json', JSON.stringify(result))
+      assert.deepStrictEqual(JSON.parse(payload[1][0].meta['_dd.appsec.json']), result)
```
simon-id: why is this needed ? string comparison doesn't work anymore ? i'm guessing the key orders changed or something ?

packages/dd-trace/src/appsec/waf/waf_manager.js
simon-id, line 1: I feel like you're adding a lot of code in this file that doesn't belong at this level of abstraction.
Why would the waf_manager have in it so many RC concepts, like config_paths, and product names. I don't think this is the right place to put all this logic. We have a rule_manager, which has all the RC and rule logic contained in there, and it's for a good reason.
In other words i think you're transgressing the layers of abstraction that we set out when we last rewrote this a couple years back. If you think doing it differently than you did would ultimately end up being worse overall then whatever keep it like this i guess

packages/dd-trace/src/appsec/waf/waf_manager.js
```diff
@@ -51,19 +54,34 @@ class WAFManager {
...
+    const diagnostics = this.ddwaf.diagnostics
 
-      Reporter.reportWafUpdate(this.ddwafVersion, this.rulesVersion, true)
-    } catch (error) {
-      Reporter.reportWafUpdate(this.ddwafVersion, 'unknown', false)
+    if (diagnostics.ruleset_version) {
+      this.rulesVersion = diagnostics.ruleset_version
+    }
+
+    if (product === 'ASM_DD' && !success && !this.ddwaf.configPaths.some(cp => cp.includes('ASM_DD'))) {
```
simon-id: so currently you're doing "remove default rule, update, and if fail add back default rule" right ?
why not just do "add new rules, if success remove old ones" ?

packages/dd-trace/src/appsec/waf/waf_manager.js
```diff
@@ -51,19 +54,34 @@ class WAFManager {
...
-      }
+    const success = this.ddwaf.createOrUpdateConfig(rules, path)
+    const diagnostics = this.ddwaf.diagnostics
 
-      Reporter.reportWafUpdate(this.ddwafVersion, this.rulesVersion, true)
-    } catch (error) {
-      Reporter.reportWafUpdate(this.ddwafVersion, 'unknown', false)
+    if (diagnostics.ruleset_version) {
+      this.rulesVersion = diagnostics.ruleset_version
+    }
```
simon-id: we should probably put this at the very end of the function

packages/dd-trace/src/appsec/rule_manager.js
```diff
@@ -26,137 +22,66 @@ function loadRules (config) {
...
-      newActions.delete(id)
+    } catch (e) {
+      wafUpdatedSuccess = false
+      item.apply_state = ERROR
+      item.apply_error = e.toString()
+      Reporter.reportWafConfigError(waf.wafManager.ddwafVersion, waf.wafManager.rulesVersion)
     }
   }
 
   for (const item of [...toApply, ...toModify]) {
```
simon-id, line 55: what happens with the new waf if you try to add the same config path twice ?
this would happen when a config item is in toModify. The path would be the same, but the payload changed. Does the WAF accept that or would it require us to delete+readd it ?

packages/dd-trace/src/appsec/rule_manager.js
```diff
@@ -26,137 +22,66 @@ function loadRules (config) {
...
-  const newRulesData = new SpyMap(appliedRulesData)
-  let newRuleset
-  let newRulesetId
-  const newRulesOverride = new SpyMap(appliedRulesOverride)
-  const newExclusions = new SpyMap(appliedExclusions)
-  const newCustomRules = new SpyMap(appliedCustomRules)
   const newActions = new SpyMap(appliedActions)
 
+  let wafUpdated = false
+  let wafUpdatedSuccess = true
```
simon-id: weird default to have, i'd like it better to have it changed to false whenever it works

packages/dd-trace/src/appsec/rule_manager.js
```diff
@@ -26,137 +22,66 @@ function loadRules (config) {
...
+      wafUpdated = true
 
-    if (product === 'ASM_DATA') {
-      newRulesData.delete(id)
-    } else if (product === 'ASM_DD') {
-      if (appliedRulesetId === id) {
-        newRuleset = defaultRules
+      if (item.product === 'ASM') {
+        newActions.delete(item.id)
       }
```
simon-id, line 47: isn't there a way to get the actions from the WAF instead of doing the state managemnt ourselves ?

packages/dd-trace/src/appsec/rule_manager.js
```diff
@@ -188,67 +113,31 @@ function concatArrays (files) {
...
-  const mergedRulesData = new Map()
-  for (const [, file] of files) {
-    for (const ruleData of file) {
-      const key = `${ruleData.id}+${ruleData.type}`
-      if (mergedRulesData.has(key)) {
-        const existingRulesData = mergedRulesData.get(key)
-        ruleData.data.reduce(rulesReducer, existingRulesData.data)
-      } else {
-        mergedRulesData.set(key, copyRulesData(ruleData))
+function extractErrors (obj) {
```
simon-id: worth a comment explaining what HOW it extract errors. I'm sure this is far from obvious for anyone that doesn't know exactly what the WAF output data looks like, even for me it's weird

integration-tests/graphql.spec.js
CarlesDD: You guessed it right. Since we are not testing the order of the keys, I think this way of comparing both values would work better.

packages/dd-trace/src/appsec/reporter.js
uurien: Yes, it could be closed. I don't usually close this type of questions because they could be useful for next reviewers to prevent asking the same.

simon-id: Ah alright up to you, I'm used to also look at the closed comments anyway :+1:

packages/dd-trace/src/appsec/waf/waf_manager.js
CarlesDD: It is not possible with the current WAF.

packages/dd-trace/src/appsec/rule_manager.js
simon-id: i think we concluded this was required because of the SDK :+1:

packages/dd-trace/src/appsec/rule_manager.js
CarlesDD: After the refactor, the method to extract the errors to be reported in RC no longer deals with an arbitrary object but will explicitly look for the the relevant keys of the object.

packages/dd-trace/src/appsec/waf/waf_manager.js
CarlesDD: That's right. Now, all the calls to this method are enclosed in a try/catch block in waf/index.

packages/dd-trace/src/appsec/waf/index.js
CarlesDD: I've refactored the code to keep this logic in waf index, so this layer nows handles (as it previously did), the fact that waf manager could be null.

packages/dd-trace/src/appsec/waf/waf_manager.js
CarlesDD: The last changes refactoring the code have tried to separate and isolate both logics in the different files (rule manager, waf manager).

packages/dd-trace/src/appsec/waf/waf_manager.js
CarlesDD: Done

packages/dd-trace/src/appsec/rule_manager.js
CarlesDD: The new WAF is able to handle this kind of situations - same config path, different payload.

packages/dd-trace/src/appsec/rule_manager.js
CarlesDD: Changed :)

packages/dd-trace/src/appsec/rule_manager.js
CarlesDD: With the new implementation the logic is so much clear - it is not dealing with arbitrary objects anymore, but it looks for the relevant keys in the object to extract the errors. I think now there is no need to explain how it works. Let me know if you think so.

packages/dd-trace/src/appsec/reporter.js
CarlesDD: I prefer the way it is currently

packages/dd-trace/src/appsec/rule_manager.js
CarlesDD: The logic to report Waf config errors has been refactored, and now the reported version is taken from the diagnostics, aligning with all the tracers.

packages/dd-trace/src/appsec/rule_manager.js
CarlesDD: The logic for error handling has been modified in the last commits, splitting the propagation of the error through RC and telemetry.

packages/dd-trace/src/appsec/rule_manager.js
CarlesDD: I'll let @simon-id answer this :)

simon-id: Ye I coded it so back then so it only gets the type and message, but honestly we've never had to dig in the backend to use these errors, i don't think they're even showed to the customers. One day if this become important we can include the stack and all. For now I'd say we can keep it like this

packages/dd-trace/src/appsec/waf/waf_manager.js
```diff
@@ -7,11 +7,14 @@ const WAFContextWrapper = require('./waf_context_wrapper')
 const contexts = new WeakMap()
 
 class WAFManager {
+  static get defaultWafConfigPath () { return 'datadog/00/ASM_DD/default/config' }
```
simon-id, line 10: static fields are available since node 12.0.0 ;)
```suggestion
  static defaultWafConfigPath = 'datadog/00/ASM_DD/default/config'
```

packages/dd-trace/src/appsec/rule_manager.js
```diff
@@ -26,137 +36,63 @@ function loadRules (config) {
...
+      if (item.product === 'ASM') {
+        newActions.delete(item.id)
       }
-    } else if (product === 'ASM') {
-      newRulesOverride.delete(id)
-      newExclusions.delete(id)
-      newCustomRules.delete(id)
-      newActions.delete(id)
+    } catch (e) {
+      item.apply_state = ERROR
```
simon-id, line 49: removeConfig doesn't provide diagnostics and need extractError() and stuff ?

packages/dd-trace/src/appsec/waf/waf_manager.js
```diff
@@ -51,20 +54,27 @@ class WAFManager {
...
+    this.setRulesVersion()
+    return updateResult
+  }
 
-      throw error
-    }
+  removeConfig (path) {
+    this.ddwaf.removeConfig(path)
+    this.setRulesVersion()
   }
```
simon-id, line 78: removeConfig also returns a boolean, why not handle it just like updateConfig() ?

packages/dd-trace/src/appsec/waf/waf_manager.js
```diff
@@ -51,20 +54,27 @@ class WAFManager {
...
+  }
 
-      Reporter.reportWafUpdate(this.ddwafVersion, this.rulesVersion, true)
-    } catch (error) {
-      Reporter.reportWafUpdate(this.ddwafVersion, 'unknown', false)
+  updateConfig (path, rules) {
+    const updateResult = this.ddwaf.createOrUpdateConfig(rules, path)
+    this.setRulesVersion()
+    return updateResult
+  }
```
simon-id, line 73: you know instead of returning updateResult and doing a weird WafUpdateError system in the layer above, you could just return null if success, and the diagnostics if not success ? hence you can remove all the error type checking in the catch ?

simon-id: something like this ? and in waf/index.js you can remove all the wafUpdateError concept
```suggestion
  updateConfig (path, rules) {
    const updateResult = this.ddwaf.createOrUpdateConfig(rules, path)
    this.setRulesVersion()
    return updateResult ? null : this.ddwaf.diagnostics
  }
```

packages/dd-trace/src/appsec/waf/index.js
```diff
@@ -34,14 +46,43 @@ function destroy () {
...
+  }
+}
+
+function updateConfig (product, configId, configPath, config) {
+  if (!waf.wafManager) throw new Error('Cannot update disabled WAF')
+
+  try {
+    if (product === 'ASM_DD') {
+      waf.wafManager.removeConfig(waf.wafManager.constructor.defaultWafConfigPath)
+    }
```
simon-id, line 65: if it was me this would be in rule_manager.js, just like it was before this PR, and to do so, you'd need to extract defaultWafConfigPath also in rule_manager, which would pass the default path to the waf constructor etc
but this PR has been going on for too long, so i won't block it

packages/dd-trace/src/appsec/waf/index.js
```diff
@@ -34,14 +46,43 @@ function destroy () {
...
+
+    const updateSucceed = waf.wafManager.updateConfig(configPath, config)
```
simon-id:
```suggestion
    const updateSucceeded = waf.wafManager.updateConfig(configPath, config)
```

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -216,17 +233,64 @@ function getCollectedHeaders (req, res, shouldCollectEventHeaders, storedRespons
...
-      metricsQueue.set('_dd.appsec.event_rules.errors', JSON.stringify(diagnosticsRules.errors))
-    }
   }
 
   incrementWafInitMetric(wafVersion, rulesVersion, success)
 }
 
+function logWafDiagnosticMessage (product, rcConfigId, configKey, message, level) {
+  const tags =
+    `log_type:rc::${product.toLowerCase()}::diagnostic,appsec_config_key:${configKey},rc_config_id:${rcConfigId}`
```
simon-id, line 232: nit: to make it fit in one line maybe ?

```suggestion
  product = product.toLowerCase()
  const tags = `log_type:rc::${product}::diagnostic,appsec_config_key:${confKey},rc_config_id:${confId}`
```

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -216,17 +233,64 @@ function getCollectedHeaders (req, res, shouldCollectEventHeaders, storedRespons
...
+    message,
+    level,
+    tags
+  })
+}
+
+function reportWafConfigUpdate (product, rcConfigId, diagnostics, wafVersion) {
+  if (diagnostics.error) {
+    logWafDiagnosticMessage(product, rcConfigId, '', diagnostics.error, 'ERROR')
+    incrementWafConfigErrorsMetric(wafVersion, diagnostics.ruleset_version || 'unknown')
```
simon-id: the 'unknown' default value should be done in incrementWafConfigErrorsMetric() directly since you do it 3 times no ?

packages/dd-trace/src/appsec/rule_manager.js
```diff
@@ -3,20 +3,30 @@
...
+const DIAGNOSTIC_KEYS = [
+  'exclusions',
+  'rules',
+  'processors',
+  'rules_override',
+  'rules_data',
+  'custom_rules',
+  'actions',
+  'scanners'
+]
```
simon-id: aren't you missing exclusion_data here ? this map is duplicated in reporter.js, make it one single source of truth instead no ?

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -216,17 +233,64 @@ function getCollectedHeaders (req, res, shouldCollectEventHeaders, storedRespons
 function reportWafInit (wafVersion, rulesVersion, diagnosticsRules = {}, success = false) {
   if (success) {
     metricsQueue.set('_dd.appsec.waf.version', wafVersion)
-
-    metricsQueue.set('_dd.appsec.event_rules.loaded', diagnosticsRules.loaded?.length || 0)
-    metricsQueue.set('_dd.appsec.event_rules.error_count', diagnosticsRules.failed?.length || 0)
-    if (diagnosticsRules.failed?.length) {
-      metricsQueue.set('_dd.appsec.event_rules.errors', JSON.stringify(diagnosticsRules.errors))
-    }
```
simon-id, line 224: are those tags deprecated ?

packages/dd-trace/src/appsec/rule_manager.js
CarlesDD: No, it does not.

packages/dd-trace/src/appsec/waf/waf_manager.js
CarlesDD: `removeConfig` returns true if configuration was removed successfully, and false if it does not exist. There is no diagnostics to extract errors or warnings from, so there is nothing to report.

packages/dd-trace/src/appsec/reporter.js
CarlesDD: Yep, they are

packages/dd-trace/src/appsec/waf/waf_manager.js
CarlesDD: Since webpack v4 does not handle static props correctly I would avoid using them for now until we have a decision after discussing it in the guild.

packages/dd-trace/src/appsec/waf/waf_manager.js
CarlesDD: Yes, this was one of the approaches I played with. However, throwing a WafUpdateError on failed update result allows me to centralize all the error handling for RC (apply_error) in one place.

packages/dd-trace/src/appsec/rule_manager.js
CarlesDD: I've extracted the keys to a separate file to have one single source of truth. 
Yes, exclusion data was missing. Keep in mind that this is a live thing and it could change in future versions of libddwaf, so it should be updated.

simon-id:
ok LGTM logic wise, some nits left with the static field and the oneliner string but not blocking. The CI probably needs a good rebase since fastify is failing too

packages/dd-trace/src/appsec/rule_manager.js
BridgeAR: Errors are actually very expensive, so if the stack is not needed, I wonder if we should even use errors in that case.

simon-id: these errors should never happen anyway, this is a cold path, it executes every 5 seconds, and the chance of having an exception is low. Is it really worth to make that optimization in this PR since it's alreayd ready to merge ? I'd say meh. Carles is off for the week anyway. Something to keep in mind for a future big optimization PR I'd say

IlyasShabi (approved):
LGTM

PR #5797, opened 2025-05-30
feat(openai): support openai v5

sabrenner:
### What does this PR do?
With the release of OpenAI v5, the only thing we needed to account for was some `del` methods changing their name to `delete`. Otherwise, our instrumentation is largely unchanged.

This PR mostly fixes tests. There were a few things impacted:
1. We had to upgrade our `nock` dev dependency since OpenAI switched to using `undici.Fetch` under the hood instead of vanilla HTTP.
2. The `beta` feature we supported is only for OpenAI v4 and does not need to be tested for v5 (as it does not exist anymore)
3. There was a small `File` error, I think just on my machine which runs Node 18 by default - it likely won't impact CI as I think we run on latest but I wanted to handle it regardless.

### Motivation
Support OpenAI v5 and unblock broken CI

MLOB-2927


packages/datadog-plugin-openai/test/index.spec.js
```diff
@@ -1180,9 +1189,13 @@ describe('Plugin', () => {
...
             })
 
-          if (semver.satisfies(realVersion, '>=4.0.0')) {
+          if (semver.satisfies(realVersion, '>=4.0.0 <5.0.0')) {
             const result = await openai.files.del('file-268aYWYhvxWwHb4nIzP9FHM6')
 
+            expect(result.deleted).to.eql(true)
+          } else if (semver.satisfies(realVersion, '>=5.0.0')) {
+            const result = await openai.files.delete('file-268aYWYhvxWwHb4nIzP9FHM6')
+
```
BridgeAR, line 1197:
```suggestion
          if (semver.satisfies(realVersion, '>=4.0.0')) {
            const method = semver.satisfies(realVersion, '>=5.0.0') ? 'delete' : 'del'
            const result = await openai.files[method]('file-268aYWYhvxWwHb4nIzP9FHM6')

```

sabrenner: applied it here and in the other spot i changed 1e03495cc0cd06a9bd849f2de823d2c7b3a23074

PR #5800, opened 2025-05-31
Update eslint rules and activate multiple more

BridgeAR:


packages/dd-trace/src/debugger/devtools_client/breakpoints.js
```diff
@@ -194,7 +194,6 @@ async function updateBreakpointInternal (breakpoint, probe) {
     } catch (err) {
       throw new Error(`Error setting breakpoint for probe ${probe.id}`, { cause: err })
     }
-    breakpoint.id = result.breakpointId
```
BridgeAR, line 197: This likely fixes a race condition.

packages/datadog-core/src/utils/src/kebabcase.js
```diff
@@ -7,8 +7,8 @@ module.exports = function kebabcase (str) {
 
   return str
     .trim()
-    .replace(/([a-z])([A-Z])/g, '$1-$2') // Convert camelCase to kebab-case
-    .replace(/[\s_]+/g, '-') // Replace spaces and underscores with a single dash
-    .replace(/^-+|-+$/g, '') // Trim leading and trailing dashes
+    .replaceAll(/([a-z])([A-Z])/g, '$1-$2') // Convert camelCase to kebab-case
+    .replaceAll(/[\s_]+/g, '-') // Replace spaces and underscores with a single dash
+    .replaceAll(/^-+|-+$/g, '') // Trim leading and trailing dashes
```
watson, line 12: This doesn't change anything, as the regexes already uses `g`. But I assume the linter just requires this?

packages/datadog-instrumentations/src/hapi.js
```diff
@@ -75,17 +75,14 @@ function wrapRebuild (rebuild) {
...
+  if (!Array.isArray(events)) {
+    events = [events]
+  }
+  return events.map(event => {
+    if (!event?.method) return event
 
-    return Object.assign({}, event, {
-      method: wrapExtension(event.method)
-    })
+    return { ...event, method: wrapHandler(event.method) }
```
watson, line 87: Are we sure `event.method` isn't an array? `wrapExtension` would have handled that situation before, whereas calling `wrapHandler` directly doesn't

packages/datadog-instrumentations/src/restify.js
```diff
@@ -70,6 +66,9 @@ function wrapFn (fn) {
       finishChannel.publish({ req })
       throw error
     } finally {
+      // TODO(BridgeAR): This might not be correct in case promises are used.
+      // It should be handled there and not trigger here anymore in case the
+      // method returns a promise.
```
watson: This sentence doesn't make sense 😅

packages/datadog-plugin-elasticsearch/src/index.js
```diff
@@ -35,7 +35,7 @@ function getBody (body) {
 }
 
 function quantizePath (path) {
-  return path && path.replace(/[0-9]+/g, '?')
+  return path && path.replaceAll(/[0-9]+/g, '?')
```
watson, line 38: This doesn't change anything, as the regex already uses `g`. But I assume the linter just requires this?

packages/datadog-plugin-google-cloud-vertexai/src/tracing.js
```diff
@@ -82,7 +82,7 @@ class GoogleCloudVertexAITracingPlugin extends TracingPlugin {
 
     const generationConfig = instance.generationConfig || {}
     for (const key of Object.keys(generationConfig)) {
-      const transformedKey = key.replace(/([a-z0-9])([A-Z])/g, '$1_$2').toLowerCase()
+      const transformedKey = key.replaceAll(/([a-z0-9])([A-Z])/g, '$1_$2').toLowerCase()
```
watson, line 85: This doesn't change anything, as the regex already uses `g`. But I assume the linter just requires this?

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -96,7 +96,7 @@ function formatHeaderName (name) {
   return name
     .trim()
     .slice(0, 200)
-    .replace(/[^a-zA-Z0-9_\-:/]/g, '_')
+    .replaceAll(/[^a-zA-Z0-9_\-:/]/g, '_')
```
watson, line 99: This doesn't change anything, as the regex already uses `g`. But I assume the linter just requires this?

packages/dd-trace/src/ci-visibility/exporters/git/git_metadata.js
```diff
@@ -38,7 +38,7 @@ function validateCommits (commits) {
       throw new Error('Invalid commit type response')
     }
     if (isValidSha1(commitSha) || isValidSha256(commitSha)) {
-      return commitSha.replace(/[^0-9a-f]+/g, '')
+      return commitSha.replaceAll(/[^0-9a-f]+/g, '')
```
watson, line 41: This doesn't change anything, as the regex already uses `g`. But I assume the linter just requires this?

packages/dd-trace/src/config.js
```diff
@@ -1533,12 +1533,12 @@ function handleOtel (tagString) {
...
     .replace(/(^|,)service\.name=/, '$1service:')
     .replace(/(^|,)service\.version=/, '$1version:')
-    .replace(/=/g, ':')
+    .replaceAll('=', ':')
 }
 
 function parseSpaceSeparatedTags (tagString) {
   if (tagString && !tagString.includes(',')) {
-    tagString = tagString.replace(/\s+/g, ',')
+    tagString = tagString.replaceAll(/\s+/g, ',')
```
watson, line 1562: This doesn't change anything, as the regex already uses `g`. But I assume the linter just requires this?

packages/dd-trace/src/datastreams/schemas/schema_builder.js
```diff
@@ -117,12 +117,12 @@ function convertKey (key) {
 function jsonStringify (obj, indent = 2) {
   // made to stringify json exactly similar to python / java in order for hashing to be the same
   const jsonString = JSON.stringify(obj, (_, value) => value, indent)
-  return jsonString.replace(/^ +/gm, ' ') // Replace leading spaces with single space
-    .replace(/\n/g, '') // Remove newlines
-    .replace(/{ /g, '{') // Remove space after '{'
-    .replace(/ }/g, '}') // Remove space before '}'
-    .replace(/\[ /g, '[') // Remove space after '['
-    .replace(/ \]/g, ']') // Remove space before ']'
+  return jsonString.replaceAll(/^ +/gm, ' ') // Replace leading spaces with single space
```
watson, line 120: This doesn't change anything, as the regex already uses `g`. But I assume the linter just requires this?

packages/dd-trace/src/debugger/devtools_client/snapshot/redaction.js
```diff
@@ -103,7 +103,7 @@ const REDACTED_IDENTIFIERS = new Set(
 
 function normalizeName (name, isSymbol) {
   if (isSymbol) name = name.slice(7, -1) // Remove `Symbol(` and `)`
-  return name.toLowerCase().replace(/[-_@$.]/g, '')
+  return name.toLowerCase().replaceAll(/[-_@$.]/g, '')
```
watson, line 106: This doesn't change anything, as the regex already uses `g`. But I assume the linter just requires this?

packages/dd-trace/src/dogstatsd.js
```diff
@@ -169,7 +169,7 @@ class DogStatsDClient {
         })
         .forEach(key => {
           // https://docs.datadoghq.com/tagging/#defining-tags
-          const value = config.tags[key].replace(/[^a-z0-9_:./-]/ig, '_')
+          const value = config.tags[key].replaceAll(/[^a-z0-9_:./-]/ig, '_')
```
watson, line 172: This doesn't change anything, as the regex already uses `g`. But I assume the linter just requires this?

packages/dd-trace/src/exporters/common/docker.js
```diff
@@ -24,7 +24,7 @@ try {
 
 const inodePath = cgroup.match(lineReg)?.[3]
 if (inodePath) {
-  const strippedPath = inodePath.replace(/^\/|\/$/g, '')
+  const strippedPath = inodePath.replaceAll(/^\/|\/$/g, '')
```
watson, line 27: This doesn't change anything, as the regex already uses `g`. But I assume the linter just requires this?

packages/dd-trace/test/profiling/profiler.spec.js
```diff
@@ -205,6 +206,7 @@ describe('profiler', function () {
       clock.tick(interval)
 
       await rejected.catch(() => {})
+      await setTimeout(1)
```
watson, line 208: Why this?

packages/dd-trace/src/payload-tagging/config/index.js
```diff
@@ -2,29 +2,25 @@ const aws = require('./aws.json')
...
-      ]
-    })
-  )
+function appendRules (requestInput = [], responseInput = []) {
+  const sdkRules = {}
+  for (const [name, sdk] of Object.entries(sdks)) {
+    sdkRules[name] = getSDKRules(sdk, requestInput, responseInput)
+  }
+  return sdkRules
 }
```
watson, line 24: Same as above

packages/dd-trace/src/payload-tagging/config/index.js
```diff
@@ -2,29 +2,25 @@ const aws = require('./aws.json')
...
-          request: serviceRules.request.concat(requestInput || []),
-          response: serviceRules.response.concat(responseInput || []),
-          expand: serviceRules.expand || []
-        }
-      ]
-    })
-  )
+  const sdkServiceRules = {}
+  for (const [service, serviceRules] of Object.entries(sdk)) {
+    sdkServiceRules[service] = {
```
watson, line 7: Could there be a performance penalty for all the hidden classes being created for every `sdkServiceRules[service] = ...` operation that `Object.fromEntries` is able to circumvent?

packages/dd-trace/src/profiling/tagger.js
```diff
@@ -5,20 +5,30 @@ const tagger = {
...
-            const value = parts.join(':').trim()
+      case 'object': {
+        if (Array.isArray(tags)) {
+          const tagObject = {}
+          for (const tag of tags) {
+            const colon = tag.indexOf(':')
+            if (colon === -1) continue
+            const key = tag.slice(0, colon).trim()
+            const value = tag.slice(colon + 1).trim()
+            if (key.length > 0 && value.length > 0) {
```
watson: Nit:

```suggestion
            if (key.length !== 0 && value.length !== 0) {
```

packages/dd-trace/src/telemetry/dependencies.js
```diff
@@ -8,70 +8,74 @@ const dc = require('dc-polyfill')
...
-    return {
-      request_type: item.reqType,
-      payload: item.payload
-    }
-  })
-
-  return batchPayload
-}
 function waitAndSend (config, application, host) {
-  if (!immediate) {
```
watson, line 34: Are you sure removing the check to see if an existing `setImmediate` is queued is a good idea?

packages/dd-trace/src/telemetry/dependencies.js
```diff
@@ -8,70 +8,74 @@ const dc = require('dc-polyfill')
...
+      // If a dependency is from the initial load, *always* send the event
+      // Otherwise, only send if dependencyCollection is enabled
+      const sendModule = isTrue(initialLoadModule) || config.telemetry?.dependencyCollection
+
+      savedDependenciesToSend.delete(dependency)
+
+      if (sendModule) {
+        dependencies.push({ name, version })
+        send++
+        if (send === 2000) {
```
watson, line 42: Nit (I personally like using `++` like this, but I know many don't)

```suggestion
        if (++send === 2000) {
```

watson (commented):
I starting commenting on changes to string `replace` calls where a regular expression was used with the `g` flag, but there turned out to so many, that I stopped half way through the review. But the initial comments apply to all

packages/datadog-core/src/utils/src/kebabcase.js
BridgeAR: Correct. It's for differentiating replacing all vs individual entries faster. I hope it's fine to resolve all similar comments :)

packages/dd-trace/test/profiling/profiler.spec.js
BridgeAR: That is something I actually wanted to follow up on. The mocks in our test have expectations that are not met due to adding one more tick until the test resolves. I am going ahead and rewrite the mock to accommodate that.

If I am not mistaken, the tests passed by chance so far and it was depending on the right ordering. Effectively this would already have had another tick.
What I changed it to use a `await Promise.all()` on top of the current await in the loop. That way it's an added tick while effectively improving the performance for cases where we want to resolve more than a single entry.

packages/dd-trace/src/payload-tagging/config/index.js
BridgeAR: Even if hidden classes are created and replaced during the loop iteration (which I am not absolutely certain about if that happens right now. I would worry more about the object being in dictionary vs non-dictionary mode) it's almost certainly way less CPU cycles with the new code.

packages/dd-trace/src/payload-tagging/config/index.js
BridgeAR: Responded in the other spot :)

packages/dd-trace/src/telemetry/dependencies.js
BridgeAR: I do feel like less people can read this well. Due to that I always favor ways that I believe are easy to read, no matter of my personal taste (I would also use the shorter notation otherwise).

packages/dd-trace/src/telemetry/dependencies.js
BridgeAR: It should not make any difference with the overall logic.

If the method is called more than once at a time (which it would only in case start is called in the same event loop twice or during an event loop afterwards that is triggered from the running recursive method), it would still just read the next n dependencies. If an extra call is now executed, it would immediately stop due to the savedDependenciesToSend.size is zero or we just continue with this one and it would be the next that stops at that point.

packages/datadog-instrumentations/src/hapi.js
BridgeAR: I tried to log the output and nothing came up in the tests and the plugin seems to work as before.
I could call `.flat()` in line 80 to be on the super safe side, while I am confident about this particular change.
There were a couple of cases where we did indeed have stacks of arrays though in other spots.
In general, using `.concat()` in the way we did is making it very hard to know what data we effectively handle. It was therefore a priority for me to make sure that's improved.

All that aside: I am fine to add the .flat call as outlined.

packages/datadog-instrumentations/src/restify.js
BridgeAR: Definitely not well phrased by me :D 
I'll check about this again and try to figure out if my thought is correct or not and remove the comment accordingly before landing.

packages/dd-trace/src/payload-tagging/config/index.js
BridgeAR: I just ran a very minimal benchmark and the new code it twice as fast as before.

packages/dd-trace/test/profiling/profiler.spec.js
BridgeAR: I had another look into it and this is actually not that bad. I could make these wait a single tick instead while this test is in general very brittle. Things are so strongly mocked that we do not really test real functionality anywhere. We just check that methods are called exactly after a specified number of ticks. That is an expectation that we should normally not have around tests, since it requires deep knowledge about the implementation details.
I would say let's keep the tests as is due to that.

packages/datadog-instrumentations/src/hapi.js
watson: Yeah. My worry is that since we're not testing all versions of the modules being instrumented, there's some version - or even some way to configure this - where it might be an array. So I suggest not relying on our current tests to cover this.

packages/dd-trace/src/telemetry/dependencies.js
watson: Yeah I know :( I really wish it would be common knowledge. Maybe my brain is just wired diffferent

packages/datadog-instrumentations/src/hapi.js
BridgeAR: I just updated it to work as before. Refactoring this is actually really difficult due to not knowing the details in all cases.

watson (dismissed):

packages/datadog-instrumentations/src/cookie-parser.js
```diff
@@ -10,7 +10,7 @@ function publishRequestCookieAndNext (req, res, next) {
     if (cookieParserReadCh.hasSubscribers && req) {
       const abortController = new AbortController()
 
-      const mergedCookies = Object.assign({}, req.cookies, req.signedCookies)
+      const mergedCookies = { ...req.cookies, ...req.signedCookies }
```
simon-id, line 13: is that equivalent performance wise ?

packages/dd-trace/src/appsec/index.js
```diff
@@ -299,12 +299,12 @@ function onResponseWriteHead ({ req, res, abortController, statusCode, responseH
...
   if (!rootSpan) return
 
-  responseHeaders = Object.assign({}, responseHeaders)
+  responseHeaders = { ...responseHeaders }
   delete responseHeaders['set-cookie']
 
   const results = waf.run({
     persistent: {
-      [addresses.HTTP_INCOMING_RESPONSE_CODE]: '' + statusCode,
+      [addresses.HTTP_INCOMING_RESPONSE_CODE]: String(statusCode),
```
simon-id, line 307: is that strictly equivalent ?

packages/dd-trace/src/config.js
```diff
@@ -178,7 +178,7 @@ function validateNamingVersion (versionString) {
  * @param {string | string[]} input
  */
 function splitJSONPathRules (input) {
-  if (!input) return null
```
simon-id, line 181: does the new lint rule forbid returning null ? i think it's a legit use case.
null !== undefined, technically and semantically

simon-id: (not talking about this line specifically but i think it makes sense to use return null in other cases)

packages/dd-trace/src/appsec/index.js
simon-id: it's an important question because this is used in public facing API, thus any change of behavior would be breaking customers

eslint.config.mjs
```diff
@@ -109,6 +120,18 @@ export default [
         }
       ]],
 
+      'no-await-in-loop': 'error',
+      'no-else-return': ['error', { allowElseIf: true }],
```
simon-id, line 124: you know i can understand that this else return case is equivalent withotu the else, but what's the problem with doing it anyway ? is there any performance downside ? because sometimes i use it explicitly as a semantic form, to organize the code in a more intuitive way.
This rule sounds like it's not fixing actual pain ?

simon-id: and why would else if be allowed but else wouldn't ? what's the reason

packages/datadog-instrumentations/src/cookie-parser.js
BridgeAR: It's faster

packages/dd-trace/src/appsec/index.js
BridgeAR: It is safer (in case .toString would be manipulated), otherwise it's identical

packages/dd-trace/src/config.js
BridgeAR: No, it does not forbid that. I changed it to be able to use a default argument in the spot that receives the values from here

eslint.config.mjs
BridgeAR: It's for better readability. Less indentation improves understanding of code. Else cases also increase the complexity for readers. The allowElseIf is just active to minimize the changes right now. I plan on changing that later.

packages/datadog-instrumentations/src/cookie-parser.js
simon-id: oh ok, i always thought it was the opposite :+1:

watson: Like with every new language feature, these things tend to shift over time as the JS engine matures

PR #5801, opened 2025-05-31
Make get-port wait between retries in tests

BridgeAR:
Just a small PR to hopefully make our tests less flaky.


rochdev:
@BridgeAR We tried several times to fix usage of `get-port` and it's always flaky. The only thing that worked was to remove it entirely and get the port automatically assigned by the kernel. I would recommend doing the same here (and I don't know why get-port was brought back to begin with given all the issues with it)

PR #5803, opened 2025-06-02
[test optimization] Improve `vitest` performance

juan-fernandez:
### What does this PR do?

* Instead of each of the workers in `vitest` reporting to datadog themselves, we make them report to the main process, which is in charge of doing the reporting.
* Additionally, we will skip `git` operations on `vitest` workers, as they're expensive and provide no value, as git tags do not modify the execution of the tests and all the information is in the main process anyway. 
  * ℹ️ All frameworks using workers could benefit from this but `vitest` is the one that benefits the most, as it _constantly_ spawns new processes to run tests. Other frameworks usually spawn a fixed number of workers. 
  * We'll tackle other frameworks in follow up PRs

### Motivation
Make `vitest` faster

### Plugin Checklist

- [x] Unit tests.


packages/datadog-plugin-jest/src/index.js
```diff
@@ -288,8 +288,8 @@ class JestPlugin extends CiPlugin {
     })
 
     this.addSub('ci:jest:worker-report:logs', (logsPayloads) => {
-      JSON.parse(logsPayloads).forEach(({ testConfiguration, logMessage }) => {
-        this.tracer._exporter.exportDiLogs(testConfiguration, logMessage)
+      JSON.parse(logsPayloads).forEach(({ logMessage }) => {
+        this.tracer._exporter.exportDiLogs(this.testEnvironmentMetadata, logMessage)
```
juan-fernandez, line 292: unrelated improvement: we don't need the `testConfiguration` provided by the worker as it's all in the main process

anmarchenko (approved):

anmarchenko (approved):

PR #5805, opened 2025-06-02
Improve agent.assertFirstTraceSpan to accept an object

BridgeAR:
The object is internally matched against a partial match.
The output is actually a significant improvement over our current output. It shows the hole trace and also other traces.

I implemented it for a few files while I believe we should switch to this pattern is most of our test code.

There is still significant room for further improvement while this is likely a major step forward in understanding a failure fast and improving the readability of the test cases.

### DIFF
--- a/packages/dd-trace/test/plugins/agent.js
+++ b/packages/dd-trace/test/plugins/agent.js
@@ -445,13 +445,17 @@ module.exports = {
    * @param {boolean} [options.rejectFirst=false] - If true, reject the first time the callback throws.
    * @returns Promise
    */
-  assertFirstTraceSpan (callback, options) {
+  assertFirstTraceSpan (callbackOrExpected, options) {
     return runCallbackAgainstTraces(function (traces) {
-      if (typeof callback !== 'function') {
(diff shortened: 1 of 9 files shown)


packages/dd-trace/test/plugins/agent.js
```diff
@@ -438,15 +439,25 @@ module.exports = {
...
-  assertFirstTraceSpan (callback, options) {
+  assertFirstTraceSpan (callbackOrExpected, options) {
     return runCallbackAgainstTraces(function (traces) {
-      return callback(traces[0][0])
+      if (typeof callbackOrExpected !== 'function') {
+        try {
+          assertObjectContains(traces[0][0], callbackOrExpected)
+        } catch (error) {
+          // eslint-disable-next-line no-console
+          console.error('Expected span %o did not match traces:\n%o', callbackOrExpected, traces)
```
rochdev, line 455: This will only log the traces for the current request. If the issue is incorrect ordering for example, it will be extremely difficult to debug with the current setup. The output should be improved ASAP, otherwise failing tests will become a big headache without rewriting the test just to get the full output.

PR #5806, opened 2025-06-03
Add blocking for fastify query and body

IlyasShabi:
### What does this PR do?
Allow blocking on fastify query and body

### Motivation

Support fastify in Appsec 

### Plugin Checklist

- [x] Unit tests.
- [x] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
+++ b/packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
@@ -195,7 +195,7 @@ withVersions('fastify', 'fastify', version => {
             type: 'object',
             required: ['validField'],
             properties: {
-              validField: { type: 'string'}
+              validField: { type: 'string' }
             }
           }
         }


packages/dd-trace/src/appsec/channels.js
IlyasShabi, line 1: sort channels alphabetically

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -106,11 +108,37 @@ function preHandler (request, reply, done) {
...
+        req,
+        res,
+        abortController,
+        query: request.query
+      })
+
+      if (abortController.signal.aborted) return
+    }
+
+    if (bodyParserReadCh.hasSubscribers && request.body !== undefined) {
```
uurien: why are you checking here `undefined` and not a falsy values like you are doing with `request.query`? 
And this for readability, why is the order here `hasSubscribers && body` and for query it is `query && hasSubscribers`?

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -0,0 +1,185 @@
...
+      server.close()
+      return agent.close({ ritmReset: false })
+    })
+
+    beforeEach(async () => {
+      requestBody = sinon.stub()
+      appsec.enable(new Config({
+        appsec: {
+          enabled: true,
+          rules: path.join(__dirname, 'express-rules.json')
```
uurien: can we rename this file? it is not only for express anymore.

IlyasShabi: Done

packages/datadog-instrumentations/src/fastify.js
IlyasShabi: Im checking falsy values for both now, order fixed too

uurien (dismissed):

packages/dd-trace/test/appsec/attacker-fingerprinting.fastify.plugin.spec.js
```diff
@@ -0,0 +1,80 @@
...
+
+    it('should report http fingerprints', async () => {
+      await axios.post('/?key=testattack',
+        {
+          headers: {
+            'User-Agent': 'test-user-agent',
+            headerName: 'headerValue',
+            'x-real-ip': '255.255.255.255'
+          }
+        }
```
CarlesDD: I'm not sure if this is intentional or not but it has caught my attention. In this request these headers are not being set, but the object with the `headers` key is what is being sent as a body.

CarlesDD:
The plugin tests fail on Node.js v18, as fastify 5 (>=3) only supports Node.js v20+. Not a big concern since v18 support should be dropped soon in the tracer, but worth mentioning.

CarlesDD (approved):
Apart from the above comments, which are not blocking, LGTM

packages/dd-trace/test/appsec/attacker-fingerprinting.fastify.plugin.spec.js
IlyasShabi: I updated the test, it's matching express fingerprint test now

CarlesDD (approved):

PR #5812, opened 2025-06-03
chore(tracing): remove async storage from mongo plugins

wconti27:
### What does this PR do?
Make mongo plugins Node 24 compatible

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/mongodb-core.js
+++ b/packages/datadog-instrumentations/src/mongodb-core.js
@@ -2,8 +2,7 @@
 
 const {
   channel,
-  addHook,
-  AsyncResource
+  addHook
 } = require('./helpers/instrument')
 const shimmer = require('../../datadog-shimmer')
 
@@ -56,22 +55,6 @@ addHook({ name: 'mongodb-core', versions: ['~3.1.10'], file: 'lib/wireprotocol/3
 addHook({ name: 'mongodb-core', versions: ['~3.1.10'], file: 'lib/wireprotocol/2_6_support.js' }, WireProtocol => {
   shimmer.wrap(WireProtocol.prototype, 'command', command => wrapUnifiedCommand(command, 'command'))
   return WireProtocol
-})
-
-addHook({ name: 'mongodb', versions: ['>=3.5.4 <4.11.0'], file: 'lib/utils.js' }, util => {
-  shimmer.wrap(util, 'maybePromise', maybePromise => function (parent, callback, fn) {
-    const asyncResource = new AsyncResource('bound-anonymous-fn')
-    const callbackIndex = arguments.length - 2
-
-    callback = arguments[callbackIndex]
-
-    if (typeof callback === 'function') {
-      arguments[callbackIndex] = asyncResource.bind(callback)
-    }
-
-    return maybePromise.apply(this, arguments)
-  })
-  return util
 })
 
 function wrapWp (wp) {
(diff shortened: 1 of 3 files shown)


packages/datadog-instrumentations/src/mongodb-core.js
```diff
@@ -58,22 +57,6 @@ addHook({ name: 'mongodb-core', versions: ['~3.1.10'], file: 'lib/wireprotocol/2
   return WireProtocol
 })
 
-addHook({ name: 'mongodb', versions: ['>=3.5.4 <4.11.0'], file: 'lib/utils.js' }, util => {
```
tlhunter: It seems weird that this hook was removed but not replaced with anything. Do you know why it is no longer needed?

wconti27: I have been following a pattern for these changes. Since we don't call channel.publish here, AFAIK we do not need this wrapper hook anymore since it only deals with binding the asyncresource to the callback. Going to refer to @rochdev since honestly im not sure why this was ever needed.

BridgeAR:
@wconti27 there seem to still be a few related test failures :)

https://github.com/DataDog/dd-trace-js/actions/runs/16173889662/job/45654105379?pr=5812

packages/datadog-instrumentations/src/mongodb-core.js
rochdev: This is technically still needed as it basically patches non-native promise implementations that can be used by `mongodb`. Since we're also instrumenting the library and not just a generic promise implementation it could be replaced with a channel, but I think it's safe to just leave it here as-is for now as it's not nesting async resources.

wconti27: okay, ive reverted the change and kept the function here.

packages/datadog-instrumentations/src/mongodb-core.js
```diff
@@ -148,61 +148,66 @@ function wrapCommand (command, operation, name) {
...
+        errorCh.publish(ctx)
       }
 
-      finishCh.publish()
-
-      if (callback) {
-        return callback.apply(this, arguments)
-      }
-    }))
+      return callback ? finishCh.runStores(ctx, callback, this, ...arguments) : finishCh.publish(ctx)
```
BridgeAR: The above code should already handle this case, since the callback is checked to be a function.

```suggestion
      return finishCh.runStores(ctx, callback, this, ...arguments)
```

wconti27: ok fixed it.

packages/datadog-plugin-mongoose/src/index.js
```diff
@@ -0,0 +1,22 @@
...
+
+class MongoosePlugin extends DatabasePlugin {
+  static get id () { return 'mongoose' }
+
+  constructor (tracer, config) {
+    super(tracer, config)
+
+    // Ensure the current store is preserved across the
+    // internal callback channel used by the instrumentation.
+    // We ignore the `ctx` argument and just return the active store so that
```
rochdev: Wouldn't that be equivalent to doing nothing since the active store will already be active by definition?

packages/datadog-instrumentations/src/mongoose.js
```diff
@@ -2,21 +2,48 @@
...
+        const ctx = {}
+        return callbackCh.runStores(ctx, original, this, ...arguments)
+      })
     }
 
     return addQueue.apply(this, arguments)
   }
 }
 
+function wrapExec (exec) {
```
rochdev: Why is this needed now if `wrapAddQueue` was sufficient before? My understanding is that `wrapAddQueue` handles all the of various execution functions and allows wrapping the callback in a centralized place. The difference might be that an additional channe; is probably needed at the time `addQueue` is called so that the plugin can capture the current store in order to reactivate it later.

wconti27: I removed this function. The change to use an additional channel at the time of `addQueue` worked

packages/datadog-plugin-mongoose/src/index.js
wconti27: Removed this. I added the subscribe and empty handler for the new channel to the mongodb plugin, so that we don't need to add a mongoose plugin.

packages/datadog-plugin-mongodb-core/src/index.js
```diff
@@ -12,6 +12,12 @@ class MongodbCorePlugin extends DatabasePlugin {
   // should be removed if one day this will be fixed
   static get peerServicePrecursors () { return [] }
 
+  constructor (tracer, config) {
+    super(tracer, config)
+    // this channel is for wrapping the callback of mongoose exec methods and handling store context, no handler needed
+    this.addBind('datadog:mongodb-core:collection:addQueue', () => {})
```
rochdev: Returning `undefined` here will make the current store `undefined` instead of restoring it as expected in `addQueue`. Instead there should be a `queue:start` and `queue:asyncStart` like for other asynchronous operations.

wconti27: Why do we need the `asyncStart`? there's only one place where this new channel is needed. But makes sense on the part where we should restore the context.

rochdev: I didn't really look at the code in depth, it may not be needed if another event already captures the current store on the context, but restoring it is definitely needed.

packages/datadog-plugin-mongodb-core/src/index.js
```diff
@@ -12,6 +12,14 @@ class MongodbCorePlugin extends DatabasePlugin {
   // should be removed if one day this will be fixed
   static get peerServicePrecursors () { return [] }
 
+  constructor (tracer, config) {
+    super(tracer, config)
+    // this channel is for wrapping the callback of mongoose exec methods and handling store context, no handler needed
+    this.addBind('datadog:mongooose:queue:start', (ctx) => {
+      return ctx.currentStore
```
rochdev: Where is `ctx.currentStore` set? It looks like the `addQueueCh` channel happens outside of the `addQueue` function, so it should be the one assigning the store to then be retrieve from another channel inside of `addQueue`.

tlhunter (approved):

PR #5815, opened 2025-06-03
Baggage update

ida613:
### What does this PR do?
Use span._baggageItems only for [legacy ot-baggage](https://github.com/DataDog/dd-trace-js/commit/2faa3e4e836be19f2a801221493e62d7538b846e) to avoid breaking changes. Use storage('baggage') for otel-compatible [baggage](https://opentelemetry.io/docs/concepts/signals/baggage/) exclusively.

Baggage will pass all [weblog system tests](https://github.com/DataDog/system-tests/blob/main/tests/test_baggage.py) after the changes in this PR. We will stop using [parametric system tests](https://github.com/DataDog/system-tests/blob/main/tests/parametric/test_headers_datadog.py) for baggage.

### Plugin Checklist

- [x] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/baggage.js
+++ b/packages/dd-trace/src/baggage.js
@@ -23,7 +23,7 @@ function removeBaggageItem (keyToRemove) {
 }
 
 function removeAllBaggageItems () {
-  storage('baggage').enterWith({})
+  storage('baggage').enterWith(undefined)
   return storage('baggage').getStore()
 }
 
--- a/packages/dd-trace/src/baggage.js
+++ b/packages/dd-trace/src/baggage.js
@@ -23,7 +23,7 @@ function removeBaggageItem (keyToRemove) {
 }
 
 function removeAllBaggageItems () {
-  storage('baggage').enterWith(undefined)
+  storage('baggage').enterWith()
   return storage('baggage').getStore()
 }
 
--- a/packages/dd-trace/src/baggage.js
+++ b/packages/dd-trace/src/baggage.js
@@ -13,7 +13,7 @@ function getBaggageItem (key) {
 }
 
 function getAllBaggageItems () {
-  return storage('baggage').getStore()
+  return storage('baggage').getStore() || {}
 }
 
 function removeBaggageItem (keyToRemove) {
--- a/packages/dd-trace/src/baggage.js
+++ b/packages/dd-trace/src/baggage.js
@@ -13,7 +13,7 @@ function getBaggageItem (key) {
 }
 
 function getAllBaggageItems () {
-  return storage('baggage').getStore() || {}
+  return storage('baggage').getStore() ?? {}
 }
 
 function removeBaggageItem (keyToRemove) {
--- a/packages/dd-trace/src/opentracing/propagation/text_map.js
+++ b/packages/dd-trace/src/opentracing/propagation/text_map.js
@@ -645,11 +645,8 @@ class TextMapPropagator {
         removeAllBaggageItems()
         return
       }
-      if (spanContext) {
-        // eslint-disable-next-line unicorn/no-lonely-if
-        if (this._config.baggageTagKeys === '*' || keysToSpanTag.has(key)) {
-          spanContext._trace.tags['baggage.' + key] = value
-        }
+      if (spanContext && (this._config.baggageTagKeys === '*' || keysToSpanTag.has(key))) {
+        spanContext._trace.tags['baggage.' + key] = value
       }
       setBaggageItem(key, value)
     }
--- a/packages/dd-trace/src/opentracing/propagation/text_map.js
+++ b/packages/dd-trace/src/opentracing/propagation/text_map.js
@@ -646,6 +646,7 @@ class TextMapPropagator {
         return
       }
       if (spanContext) {
+        // eslint-disable-next-line unicorn/no-lonely-if
         if (this._config.baggageTagKeys === '*' || keysToSpanTag.has(key)) {
           spanContext._trace.tags['baggage.' + key] = value
         }
--- a/packages/dd-trace/src/opentracing/propagation/text_map.js
+++ b/packages/dd-trace/src/opentracing/propagation/text_map.js
@@ -136,7 +136,7 @@ class TextMapPropagator {
       let itemCounter = 0
       let byteCounter = 0
 
-      const baggageItems = spanContext ? spanContext._baggageItems : getAllBaggageItems()
+      const baggageItems = getAllBaggageItems()
       if (!baggageItems) return
       for (const [key, value] of Object.entries(baggageItems)) {
         const item = `${this._encodeOtelBaggageKey(String(key).trim())}=${encodeURIComponent(String(value).trim())},`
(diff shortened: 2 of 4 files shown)


packages/dd-trace/src/opentracing/propagation/text_map.js
```diff
@@ -629,33 +629,29 @@ class TextMapPropagator {
...
       if (spanContext) {
-        if (Object.hasOwn(spanContext._baggageItems, key)) continue
-        spanContext._baggageItems[key] = value
+        // eslint-disable-next-line unicorn/no-lonely-if
         if (this._config.baggageTagKeys === '*' || keysToSpanTag.has(key)) {
           spanContext._trace.tags['baggage.' + key] = value
         }
-      } else {
-        setBaggageItem(key, value)
       }
```
BridgeAR: Deactivating the lint rule will of course work while that should only ever be done in case the rule is making a mistake :) So instead of deactivating the rule, let's address what it's complaining about:

```suggestion
        if (spanContext && (this._config.baggageTagKeys === '*' || keysToSpanTag.has(key))) {
          spanContext._trace.tags['baggage.' + key] = value
        }
```

packages/dd-trace/test/opentracing/propagation/text_map.spec.js
```diff
@@ -431,7 +431,8 @@ describe('TextMapPropagator', () => {
         baggage: '%22%2C%3B%5C%28%29%2F%3A%3C%3D%3E%3F%40%5B%5D%7B%7D=%22%2C%3B%5C'
       }
       const spanContext = propagator.extract(carrier)
-      expect(spanContext._baggageItems).to.deep.equal({ '",;\\()/:<=>?@[]{}': '",;\\' })
+      expect(spanContext._baggageItems).to.deep.equal({})
```
BridgeAR, line 434: Should this not be undefined? I guess for legacy reasons it's kept?

packages/dd-trace/test/opentracing/propagation/text_map.spec.js
```diff
@@ -104,55 +104,35 @@ describe('TextMapPropagator', () => {
       expect(carrier['ot-baggage-bool']).to.equal('true')
       expect(carrier['ot-baggage-array']).to.equal('foo,bar')
       expect(carrier['ot-baggage-object']).to.equal('[object Object]')
```
BridgeAR, line 106: Unrelated to PR question: seeing this, I wonder if that is spected somewhere or if we could improve the handling for such input. Do you know if there is a definition for the behavior for that somewhere?

packages/dd-trace/src/opentracing/propagation/text_map.js
```diff
@@ -629,33 +629,29 @@ class TextMapPropagator {
...
     if (!carrier || !carrier.baggage) return
-    if (!spanContext) removeAllBaggageItems()
     const baggages = carrier.baggage.split(',')
     const keysToSpanTag = this._config.baggageTagKeys === '*'
       ? undefined
       : new Set(this._config.baggageTagKeys.split(','))
     for (const keyValue of baggages) {
       if (!keyValue.includes('=')) {
-        if (spanContext) spanContext._baggageItems = {}
+        removeAllBaggageItems()
```
BridgeAR, line 638: Just for me to understand: the spec defines that as soon as a single entry can not be parsed properly to discard everything?
I am surprised it would not just ignore the individual entry.

BridgeAR (commented):
I had a brief look and this looks solid to me!
One thing to update and just some questions for more context for me :)

The parametric test is currently failing, so I guess that would be removed before this should land?

packages/dd-trace/src/opentracing/propagation/text_map.js
ida613: Thank you Ruben for teaching me how to address this! :)

packages/dd-trace/test/opentracing/propagation/text_map.spec.js
ida613: yup it's for [legacy reasons](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/opentracing/span.js#L336)

packages/dd-trace/test/opentracing/propagation/text_map.spec.js
ida613: This behaviour was [added 8 years ago](https://github.com/DataDog/dd-trace-js/commit/2faa3e4e836be19f2a801221493e62d7538b846e) by @rochdev, following opentracing spec. I kept it for legacy purposes, you can ask Roch for the details. :)

packages/dd-trace/src/opentracing/propagation/text_map.js
ida613: yes this is specified in the [RFC](https://docs.google.com/document/d/1cX4bWYW-EPqKc64oynXpeHOI8FCW3XI-h0uH1tnzV3Q/edit?tab=t.0#heading=h.5xra5z6871b3)

packages/dd-trace/src/opentracing/propagation/text_map.js
BridgeAR: You're welcome :)

packages/dd-trace/test/opentracing/propagation/text_map.spec.js
BridgeAR: Hm, I can't fully follow. That would also work in case `_baggageItems` is undefined (`Object.assign({}, undefined)` equals {})

packages/dd-trace/test/opentracing/propagation/text_map.spec.js
rochdev: Indeed this was for OT and should be considered deprecated at this point, so I wouldn't touch it.

packages/dd-trace/src/opentracing/propagation/text_map.js
BridgeAR: Thank you!

packages/dd-trace/test/opentracing/propagation/text_map.spec.js
ida613: sorry [this](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/opentracing/span_context.js#L23) is probably a better file

BridgeAR: I am still not following. I guess it is not critical while I don't believe that _baggageItems would have to be an empty object. I believe it's fine to change it to just be undefined. Or did we ever declare this property for a customer to be used somewhere? It's an underscored property so it is meant to be internal.

ida613: We could, but since we are only using it for legacy opentracing baggage, I'm not sure if we should change it

BridgeAR: We could maybe clean this up in our next major.

BridgeAR (approved):

packages/dd-trace/test/opentracing/propagation/text_map.spec.js
```diff
@@ -442,6 +443,7 @@ describe('TextMapPropagator', () => {
       }
       const spanContextA = propagator.extract(carrierA)
       expect(spanContextA._baggageItems).to.deep.equal({})
+      expect(getAllBaggageItems()).to.be.undefined
```
rochdev: Is it expected that this can return undefined? As a user I would expect to get an empty object when there are no items. We don't necessarily need to actually store an empty object underneath but I think we should return one when the store is empty.

rochdev (changes requested):

packages/dd-trace/test/opentracing/propagation/text_map.spec.js
BridgeAR: I asked for clarification in the RFC

ida613: updated getAllBaggageItems to return `{}` when `storage('baggage').getStore()` is undefined

packages/dd-trace/src/baggage.js
```diff
@@ -13,7 +13,7 @@ function getBaggageItem (key) {
 }
 
 function getAllBaggageItems () {
-  return storage('baggage').getStore()
+  return storage('baggage').getStore() || {}
```
BridgeAR: Nit
```suggestion
  return storage('baggage').getStore() ?? {}
```

rochdev (approved):
LGTM, but let's wait for the RFC clarification before merging.

ida613:
> LGTM, but let's wait for the RFC clarification before merging.

Thank you Roch! @lucaspimentel [replied](https://docs.google.com/document/d/1cX4bWYW-EPqKc64oynXpeHOI8FCW3XI-h0uH1tnzV3Q/edit?disco=AAABlVlekcE) saying we should follow the convention for each language, so I think an empty object is good for our purposes

PR #5825, opened 2025-06-04
fix store binding transform not executing with undefined current store

rochdev:
### What does this PR do?

Fix store binding transform not executing with undefined current store.

### Motivation

This check was originally added to allow running transforms for bindings when the current store has `noop: true` so that the previous store (without noop) would be restored. When I implemented the check I didn't consider cases where the current store is `undefined`.

### DIFF
--- a/packages/dd-trace/src/plugins/plugin.js
+++ b/packages/dd-trace/src/plugins/plugin.js
@@ -34,7 +34,7 @@ class StoreBinding {
     this._transform = data => {
       const store = storage('legacy').getStore()
 
-      return !store || !store.noop || data?.currentStore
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/plugins/plugin.js
```diff
@@ -34,7 +34,7 @@ class StoreBinding {
     this._transform = data => {
       const store = storage('legacy').getStore()
 
-      return !store || !store.noop || data?.currentStore
+      return !store || !store.noop || data?.hasOwnProperty('currentStore')
```
BridgeAR:
```suggestion
      return !store || !store.noop || data && Object.hasOwn(data, 'currentStore')
```

BridgeAR (approved):

PR #5830, opened 2025-06-05
[test optimization] Remove deprecated async resource from mocha

juan-fernandez:
### What does this PR do?
Remove async resource usage from mocha. 

### Motivation
It's no longer maintained and might be removed soon.

### Plugin Checklist
Make sure current mocha tests pass

### DIFF
--- a/integration-tests/ci-visibility/test-custom-tags/custom-tags.js
+++ b/integration-tests/ci-visibility/test-custom-tags/custom-tags.js
@@ -1,21 +1,21 @@
 const { expect } = require('chai')
 const sum = require('../test/sum')
-const ddTrace = require('dd-trace')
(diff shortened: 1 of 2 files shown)


integration-tests/ci-visibility/test-custom-tags/custom-tags.js
```diff
@@ -1,21 +1,21 @@
 const { expect } = require('chai')
 const sum = require('../test/sum')
-const ddTrace = require('dd-trace')
+const tracer = require('dd-trace')
 
-describe('ci visibility', () => {
+describe('test optimization custom tags', () => {
```
nikita-tkachenko-datadog, line 5: the renaming never stops :)

PR #5831, opened 2025-06-05
Get response headers on Fastify by the end of the request

IlyasShabi:
### What does this PR do?

Fastify calls `writeHead` directly to set response headers, but Node.js doesn't store these headers in the internal structure that `res.getHeaders()` reads from. 
This causes [getHeaders](https://github.com/DataDog/dd-trace-js/blob/d0de1c38ac9d7767642d5884a7531f0d62e66326/packages/dd-trace/src/appsec/reporter.js#L150) to return empty object.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/integration-tests/appsec/iast.esm-security-controls.spec.js
+++ b/integration-tests/appsec/iast.esm-security-controls.spec.js
@@ -8,114 +8,118 @@ const { assert } = require('chai')
 describe('ESM Security controls', () => {
   let axios, sandbox, cwd, appFile, agent, proc
 
-  before(async function () {
-    this.timeout(process.platform === 'win32' ? 90000 : 30000)
-    sandbox = await createSandbox(['express@4']) // TODO: Remove pinning once our tests support Express v5
-    cwd = sandbox.folder
-    appFile = path.join(cwd, 'appsec', 'esm-security-controls', 'index.mjs')
-  })
+  ['4', '5'].forEach(version => {
+    describe(`With express v${version}`, () => {
+      before(async function () {
+        this.timeout(process.platform === 'win32' ? 90000 : 30000)
+        sandbox = await createSandbox([`express@${version}`])
+        cwd = sandbox.folder
+        appFile = path.join(cwd, 'appsec', 'esm-security-controls', 'index.mjs')
+      })
 
-  after(async function () {
-    await sandbox.remove()
-  })
+      after(async function () {
+        await sandbox.remove()
+      })
 
-  const nodeOptions = '--import dd-trace/initialize.mjs'
+      const nodeOptions = '--import dd-trace/initialize.mjs'
 
-  beforeEach(async () => {
-    agent = await new FakeAgent().start()
+      beforeEach(async () => {
+        agent = await new FakeAgent().start()
 
-    proc = await spawnProc(appFile, {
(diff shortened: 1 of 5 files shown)


packages/dd-trace/src/appsec/reporter.js
```diff
@@ -140,14 +140,14 @@ function filterExtendedHeaders (headers, excludedHeaderNames, tagPrefix, limit =
...
-function getCollectedHeaders (req, res, shouldCollectEventHeaders) {
+function getCollectedHeaders (req, res, shouldCollectEventHeaders, storedResponseHeaders) {
   // Mandatory
   const mandatoryCollectedHeaders = filterHeaders(req.headers, REQUEST_HEADERS_MAP)
 
   // Basic collection
   if (!shouldCollectEventHeaders) return mandatoryCollectedHeaders
 
-  const responseHeaders = res.getHeaders()
+  const responseHeaders = storedResponseHeaders || res.getHeaders()
```
uurien: shouldn't this be a merge of these things? whatever is in `storedResponseHeaders` + `res.getHeaders()`? Like:
```javascript
const responseHeaders = { ...storedResponseHeaders, ...res.getHeaders() }
```

IlyasShabi: Is there a case where we could have different headers from getHeaders and stored ones?

uurien: Yes, in fastify for example, if for some reason (a lib or whatever) is using `setHeaders` in `reply.raw` we will have different data in `storedResponseHeaders` and in `res.getHeaders`.

Corner cases, but they are cases anyway.

IlyasShabi: Yes Indeed, done

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -140,14 +140,14 @@ function filterExtendedHeaders (headers, excludedHeaderNames, tagPrefix, limit =
...
-function getCollectedHeaders (req, res, shouldCollectEventHeaders) {
+function getCollectedHeaders (req, res, shouldCollectEventHeaders, storedResponseHeaders) {
   // Mandatory
   const mandatoryCollectedHeaders = filterHeaders(req.headers, REQUEST_HEADERS_MAP)
 
   // Basic collection
   if (!shouldCollectEventHeaders) return mandatoryCollectedHeaders
 
-  const responseHeaders = res.getHeaders()
+  const responseHeaders = { ...storedResponseHeaders, ...res.getHeaders() }
```
uurien: I know that I suggested this change, but we can't prevent creating copies of the objects when they are not necessary. for example:

```suggestion
  const responseHeaders = storedResponseHeaders ? { ...storedResponseHeaders, ...res.getHeaders() } : res.getHeaders()
```

IlyasShabi: Good catch, Solved

uurien (approved):

PR #5832, opened 2025-06-05
Crysmags/ws integration

crysmags:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/ws.js
+++ b/packages/datadog-instrumentations/src/ws.js
@@ -1,9 +1,7 @@
 'use strict'
 
 const {
-  addHook,
-  channel,
-  AsyncResource
+  addHook
 } = require('./helpers/instrument')
 const shimmer = require('../../datadog-shimmer')
 
@@ -21,6 +19,21 @@ function createWrapRequest (ws, options) {
     }
   }
 }
+function createWrapEmit (ctx) {
+  return function wrapEmit (emit) {
+    return function (title, headers, req) {
+      ctx.resStatus = headers[0].split(' ')[1]
+
+      return ch.asyncStart.runStores(ctx, () => {
+        try {
+          return emit.apply(this, arguments)
+        } finally {
+          ch.asyncEnd.publish(ctx)
+        }
+      })
+    }
+  }
+}
 
 function wrapHandleUpgrade (handleUpgrade) {
   return function () {
@@ -30,6 +43,7 @@ function wrapHandleUpgrade (handleUpgrade) {
     const ctx = { req }
 
     return ch.tracePromise(() => {
+      shimmer.wrap(this, 'emit', createWrapEmit(ctx))
       handleUpgrade.call(this, req, socket, head, cb)
     }, ctx)
   }
--- a/packages/datadog-instrumentations/src/ws.js
+++ b/packages/datadog-instrumentations/src/ws.js
@@ -1,0 +1,45 @@
+'use strict'
+
+const {
+  addHook,
+  channel,
+  AsyncResource
+} = require('./helpers/instrument')
+const shimmer = require('../../datadog-shimmer')
+
+const tracingChannel = require('dc-polyfill').tracingChannel
+const ch = tracingChannel('ws:client:connect')
+
+function createWrapRequest (ws, options) {
+  return function wrapRequest (request) {
+    return function (headers) {
+      if (!ch.start.hasSubscribers) return request.apply(this, arguments)
+
+      const ctx = { headers, ws, options }
+
+      return ch.tracePromise(() => request.call(this), ctx)
+    }
+  }
+}
+
+function wrapHandleUpgrade (handleUpgrade) {
+  return function () {
+    if (!ch.start.hasSubscribers) return handleUpgrade.apply(this, arguments)
+
+    const [req, socket, head, cb] = arguments
+    const ctx = { req }
+
+    return ch.tracePromise(() => {
+      handleUpgrade.call(this, req, socket, head, cb)
+    }, ctx)
+  }
+}
+
+addHook({
+  name: 'ws',
+  file: 'lib/websocket-server.js'
+}, ws => {
+  shimmer.wrap(ws.prototype, 'handleUpgrade', wrapHandleUpgrade)
+
+  return ws
+})
--- a/packages/dd-trace/src/service-naming/schemas/v0/websocket.js
+++ b/packages/dd-trace/src/service-naming/schemas/v0/websocket.js
@@ -1,0 +1,18 @@
+const { identityService } = require('../util')
+
+const websocket = {
+  consumer: {
+    ws: {
+      opName: () => 'ws.request',
+      serviceName: identityService
+    }
+  },
+  producer: {
+    ws: {
+      opName: () => 'ws.request',
+      serviceName: identityService
+    }
+  }
+}
+
+module.exports = websocket
--- a/packages/dd-trace/src/service-naming/schemas/v1/websocket.js
+++ b/packages/dd-trace/src/service-naming/schemas/v1/websocket.js
@@ -1,0 +1,18 @@
+const { identityService } = require('../util')
+
+const websocket = {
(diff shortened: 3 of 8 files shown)


packages/dd-trace/src/service-naming/schemas/v0/websocket.js
```diff
@@ -0,0 +1,32 @@
...
+
+const { identityService } = require('../util')
+
+const websocket = {
+  request: {
+    websocket: {
+      opName: () => 'websocket.request',
+      serviceName: identityService
+    }
+  },
```
tlhunter: This seems superfluous

packages/dd-trace/src/service-naming/schemas/v1/websocket.js
```diff
@@ -0,0 +1,32 @@
...
+
+const { identityService } = require('../util')
+
+const websocket = {
+  request: {
+    websocket: {
+      opName: () => 'websocket.request',
+      serviceName: identityService
+    }
+  },
```
tlhunter: This also seems superfluous

tlhunter:
The remaining errors seem to all be related to `selenium`. Wouldn't you know it, [selenium depends on `ws`](https://www.npmjs.com/package/selenium-webdriver?activeTab=dependencies).

It might be that we're now instrumenting the `ws` used by `selenium` which is then breaking tests which rely on specific span data being produced. Or it could mean that we fundamentally broke `ws`.

packages/dd-trace/src/service-naming/schemas/v1/websocket.js
```diff
@@ -0,0 +1,30 @@
+'use strict'
+
+const websocket = {
+  request: {
+    ws: {
+      opName: () => 'websocket.request',
+      serviceName: ({ pluginConfig, tracerService }) => pluginConfig.service || tracerService
+    }
+  },
```
tlhunter: Is `websocket.request` ever emitted by the tracer?

packages/datadog-plugin-ws/test/index.spec.js
```diff
@@ -0,0 +1,331 @@
...
+            connectionReceived = true
+            ws.send('test message')
+          })
+
+          client.on('message', (msg) => {
+            expect(msg.toString()).to.equal('test message')
+          })
+
+          return agent.assertSomeTraces(traces => {
+            expect(traces[0][0]).to.have.property('name', 'websocket.request')
```
tlhunter: I believe this should be `websocket.send` or `websocket.receive`.

packages/datadog-plugin-ws/test/index.spec.js
```diff
@@ -0,0 +1,331 @@
...
+        })
+
+        it('should work with custom service configuration', () => {
+          wsServer.on('connection', (ws) => {
+          })
+          messageReceived = false
+
+          return agent.assertSomeTraces(traces => {
+            expect(traces[0][0]).to.have.property('service', 'custom-ws-service')
+            expect(traces[0][0]).to.have.property('name', 'websocket.request')
```
tlhunter: I believe this should be `websocket.send` or `websocket.receive`.

packages/datadog-plugin-ws/test/index.spec.js
```diff
@@ -0,0 +1,331 @@
...
+          messageReceived = false
+
+          client.on('message', (data) => {
+            expect(data.toString()).to.equal('test message')
+            messageReceived = true
+          })
+
+          return agent.assertSomeTraces(traces => {
+            expect(traces[0][0]).to.have.property('service', 'custom-ws-service')
+            expect(traces[0][0]).to.have.property('name', 'websocket.request')
```
tlhunter: I believe this should be `websocket.send` or `websocket.receive`.

packages/datadog-plugin-ws/test/index.spec.js
```diff
@@ -0,0 +1,331 @@
...
+        })
+
+        it('should not produce close event spans when traceWebsocketMessagesEnabled is not set to true', () => {
+          wsServer.on('connection', (ws) => {
+            ws.close()
+          })
+
+          return agent.assertSomeTraces(traces => {
+            expect(traces[0][0]).to.have.property('service', 'custom-ws-service')
+            expect(traces[0][0]).to.have.property('name', 'websocket.request')
```
tlhunter: I believe this should be `websocket.send` or `websocket.receive`.

packages/datadog-instrumentations/src/ws.js
```diff
@@ -0,0 +1,161 @@
...
+})
+
+addHook({
+  name: 'ws',
+  file: 'lib/websocket.js',
+  versions: ['>=8.0.0']
+}, ws => {
+  shimmer.wrap(ws.prototype, 'close', wrapClose)
+  return ws
+})
```
tlhunter:
```suggestion
addHook({
  name: 'ws',
  file: 'lib/websocket-server.js',
  versions: ['>=8.0.0']
}, ws => {
  shimmer.wrap(ws.prototype, 'handleUpgrade', wrapHandleUpgrade)
  shimmer.wrap(ws.prototype, 'emit', createWrapEmit)
  return ws
})

addHook({
  name: 'ws',
  file: 'lib/websocket.js',
  versions: ['>=8.0.0']
}, ws => {
  shimmer.wrap(ws.prototype, 'send', wrapSend)
  shimmer.wrap(ws.prototype, 'on', wrapListener)
  shimmer.wrap(ws.prototype, 'close', wrapClose)
  return ws
})
```

packages/dd-trace/src/service-naming/schemas/v1/websocket.js
crysmags: Yes, this is the http handshake trace. The server plugin has `kind` set to `request`

packages/datadog-plugin-ws/test/index.spec.js
crysmags: in this case we are not producing message spans so it will be `web.request`

packages/datadog-plugin-ws/test/index.spec.js
crysmags: not sending any messages in this case so I only have the handshake trace

packages/datadog-plugin-ws/test/index.spec.js
crysmags: this is now `web.request`

packages/datadog-plugin-ws/test/index.spec.js
```diff
@@ -0,0 +1,331 @@
...
+
+const { expect } = require('chai')
+const { withVersions } = require('../../dd-trace/test/setup/mocha')
+const agent = require('../../dd-trace/test/plugins/agent')
+
+describe('Plugin', () => {
+  let WebSocket
+  let wsServer
+  let connectionReceived
+  let clientPort = 6015
```
tlhunter, line 11: Hard-coding a port like this may cause test flakiness in the future. To get around it you can listen on port 0, then the OS picks a random free port, and then later once the server is listening the callback should provide the chosen port that you can use with the clients. It's probably fine now but should def be changed in the future.

packages/datadog-instrumentations/src/ws.js
```diff
@@ -0,0 +1,136 @@
...
+          serverCh.asyncEnd.publish(ctx)
+        }
+      }, this, ...arguments)
+    }
+    return serverCh.traceSync(handleUpgrade, ctx, this, ...arguments)
+  }
+}
+
+function wrapSend (send) {
+  return function wrappedSend (...args) {
```
tlhunter, line 39:
```suggestion
  return function wrappedSend () {
```

I don't think `args` is being used and should be safe to remove. That said I'm not sure why the linter didn't complain.

PR #5837, opened 2025-06-05
Fix kafka clusterId gathering possible to crash

BridgeAR:
This improves the kafka client performance as well as fixing a bug in case the clusterId gathering would have errored. In that case, the library would have ended with an unhandled rejection.

### DIFF
--- a/packages/datadog-plugin-kafkajs/src/producer.js
+++ b/packages/datadog-plugin-kafkajs/src/producer.js
@@ -88,11 +88,6 @@ class KafkajsProducerPlugin extends ProducerPlugin {
     }
     for (const message of messages) {
       if (message !== null && typeof message === 'object') {
-        // message headers are not supported for kafka broker versions <0.11
-        if (!disableHeaderInjection) {
-          message.headers ??= {}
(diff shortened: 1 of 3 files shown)


packages/datadog-plugin-kafkajs/src/producer.js
```diff
@@ -86,25 +86,20 @@ class KafkajsProducerPlugin extends ProducerPlugin {
...
+        }
 
-          const dataStreamsContext = this.tracer.setCheckpoint(edgeTags, span, payloadSize)
-          if (!disableHeaderInjection) {
-            DsmPathwayCodec.encode(dataStreamsContext, message.headers)
-          }
+        const dataStreamsContext = this.tracer.setCheckpoint(edgeTags, span, payloadSize)
+        if (!disableHeaderInjection) {
+          // Message headers are not supported for kafka broker versions <0.11
+          this.tracer.inject(span, 'text_map', message.headers)
```
BridgeAR, line 101: I am not sure if this move is actually correct. That would have an impact on other changes in here as well.

BridgeAR:
Superseded

PR #5839, opened 2025-06-05
[SVLS-5649] Add HTTP Context Propagation to Azure Functions

jcstorms1:
### What does this PR do?

Updates the current Azure Function plugin to include HTTP context propagation.

image

### Motivation

In order to provide customers with a complete distributed trace, HTTP triggered functions need to propagate trace context correctly. Previously HTTP triggered functions were not parenting correctly due to the lack of context propagation. 

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [x] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -21,12 +21,13 @@ class AzureFunctionsPlugin extends TracingPlugin {
...
 
   static get prefix () { return 'tracing:datadog:azure:functions:invoke' }
-
   bindStart (ctx) {
-    const { functionName, methodName } = ctx
+    const { functionName, methodName, httpRequest } = ctx
     const store = storage('legacy').getStore()
-
+    const ddTraceContext = Object.fromEntries(httpRequest.headers.entries())
+    const childOf = this._tracer.extract('http_headers', ddTraceContext) || null
```
BridgeAR: Nit
```suggestion
    const childOf = this._tracer.extract('http_headers', ddTraceContext) ?? null
```

rochdev (approved):

packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -21,12 +21,13 @@ class AzureFunctionsPlugin extends TracingPlugin {
   static get type () { return 'serverless' }
 
   static get prefix () { return 'tracing:datadog:azure:functions:invoke' }
-
   bindStart (ctx) {
-    const { functionName, methodName } = ctx
+    const { functionName, methodName, httpRequest } = ctx
     const store = storage('legacy').getStore()
-
+    const ddTraceContext = Object.fromEntries(httpRequest.headers.entries())
```
BridgeAR: Should this just copy the headers? Because we now create an object instead of keeping it an array? And do we really need to copy them in the first place or could we just pass the headers through directly?

```suggestion
    const ddTraceContext = [...httpRequest.headers]
```

[edit] it turned out it's a map being used here. So it's all good :)

jcstorms1: It doesn't pass the integration tests when I do this unfortunately.

rochdev: It needs to be an object in a specific format so that the propagator can extract and inject consistently.

PR #5842, opened 2025-06-05
Fix startuplog crashing by serializing bigint

### LINKED ISSUE #5829: [BUG]: TypeError: Do not know how to serialize a BigInt
### Tracer Version(s)

5.54.0

### Node.js Version(s)

20x

### Bug Report

after upgrading app crashes like such:
TypeError: Do not know how to serialize a BigInt
    at JSON.stringify (<anonymous>)
    at Object.toString (/usr/src/app/node_modules/dd-trace/packages/dd-trace/src/startup-log.js:64:19)
    at startupLog (/usr/src/app/node_modules/dd-trace/packages/dd-trace/src/startup-log.js:46:42)
    at /usr/src/app/node_modules/dd-trace/packages/dd-trace/src/exporters/agent/writer.js:98:7
    at IncomingMessage.<anonymous> (/usr/src/app/node_modules/dd-trace/packages/dd-trace/src/exporters/common/request.js:93:11)
    at IncomingMessage.emit (node:events:536:35)
    at endReadableNT (node:internal/streams/readable:1698:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)

### Reproduction Code

_No response_

### Error Logs

_No response_

### Tracer Config

_No response_

### Operating System

_No response_

### Bundling

Unsure

BridgeAR:
Instead, handle bigint and symbols in the future.
Also make the bigint which caused it private and use a getter for
tests instead.

Fixes #5829

### DIFF
--- a/packages/dd-trace/src/sampler.js
+++ b/packages/dd-trace/src/sampler.js
@@ -21,6 +21,8 @@ class Sampler {
    * @param {number} rate
    */
   constructor (rate) {
+    // TODO: Should this be moved up to the calling parts?
+    rate = Math.min(Math.max(rate, 0), 1)
     this._rate = rate
     this.#threshold = BigInt(Math.floor(rate * MAX_TRACE_ID))
   }
--- a/packages/dd-trace/src/sampler.js
+++ b/packages/dd-trace/src/sampler.js
@@ -16,12 +16,13 @@ const SAMPLING_KNUTH_FACTOR = 1_111_111_111_111_111_111n
  * This class uses a deterministic sampling algorithm that is consistent across all languages.
  */
 class Sampler {
+  #threshold = BigInt(0)
   /**
    * @param {number} rate
    */
   constructor (rate) {
     this._rate = rate
-    this._threshold = BigInt(Math.floor(rate * MAX_TRACE_ID))
+    this.#threshold = BigInt(Math.floor(rate * MAX_TRACE_ID))
   }
 
   /**
@@ -29,6 +30,10 @@ class Sampler {
    */
   rate () {
     return this._rate
+  }
+
+  get threshold () {
+    return this.#threshold
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/dd-trace/src/sampler.js
```diff
@@ -16,12 +16,15 @@ const SAMPLING_KNUTH_FACTOR = 1_111_111_111_111_111_111n
  * This class uses a deterministic sampling algorithm that is consistent across all languages.
  */
 class Sampler {
+  #threshold = BigInt(0)
```
watson: Nit:

```suggestion
  #threshold = 0n
```

packages/dd-trace/src/sampler.js
```diff
@@ -16,12 +16,15 @@ const SAMPLING_KNUTH_FACTOR = 1_111_111_111_111_111_111n
...
   /**
```
watson, line 21: Nit:

```suggestion

  /**
```

packages/dd-trace/test/startup-log.spec.js
```diff
@@ -5,10 +5,13 @@ require('./setup/tap')
 const os = require('os')
 const tracerVersion = require('../../../package.json').version
 const Config = require('../src/config')
+const SamplingRule = require('../src/sampling_rule')
+const assert = require('node:assert')
```
watson: Nit: I like Node.js core modules to be at the top when possible (we should have a linting rule for that)

```suggestion
const assert = require('node:assert')
const tracerVersion = require('../../../package.json').version
const Config = require('../src/config')
const SamplingRule = require('../src/sampling_rule')
```

BridgeAR: While I agree that it's easier to read it that way, I think we should just let a linter handle it with autofix at a later point.

watson (approved):

PR #5844, opened 2025-06-06
[test optimization] Fix test optimization tests in v5 release line

juan-fernandez:
### What does this PR do?

* Add polyfills to `cypress` code to include `Object#hasOwn` and `Array#at` functions, which are executed in the cypress plugin code path. 
* Do not attempt to call `.addTags` in playwright's older versions, since active test span is not supported there.

### Motivation
Fix cypress and playwright integration tests in v5 (which run in a lower version)

### Plugin Checklist
The ones we already have


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dd-trace",
-  "version": "6.0.0-pre",
+  "version": "5.0.0",
```
juan-fernandez: ℹ️ to be removed before merging, just to test

packages/datadog-plugin-cypress/test/app/cypress/plugins/index.js
```diff
@@ -1,7 +1,16 @@
...
   // run in a different process. We need to init a different tracer reporting to the
   // url set by the plugin agent
-  require('../../../../../dd-trace').init({ startupLogs: false })
+
+  const tracer = require('../../../../../dd-trace').init({
+    startupLogs: false,
+    isCiVisibility: true,
+  })
+  tracer.use('fs', false)
+  tracer.use('child_process', false)
```
juan-fernandez: this is just a testing quirk. We want the tracer init in the plugin test to look like `ci/init` as much as possible

packages/datadog-plugin-cypress/test/app/cypress/plugins/index.js
```diff
@@ -1,7 +1 @@
-module.exports = (on, config) => {
-  // We can't use the tracer available in the testing process, because this code is
-  // run in a different process. We need to init a different tracer reporting to the
-  // url set by the plugin agent
-  require('../../../../../dd-trace').init({ startupLogs: false })
-  require('../../../../src/plugin')(on, config)
-}
+module.exports = require('../../../../../../ci/cypress/plugin')
```
juan-fernandez, line 1: ℹ️ this change is just to run this test more closely matching what the user would do

packages/datadog-plugin-cypress/test/index.spec.js
```diff
@@ -34,6 +34,9 @@ describe('Plugin', function () {
       return agent.load()
     })
     beforeEach(function (done) {
+      // This test does not check test optimization agentless protocol,
+      // so we'll make the tracer default to reporting to v0.4/traces
+      agent.setAvailableEndpoints([])
```
juan-fernandez, line 39: ℹ️ Necessary because of the change in `packages/datadog-plugin-cypress/test/app/cypress/plugins/index.js`. Now the tracer is initialized in test visibility mode, as it should.

tlhunter (dismissed):

PR #5848, opened 2025-06-06
test: reduce usage of get-port dependency

watson:
### What does this PR do?

Remove the usage of the `get-port` dependency everywhere that we use the `spawnProc` helper method. This method have a way to communicate the bound port back to the parent process, so we don't need to pass a known one into the process when spawning it.

### Motivation

Using the `get-port` dependency comes with an inherent risk of a race condition, in which the returned port might be claimed by another process between it's returned by `get-port` and we bind to it. Therefore we should rely on other means of knowning which port our test-apps bind to when testing.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes

I left a reference to the `APP_PORT` environment variable in the sandboxed applications, so that it's possible to easily run these scripts manually for debugging purposes. I've found that useful myself many times.


watson:
* **#5848** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5848?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5848?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #5853, opened 2025-06-07
test: destroy active connections when stopping fake agent

watson:
The fake agent is used in tests and is stopped after the test is done.

When calling its `stop` function, it will wait for active connections (sockets) to close before considering itself stopped. This can add an unnecessary waiting period after each test, as the timeout on the socket by default is two seconds in our code-base:

https://github.com/DataDog/dd-trace-js/blob/90fe64b9845d04d6273ec2b7a36ba6a7844b8253/packages/dd-trace/src/exporters/common/request.js#L56

For integration tests making heavy use of fake agents, this can add several minutes of overhead that can be avoided by destroying the sockets when calling the `stop` function on the fake agent.

This mainly affects tests running on Node.js 18, as sockets there for some reason takes a little longer to clear out. I haven't spend time figuring out why, but [here's a recent CI run](https://github.com/DataDog/dd-trace-js/actions/runs/15504375601) of `Debugger` on `master`:

image

Compare that, to [the same run in this PR](https://github.com/DataDog/dd-trace-js/actions/runs/15505590558):

image

### DIFF
--- a/integration-tests/helpers/fake-agent.js
+++ b/integration-tests/helpers/fake-agent.js
@@ -14,6 +14,7 @@ module.exports = class FakeAgent extends EventEmitter {
     super({ captureRejections: true })
     this.port = port
     this.resetRemoteConfig()
+    this._sockets = new Set()
   }
 
   async start () {
@@ -23,6 +24,15 @@ module.exports = class FakeAgent extends EventEmitter {
       }, 10000)
       this.server = http.createServer(buildExpressServer(this))
       this.server.on('error', reject)
+
+      // Track connections to force close them later
+      this.server.on('connection', (socket) => {
+        this._sockets.add(socket)
+        socket.on('close', () => {
+          this._sockets.delete(socket)
+        })
+      })
+
       this.server.listen(this.port, () => {
         this.port = this.server.address().port
         clearTimeout(timeoutObj)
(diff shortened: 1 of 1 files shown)


integration-tests/helpers/fake-agent.js
```diff
@@ -32,7 +42,14 @@ module.exports = class FakeAgent extends EventEmitter {
...
+
     return new Promise((resolve) => {
+      if (!this.server?.listening) return resolve()
+      for (const socket of this._sockets) {
+        socket.destroy()
+      }
+      this._sockets.clear()
       this.server.on('close', resolve)
       this.server.close()
     })
```
BridgeAR, line 55: We could make it very readable using `once` from the built-in events module.

```suggestion
  async stop () {
    if (!this.server?.listening) return

     for (const socket of this._sockets) {
      socket.destroy()
    }
    this._sockets.clear()
    this.server.close()
    return once(this.server, 'close')
```

BridgeAR (approved):
Interesting finding!
LGTM, just left a suggestion to improve readability of touched code (that existed before)

integration-tests/helpers/fake-agent.js
watson: Addressed in follow up PR: https://github.com/DataDog/dd-trace-js/pull/5854

PR #5857, opened 2025-06-09
[test-optimization] [SDTEST-2144] Fix `it.failing` in combination with EFD

Mariovido:
### What does this PR do?

This update resolves an issue that occurred when using EFD in combination with `it.failing`. The new behavior is:
- When both `it.failing` and EFD are enabled in a test &rarr; EFD will be ignored.

### Motivation

We want to solve an issue that was reported by a client.

### Plugin Checklist

- [x] Integration tests.

### DIFF
--- a/integration-tests/ci-visibility/test/failing-test.js
+++ b/integration-tests/ci-visibility/test/failing-test.js
@@ -1,0 +1,7 @@
+const { expect } = require('chai')
+
+describe('failing', () => {
+  it.failing('can report failed tests', () => {
+    expect(1 + 2).to.equal(4)
+  })
+})
--- a/integration-tests/ci-visibility/test/failing-test.js
+++ b/integration-tests/ci-visibility/test/failing-test.js
@@ -1,7 +1,0 @@
-const { expect } = require('chai')
-
-describe('failing', () => {
-  it.failing('can report failed tests', () => {
-    expect(1 + 2).to.equal(4)
-  })
-})
--- a/integration-tests/jest/jest.spec.js
+++ b/integration-tests/jest/jest.spec.js
@@ -2365,6 +2365,58 @@ describe('jest CommonJS', () => {
         }).catch(done)
       })
     })
+
+    it('does not retry when it.failing is used', (done) => {
+      receiver.setInfoResponse({ endpoints: ['/evp_proxy/v4'] })
+      const NUM_RETRIES_EFD = 3
+      receiver.setSettings({
+        early_flake_detection: {
+          enabled: true,
+          slow_test_retries: {
+            '5s': NUM_RETRIES_EFD
+          },
+          faulty_session_threshold: 100
+        },
+        known_tests_enabled: true
+      })
+      const eventsPromise = receiver
+        .gatherPayloadsMaxTimeout(({ url }) => url.endsWith('/api/v2/citestcycle'), (payloads) => {
+          const events = payloads.flatMap(({ payload }) => payload.events)
+
+          const testSession = events.find(event => event.type === 'test_session_end').content
+          assert.propertyVal(testSession.meta, TEST_EARLY_FLAKE_ENABLED, 'true')
+
+          const tests = events.filter(event => event.type === 'test').map(event => event.content)
+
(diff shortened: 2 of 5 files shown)


integration-tests/my-nyc.config.js
```diff
@@ -1,5 +1,5 @@
 // non default name so that it only gets picked up intentionally
 module.exports = {
-  exclude: ['node_modules/**'],
+  exclude: ['node_modules/**', 'ci-visibility/test/failing-test.js'],
```
juan-fernandez: why do we need this?

integration-tests/jest/jest.spec.js
```diff
@@ -2365,6 +2365,58 @@ describe('jest CommonJS', () => {
...
+          newTests.forEach(test => {
+            assert.notProperty(test.meta, TEST_IS_NEW)
+          })
+          assert.equal(newTests.length, 1)
+
+          const retriedTests = newTests.filter(test => test.meta[TEST_IS_RETRY] === 'true')
+          assert.equal(retriedTests.length, 0)
+          newTests.forEach(test => {
+            assert.equal(test.meta[TEST_NAME], 'failing can report failed tests')
+          })
```
juan-fernandez: nitpick: just to have the `newTests` assertions together
```suggestion
          assert.equal(newTests.length, 1)
          newTests.forEach(test => {
            assert.equal(test.meta[TEST_NAME], 'failing can report failed tests')
          })
          const retriedTests = newTests.filter(test => test.meta[TEST_IS_RETRY] === 'true')
          assert.equal(retriedTests.length, 0)
```

integration-tests/jest/jest.spec.js
```diff
@@ -2365,6 +2365,58 @@ describe('jest CommonJS', () => {
...
+        .gatherPayloadsMaxTimeout(({ url }) => url.endsWith('/api/v2/citestcycle'), (payloads) => {
+          const events = payloads.flatMap(({ payload }) => payload.events)
+
+          const testSession = events.find(event => event.type === 'test_session_end').content
+          assert.propertyVal(testSession.meta, TEST_EARLY_FLAKE_ENABLED, 'true')
+
+          const tests = events.filter(event => event.type === 'test').map(event => event.content)
+
+          const newTests = tests.filter(test =>
+            test.meta[TEST_SUITE] === 'ci-visibility/test/failing-test.js'
```
juan-fernandez: it'd also be interesting to assert the `test.status` for `it.failing` tests: 

If the test function throws, the test should be reported as passing, and if it doesn't, the test should be reported as failing. Could we check this?

integration-tests/ci-visibility/test/failing-test.js
```diff
@@ -0,0 +1,7 @@
+const { expect } = require('chai')
+
+describe('failing', () => {
+  it.failing('can report failed tests', () => {
+    expect(1 + 2).to.equal(4)
+  })
```
juan-fernandez: could we add a `it.failing` test that fails because its test function doesn't throw? Just to check complete compatibility 

```javascript
  it.failing('can report failing tests as failures', () => {
    expect(1 + 2).to.equal(3) // this passes but it should fail! So the test.status should be fail
  })
```

juan-fernandez (commented):
looking good! Some minor comments

integration-tests/my-nyc.config.js
Mariovido: There is no need. I moved the test to other folder to avoid having to add this line

integration-tests/jest/jest.spec.js
Mariovido: Changed!

integration-tests/jest/jest.spec.js
Mariovido: Added tthe assertions to check the `test.status`

integration-tests/ci-visibility/test/failing-test.js
Mariovido: Added a new test to check that case.

juan-fernandez (approved):

PR #5858, opened 2025-06-09
wip(vercel-ai): add vercel ai integration

sabrenner:
### What does this PR do?
WIP - assessing strategy of parsing OTel spans generated from the Vercel AI SDK (which works great, but seeing if it fits into library standards/if we can set a new precedent)

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/helpers/hooks.js
+++ b/packages/datadog-instrumentations/src/helpers/hooks.js
@@ -28,6 +28,7 @@ module.exports = {
   '@smithy/smithy-client': () => require('../aws-sdk'),
   '@vitest/runner': { esmFirst: true, fn: () => require('../vitest') },
   aerospike: () => require('../aerospike'),
+  ai: { esmFirst: true, fn: () => require('../vercel-ai') },
   amqp10: () => require('../amqp10'),
   amqplib: () => require('../amqplib'),
   avsc: () => require('../avsc'),
--- a/packages/datadog-instrumentations/src/vercel-ai.js
+++ b/packages/datadog-instrumentations/src/vercel-ai.js
@@ -1,0 +1,70 @@
+'use strict'
+
+const { addHook } = require('./helpers/instrument')
+const tracer = require('../../dd-trace')
+const shimmer = require('../../datadog-shimmer')
+
+const { TracerProvider } = tracer
+const provider = new TracerProvider()
+provider.register()
+
+const { channel } = require('dc-polyfill')
+const toolCreationChannel = channel('dd-trace:vercel-ai:tool')
+
+const TRACED_FUNCTIONS = {
+  generateText: wrapWithTracer,
+  streamText: wrapWithTracer,
+  generateObject: wrapWithTracer,
+  streamObject: wrapWithTracer,
+  embed: wrapWithTracer,
+  embedMany: wrapWithTracer,
+  tool: wrapTool
+}
+
+function wrapWithTracer (fn) {
+  return function () {
+    const options = arguments[0]
+    if (options.experimental_telemetry != null) return fn.apply(this, arguments)
+
+    options.experimental_telemetry = {
+      isEnabled: true,
+      // TODO(sabrenner): need to figure out how a user can configure this tracer
+      // maybe we advise they do this manually?
+      tracer: provider.getTracer()
+    }
+
+    return fn.apply(this, arguments)
+  }
+}
+
+function wrapTool (tool) {
+  return function () {
+    const args = arguments[0]
+    toolCreationChannel.publish(args)
+
+    return tool.apply(this, arguments)
+  }
+}
+
+addHook({
+  name: 'ai',
+  versions: ['>=4.0.0'],
+}, exports => {
+  // the exports from this package are not configurable
+  // to return wrapped functions with the tracer provided above,
+  // we need to copy the exports
+  const wrappedExports = {}
+
+  for (const [fnName, patchingFn] of Object.entries(TRACED_FUNCTIONS)) {
+    const original = exports[fnName]
+    wrappedExports[fnName] = shimmer.wrapFunction(original, patchingFn)
+  }
+
+  Object.getOwnPropertyNames(exports).forEach(prop => {
+    if (!Object.keys(TRACED_FUNCTIONS).includes(prop)) {
+      wrappedExports[prop] = exports[prop]
+    }
+  })
+
+  return wrappedExports
+})
--- a/packages/dd-trace/src/id.js
+++ b/packages/dd-trace/src/id.js
@@ -182,6 +182,6 @@ function writeUInt32BE (buffer, value, offset) {
   buffer[0 + offset] = value & 255
 }
 
-module.exports = function createIdentifier (value, radix) {
+module.exports = Object.assign(function createIdentifier (value, radix) {
   return new Identifier(value, radix)
-}
+}, { Identifier })
--- a/packages/dd-trace/src/llmobs/plugins/vercel-ai.js
+++ b/packages/dd-trace/src/llmobs/plugins/vercel-ai.js
@@ -1,0 +1,423 @@
+'use strict'
+
+const BaseLLMObsPlugin = require('./base')
+
+const { channel } = require('dc-polyfill')
+
+const otelSpanStartCh = channel('dd-trace:otel:span:start')
+const otelSpanFinishCh = channel('dd-trace:otel:span:finish')
+const toolCreationCh = channel('dd-trace:vercel-ai:tool')
+
+const { isVercelAISpan } = require('../../../../datadog-plugin-vercel-ai/src/util')
+const { MODEL_NAME, MODEL_PROVIDER, NAME } = require('../constants/tags')
+
+const SPANS_TO_USE_LLMOBS_PARENT = new Set([
+  'generateText',
+  'streamText',
+  'embed',
+  'embedMany',
+  'generateObject',
+  'streamObject'
+])
+
+const SPAN_NAME_TO_KIND_MAPPING = {
+  // embeddings
+  embed: 'workflow',
+  embedMany: 'workflow',
+  doEmbed: 'embedding',
+  // object generation
+  generateObject: 'workflow',
+  streamObject: 'workflow',
+  // text generation
+  generateText: 'workflow',
+  streamText: 'workflow',
+  // llm operations
+  doGenerate: 'llm',
+  doStream: 'llm',
+  // tools
+  toolCall: 'tool'
+}
+
+const MODEL_PROVIDER_MAPPING = {
+  'amazon-bedrock': 'bedrock',
+  vertex: 'vertexai',
+  'generative-ai': 'genai' // TODO(sabrenner): double check this
+}
+
+const MODEL_METADATA_KEYS = new Set([
+  'frequency_penalty',
+  'max_tokens',
+  'presence_penalty',
+  'temperature',
+  'top_p',
+  'top_k',
+  'stop_sequences'
+])
+
+/**
+ * @param {import('../../../../dd-trace/src/opentracing/span')} span
+ * @param {string} tag
+ * @returns {string}
+ */
+function getSpanTag (span, tag) {
+  const value = span.context()._tags[tag]
+  if (!value) return
+
+  return value
+}
+
+/**
+ * @param {import('../../../../dd-trace/src/opentracing/span')} span
+ * @returns {string}
+ */
+function getOperation (span) {
+  const name = span._name
+  if (!name) return
+
+  return name.split('.').pop()
+}
+
+function getUsage (span) {
+  const inputTokens = getSpanTag(span, 'ai.usage.promptTokens')
+  const outputTokens = getSpanTag(span, 'ai.usage.completionTokens')
+
+  return {
+    inputTokens,
+    outputTokens,
+    totalTokens: inputTokens + outputTokens
+  }
+}
+
+/**
+ * @param {import('../../../../dd-trace/src/opentracing/span')} span
+ * @returns {string}
+ */
+function getModelProvider (span) {
+  const modelProviderTag = span.context()._tags['ai.model.provider']
+  const providerParts = modelProviderTag?.split('.')
+  const provider = providerParts?.[0]
+
+  // TODO(sabrenner): explain or simplify this logic
+  switch (provider) {
+    case 'google':
+      return MODEL_PROVIDER_MAPPING[providerParts?.[1]] ?? provider
+    default:
+      return MODEL_PROVIDER_MAPPING[provider] ?? provider
+  }
+}
+
+function getJsonStringValue (str, defaultValue) {
+  let maybeValue = defaultValue
+  try {
+    maybeValue = JSON.parse(str)
+  } catch {
+    // do nothing
+  }
+
+  return maybeValue
+}
+
+function getModelMetadata (span) {
+  const modelMetadata = {}
+  for (const metadata of MODEL_METADATA_KEYS) {
+    const metadataTagKey = `gen_ai.request.${metadata}`
+    const metadataValue = getSpanTag(span, metadataTagKey)
+    if (metadataValue) {
+      modelMetadata[metadata] = metadataValue
+    }
+  }
+
+  return modelMetadata
+}
+
+function getGenerationMetadata (span) {
+  const metadata = {}
+  for (const tag of Object.keys(span.context()._tags)) {
+    if (!tag.startsWith('ai.settings')) continue
+
+    const settingKey = tag.split('.').pop()
+    const transformedKey = settingKey.replaceAll(/[A-Z]/g, letter => '_' + letter.toLowerCase())
+    if (MODEL_METADATA_KEYS.has(transformedKey)) continue
+
+    const settingValue = getSpanTag(span, tag)
+    metadata[settingKey] = settingValue
+  }
+
+  return metadata
+}
+
+class VercelAILLMObsPlugin extends BaseLLMObsPlugin {
+  static get id () { return 'vercel-ai' }
+  static get integration () { return 'vercel-ai' }
+
+  /** @type {Set<Record<string, any>>} */
+  #availableTools
+
+  /** @type {Record<string, string>} */
+  #toolCallIdsToName
+
+  constructor (...args) {
+    super(...args)
+
+    otelSpanStartCh.subscribe(({ ddSpan }) => {
+      if (!isVercelAISpan(ddSpan)) return
+
+      // holding context for llmobs parentage
+      const ctx = {
+        currentStore: { span: ddSpan }
+      }
+
+      const operation = getOperation(ddSpan)
+      const useLlmObsParent = SPANS_TO_USE_LLMOBS_PARENT.has(operation)
+
+      this.start(ctx, {
+        useLlmObsParent,
+        enterIntoLlmObsStorage: false
+      }) // triggers the getLLMObsSpanRegisterOptions
+    })
+
+    otelSpanFinishCh.subscribe(({ ddSpan }) => {
+      if (!isVercelAISpan(ddSpan)) return
+
+      const ctx = {
+        currentStore: { span: ddSpan }
+      }
+
+      this.asyncEnd(ctx) // triggers the setLLMObsTags
+    })
+
+    this.#toolCallIdsToName = {}
+    this.#availableTools = new Set()
+    toolCreationCh.subscribe(toolArgs => {
+      this.#availableTools.add(toolArgs)
+    })
+  }
+
+  findToolName (toolDescription) {
+    for (const availableTool of this.#availableTools) {
+      const description = availableTool.description
+      if (description === toolDescription) {
+        return availableTool.id
+      }
+    }
+  }
+
+  getLLMObsSpanRegisterOptions (ctx) {
+    const span = ctx.currentStore.span
+    const operation = getOperation(span)
+    const kind = SPAN_NAME_TO_KIND_MAPPING[operation]
+    if (!kind) return
+
+    return { kind, name: operation }
+  }
+
+  setLLMObsTags (ctx) {
+    const span = ctx.currentStore.span
+    if (!span) return
+
+    const operation = getOperation(span)
+    const kind = SPAN_NAME_TO_KIND_MAPPING[operation]
+    if (!kind) return
+
+    if (['embedding', 'llm'].includes(kind)) {
+      this._tagger._setTag(span, MODEL_NAME, getSpanTag(span, 'ai.model.id'))
+      this._tagger._setTag(span, MODEL_PROVIDER, getModelProvider(span))
+    }
+
+    switch (operation) {
+      case 'embed':
+      case 'embedMany':
+        this.setEmbeddingWorkflowTags(span)
+        break
+      case 'doEmbed':
+        this.setEmbeddingTags(span)
+        break
+      case 'generateObject':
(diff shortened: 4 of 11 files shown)


packages/dd-trace/src/id.js
```diff
@@ -182,6 +182,6 @@ function writeUInt32BE (buffer, value, offset) {
   buffer[0 + offset] = value & 255
 }
 
-module.exports = function createIdentifier (value, radix) {
+module.exports = Object.assign(function createIdentifier (value, radix) {
   return new Identifier(value, radix)
-}
+}, { Identifier })
```
sabrenner: this isn't necessary but is a helpful export for some `instanceof` checks we do in `llmobs/tagger.js` - seems all tests still pass as-is, if we're ok to do this!

packages/datadog-instrumentations/src/helpers/hooks.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
   '@smithy/smithy-client': () => require('../aws-sdk'),
   '@vitest/runner': { esmFirst: true, fn: () => require('../vitest') },
   aerospike: () => require('../aerospike'),
+  ai: { esmFirst: true, fn: () => require('../vercel-ai') },
```
rochdev: Why not just call the file `ai` if the module is `ai`?

packages/datadog-instrumentations/src/vercel-ai.js
```diff
@@ -0,0 +1,70 @@
...
+function wrapWithTracer (fn) {
+  return function () {
+    const options = arguments[0]
+    if (options.experimental_telemetry != null) return fn.apply(this, arguments)
+
+    options.experimental_telemetry = {
+      isEnabled: true,
+      // TODO(sabrenner): need to figure out how a user can configure this tracer
+      // maybe we advise they do this manually?
+      tracer: provider.getTracer()
```
rochdev: Ideally we wouldn't go through translation as that could have unintended side-effects and won't be as flexible. Is there a centralized place in the module that handles tracing? Maybe we could patch something there instead?

packages/dd-trace/src/id.js
rochdev: Let's not expose this as we should replace the current implementation with `BigInt` at some point, and any exposed internals will make that more difficult.

packages/dd-trace/src/llmobs/tagger.js
```diff
@@ -89,10 +91,15 @@ class LLMObsTagger {
 
     this._setTag(span, ML_APP, spanMlApp)
 
-    const parentId =
-      parent?.context().toSpanId() ??
-      span.context()._trace.tags[PROPAGATED_PARENT_ID_KEY] ??
-      ROOT_PARENT_ID
+    let parentId
+    if (parent instanceof Identifier) {
```
rochdev: Why is an identifier as the parent needed?

packages/dd-trace/src/llmobs/plugins/vercel-ai.js
```diff
@@ -0,0 +1,475 @@
...
+      const ctx = {
+        currentStore: { span: ddSpan }
+      }
+
+      const operation = getOperation(ddSpan)
+      const useLlmObsParent = SPANS_TO_USE_LLMOBS_PARENT.has(operation)
+
+      this.start(ctx, {
+        useLlmObsParent,
+        enterIntoLlmObsStorage: false
```
rochdev: Could these be replaced by a different class hierarchy or by runStores/bindStore?

packages/datadog-instrumentations/src/helpers/hooks.js
sabrenner: i thought maybe it wasn't specific enough and could be confusing initially 😅 :ai-intesifies:, but yes i can rename it!

packages/datadog-instrumentations/src/vercel-ai.js
sabrenner: hey do have a `getTracer` [function](https://github.com/vercel/ai/blob/main/packages/ai/core/telemetry/get-tracer.ts), but as this whole module is bundled it seems into one `dist/index.js` ([ref](https://www.npmjs.com/package/ai?activeTab=code)), it's not actually available on module exports for us to hook into.

packages/dd-trace/src/llmobs/tagger.js
sabrenner: oh i suppose i was being silly with what i was doing here haha. i can change this back to what i had before

packages/dd-trace/src/llmobs/plugins/vercel-ai.js
sabrenner: this was a bit tricky to me - because we're using the points of the otel spans, which are the start and end, it doesn't leave a lot of flexibility for running/binding stores over a context if the only points of the context we have are disjoint start and end events. but, i could be thinking about this wrong - lmk if so!

since the otel spans started by the vercel ai sdk are activated spans in the proper scope, i wanted to just add an option to use the APM span parentage as opposed to the LLMObs parentage, since these operations should be isolated from whatever manual LLMObs instrumentation users might do.

packages/dd-trace/src/llmobs/tagger.js
sabrenner: ah actually i had it this way because i don't think i have a reference to the parent span at the point of the vercel ai integration for llmobs parentage, just the current span, and on it, its parent id, which i opted to pass in here (but, correct me if i'm wrong about spans holding a reference to their entire parent span)

packages/datadog-instrumentations/src/vercel-ai.js
rochdev: Can we use orchestrion for this? I think we're not ready to use it for APM yet, but for LLMObs I think it's already used? If not I guess passing in our OTel tracer might be the only option, although I'd probably prefer to pass a "fake" tracer that only works for this integration at that point to avoid having to go through the OTel SDK.

sabrenner: yeah i thought about this too. aside from the node version problem (it's possible we have customers not above 20.6 yet in order to use orchestrion), it would just mean more on our side for maintaining patching for something they already instrument with the otel sdk themselves.

imho, i'm more in favor of just re-using their otel instrumentation and translating it as-is, although it wouldn't change the existing logic in the plugins too much, just our maintenance of patching in the instrumentation.

> I'd probably prefer to pass a "fake" tracer that only works for this integration at that point to avoid having to go through the OTel SDK

could you clarify this a bit - i think i'm confused, i'm not seeing the harm in passing ours in as is 😅 it's only currently passed in if there's no tracer passed in which, by default, there isn't anyways.

rochdev: What I mean is that for example if it expected to receive an object with `startSpan` on it, you could just pass an object created in the instrumentation that would just send a message on a channel like other integrations do instead of passing in a provider. Basically converting OTel to what the library should actually be doing itself 😅 Then the plugin can listen to that as normal without any hacks to make it work with OTel.

For example:

```
tracer: {
  startSpan ( ... ) {
    startCh.publish( ... )
  }
}
```

sabrenner: ok, i've tried this in https://github.com/DataDog/dd-trace-js/pull/5858/commits/a4014ccba099ae520932446476baab0741c67798. it works (yeah there are no tests but it looks the same w the dummy app i'm using for now 😆), but my main concern is if someone wants to pass in their own dd-trace provider otel tracer (as noted is possible in the PR description), we'll need to still handle those events (or just always override the tracer option, which i don't think is a good solution).

at that point, instead of handling the case where they pass in their own dd-trace otel tracer, and us using this custom one in instrumentation, it would be better to just keep the implementation before this commit. 

let me know what you think - i actually do agree it's easier for parent context setting/manipulation if we do it ourselves as i implemented in the commit, but i wonder if we should have consideration for folks passing in their own dd-trace otel tracer.

sabrenner: to further that point a bit, the dummy 'span' passed into the function in the instrumentation for `startActiveSpan` doesn't have all of the otel methods, just the ones used by vercel-ai. should they use one not on there, we break. imho any solution there that's like a) subclass the span from a noop otel span kind or b) just have all the methods, points us back to just listening for otel spans in the first place 😅

sabrenner: > my main concern is if someone wants to pass in their own dd-trace provider otel tracer (as noted is possible in the PR description), we'll need to still handle those events (or just always override the tracer option, which i don't think is a good solution)

i've pushed up one more change in https://github.com/DataDog/dd-trace-js/pull/5858/commits/407d6dc1aec6e8ebe65559e3adc10558eb30dfc1 that addresses this concern, which seems a lot cleaner. but i guess i still have a question on reusability of otel tracer spans from integrations we want to have. if we add more integrations that have their own otel tracing under the hood, is this a pattern we want to do every time?

packages/datadog-instrumentations/src/ai.js
```diff
@@ -0,0 +1,124 @@
...
+      end () {},
+      setAttributes () { return this },
+      addEvent () { return this },
+      recordException () { return this },
+      setStatus () { return this }
+    }
+
+    return fn(span)
+  }
+}
```
sabrenner: i guess this could be extracted out into an otel noop tracer that could be shared

rochdev: Why even define a noop tracer that will ultimately be patched in the first place? Why not just returning a fake span directly?

packages/dd-trace/src/llmobs/tagger.js
rochdev: In general you should have access to the span you need, otherwise it's probably not your parent. Why would that not be the case here?

packages/datadog-instrumentations/src/vercel-ai.js
rochdev: _Note: this was written a month ago but I forgot to submit the review. Most of the points have been addressed but posting anyway for completeness._

> it would be better to just keep the implementation before this commit

The problem with that is that it relies on code that may not even be available in the future. As we start separating the SDK from the APIs, at some point the API will be an external module that is not included with dd-trace, so we should reduce our reliance on using public APIs internally (right now the Datadog API is an exception to that rule but that's because the internal API matches the public API right now, but that will also eventually change). This is not a problem when a provider is passed explicitly since at that point the user explicitly decided to use that API directly, with a version they decided.

> we'll need to still handle those events

Will we? If you pass a provider to a library then I would expect that we don't need an integration at all at that point and it will just be used like with any other provider usage.

> but i guess i still have a question on reusability of otel tracer spans from integrations we want to have

I guess we could have a generic plugin-system-specific implementation that just forwards calls to DC given a specific channel name.

> if we add more integrations that have their own otel tracing under the hood, is this a pattern we want to do every time?

This is actually an anti-pattern on their end and we should work with both OpenTelemetry and library authors to avoid this going forward. They should instead provide tracing channels that both an external OpenTelemetry implementation can listen to, as well as any APM vendors or any other subscriber interested in the data.

sabrenner:
for reviewers (when i open this up): currently this is all one PR - apm + llmobs. just the way i did the poc + clean up, but if this pr is too big i'm happy to separate it out!

packages/datadog-instrumentations/src/ai.js
sabrenner: i've cleaned this up, although it's still local to the instrumentation, but i'm gonna resolve for now anyways since it's only used in this instrumentation. if we need a no-op dummy otel tracer for other instrumentations down the line i think we can refactor then.

packages/datadog-instrumentations/src/ai.js
```diff
@@ -0,0 +1,162 @@
...
+
+const noopTracer = {
+  startActiveSpan () {
+    const fn = arguments[arguments.length - 1]
+
+    const span = {
+      spanContext () {
+        return { traceId: '', spanId: '', traceFlags: 0 }
+      },
+      setAttribute () {
```
rochdev: Nit: these could be one-liners instead of three-liners to to take less space.

packages/datadog-instrumentations/src/ai.js
```diff
@@ -0,0 +1,162 @@
...
+  })
+
+  Object.defineProperty(tracer, Symbol.for('_dd.wrapped'), { value: true })
+}
+
+function wrapWithTracer (fn) {
+  return function () {
+    const options = arguments[0]
+
+    options.experimental_telemetry ??= { isEnabled: true, tracer: noopTracer }
```
rochdev, line 101: If `noopTracer` is only ever set here, and then immediately patched, any reason it's even needed in the first place?

packages/datadog-plugin-ai/test/index.spec.js
```diff
@@ -0,0 +1,277 @@
...
+    before(() => agent.load('ai'))
+
+    after(() => agent.close({ ritmReset: false }))
+
+    beforeEach(function () {
+      if (semifies(moduleVersion, '<4.0.2') && NODE_MAJOR < 22) {
+        /**
+         * Resolves the following error:
+         *
+         * Error [ERR_REQUIRE_ESM]: require() of ES Module  from ... not supported.
```
rochdev: Not sure I understand this comment. If it's supported in CommonJS (which should be true according to the instrumentation) it should means we can also import it here no?

packages/dd-trace/src/llmobs/plugins/ai.js
```diff
@@ -0,0 +1,469 @@
...
+  'stop_sequences'
+])
+
+/**
+ * Get the span tags from the context (either the attributes or the span tags).
+ *
+ * @param {Record<string, any>} ctx
+ * @returns {Record<string, any>}
+ */
+function getSpanTags (ctx) {
```
rochdev: This file would be easier to read if all these helpers outside of the class would be in a separate helpers file.

packages/dd-trace/src/llmobs/plugins/ai.js
```diff
@@ -0,0 +1,469 @@
...
+    const settingValue = tags[tag]
+    metadata[settingKey] = settingValue
+  }
+
+  return Object.keys(metadata).length ? metadata : null
+}
+
+class VercelAILLMObsPlugin extends BaseLLMObsPlugin {
+  static get id () { return 'ai' }
+  static get integration () { return 'vercel-ai' } // for LLMObs telemetry - "vercel-ai" makes more sense than "ai"
```
rochdev: Not sure about this one. If the package is `ai`, the author of the package doesn't matter. It would be a different story if it was `@vercel/ai` but it's not. If we do want to name it differently (which we definitely can, plugin names are technically arbitrary) then let's rename it everywhere. I personally prefer to keep it standardized as the other plugins, but consistency within a plugin is even more important.

packages/dd-trace/test/llmobs/util.js
```diff
@@ -190,11 +194,74 @@ function fromBuffer (spanProperty, isNumber = false) {
...
+    throw new TypeError(
+      '`loadOptions.llmobs` is required when using `useLlmobs`'
+    )
+  }
+
+  let agentPromise
+  let llmobsEvents = []
+
+  before(() => {
+    sinon.stub(LLMObsSpanWriter.prototype, 'append').callsFake(event => {
```
rochdev: Why is mocking needed here? Usually for plugin tests we go through the entirety of the real code.

packages/datadog-instrumentations/src/ai.js
sabrenner: it's more that we default the tracer here to the noop tracer, and then in `wrapTracer` we generically patch any tracer passed in (say the user of the vercel ai sdk passes in their own otel compatible tracer, we don't want to override it with our dummy tracer). but lmk if i'm not understanding!

packages/datadog-plugin-ai/test/index.spec.js
sabrenner: i think it's not supported in commonjs for the restrictions specified in the condition below (early version 4 of vercel ai sdk + node < 22 not supported. i get this when just running a dummy script requiring `ai` with `ai@4.0.0.0` and Node 20

packages/dd-trace/src/llmobs/plugins/ai.js
sabrenner: sure - this is just for llmobs telemetry, so it doesn't really matter. can rename to `ai` for internal consistency

packages/dd-trace/test/llmobs/util.js
sabrenner: this is how we do the rest of the tests as well (for our other llmobs auto-instrumentations). lemme see if i can do something better for this, maybe by adding a handler function in our test agent. we just want to distinguish the llmobs payloads from the apm ones.

packages/datadog-instrumentations/src/ai.js
sabrenner: > Why not just returning a fake span directly?

mostly because we could be patching an actual otel-compatible tracer that someone is already using. i agree if we were just concerned about a dummy tracer/spans, then yeah we could patch in-place with respect to the dummy tracer, but i did it this way bc someone could actually be using a real tracer. lmk if that answers your question!

packages/datadog-plugin-ai/test/index.spec.js
sabrenner: if that makes sense, i can update the comment so it's not as confusing 😅

packages/dd-trace/test/llmobs/util.js
sabrenner: i've resolved this by using our fake "test agent" locally to assert traces were sent, in this commit (https://github.com/DataDog/dd-trace-js/pull/5858/commits/c03fc82a45cc885a9b382c2c871037c8fb504abe). to help with this, i've also added some configurability for llmobs flushing (we flush on an interval every second, which is annoying for tests), and that was done in this commit (https://github.com/DataDog/dd-trace-js/pull/5858/commits/0f1acf67abf2f69fe023f2d898bb1702190f13de) - let me know what you think (if it looks good, i'll have another ticket to use this for our other tests too)!

.github/actions/testagent-llmobs/start/action.yml
```diff
@@ -0,0 +1,8 @@
+name: "Start the APM Test Agent"
+description: "Starts the APM Test Agent image with environment, with LLM Observability cassettes attached."
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+    - run: docker compose up -d testagent-llmobs-vcr || docker compose up -d testagent-llmobs-vcr
```
sabrenner: i duped this action because i needed a way to use this testagent-llmobs-vcr image, which is the same as the testagent image we already have but mounts the locally added cassettes to mock responses.

doing this generically for testagent failed openai tests, which do not have cassettes locally. i plan on coalescing this back into the `testagent` fixture once i move those cassettes to be local as well

packages/datadog-instrumentations/src/vercel-ai.js
sabrenner: yeah i think this answers all of my questions as well, but i did implement it the other way as stated. will resolve this for now but lmk if there are points in here i missed addressing in code!

.github/actions/testagent-llmobs/start/action.yml
sabrenner: i will actually resolve this after merging #6152

packages/datadog-instrumentations/src/ai.js
rochdev: Oh I didn't realize that we may not be patching the no-op tracer and a different one may be passed in, then it makes sense!

packages/datadog-plugin-ai/test/index.spec.js
rochdev: I guess my concern is that if it's supported, it should work in tests, and if it's not supported, then we should change the range. But I may still not be grasping the issue correctly 😅

packages/dd-trace/test/llmobs/util.js
```diff
@@ -221,37 +226,37 @@ function useLlmobs ({
...
+      return llmobsTraces
+        .flatMap(trace => trace)
+        .map(trace => trace.spans[0])
+        .sort((a, b) => a.start_ns - b.start_ns)
+    })
   })
 
   after(() => {
-    LLMObsSpanWriter.prototype.append.restore()
+    process.removeAllListeners()
```
rochdev: Is this safe? I think the test runner might have listeners on the process.

packages/dd-trace/test/llmobs/util.js
rochdev: LGTM!

Scalahansolo:
Hey hey crew. Fwiw... I've been silently following this PR in wait to try and weave this into the Motion codebase to get our stack onto LLMObservability. Currently we don't use dd-trace, and just export spans / traces through otel. I imagine once this lands I'll need to pull in dd-trace for LLM Observability. 

Would it be helpful at all if I were to test this (somehow?) and report back here?

packages/dd-trace/test/llmobs/util.js
sabrenner: oh it might not be lol but it doesn't currently seem to be causing any problems. i just had it there because it complained about a max listeners thing 😅 but tbh i'll just remove it

packages/datadog-plugin-ai/test/index.spec.js
sabrenner: yeah it's just not supported, but i did it like this because we run CI against oldest node version supported, which hits this condition. if i could, i would move it out to a version variable outside the scope of the describe blocks, but since it's dependent on the `ai` version, i still think this is the best spot for it.

```bash
> npm install ai@4.0.0
> nvm use 20
> node -e "require('ai')"
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/sam.brenner/Development/ml-obs/sandboxes/vercel-ai-tests/node_modules/@ai-sdk/ui-utils/node_modules/nanoid/non-secure/index.js from /Users/sam.brenner/Development/ml-obs/sandboxes/vercel-ai-tests/node_modules/@ai-sdk/ui-utils/node_modules/@ai-sdk/provider-utils/dist/index.js not supported.
```
unrelated to `dd-trace`

sabrenner: oh nvm i should still be able to just have a version range anyhow i think. let me push that up with a comment

packages/dd-trace/test/llmobs/util.js
sabrenner: removed 9fc3d13f8c8fedcd7fd850444b98c70877c9466a 👍

packages/datadog-plugin-ai/test/index.spec.js
sabrenner: lmk if this is any better! 9fc3d13f8c8fedcd7fd850444b98c70877c9466a

rochdev (approved):

sabrenner:
Hi @ctamulonis-glossy @gabrielsch! Yes we are looking to get this merged soon (within the next couple of weeks), and it should support AI SDK v5 😄

Kyle-Verhoog (dismissed):

rochdev (approved):

PR #5859, opened 2025-06-09
Enable log injection by default for structured loggers

ida613:
### What does this PR do?
Set `config.logInjection` default to `structured`, in which case log injection is enabled for structured loggers and disabled for unstructured loggers.
When `config.logInjection=true`, log injection is enabled for both structured and unstructured loggers
When `config.logInjection=false`, log injection is disabled for both structured and unstructured loggers

This change breaks the configuration default, so please let me know if there are any concerns regarding how this change is handled!

### Plugin Checklist

- [x] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/dd-trace/src/plugins/log_plugin.js
```diff
@@ -3,6 +3,7 @@
 const { LOG } = require('../../../../ext/formats')
 const Plugin = require('./plugin')
 const { storage } = require('../../../datadog-core')
+const structuredLoggers = new Set(['bunyan', 'pino', 'winston'])
```
rochdev: A base class shouldn't know about its subclasses. I would say to either create a `StructuredLogPlugin` base class that these 3 can inherit from instead, or add a property like `this.structured = true` that could be added that the 3 classes could set.

rochdev (changes requested):

packages/dd-trace/src/plugins/log_plugin.js
ida613: thank you Roch! updated

packages/dd-trace/src/plugins/log_plugin.js
```diff
@@ -43,12 +43,21 @@ module.exports = class LogPlugin extends Plugin {
...
     })
+    this.structured = false // default unstructured logger
+  }
+
+  _injectLogs (config) {
+    if (this.structured) {
+      return (config.logInjection === true || config.logInjection === 'structured')
+    }
+
+    return config.logInjection === true
```
rochdev: The whole function body can be replaced by `config.logInjection === true || (this.structured && config.logInjection === 'structured')`

packages/dd-trace/src/plugins/log_plugin.js
```diff
@@ -43,12 +43,21 @@ module.exports = class LogPlugin extends Plugin {
       this.tracer.inject(span, LOG, holder)
       arg.message = messageProxy(arg.message, holder)
     })
+    this.structured = false // default unstructured logger
+  }
+
+  _injectLogs (config) {
```
rochdev: This is not injecting so the method name is confusing. Maybe something like `_shouldInject`?

packages/datadog-plugin-winston/src/index.js
```diff
@@ -3,6 +3,11 @@
 const LogPlugin = require('../../dd-trace/src/plugins/log_plugin')
 
 class WinstonPlugin extends LogPlugin {
+  constructor (...args) {
+    super(...args)
+    this.structured = true
```
rochdev: Let's use a static getter instead as this is really a property of the class and not the instance (we can't use a static property as that's not supported by Webpack 4).

rochdev: Actually at that point we might as well just use a new subclass that overrides `configure` instead so this property is not needed.

packages/dd-trace/src/plugins/log_plugin.js
BridgeAR: Can we please change it to a private method `#isEnabled()` and move the overall enabled logic in here?

packages/dd-trace/src/plugins/log_plugin.js
```diff
@@ -43,12 +43,21 @@ module.exports = class LogPlugin extends Plugin {
       this.tracer.inject(span, LOG, holder)
       arg.message = messageProxy(arg.message, holder)
     })
+    this.structured = false // default unstructured logger
```
BridgeAR: This could be a private property if I am not mistaken :)

That way it would not potentially be exposed to the user.

packages/datadog-plugin-winston/src/index.js
BridgeAR: Can we actually pass through structured as additional part of the args? I believe so?

packages/dd-trace/src/plugins/log_plugin.js
ida613: I added the `StructuredLogPlugin` [subclass](https://github.com/DataDog/dd-trace-js/blob/99496e825ed55461dc37bdea4ff471393c7d75cd/packages/dd-trace/src/plugins/structured_log_plugin.js) which has its own implementation of `_shouldInjectLogs`. If I encapsulate the entire `isEnabled` logic in one function, the subclass won't have its own implementation of `_shouldInjectLogs`, which I'm okay with. Do you still think we should get rid of `_shouldInjectLogs` and go with `isEnabled`? I can also add an `UnstructuredLogPlugin` subclass, it's just that no plugin will be extending it at the moment.

packages/datadog-plugin-winston/src/index.js
rochdev: A log plugin class is either structured or not, it cannot be configured one way or the other so an argument doesn't quite make sense for this IMO.

packages/dd-trace/src/plugins/log_plugin.js
ida613: good call! updated

packages/dd-trace/src/plugins/log_plugin.js
ida613: updated!

packages/dd-trace/src/plugins/log_plugin.js
```diff
@@ -43,12 +43,17 @@ module.exports = class LogPlugin extends Plugin {
       this.tracer.inject(span, LOG, holder)
       arg.message = messageProxy(arg.message, holder)
     })
+    this._structured = false // default unstructured logger
```
rochdev: This is no longer needed with the subclass as it has its own `shouldInject` method.

BridgeAR (changes requested):
Winston does not always produce structured logs. We need to check the configuration and decide upon that or check each individual message if it's an object or not.

ida613:
> Winston does not always produce structured logs. We need to check the configuration and decide upon that or check each individual message if it's an object or not.

Thank you Ruben! I added [this commit](https://github.com/DataDog/dd-trace-js/pull/5859/commits/ab605e8ac1ccadd908d6fa9b315ef127819d3c88) to check if each message is structured for winston. This might cause some CI failures but you can treat it as a draft for now. Please let me know if you have any feedback!

ida613:
I've removed winston for now for a separate PR

packages/dd-trace/src/plugins/structured_log_plugin.js
```diff
@@ -0,0 +1,13 @@
+'use strict'
+
+const LogPlugin = require('./log_plugin')
+
+module.exports = class StructuredLogPlugin extends LogPlugin {
+  static get _structured () {
+    return true
+  }
+
```
BridgeAR:
```suggestion
```

packages/dd-trace/src/plugins/structured_log_plugin.js
```diff
@@ -0,0 +1,13 @@
...
+  _shouldInjectLogs (config) {
+    return config.logInjection === true || config.logInjection === 'structured'
```
BridgeAR:
```suggestion
  _isEnabled(config) {
    return super._isEnabled(config) || config.enabled && config.logInjection === 'structured'
```

packages/dd-trace/src/plugins/log_plugin.js
```diff
@@ -45,10 +45,14 @@ module.exports = class LogPlugin extends Plugin {
...
 
+  _shouldInjectLogs (config) {
+    return config.logInjection === true
+  }
+
   configure (config) {
     return super.configure({
       ...config,
-      enabled: config.enabled && (config.logInjection || config.ciVisAgentlessLogSubmissionEnabled)
+      enabled: config.enabled && (this._shouldInjectLogs(config) || config.ciVisAgentlessLogSubmissionEnabled)
```
BridgeAR:
```suggestion
      enabled: this._isEnabled(config)
```

packages/dd-trace/src/plugins/log_plugin.js
```diff
@@ -45,10 +45,14 @@ module.exports = class LogPlugin extends Plugin {
     })
   }
 
+  _shouldInjectLogs (config) {
+    return config.logInjection === true
```
BridgeAR:
```suggestion
  _isEnabled (config) {
    return config.enabled && (config.logInjection === true || config.ciVisAgentlessLogSubmissionEnabled)
```

integration-tests/pino.spec.js
```diff
@@ -32,6 +32,27 @@ describe('pino test', () => {
...
+      assert.containsAllKeys(stdoutData, ['dd'])
+      assert.containsAllKeys(stdoutData.dd, ['trace_id', 'span_id'])
+      assert.strictEqual(
+        stdoutData.dd.trace_id,
+        stdoutData.custom.trace_id
+      )
+      assert.strictEqual(
+        stdoutData.dd.span_id,
+        stdoutData.custom.span_id
+      )
```
BridgeAR: Nit (I am currently working on removing chai from all tests)

The helper has to be required first, if used.

```suggestion
      assertObjectContains(stdoutData, {
        dd: {
          trace_id: stdoutData.custom.trace_id,
          span_id: stdoutData.custom.span_id,
        },
        custom: {
          trace_id: stdoutData.dd.trace_id,
          span_id: stdoutData.dd.span_id,
        },
      })
```

integration-tests/automatic-log-submission.spec.js
```diff
@@ -174,7 +174,7 @@ describe('test visibility automatic log submission', () => {
         )
         childProcess.on('exit', () => {
           assert.include(testOutput, 'Hello simple log!')
-          assert.notInclude(testOutput, 'span_id')
+          assert.include(testOutput, 'span_id')
```
BridgeAR, line 177: The test message is now a bit confusing, since this is actually the part that should verify that the logs are not submitted.
The test uses winston, so that's why it is now automatically enabled and for the test to behave as expected, we'd have to use a different logger instead.

It is the logger being used in `automatic-log-submission-test.js`. That requires `./logger` and that instruments winston. That would have to change.

juan-fernandez: if this PR is merged as is, without changes to the test optimization logic, this test would not make sense, so it'd have to be removed

juan-fernandez: I think we can leave this change and the name of the test as is, but we need to add some logic to the test. Since now what we're checking is that the logs are _not_ submitted, we'd have to *add* something like:

```javascript
              const logsPromise = receiver
                .gatherPayloadsMaxTimeout(({ url }) => url.includes('/api/v2/logs'), payloads => {
                  if (payloads.length > 0) {
                    throw new Error('Unexpected logs')
                  }
                }, 5000)
```

ida613: updated! let me know if it looks alright :)

juan-fernandez: the problem with this is that the error can be thrown after the test finishes. Let's do this:

```javascript
       childProcess.on('exit', () => {
          assert.include(testOutput, 'Hello simple log!')
          assert.include(testOutput, 'span_id')
          logsPromise.then(() => {
             done()
          }).catch(done)
        })
        ...
        const logsPromise = receiver.gatherPayloadsMaxTimeout(({ url }) => url.includes('/api/v2/logs'), payloads => {
       ...
```

ida613: thank you Juan! updated

BridgeAR (approved):
LGTM. Thanks for the follow-up!

packages/datadog-plugin-winston/src/index.js
BridgeAR: Winston actually supports both :D
But our plugin can't handle that

rochdev: Does it though? I thought the end result of our investigation was that it's always structured internally?

PR #5861, opened 2025-06-10
[DI] Support custom upload interval

watson:
### What does this PR do?

Add a new config option to control the upload interval used by Dynamic Instrumentation (DI). By default, data collected by DI is buffered in memory, and only sent to the agent once every second (if there is data in the buffer).
    
The new config option, allows the timeout to be controlled by the following environment variable:
    
```
DD_DYNAMIC_INSTRUMENTATION_UPLOAD_INTERVAL_SECONDS
```
    
There is also a code config alias:
    
```js
{ dynamicInstrumentation: { uploadIntervalSeconds: 1 } }
```
    
The value of this config option is parsed as a float.

### Motivation

The main use-case for this config option is to make tests run faster, as they don't have to wait for the one second timeout. It's not expected that end-users would need to modify the default value.

As a result of this, the Debugger tests now run over a minute faster in CI:

| Before | After |
|--------|------|
| image | image |

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes

- Added the new config option to allow list in `dd-go`: https://github.com/DataDog/dd-go/pull/187337
- Should this be `semver-minor`?
- Tip: View changes with whitespace diff disabled, for an easier review time


watson:
@BridgeAR I don't think we plan to document it. But we have agreed across all the tracers to implement it. At least the environment variable version of it. The config option version is just a Node.js convention.

rochdev (approved):

watson:
Backstory: Some of the other tracers already have config options to control this. They are called something a bit different and we're trying to standardize this and align on what it controls (e.g. some tracers, like ours, have two buffers, others share one buffer).

I agree: When the config option present, it's public. What I meant by potentially not documenting it, I was referring to your comment about not overwhelming people with that many knobs. I feel like we could have config options that people can use if they really want to, but that we don't technically have do document if we feel it would just confuse the vast majority of users. Or we could put it in an advanced section of the docs.

PR #5864, opened 2025-06-10
test: speed up DI integration tests even more

watson:
### What does this PR do?

Don't wait for script loading to stabilize in Dynamic Instrumentation integration tests.
    
This might make them temporarily set the test breakpoint in the wrong location, but it will quickly get moved to the right location and the test should still trigger as expected.

Debugger CI run performance improvement:

| Before | After |
|--------|------|
| image | image |

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/integration-tests/debugger/utils.js
+++ b/integration-tests/debugger/utils.js
@@ -97,6 +97,7 @@ function setup ({ env, testApp, testAppSource, dependencies } = {}) {
       env: {
         DD_DYNAMIC_INSTRUMENTATION_ENABLED: 'true',
         DD_DYNAMIC_INSTRUMENTATION_UPLOAD_INTERVAL_SECONDS: '0',
+        _DD_INTERNAL_DYNAMIC_INSTRUMENTATION_SCRIPT_LOADING_STABILIZATION_DELAY_MS: '0',
         DD_TRACE_AGENT_PORT: t.agent.port,
         DD_TRACE_DEBUG: process.env.DD_TRACE_DEBUG, // inherit to make debugging the sandbox easier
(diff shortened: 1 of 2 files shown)


rochdev (commented):
Any way to do this without creating such a variable?

rochdev:
Actually what does this even do exactly? According to the description it seems like the right breakpoint will eventually happen anyway, so why have the delay to begin with?

watson:
> Any way to do this without creating such a variable?

I haven't been able to think of anything, as it has to be configurable from outside the sandbox 🤷 

> Actually what does this even do exactly? According to the description it seems like the right breakpoint will eventually happen anyway, so why have the delay to begin with?

It's to avoid the false positive. Both the data collected would not be what was expected and confuse users, and the incorrectly set breakpoint might be in a hot code-path, which could be a big problem (worst case scenario)

rochdev:
Seems to me like a design flaw more than something that should be configurable. I'm going to agree with @BridgeAR here. This is equivalent to adding a sleep which wouldn't fix the issue but only make it less likely, and I think we should figure out an actual fix for the underlying issue instead.

watson:
Haha, yeah I knew this change was going to be controversial - and to be honest I don't really like it myself either. But I'm really happy for the discussion. I'll mark this as draft for now and think a bit more about it. Also, it's not super important to get this merged. It would just be good DevEx ☺️

watson:
I've decided it's not worth it, as it's better to test closer to the real world experience

PR #5867, opened 2025-06-10
[test optimization] Review changes for #4581

juan-fernandez:


integration-tests/ci-visibility/run-jest.js
```diff
@@ -7,8 +7,7 @@ const options = {
   testRegex: process.env.TESTS_TO_RUN ? new RegExp(process.env.TESTS_TO_RUN) : /test\/ci-visibility-test/,
   coverage: !process.env.DISABLE_CODE_COVERAGE,
   runInBand: true,
-  shard: process.env.TEST_SHARD || undefined,
-  testSequencer: process.env.CUSTOM_TEST_SEQUENCER
```
juan-fernandez, line 11: not used anymore

packages/datadog-instrumentations/src/jest.js
```diff
@@ -90,6 +89,7 @@ const originalHookFns = new WeakMap()
 const retriedTestsToNumAttempts = new Map()
 const newTestsTestStatuses = new Map()
 const attemptToFixRetriedTestsStatuses = new Map()
+const wrappedWorkers = new WeakSet()
```
juan-fernandez, line 92: instead of using `_ddWorker`, this is better (we don't pollute the original object)

packages/datadog-instrumentations/src/jest.js
```diff
@@ -598,15 +598,15 @@ function getWrappedEnvironment (BaseEnvironment, jestVersion) {
       }
     }
 
-    async teardown () {
-      if (satisfies(jestVersion, '>=30.0.0') && this._globalProxy) {
+    teardown () {
+      if (this._globalProxy?.propertyToValue) {
```
juan-fernandez, line 602: checking this property should be enough, instead of having to check the version

packages/datadog-instrumentations/src/jest.js
```diff
@@ -663,308 +663,305 @@ function getWrappedScheduleTests (scheduleTests, frameworkVersion) {
...
-    }
-    // If suite skipping is enabled and has not filtered skippable suites yet, we'll attempt to do it
-    const scheduler = await oldCreateTestScheduler.apply(this, arguments)
-    shimmer.wrap(scheduler, 'scheduleTests', scheduleTests => getWrappedScheduleTests(scheduleTests, frameworkVersion))
-    return scheduler
-  }
-  testSchedulerPackage.createTestScheduler = newCreateTestScheduler
-  return testSchedulerPackage
-})
+function searchSourceWrapper (searchSourcePackage, frameworkVersion) {
```
juan-fernandez, line 666: no logic change here: just moving `searchSourceWrapper` up

packages/datadog-instrumentations/src/jest.js
```diff
@@ -663,308 +663,305 @@ function getWrappedScheduleTests (scheduleTests, frameworkVersion) {
...
-        earlyFlakeDetectionNumRetries = libraryConfig.earlyFlakeDetectionNumRetries
-        earlyFlakeDetectionFaultyThreshold = libraryConfig.earlyFlakeDetectionFaultyThreshold
-        isKnownTestsEnabled = libraryConfig.isKnownTestsEnabled
-        isTestManagementTestsEnabled = libraryConfig.isTestManagementEnabled
-        testManagementAttemptToFixRetries = libraryConfig.testManagementAttemptToFixRetries
-        isImpactedTestsEnabled = libraryConfig.isImpactedTestsEnabled
-      }
-    } catch (err) {
-      log.error('Jest library configuration error', err)
+function getCliWrapper (isNewJestVersion) {
```
juan-fernandez, line 710: instead of having `cliWrapper` check the jest version, we have `getCliWrapper` that receives whether the version is new or not, and wrap `SearchSource` depending on that

packages/datadog-instrumentations/src/jest.js
```diff
@@ -988,6 +985,55 @@ function coverageReporterWrapper (coverageReporter) {
...
+    if (!shardedTests.length || !isSuitesSkippingEnabled || !skippableSuites.length) {
+      return shardedTests
+    }
+    const [test] = shardedTests
+    const rootDir = test?.context?.config?.rootDir
+
+    return applySuiteSkipping(shardedTests, rootDir, frameworkVersion)
+  })
+  return sequencerPackage
+})
```
juan-fernandez, line 1035: no changes here, just moving things around a bit

packages/datadog-instrumentations/src/jest.js
```diff
@@ -1277,77 +1279,81 @@ addHook({
...
+          ...config.testEnvironmentOptions,
+          _ddKnownTests: suiteKnownTests,
+          _ddTestManagementTests: suiteTestManagementTests,
+          _ddModifiedTests: suiteModifiedTests
+        }
+      }
+    }
+    return send.apply(this, arguments)
   }
-  return _onMessage.apply(this, arguments)
```
juan-fernandez, line 1336: `onMessageWrapper` and `sendWrapper` are the same, they're just `function`s now

packages/datadog-instrumentations/src/jest.js
```diff
@@ -1277,77 +1279,81 @@ addHook({
...
+    shimmer.wrap(arguments[0], 'onStart', onStart => function (worker) {
+      if (worker && !wrappedWorkers.has(worker)) {
+        shimmer.wrap(worker._child, 'send', sendWrapper)
+        shimmer.wrap(worker, '_onMessage', onMessageWrapper)
+        worker._child.on('message', worker._onMessage.bind(worker))
+        wrappedWorkers.add(worker)
+      }
+      return onStart.apply(this, arguments)
+    })
+    return enqueue.apply(this, arguments)
```
juan-fernandez, line 1355: changes:
* `function enqueueWrapper` instead of `const enqueueWrapper` (just more consistent and it has other advantages)
* use `arguments` instead of `...args`
* use `wrappedWorkers` instead of `_ddWorker`
* use `shimmer.wrap` whenever possible instead of directly wrapping

packages/datadog-instrumentations/src/jest.js
```diff
@@ -598,15 +598,15 @@ function getWrappedEnvironment (BaseEnvironment, jestVersion) {
       }
     }
 
-    async teardown () {
-      if (satisfies(jestVersion, '>=30.0.0') && this._globalProxy) {
+    teardown () {
```
Mariovido, line 601: Is there any reason why you removed the `async`? As far as I know the original function was an `async`

Mariovido (approved):
LGTM, just a small comment

packages/datadog-instrumentations/src/jest.js
juan-fernandez: it doesn't really matter as long as the function returns a promise, which it does. It signals your intention to use await, but we don't, so it's not necessary

PR #5871, opened 2025-06-10
telemetry: add transmit:false flag for logs

tlhunter:
### What does this PR do?
- allows log.error and log.warn calls to optionally disable telemetry

### Motivation
- host connection errors are useful for users but not useful for us to debug


packages/dd-trace/src/exporters/agent/writer.js
```diff
@@ -42,7 +42,7 @@ class Writer extends BaseWriter {
       startupLog({ agentError: err })
 
       if (err) {
-        log.error('Error sending payload to the agent (status code: %s)', err.status, err)
+        log.error('Error sending payload to the agent (status code: %s)', err.status, { transmit: false }, err)
```
watson: To reduce the risk of false positives, shouldn't we use a symbol?

```suggestion
        log.error('Error sending payload to the agent (status code: %s)', err.status, { [Symbol.for('datadog:transmit')]: false }, err)
```

BridgeAR: We could even move this to the last argument and just check if it contains a datadog internal symbol. That way it's easier to keep everything else as is.

tlhunter: ~~I suppose I could make the last argument _just_ a symbol (not an object with a symbol property). Then check if the last argument is a symbol and narrow it down to the transmit skip symbol.~~

Actually that defeats the purpose of a general config object that can have further properties attached.

tlhunter: I could export a `log.Config` class and then do an instanceof and that Config class can hold all possible config. Furthermore I wouldn't need to instanciate a bunch at runtime, just once per file that needs to do special logging.

packages/dd-trace/src/log/index.js
```diff
@@ -15,7 +15,12 @@ const config = {
   logLevel: 'debug'
 }
 
+const mute = new LogConfig(false)
+
 const log = {
+  LogConfig,
+  MUTE: mute,
```
watson: The word "mute" will easily be misinterpreted, as we're not muting the log in general, but just not sending it to telemetry. A more descriptive name would be:

```suggestion
  NO_TELEMETRY: mute,
```

packages/dd-trace/src/telemetry/logs/log-collector.js
```diff
@@ -27,7 +27,7 @@ function createHash (logEntry) {
 }
 
 function isValid (logEntry) {
-  return logEntry?.level && logEntry.message
+  return logEntry?.level && logEntry.message && logEntry.transmit !== false
```
watson: Maybe I'm reading the code wrong, but isn't `transmit` only a property of the `LogConfig` instance? A `Log` instance would have `sendViaTelemetry` set instead, right?

```suggestion
  return logEntry?.level && logEntry.message && logEntry.sendViaTelemetry !== false
```

packages/dd-trace/test/telemetry/logs/log-collector.spec.js
```diff
@@ -49,6 +49,12 @@ describe('telemetry log collector', () => {
       ).to.be.false
     })
 
+    it('should not store logs with transmit flag set to false', () => {
+      expect(logCollector.add({ message: 'Error 1', level: 'ERROR' })).to.be.true
+      expect(logCollector.add({ message: 'Error 2', level: 'ERROR', transmit: true })).to.be.true
+      expect(logCollector.add({ message: 'Error 3', level: 'ERROR', transmit: false })).to.be.false
```
watson: If I'm correct about my question about `transmit` vs `sendViaTelemetry` above, this test is not actually testing what you think it is. In that case it would be more robust to not just set the flag manually like you do here, but to at least go through the `Log` constructor.

packages/dd-trace/src/exporters/agent/writer.js
```diff
@@ -42,7 +42,7 @@ class Writer extends BaseWriter {
       startupLog({ agentError: err })
 
       if (err) {
-        log.error('Error sending payload to the agent (status code: %s)', err.status, err)
+        log.error('Error sending payload to the agent (status code: %s)', err.status, err, log.NO_TELEMETRY)
```
bouwkast: Awesome thanks!

rochdev: This pushes knowledge of telemetry reporting (or any logging subscriber in general) to the logging client. I'd prefer if it can be handled downstream, and if not, a more generic flag would make more sense, for example `log.SENSITIVE` which subscribers can then use to decide whether to report sensitive data.

rochdev: Actually, it sounds like all we need to do is make sure the message is the non-interpolated string for telemetry, without parsing the parameters. This should be the case for _all_ logs, so we don't need a flag.

bouwkast: This one specifically is not actionable by us in the library code (for the most part) and is usually a configuration or transient connection issue. At the time, it was taking up a large portion of our log quota. I've put a large exclusion filter on it so that we can ingest other logs.

rochdev: My understanding from yesterday's discussion is that we should disable all errors for now anyway to avoid issues, and then come up with a new strategy entirely to be able to aggregate errors. I think we should simply do that for now.

rochdev: > This one specifically is not actionable by us in the library code (for the most part) and is usually a configuration or transient connection issue.

That's an interesting one actually, as we currently don't really have a distinction between errors caused by users or caused by us. Might be something to look into as well.

packages/dd-trace/src/telemetry/logs/index.js
```diff
@@ -35,8 +35,8 @@ function onLog (log) {
 }
 
 function onErrorLog (msg) {
-  const { message, cause } = msg
-  if (!message && !cause) return
+  const { message, cause, sendViaTelemetry } = msg
+  if (!sendViaTelemetry || (!message && !cause)) return
```
tlhunter, line 39: Here's where the decision to send an error to telemetry needed to be made.

packages/dd-trace/src/exporters/common/request.js
```diff
@@ -106,11 +106,12 @@ function request (data, options, callback) {
         } catch {
           // ignore error
         }
+
         const responseData = buffer.toString()
         if (responseData) {
           errorMessage += ` Response from the endpoint: "${responseData}"`
         }
-        const error = new Error(errorMessage)
+        const error = new log.NoTransmitError(errorMessage)
```
tlhunter, line 114: This prevents us from sending the HTTP response via telemetry which is where PII could slip in.

packages/dd-trace/src/log/log.js
```diff
@@ -22,10 +26,22 @@ class Log {
...
     let message, cause, delegate
+    let sendViaTelemetry = true
+
+    {
+      const lastArg = args.at(-1)
+      if (lastArg instanceof LogConfig) {
+        args.pop()
+        sendViaTelemetry = lastArg.transmit
+      }
+    }
```
BridgeAR, line 35: I think it would be nicer not to use the extra block scope.

marcotc:
Just to give a bit of context, we log two different error messages in some cases in Ruby:
1. One for the library user, to debug the issue, with all the information we can provide that's useful.
2. One for telemetry, with a fixed string.

https://github.com/DataDog/dd-trace-rb/blob/a09772446ae2575d6f0a80e7ce544e088c483ce4/lib/datadog/profiling/http_transport.rb#L50-L51

packages/dd-trace/src/log/log.js
tlhunter: removed

PR #5872, opened 2025-06-11
perf: improve the internal config set function

watson:
### What does this PR do?

Just an resource/peformance improvement: The function is now almost twice as fast and doesn't create any temprary objects in memory.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


BridgeAR (approved):

packages/datadog-core/src/utils/src/set.js
```diff
@@ -1,16 +1,21 @@
...
+
+  let property = object[path.slice(0, index)] ??= {}
+
+  while (true) {
+    const nextIndex = path.indexOf('.', index + 1)
+    if (nextIndex === -1) {
+      property[path.slice(index + 1)] = value
+      return
     }
+    property = property[path.slice(index + 1, nextIndex)] ??= {}
```
BridgeAR:
```suggestion
  let index = -1
  while (true) {
    const nextIndex = path.indexOf('.', index + 1)
    if (nextIndex === -1) {
      object[path.slice(index + 1)] = value
      return
    }
    object = object[path.slice(index + 1, nextIndex)] ??= {}
```

BridgeAR (commented):
LGTM while I believe we can simplify it quite a bit with my suggestion

packages/datadog-core/src/utils/src/set.js
watson: I thought about that, but wanted to not have to run `path.slice(index + 1)` for the cases where there's no `.` in the path at all (which is the normal case for a lot of config vars). Don't you think that's worth the extra complexity?

BridgeAR: The only added overhead is the loop start. I agree that most cases won't hit this case while it's only used during configuration anyway and this is not going to change much as such. I think keeping the code simpler while still improving the performance significantly is a good trade-off.

BridgeAR (approved):

PR #5879, opened 2025-06-11
fix audit running on pull requests not touching dependencies

rochdev:
### What does this PR do?

Fix audit running on pull requests not touching dependencies.

### Motivation

Otherwise, whenever a new vulnerability is detected, all PRs become blocked until we fix the vulnerability. With this change, only PRs that are touching the package.json will be affected.


.github/workflows/audit.yml
```diff
@@ -0,0 +1,22 @@
+name: Audit
+
+on:
+  pull_request:
+    paths:
+      - package.json
```
watson: We should probably add yarn.lock as well

.github/workflows/audit.yml
```diff
@@ -0,0 +1,22 @@
...
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  dependencies:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: ./.github/actions/node/active-lts
+      - run: yarn audit --groups dependencies
```
watson: Should we remove the --groups flag?

.github/workflows/audit.yml
watson: Or maybe just the lock file 🤔

rochdev: The lockfile is only used for local development, not when installing the library.

.github/workflows/audit.yml
rochdev: I don't know, I just kept what was already there, and the change to pick only dependencies seemed to be popular when it was discussed this morning.

rochdev: Actually I would say let's decide that outside the scope of this PR as it preserves the current behaviour.

.github/workflows/audit.yml
watson: I know, but this is about looking for a signal of when the dependencies change, so we know to run the audit. And in development, we always update the lock file in a PR if we bump or add a dependency. You can even update the lock file without updating `package.json` - something we also want to catch, so the correct line here would be:

```suggestion
      - yarn.lock
```

.github/workflows/audit.yml
watson: I added the `--groups dependencies` just earlier today to get around the issue that blocked all PRs from being merged. It has not been there before

watson: But feel free to leave it - just wanted to make sure you had the context

.github/workflows/audit.yml
rochdev: At that point, shouldn't we just use the lockfile? As you said we always update it when changing dependencies, and it would also allow changes to package.json that are unrelated to dependencies to not trigger the audit.

rochdev: I just realized that the suggested change is exactly that.

watson (approved):

PR #5883, opened 2025-06-12
test: clean up usage of agent.assertTelemetryReceived

watson:
### What does this PR do?

Ensure that:
- We always await calls to `agent.assertTelemetryReceived` (instead of either ignoring the promise, or chaining a `.then()` call)
- We always use a reasonable timeout.
    
Addionally, `agent.assertTelemetryReceived` is updated to allow the function to be optional.
    
A bug in `assertObjectContains` is also fixed to allow arrays to be checked.

### Motivation

This is essentially a follow up PR to: https://github.com/DataDog/dd-trace-js/pull/5374

In the cases where the promise was ignored, the test would time out if the content of the function ever threw, instead of actually showing the thrown error.

I also wanted a chance to fix the bug in `assertObjectContains`.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


watson:
* **#5883** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5883?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5883?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

integration-tests/helpers/fake-agent.js
```diff
@@ -188,7 +188,7 @@ module.exports = class FakeAgent extends EventEmitter {
       if (msg.payload.request_type !== requestType) return
       msgCount += 1
       try {
-        fn(msg)
+        fn?.(msg)
```
BridgeAR: I saw it should be optional in some cases while I believe we should just check for two arguments at the method start and define a noop for fn in such cases.

integration-tests/helpers/index.js
```diff
@@ -385,6 +385,26 @@ function setShouldKill (value) {
...
 const assertObjectContains = assert.partialDeepStrictEqual || function assertObjectContains (actual, expected) {
+  if (Array.isArray(expected)) {
+    assert.ok(Array.isArray(actual), `Expected array but got ${typeof actual}`)
+    for (const expectedItem of expected) {
+      const found = actual.some(actualItem => {
+        try {
+          if (expectedItem !== null && typeof expectedItem === 'object') {
+            assertObjectContains(actualItem, expectedItem)
+          } else {
+            assert.strictEqual(actualItem, expectedItem)
```
BridgeAR: I understand that arrays where always fully matched so far even though we only wanted to match the order, not the exact position.

The new code seems to have another issue though: it may match a single entry more than once and in an arbitrary order. That is even more critical in my opinion. To prevent that, we'd have to a) skip any entry that has already been successfully matched before and b) make sure we only match parts that are in the correct order.

You could have a look at the Node.js assert implementation as reference (ignoring sparse handling in it) https://github.com/nodejs/node/blob/708477bd8d68f3571b02bb311dccc3d892e447d1/lib/internal/util/comparisons.js#L914-L934

integration-tests/profiler/profiler.spec.js
```diff
@@ -539,21 +539,21 @@ describe('profiler', () => {
...
+    it('records profile on process exit', async () => {
       proc = fork(profilerTestFile, {
         cwd,
         env: {
           DD_TRACE_AGENT_PORT: agent.port,
           DD_PROFILING_ENABLED: 1
         }
       })
-      const checkTelemetry = agent.assertTelemetryReceived(_ => {}, 1000, 'generate-metrics')
+      const checkTelemetry = agent.assertTelemetryReceived(null, 1000, 'generate-metrics')
```
BridgeAR:
```suggestion
      const checkTelemetry = agent.assertTelemetryReceived(1000, 'generate-metrics')
```

I think we should just handle it in the function to be optional like that.

integration-tests/profiler/profiler.spec.js
```diff
@@ -539,21 +539,21 @@ describe('profiler', () => {
...
       // SSI telemetry is not supposed to have been emitted when DD_INJECTION_ENABLED is absent,
       // so expect telemetry callback to time out
-      return Promise.all([checkProfiles(agent, proc, timeout), expectTimeout(checkTelemetry)])
+      await Promise.all([checkProfiles(agent, proc, timeout), expectTimeout(checkTelemetry)])
```
BridgeAR, line 553: I guess you wanted to make it explicit that the methods are async? Because this only changes it to add an extra tick effectively.

watson: Correct. I felt that it was easier to understand what was going on when reading the test

watson: It also aligns this test with all the others

integration-tests/profiler/profiler.spec.js
watson: Fixed 👍

integration-tests/helpers/fake-agent.js
watson: Fixed 👍

integration-tests/helpers/index.js
watson: Fixed 👍

PR #5884, opened 2025-06-12
ci: use official system tests GitHub Actions workflow

watson:
### What does this PR do?

This runs _all_ system tests instead of just a subset and tries to balance them out as much as possible, to get below the `desired_execution_time`.

The `desired_execution_time` is set to 5 minutes in this PR. In reality this makes the slowest system test [run in 7m 17s](https://github.com/DataDog/dd-trace-js/actions/runs/15925000147/usage?pr=5884), so I still consider that within what we can accept.

### Motivation

- No need to maintain our own system tests workflow, when an official one exists
- Better test coverage in PRs

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


watson:
* **#5884** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5884?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/5884?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

.github/workflows/system-tests.yml
```diff
@@ -4,10 +4,14 @@ on:
   pull_request:
   push:
     branches: [master]
-  workflow_dispatch: {}
+  workflow_dispatch:
```
simon-id, line 7: the empty line as a replacement for empty object is not the most readable thing imo. It almost looks like `schedule` is inside `workflow_dispatch` don't you think ?

simon-id:
otherwise lgtm

.github/workflows/system-tests.yml
watson: This was just a fly-by change to bring this in line with how we otherwise specify empty yaml properties in the `.github` folder. See for example the `pull_request` propert a few lines above

watson: It was the _only_ property where we used `{}`

szegedi (approved):

simon-id (approved):

rochdev:
This is causing regressions compared to some optimizations we had in place before. I think this should either be reverted or we need to find how to apply those optimizations in the official system tests (which I believe should have been done first).

PR #5885, opened 2025-06-12
[test-optimization] [SDTEST-1060] Add mocked files of Jest to the list of dependencies in TIA

Mariovido:
### What does this PR do?

With this change we will not longer skip tests if some of their mocked files has been changed/deleted.

### Motivation

There was an incident where we were skipping a test and one of their mocked files was deleted in that same PR. In Jest mocked files don't fail in typecheck, so there is a need to add this mocked files to the list of dependencies.
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/jest.js
+++ b/packages/datadog-instrumentations/src/jest.js
@@ -90,6 +90,7 @@ const retriedTestsToNumAttempts = new Map()
 const newTestsTestStatuses = new Map()
 const attemptToFixRetriedTestsStatuses = new Map()
 const wrappedWorkers = new WeakSet()
+const testSuiteMockedFiles = new Map()
 
 const BREAKPOINT_HIT_GRACE_PERIOD_MS = 200
 
@@ -1095,10 +1096,12 @@ function jestAdapterWrapper (jestAdapter, jestVersion) {
       if (environment.testEnvironmentOptions?._ddTestCodeCoverageEnabled) {
         const root = environment.repositoryRoot || environment.rootDir
 
-        const coverageFiles = getCoveredFilenamesFromCoverage(environment.global.__coverage__)
-          .map(filename => getTestSuitePath(filename, root))
+        const getFilesWithPath = (files) => files.map(file => getTestSuitePath(file, root))
 
-        testSuiteCodeCoverageCh.publish({ coverageFiles, testSuite: environment.testSourceFile })
+        const coverageFiles = getFilesWithPath(getCoveredFilenamesFromCoverage(environment.global.__coverage__))
+        const mockedFiles = getFilesWithPath(testSuiteMockedFiles.get(environment.testSuite) || [])
+
+        testSuiteCodeCoverageCh.publish({ coverageFiles, testSuite: environment.testSourceFile, mockedFiles })
       }
       testSuiteFinishCh.publish({ status, errorMessage })
       return suiteResults
@@ -1273,7 +1276,21 @@ addHook({
       // To bypass jest's own require engine
       return this._requireCoreModule(moduleName)
     }
-    return requireModuleOrMock.apply(this, arguments)
+    const result = requireModuleOrMock.apply(this, arguments)
+
+    const suiteFileName = this._mainModule?.filename
+    if (suiteFileName) {
+      const testSuite = getTestSuitePath(suiteFileName, process.cwd())
+      for (const [key] of this._mockRegistry?.entries() || []) {
+        const path = key?.match(/^user:(.*?):/)?.[1]
+        if (path) {
+          const existingMockedFiles = testSuiteMockedFiles.get(testSuite) || []
+          testSuiteMockedFiles.set(testSuite, [...existingMockedFiles, path])
+        }
+      }
+    }
+
+    return result
   })
 
   return runtimePackage
--- a/packages/datadog-plugin-jest/src/index.js
+++ b/packages/datadog-plugin-jest/src/index.js
@@ -318,11 +318,11 @@ class JestPlugin extends CiPlugin {
      * because this subscription happens in a different process from the one
      * fetching the ITR config.
      */
-    this.addSub('ci:jest:test-suite:code-coverage', ({ coverageFiles, testSuite }) => {
+    this.addSub('ci:jest:test-suite:code-coverage', ({ coverageFiles, testSuite, mockedFiles }) => {
       if (!coverageFiles.length) {
         this.telemetry.count(TELEMETRY_CODE_COVERAGE_EMPTY)
       }
-      const files = [...coverageFiles, testSuite]
+      const files = [...coverageFiles, ...mockedFiles, testSuite]
 
       const { _traceId, _spanId } = this.testSuiteSpan.context()
       const formattedCoverage = {


integration-tests/jest/jest.spec.js
```diff
@@ -1560,6 +1560,71 @@ describe('jest CommonJS', () => {
...
+
+        const [coveragePayload] = codeCovRequest.payload
+        assert.propertyVal(codeCovRequest.headers, 'dd-api-key', '1')
+
+        assert.propertyVal(coveragePayload, 'name', 'coverage1')
+        assert.propertyVal(coveragePayload, 'filename', 'coverage1.msgpack')
+        assert.propertyVal(coveragePayload, 'type', 'application/msgpack')
+        assert.include(coveragePayload.content, {
+          version: 2
+        })
```
juan-fernandez: we don't need to check this for this particular test

integration-tests/jest/jest.spec.js
```diff
@@ -1560,6 +1560,71 @@ describe('jest CommonJS', () => {
...
+      )
+      const codeCovRequestPromise = receiver.payloadReceived(({ url }) => url === '/api/v2/citestcov')
+      const eventsRequestPromise = receiver.payloadReceived(({ url }) => url === '/api/v2/citestcycle')
+
+      Promise.all([
+        libraryConfigRequestPromise,
+        codeCovRequestPromise,
+        eventsRequestPromise
+      ]).then(([libraryConfigRequest, codeCovRequest, eventsRequest]) => {
+        assert.propertyVal(libraryConfigRequest.headers, 'dd-api-key', '1')
```
juan-fernandez: we don't need to check this for this particular test

integration-tests/jest/jest.spec.js
```diff
@@ -1560,6 +1560,71 @@ describe('jest CommonJS', () => {
...
+
+        const testSession = eventsRequest.payload.events.find(event => event.type === 'test_session_end').content
+        assert.exists(testSession.metrics[TEST_CODE_COVERAGE_LINES_PCT])
+
+        const eventTypes = eventsRequest.payload.events.map(event => event.type)
+        assert.includeMembers(eventTypes, ['test', 'test_suite_end', 'test_module_end', 'test_session_end'])
+        const numSuites = eventTypes.reduce(
+          (acc, type) => type === 'test_suite_end' ? acc + 1 : acc, 0
+        )
+        assert.equal(numSuites, 1)
```
juan-fernandez: let's just check that `allCoverageFiles` includes the files to simplify

integration-tests/jest/jest.spec.js
```diff
@@ -1560,6 +1560,71 @@ describe('jest CommonJS', () => {
         }).catch(done)
       })
     })
+
+    it('report code coverage with all mocked files', (done) => {
+      const libraryConfigRequestPromise = receiver.payloadReceived(
+        ({ url }) => url === '/api/v2/libraries/tests/services/setting'
+      )
+      const codeCovRequestPromise = receiver.payloadReceived(({ url }) => url === '/api/v2/citestcov')
```
juan-fernandez, line 1565: we need only to check `codeCovRequestPromise` for this test

integration-tests/jest/jest.spec.js
```diff
@@ -1560,6 +1560,71 @@ describe('jest CommonJS', () => {
...
+          cwd,
+          env: {
+            ...getCiVisAgentlessConfig(receiver.port),
+            TESTS_TO_RUN: 'jest/mocked-test.js',
+          },
+          stdio: 'pipe'
+        }
+      )
+      childProcess.stdout.on('data', (chunk) => {
+        testOutput += chunk.toString()
```
juan-fernandez: I don't think we need the `testOutput` either

packages/datadog-instrumentations/src/jest.js
```diff
@@ -1267,6 +1271,24 @@ addHook({
 }, (runtimePackage) => {
   const Runtime = runtimePackage.default ?? runtimePackage
 
+  shimmer.wrap(Runtime.prototype, '_createJestObjectFor', _createJestObjectFor => function (from) {
+    const result = _createJestObjectFor.apply(this, arguments)
+    const suiteFilePath = this._testPath
+
+    shimmer.wrap(result, 'mock', mock => function (moduleName, mockFactory, options) {
```
juan-fernandez: we only need the `moduleName`, right? 
```suggestion
    shimmer.wrap(result, 'mock', mock => function (moduleName) {
```

packages/datadog-instrumentations/src/jest.js
```diff
@@ -1267,6 +1271,24 @@ addHook({
...
+      if (suiteFilePath) {
+        const testSuite = getTestSuitePath(suiteFilePath, process.cwd())
+        const existingMockedFiles = testSuiteMockedFiles.get(testSuite) || []
+        const suiteDir = path.dirname(suiteFilePath)
+        const mockPath = path.resolve(suiteDir, moduleName)
+        const relativeMockPath = path.relative(process.cwd(), mockPath)
+        testSuiteMockedFiles.set(testSuite, [...existingMockedFiles, relativeMockPath])
+      }
+      return mock.apply(this, arguments)
```
juan-fernandez: `suiteFilePath` is the absolute test suite path, right? I wonder if we could simplify things by keeping the absolute path everywhere and only calculating the relative path before `testSuiteCodeCoverageCh.publish`
```suggestion
      if (suiteFilePath) {
        const existingMockedFiles = testSuiteMockedFiles.get(suiteFilePath) || []
        const suiteDir = path.dirname(suiteFilePath)
        const mockPath = path.resolve(suiteDir, moduleName)
        existingMockedFiles.push(mockPath)
      }
      return mock.apply(this, arguments)
```
caveat: `testSuiteMockedFiles` now includes absolute paths, so when extracting `mockedFiles` you should use the absolute path

```javascript
const mockedFiles = getFilesWithPath(testSuiteMockedFiles.get(environment.testSuiteAbsolutePath) || [])
```

`environment.testSuiteAbsolutePath` would be undefined but I think you can add it simply by doing something like 

```javascript
this.testSuiteAbsolutePath = context.testPath
```

in the constructor.

juan-fernandez (commented):
good job finding `_createJestObjectFor`! My comments are mostly minor things

integration-tests/jest/jest.spec.js
Mariovido: Removed!

integration-tests/jest/jest.spec.js
Mariovido: Removed!

integration-tests/jest/jest.spec.js
Mariovido: Changed!

integration-tests/jest/jest.spec.js
Mariovido: Removed!

packages/datadog-instrumentations/src/jest.js
Mariovido: You're right, removed the other two inputs

integration-tests/jest/jest.spec.js
Mariovido: Removed the other two promises

packages/datadog-instrumentations/src/jest.js
Mariovido: Already refactored :)

PR #5888, opened 2025-06-12
Improve support for esbuild + esm with DD_BUILD_ESM env var

uurien:
### What does this PR do?

Improves support for esbuild + esm with DD_BUILD_ESM env var, fixing the issue with `__dirname` and `__filename`.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


integration-tests/esbuild/index.spec.js
```diff
@@ -44,6 +44,23 @@ esbuildVersions.forEach((version) => {
...
+      chproc.execSync('npm run build:esm')
+
+      console.log('npm run built')
+      try {
+        chproc.execSync('npm run built:esm', {
+          timeout: 1000 * 30
+        })
+      } catch (err) {
+        console.error(err)
+        process.exit(1)
```
BridgeAR:
```suggestion
```

integration-tests/esbuild/package.json
```diff
@@ -22,7 +24,7 @@
     "@koa/router": "*",
     "aws-sdk": "*",
     "axios": "*",
-    "esbuild": "*",
+    "esbuild": "^0.16.12",
```
BridgeAR:
```suggestion
    "esbuild": "*",
```

That should be a start with how the tests are currently written. We should use `withVersions` but it was simpler before to just run it with more than one version as is.

packages/datadog-esbuild/index.js
```diff
@@ -55,6 +56,18 @@ module.exports.name = 'datadog-esbuild'
 
 module.exports.setup = function (build) {
   const externalModules = new Set(build.initialOptions.external || [])
+
+  if (DD_BUILD_ESM) {
```
```suggestion
  if (build.initialOptions.format === 'esm' || build.initialOptions.outfile.endsWith('.mjs')) {
```

The second case would need an additional test case where we do not use `format`.

packages/datadog-esbuild/index.js
```diff
@@ -53,8 +53,29 @@ for (const pkg of INSTRUMENTED) {
 
 module.exports.name = 'datadog-esbuild'
 
+function isESMBuild (build) {
+  // check toLowerCase? to be safe if unexpected object is there instead of a string
+  const format = build.initialOptions.format?.toLowerCase?.()
+  const outputFile = build.initialOptions.outfile?.toLowerCase?.()
+  const outExtension = build.initialOptions.outExtension?.['.js']
```
tlhunter, line 60: So this is looking for a value like so?

```js
build.initialOptions.outExtension['.js'] = '.mjs'
```

Based on property names I would expect it to be more like:

```js
build.initialOptions.outExtension = '.mjs'
```

BridgeAR: Judging by the documentation it's correct as is: https://esbuild.github.io/api/#out-extension :)

packages/datadog-esbuild/index.js
```diff
@@ -53,8 +53,29 @@ for (const pkg of INSTRUMENTED) {
...
+    build.initialOptions.banner ??= {}
+    build.initialOptions.banner.js ??= ''
+    build.initialOptions.banner.js = `import { createRequire as $dd_createRequire } from 'module';
+import { fileURLToPath as $dd_fileURLToPath } from 'url';
+import { dirname as $dd_dirname } from 'path';
+(function(globals) {
+  globals.require ??= $dd_createRequire(import.meta.url);
+  globals.__filename ??= $dd_fileURLToPath(import.meta.url);
+  globals.__dirname ??= $dd_dirname(__filename);
+}((1, eval)('this')));${build.initialOptions.banner.js}`
```
BridgeAR: I would just access globalThis instead of using eval for it :)

```suggestion
globalThis.require ??= $dd_createRequire(import.meta.url);
globalThis.__filename ??= $dd_fileURLToPath(import.meta.url);
globalThis.__dirname ??= $dd_dirname(__filename);
${build.initialOptions.banner.js}`
```

packages/datadog-esbuild/index.js
uurien: Yes, I've added this because I read that in the docs :-)
It expects `.css` or `.js` to convert to a different ending.

integration-tests/esbuild/index.spec.js
uurien: tbh, the only reason why I added this, is because it is also in the previous test

integration-tests/esbuild/index.spec.js
```diff
@@ -75,5 +76,71 @@ esbuildVersions.forEach((version) => {
...
+        fs.rmSync('./out-with-patched-global-banner.mjs', { force: true })
+        fs.rmSync('./out-with-patched-const-banner.mjs', { force: true })
+        fs.rmSync('./out-relying-in-extension.mjs', { force: true })
+        fs.rmSync('./out.js', { force: true })
+        fs.rmSync('./basic-test.mjs', { force: true })
+        fs.rmSync('./out-non-esm.js', { force: true })
+      })
+
+      it('works', () => {
+        console.log('npm run built')
```
tlhunter:
```suggestion
        console.log('npm run built:esm')
```

tlhunter (dismissed):

BridgeAR (approved):
LGTM. Would this potentially also fix an open issue?

uurien:
> Would this potentially also fix an open issue?

I would say no. I can't find any issue related with esbuild build process with ESM, or esbuild app crashing in the issues.

Checking here:
https://github.com/DataDog/dd-trace-js/issues?q=is%3Aissue%20state%3Aopen%20esbuild

integration-tests/esbuild/build-and-run.esm-patched-const-banner.js
```diff
@@ -0,0 +1,33 @@
...
+  if (stderr.length) {
+    console.error(stderr.toString())
+  }
+  if (status) {
+    throw new Error('generated script failed to run')
+  }
+  console.log('ok')
+}).catch((err) => {
+  console.error(err)
+  process.exit(1)
```
BridgeAR: Nit: calling process.exit(1) is mostly an indicator that something is not working as expected.
Here, if we'd change the files to mjs files, we could just use top level await and things should work as expected.

The same applies to the other files.

integration-tests/esbuild/index.spec.js
```diff
@@ -75,5 +76,93 @@ esbuildVersions.forEach((version) => {
         timeout: 1000 * 30
       })
     })
+
+    describe('ESM', () => {
+      afterEach(() => {
+        fs.rmSync('./out.mjs', { force: true })
+        fs.rmSync('./out.js', { force: true })
+        fs.rmSync('./basic-test.mjs', { force: true })
```
BridgeAR, line 84: If we move the cleanup to the file that runs the test by using a finally block, we'd not have to list them here.
That feels a tad cleaner.

BridgeAR (dismissed):
LGTM. I just left some suggestions that would make things a bit cleaner in my opinion.

integration-tests/esbuild/index.spec.js
uurien: There are some tests that are checking the content of the files, like this one: https://github.com/DataDog/dd-trace-js/pull/5888/files/13e26af2feb6a6d3a14d1677cdbebe29bfd08a07

I don't want to move these checks to the file running the tests, imo the test logic is better here. imho build & run are more than enough to have in separate files.

BridgeAR (approved):

PR #5889, opened 2025-06-12
Ishabi/fastify path params

IlyasShabi:
### What does this PR do?
Allow blocking on fastify path params

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -262,6 +262,174 @@ withVersions('fastify', 'fastify', version => {
...
+        } catch (e) {
+          assert.equal(e.response.status, 403)
+          assert.deepEqual(e.response.data, JSON.parse(json))
+        }
+      })
+    })
+
+    describe('path parameter with hook', () => {
+      it('should not block the request when attack is not detected', async () => {
+        const res = await axios.get('/callback-path-param/safe_param')
```
uurien, line 425:
```suggestion
        const res = await axios.get('/callback-path-param/safe_param')

```

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -262,6 +262,174 @@ withVersions('fastify', 'fastify', version => {
...
+    before((done) => {
+      const fastify = require(`../../../../versions/fastify@${version}`).get()
+
+      const app = fastify()
+      app.get('/multiple-path-params/:parameter1/:parameter2', (request, reply) => {
+        reply.send('DONE')
+      })
+
+      app.register(async function (nested) {
+        nested.get('/:nestedDuplicatedParameter', async (request, reply) => {
```
uurien: The reason why we have this `nestedDuplicatedParameter` in express test, is that we have two parameters with the same name in different routes.
Here, we don't have any other parameter called `nestedDuplicatedParameter`, so the property name is wrong, and should be something like `secondParam`, or in the prefix, it should be also `nestedDuplicatedParameter`.
Maybe this example does make no sense in fastify.

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -262,6 +262,174 @@ withVersions('fastify', 'fastify', version => {
...
+      // Route with hook for path parameters
+      const paramHook = (request, reply, done) => {
+        done()
+      }
+
+      paramHookSpy = sinon.spy(paramHook)
+
+      app.addHook('preHandler', paramHookSpy)
+
+      app.get('/callback-path-param/:callbackedParameter', (request, reply) => {
```
uurien: Same for this test, in express tests, `:callbackedParameter` is called as it is, because we are listening for `app.param('callbackedParameter'...)`. Does it make sense here?

IlyasShabi: Parameter name doesn't make sense here because in Fastify we have a global hook that runs for all routes using `app.addHook('preHandler', paramHookSpy)`

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -262,6 +262,175 @@ withVersions('fastify', 'fastify', version => {
...
+
+      it('should block the request when attack is detected in both parameters', async () => {
+        try {
+          await axios.get('/multiple-path-params/testattack/testattack')
+
+          return Promise.reject(new Error('Request should not return 200'))
+        } catch (e) {
+          assert.equal(e.response.status, 403)
+          assert.deepEqual(e.response.data, JSON.parse(json))
+        }
```
BridgeAR, line 355: Instead of using try / catch, we could switch to the native `assert` and use `assert.rejects`. That provides more information and is IMO easier to reason about.

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -133,6 +134,17 @@ function preValidation (request, reply, done) {
       if (abortController.signal.aborted) return
     }
 
+    if (pathParamsReadCh.hasSubscribers && request.params) {
+      const abortController = new AbortController()
```
BridgeAR: Creating new AbortControllers is quite expensive. Could we just reuse an already defined controller?
I believe it's possible to unify the up to three controllers to a single one.

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -133,6 +134,17 @@ function preValidation (request, reply, done) {
...
+
+      pathParamsReadCh.publish({
+        req,
+        res,
+        abortController,
+        params: request.params
+      })
+    }
```
simon-id: uhm where is the blocking ?

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -126,13 +127,22 @@ function preValidation (request, reply, done) {
...
       bodyParserReadCh.publish({ req, res, body: request.body, abortController })
 
       if (abortController.signal.aborted) return
     }
 
+    if (pathParamsReadCh.hasSubscribers && request.params) {
+      pathParamsReadCh.publish({
+        req,
+        res,
+        abortController,
```
simon-id, line 142: if you do this for the 3 different calls, you don't create it until it's needed, but you also cache it for the others. Wdyt ?
```suggestion
        abortController: abortController ?? abortController = new AbortController(),
```

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -264,7 +264,11 @@ withVersions('fastify', 'fastify', version => {
   })
 
   describe('Suspicious request blocking - path parameters', () => {
-    let server, paramHookSpy, axios
+    // Skip Fastify v1 - preValidation hook is not supported
+    if (semver.lt(semver.coerce(version), '2.0.0')) {
```
simon-id, line 268: if we don't really support blocking before 2.0.0 we should consider appsec only support fastify starting from v2 and add that to the relevant doc

packages/datadog-instrumentations/src/fastify.js
uurien: it could be also:

```javascript
abortController: abortController ??= new AbortController()
```

But I think that inlining everything here makes this more confusing than cleaner.

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -112,9 +113,10 @@ function preValidation (request, reply, done) {
   const parsingResource = parsingResources.get(req)
 
   const processInContext = () => {
-    if (queryParamsReadCh.hasSubscribers && request.query) {
-      const abortController = new AbortController()
+    let abortController
 
+    if (queryParamsReadCh.hasSubscribers && request.query) {
+      abortController = abortController ?? new AbortController()
```
uurien:
```suggestion
      abortController ??= new AbortController()
```

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -262,6 +262,187 @@ withVersions('fastify', 'fastify', version => {
       }
     })
   })
+
+  describe('Suspicious request blocking - path parameters', () => {
+    // Skip Fastify v1 - preValidation hook is not supported
+    if (semver.lt(semver.coerce(version), '2.0.0')) {
+      return
+    }
+    let server, preHandlerHookSpy, preValidationHookSpy, axios
```
uurien, line 272:
```suggestion

    let server, preHandlerHookSpy, preValidationHookSpy, axios
```

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
IlyasShabi: I will create another PR to use `assert.rejects` in all tests/appsec files

PR #5891, opened 2025-06-12
add paths config to workflows for pull_request trigger

rochdev:
### What does this PR do?

Add paths config to workflows for pull_request trigger.

### Motivation

Otherwise, editing unrelated files end up running unrelated workflows. For now I basically isolated the change to obvious files like markdown and any workflow YAML that is not the one being currently updated. As our codebase becomes more granular, we could extend this to code and test as well.


.github/workflows/apm-capabilities.yml
```diff
@@ -2,6 +2,23 @@ name: APM Capabilities
 
 on:
   pull_request:
+    paths:
+      # Paths for whole project (should be same in all workflows)
```
BridgeAR, line 6: I think it would be good to abstract that. Otherwise it is difficult to keep it aligned over time and the duplication is also not ideal.

rochdev: I agree in principle but to my knowledge there is no way to do this since there is no reusability constructs for the triggers section. This is even more true to keep the list in sync between GitHub and GitLab. The only thing we could potentially do is validate the list (or at least the minimum required entries).

BridgeAR: I see, thank you

.github/workflows/apm-capabilities.yml
```diff
@@ -2,6 +2,23 @@ name: APM Capabilities
...
+      - '!.gitlab/**'
+      - '!.vscode/**'
+      - '!devdocs/**'
+      - '!examples/**'
+      - '!*.md'
+      - '!.gitlab-ci.yml'
+      - '!LICENSE'
+      - '!LICENSE.Apache'
+      - '!LICENSE.BSD3'
+      - '!LICENSE'
```
BridgeAR:
```suggestion
      - '!LICENSE*'
      - 'LICENSE-3rdparty.csv'
```

That applies to all :)

rochdev: It also applies to `LICENSE-3rdparty.csv` which could affect some CI workflows.

BridgeAR: Let's add it to the accepted ones in that case. That way we are able to remove two lines each time :)

rochdev: Done!

BridgeAR (approved):
Still LGTM

PR #5892, opened 2025-06-12
add explicit main ref when checking out system tests in workflow

rochdev:
### What does this PR do?

Add explicit `main` ref when checking out system tests in workflow.

### Motivation

Otherwise each job in the matrix needs an API call to figure out the default branch, and since we have hundreds of jobs in the matrix it resulted in API call limits being reached.


simon-id (approved):
beautifully simple fix, i've also checked all other occurences of checkout action in the repo and couldn't find a place where this fix would fit :ok_hand:

PR #5894, opened 2025-06-13
Support schema extraction in fastify response objects

IlyasShabi:
### What does this PR do?
Support schema extraction in fastify response objects

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/datadog-instrumentations/src/fastify.js
```diff
@@ -153,10 +155,14 @@ function preParsing (request, reply, payload, done) {
...
-    if (error instanceof Error) {
-      errorChannel.publish({ req, error })
+function wrapSend (send, req, res) {
+  return function sendWithTrace (payload) {
+    if (payload instanceof Error) {
+      errorChannel.publish({ req, error: payload })
+    } else {
+      if (responseJsonReadCh.hasSubscribers) {
+        responseJsonReadCh.publish({ req, res, body: payload })
+      }
```
BridgeAR:
```suggestion
    } else if (responseJsonReadCh.hasSubscribers) {
      responseJsonReadCh.publish({ req, res, body: payload })
```

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -100,8 +101,9 @@ function preHandler (request, reply, done) {
   if (!reply || typeof reply.send !== 'function') return done()
 
   const req = getReq(request)
+  const res = getRes(reply)
 
-  reply.send = wrapSend(reply.send, req)
+  reply.send = wrapSend(reply.send, req, res)
```
BridgeAR: Do we have to pass through the actual res? If not, I would just pass through reply and only read it when it's actually used. That way the case where appsec is not used won't be impacted (not that it's a big hit but it's read on every call).

IlyasShabi: Good point! I removed res and get it inside wrapSend when needed

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -264,6 +265,91 @@ withVersions('fastify', 'fastify', version => {
...
+    }
+
+    it('should get the response body schema with reply.send', async () => {
+      const expectedResponseBodySchema = formatSchema([{ sendResKey: [8] }])
+      const res = await axios.post('/send', { key: 'value' })
+
+      await agent.assertSomeTraces((traces) => {
+        const span = traces[0][0]
+        assert.equal(span.meta['_dd.appsec.s.res.body'], expectedResponseBodySchema)
+      })
```
BridgeAR: You could use the new `assertFirstTraceSpan` helper :)

The same in the spot below.

```suggestion
      await agent.assertFirstTraceSpan({
        meta: {
          '_dd.appsec.s.res.body': expectedResponseBodySchema
        }
      })
```

IlyasShabi: Good to know. Done

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -154,9 +155,12 @@ function preParsing (request, reply, payload, done) {
...
 function wrapSend (send, req) {
-  return function sendWithTrace (error) {
-    if (error instanceof Error) {
-      errorChannel.publish({ req, error })
+  return function sendWithTrace (payload) {
+    if (payload instanceof Error) {
+      errorChannel.publish({ req, error: payload })
+    } else if (responseJsonReadCh.hasSubscribers) {
+      const res = getRes(this)
+      responseJsonReadCh.publish({ req, res, body: payload })
```
uurien: `reply.send` expects a lot of different types: `Error`, `string`, `Buffer`, `Stream`/`ReadableStream`, TypedArrays, and other objects. We are only interested in other objects, and the subscriber shouldn't understand about different options that this method support. We should discar, at least, all the `Buffer`, `Stream`/`ReadableStream`, TypedArrays and send only when we have the real and final object.

uurien: Here the link to the docs: https://fastify.dev/docs/latest/Reference/Reply/#senddata

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -264,6 +266,93 @@ withVersions('fastify', 'fastify', version => {
...
+
+    afterEach(() => {
+      appsec.disable()
+    })
+
+    function formatSchema (body) {
+      return zlib.gzipSync(JSON.stringify(body)).toString('base64')
+    }
+
+    it('should get the response body schema with reply.send', async () => {
```
uurien, line 527: Related with the previous comment, I would add more tests to check that no api security tags are generated for Buffer, Streams etc.

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -200,6 +204,24 @@ function onRoute (routeOptions) {
...
+    case typeof payload !== 'object':
+    case Buffer.isBuffer(payload):
+    case typeof payload.pipe === 'function': // Node streams (.pipe)
+    case payload.body && typeof payload.body.pipe === 'function': // Response with body stream
+    case payload instanceof ArrayBuffer: // ArrayBuffer
+    case ArrayBuffer.isView(payload): // TypedArray
+      return false
+    default:
+      return true
+  }
```
BridgeAR: I would argue this is not ideal to return true or false 😅 

```suggestion
  return responsePayloadReadCh.hasSubscribers &&
    payload &&
    typeof payload === 'object' &&
    !Buffer.isBuffer(payload) &&
    typeof payload.pipe !== 'function' && // Node streams (.pipe)
    typeof payload.body?.pipe !== 'function' && // Response with body stream
    !util.types.isArrayBuffer(payload) && // ArrayBuffer
    !util.types.isArrayBufferView(payload) // TypedArray
```

That aside: I wonder if it's not possible to check for a specific type instead of checking for not being the wrong type. The current check is quite difficult and might fail over time.

IlyasShabi: Checking every type feels a bit messy, but it’s safer because it lines up with what Fastify actually returns.  
I tried the simpler check like `Object.getPrototypeOf(payload) === Object.prototype`  but that could let through new payload types Fastify adds in the future, which we don’t want to collect.

uurien: We cannot check for a specific type, because any other type would be supported, if an application is returning a DTO object extending from other, the `Object.getPrototype(B)` will return `A`, and not `Object.prototype`. (of course, if `B` extends `A`)

uurien (approved):

PR #5898, opened 2025-06-16
[test optimization] Improve flakiness of performance overhead test

juan-fernandez:
### What does this PR do?
Reduce flakiness of performance overhead test.

### Motivation
Sometimes the script fails because `jobs` is undefined. This is likely due to a malformed response from github: https://github.com/DataDog/dd-trace-js/actions/runs/15613751784/job/43981048645

### DIFF
--- a/benchmark/e2e-test-optimization/benchmark-run.js
+++ b/benchmark/e2e-test-optimization/benchmark-run.js
@@ -3,10 +3,13 @@
 /* eslint-disable no-console */
 
 const https = require('https')
+const { setTimeout } = require('timers/promises')
 
 const API_REPOSITORY_URL = 'https://api.github.com/repos/DataDog/test-environment'
 const DISPATCH_WORKFLOW_URL = `${API_REPOSITORY_URL}/actions/workflows/dd-trace-js-tests.yml/dispatches`
 const GET_WORKFLOWS_URL = `${API_REPOSITORY_URL}/actions/runs`
+
+const MAX_ATTEMPTS = 30 * 60 / 5 // 30 minutes, polling every 5 seconds = 360 attempts
 
 function getBranchUnderTest () {
   /**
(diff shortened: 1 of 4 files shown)


rochdev (approved):

benchmark/e2e-test-optimization/benchmark-run.js
```diff
@@ -148,22 +148,27 @@ async function main () {
...
-        clearInterval(intervalId)
-      } else if (hasEveryJobPassed) {
-        console.log('Performance overhead test successful.')
-        resolve()
-        clearInterval(intervalId)
-      } else {
-        console.log(`Workflow https://github.com/DataDog/test-environment/actions/runs/${runId} is not finished yet.`)
+      try {
+        const currentWorkflow = await getCurrentWorkflowJobs(runId)
+        const { jobs } = currentWorkflow
```
BridgeAR:
```suggestion
      const currentWorkflow = await getCurrentWorkflowJobs(runId)
      const { jobs } = currentWorkflow
      if (!jobs) {
        console.error(`Workflow check returned unknown object %o. Retry in 5 seconds.`, currentWorkflow)
        return
      }  
```

That way it's handling the condition as shown in the failed job and we get the actual output as well instead of seeing accessing `some of undefined`.

BridgeAR (commented):
Could we change this to a loop that has a maximum count of e.g., 720 (maximum run time is one hour). That way the code is likely easier to read:

```js
const pollTimeInMS = 5000
for (let run = 0: run < 60 * 60 * 1000; run += pollTimeInMS) {
  // do stuff
  if (endCodition) {
    return
  }
  await timers.promises.setTimeout(pollTimeInMS)
}

```

benchmark/e2e-test-optimization/benchmark-run.js
juan-fernandez: yeah that makes sense. Done in https://github.com/DataDog/dd-trace-js/pull/5898/commits/78764fd6b079884b34fc106b3144b50d76bb4b7e

BridgeAR (approved):
LGTM, thanks for the cleanup :)

PR #5899, opened 2025-06-16
test: sort config value assertions

watson:
### What does this PR do?

Sort the assertions in the major tests in `config.spec.js`.

### Motivation

Less bike-shedding in the future when modifying these tests.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes

One duplicate assertion found during refactoring which has been removed. Besides that, the only changes are in the order

### DIFF
--- a/packages/dd-trace/test/config.spec.js
+++ b/packages/dd-trace/test/config.spec.js
@@ -262,88 +262,88 @@ describe('Config', () => {
   it('should initialize with the correct defaults', () => {
     const config = new Config()
 
-    expect(config).to.have.property('service', 'node')
-    expect(config).to.have.property('tracing', true)
-    expect(config).to.have.property('debug', false)
-    expect(config).to.have.property('protocolVersion', '0.4')
-    expect(config).to.have.nested.property('dogstatsd.hostname', '127.0.0.1')
-    expect(config).to.have.nested.property('dogstatsd.port', '8125')
-    expect(config).to.have.property('flushInterval', 2000)
-    expect(config).to.have.property('flushMinSpans', 1000)
-    expect(config).to.have.property('queryStringObfuscation').with.length(626)
-    expect(config).to.have.property('clientIpEnabled', false)
-    expect(config).to.have.property('clientIpHeader', null)
-    expect(config).to.have.property('middlewareTracingEnabled', true)
-    expect(config).to.have.nested.property('crashtracking.enabled', true)
-    expect(config).to.have.property('sampleRate', undefined)
-    expect(config).to.have.property('runtimeMetrics', false)
-    expect(config).to.have.property('runtimeMetricsRuntimeId', false)
-    expect(config.tags).to.have.property('service', 'node')
-    expect(config).to.have.property('plugins', true)
-    expect(config).to.have.property('traceEnabled', true)
-    expect(config).to.have.property('env', undefined)
-    expect(config).to.have.property('reportHostname', false)
-    expect(config).to.have.property('scope', undefined)
-    expect(config).to.have.property('logLevel', 'debug')
-    expect(config).to.have.nested.property('codeOriginForSpans.enabled', true)
-    expect(config).to.have.nested.property('codeOriginForSpans.experimental.exit_spans.enabled', false)
-    expect(config).to.have.nested.property('dynamicInstrumentation.enabled', false)
-    expect(config).to.have.nested.deep.property('dynamicInstrumentation.redactedIdentifiers', [])
-    expect(config).to.have.nested.deep.property('dynamicInstrumentation.redactionExcludedIdentifiers', [])
-    expect(config).to.have.nested.property('dynamicInstrumentation.uploadIntervalSeconds', 1)
-    expect(config).to.have.property('traceId128BitGenerationEnabled', true)
-    expect(config).to.have.property('traceId128BitLoggingEnabled', true)
-    expect(config).to.have.property('spanAttributeSchema', 'v0')
-    expect(config.grpc.client.error.statuses).to.deep.equal(GRPC_CLIENT_ERROR_STATUSES)
-    expect(config.grpc.server.error.statuses).to.deep.equal(GRPC_SERVER_ERROR_STATUSES)
-    expect(config).to.have.property('spanComputePeerService', false)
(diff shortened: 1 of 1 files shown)


packages/dd-trace/test/config.spec.js
```diff
@@ -262,88 +262,88 @@ describe('Config', () => {
...
-    expect(config).to.have.property('service', 'node')
-    expect(config).to.have.property('tracing', true)
-    expect(config).to.have.property('debug', false)
-    expect(config).to.have.property('protocolVersion', '0.4')
-    expect(config).to.have.nested.property('dogstatsd.hostname', '127.0.0.1')
-    expect(config).to.have.nested.property('dogstatsd.port', '8125')
-    expect(config).to.have.property('flushInterval', 2000)
-    expect(config).to.have.property('flushMinSpans', 1000)
-    expect(config).to.have.property('queryStringObfuscation').with.length(626)
+    expect(config).to.have.nested.property('apmTracingEnabled', true)
```
BridgeAR, line 265: If we change these, can we please change it to a deepStrictEqual call instead? That way it should be much easier to read.

watson: > I doubt that it will stay sorted over time 😅

They might not stay 100% sorted over time, but it's always easier to come back and tidy things up when it started out as sorted. Today, you just give up and add your new config where ever, but if it was 99% sorted already, it's easier to tidy things up when you're making other changes.

> If we change these, can we please change it to a deepStrictEqual call instead? That way it should be much easier to read.

I prefer to keep this PR as quick and simple as possible. I don't want to spend any more time on it than I have to as I'm working on adding a new config and didn't want to keep just adding it where ever there was room 😅

BridgeAR: Could you not use copilot to refactor that while looking at something else? :)

I guess it would be capable of doing that quite well.

watson: It's going to make it complicated that some of the tests are checking the length of a given value and that the tap output for large assertions is garbage. E.g. it will just give you assertion errors like this:

> expected Config{ …(109) } to have deep property 'appsec' of { apiSecurity: { …(2) }, …(14) }, but got { apiSecurity: { …(2) }, …(14) }

watson: `deepStrictEqual` doesn't make it much better 🤔 

image

watson: I think this should be in a separate PR

BridgeAR: Tap is sadly making it very hard to get a proper output.
Doing the review here is difficult for me. If it's just a reordering, I would expect the lines to be identical but that does not seem to be the case.

watson: There were a few extra redundant linebreaks that I removed because it made automatic sorting easier

watson: For example

```diff
-     expect(config).to.have.deep.property('serviceMapping', {
-       a: 'aa',
-       b: 'bb'
-     })
+     expect(config).to.have.deep.property('serviceMapping', { a: 'aa', b: 'bb' })
```

watson: About half the changes in this PR was automatic sorting, the other half, I moved lines around manually where the automatic sorting came up short

watson: Though a bit of a hack, you can better get an idea of the actual differences if you run this command:

```sh
diff <(sort packages/dd-trace/test/config.spec.js) <(git show HEAD^:packages/dd-trace/test/config.spec.js | sort)
```

watson: Actually, this diff is smaller as it ignores trailing commas (which has been altered when items moved to/from the end of lists/arrays):

```sh
git show HEAD^:packages/dd-trace/test/config.spec.js | sed 's/,\s*$//' | sort | diff - <(sed 's/,\s*$//' packages/dd-trace/test/config.spec.js | sort)
```

watson: And here's the output of that command:

```diff
644d643
<         name: 'sampler.rules'
655d653
<         origin: 'env_var'
723d720
<         value: process.env.DD_TRACE_SAMPLING_RULES
766c763
<       {
---
>       { id: '68e75c48-57ca-4a12-adfc-575c4b05fcbe', type: 'k8s_single_step', time: '1703188212' })
958a956
>       { name: 'sampler.rules', value: process.env.DD_TRACE_SAMPLING_RULES, origin: 'env_var' }
1067d1064
<       }
1122,1123d1118
<       a: 'aa'
<       a: 'aa'
1144,1145d1138
<       b: 'bb'
<       b: 'bb'
1151d1143
<       c: 'cc'
1217d1208
<       d: 'dd'
1378d1368
<       id: '68e75c48-57ca-4a12-adfc-575c4b05fcbe'
1564d1553
<       time: '1703188212'
1575d1563
<       type: 'k8s_single_step'
1679,1682d1666
<     })
<     })
<     })
<     })
1847d1830
<     expect(config.tags).to.have.property('foo', 'bar')
1888,1889c1871,1872
<     expect(config).to.have.deep.property('installSignature', {
<     expect(config).to.have.deep.property('peerServiceMapping', {
---
>     expect(config).to.have.deep.property('installSignature'
>     expect(config).to.have.deep.property('peerServiceMapping', { c: 'cc', d: 'dd' })
1892,1893c1875,1876
<     expect(config).to.have.deep.property('serviceMapping', {
<     expect(config).to.have.deep.property('serviceMapping', {
---
>     expect(config).to.have.deep.property('serviceMapping', { a: 'aa', b: 'bb' })
>     expect(config).to.have.deep.property('serviceMapping', { a: 'aa', b: 'bb' })
```

watson:
Just force pushed to try and trigger CI again as there's a bug in CI that might be fixable by re-running from a fresh SHA. No code-changes was added in the force-push

BridgeAR (approved):
I don't think this provides a big benefit but it does not make things worse either.

PR #5901, opened 2025-06-16
Detect destructuring process.env and use the config-helper instead

BridgeAR:


eslint-rules/eslint-process-env.mjs
```diff
@@ -23,27 +23,59 @@ export default {
...
-            isProcessEnvObject(node.object)) {
+        // direct `process.env`
+        if (isProcessEnvObject(node)) {
+          report(node)
+        }
+        // nested `process.env.FOO`
+        else if (
+          node.object?.type === 'MemberExpression' &&
+          isProcessEnvObject(node.object)
+        ) {
```
watson:
```suggestion
        if (isProcessEnvObject(node)) {
          // direct `process.env`
          report(node)
        } else if (
          node.object?.type === 'MemberExpression' &&
          isProcessEnvObject(node.object)
        ) {
          // nested `process.env.FOO`
  ```

watson: I'd fix the linting error like this. And same for the rest below

tlhunter:
I think this PR needs a description.

I like the idea of disallowing 100% of process.env usage and requiring that all access be done through a single location (e.g. config). But I don't think this PR does that.

watson (approved):

PR #5903, opened 2025-06-17
Remove SSI profiler heuristics telemetry

szegedi:
### What does this PR do?
* Removes SSI profiler heuristics telemetry.
* Removes recognition of `DD_INJECTION_ENABLED=profiler`, as it was never implemented in the injection platform and is superseded by Stable APM Configuration through #5126.

(These look unrelated, but aren't, both were initial SSI features and the env var did also affect SSI profiler operation.)

### Motivation
SSI heuristics telemetry is not used anymore, and `DD_INJECTION_ENABLED=profiler` never obtained support in the injection platform , thus removing the code for these unused features reduces maintenance burden.

### Additional Notes
Jira: [PROF-11969]

[PROF-11969]: https://datadoghq.atlassian.net/browse/PROF-11969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dd-trace/src/profiling/exporter_cli.js
```diff
@@ -24,9 +24,7 @@ function exporterFromURL (url) {
     ? 'manual'
     : profilingEnabled === 'auto'
       ? 'auto'
-      : injectionEnabled.includes('profiling')
```
nsavoire, line 27: It was not even correct (`profiling` instead of `profiler`) 😅

szegedi: 🤣

PR #5906, opened 2025-06-17
feat(profiling): Start profilers synchronously within tracer initialization

szegedi:
### What does this PR do?
Start profilers synchronously within tracer initialization.

### Motivation
Profiler was started asynchronously because of source map loading – it uses async file IO. There are drawbacks to starting the profiler asynchronously though. A minor issue is that it won't start capturing CPU samples for the root program. A larger issue is that any long-lived allocations made after tracer init in the root program also won't ever be shown in the heap live size snapshots. We've had customers complaining about this.

### Additional Notes
* Source maps are still loaded asynchronously. Since they aren't needed until the first time profiles are emitted (65s into the runtime) they should have plenty of time to get loaded.
* We moved initialization of compression functions (thus, loading of zlib etc.) also out of the synchronous startup path as they, too, are only required first time profiles are emitted.

Jira: [PROF-14055](https://datadoghq.atlassian.net/browse/PROF-14055)

[PROF-14055]: https://datadoghq.atlassian.net/browse/PROF-14055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/src/profiling/profiler.js
+++ b/packages/dd-trace/src/profiling/profiler.js
@@ -352,12 +352,12 @@ class Profiler extends EventEmitter {
 
       const encodedProfiles = {}
       const infos = this.#createInitialInfos()
+      const compressionFn = this.#getCompressionFn()
 
       // encode and export asynchronously
       await Promise.all(profiles.map(async ({ profiler, profile, info }) => {
         try {
           const encoded = await profiler.encode(profile)
-          const compressionFn = this.#getCompressionFn()
           const compressed = encoded instanceof Buffer && compressionFn !== undefined
             ? await compressionFn(encoded, this.#compressionOptions)
             : encoded


packages/dd-trace/src/profiling/profiler.js
```diff
@@ -337,8 +357,9 @@ class Profiler extends EventEmitter {
       await Promise.all(profiles.map(async ({ profiler, profile, info }) => {
         try {
           const encoded = await profiler.encode(profile)
-          const compressed = encoded instanceof Buffer && this.#compressionFn !== undefined
-            ? await this.#compressionFn(encoded, this.#compressionOptions)
+          const compressionFn = this.#getCompressionFn()
```
IlyasShabi: Nit: `#getCompressionFn()` returns the same function for all profilers. Maybe  consider calling it before the `Promise.all`

PR #5909, opened 2025-06-17
wrapCommandQueueClass fix super call

johnhuffsmith:
### What does this PR do?

Fix the super call on wrapCommandQueueClass

### Motivation

Currently breaks redis pub/sub on node-redis 5+ (and is probably causing other unexpected issues)

### DIFF
--- a/packages/datadog-instrumentations/src/redis.js
+++ b/packages/datadog-instrumentations/src/redis.js
@@ -39,8 +39,8 @@ function wrapAddCommand (addCommand) {
 
 function wrapCommandQueueClass (cls) {
   const ret = class RedisCommandQueue extends cls {
-    constructor () {
-      super(arguments)
+    constructor (...args) {
+      super(...args)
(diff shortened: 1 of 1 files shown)


BridgeAR (approved):
Thanks, the change is LGTM!
Would you mind adding a regression test as well?

We also can't directly accept the change, since we can't run the CI on external code but we'd port the change over.

johnhuffsmith:
> Thanks, the change is LGTM! Would you mind adding a regression test as well?
> 
> We also can't directly accept the change, since we can't run the CI on external code but we'd port the change over.

Not sure where to start with adding a regression test (not very familiar with the inner working of this project). Is there a similar test I can reference in the codebase?

BridgeAR:
I had to look into it to figure out where to place the test and how it might look like, so I went ahead and wrote the test already.

I opened the port of your PR, so thanks again for your contribution!

PR #5910, opened 2025-06-18
Allow blocking on fastify cookie

IlyasShabi:
### What does this PR do?
Allow blocking on fastify cookie @fastify/cookie library

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/fastify.js
+++ b/packages/datadog-instrumentations/src/fastify.js
@@ -58,7 +58,9 @@ function wrapAddHook (addHook) {
           arguments[arguments.length - 1] = function (err) {
             publishError(err, req)
 
-            if (cookieParserReadCh.hasSubscribers && !cookiesPublished.has(req)) {
+            const hasCookies = request.cookies && Object.keys(request.cookies).length > 0
+
+            if (cookieParserReadCh.hasSubscribers && hasCookies && !cookiesPublished.has(req)) {
               const res = getRes(reply)
               const abortController = new AbortController()
 
--- a/packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
+++ b/packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
@@ -470,6 +470,9 @@ withVersions('fastify', 'fastify', '>=2', version => {
               hook
             })
 
+            // Dummy hook
+            app.addHook('onRequest', (req, reply, done) => done())
+
             app.post('/', (request, reply) => {
               requestCookie()
               reply.send('DONE')


packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -4,13 +4,12 @@ const Axios = require('axios')
...
 const getPort = require('get-port')
 const path = require('path')
-const semver = require('semver')
 const agent = require('../plugins/agent')
 const appsec = require('../../src/appsec')
 const Config = require('../../src/config')
 const { json } = require('../../src/appsec/blocked_templates')
 
-withVersions('fastify', 'fastify', version => {
+withVersions('fastify', 'fastify', '>=2', version => {
```
IlyasShabi, line 12: since we support fastify >=2

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -45,16 +46,35 @@ function wrapAddHook (addHook) {
...
-        if (typeof done === 'function') {
-          done = arguments[arguments.length - 1]
+        // done callback is always the last argument
+        const done = arguments[arguments.length - 1]
 
+        if (typeof done === 'function') {
           arguments[arguments.length - 1] = function (err) {
             publishError(err, req)
 
+            const hasCookies = request.cookies && Object.keys(request.cookies).length > 0
```
simon-id, line 61: eeeeh, usually we check that in the listener instead, not sure it makes a big difference, but when debugging it's nice to still have the DC event fired, even if it is with empty object

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -45,16 +46,35 @@ function wrapAddHook (addHook) {
...
+    arguments[arguments.length - 1] = shimmer.wrapFunction(fn, fn => function () {
+      const request = arguments[0]
+      const reply = arguments[1]
       const req = getReq(request)
 
       try {
-        if (typeof done === 'function') {
-          done = arguments[arguments.length - 1]
+        // done callback is always the last argument
+        const done = arguments[arguments.length - 1]
```
simon-id: why add this ? there can be more arguments between reply and done ?
if that's the case, there was a bug before no ?

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -45,16 +46,35 @@ function wrapAddHook (addHook) {
 
     if (typeof fn !== 'function') return addHook.apply(this, arguments)
 
-    arguments[arguments.length - 1] = shimmer.wrapFunction(fn, fn => function (request, reply, done) {
+    arguments[arguments.length - 1] = shimmer.wrapFunction(fn, fn => function () {
```
simon-id: if you remove the arguments from the function signature, you potentially break any behavior in the framework, that might be checking "func.length" to detect the type of callback it is and what arguments to pass it

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -4,15 +4,14 @@ const Axios = require('axios')
...
-const semver = require('semver')
 const zlib = require('zlib')
 const fs = require('node:fs')
 const agent = require('../plugins/agent')
 const appsec = require('../../src/appsec')
 const Config = require('../../src/config')
 const { json } = require('../../src/appsec/blocked_templates')
 
-withVersions('fastify', 'fastify', version => {
+withVersions('fastify', 'fastify', '>=2', version => {
```
simon-id, line 14: aren't you removing coverage for fastify v1 by doing this ?

packages/dd-trace/test/plugins/externals.json
```diff
@@ -151,6 +151,10 @@
     {
       "name": "middie",
       "versions": ["5.1.0"]
+    },
+    {
+      "name": "@fastify/cookie",
+      "versions": [">=6"]
```
simon-id, line 157: i was gonna say "can't we support older versions ?" but this package started at v6.0.0 lol

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
IlyasShabi: We do not support fastify v1, I will add this to the documentation

packages/datadog-instrumentations/src/fastify.js
IlyasShabi: I don't want to create an `AbortController` as it's an expensive operation

simon-id: I talked with Ruben in slack about this last time, and from what I understand, it's really not THAT expensive, it's just a new object, nothing crazy

simon-id: I think his definition of "expensive" wasn't what we understood from it

packages/datadog-instrumentations/src/fastify.js
uurien: it is checked 2 lines above

simon-id: what ? I think you're misunderstanding my comment

simon-id: he removed the request, reply, done arguments from the function signature, thus if code in fastify checks for this callback length, it will be wrong no ?

uurien: yes sorry, I misunderstood it. you're right.

packages/datadog-instrumentations/src/fastify.js
IlyasShabi: Hook could have 3 or 4 args https://fastify.dev/docs/latest/Reference/Hooks
Could you check now

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -49,26 +51,44 @@ function wrapAddHook (addHook) {
...
-        if (typeof done === 'function') {
-          done = arguments[arguments.length - 1]
+        // done callback is always the last argument
+        const doneCallback = arguments[arguments.length - 1]
 
+        if (typeof doneCallback === 'function') {
           arguments[arguments.length - 1] = function (err) {
             publishError(err, req)
 
+            if (cookieParserReadCh.hasSubscribers && !cookiesPublished.has(req)) {
```
uurien: why does this work?
we are executing this peace of code each time that the hook is executed, this only should work if the cookies are parsed the first time a hook is executed.
If the cookies hook is the second one, we will mark add the request in `cookiesPublished` and we will ignore the good one. To make this work, we should check also if the cookies object has something, as you did in your first implementation.

IlyasShabi: It works because the first hook have already cookies parsed, but adding
```
app.addHook('onRequest', (req, reply, done) => done())
```
Will break the `onRequest` test.
I fixed the implementation and improve tests

packages/datadog-instrumentations/src/fastify.js
simon-id: so that means this is a bug existing in master no ?

IlyasShabi: Yes there is a bug on master with `preParsing` hook

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
simon-id: why not ? the instrumentation is there no ?

IlyasShabi: Fasitfy v1 is missing a lot of hooks like `preValidation` and `preParsing` that we use to block a request + we have 0 customer using fastify v1

simon-id (approved):

PR #5913, opened 2025-06-18
add event loop and gc configuration for runtime metrics

rochdev:
### What does this PR do?

Add event loop and GC configuration for runtime metrics.

### Motivation

Allow finer control on the types of metrics to collect, especially on the native front. Allowing to pick which GC collector to use can also be used to work around issues with one of the collector by temporarily switching to a different one.


packages/dd-trace/src/config.js
```diff
@@ -567,7 +573,10 @@ class Config {
     defaults['remoteConfig.enabled'] = true
     defaults['remoteConfig.pollInterval'] = 5 // seconds
     defaults.reportHostname = false
-    defaults.runtimeMetrics = false
+    defaults['runtimeMetrics.enabled'] = false
+    defaults['runtimeMetrics.eventLoop'] = true
+    defaults['runtimeMetrics.gc'] = true
+    defaults['runtimeMetrics.gcCollector'] = 'default'
```
szegedi: Does it make sense to have these as two separate config properties? You could have just one (just keep `.gc`?) and introduce an `'off'` value (and maybe use `'on'` instead of `'default'` for consistency.) You could still have config translate incoming env var values of '0', '1', 'true', 'false' to off and on. Of course, for consistency you should then probably allow use of off and on as keywords for the other two properties a well (`enabled` and `eventLoop`.) Maybe we should just add `on` and `off` as recognized values in [util.js](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/util.js#L5-L13).

We recently added a config for choosing the compression method for uploading profiles and use on/off there in addition to specifically naming compression methods.

Of course, if you envision there could be even more config options for GC metrics in the future, then it may make sense to keep them as separate properties, but we should assess how likely is that.

BridgeAR (commented):
I did not yet properly look into it. I just wonder if we really want to support different implementations almost indefinitely, since that is likely the outcome for `gcCollector` (or we would have to make it a noop).

packages/dd-trace/src/config.js
tlhunter: I'm personally not a fan of "enabled" (or "disabled") as key names, only as values, but we do it everywhere so that'll never change.

But in this case it looks like the gc runtime metrics are either enabled or disabled, so for consistency maybe it should looks more like this:

```
    defaults['runtimeMetrics.gc.enabled'] = true
    defaults['runtimeMetrics.gc.collector'] = 'default'
```

Otherwise I'm not sure if we want to break the `enabled=true` pattern that we have everywhere.

rochdev: > Does it make sense to have these as two separate config properties?

I think it does, otherwise, in order to disable GC metrics you would have to lose the information about what collector you want to use, and then figure out that value again to re-enable them. Keeping them separate keeps their purposes usable on their own without impacting each other.

rochdev:
> I did not yet properly look into it. I just wonder if we really want to support different implementations almost indefinitely, since that is likely the outcome for gcCollector (or we would have to make it a noop).

I don't think we'd want to keep them indefitely, but they are likely to stay for at least a few years, and after that we can indeed deprecate them and make configuring any implementation that no longer exists fall back to the default.

packages/dd-trace/src/config.js
rochdev: > But in this case it looks like the gc runtime metrics are either enabled or disabled, so for consistency maybe it should looks more like this:

Makes sense, I'll make the change. Would you also rename the environment variables at that point?

szegedi: > I think it does, otherwise, in order to disable GC metrics you would have to lose the information about what collector you want to use, and then figure out that value again to re-enable them. 

I can buy that argument.

szegedi: > Would you also rename the environment variables at that point?

I think yes – we generally add `_ENABLED` at the end of the name of an env var that enables a functionality, so:
```
DD_RUNTIME_METRICS_EVENT_LOOP_ENABLED
DD_RUNTIME_METRICS_GC_ENABLED
```

rochdev: Renamed the environment variables as well 👍

tlhunter (dismissed):

packages/dd-trace/src/runtime_metrics/runtime_metrics.js
```diff
@@ -34,15 +34,27 @@ reset()
...
-      if (hasGCObserver) {
-        nativeMetrics.start('loop') // Only add event loop watcher and not GC.
+      if (hasGCProfiler && (gcCollector === 'profiler' || gcCollector === 'default')) {
+        startGCProfiler()
+      } else if (hasGCObserver && (gcCollector === 'observer' || gcCollector === 'default')) {
+        startGCObserver()
       } else {
-        nativeMetrics.start()
+        watchers.push('gc')
       }
```
BridgeAR:
```suggestion
      if (hasGCProfiler) {
        if (gcCollector === 'profiler' || gcCollector === 'default') {
          startGCProfiler()
        } else {
          startGCObserver()
        }
      } else {
        watchers.push('gc')
      }
```

The condition checked in two different spots for default. That is definitely off.

BridgeAR (approved):
LGTM

PR #5920, opened 2025-06-20
add automatic heap snapshots to help debug memory issues

rochdev:
### What does this PR do?

Add automatic heap snapshots to help debug memory issues.

### Motivation

Sometimes heap dumps are needed to debug memory issues. This can be annoying for users as they have to manually add code to generate the heap dumps, and sometimes they are not generated correctly (wrong interval, only one sample, without garbage collection, etc). This change adds the ability to generate heap snapshots directly from dd-trace by simply passing an environment variable, making the whole process a lot easier.

### Additional Notes

@BridgeAR I know you have been working on logic to write the heap snapshot manually so that we can also send it elsewhere, let me know if you want me to add it to this PR, otherwise I think it could be a future improvement.

I thought about using telemetry for this but I don't like the security implications of sending a full memory dump directly to us.


packages/dd-trace/src/heap_snapshots.js
```diff
@@ -0,0 +1,62 @@
...
+      const log = require('./log')
+      const folder = config.heapSnapshot.folder
+
+      try {
+        await scheduleSnapshot(config, 1)
+        log.debug('Wrote heap snapshots to %s.', folder)
+      } catch (e) {
+        log.error('Failed to write heap snapshots to %s.', folder, e)
+      }
+    }
```
watson: Nit: Personally, I like to reverse the if-statement and return instead of indenting everything in one large block, but feel free to ignore

```suggestion
    if (config.heapSnapshot.count === 0 || !globalThis.gc) return
    const log = require('./log')
    const folder = config.heapSnapshot.folder

    try {
      await scheduleSnapshot(config, 1)
      log.debug('Wrote heap snapshots to %s.', folder)
    } catch (e) {
      log.error('Failed to write heap snapshots to %s.', folder, e)
    }
```

packages/dd-trace/src/heap_snapshots.js
```diff
@@ -0,0 +1,62 @@
...
+  writeHeapSnapshot(getName(config.heapSnapshot.folder))
+  await scheduleSnapshot(config, total + 1)
+}
+
+async function clearMemory () {
+  const { setImmediate } = require('timers/promises')
+
+  globalThis.gc()
+  await setImmediate()
+  globalThis.gc() // Run full GC a second time for anything missed in first GC.
```
watson, line 23: If this is a know issue, I prefer if we reference a bug report/documentation/blog post specifying why this is required

packages/dd-trace/src/heap_snapshots.js
```diff
@@ -0,0 +1,62 @@
...
+    'Heap-%s%s%s-%s%s%s-%s-%s.heapsnapshot',
+    date.getFullYear(),
+    pad(date.getMonth()),
+    pad(date.getDate()),
+    pad(date.getHours()),
+    pad(date.getMinutes()),
+    pad(date.getSeconds()),
+    process.pid,
+    threadId
+  )
```
watson, line 42: Unfortunate that we have to replicate this from V8 just because we want to specify the folder. It would be nice if specifying a folder without having to specify the filename was supported by V8

packages/dd-trace/test/heap_snapshots.spec.js
```diff
@@ -0,0 +1,34 @@
+'use strict'
+
+require('./setup/tap')
+
+const { mkdtempSync, readdirSync } = require('fs')
+const { tmpdir } = require('os')
+const { join } = require('path')
+const { start } = require('../src/heap_snapshots')
+
+const folder = mkdtempSync(join(tmpdir(), 'dd-trace-heap-snapshot-'))
```
watson: Not that it matters, but I assume the trailing `-` was a mistake?

```suggestion
const folder = mkdtempSync(join(tmpdir(), 'dd-trace-heap-snapshot'))
```

packages/dd-trace/test/heap_snapshots.spec.js
```diff
@@ -0,0 +1,34 @@
...
+      heapSnapshot: {
+        count: 2,
+        folder,
+        interval: 1
+      }
+    })
+
+    clearInterval(interval)
+
+    const pattern = /^Heap-\d{8}-\d{6}-\d+-\d+\.heapsnapshot$/
```
watson: Not a big deal, but just a slight improvement to the validation now that you _are_ generating the filename manually (where `threadId` is imported at the top of the spec file). This would also catch cases where somehow files are reused between runs for some reason

```suggestion
    const pattern = new RegExp(`^Heap-\\d{8}-\\d{6}-${process.pid}-${threadId}\\.heapsnapshot$`)
```

packages/dd-trace/src/config.js
```diff
@@ -507,6 +507,9 @@ class Config {
     defaults['grpc.client.error.statuses'] = GRPC_CLIENT_ERROR_STATUSES
     defaults['grpc.server.error.statuses'] = GRPC_SERVER_ERROR_STATUSES
     defaults.headerTags = []
+    defaults['heapSnapshot.count'] = 0
```
BridgeAR, line 510: I would just use a hardcoded default, since changing the duration between each should be customization enough. I would therefore remove this option.

I would just use a default of e.g., 4-5 snapshots. Adding this as feature like that might however be tricky, since each instance would now create these. I would rather have this functionality being triggered by our remote configuration for a few dedicated services and only for one instance of these. Otherwise the customer would create lots of snapshots due to the potential amount of instances running.

packages/dd-trace/src/config.js
```diff
@@ -507,6 +507,9 @@ class Config {
...
+    defaults['heapSnapshot.folder'] = undefined
```
BridgeAR: I think naming it option is limiting the possibility to change the destination later to e.g., a url. Thus, I would use

```suggestion
    defaults['heapSnapshot.destination'] = undefined
```

As name

packages/dd-trace/src/heap_snapshots.js
BridgeAR: I do not know about a blog post but the GC is working async in background tasks. As such, calling something a full GC is tricky and calling it more than once helps.

packages/dd-trace/src/heap_snapshots.js
BridgeAR: That is actually the default

> If not specified, a file name with the pattern 'Heap-${yyyymmdd}-${hhmmss}-${pid}-${thread_id}.heapsnapshot'

https://nodejs.org/api/v8.html#v8writeheapsnapshotfilenameoptions

packages/dd-trace/src/heap_snapshots.js
```diff
@@ -0,0 +1,62 @@
...
+
+async function scheduleSnapshot (config, total) {
+  if (total > config.heapSnapshot.count) return
+
+  const { setTimeout } = require('timers/promises')
+  const { writeHeapSnapshot } = require('v8')
+
+  await setTimeout(config.heapSnapshot.interval * 1000, null, { ref: false })
+  await clearMemory()
+  writeHeapSnapshot(getName(config.heapSnapshot.folder))
```
BridgeAR: I would not write to the file system by default. A lot of applications won't be able to get that snapshot from their hdd easily.
As such, I would also allow to specify a URL instead to support them writing things to e.g., S3 with a token.

packages/dd-trace/src/heap_snapshots.js
```diff
@@ -0,0 +1,62 @@
+'use strict'
+
+async function scheduleSnapshot (config, total) {
+  if (total > config.heapSnapshot.count) return
+
+  const { setTimeout } = require('timers/promises')
+  const { writeHeapSnapshot } = require('v8')
```
BridgeAR: I would not call these in a method that is used in the recursion. Either place it at the top (which I think is fine. I would just only load the file if we indeed activate the feature instead of checking here if the feature is active) or lazy load it in the enablement code.

packages/dd-trace/src/heap_snapshots.js
```diff
@@ -0,0 +1,62 @@
...
+
+  await setTimeout(config.heapSnapshot.interval * 1000, null, { ref: false })
```
BridgeAR, line 13: I would wait a fixed time at the beginning and not use the regular timeout as starting point.

The fixed timeout should IMO be 5 minutes and just automatically adjust to earlier in case memory builds up faster.

packages/dd-trace/src/heap_snapshots.js
rochdev: Exactly, I know this from personal experience and because Node does it as well, but I'm not aware of any documentation about it.

packages/dd-trace/test/heap_snapshots.spec.js
rochdev: The trailing dash is needed so that the result is something like `dd-trace-heap-snapshot-e7a5e5` instead of `dd-trace-heap-snapshote7a5e5`.

packages/dd-trace/src/heap_snapshots.js
rochdev: We unfortunately can't as you have to specify the full path. This means we have to repeat the filename.

packages/dd-trace/src/config.js
rochdev: What would be the use case for sending a heap snapshot to an arbitrary URL?

packages/dd-trace/src/config.js
rochdev: > I would just use a hardcoded default, since changing the duration between each should be customization enough. I would therefore remove this option.

I think I tend to agree that this doesn't necessarily need to be configurable and it could always be added in the future if we need. We could instead have something like `heapSnapshot.enabled` and only send what we need (which is probably 2, I don't know why we're want any more than that if they're far apart enough).

> I would rather have this functionality being triggered by our remote configuration for a few dedicated services and only for one instance of these.

I don't know that remote config is the right place for this either. The best place in theory would be when requesting a tracer flare, but there are security implications with that given the whole memory is sent over, and also right now it applies to all services connected to the agent which is even worse than what you're describing. I see this pretty much equivalent to other flags like `DD_TRACE_DEBUG` which already exist as environment variables, so I think this is consistent even if not ideal, and it would allow us to get this much sooner.

packages/dd-trace/src/heap_snapshots.js
rochdev: > I would not write to the file system by default. A lot of applications won't be able to get that snapshot from their hdd easily.

Most application (in fact almost all of them in my experience) will be able to do this, so I think the gains from being able to quickly write a heap dump to a file is much higher than the potential problem of a write access issue.

> As such, I would also allow to specify a URL instead to support them writing things to e.g., S3 with a token.

This is a much more complicated solution that require a lot of setup from users. I prefer a simpler one-click solution even if it doesn't work in very rare edge cases. We can always expand to more complex options later on.

packages/dd-trace/src/heap_snapshots.js
rochdev: I thought about having a separate initial delay but I ended up not really seeing the value in that. What would be the benefit that you see and should it be configurable? (especially if the interval is configurable and could be less than 5 minutes)

szegedi: Agree with @rochdev here; if there's no separate initial delay setting then intuitively it should be equal to the regular snapshot period. 5 minutes feels like an arbitrary magic number.

packages/dd-trace/src/heap_snapshots.js
szegedi: I agree with @rochdev, it's more valuable to ship the initial version with file-writing capability and expand with a S3 bucket writing capability later.

packages/dd-trace/src/heap_snapshots.js
szegedi: I'm rather baffled by the need for this. I don't question it, as I don't have enough expertise in V8 GC, but I'm surprised. I would've expected the V8 heap snapshots either wouldn't include garbage, or if they do, an analysis tool loading them could still identify garbage as objects disconnected from GC roots and ignore it.

rochdev: To be honest, doing a GC here is not really strictly needed anyway. It's just done to remove any memory that can be recollected before the snapshot to avoid noise from normal memory usage as much as possible and make the snapshot smaller. In that sense a single pass would probably be more than enough, but at that point we might as well do 2 for good measure.

packages/dd-trace/src/heap_snapshots.js
BridgeAR: I would sometimes expect a timeout of e.g., an hour. To wait an hour for the first snapshot it quite a long time though. We could therefore use an upper bound of five or ten minutes instead. That would work for me.

rochdev: I'm still not convinced. If the interval is high, then the time between snapshots and thus the total time to get all the data will be high anyway. At that point why would we want to force the initial delay to be shorter? This is even more true when more than 2 snapshots are captured. It just seems a bit arbitrary as @szegedi mentioned.

packages/dd-trace/src/config.js
tlhunter: Considering the two main scenarios (app is running on a staging/prod service, or app is running on a dev machine) I suppose the answer is not, but...

Should we provide a default directory? E.g. most (all?) systems (including Lambda) support writing to /tmp. I suppose that would be annoying for a macOS developer who needs to navigate to the directory using Finder. They'd probably prefer the snapshots be written to, say, the root of the project.

tlhunter: I agree that destination is more future-proof than folder. While it makes sense for the first version to write to disk I imagine that if this feature is successful we'll later want different backends.

tlhunter (dismissed):
Overall I think this is fine in the current incantation. It seems pretty low-risk if we decide to completely overhaul how everything works in the future as this is just to aid with debugging. Changing an env var name or removing the feature outright shouldn't be a breaking change.

packages/dd-trace/src/config.js
BridgeAR: The URL could be a couple of things. Any blob storage in fact.

rochdev: The default is the root of the project which is why I left it undefined by default. I'm open to writing to other places, it just seemed the easiest to find by default.

packages/dd-trace/src/config.js
BridgeAR: I like to use more than two snapshots since it could also be e.g., a cron job that only runs from time to time that uses specific features or it could just generally happen at a later point. Having more in-between snapshots helps for that case.
I still worry about the amount of snapshots this will automatically create on customer side if not used with extreme care, no matter if it is only two or more per instance.

I would really like something triggering dedicated machines from our end.

packages/dd-trace/src/heap_snapshots.js
BridgeAR: The initial timeout is just for the application to get into its regular load. As soon as that's the case, I do want to get the first snapshot.

packages/dd-trace/src/heap_snapshots.js
BridgeAR: I totally agree that it should ship with the file writing capability. What I meant is that we could just use the stream and direct that stream to a file instead of using the helper to write to disk directly.

So:

`stream.promises.pipeline(v8.getHeapSnapshot(), fs.createWriteStream(path))`

instead of

`v8.writeHeapSnapshot(path)`

And yes, we can also change that as soon as we change the feature.

packages/dd-trace/src/heap_snapshots.js
BridgeAR: Hm... yeah, it would probably be a good idea to accept a folder in the Node.js API and just automatically write them there with the default name instead of having to pass through the full name.

szegedi (dismissed):

packages/dd-trace/src/heap_snapshots.js
rochdev: > What I meant is that we could just use the stream and direct that stream to a file instead of using the helper to write to disk directly.

I see this as out of scope as for now they are effectively equivalent. We can refactor this when we actually add other transports.

packages/dd-trace/src/heap_snapshots.js
rochdev: That makes assumptions about the application. What if the pod takes 10 minutes to become healthy? Now we'd be getting a heap snapshot before there was ever any load. I don't mind adding a delay configuration, but I also don't really want to make any assumption by default.

rochdev: @BridgeAR Can this be discussed in an improvement PR? I feel like this PR is important enough to debug active issues that it warrants moving forward with the simplest implementation and iterate later on anything we feel could be improved instead of blocking it.

szegedi (approved):

packages/dd-trace/src/heap_snapshots.js
```diff
@@ -0,0 +1,59 @@
...
+    pad(date.getMonth()),
+    pad(date.getDate()),
+    pad(date.getHours()),
+    pad(date.getMinutes()),
+    pad(date.getSeconds()),
+    process.pid,
+    threadId
+  )
+
+  return join(destination, filename)
```
IlyasShabi, line 44: According to the `config.js` the destination could be `undefined` and this will fail with a invalid type error I think

packages/dd-trace/src/heap_snapshots.js
BridgeAR: Yes, we can still change it later

packages/dd-trace/src/heap_snapshots.js
BridgeAR: True, in case the env is set to an empty string. In that case the helper converts it to `undefined` (which is rather unfortunate in my opinion but nothing to change now). @rochdev I guess we should address that before landing?

rochdev: The default value is now always a string.

BridgeAR: Ah, the default is replacing the undefined by the env? I was not aware about that. It is difficult to see that in the code, since it's just a side effect.

BridgeAR (approved):
LGTM

PR #5921, opened 2025-06-20
remove async storage

wconti27:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/redis.js
+++ b/packages/datadog-instrumentations/src/redis.js
@@ -2,8 +2,7 @@
 
 const {
   channel,
-  addHook,
-  AsyncResource
+  addHook
 } = require('./helpers/instrument')
 const shimmer = require('../../datadog-shimmer')
 
@@ -22,15 +21,11 @@ function wrapAddCommand (addCommand) {
     const name = command[0]
     const args = command.slice(1)
 
-    const asyncResource = new AsyncResource('bound-anonymous-fn')
-    return asyncResource.runInAsyncScope(() => {
-      start(this, name, args, this._url)
-
+    const ctx = getStartCtx(this, name, args, this._url)
+    return startCh.runStores(ctx, () => {
       const res = addCommand.apply(this, arguments)
-      const onResolve = asyncResource.bind(() => finish(finishCh, errorCh))
-      const onReject = asyncResource.bind(err => finish(finishCh, errorCh, err))
 
-      res.then(onResolve, onReject)
+      res.then(() => finish(finishCh, errorCh, ctx), err => finish(finishCh, errorCh, ctx, err))
 
       return res
     })
@@ -94,14 +89,9 @@ addHook({ name: 'redis', versions: ['>=2.6 <4'] }, redis => {
 
     if (!options.callback) return internalSendCommand.apply(this, arguments)
 
-    const callbackResource = new AsyncResource('bound-anonymous-fn')
-    const asyncResource = new AsyncResource('bound-anonymous-fn')
-    const cb = callbackResource.bind(options.callback)
-
-    return asyncResource.runInAsyncScope(() => {
-      start(this, options.command, options.args)
-
-      options.callback = asyncResource.bind(wrapCallback(finishCh, errorCh, cb))
+    const ctx = getStartCtx(this, options.command, options.args)
+    return startCh.runStores(ctx, () => {
+      options.callback = wrapCallback(finishCh, errorCh, ctx, options.callback)
 
       try {
         return internalSendCommand.apply(this, arguments)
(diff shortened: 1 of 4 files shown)


packages/datadog-plugin-redis/src/index.js
```diff
@@ -13,10 +13,14 @@ class RedisPlugin extends CachePlugin {
...
 
-  start ({ db, command, args, connectionOptions = {}, connectionName }) {
+  bindStart (ctx) {
+    const { db, command, args, connectionOptions, connectionName } = ctx
+
     const resource = command
     const normalizedCommand = command.toUpperCase()
-    if (!this.config.filter(normalizedCommand)) return this.skip()
+    if (!this.config.filter(normalizedCommand)) {
+      return this.skip(ctx)
```
BridgeAR:
```suggestion
      return this.skip()
```

If changed, `ctx` is always truthy and the method would immediately return that without doing anything else, if I read the correct. Looking at other implementations, this should likely not change (see e.g., fs plugin).

packages/datadog-plugin-redis/src/index.js
```diff
@@ -13,10 +13,14 @@ class RedisPlugin extends CachePlugin {
     this._spanType = 'redis'
   }
 
-  start ({ db, command, args, connectionOptions = {}, connectionName }) {
+  bindStart (ctx) {
+    const { db, command, args, connectionOptions, connectionName } = ctx
```
BridgeAR, line 17: Nice spot that `connectionOptions` are always defined by using default argument in `url` above!

packages/dd-trace/src/plugins/plugin.js
```diff
@@ -70,7 +70,9 @@ module.exports = class Plugin {
 
   // TODO: Implement filters on resource name for all plugins.
   /** Prevents creation of spans here and for all async descendants. */
-  skip () {
+
+  skip (ctx = null) {
+    if (ctx) return ctx
```
BridgeAR:
```suggestion
  skip () {
```

As follow-up of the comment above.

wconti27: I removed the `this.skip()` call, I don't believe we need it. @rochdev Can you confirm whether we need the `this.skip` call within the redis plugin? It simply does `storage('legacy').enterWith({ noop: true })`, which was preventing the later redis calls from being traced despite not being on the blocklist.

packages/datadog-instrumentations/src/redis.js
```diff
@@ -149,24 +134,28 @@ addHook({ name: 'redis', versions: ['>=0.12 <2.6'] }, redis => {
...
+  return {
+    db: client.selected_db,
+    command,
+    args,
+    connectionOptions: client.connection_options || client.connection_option || client.connectionOption || url
+  }
 }
 
-function wrapCallback (finishCh, errorCh, callback) {
+function wrapCallback (finishCh, errorCh, ctx, callback) {
```
rochdev, line 146: It doesn't look like the parent context is restored in the callback. There is probably no test for this which is why everything is passing, but the callback should be wrapped with `finishCh.runStore` and a corresponding `bindFinish` handler on the subscriber side.

rochdev (changes requested):

packages/datadog-instrumentations/src/redis.js
wconti27: okay, i made that change and added a test for the parent context restoration.

wconti27: Also, can you check / respond to the thread above please.

packages/datadog-instrumentations/src/redis.js
```diff
@@ -149,24 +134,28 @@ addHook({ name: 'redis', versions: ['>=0.12 <2.6'] }, redis => {
...
   if (error) {
-    errorCh.publish(error)
+    ctx.error = error
+    errorCh.publish(ctx)
+  }
+  if (callback) {
+    return finishCh.runStores(ctx, callback, thisArg, ...args)
   }
-  finishCh.publish()
+  finishCh.publish(ctx)
```
rochdev, line 160: This will call `publish` a second time since the previous line also calls `publish` through `runStores`.

packages/dd-trace/src/plugins/plugin.js
rochdev: It's needed as it's what prevents new spans from being created. Without it, the context will be undefined but new spans will still be created, just in a new trace. The actual `skip` call itself is not needed in a `bind` function though as you can just `return { noop: true }` instead.

packages/datadog-instrumentations/src/redis.js
wconti27: how? the other line above returns

packages/dd-trace/src/plugins/plugin.js
wconti27: okay, makes sense, i made the change to return `{ noop: true }`

packages/datadog-instrumentations/src/redis.js
rochdev: Oh, you're right I totally missed that this only happens when there is no callback. Disregard the above.

packages/dd-trace/src/plugins/outbound.js
```diff
@@ -27,6 +27,7 @@ class OutboundPlugin extends TracingPlugin {
   }
 
   bindFinish (ctx) {
+    this.finish(ctx)
```
rochdev: This should already happen automatically as `runStores` also calls `publish` so `bindFinish` is always followed by `finish`.

packages/dd-trace/src/plugins/plugin.js
```diff
@@ -70,6 +70,7 @@ module.exports = class Plugin {
 
   // TODO: Implement filters on resource name for all plugins.
   /** Prevents creation of spans here and for all async descendants. */
+
```
rochdev: Unrelated change that could cause conflicts down the line, better to remove this from the PR instead.

packages/dd-trace/src/plugins/outbound.js
wconti27: ah good catch, understood

packages/dd-trace/src/plugins/plugin.js
wconti27: removing!

rochdev (approved):

PR #5927, opened 2025-06-23
Use proper API to get number of bytes in data

watson:
### What does this PR do?

Relying on string length to calculate the number of bytes in the string is incorrect as the `length` property returns number of characters - not the number of bytes those characters take up.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes

Reviewers should be careful to check the input to `Buffer.byteLength()` as it will throw un unsupported input, whereas `input.length` will in most cases just return `undefined`. However, while the old behavior is not throwing, it's most likely still incorrect and propably means that the input is of an unexpected type.


packages/dd-trace/src/ci-visibility/exporters/agentless/writer.js
```diff
@@ -48,7 +48,7 @@ class Writer extends BaseWriter {
     const startRequestTime = Date.now()
 
     incrementCountMetric(TELEMETRY_ENDPOINT_PAYLOAD_REQUESTS, { endpoint: 'test_cycle' })
-    distributionMetric(TELEMETRY_ENDPOINT_PAYLOAD_BYTES, { endpoint: 'test_cycle' }, data.length)
+    distributionMetric(TELEMETRY_ENDPOINT_PAYLOAD_BYTES, { endpoint: 'test_cycle' }, Buffer.byteLength(data))
```
watson, line 51: Can we be sure `data` is always compatible with `Buffer.byteLength()`? If not, I'm not sure calling `.length` either is correct.

For example, the `CoverageCIVisibilityEncoder` class has a `makePayload` function which returns an instance of `FormData`. Can that theoretically end up here?

packages/dd-trace/test/ci-visibility/exporters/agentless/writer.spec.js
```diff
@@ -23,7 +23,7 @@ describe('CI Visibility Writer', () => {
     encoder = {
       encode: sinon.stub(),
       count: sinon.stub().returns(0),
-      makePayload: sinon.stub().returns([])
+      makePayload: sinon.stub().returns(Buffer.from(''))
```
watson, line 26: Dear reviewer: I'm pretty sure that this `makePayload` mock should return a buffer and not an array - please verify

simon-id (dismissed):
RSLGTM

packages/dd-trace/test/ci-visibility/exporters/agentless/writer.spec.js
BridgeAR: Verified

packages/dd-trace/src/ci-visibility/exporters/agentless/writer.js
BridgeAR: I don't think so. It would definitely fail for these cases as well, so we should be good here.

BridgeAR (dismissed):

packages/datadog-plugin-openai/src/tracing.js
```diff
@@ -687,7 +687,9 @@ function commonFineTuneResponseExtraction (tags, body) {
 // the OpenAI package appears to stream the content download then provide it all as a singular string
 function downloadFileResponseExtraction (tags, body) {
   if (!body.file) return
-  tags['openai.response.total_bytes'] = body.file.length
+  tags['openai.response.total_bytes'] = typeof body.file === 'string'
+    ? Buffer.byteLength(body.file)
+    : body.file.length // TODO: This will most likely return undefined and is probably not what we want
```
BridgeAR: I think we could even go and always use `byteLength()` here. Because if it's not a string, it should be a buffer which should also be fine.

sabrenner: it's possible for `body.file` to be a `Response` (in newer OpenAI versions). i think we should just add a check to skip this tag if so. getting the byte length would require awaiting `response.arrayBuffer()`, which I don't think we can do here, and we are already just tagging `undefined` which gets dropped anyways, but doing `Buffer.byteLength` will not submit the span because it will throw (although it does just get caught and logged)

```
Error: Error in plugin handler:
    at wrappedHandler (/Users/sam.brenner/dd/dd-trace-js/packages/dd-trace/src/plugins/plugin.js:83:16)
    at Subscription._handler (/Users/sam.brenner/dd/dd-trace-js/packages/dd-trace/src/plugins/plugin.js:15:9)
    at Channel.publish (node:diagnostics_channel:141:9)
    at finish (/Users/sam.brenner/dd/dd-trace-js/packages/datadog-instrumentations/src/openai.js:388:15)
    at /Users/sam.brenner/dd/dd-trace-js/packages/datadog-instrumentations/src/openai.js:362:9
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async main (/Users/sam.brenner/Development/ml-obs/sandboxes/vercel-ai-tests/openai.js:12:20)
TypeError [ERR_INVALID_ARG_TYPE]: The "string" argument must be of type string or an instance of Buffer or ArrayBuffer. Received an instance of Response
    at new NodeError (node:internal/errors:405:5)
    at Function.byteLength (node:buffer:779:11)
    at downloadFileResponseExtraction (/Users/sam.brenner/dd/dd-trace-js/packages/datadog-plugin-openai/src/tracing.js:689:56)
    at responseDataExtractionByMethod (/Users/sam.brenner/dd/dd-trace-js/packages/datadog-plugin-openai/src/tracing.js:573:7)
    at OpenAiTracingPlugin.asyncEnd (/Users/sam.brenner/dd/dd-trace-js/packages/datadog-plugin-openai/src/tracing.js:246:5)
    at /Users/sam.brenner/dd/dd-trace-js/packages/dd-trace/src/plugins/tracing.js:77:22
    at wrappedHandler (/Users/sam.brenner/dd/dd-trace-js/packages/dd-trace/src/plugins/plugin.js:81:24)
    at Subscription._handler (/Users/sam.brenner/dd/dd-trace-js/packages/dd-trace/src/plugins/plugin.js:15:9)
    at Channel.publish (node:diagnostics_channel:141:9)
    at finish (/Users/sam.brenner/dd/dd-trace-js/packages/datadog-instrumentations/src/openai.js:388:15) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

Suggestion:

```javascript
if (!body.file || body.file instanceof Response) return
```

I believe with this we should otherwise be fine (and in any case, I don't think our instrumentation of the files content method is too popular anyhow 😂)

sabrenner (changes requested):

packages/dd-trace/src/ci-visibility/exporters/agentless/writer.js
watson: No, in case data is an a random object, calling `.length` just returns `undefined` most likely. This will not throw, and the program will not crash. Calling `Buffer.byteLength(randomObject)` will throw. That's what I'm worried about

packages/datadog-plugin-openai/src/tracing.js
watson: Do you mean the global `Response` class?

sabrenner: yes ([ref](https://github.com/openai/openai-node/blob/ab73eaf4b5432cc51fa00a2656a2a610c86371a2/src/resources/files.ts#L69) - this is what this function is tagging the response for. it used to resolve directly to a string but it is now a response)

watson: Unfortunately OpenAI is using it's own `node-fetch` polyfill, so we can't rely on the global `Response` to do a `instanceof`. We can't just require `node-fetch` and use its `Response` either, as we might get a different one depending on how the require call resolves the dependency.

I think the best I can do is the following:

```js
function downloadFileResponseExtraction (tags, body) {
  if (typeof body.file !== 'string') return
  tags['openai.response.total_bytes'] = Buffer.byteLength(body.file)
}
```

packages/dd-trace/src/ci-visibility/exporters/agentless/writer.js
watson: Tracing back the origin of `data`, it's the result from calling `this._encoder.makePayload()` in `packages/dd-trace/src/exporters/common/writer.js`. This file contains the `BaseWriter` class used by ci-visibility in 3 locations - all of which implement their own `_encoder.makePayload` methods. Each of these `makePayload` methods return a `Buffer`, `string` or `FormData` object respectively. I can't rule out (@juan-fernandez will have to verify this) if we can be sure that `data` will only be in the form of either the `Buffer` or the `string` in this case (tests indicate it's just a `Buffer`, but there might be gaps in the tests), or if it can in some circomstances also be a `FormData` object (in which case calling `Buffer.byteLength` will throw).

packages/datadog-plugin-openai/src/tracing.js
watson: I updated the PR with the above mentioned change

packages/dd-trace/src/ci-visibility/exporters/agentless/writer.js
juan-fernandez: no `FormData` should go through this code, just `Buffer`

packages/dd-trace/test/ci-visibility/exporters/agentless/writer.spec.js
juan-fernandez: yeah, good catch

juan-fernandez (approved):
looks good from test optimization's side 👍

sabrenner (approved):
lgtm for ml-obs changes!

PR #5125, opened 2025-01-17
DBM: Add option to SQL comment propagation

andreas:
### What does this PR do?

This PR is a port of https://github.com/DataDog/dd-trace-rb/pull/3809 to `dd-trace-js`.

It adds an option to append the DBM trace propagation comment instead, of prepending it. This allow users using database monitor tools to see the SQL query itself first, instead of the Datadog trace propagation comment first.

### Motivation

External tools, such as AWS RDS Performance Insights, are very hard to use with DBM comments prepended.

Further, it's currently impossible to use [pg_hint_plan](https://github.com/ossc-db/pg_hint_plan) when `dd-trace-js` prefixes queries with a comment.

### Plugin Checklist

- [x] Unit tests.
- [x] TypeScript [definitions][1].
- [x] TypeScript [tests][2].
- [ ] ~API [documentation][3].~

[1]: https://github.com/DataDog/dd-trace-js/blob/master/index.d.ts
[2]: https://github.com/DataDog/dd-trace-js/blob/master/docs/test.ts
[3]: https://github.com/DataDog/documentation/blob/master/content/en/tracing/trace_collection/library_config/nodejs.md
[4]: https://github.com/DataDog/dd-trace-js/blob/master/.circleci/config.yml

PR #5929, opened 2025-06-23
Add option to append comment to pg queries

BridgeAR:
This allows to opt into a different way to place pg query comments.

It ports #5125 by @andreas and I decided to rename the option and actually re-add the original added environment variable.

Closes #5125


BridgeAR:
@rochdev I know of Ruby and Java having this implemented. I did not check other platforms. I guess it would be good to have a system test for that. @bm1549 do you by chance know about that or could you ask someone to write a system test for the functionality? :)

rochdev (approved):

PR #5930, opened 2025-06-23
Activate weekly dependabot PRs for npm and switch actions to weekly

BridgeAR:
We should be up to date with dependencies in an automated way.

I decided not to ignore anything for now to see how many PRs are incoming and potentially limit it later by ignoring a couple of pinned versions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,8 +11,28 @@ updates:
       - "/.github/actions/*"
       - "/.github/actions/*/*"
     schedule:
-      interval: "monthly"
+      interval: "weekly"
     groups:
       gh-actions-packages:
         patterns:
           - "*"
+  - package-ecosystem: "npm"
+    directories:
+      - "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      dev-minor-and-patch-dependencies:
+        dependency-type: "development"
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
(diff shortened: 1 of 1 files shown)


watson (commented):
Will this also update `yarn.lock` in those cases where it's a sub-dependencies that we don't have a direct reference to in `package.json`? (I notice that you have a hardcoded reference to `/package.json` in the `dependabot.yml` file).

.github/dependabot.yml
```diff
@@ -11,8 +11,28 @@ updates:
...
+    schedule:
+      interval: "weekly"
+    groups:
+      dev-minor-and-patch-dependencies:
+        dependency-type: "development"
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
```
IlyasShabi, line 31: why ignoring major updates?

BridgeAR: This is a grouping rule. So all patch and minor updates are opened as a single PR while major versions should always open as an individual PR this way.

IlyasShabi (approved):

rochdev (changes requested):
This won't check whether it works on any legacy release lines. We often depend on older versions explicitly because of old Node version support. Because of this, I think this should be either a manual task or we should have some automation around testing older versions of Node than what is supported on the main branch.

rochdev (approved):
I guess we can use the daily proposal updates to let us know if something goes wrong on older release lines, and we can improve that to be at the PR level later on.

PR #5941, opened 2025-06-24
[DI] Add support for loading probes from JSON file

watson:
### What does this PR do?

Add support for loading Dynamic Instrumentation / Live Debugger probes via a JSON file instead of Remote Configuration (RC).

To use this feature, specify a path to the JSON file using either the environment variable `DD_DYNAMIC_INSTRUMENTATION_PROBE_FILE` or the programmatic configuration `dynamicInstrumentation.probeFile`.

This can be used in combination with RC.

The JSON should be an array of probe objects in the same format as received via the RC `config` object, for example:

```js
[
  {
    id: '100c9a5c-45ad-49dc-818b-c570d31e11d1',
    version: 0,
    type: 'LOG_PROBE',
    where: { sourceFile: 'index.js', lines: ['25'] },
    template: 'Hello World',
    segments: [{ str: 'Hello World' }],
    captureSnapshot: true,
    capture: { maxReferenceDepth: 3 },
    sampling: { snapshotsPerSecond: 100 }
  }
]
```

### Motivation

This is useful for testing and for environments that doesn't have RC enabled.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes

The reduction in test coverage is incorrect. We cover these lines in the integration tests, but these runs are not reported to Codecov.

### DIFF
--- a/.github/workflows/test-optimization.yml
+++ b/.github/workflows/test-optimization.yml
@@ -75,8 +75,9 @@ jobs:
   integration-ci:
     strategy:
       matrix:
+        # TODO: Add cucumber and selenium once cucumber+12 is fixed
         version: [oldest, latest]
-        framework: [cucumber, selenium, jest, mocha]
+        framework: [jest, mocha]
     runs-on: ubuntu-latest
     env:
       DD_SERVICE: dd-trace-js-integration-tests
@@ -160,13 +161,14 @@ jobs:
         env:
           NODE_OPTIONS: '-r ./ci/init'
 
-  plugin-cucumber:
-    runs-on: ubuntu-latest
-    env:
-      PLUGINS: cucumber
-    steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-      - uses: ./.github/actions/plugins/test
+  # TODO: Remove comment once cucumber+12 is fixed
+  # plugin-cucumber:
+  #   runs-on: ubuntu-latest
+  #   env:
+  #     PLUGINS: cucumber
+  #   steps:
+  #     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+  #     - uses: ./.github/actions/plugins/test
 
   # TODO: fix performance issues and test more Node versions
   plugin-cypress:
--- a/integration-tests/ci-visibility/automatic-log-submission.spec.js
+++ b/integration-tests/ci-visibility/automatic-log-submission.spec.js
@@ -59,10 +59,11 @@ describe('test visibility automatic log submission', () => {
       name: 'jest',
       command: 'node ./node_modules/jest/bin/jest --config ./ci-visibility/automatic-log-submission/config-jest.js'
     },
-    {
-      name: 'cucumber',
-      command: './node_modules/.bin/cucumber-js ci-visibility/automatic-log-submission-cucumber/*.feature'
-    },
+    // TODO: Uncomment once cucumber+12 is fixed
+    // {
+    //   name: 'cucumber',
+    //   command: './node_modules/.bin/cucumber-js ci-visibility/automatic-log-submission-cucumber/*.feature'
+    // },
     {
       name: 'playwright',
       command: './node_modules/.bin/playwright test -c playwright.config.js',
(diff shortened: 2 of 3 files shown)


tylfin:
FYI - the Python implementation is here to make sure we aligned https://github.com/DataDog/dd-trace-py/pull/13747

packages/dd-trace/src/debugger/index.js
```diff
@@ -23,17 +24,26 @@ function start (config, rc) {
...
+  const probeChannel = new MessageChannel()
   configChannel = new MessageChannel()
 
   process[Symbol.for('datadog:node:util:types')] = types
 
-  rc.setProductHandler('LIVE_DEBUGGING', (action, conf, id, ack) => {
+  readProbeFile(config.dynamicInstrumentation.probeFile, (probes) => {
+    const action = 'apply'
+    for (const probe of probes) {
+      probe.shouldAcknowledge = false
```
tylfin: Could you explain what `shouldAcknowledge` is doing here? Are you still emitting probe statuses when a probe is configured via a file? I ask because there are instances where we may want to validate these as part of a testing suite without remote configuration enabled.

packages/dd-trace/src/debugger/devtools_client/status.js
```diff
@@ -34,7 +34,8 @@ const STATUSES = {
   BLOCKED: 'BLOCKED' // TODO: Implement once support for allow list, deny list or max probe limit has been added
 }
 
-function ackReceived ({ id: probeId, version }) {
+function ackReceived ({ id: probeId, version, shouldAcknowledge }) {
+  if (shouldAcknowledge === false) return
```
szegedi: Explicit comparisons of a simple value with a boolean literal always catch my eye, especially when it's for a false value. I think what you want to express here is that the absence of `shouldAcknowledge` property in the function parameter object is to be treated as if it were set to true. I think the intent would be better communicated with:
```suggestion
function ackReceived ({ id: probeId, version, shouldAcknowledge }) {
  if (shouldAcknowledge ?? true) return
```
It obviously applies to all places where this is used.

szegedi (changes requested):

packages/dd-trace/src/debugger/index.js
watson: No I'm disabling probe acknowledgements that would normally be sent via the diagnostics endpoint if the probe is loaded via a file. There's nothing acknowledge as there's no matching ID in the backend

packages/dd-trace/src/debugger/devtools_client/status.js
watson: > Explicit comparisons of a simple value with a boolean literal always catch my eye, especially when it's for a false value.

I know it's not going to matter in the grand scheme of things, but I'm annoyed how in JavaScript `if (foo) {}` is going to make it a truthy check, when all you want is a boolean check.

Despite that, your suggested change means the opposite of what's intended, as it's going to abort if `shouldAcknowledge === true`, which is not what we want.

If it makes it easier to read, I could reverse the boolean, so that it becomes `skipAcknowledge`. Is that better? In that case, the diff would read:

```suggestion
function ackReceived ({ id: probeId, version, skipAcknowledge }) {
  if (skipAcknowledge === true) return
```

packages/dd-trace/src/debugger/index.js
watson: Hang on, what am I saying?!? I must say that I haven't thought about what happens when the probe result is sent to the input intake - here the ID obviously need to match if I'm not mistaken? And I possibly have been thinking too much about Exception Replay, where there's no ID and mixing it up with this one.

Am I understanding your correctly, in that the probe _does_ need to exist in the backend?

tylfin: Hmm, I didn't touch the acknowledgements in the Python implementation - we should chat about this during agent sync and see what JP did

packages/dd-trace/src/debugger/devtools_client/status.js
szegedi: You're right, it should've been
```suggestion
function ackReceived ({ id: probeId, version, shouldAcknowledge }) {
  if (!(shouldAcknowledge ?? true)) return
```
Anyhow, I don't think this is worth blocking this PR, it was just a stylistic suggestion. Going to approve as-is.

szegedi (dismissed):

BridgeAR:
@watson should this be merged? It's approved for a while.

packages/dd-trace/src/debugger/index.js
watson: I've re-added the probe diagnostics

tylfin (approved):

PR #5944, opened 2025-06-24
[test optimization] Fix playwright tests in v5

juan-fernandez:
### What does this PR do?
Fix logic for reporting attempt to fix compatibility in playwright. 

### Motivation
It's broken right now: https://github.com/DataDog/dd-trace-js/actions/runs/15841776117/job/44693661227?pr=5893

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dd-trace",
-  "version": "5.0.0-pre",
+  "version": "6.0.0-pre",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dd-trace",
-  "version": "6.0.0-pre",
+  "version": "5.0.0-pre",
```
juan-fernandez: ⚠️ to be removed before merging, just for testing

PR #5947, opened 2025-06-24
remove async storage from http2 server instrumentation

wconti27:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/http2/server.js
+++ b/packages/datadog-instrumentations/src/http2/server.js
@@ -5,8 +5,7 @@
 
 const {
   channel,
-  addHook,
-  AsyncResource
+  addHook
 } = require('../helpers/instrument')
 const shimmer = require('../../../datadog-shimmer')
 
@@ -30,14 +29,11 @@ function wrapCreateServer (createServer) {
   }
 }
 
-function wrapResponseEmit (emit) {
-  const asyncResource = new AsyncResource('bound-anonymous-fn')
+function wrapResponseEmit (emit, ctx) {
   return function (eventName, event) {
-    return asyncResource.runInAsyncScope(() => {
-      if (eventName === 'close' && finishServerCh.hasSubscribers) {
-        finishServerCh.publish({ req: this.req })
-      }
+    if (eventName !== 'close') return emit.apply(this, arguments)
 
+    return finishServerCh.runStores({ req: this.req, ...ctx }, () => {
       return emit.apply(this, arguments)
     })
   }
@@ -51,18 +47,17 @@ function wrapEmit (emit) {
     if (eventName === 'request') {
       res.req = req
 
-      const asyncResource = new AsyncResource('bound-anonymous-fn')
-      return asyncResource.runInAsyncScope(() => {
-        startServerCh.publish({ req, res })
-
-        shimmer.wrap(res, 'emit', wrapResponseEmit)
+      const ctx = { req, res }
+      return startServerCh.runStores(ctx, () => {
+        shimmer.wrap(res, 'emit', emit => wrapResponseEmit(emit, ctx))
 
         try {
           return emit.apply(this, arguments)
-        } catch (err) {
-          errorServerCh.publish(err)
+        } catch (error) {
+          ctx.error = error
+          errorServerCh.publish(ctx)
 
-          throw err
+          throw error
         }
       })
     }
--- a/packages/datadog-plugin-http2/src/server.js
+++ b/packages/datadog-plugin-http2/src/server.js
@@ -16,7 +16,13 @@ class Http2ServerPlugin extends ServerPlugin {
     this.addSub(`apm:${this.constructor.id}:server:${this.operation}:${eventName}`, handler)
   }
 
-  start ({ req, res }) {
+  addTraceBind (eventName, handler) {
+    this.addBind(`apm:${this.constructor.id}:server:${this.operation}:${eventName}`, handler)
+  }
+
+  bindStart (ctx) {
+    const { req, res } = ctx
+
     const store = storage('legacy').getStore()
     const span = web.startSpan(
       this.tracer,
@@ -39,14 +45,20 @@ class Http2ServerPlugin extends ServerPlugin {
       context.res.writeHead = web.wrapWriteHead(context)
       context.instrumented = true
     }
+
+    return ctx.currentStore
   }
 
-  finish ({ req }) {
+  bindFinish (ctx) {
+    const { req } = ctx
+
     const context = web.getContext(req)
 
     if (!context || !context.res) return // Not created by a http.Server instance.
 
     web.finishAll(context)
+
+    return ctx.parentStore
   }
 
   error (error) {
--- a/packages/dd-trace/src/plugins/util/inferred_proxy.js
+++ b/packages/dd-trace/src/plugins/util/inferred_proxy.js
@@ -20,7 +20,7 @@ const supportedProxies = {
   }
 }
 
-function createInferredProxySpan (headers, childOf, tracer, context) {
+function createInferredProxySpan (headers, childOf, tracer, reqCtx, store, traceCtx) {
   if (!headers) {
     return null
   }
@@ -57,12 +57,17 @@ function createInferredProxySpan (headers, childOf, tracer, context) {
   )
 
   tracer.scope().activate(span)
-  context.inferredProxySpan = span
+  reqCtx.inferredProxySpan = span
   childOf = span
 
   log.debug('Successfully created inferred proxy span.')
 
   setInferredProxySpanTags(span, proxyContext)
+
+  if (traceCtx) {
+    traceCtx.parentStore = store
+    traceCtx.currentStore = { ...store, span }
+  }
 
   return childOf
 }
--- a/packages/dd-trace/src/plugins/util/web.js
+++ b/packages/dd-trace/src/plugins/util/web.js
@@ -10,6 +10,7 @@ const kinds = require('../../../../../ext/kinds')
 const urlFilter = require('./urlfilter')
 const { ERROR_MESSAGE, ERROR_TYPE, ERROR_STACK } = require('../../constants')
 const { createInferredProxySpan, finishInferredProxySpan } = require('./inferred_proxy')
+const { storage } = require('../../../../datadog-core')
 
 let extractIp
 
@@ -92,7 +93,7 @@ const web = {
     analyticsSampler.sample(span, config.measured, true)
   },
 
-  startSpan (tracer, config, req, res, name) {
+  startSpan (tracer, config, req, res, name, traceCtx) {
     const context = this.patch(req)
 
     let span
@@ -101,7 +102,7 @@ const web = {
       context.span.context()._name = name
       span = context.span
     } else {
-      span = web.startChildSpan(tracer, name, req)
+      span = web.startChildSpan(tracer, name, req, traceCtx)
     }
 
     context.tracer = tracer
@@ -257,20 +258,27 @@ const web = {
   },
 
   // Extract the parent span from the headers and start a new span as its child
-  startChildSpan (tracer, name, req) {
+  startChildSpan (tracer, name, req, traceCtx) {
     const headers = req.headers
-    const context = contexts.get(req)
+    const reqCtx = contexts.get(req)
     let childOf = tracer.extract(FORMAT_HTTP_HEADERS, headers)
+
+    const store = storage('legacy').getStore()
 
     // we may have headers signaling a router proxy span should be created (such as for AWS API Gateway)
     if (tracer._config?.inferredProxyServicesEnabled) {
-      const proxySpan = createInferredProxySpan(headers, childOf, tracer, context)
+      const proxySpan = createInferredProxySpan(headers, childOf, tracer, reqCtx, store, traceCtx)
       if (proxySpan) {
         childOf = proxySpan
       }
     }
 
     const span = tracer.startSpan(name, { childOf, links: childOf?._links })
+
+    if (traceCtx) {
+      traceCtx.parentStore = store
+      traceCtx.currentStore = { ...store, span }
+    }
 
     return span
   },


packages/dd-trace/src/plugins/util/web.js
```diff
@@ -257,21 +258,28 @@ const web = {
...
-      const proxySpan = createInferredProxySpan(headers, childOf, tracer, context)
+      const proxySpan = createInferredProxySpan(headers, childOf, tracer, reqCtx, store, traceCtx)
       if (proxySpan) {
         childOf = proxySpan
       }
     }
 
     const span = tracer.startSpan(name, { childOf, links: childOf?._links })
 
+    if (traceCtx) {
```
rochdev: Why can't we rely on `startSpan` for this?

packages/datadog-plugin-http2/src/server.js
```diff
@@ -16,7 +16,13 @@ class Http2ServerPlugin extends ServerPlugin {
     this.addSub(`apm:${this.constructor.id}:server:${this.operation}:${eventName}`, handler)
   }
 
-  start ({ req, res }) {
+  addTraceBind (eventName, handler) {
+    this.addBind(`apm:${this.constructor.id}:server:${this.operation}:${eventName}`, handler)
```
rochdev: Why is this needed in addition to the existing methods?

packages/dd-trace/src/plugins/util/web.js
wconti27: The logic for ctx being set is within `tracingPlugin.startSpan`, not `tracer.startSpan`

packages/datadog-plugin-http2/src/server.js
wconti27: `bindStart` and `bindFinish` never get called without adding this method.

rochdev: Why though? That should just work if everything is named properly. Maybe there is a typo somewhere?

packages/dd-trace/src/plugins/util/web.js
rochdev: Ah yes I guess we'd need to have some WebPlugin class instead of this helper, or switch everything to composition. For now it's fine then, we can figure this out later.

packages/datadog-plugin-http2/src/server.js
wconti27: Yup, you're right, the naming was wrong, the `tracingPlugin` was using the channel prefix `apm:http2:request`, but the prefix should be `apm:http2:server:request`. I fixed it for http2 server and http server. And removed the addTraceSub method calls in the http plugins.

packages/datadog-instrumentations/src/http2/server.js
```diff
@@ -30,14 +29,11 @@ function wrapCreateServer (createServer) {
...
-  const asyncResource = new AsyncResource('bound-anonymous-fn')
+function wrapResponseEmit (emit, ctx) {
   return function (eventName, event) {
-    return asyncResource.runInAsyncScope(() => {
-      if (eventName === 'close' && finishServerCh.hasSubscribers) {
-        finishServerCh.publish({ req: this.req })
-      }
+    if (eventName !== 'close') return emit.apply(this, arguments)
 
+    return finishServerCh.runStores({ req: this.req, ...ctx }, () => {
```
rochdev: The context should be altered and passed directly instead of creating a new object.

packages/datadog-instrumentations/src/http2/server.js
```diff
@@ -30,14 +29,11 @@ function wrapCreateServer (createServer) {
...
 
-function wrapResponseEmit (emit) {
-  const asyncResource = new AsyncResource('bound-anonymous-fn')
+function wrapResponseEmit (emit, ctx) {
   return function (eventName, event) {
-    return asyncResource.runInAsyncScope(() => {
-      if (eventName === 'close' && finishServerCh.hasSubscribers) {
-        finishServerCh.publish({ req: this.req })
-      }
+    if (eventName !== 'close') return emit.apply(this, arguments)
```
rochdev: The check for `hasSubscribers` should be restored.

packages/dd-trace/src/plugins/util/web.js
```diff
@@ -92,7 +93,7 @@ const web = {
     analyticsSampler.sample(span, config.measured, true)
   },
 
-  startSpan (tracer, config, req, res, name) {
+  startSpan (tracer, config, req, res, name, traceCtx) {
```
rochdev, line 114: I can't find where `traceCtx` is passed as a parameter in this PR.

packages/datadog-instrumentations/src/http2/server.js
rochdev: Also, if I understand the new code correctly, it now restores the parent context here where before it was restoring the current context. This means event listeners will no longer be in the context of the span. I'm not sure which is the expected behaviour, especially if there was no test for this, but to be safe it's probably best to keep the previous behaviour.

packages/dd-trace/src/plugins/util/inferred_proxy.js
```diff
@@ -57,13 +57,18 @@ function createInferredProxySpan (headers, childOf, tracer, context) {
...
   tracer.scope().activate(span)
-  context.inferredProxySpan = span
+  reqCtx.inferredProxySpan = span
   childOf = span
 
   log.debug('Successfully created inferred proxy span.')
 
   setInferredProxySpanTags(span, proxyContext)
 
+  if (traceCtx) {
```
rochdev: This is repeated in a few places at this point. Could we maybe create a `WebPlugin` that would reuse the existing code instead in addition to call the util?

rochdev (changes requested):

packages/datadog-instrumentations/src/http2/server.js
wconti27: fixed.

packages/dd-trace/src/plugins/util/web.js
wconti27: fixed this.

packages/datadog-instrumentations/src/http2/server.js
wconti27: I fixed this behavior, now it restores the current context.

packages/dd-trace/src/plugins/util/inferred_proxy.js
wconti27: made a helper method that calls `TracingPlugin.startSpan`, which allowed me to delete this repeated code.

BridgeAR:
@wconti27 I believe the system test failures are related to the current implementation. If that's the case, I guess we are missing local tests, since it would be good to be aware about the behavior in our direct tests as well. As such, it would be great if additional tests are added as part of this PR when the failures are fixed.

wconti27:
> @wconti27 I believe the system test failures are related to the current implementation. If that's the case, I guess we are missing local tests, since it would be good to be aware about the behavior in our direct tests as well. As such, it would be great if additional tests are added as part of this PR when the failures are fixed.

Its odd, we do have tests. But they said they passed even though I had originally messed up the start span call and it was failing to run and create inferred proxy spans. Looking into it now

packages/datadog-instrumentations/src/http2/server.js
```diff
@@ -30,14 +29,11 @@ function wrapCreateServer (createServer) {
...
+function wrapResponseEmit (emit, ctx) {
   return function (eventName, event) {
-    return asyncResource.runInAsyncScope(() => {
-      if (eventName === 'close' && finishServerCh.hasSubscribers) {
-        finishServerCh.publish({ req: this.req })
-      }
-
+    ctx.req = this.req
+    ctx.eventName = eventName
+    return finishServerCh.runStores(ctx, () => {
```
rochdev: Looking at the previous code it looks like the binding of emit was the same as when the span is started and not the parent. This means the new behaviour is different (which might also explain the system test failures)

wconti27:
> > @wconti27 I believe the system test failures are related to the current implementation. If that's the case, I guess we are missing local tests, since it would be good to be aware about the behavior in our direct tests as well. As such, it would be great if additional tests are added as part of this PR when the failures are fixed.
> 
> Its odd, we do have tests. But they said they passed even though I had originally messed up the start span call and it was failing to run and create inferred proxy spans. Looking into it now
> 
> They were run here: https://github.com/DataDog/dd-trace-js/actions/runs/16197859525/job/45728961958?pr=5947#:~:text=packages/dd%2Dtrace,to%20be%20off

@BridgeAR Fixed this, we previously were catching the thrown error in the callback of `agent.assertSomeTraces`, which meant that when they failed, we never actually threw a test error. I've fixed this behavior, and now the tests should be correct.

packages/datadog-instrumentations/src/http2/server.js
```diff
@@ -30,14 +30,12 @@ function wrapCreateServer (createServer) {
...
-function wrapResponseEmit (emit) {
-  const asyncResource = new AsyncResource('bound-anonymous-fn')
+function wrapResponseEmit (emit, ctx, store) {
   return function (eventName, event) {
-    return asyncResource.runInAsyncScope(() => {
-      if (eventName === 'close' && finishServerCh.hasSubscribers) {
-        finishServerCh.publish({ req: this.req })
-      }
-
+    storage('legacy').enterWith(store)
```
rochdev: The storage shouldn't be accessed from instrumentation. A new channel should be used instead which handles the storage on the subscriber side.

rochdev (changes requested):

packages/datadog-instrumentations/src/http2/server.js
wconti27: Okay, I've removed the direct use of storage and added a new channel instead.

packages/datadog-instrumentations/src/http2/server.js
```diff
@@ -5,14 +5,15 @@
...
-  addHook,
-  AsyncResource
+  addHook
 } = require('../helpers/instrument')
 const shimmer = require('../../../datadog-shimmer')
 
 const startServerCh = channel('apm:http2:server:request:start')
 const errorServerCh = channel('apm:http2:server:request:error')
 const finishServerCh = channel('apm:http2:server:request:finish')
+// this channel is for wrapping the response emit method and handling store context, it doesn't have any subscribers
```
rochdev: This needs a subscriber to have any effect.

packages/datadog-instrumentations/src/http2/server.js
```diff
@@ -51,18 +47,19 @@ function wrapEmit (emit) {
...
       res.req = req
 
-      const asyncResource = new AsyncResource('bound-anonymous-fn')
-      return asyncResource.runInAsyncScope(() => {
-        startServerCh.publish({ req, res })
-
-        shimmer.wrap(res, 'emit', wrapResponseEmit)
+      const ctx = { req, res }
+      return startServerCh.runStores(ctx, () => {
+        shimmer.wrap(res, 'emit', emit => responseCh.runStores(ctx, () => {
```
rochdev: This should happen at emit time and not at wrap time. From my understanding of the new code it should look something like this:

```js
function wrapResponseEmit (emit, ctx) {
  return function (eventName, event) {
    ctx.req = this.req
    ctx.eventName = eventName

    return eventName === 'close' && finishServerCh.hasSubscribers
      ? finishServerCh.runStores(ctx, emit, this, ...arguments)
      : responseCh.runStores(ctx, emit, this, ...arguments)
  }
}
```

packages/datadog-instrumentations/src/http2/server.js
wconti27: done.

wconti27: interestingly enough, adding the subscriber caused tests to fail and doesn't work as anticipated, but without the subscriber, the tests pass and context gets past around correctly (and the wrapped code for the channel is correctly called despite no subscriber).

wconti27: Fixed in our review session

packages/datadog-instrumentations/src/http2/server.js
wconti27: Fixed in our review session.

packages/dd-trace/src/plugins/tracing.js
```diff
@@ -102,25 +102,28 @@ class TracingPlugin extends Plugin {
     }
   }
 
-  startSpan (name, { childOf, kind, meta, metrics, service, resource, type } = {}, enterOrCtx = true) {
+  startSpan (name, options = {}, enterOrCtx = true) {
```
rochdev, line 105:
```suggestion
  startSpan (name, options = {}, enterOrCtx = true) {
  let { childOf, kind, meta, metrics, service, resource, type } = options
```

It's best to not alter object received as a parameter to a function. This also means restoring direct usage below.

packages/dd-trace/src/plugins/util/web.js
```diff
@@ -37,6 +37,17 @@ const HTTP2_HEADER_PATH = ':path'
 const contexts = new WeakMap()
 const ends = new WeakMap()
 
+const TracingPlugin = require('../tracing')
+
+function startSpanHelper (tracer, name, options, traceCtx, config = {}) {
+  return TracingPlugin.prototype.startSpan.call(
```
rochdev: This seems like an anti-pattern, prototype methods are meant to become available on instances of the class, not be accessed directly.

rochdev (changes requested):
The change overall LGTM, just a couple of small things to address and it's good to go!

packages/dd-trace/src/plugins/tracing.js
wconti27: fixed.

packages/dd-trace/src/plugins/util/web.js
wconti27: Changed the implementation to create a new instance of tracing plugin for web to use.

rochdev: What I meant is that if what is needed is a function and only a function without the object containing it as a method, then it should be extracted elsewhere so that the class/object is not needed at all.

rochdev (dismissed):

tlhunter (dismissed):

tlhunter:
You might want to ignore my merge conflict fix, squash all your commits locally, pull master, then rebase.

rochdev (approved):

PR #5948, opened 2025-06-24
remove async storage from couchbase

wconti27:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/couchbase.js
+++ b/packages/datadog-instrumentations/src/couchbase.js
@@ -191,7 +191,9 @@ function wrapV3Query (query) {
 
 // semver >=2 <3
 addHook({ name: 'couchbase', file: 'lib/bucket.js', versions: ['^2.6.12'] }, Bucket => {
-  shimmer.wrap(Bucket.prototype, '_maybeInvoke', maybeInvoke => wrapMaybeInvoke(maybeInvoke, 'apm:couchbase:bucket:maybeInvoke'))
+  shimmer.wrap(Bucket.prototype, '_maybeInvoke', maybeInvoke => {
+    return wrapMaybeInvoke(maybeInvoke, 'apm:couchbase:bucket:maybeInvoke')
+  })
 
   const startCh = channel('apm:couchbase:query:start')
   const finishCh = channel('apm:couchbase:query:finish')
@@ -241,7 +243,9 @@ addHook({ name: 'couchbase', file: 'lib/bucket.js', versions: ['^2.6.12'] }, Buc
 })
 
 addHook({ name: 'couchbase', file: 'lib/cluster.js', versions: ['^2.6.12'] }, Cluster => {
-  shimmer.wrap(Cluster.prototype, '_maybeInvoke', maybeInvoke => wrapMaybeInvoke(maybeInvoke, 'apm:couchbase:cluster:maybeInvoke'))
+  shimmer.wrap(Cluster.prototype, '_maybeInvoke', maybeInvoke => {
+    return wrapMaybeInvoke(maybeInvoke, 'apm:couchbase:cluster:maybeInvoke')
+  })
 
   shimmer.wrap(Cluster.prototype, 'query', query => wrapQuery(query))
   shimmer.wrap(Cluster.prototype, 'openBucket', openBucket => {


packages/datadog-plugin-couchbase/src/index.js
```diff
@@ -63,10 +72,11 @@ class CouchBasePlugin extends StoragePlugin {
...
   _addCommandSubs (name) {
-    this.addSubs(name, ({ bucket, collection, seedNodes }) => {
-      const store = storage('legacy').getStore()
-      const span = this.startSpan(name, {}, store, { bucket, collection, seedNodes })
-      this.enter(span, store)
+    this.addBinds(name, (ctx) => {
+      const { bucket, collection, seedNodes } = ctx
+
+      const span = this.startSpan(name, {}, { bucket, collection, seedNodes }, ctx)
+      this.enter(span, ctx.currentStore)
```
rochdev: What ensures that this is bound to a specific scope without AsyncResource? Also, why use `addBind` if the context is not returned?

packages/datadog-plugin-couchbase/src/index.js
```diff
@@ -7,13 +7,18 @@ class CouchBasePlugin extends StoragePlugin {
...
 
-  addSubs (func, start) {
-    this.addSub(`apm:couchbase:${func}:start`, start)
-    this.addSub(`apm:couchbase:${func}:error`, error => this.addError(error))
-    this.addSub(`apm:couchbase:${func}:finish`, message => this.finish(message))
+  addBinds (func, start) {
+    this.addBind(`apm:couchbase:${func}:start`, start)
+    this.addSub(`apm:couchbase:${func}:error`, ({ error }) => this.addError(error))
+    this.addSub(`apm:couchbase:${func}:finish`, ctx => this.finish(ctx))
+    this.addBind('apm:couchbase:query:callback:start', ctx => {
```
rochdev: This will re-add the same handler multiple times that will apply to every publisher.

packages/datadog-instrumentations/src/couchbase.js
```diff
@@ -176,24 +212,31 @@ addHook({ name: 'couchbase', file: 'lib/bucket.js', versions: ['^2.6.12'] }, Buc
...
-        finishCh.publish()
-      }))
-
-      emitter.once(errorMonitor, asyncResource.bind((error) => {
-        errorCh.publish(error)
-        finishCh.publish()
-      }))
+    const ctx = { resource: n1qlQuery, bucket: { name: this.name || this._name }, seedNodes: this._dd_hosts }
+    return startCh.runStores(ctx, () => {
+      callbackStartCh.runStores(ctx, () => {
```
rochdev: This seems either redundant or in the wrong place? Not sure why 2 nested `runStores` would be needed with the same context.

packages/datadog-instrumentations/src/couchbase.js
```diff
@@ -104,36 +133,42 @@ function wrapCBandPromise (fn, name, startData, thisArg, args) {
...
+          return wrapCallbackFinish(cb, thisArg, args, errorCh, finishCh, ctx)
+        })
       }
       const res = fn.apply(thisArg, args)
 
       // semver >=3 will always return promise by default
-      res.then(
-        asyncResource.bind((result) => finishCh.publish({ result })),
-        asyncResource.bind((err) => errorCh.publish(err)))
+      callbackStartCh.runStores(ctx, () => {
```
rochdev: Promises already bind properly out of the box, so probably none of the `runStores` calls are needed in this block.

packages/datadog-instrumentations/src/couchbase.js
```diff
@@ -3,11 +3,14 @@
...
 const {
   channel,
-  addHook,
-  AsyncResource
+  addHook
 } = require('./helpers/instrument')
 const shimmer = require('../../datadog-shimmer')
 
+// these channels are for maintaining store context for their respective methods
+const callbackStartCh = channel('apm:couchbase:query:callback:start')
```
rochdev: Why not use a TracingChannel for each channel in the file at that point instead?

packages/datadog-plugin-couchbase/src/index.js
wconti27: Fixed.

packages/datadog-instrumentations/src/couchbase.js
wconti27: Good catch, its not needed.

packages/datadog-instrumentations/src/couchbase.js
wconti27: Makes sense, I removed the usage of runstores in promises.

packages/datadog-instrumentations/src/couchbase.js
wconti27: I just tried to switch over tracing channels and was struggling, we can always do that in a followup PR.

packages/datadog-plugin-couchbase/src/index.js
wconti27: I removed the `this.enter` and returned the `currentStore`. I believe this should fix that binding issue.

packages/datadog-instrumentations/src/couchbase.js
rochdev: This will need to be split in multiple channels anyway to address every subscriber receiving the message from all the publishers anyway, so at that point it might actually be simpler to do that at the same time.

wconti27: Fixed, though I did not make the tracingChannel change. I must be missing something, but I couldn't get stuff working properly (mainly subscriptions).

rochdev (approved):

PR #5956, opened 2025-06-25
remove async storage from moleculer

wconti27:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/moleculer/client.js
+++ b/packages/datadog-instrumentations/src/moleculer/client.js
@@ -24,12 +24,14 @@ function wrapCall (call) {
       return promise
         .then(
           result => {
-            finishChannel.runStores(ctx, () => { return result })
+            finishChannel.publish(ctx)
+            return result
           },
           error => {
             ctx.error = error
             errorChannel.publish(ctx)
-            finishChannel.runStores(ctx, () => { throw error })
+            finishChannel.publish(ctx)
+            throw error
           }
         )
     })
--- a/packages/datadog-plugin-moleculer/src/client.js
+++ b/packages/datadog-plugin-moleculer/src/client.js
@@ -21,7 +21,7 @@ class MoleculerClientPlugin extends ClientPlugin {
     return ctx.currentStore
   }
 
-  bindFinish (ctx) {
+  finish (ctx) {
     const { promiseCtx, broker } = ctx
 
     const span = ctx.currentStore.span || this.activeSpan
@@ -36,8 +36,6 @@ class MoleculerClientPlugin extends ClientPlugin {
     }
 
     super.finish(ctx)
-
-    return ctx.parentStore
   }
 }
 
--- a/packages/datadog-instrumentations/src/moleculer/server.js
+++ b/packages/datadog-instrumentations/src/moleculer/server.js
@@ -30,18 +30,20 @@ function createMiddleware () {
           try {
             return next(middlewareCtx).then(
               result => {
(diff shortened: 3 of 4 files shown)


packages/datadog-instrumentations/src/moleculer/client.js
```diff
@@ -9,30 +9,27 @@ const errorChannel = channel('apm:moleculer:call:error')
...
-      const ctx = promise.ctx
+      ctx.promiseCtx = promise.ctx
+      ctx.broker = broker
 
       return promise
         .then(
           result => {
-            finishChannel.publish({ broker, ctx })
-            return result
+            finishChannel.runStores(ctx, () => { return result })
```
rochdev: `runStores` is not needed here since there is no callback.

packages/datadog-instrumentations/src/moleculer/client.js
```diff
@@ -9,30 +9,27 @@ const errorChannel = channel('apm:moleculer:call:error')
...
           },
           error => {
-            errorChannel.publish(error)
-            finishChannel.publish({ broker, ctx })
-            throw error
+            ctx.error = error
+            errorChannel.publish(ctx)
+            finishChannel.runStores(ctx, () => { throw error })
```
rochdev: Same as the above.

packages/datadog-plugin-moleculer/src/client.js
```diff
@@ -7,29 +7,37 @@ class MoleculerClientPlugin extends ClientPlugin {
...
+    }, ctx)
 
     this.tracer.inject(span, 'text_map', opts.meta)
+
+    return ctx.currentStore
   }
 
-  finish ({ broker, ctx }) {
-    const span = this.activeSpan
+  bindFinish (ctx) {
```
rochdev: `bindFinish` is not needed since `runStores` is also not needed as described above, because promises already handle binding for us.

rochdev (changes requested):

packages/datadog-instrumentations/src/moleculer/client.js
wconti27: made this change 👍

packages/datadog-instrumentations/src/moleculer/client.js
wconti27: made this change 👍

packages/datadog-plugin-moleculer/src/client.js
wconti27: made this change 👍

packages/datadog-instrumentations/src/moleculer/client.js
```diff
@@ -9,29 +9,28 @@ const errorChannel = channel('apm:moleculer:call:error')
...
 
-    return callResource.runInAsyncScope(() => {
-      startChannel.publish({ actionName, params, opts })
-
+    const ctx = { actionName, params, opts }
+    return startChannel.runStores(ctx, () => {
       const promise = call.apply(this, arguments)
       const broker = this
-      const ctx = promise.ctx
+      ctx.promiseCtx = promise.ctx
```
rochdev, line 21: It would be better to pass what we need directly here, but that's a nit and could be cleaned up later if needed so not blocking on this.

rochdev (approved):

PR #5959, opened 2025-06-25
fix mongoose flaky tests

wconti27:
### What does this PR do?
Fixes mongoose flakes due to race condition setting peer service tag and the span finishing.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/test/setup/mocha.js
+++ b/packages/dd-trace/test/setup/mocha.js
@@ -129,16 +129,32 @@ function withPeerService (tracer, pluginName, spanGenerationFn, service, service
     })
 
     it('should compute peer service', done => {
-      agent
-        .assertSomeTraces(traces => {
-          const span = traces[0][0]
-          expect(span.meta).to.have.property('peer.service', typeof service === 'function' ? service() : service)
-          expect(span.meta).to.have.property('_dd.peer.service.source', serviceSource)
-        })
-        .then(done)
-        .catch(done)
+      const assertPeerService = traces => {
+        const span = traces[0][0]
+        expect(span.meta).to.have.property('peer.service', typeof service === 'function' ? service() : service)
+        expect(span.meta).to.have.property('_dd.peer.service.source', serviceSource)
+      }
 
-      spanGenerationFn(done)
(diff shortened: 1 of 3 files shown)


BridgeAR (changes requested):
This is sync code, so it should effectively not change anything.

tlhunter (changes requested):
The changes made in this file shouldn't have any affect on the order that code runs in since the `tagPeerService` method doesn't involve any asynchrony. In both the before and after case `super.finish` is always guaranteed to call after all the work that's done inside of `tagPeerService`.

packages/dd-trace/test/setup/mocha.js
```diff
@@ -129,16 +129,32 @@ function withPeerService (tracer, pluginName, spanGenerationFn, service, service
...
-          expect(span.meta).to.have.property('_dd.peer.service.source', serviceSource)
+      const assertPeerService = traces => {
+        const span = traces[0][0]
+        expect(span.meta).to.have.property('peer.service', typeof service === 'function' ? service() : service)
+        expect(span.meta).to.have.property('_dd.peer.service.source', serviceSource)
+      }
+
+      // we were hitting race condition where the test was finishing (with error) before the span
+      // generation was complete, so add a wait for the span to be finished before asserting the peer service
+      if (opts.waitForFinish) {
```
rochdev, line 140: Any reason not to always wait for finish at that point?

wconti27: it requires touching pretty much every integration test file

BridgeAR: With the current change (if I am reading the code correct), it would improve the test case by not calling `done(error)` in the `spanGeneration` and then in the assertions again (that call could theoretically also pass).

If that is the case, it would end up with an improved error message (no double done calls anymore), it would just still error, if I am not mistaken.

PR #5963, opened 2025-06-26
feat(openai): support OpenAI v5

sabrenner:
### What does this PR do?
Supports OpenAI v5 buy adding proper patching support for `client.files.delete` and `client.models.delete`, previously both `.del` instead.

Additionally, there are some Node.js 20 stipulations for files that are added into the tests as well.

### Motivation
Support most recent OpenAI major, remove version cap on CI.

### DIFF
--- a/packages/datadog-plugin-openai/src/tracing.js
+++ b/packages/datadog-plugin-openai/src/tracing.js
@@ -366,6 +366,7 @@ function normalizeMethodName (methodName) {
     case 'files.retrieve':
       return 'retrieveFile'
     case 'files.del':
+    case 'files.delete':
       return 'deleteFile'
     case 'files.retrieveContent':
     case 'files.content':
@@ -410,6 +411,7 @@ function normalizeMethodName (methodName) {
     case 'models.retrieve':
       return 'retrieveModel'
     case 'models.del':
+    case 'models.delete':
       return 'deleteModel'
     default:
       return methodName
@@ -952,6 +954,7 @@ function normalizeRequestPayload (methodName, args) {
 
     case 'deleteFile':
     case 'files.del':
+    case 'files.delete':
     case 'retrieveFile':
     case 'files.retrieve':
     case 'downloadFile':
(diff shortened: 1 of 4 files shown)


packages/datadog-plugin-openai/test/index.spec.js
```diff
@@ -1785,7 +1807,7 @@ describe('Plugin', () => {
       })
 
       it('makes a successful call with the beta chat completions', async function () {
-        if (semver.satisfies(realVersion, '<4.59.0')) {
+        if (semver.satisfies(realVersion, '<4.59.0 || >=5.0.0')) {
```
BridgeAR: I believe that feature is now just stable and we should just directly access it instead of skipping the test here?

See https://github.com/openai/openai-node/commit/4c622f9f55529e3aab30c834349d341038499db1

sabrenner: ah good find let me try that out. i was inclined to skip it anyhow as it was experimental support we were going to drop anyways, and then officially support later, but if the test still passes with changing then a win's a win.

if for some reason it doesn't work, i will leave a TODO with a ticket to separately officially support it.

sabrenner: it just works 🎉 thanks again for pointing this out!

BridgeAR (approved):
LGTM, while I would adjust the test title now, that the chat completions is not a beta feature anymore :)

packages/datadog-plugin-openai/src/tracing.js
```diff
@@ -410,6 +411,7 @@ function normalizeMethodName (methodName) {
     case 'models.retrieve':
       return 'retrieveModel'
     case 'models.del':
+    case 'models.delete':
```
BridgeAR, line 414: Not something for this PR, while I do believe we should use a map for the lookup here. That is very likely much faster than going through all these entries as we currently do and we seemingly have to do that every time.
I would check a map and an object and compare that performance.

PR #5966, opened 2025-06-26
remove async storage from graphql

wconti27:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/graphql.js
+++ b/packages/datadog-instrumentations/src/graphql.js
@@ -2,8 +2,7 @@
 
 const {
   addHook,
-  channel,
-  AsyncResource
+  channel
 } = require('./helpers/instrument')
 const shimmer = require('../../datadog-shimmer')
 
@@ -89,10 +88,8 @@ function wrapParse (parse) {
       return parse.apply(this, arguments)
     }
 
-    const asyncResource = new AsyncResource('bound-anonymous-fn')
-
-    return asyncResource.runInAsyncScope(() => {
-      parseStartCh.publish()
+    const ctx = {}
+    return parseStartCh.runStores(ctx, () => {
       let document
       try {
         document = parse.apply(this, arguments)
@@ -107,11 +104,12 @@ function wrapParse (parse) {
         return document
       } catch (err) {
         err.stack
-        parseErrorCh.publish(err)
+        ctx.error = err
+        parseErrorCh.publish(ctx)
 
         throw err
       } finally {
-        parseFinishCh.publish({ source, document, docSource: documentSources.get(document) })
+        parseFinishCh.publish({ source, document, docSource: documentSources.get(document), ...ctx })
       }
     })
   }
@@ -123,25 +121,24 @@ function wrapValidate (validate) {
       return validate.apply(this, arguments)
     }
 
-    const asyncResource = new AsyncResource('bound-anonymous-fn')
-
-    return asyncResource.runInAsyncScope(() => {
-      validateStartCh.publish({ docSource: documentSources.get(document), document })
-
+    const ctx = { docSource: documentSources.get(document), document }
+    return validateStartCh.runStores(ctx, () => {
       let errors
       try {
         errors = validate.apply(this, arguments)
         if (errors && errors[0]) {
-          validateErrorCh.publish(errors && errors[0])
+          ctx.error = errors && errors[0]
+          validateErrorCh.publish(ctx)
         }
         return errors
       } catch (err) {
         err.stack
-        validateErrorCh.publish(err)
+        ctx.error = err
+        validateErrorCh.publish(ctx)
 
         throw err
       } finally {
-        validateFinishCh.publish({ document, errors })
+        validateFinishCh.publish({ errors, ...ctx })
       }
     })
   }
@@ -155,15 +152,15 @@ function wrapExecute (execute) {
         return exe.apply(this, arguments)
       }
 
-      const asyncResource = new AsyncResource('bound-anonymous-fn')
-      return asyncResource.runInAsyncScope(() => {
-        const args = normalizeArgs(arguments, defaultFieldResolver)
-        const schema = args.schema
-        const document = args.document
-        const source = documentSources.get(document)
-        const contextValue = args.contextValue
-        const operation = getOperation(document, args.operationName)
+      const args = normalizeArgs(arguments, defaultFieldResolver)
+      const schema = args.schema
+      const document = args.document
+      const source = documentSources.get(document)
+      const contextValue = args.contextValue
+      const operation = getOperation(document, args.operationName)
 
+      const ctx = { operation, args, docSource: documentSources.get(document) }
+      return startExecuteCh.runStores(ctx, () => {
         if (contexts.has(contextValue)) {
           return exe.apply(this, arguments)
         }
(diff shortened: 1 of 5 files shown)


packages/datadog-instrumentations/src/graphql.js
```diff
@@ -123,25 +121,24 @@ function wrapValidate (validate) {
...
       } catch (err) {
         err.stack
-        validateErrorCh.publish(err)
+        ctx.error = err
+        validateErrorCh.publish(ctx)
 
         throw err
       } finally {
-        validateFinishCh.publish({ document, errors })
+        validateFinishCh.publish({ errors, ...ctx })
```
rochdev: The context should be passed by reference in case the subscriber wants to attach something to it. It also means no additional object allocation is needed.

packages/datadog-instrumentations/src/graphql.js
```diff
@@ -211,8 +204,8 @@ function wrapResolve (resolve) {
 
     const field = assertField(context, info, args)
 
-    return callInAsyncScope(resolve, field.asyncResource, this, arguments, context.abortController, (err) => {
-      updateFieldCh.publish({ field, info, err })
+    return callInAsyncScope(resolve, this, arguments, context.abortController, (err) => {
+      updateFieldCh.publish({ field, info, err, ...field.ctx })
```
rochdev: Similar to the above, this doesn't persist the data on the context nor keep the reference.

packages/datadog-instrumentations/src/graphql.js
```diff
@@ -271,27 +262,15 @@ function assertField (context, info, args) {
...
-        })
-      })
-
-      field = fields[pathString] = {
-        parent,
-        asyncResource: childResource,
-        error: null
-      }
-    })
+    // we need to pass the parent span to the field if it exists for correct span parenting
```
rochdev: This should be handled on the plugin side. Instrumentation should never know about spans.

rochdev: Basically, the plugin should be tracking the hierarchy of resolvers through the context and any information on the context so that it can do the correct parentage. It should really always have been the plugin handling this, but with `AsyncResource` that was not really possible.

rochdev (changes requested):

packages/datadog-instrumentations/src/graphql.js
wconti27: Makes sense, fixing this.

packages/datadog-instrumentations/src/graphql.js
wconti27: Fixed

packages/datadog-instrumentations/src/graphql.js
```diff
@@ -259,59 +259,26 @@ function pathToArray (path) {
...
-}
-
-function getParentField (context, path) {
-  for (let i = path.length - 1; i > 0; i--) {
-    const field = getField(context, path.slice(0, i))
-    if (field) {
-      return field
+    const fieldCtx = { info, ctx: parentCtx, args }
+    startResolveCh.publish(fieldCtx)
+    field = fields[pathString] = {
```
uurien, line 276: you are replacing `context` property by `ctx` here, is the reason why appsec tests are failing.

If you want to maintain that, you have to modify the references of `context` in appsec subscriptions, like here:
https://github.com/DataDog/dd-trace-js/blob/c17708e05dec0fda4992c54065fc267abfbd52e7/packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js#L124

wconti27: Ahh i didn't catch that, thank you for this, just changed it hopefully all the appsec failures dissappear.

packages/datadog-instrumentations/src/graphql.js
```diff
@@ -155,15 +153,22 @@ function wrapExecute (execute) {
...
+
+      const ctx = {
+        operation,
+        args,
+        docSource: documentSources.get(document),
+        source,
+        fields: {},
+        abortController: new AbortController()
+      }
+      return startExecuteCh.runStores(ctx, () => {
```
rochdev: The previous code would skip this channel if `contexts` container `contextValue` to avoid double patching. This doesn't seem to be handled with the new code (it probably needs to be checked somehow in the plugin now since `startExecuteCh` handles both starting the span and context management which were separated before).

packages/datadog-instrumentations/src/graphql.js
```diff
@@ -259,59 +259,26 @@ function pathToArray (path) {
...
-
-  return field
-}
-
-function getParentField (context, path) {
-  for (let i = path.length - 1; i > 0; i--) {
-    const field = getField(context, path.slice(0, i))
-    if (field) {
-      return field
+    const fieldCtx = { info, ctx: parentCtx, args }
```
rochdev: Renaming this `ctx` will mean there is a `field.ctx.ctx`. I'd probably keep the `parentCtx` name instead.

packages/datadog-plugin-graphql/src/resolve.js
```diff
@@ -9,40 +9,50 @@ class GraphQLResolvePlugin extends TracingPlugin {
   static get id () { return 'graphql' }
   static get operation () { return 'resolve' }
 
-  start ({ info, context, args }) {
+  start (fieldCtx) {
+    const { info, ctx: parentCtx, args } = fieldCtx
+
     const path = getPath(info, this.config)
 
+    // we need to get the parent span to the field if it exists for correct span parenting
```
rochdev, line 17: Isn't `parentCtx` already the context of the parent field? Why is a search needed if that's the case?

rochdev (changes requested):
A few nits/questions but the code LGTM, it actually simplifies things a lot which is great!

I think now that `AsyncResource` is gone it will even be possible to refactor this integration to be even simpler in the future 👍

packages/datadog-instrumentations/src/graphql.js
wconti27: Is moving it before the start ch fine? It's an easy change.

packages/datadog-instrumentations/src/graphql.js
wconti27: done.

packages/datadog-plugin-graphql/src/resolve.js
wconti27: Good catch, this `parentCtx` could be better described as the root context of the graphql operation, and corresponds to the root `graphql.execute` span. I have updated the naming to be `rootCtx` for this code. From the rootCtx, we still need the search to get the correct parent field. Hope that clears things up.

packages/datadog-instrumentations/src/graphql.js
wconti27: switched to `rootCtx`

packages/datadog-plugin-graphql/src/resolve.js
rochdev: Ah yeah it's way clearer now, I got confused the it was the direct parent without of the previous name.

rochdev (approved):

PR #5967, opened 2025-06-26
Enhance Service Representation for Serverless

zarirhamza:
### What does this PR do?

Rollout of span naming changes to align tracer with serverless product to create streamlined Service Representation for Serverless

Key Changes:
- Apply all changes ONLY in a Serverless scenario
- Subscribe to `http/http2/tcp/dns` spans called by the aws-sdk client 
- Subscribe to start events for cold starts
- Apply explicit `peer.service` tag equal to the hostname based on aws service type and region grouping all spans underneath one explicitly defined inferred service
- Extract hostname within the region event for accuracy
- Enhance storage with region and params in a Serverless environment

### Motivation

Improve Service Map for Serverless. This allows for AWS client spans to have the same service name as the underlying http/tcp/dns calls within Serverless environments thus cleaning up the Service Map.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -69,8 +116,11 @@ class BaseAwsSdkPlugin extends ClientPlugin {
       }
 
       const store = storage('legacy').getStore()
-
-      this.enter(span, store)
+      if (this._tracerConfig?._isInServerlessEnvironment()) {
+        this.enter(span, { ...store, span, awsParams: request.params, awsService })
+      } else {
+        this.enter(span)
```
watson: Are you sure you want to not pass along `store` in the normal case?

```suggestion
        this.enter(span, store)
```

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -80,6 +130,28 @@ class BaseAwsSdkPlugin extends ClientPlugin {
...
+            case 'Kinesis': return `kinesis.${region}.amazonaws.com`
+            case 'DynamoDBDocument': // or 'DynamoDB'
+              return `dynamodb.${region}.amazonaws.com`
+            case 'S3':
+              return awsParams?.Bucket
+                ? `${awsParams.Bucket}.s3.${region}.amazonaws.com`
+                : `s3.${region}.amazonaws.com`
+            default: return null
+          }
+        })()
```
watson: I think it's a bit verbose and potentially a performance issue to create an in-line lambda just for allowing you to use `return` in a switch statement instead of either just set `hostname` in each `case` block, or move the entire switch block to a helper function.

Option 1:

```suggestion
        const { awsParams, awsService } = store
        let hostname
        switch (awsService) {
          case 'EventBridge':
            hostname = `events.${region}.amazonaws.com`
            break
          case 'SQS':
            hostname = `sqs.${region}.amazonaws.com`
            break
          case 'SNS':
            hostname = `sns.${region}.amazonaws.com`
            break
          case 'Kinesis':
            hostname = `kinesis.${region}.amazonaws.com`
            break
          case 'DynamoDBDocument': // or 'DynamoDB'
            hostname = `dynamodb.${region}.amazonaws.com`
            break
          case 'S3':
            hostname = awsParams?.Bucket
              ? `${awsParams.Bucket}.s3.${region}.amazonaws.com`
              : `s3.${region}.amazonaws.com`
        }
```

Option 2:

```suggestion
        const hostname = getHostname(store, region)
```

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -34,6 +34,53 @@ class BaseAwsSdkPlugin extends ClientPlugin {
...
     super(...args)
 
+    if (this._tracerConfig?._isInServerlessEnvironment()) {
+      // subscribe to http channels for cold start
+      const httpChannelsStart = [
+        'apm:http:client:request:start',
+        'apm:http2:client:request:start',
+        'apm:net:tcp:start',
+        'apm:dns:lookup:start'
+      ]
```
watson: I would move this to the top of the tile, so it's only define once and not on every invocation

packages/datadog-plugin-aws-sdk/src/base.js
zarirhamza: Fixed

packages/datadog-plugin-aws-sdk/src/base.js
zarirhamza: Implemented a helper function

packages/datadog-plugin-aws-sdk/src/base.js
zarirhamza: Fixed

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -34,6 +42,46 @@ class BaseAwsSdkPlugin extends ClientPlugin {
...
+        const span = ctx?.currentStore?.span
+        if (!span) return
+        const { peerHostname } = storage('legacy').getStore() || {}
+        if (span && peerHostname) {
+          span.setTag('peer.service', peerHostname)
+        }
+      })
+
+      this.addSub('apm:dns:lookup:finish', (ctx) => {
+        const span = ctx?.currentStore?.span
```
wconti27: All these handlers use very similar code, can you abstract the duplication and make 1 function?

wconti27 (changes requested):
The feature also needs tests, they can go in the current aws-sdk test suite.

packages/datadog-plugin-aws-sdk/src/base.js
zarirhamza: Fixed

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -34,6 +42,46 @@ class BaseAwsSdkPlugin extends ClientPlugin {
   constructor (...args) {
     super(...args)
 
+    if (this._tracerConfig?._isInServerlessEnvironment()) {
+      // subscribe to http channels for cold start
+      for (const ch of httpChannelsStart) {
+        this.addSub(ch, () => {})
+      }
```
wconti27: This can be removed if there's no functionality needed to run on your start channels (the callback is a no-op)

zarirhamza: Fixed

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -243,4 +297,28 @@ function getHooks (config) {
...
+      return `kinesis.${region}.amazonaws.com`
+    case 'DynamoDBDocument':
+    case 'DynamoDB':
+      return `dynamodb.${region}.amazonaws.com`
+    case 'S3':
+      return awsParams?.Bucket
+        ? `${awsParams.Bucket}.s3.${region}.amazonaws.com`
+        : `s3.${region}.amazonaws.com`
+    default:
+      return null
```
watson: Tiny nit: This default case is not needed, as JavaScript it will just "return" `undefined` if there's no `return` statement executed. And you're not checking if the value is `null` anyway when calling the function.

```suggestion
```

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -243,4 +297,28 @@ function getHooks (config) {
   return { request }
 }
 
+function getHostname (store, region) {
+  if (!store) return null
```
watson: Tiny nit: you can just use `return` without any value. That will make the function return `undefined` and since you're not actively checking if the return value is `null`, it's going to result in the same behavior 😃 

```suggestion
  if (!store) return
```

packages/datadog-plugin-aws-sdk/test/serverless-peer-service.spec.js
```diff
@@ -0,0 +1,363 @@
...
+          s3 = new AWS.S3({ endpoint: 'http://127.0.0.1:4566', s3ForcePathStyle: true, region: 'us-east-1' })
+
+          if (s3ClientName === 'aws-sdk') {
+            s3.api.globalEndpoint = '127.0.0.1'
+          }
+
+          s3.createBucket({ Bucket: bucketName }, (err) => {
+            if (err) return done(err)
+            done()
+          })
```
watson: Same behavior, but less verbose:

```suggestion
          s3.createBucket({ Bucket: bucketName }, done)
```

packages/datadog-plugin-aws-sdk/test/serverless-peer-service.spec.js
```diff
@@ -0,0 +1,363 @@
...
+
+        before(done => {
+          AWS = require(`../../../versions/${sqsClientName}@${version}`).get()
+
+          sqs = new AWS.SQS({ endpoint: 'http://127.0.0.1:4566', region: 'us-east-1' })
+          sqs.createQueue(queueOptions, (err, res) => {
+            if (err) return done(err)
+
+            done()
+          })
```
watson: Same behavior, but less verbose:

```suggestion
          sqs.createQueue(queueOptions, done)
```

packages/datadog-plugin-aws-sdk/test/serverless-peer-service.spec.js
```diff
@@ -0,0 +1,363 @@
...
+
+            if (result && typeof result.then === 'function') {
+              return result
+            }
+
+            if (result && typeof result.promise === 'function') {
+              return result.promise()
+            }
+
+            return require('util').promisify(fn)(...args)
```
watson: Please move `const { promisify } = require('util')` up to the top require section of the file, and then:

```suggestion
            return promisify(fn)(...args)
```

packages/datadog-plugin-aws-sdk/test/serverless-peer-service.spec.js
```diff
@@ -0,0 +1,363 @@
...
+        }
+
+        before(async () => {
+          AWS = require(`../../../versions/${dynamoClientName}@${version}`).get()
+
+          dynamo = new AWS.DynamoDB({ endpoint: 'http://127.0.0.1:4566', region: 'us-east-1' })
+
+          if (typeof dynamo.createTable === 'function') {
+            const createTable = toPromise(dynamo, dynamo.createTable)
+            try { await createTable(getCreateTableParams()) } catch (_) {}
```
watson, line 49: Consider adding a comment of why this error is ignored

packages/datadog-plugin-aws-sdk/test/serverless-peer-service.spec.js
```diff
@@ -0,0 +1,363 @@
...
+
+            params.requestHandler = new NodeHttpHandler()
+          }
+
+          kinesis = new AWS.Kinesis(params)
+
+          kinesis.createStream({
+            StreamName: streamName,
+            ShardCount: 1
+          }, (err, _) => {
```
watson: No need to specify the 2nd argument when you don't need it:

```suggestion
          }, (err) => {
```

packages/datadog-plugin-aws-sdk/test/serverless-peer-service.spec.js
```diff
@@ -0,0 +1,363 @@
...
+
+          helpers.putTestRecord(kinesis, streamName, helpers.dataBuffer, e => e && done(e))
+        })
+      })
+
+      describe('SNS-Serverless', () => {
+        let sns
+        let sqs
+        let TopicArn
+        let QueueUrl
```
watson: Tiny nit (just less verbose):

```suggestion
        let sns, sqs, TopicArn, QueueUrl
```

packages/datadog-plugin-aws-sdk/test/serverless-peer-service.spec.js
```diff
@@ -0,0 +1,361 @@
...
+            expect(awsSpan.meta['peer.service'],
+              'expected the aws span to have a peer.service tag'
+            ).to.exist
+            expect(httpSpan.meta['peer.service'],
+              'expected the underlying http span to have a peer.service tag'
+            ).to.exist
+            expect(awsSpan.meta['peer.service'],
+              'expected the aws span to have the same peer.service tag as the underlying http span'
+            ).to.equal(httpSpan.meta['peer.service'])
+          }).then(done, done)
```
wconti27: tests and changes look good to me, nit, we can abstract this assertion code, its duplicated multiple times.

wconti27 (commented):
just some small nits

wconti27 (approved):
LGTM

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -34,6 +59,12 @@ class BaseAwsSdkPlugin extends ClientPlugin {
   constructor (...args) {
     super(...args)
 
+    // subscribe to network channels for serverless environment peer.service propagation
+    if (this._tracerConfig?._isInServerlessEnvironment()) {
+      networkChannelsStart.forEach(ch => this.addSub(ch, () => {}))
+      networkChannelsEnd.forEach(ch => this.addSub(ch, setPeerServiceTag))
```
rochdev: Why are these in `aws-sdk` instead of in the corresponding plugins?

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -34,6 +59,12 @@ class BaseAwsSdkPlugin extends ClientPlugin {
   constructor (...args) {
     super(...args)
 
+    // subscribe to network channels for serverless environment peer.service propagation
+    if (this._tracerConfig?._isInServerlessEnvironment()) {
+      networkChannelsStart.forEach(ch => this.addSub(ch, () => {}))
```
rochdev: Why the empty subscriber?

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -8,6 +8,31 @@ const coalesce = require('koalas')
...
+
+// channels for Lambda to carry peer.service
+const networkChannelsEnd = [
+  'apm:http:client:request:finish',
+  'apm:http2:client:request:end',
+  'apm:net:tcp:finish',
+  'apm:dns:lookup:finish'
+]
+
+// handler to attach peer.service to underlying network spans
```
rochdev: This should be done at that level instead of here.

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -34,6 +59,12 @@ class BaseAwsSdkPlugin extends ClientPlugin {
   constructor (...args) {
     super(...args)
 
+    // subscribe to network channels for serverless environment peer.service propagation
+    if (this._tracerConfig?._isInServerlessEnvironment()) {
```
rochdev: Why is this needed? What would be the downside of always setting the peer service in the same way?

rochdev (changes requested):

packages/datadog-plugin-aws-sdk/src/base.js
zarirhamza: The corresponding plugins do not have the necessary information at their span level to determine hostname (via things like awsRegion and awsService)

packages/datadog-plugin-aws-sdk/src/base.js
zarirhamza: Lambdas have two modes - cold/warm start. During a cold start, the serverless team disables all plugins unless explicitly required. This short circuits the channel pub/sub which can be seen via live testing as well. The start subscription ensures that the channel is still in use even in a cold start

packages/datadog-plugin-aws-sdk/src/base.js
zarirhamza: Addressed in previous comment - underlying network spans do not have the means to reproduce the correct peer.service value

packages/datadog-plugin-aws-sdk/src/base.js
zarirhamza: These fixes are targeted towards serverless systems and their interactions on the service map. The serverless repo produces spans that are incompatible with inferred spans and these fixes are to mitigate that

packages/datadog-plugin-http2/src/client.js
```diff
@@ -110,16 +108,14 @@ class Http2ClientPlugin extends ClientPlugin {
     addHeaderTags(store.span, headers, HTTP_RESPONSE_HEADERS, this.config)
   }
 
-  _onError (ctx) {
-    const { span, error } = ctx
+  _onError (currentStore, error) {
+    const { span } = currentStore
     span.setTag('error', error)
-    super.finish(ctx)
+    super.finish({ currentStore })
```
rochdev: This is making assumptions about what the parent `finish` expects. The `ctx` should be passed as before instead, but with the correct parameters received.

zarirhamza: Fixed

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -85,6 +101,17 @@ class BaseAwsSdkPlugin extends ClientPlugin {
...
+      if (this._tracerConfig?._isInServerlessEnvironment()) {
+        const hostname = getHostname(store, region)
+        if (!hostname) return
+        span.setTag('peer.service', hostname)
+
+        const peerServerlessStore = storage('peerServerless').getStore()
+        if (peerServerlessStore) {
+          peerServerlessStore.peerHostname = hostname
+        }
+      }
```
rochdev:
```suggestion
      if (!this._tracerConfig?._isInServerlessEnvironment()) return

      const hostname = getHostname(store, region)
      if (!hostname) return
      span.setTag('peer.service', hostname)

      const peerServerlessStore = storage('peerServerless').getStore()
      if (peerServerlessStore) {
        peerServerlessStore.peerHostname = hostname
      }
```

Nit: this could be a short-circuit to avoid additional indentation

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -73,9 +73,25 @@ class BaseAwsSdkPlugin extends ClientPlugin {
         span.addTags(requestTags)
       }
 
+      this.enter(span)
       const store = storage('legacy').getStore()
 
-      this.enter(span, store)
+      const peerServerlessStorage = storage('peerServerless')
+      if (this._tracerConfig?._isInServerlessEnvironment()) {
```
rochdev: Nit: similar to below, this could be a short-circuit for the rest of the function:

```js
this.enter(span)

if (!this._tracerConfig?._isInServerlessEnvironment()) return

...
```

packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -243,4 +270,27 @@ function getHooks (config) {
   return { request }
 }
 
+function getHostname (store, region) {
```
rochdev, line 273: Not a blocker, but wouldn't we know this from the HTTP layer? At some point we will be actually calling out to the endpoint so we for sure have the information somewhere.

packages/datadog-plugin-aws-sdk/test/serverless-peer-service.spec.js
```diff
@@ -0,0 +1,319 @@
...
+  describe('Serverless', function () {
+    let tracer
+    setup()
+
+    withVersions('aws-sdk', ['aws-sdk', '@aws-sdk/smithy-client'], (version, moduleName) => {
+      let AWS
+
+      before(async () => {
+        tracer = require('../../dd-trace')
+        sinon.stub(Config.prototype, '_isInServerlessEnvironment').returns(true)
```
rochdev: Can we use the real configuration instead? Testing in this way doesn't guarantee that this actually works.

packages/datadog-plugin-aws-sdk/test/serverless-peer-service.spec.js
```diff
@@ -0,0 +1,319 @@
...
+        // force a reload of the plugins by deleting the plugin instance
+        delete tracer._pluginManager._pluginsByName['aws-sdk']
```
rochdev: Why is this needed?

packages/datadog-plugin-aws-sdk/test/serverless-peer-service.spec.js
```diff
@@ -0,0 +1,319 @@
...
+          if (typeof dynamo.createTable === 'function') {
+            const createTable = toPromise(dynamo, dynamo.createTable)
+            try { await createTable(getCreateTableParams()) } catch (_) {}
+          }
+        })
+
+        it('propagates peer.service from aws span to underlying http span', async () => {
+          const send = toPromise(dynamo, dynamo.putItem)
+
+          const tracesPromise = agent.assertSomeTraces(assertPeerServicePropagation)
```
rochdev: Can real values be asserted explicitly instead of just checking for the general structure?

packages/datadog-plugin-aws-sdk/src/base.js
zarirhamza: This isn't guaranteed varying on the service called and version of the sdk. The hostname we are using could be translated differently before making an outgoing http call. Also this does not exist on the DNS/TCP layer

packages/datadog-plugin-aws-sdk/test/serverless-peer-service.spec.js
zarirhamza: The way the check is normally done is by reading an env var that is set by the Lambda. However, if we trigger that here we create all sorts of problems because the environment we are running tests on is not a lambda environment. The test agent does not have a LogExporter set if triggered in a lambda environment, etc, etc.

For the scope of this feature, mocking the config call should be enough otherwise we would have to change the testagent

packages/datadog-plugin-aws-sdk/test/serverless-peer-service.spec.js
zarirhamza: The config object is only read at the time of loading, so we need to induce a reload by deleting the plugin. That way the newly changed Config var is read at the proper time.

packages/datadog-plugin-aws-sdk/test/serverless-peer-service.spec.js
rochdev: Would something like this work? https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-plugin-fetch/test/index.spec.js#L616-L617

packages/datadog-plugin-aws-sdk/test/serverless-peer-service.spec.js
rochdev: What I meant is to have the individual expectations in the test itself, with the values directly. This is for 3 reasons:

1) By checking that a value exists but not checking the value, we're not really ensuring that the values are correct.
2) Indirection in tests make them difficult to read and figure out what the expectation actually is.
3) Debugging a test is made a lot more difficult with code reuse as the same helper ends up running for every test.

PR #5977, opened 2025-06-27
test: skip undici v7.11 tests on Node.js 18

BridgeAR:
Undici 7 only supports Node.js 20 and higher and uses newer methods that are not supported in Node.js 18.

### DIFF
--- a/packages/datadog-plugin-undici/test/index.spec.js
+++ b/packages/datadog-plugin-undici/test/index.spec.js
@@ -6,6 +6,8 @@ const tags = require('../../../ext/tags')
 const { expect } = require('chai')
 const { rawExpectedSchema } = require('./naming')
 const { ERROR_MESSAGE, ERROR_TYPE, ERROR_STACK } = require('../../dd-trace/src/constants')
+const semver = require('semver')
+const { NODE_MAJOR } = require('../../../version')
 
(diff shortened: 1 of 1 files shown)


packages/datadog-plugin-undici/test/index.spec.js
```diff
@@ -18,7 +20,10 @@ describe('Plugin', () => {
   let appListener
 
   describe('undici-fetch', () => {
-    withVersions('undici', 'undici', version => {
+    withVersions('undici', 'undici', (version, _moduleName_, realVersion) => {
+      if (NODE_MAJOR < 20 && semver.satisfies(realVersion, '>=7.11.0')) {
+        return
+      }
```
watson: I haven't tested it, but shouldn't this do the same, but produce better CI output?

```suggestion
    withVersions('undici', 'undici', NODE_MAJOR < 20 ? '<7.11.0' : '*', (version, _moduleName_, realVersion) => {
```

BridgeAR: The describe blocks did not show up, so the output should be the same, while I still like this more and it will actually test the latest v7 as well in Node.js 18, so that's nice!

BridgeAR: Sadly, our tests still do not pick up v7 in Node.js 18. Even with this change.

PR #5980, opened 2025-06-27
support blocking on fastify multipart

IlyasShabi:
### What does this PR do?
Support blocking on fastify multipart body

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -51,8 +51,7 @@ export default [
       'integration-tests/esbuild/aws-sdk-out.js', // Generated
       'packages/datadog-plugin-graphql/src/tools/index.js', // Inlined from apollo-graphql
       'packages/datadog-plugin-graphql/src/tools/signature.js', // Inlined from apollo-graphql
-      'packages/datadog-plugin-graphql/src/tools/transforms.js', // Inlined from apollo-graphql
-      'packages/dd-trace/src/guardrails/**/*' // Guardrails contain very old JS
+      'packages/datadog-plugin-graphql/src/tools/transforms.js' // Inlined from apollo-graphql
     ]
   },
   { name: '@eslint/js/recommended', ...eslintPluginJs.configs.recommended },
@@ -301,21 +300,17 @@ export default [
   },
   {
     name: 'dd-trace/defaults',
-
     plugins: {
       '@stylistic': eslintPluginStylistic,
       import: eslintPluginImport,
       n: eslintPluginN
     },
-
     languageOptions: {
       globals: {
         ...globals.node
       },
-
       ecmaVersion: 2022
     },
-
     settings: {
       node: {
         // Used by `eslint-plugin-n` to determine the minimum version of Node.js to support.
@@ -325,7 +320,6 @@ export default [
         version: '>=18.0.0'
       }
     },
-
     rules: {
       '@stylistic/max-len': ['error', { code: 120, tabWidth: 2, ignoreUrls: true, ignoreRegExpLiterals: true }],
       '@stylistic/object-curly-newline': ['error', { multiline: true, consistent: true }],
@@ -437,6 +431,81 @@ export default [
     }
   },
   {
+    name: 'dd-trace/defaults/v0.8-oldest',
+    plugins: {
+      n: eslintPluginN
+    },
+    files: [
+      'init.js',
+      'packages/dd-trace/src/guardrails/**/*',
+      'version.js'
+    ],
+    settings: {
+      node: {
+        version: '>=0.8.0'
+      }
+    },
+    rules: {
+      'eslint-rules/eslint-process-env': 'off', // Would require us to load a module outside the guardrails directory
+      'n/no-unsupported-features/es-builtins': ['error', {
+        // The following are false positives that are supported in Node.js 0.8.0
+        ignores: [
+          'JSON',
+          'JSON.stringify',
+          'parseInt',
+          'String'
+        ]
+      }],
+      'n/no-unsupported-features/es-syntax': ['error', {
+        // The following are false positives that are supported in Node.js 0.8.0
+        ignores: [
+          'array-prototype-indexof',
+          'json'
+        ]
+      }],
+      'no-var': 'off', // Only supported in Node.js 6+
+      'object-shorthand': 'off', // Only supported in Node.js 4+
+      'unicorn/prefer-includes': 'off', // Only supported in Node.js 6+
+      'unicorn/prefer-number-properties': 'off', // Only supported in Node.js 0.12+
+      'unicorn/prefer-optional-catch-binding': 'off', // Only supported in Node.js 10+
+      'unicorn/prefer-set-has': 'off', // Only supported in Node.js 0.12+
+      'unicorn/prefer-string-replace-all': 'off' // Only supported in Node.js 15+
+    }
+  },
+  {
+    name: 'dd-trace/defaults/v16-oldest',
+    plugins: {
+      n: eslintPluginN
+    },
+    files: [
+      'packages/datadog-plugin-cypress/src/support.js'
+    ],
+    settings: {
+      node: {
+        version: '>=16.0.0'
+      }
+    }
+  },
+  {
+    name: 'dd-trace/defaults/v18-latest',
+    plugins: {
+      n: eslintPluginN
+    },
+    files: [
+      'benchmark/**/*',
+      'scripts/**/*',
+      ...TEST_FILES
+    ],
+    settings: {
+      node: {
+        version: '>=18' // These files don't have to support the oldest v18 release
+      }
+    },
+    rules: {
+      'n/no-unsupported-features/node-builtins': ['error', { allowExperimental: true }]
+    }
+  },
+  {
     ...eslintPluginCypress.configs.recommended,
     files: [
       'packages/datadog-plugin-cypress/src/support.js'
@@ -462,9 +531,7 @@ export default [
       'scripts/**/*'
     ],
     rules: {
-      'n/no-unsupported-features/node-builtins': ['error', {
-        allowExperimental: true
-      }]
+      'n/no-unsupported-features/node-builtins': ['error', { allowExperimental: true }]
     }
   },
   {
@@ -494,9 +561,7 @@ export default [
       'mocha/no-top-level-hooks': 'off',
       'n/handle-callback-err': 'off',
       'n/no-missing-require': 'off',
-      'n/no-unsupported-features/node-builtins': ['error', {
-        allowExperimental: true
-      }],
+      'n/no-unsupported-features/node-builtins': ['error', { allowExperimental: true }],
       'require-await': 'off'
     }
   },
--- a/init.js
+++ b/init.js
@@ -1,7 +1,5 @@
 'use strict'
 
-/* eslint-disable no-var */
-
 // TODO: It shouldn't be necessary to disable n/no-unpublished-require - Research
 // eslint-disable-next-line n/no-unpublished-require
 var guard = require('./packages/dd-trace/src/guardrails')
--- a/integration-tests/automatic-log-submission.spec.js
+++ b/integration-tests/automatic-log-submission.spec.js
@@ -69,7 +69,8 @@ describe('test visibility automatic log submission', () => {
       command: './node_modules/.bin/playwright test -c playwright.config.js',
       getExtraEnvVars: () => ({
         PW_BASE_URL: `http://localhost:${webAppPort}`,
-        TEST_DIR: 'ci-visibility/automatic-log-submission-playwright'
+        TEST_DIR: 'ci-visibility/automatic-log-submission-playwright',
+        DD_TRACE_DEBUG: 1
       })
     }
   ]
@@ -135,6 +136,10 @@ describe('test visibility automatic log submission', () => {
 
         childProcess.on('exit', () => {
           Promise.all([logsPromise, eventsPromise]).then(() => {
+            if (name === 'playwright') {
+              // eslint-disable-next-line no-console
+              console.log(testOutput)
+            }
             const { logSpanId, logTraceId } = logIds
             const { testSpanId, testTraceId } = testIds
             assert.include(testOutput, 'Hello simple log!')
--- a/packages/datadog-plugin-aws-sdk/test/sqs.spec.js
+++ b/packages/datadog-plugin-aws-sdk/test/sqs.spec.js
@@ -1,15 +1,14 @@
 'use strict'
 
+const { randomUUID } = require('crypto')
 const sinon = require('sinon')
 const { withNamingSchema, withPeerService, withVersions } = require('../../dd-trace/test/setup/mocha')
 const agent = require('../../dd-trace/test/plugins/agent')
 const { setup } = require('./spec_helpers')
 const semver = require('semver')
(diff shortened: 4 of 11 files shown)


packages/dd-trace/test/plugins/externals.json
```diff
@@ -146,7 +146,7 @@
   "fastify": [
     {
       "name": "fastify",
-      "versions": ["2.15.0", "3.0.0", "3.9.2"]
```
simon-id: not sure it's a good idea to remove this. We should ask @rochdev who introduced it in the first place in this commit https://github.com/DataDog/dd-trace-js/commit/0a1befc806751bd7ae48f2b8099ef3a56ad33449

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -536,6 +537,94 @@ withVersions('fastify', 'fastify', '>=2', version => {
...
+
+      before(function () {
+        // @fastify/multipart is not compatible with Fastify 2.x
+        if (version.startsWith('2')) {
+          this.skip()
+        }
+
+        // @fastify/multipart v6.x is not compatible with Fastify >3
+        // For Fastify v5.x, you need @fastify/multipart v9.x or newer
+        if (semver.intersects(version, '>3.0.0') && multipartVersion === '6.0.0') {
```
simon-id: you're only testing if it's exactly version 6.0.0, not if it's 6.x, is that intended ?

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -536,6 +537,94 @@ withVersions('fastify', 'fastify', '>=2', version => {
...
+      })
+
+      afterEach(() => {
+        appsec.disable()
+      })
+
+      after(() => {
+        if (server) {
+          server.close()
+        }
```
watson: We can just use optional chaining:

```suggestion
        server?.close()
```

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -536,6 +537,94 @@ withVersions('fastify', 'fastify', '>=2', version => {
...
+        })
+
+        getPort().then((port) => {
+          app.listen({ port }, () => {
+            axios = Axios.create({ baseURL: `http://localhost:${port}` })
+            done()
+          })
+          server = app.server
+        }).catch(done)
+      })
```
watson: The Fastify `listen` function returns a promise if not provided a callback, so you can make the `before` hook async:

```suggestion
      before(async () => {
        const fastify = require(`../../../../versions/fastify@${version}`).get()
        const fastifyMultipart = require(`../../../../versions/@fastify/multipart@${multipartVersion}`).get()

        const app = fastify()

        app.register(fastifyMultipart, { attachFieldsToBody: true })

        app.post('/', (request, reply) => {
          uploadSpy()
          reply.send('DONE')
        })

        const port = await getPort()
        await app.listen({ port })

        axios = Axios.create({ baseURL: `http://localhost:${port}` })
        server = app.server
      })
```

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
IlyasShabi: Im avoiding the first version `6.0.0` defined by `6>=` as an example of 6.x

simon-id: yes but it's not the only version that you need to skip no ? you wrote in the comment it's not compatible with v6.x, thus you should skip all v6.x no ?

packages/dd-trace/test/plugins/externals.json
uurien: I would say that we should not remove this. Specific versions added here are to force the CI to be executed in these versions, probably something changed in these releases and we don't want to add a regression for these versions in the future.

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -124,6 +124,16 @@ function preHandler (request, reply, done) {
   if (!reply || typeof reply.send !== 'function') return done()
 
   const req = getReq(request)
+  const res = getRes(reply)
+
+  // For multipart/form-data, the body is not available until after preValidation hook
+  if (bodyParserReadCh.hasSubscribers && request.body) {
+    const abortController = new AbortController()
+
+    bodyParserReadCh.publish({ req, res, body: request.body, abortController })
```
uurien, line 136: For not `multipart/form-data`, aren't we sending the event twice? One in the original `preValidation` and other here.

Should we use only this one? Or publish this one only if this is a `multipart/form-data`?

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -536,6 +522,89 @@ withVersions('fastify', 'fastify', '>=2', version => {
...
+
+      before(function () {
+        // @fastify/multipart is not compatible with Fastify 2.x
+        if (version.startsWith('2')) {
+          this.skip()
+        }
+
+        // @fastify/multipart v6.x is not compatible with Fastify >3
+        // For Fastify v5.x, you need @fastify/multipart v9.x or newer
+        if (semver.intersects(version, '>3.0.0') && multipartVersion === '6.0.0') {
```
simon-id: is it >=3 instead ?
```suggestion
        if (semver.intersects(version, '>=3.0.0') && multipartVersion === '6.0.0') {
```

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
simon-id:
```suggestion
        if (semver.intersects(version, '>=3.0.0') && semver.intersects(multipartVersion, '<7.0.0')) {
```

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -536,6 +522,89 @@ withVersions('fastify', 'fastify', '>=2', version => {
...
     })
   })
+
+  describe('Suspicious request blocking - multipart', () => {
+    withVersions('fastify', '@fastify/multipart', multipartVersion => {
+      let server, uploadSpy, axios
+
+      before(function () {
+        // @fastify/multipart is not compatible with Fastify 2.x
+        if (version.startsWith('2')) {
```
CarlesDD: version check with semver is more accurate
```suggestion
        if (semver.intersects(version, '<= 2.x')) {
```

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -536,6 +522,89 @@ withVersions('fastify', 'fastify', '>=2', version => {
...
+      after(() => {
+        server?.close()
+        return agent.close({ ritmReset: false })
+      })
+
+      it('should not block the request without an attack', async () => {
+        const form = new FormData()
+        form.append('key', 'value')
+
+        const res = await axios.post('/', form)
```
CarlesDD, line 602: worth checking res status is 200 if not blocked

packages/datadog-instrumentations/src/fastify.js
IlyasShabi: We could not have it in one place:
- we should block before schema validation so it's mandatory to have it in preValidation
- we should block when we have the body, and on multipart use case the body is here after preValidation
However I added a WeakSet to avoid sending the body to the waf twice

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
IlyasShabi: Done

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
IlyasShabi: Done

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -536,6 +520,93 @@ withVersions('fastify', 'fastify', '>=2', version => {
...
     })
   })
+
+  describe('Suspicious request blocking - multipart', () => {
+    withVersions('fastify', '@fastify/multipart', multipartVersion => {
+      let server, uploadSpy, axios
+
+      before(function () {
+        // @fastify/multipart is not compatible with Fastify 2.x
+        if (semver.intersects(version, '<= 2.x')) {
```
uurien: I don't know if this is exactly the case here, but to compare versions, we should use the loadedVersion, and not the version. For example, `version` could be `>0.10`, and the loaded version for that will be the latest, maybe `6.0`.

The callback of `withVersions` method returns 3 parameters, `version` (what we are using now), `module` (IIRC, module name), `loadedVersion` (the real version that will be loading).

And other thing here, why are we using `2.x` and not just `2`?
```suggestion
        if (semver.intersects(version, '<= 2')) {
```

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -536,6 +520,93 @@ withVersions('fastify', 'fastify', '>=2', version => {
...
+    withVersions('fastify', '@fastify/multipart', (multipartVersion, _, multipartLoadedVersion) => {
+      let server, uploadSpy, axios
+
+      before(function () {
+        // @fastify/multipart is not compatible with Fastify 2.x
+        if (semver.intersects(fastifyLoadedVersion, '2')) {
+          this.skip()
+        }
+
+        // 6.x works for this specific Fastify patch version.
```
uurien: I think that this convination DOESN'T work, and it is why we are skipping.
```suggestion
        // 6.x doesn't work for this specific Fastify patch version.
```

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
IlyasShabi: I modified the logic and >3 is the expected one

uurien (approved):

packages/dd-trace/test/plugins/externals.json
simon-id: @IlyasShabi what did Roch say ?

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -536,6 +520,93 @@ withVersions('fastify', 'fastify', '>=2', version => {
...
+        }
+
+        // 6.x doesn't work for this specific Fastify patch version.
+        if (semver.intersects(fastifyLoadedVersion, '3.9.2') && semver.intersects(multipartLoadedVersion, '>6')) {
+          this.skip()
+        }
+
+        if (semver.intersects(fastifyLoadedVersion, '>3') && semver.intersects(multipartLoadedVersion, '<7.0.0')) {
+          this.skip()
+        }
```
simon-id: are you sure this is what you mean ?
`>6` won't match 6.0.0, but it's roughly equivalent to >=7.0.0 no ?
same thing for `>3`

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -29,13 +30,10 @@ withVersions('fastify', 'fastify', '>=2', version => {
...
       })
 
-      getPort().then((port) => {
-        app.listen({ port }, () => {
-          axios = Axios.create({ baseURL: `http://localhost:${port}` })
-          done()
-        })
-        server = app.server
-      })
+      const port = await getPort()
```
simon-id: we need to stop using getPort, just let the OS choose a port for you, by passing no port argument, and getting the assigned one with app.server.address().port or soemthing like that

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -95,13 +93,10 @@ withVersions('fastify', 'fastify', '>=2', version => {
...
+      await app.listen({ port })
```
simon-id: same thing

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
simon-id: check https://github.com/DataDog/dd-trace-js/pull/6060 for examples

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -205,13 +200,10 @@ withVersions('fastify', 'fastify', '>=2', version => {
...
```
simon-id: same

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -295,13 +287,10 @@ withVersions('fastify', 'fastify', '>=2', version => {
...
+      axios = Axios.create({ baseURL: `http://localhost:${port}` })
```
simon-id: same

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -577,13 +648,10 @@ describe('Api Security - Fastify', () => {
...
 
-      getPort().then((port) => {
-        app.listen({ port }, () => {
-          axios = Axios.create({ baseURL: `http://localhost:${port}` })
-          done()
-        })
-        server = app.server
-      })
+      const port = await getPort()
+      await app.listen({ port })
```
simon-id: same thing

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -536,6 +520,93 @@ withVersions('fastify', 'fastify', '>=2', version => {
...
+
+        app.register(fastifyMultipart, { attachFieldsToBody: true })
+
+        app.post('/', (request, reply) => {
+          uploadSpy()
+          reply.send('DONE')
+        })
+
+        const port = await getPort()
+        await app.listen({ port })
```
simon-id: same thing

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
IlyasShabi: I asked Roch that I will take care of Appsec because of this PR, I will do the same or maybe wait for his PR https://github.com/DataDog/dd-trace-js/pull/6060/files#diff-854734aa9e96d9487e086516a67e0b9b0255b7e2b8f51dc37f303bc76b07ad4b to land

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
IlyasShabi: This will skip every combination where:
 – Fastify is higher than `3.0.0` (3.1.x, 4.x, …) and
 – @fastify/multipart is lower than `7.0.0` (6.x or 5.x)

So the test with fastify `3.0.0` and @fastify/multipart `6.0.0` pass

PR #5981, opened 2025-06-27
chore: update dependencies and dependabot config

BridgeAR:
This skips the Node.js types, since we need to manually update
those and pins most dependencies. The pinning is useful for security
and we are able to update the dependencies weekly anyway with
dependabot running.

The dependencies now all use the latest version. This is done to automatically
include security fixes and features by the peer dependencies for our users.
They should use a package-lock.json anyway, so that they can reliably get the
best experience.

We also don't need to open PRs anymore for our own minor and patch updates
that way. They will automatically be installed for users.

Graphql was installed as dev dependency in an old version, while it's unclear
why that was the case. I removed it as it was not required in the code.

### DIFF
--- a/integration-tests/cucumber/cucumber.spec.js
+++ b/integration-tests/cucumber/cucumber.spec.js
@@ -55,6 +55,7 @@ const {
   DD_CAPABILITIES_TEST_MANAGEMENT_QUARANTINE,
   DD_CAPABILITIES_TEST_MANAGEMENT_DISABLE,
   DD_CAPABILITIES_TEST_MANAGEMENT_ATTEMPT_TO_FIX,
+  DD_CAPABILITIES_FAILED_TEST_REPLAY,
   TEST_MANAGEMENT_IS_ATTEMPT_TO_FIX,
   TEST_HAS_FAILED_ALL_RETRIES,
   TEST_MANAGEMENT_ATTEMPT_TO_FIX_PASSED,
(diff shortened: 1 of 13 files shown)


.github/workflows/system-tests.yml
```diff
@@ -35,6 +35,7 @@ jobs:
     runs-on: ${{ contains(fromJSON('["CROSSED_TRACING_LIBRARIES", "INTEGRATIONS"]'), matrix.scenario) && 'ubuntu-latest-16-cores' || 'ubuntu-latest' }}
     needs:
       - get-scenarios
+    if: github.actor != 'dependabot[bot]'
```
rochdev: Shouldn't we be running system tests when updating dependencies?

BridgeAR: Theoretically yes, I would like them to run. Effectively we can't. They rely upon secrets that dependabot has no access to.

rochdev: Any way for give it access? We do that for our own bots, ideally we could do that for dependabot as well.

BridgeAR: I would rather not

rochdev: I think this is a serious enough limitation that we should find a way to make it work. Since all bots are effectively own by GitHub (including our own bots since they're hosted by GitHub) I don't really see the difference between accessing secrets with our own bots and with dependabot.

BridgeAR: I can also remove the skipping for now, while that means we can't merge any dependabot PR.
Thus, I definitely believe that is the better way around.

watson: I've added the `DD_API_KEY` from org2 prod to the Dependabot secrets vault (I'm pretty sure using prod is right, but please correct me here if I'm mistaken). This allows system tests to run successfully. I don't see an issue giving Dependabot access to this secret, as it's already owned and operated by GitHub.

PR #5982, opened 2025-06-27
ci: auto-dedupe yarn.lock file for PRs created by bots

watson:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


.github/scripts/yarn-dedupe.sh
```diff
@@ -0,0 +1,44 @@
...
+if git diff --exit-code yarn.lock; then
+  echo "✅ yarn.lock is already properly deduplicated"
+else
+  echo "📝 yarn.lock was modified by dedupe command"
+
+  PR_AUTHOR="${PR_AUTHOR:-}"
+  PR_USER_TYPE="${PR_USER_TYPE:-}"
+  IS_BOT=false
+
+  if [[ "$PR_USER_TYPE" == "Bot" ]] || [[ "$PR_AUTHOR" == *"bot"* ]] || [[ "$PR_AUTHOR" == "renovate"* ]] || [[ "$PR_AUTHOR" == "github-actions"* ]]; then
```
BridgeAR: I think you wanted to write dependabot, not renovate? :)

watson: No it was actually intentional. It was based of a question I asked AI on how to match bots in general in a GitHub Actions workflow. The main way to detect it is if `PR_USER_TYPE` is set to `Bot`, which many bots use these days. If that's not used by the bot correctly, fall back to check if it has the word `bot` in its name (which Dependabot does), if not, then check if it starts with either `renovate` or `github-actions`. But if you want me to simplify it I can. E.g. just check for `PR_USER_TYPE == "Bot"`, which I believe Dependabot is correctly is identifying itself as

BridgeAR: Let's limit it to dependabot for now, otherwise we try to accommodate situations that we likely won't face anytime soon and I like to limit code to what we need.

watson: Fixed 👍

BridgeAR (approved):

PR #5985, opened 2025-06-27
add experimental flag on plugins to disable them by default

rochdev:
### What does this PR do?

Add experimental flag on plugins to disable them by default.

### Motivation

Right now only disabling the instrumentation is possible by default. However, this is problematic because users interact with plugins, not instrumentation, so they would only know the name for the plugins. Also, instrumentations do more than just back a single plugin, so when the goal is to disable just tracing for a library for example, it shouldn't also disable any other existing features already built on top of the same instrumentation. Doing this at the plugin level instead solves these issues and provides a better and more granular user experience that focuses only on the functionality.

### DIFF
--- a/packages/datadog-instrumentations/src/helpers/register.js
+++ b/packages/datadog-instrumentations/src/helpers/register.js
@@ -9,7 +9,7 @@ const log = require('../../../dd-trace/src/log')
 const checkRequireCache = require('./check-require-cache')
 const telemetry = require('../../../dd-trace/src/guardrails/telemetry')
 const { isInServerlessEnvironment } = require('../../../dd-trace/src/serverless')
-const { isFalse, isTrue, normalizePluginEnvName } = require('../../../dd-trace/src/util')
+const { isFalse } = require('../../../dd-trace/src/util')
 const { getEnvironmentVariables } = require('../../../dd-trace/src/config-helper')
 
 const envs = getEnvironmentVariables()
@@ -25,20 +25,15 @@ const names = Object.keys(hooks)
 const pathSepExpr = new RegExp(`\\${path.sep}`, 'g')
 
 const disabledInstrumentations = new Set(
-  DD_TRACE_DISABLED_INSTRUMENTATIONS?.split(',').map(name => normalizePluginEnvName(name, true)) ?? []
+  DD_TRACE_DISABLED_INSTRUMENTATIONS?.split(',') ?? []
 )
-const reenabledInstrumentations = new Set()
 
 // Check for DD_TRACE_<INTEGRATION>_ENABLED environment variables
 for (const [key, value] of Object.entries(envs)) {
   const match = key.match(/^DD_TRACE_(.+)_ENABLED$/)
-  if (match && value) {
-    const integration = normalizePluginEnvName(match[1], true)
-    if (isFalse(value)) {
-      disabledInstrumentations.add(integration)
-    } else if (isTrue(value)) {
-      reenabledInstrumentations.add(integration)
-    }
+  if (match && isFalse(value)) {
+    const integration = match[1].toLowerCase()
+    disabledInstrumentations.add(integration)
   }
 }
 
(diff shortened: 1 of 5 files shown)


packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -25,20 +25,15 @@ const names = Object.keys(hooks)
 const pathSepExpr = new RegExp(`\\${path.sep}`, 'g')
 
 const disabledInstrumentations = new Set(
-  DD_TRACE_DISABLED_INSTRUMENTATIONS?.split(',').map(name => normalizePluginEnvName(name, true)) ?? []
+  DD_TRACE_DISABLED_INSTRUMENTATIONS?.split(',') ?? []
```
watson: No need for a fallback. These all result in an empty set: `new Set()`, `new Set([])`, `new Set(undefined)`

```suggestion
  DD_TRACE_DISABLED_INSTRUMENTATIONS?.split(',')
```

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -75,10 +69,6 @@ for (const packageName of names) {
   if (hook !== null && typeof hook === 'object') {
     if (hook.serverless === false && isInServerlessEnvironment()) continue
 
-    // some integrations are disabled by default, but can be enabled by setting
-    // the DD_TRACE_<INTEGRATION>_ENABLED environment variable to true
-    if (hook.disabled && !reenabledInstrumentations.has(normalizedPackageName)) continue
```
watson, line 80: I've not been able to find any hooks that are disabled by default currently. Am I correct in assuming that this was just an option, but that we currently didn't use it?

Does this PR also remove support for the `disabled` property on hooks?

watson (commented):
Maybe it's just me, but I find the PR description a little confusing, so I'm not sure exactly what it's supposed to do in details. E.g. as I wrote in one of my comments below, it seems like this PR is also deprecating a `disabled` property, though I'm not sure.

packages/datadog-instrumentations/src/helpers/register.js
rochdev: > I've not been able to find any hooks that are disabled by default currently. Am I correct in assuming that this was just an option, but that we currently didn't use it?

Yes this was a request to be able to disable new integrations before they go GA but hasn't been used yet. The first time we wanted to use it it didn't quite work because it operates at the wrong layer, which makes usage confusing the the implementation prone to the issues described in the PR description.

> Does this PR also remove support for the disabled property on hooks?

It does as instrumentation should always be enabled by default and there are other means to disable them when needed.

watson: Got'ya 👍 Can we be sure that it's not used by 3rd part instrumentations in the wild, and hence should be treated as a breaking change?

rochdev: Third party instrumentation is not currently something that can exist as all of it is internal (and would basically require patching dd-trace to get it working).

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -25,20 +25,15 @@ const names = Object.keys(hooks)
...
-  if (match && value) {
-    const integration = normalizePluginEnvName(match[1], true)
-    if (isFalse(value)) {
-      disabledInstrumentations.add(integration)
-    } else if (isTrue(value)) {
-      reenabledInstrumentations.add(integration)
-    }
+  if (match && isFalse(value)) {
+    const integration = match[1].toLowerCase()
+    disabledInstrumentations.add(integration)
```
BridgeAR: I believe we should actually also handle this part different as well.
Until then, keeping the normalization here seems better than not doing that (otherwise it would not be possible to match a couple of integrations).

```suggestion
    disabledInstrumentations.add(normalizePluginEnvName(integration))
```

rochdev: This code has now been removed entirely.

rochdev: @BridgeAR Any other concerns?

BridgeAR (approved):
LGTM

Nice that the state is now coherent again :)

PR #5987, opened 2025-06-27
chore(llmobs): add enable and disable debug logs

sabrenner:
### What does this PR do?
Adds enable and disable debug logs for easier troubleshooting

### Motivation
Ease of troubleshooting escalations and issues.

### DIFF
--- a/packages/dd-trace/src/llmobs/index.js
+++ b/packages/dd-trace/src/llmobs/index.js
@@ -75,6 +75,7 @@ function enable (config) {
     spanWriter?.setAgentless(useAgentless)
 
     telemetry.recordLLMObsEnabled(startTime, config)
+    log.debug(`[LLMObs] Enabled LLM Observability with configuration: ${JSON.stringify(config.llmobs)}`)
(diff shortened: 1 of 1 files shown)


packages/dd-trace/src/llmobs/index.js
```diff
@@ -75,6 +75,7 @@ function enable (config) {
     spanWriter?.setAgentless(useAgentless)
 
     telemetry.recordLLMObsEnabled(startTime, config)
+    log.debug(`[LLMObs] Enabled LLM Observability with configuration: ${JSON.stringify(config.llmobs)}`)
```
BridgeAR:
```suggestion
    log.debug(`[LLMObs] Enabled LLM Observability with configuration: ${util.inspect(config.llmobs)}`)
```

JSON.stringify fails with e.g., symbols and BigInt. So it's safer to use that :)

sabrenner: makes sense! TIL 😄

PR #5988, opened 2025-06-27
guardrails: bail out if esbuild bundle

tlhunter:
### What does this PR do?
- instructs guardrails to disable SSI if the application is an ESBuild bundle
  - we cannot properly instrument any bundle externally
  - we can only instrument ESBuild bundles and only if the user built it with our plugin

### Motivation
- if a user has bundled their application with our ESBuild plugin and...
  - included a copy of the tracer using `require('dd-trace')` then...
    - it can get into a weird double-tracer state
    - there will be an internal (bundled) copy and an external (`--require` via guardrails) tracer
    - they fight with each other
  - but without a copy of the tracer then...
    - it can get into a weird state where the bundle does emit diagnostic channel messages
    - but the external tracer doesn't properly handle them
    - we could possibly support this in the future but it will require a lot of edge case testing
- if a user has bundled their application...
  - with Webpack or with ESBuild but without our ESBuild plugin or with another bundler then...
    - it can get into a weird state where the tracer only instruments internal module `require('http')` calls
    - but doesn't instrument userland module `require('express')` calls

### DIFF
--- a/packages/dd-trace/src/guardrails/index.js
+++ b/packages/dd-trace/src/guardrails/index.js
@@ -1,5 +1,6 @@
 'use strict'
 
+var fs = require('fs')
 var path = require('path')
 var Module = require('module')
 
@@ -8,11 +9,10 @@ var isTrue = require('./util').isTrue
 var log = require('./log')
 var telemetry = require('./telemetry')
 
+var BUNDLE_THRESHOLD = 1024 * 256 // arbitrary 256 KB filesize threshold
 var NODE_MAJOR = nodeVersion.NODE_MAJOR
 
 function guard (fn) {
-  var initBailout = false
-  var clobberBailout = false
   var forced = isTrue(process.env.DD_INJECT_FORCE)
   var engines = require('../../../../package.json').engines
   var versions = engines.node.match(/^>=(\d+) <(\d+)$/)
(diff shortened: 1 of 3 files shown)


packages/dd-trace/src/guardrails/index.js
```diff
@@ -59,6 +81,18 @@ function guard (fn) {
     }
   }
 
+  if (bundleBailout) {
+    initBailout = true
+    telemetry([
+      { name: 'abort', tags: ['reason:incompatible_bundle'] }, // TODO: what reason code
```
BridgeAR: Should this be defined before we land this?

packages/dd-trace/src/guardrails/index.js
```diff
@@ -59,6 +81,18 @@ function guard (fn) {
     }
   }
 
+  if (bundleBailout) {
```
BridgeAR: Should we report this reason as well, when we have already bailed and logged an incompatible version? I could imagine either is correct.

packages/dd-trace/src/guardrails/index.js
```diff
@@ -59,6 +81,18 @@ function guard (fn) {
...
+    initBailout = true
+    telemetry([
+      { name: 'abort', tags: ['reason:incompatible_bundle'] }, // TODO: what reason code
+      { name: 'abort.runtime', tags: [] }
+    ])
+    log.info('Aborting application instrumentation since application is bundled.')
+    if (forced) {
+      log.info('DD_INJECT_FORCE enabled, allowing unsupported bundling and continuing.')
+    }
```
BridgeAR: Please move this inside of the actual executed part. That way we can deduplicate the code and only run it if we actually get there.

packages/dd-trace/src/guardrails/index.js
```diff
@@ -38,6 +41,25 @@ function guard (fn) {
...
+      // this only checks for ESBuild bundles, but there are many other bundlers to consider
+      if (
+        contents.length > BUNDLE_THRESHOLD &&
+        contents.indexOf('__defProp') !== -1 &&
+        contents.indexOf('Bundled license information') !== -1
+      ) {
+        bundleBailout = true
+      }
+    } catch (_err) {
+      // it's a noop if we wan't access the entrypoint
```
BridgeAR: I am not sure what the comment stands for. If the entrypoint is not accessible, it would likely fail as soon as it is loaded by Node.js, right?

packages/dd-trace/src/guardrails/index.js
```diff
@@ -38,6 +41,25 @@ function guard (fn) {
...
+
+    try {
+      // If the application has been bundled then we probably cannot instrument it since the require() calls are gone
+      // If the app was bundled using our ESBuild plugin it should have a copy of the tracer insde of it anyway.
+
+      var contents = fs.readFileSync(entrypoint, 'utf8')
+
+      // this only checks for ESBuild bundles, but there are many other bundlers to consider
+      if (
+        contents.length > BUNDLE_THRESHOLD &&
```
BridgeAR: If we limit this, should we not just check for the size first and then read the file? Otherwise I don't think we should limit the size at all, since we already did the heavy lifting by reading the file.

packages/dd-trace/src/guardrails/index.js
tlhunter: I do believe this should be some sort of standardized / centralized value but I'm not sure who would own that.

packages/dd-trace/src/guardrails/index.js
tlhunter: What if someone does `node --require dd-trace/init -e 'console.log("hi")'`?

BridgeAR (commented):
The logic seems fine, while the state handling becomes difficult to follow.

We now have four states (2 ** 4 === 16) and at least for me, it is starting to become difficult to follow what is happening. I wonder if we could simplify that a bit.

PR #6003, opened 2025-07-01
instrumentation: fix oracle db connection string

BridgeAR:
Oracle supports three different ways to pass through conenctions. We only supported URLs so far. Others were unsupported. Instead of relying on the input, this changes the implementation to use the drivers own information to get that. This reduces the overhead of the instrumentation and we should always get the correct information, no matter how the user passes the information through.

### DIFF
--- a/packages/datadog-plugin-oracledb/src/connection-parser.js
+++ b/packages/datadog-plugin-oracledb/src/connection-parser.js
@@ -1,0 +1,35 @@
+const { URL } = require('url')
+const log = require('../../dd-trace/src/log')
+
+function parseOracleDescriptor (descriptor) {
+  const hostnameMatch = descriptor.match(/HOST\s*=\s*([^)]+)/i)
+  const hostname = hostnameMatch?.[1]
+
+  const portMatch = descriptor.match(/PORT\s*=\s*([^)]+)/i)
+  const port = portMatch?.[1]
+
+  const sidMatch = descriptor.match(/SID\s*=\s*([^)]+)/i)
+
+  const dbInstance = sidMatch?.[1] ?? descriptor.match(/SERVICE_NAME\s*=\s*([^)]+)/i)?.[1] ?? ''
+
+  return { hostname, port, dbInstance }
+}
+
+module.exports = function getDBInformation (connAttrs) {
+  // Users can pass either connectString or connectionString
+  const connectString = (connAttrs.connectString || connAttrs.connectionString).trim()
+  if (connectString.startsWith('(')) {
+    return parseOracleDescriptor(connectString)
+  }
+  try {
+    const url = new URL(`oracle://${connectString}`)
+    return {
+      hostname: url.hostname || 'localhost', // Default Oracle hostname
+      port: url.port || '1521', // Default Oracle port
+      dbInstance: url.pathname && url.pathname.slice(1) || 'XEPDB1' // Default Oracle service name
+    }
+  } catch (error) {
+    log.error('Invalid oracle connection string', error)
+    return {}
+  }
+}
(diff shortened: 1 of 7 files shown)


packages/datadog-plugin-oracledb/src/connection-parser.js
```diff
@@ -0,0 +1,35 @@
...
+  const connectString = (connAttrs.connectString || connAttrs.connectionString).trim()
+  if (connectString.startsWith('(')) {
+    return parseOracleDescriptor(connectString)
+  }
+  try {
+    const url = new URL(`oracle://${connectString}`)
+    return {
+      hostname: url.hostname || 'localhost', // Default Oracle hostname
+      port: url.port || '1521', // Default Oracle port
+      dbInstance: url.pathname && url.pathname.slice(1) || 'XEPDB1' // Default Oracle service name
```
BridgeAR, line 29: I wonder if we just want to use defaults in all cases, instead of only when a user does not provide a full connection string.

wconti27: I think its fine to use defaults in the case that the full string wasn't provided

wconti27 (dismissed):
This looks great, thank you for taking this on!

packages/datadog-plugin-oracledb/test/index.spec.js
```diff
@@ -295,18 +343,20 @@ describe('Plugin', () => {
             }
           )
 
-          it('should set the service name', done => {
-            agent.assertSomeTraces(traces => {
-              expect(traces[0][0]).to.have.property('name', expectedSchema.outbound.opName)
-              expect(traces[0][0]).to.have.property('service', 'custom')
-            }).then(done, done)
-            connection.execute(dbQuery)
+          it('should set the service name', async () => {
```
rochdev, line 348: I think the test case here is reversed with the above? Meaning that it says "with service undefined" but then defines a service string and vice versa.

packages/datadog-plugin-oracledb/test/index.spec.js
```diff
@@ -327,21 +377,23 @@ describe('Plugin', () => {
             {
               v0: {
                 opName: 'oracle.query',
-                serviceName: 'oracledb:1521/xepdb1'
+                serviceName: config.connectString,
```
rochdev, line 387: Nit: if the service function doesn't work properly and `connAttrs.connectString` is undefined, then this will match. Checking for a hard-coded value as before avoids this issue.

packages/dd-trace/src/service-naming/schemas/definition.js
```diff
@@ -3,20 +3,13 @@ class SchemaDefinition {
...
-  getSchemaItem (type, kind, plugin) {
-    const schema = this.schema
-    if (schema && schema[type] && schema[type][kind] && schema[type][kind][plugin]) {
-      return schema[type][kind][plugin]
-    }
-  }
-
   getOpName (type, kind, plugin, opts) {
-    const item = this.getSchemaItem(type, kind, plugin)
+    const item = this.schema[type][kind][plugin]
```
rochdev, line 7: Is this safe? Why was there a check before?

BridgeAR: I accidentally stumbled upon that method. The usage of the call required it to always be truthy before, otherwise it would have already failed.

packages/datadog-plugin-oracledb/test/index.spec.js
BridgeAR: Good catch!

packages/datadog-plugin-oracledb/src/connection-parser.js
BridgeAR: We can't definitely say it was provided or not, while that was my intention.
Sadly we can't differentiate the string in case a name like `sales` would be used (but that's also currently an issue). It would just end up being the wrong value in v5. In v6 it'll be fine.

packages/datadog-plugin-oracledb/test/index.spec.js
BridgeAR: I added a check for the string being passed through correctly.
It would otherwise not work between local / CI.

szegedi (approved):

PR #6004, opened 2025-07-01
tracing: add support for graphql esm applications 

wconti27:
### What does this PR do?
`graphql-yoga`, a package used for building graphql servers, uses `graphql-tools/executor` under the hood. `graphql-tools/executor` has separate code paths for `cjs` versus `esm` code. We already had tracing for the `cjs` code path, this PR adds tracing for the `esm` code path as well. Also adds esm tests for both a pure graphql server and a graphql-yoga server.

### Motivation
escalation

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-plugin-graphql/test/esm-test/esm.spec.js
+++ b/packages/datadog-plugin-graphql/test/esm-test/esm.spec.js
@@ -1,0 +1,102 @@
+'use strict'
+
+const {
+  FakeAgent,
+  createSandbox,
+  checkSpansForServiceName,
+  spawnPluginIntegrationTestProc
+} = require('../../../../integration-tests/helpers')
+const { assert } = require('chai')
+const { withVersions } = require('../../../dd-trace/test/setup/mocha')
+const axios = require('axios')
+const semver = require('semver')
+
+describe('Plugin (ESM)', () => {
+  describe('graphql (ESM)', () => {
+    let agent
+    let proc
+    let sandbox
+
+    withVersions('graphql', ['graphql'], version => {
+      before(async function () {
+        this.timeout(50000)
+        sandbox = await createSandbox([`'graphql@${version}'`, "'graphql-yoga@3.6.0'"], false, [
+          './packages/datadog-plugin-graphql/test/esm-test/*'])
+      })
+
+      after(async function () {
+        await sandbox.remove()
+      })
+
+      beforeEach(async () => {
+        agent = await new FakeAgent().start()
+      })
+
+      afterEach(async () => {
+        proc && proc.kill()
+        await agent.stop()
+      })
+
+      it('should instrument GraphQL execution with ESM', async () => {
+        const res = agent.assertMessageReceived(({ headers, payload }) => {
+          assert.propertyVal(headers, 'host', `127.0.0.1:${agent.port}`)
+          assert.isArray(payload)
+          assert.strictEqual(checkSpansForServiceName(payload, 'graphql.execute'), true)
+        })
+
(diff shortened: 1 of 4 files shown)


packages/datadog-plugin-graphql/test/esm-test/esm.spec.js
```diff
@@ -0,0 +1,102 @@
...
+describe('Plugin (ESM)', () => {
+  describe('graphql (ESM)', () => {
+    let agent
+    let proc
+    let sandbox
+
+    withVersions('graphql', ['graphql'], version => {
+      before(async function () {
+        this.timeout(50000)
+        sandbox = await createSandbox([`'graphql@${version}'`, "'graphql-yoga@3.6.0'"], false, [
```
uurien: I think that the version that you want to test here is the resolved version and not the version. `withVersions` could resolve multipe library versions for the same version, for example in this case, I can see in the logs that for version `>0.10.0` it is executing tests with. `0.10.0` and `16.11.0`. The version field for these values is going to be `>0.10.0` and iirc, `createSandbox` is going to do a `npm install graphql@>0.10.0` and you are going to test the same twice.

```suggestion
    withVersions('graphql', ['graphql'], (version, moduleName, resolvedVersion) => {
      before(async function () {
        this.timeout(50000)
        sandbox = await createSandbox([`'graphql@${resolvedVersion}'`, "'graphql-yoga@3.6.0'"], false, [
```

packages/datadog-plugin-graphql/test/esm-test/esm.spec.js
```diff
@@ -0,0 +1,102 @@
...
+        await res
+      }).timeout(50000)
+
+      // Only run GraphQL Yoga test for newer GraphQL versions (>= 15.0.0)
+      // GraphQL Yoga 3.6.0 requires newer GraphQL versions that have versionInfo export
+      // Extract version number from range strings like ">=0.10" or "^15.2.0"
+      const cleanVersion = version.replace(/^[>=^~]+/, '')
+      const coercedVersion = semver.coerce(cleanVersion)
+      if (coercedVersion && semver.gte(coercedVersion, '15.0.0')) {
+        it('should instrument GraphQL Yoga execution with ESM', async () => {
```
uurien, line 79: For the same reason, this test is not executed with graphql 16.11.0: https://github.com/DataDog/dd-trace-js/actions/runs/16010200086/job/45166110956?pr=6004
```
      with graphql >=0.10 (16.11.0)
```

packages/datadog-plugin-graphql/test/esm-test/esm.spec.js
uurien: Nah, you can ignore this comment

packages/datadog-plugin-graphql/test/esm-test/esm.spec.js
uurien: But here yes, you need the resolved version, because `>0.10.0` is resolving also to 16.11.0, and you are comparing `0.10.0 gte 15.0.0` instead of `16.11.0 gte 15.0.0`

wconti27: Good catch! I fixed it to use the resolved version, now the correct tests are running per the resolved version.

PR #6010, opened 2025-07-02
feat(tracer): add `recordException` API

dubloom:
### What does this PR do?
- Adds `recordException` API following this [RFC](https://docs.google.com/document/d/1-7CsG8GZVnNU198FfzavCNMI6IF0cGzFfS8cIz6Pu1g/edit?tab=t.0#heading=h.i8a6l13dr0px). `recordException` attaches error information to span through a span event.
- Refactor `sanitizeEventAttributes()`. Now if an attribute values does not follow the requirements (defined [here](https://docs.google.com/document/d/1cVod_VI7Yruq8U9dfMRFJd7npDu-uBpste2IB04GyaQ/edit?pli=1&tab=t.0#heading=h.z8rsw69ccced), the key is fully dropped.

### Motivation
We want to add a recordException API in every tracers so customers can use the feature without using the OpenTelemetry one which is not consistent across tracers. Note that the function is similar to the OpenTelemetry function but is not meant to be a replica.

About the second change, while implementing the API, I found out that `sanitizeEventAttributes` was not complete (lacking of check for integer range for instance) and not following the RFC.
Example:
[1, [1]] is not accepted and currently the function was only removing [1].

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -271,6 +271,15 @@ declare namespace tracer {
      * @returns {void}
      */
     addLink (context: SpanContext, attributes?: Object): void;
+
+    /**
+     * Records an exception as a span event.
+     *
+     * @param exception the exception to record.
+     * @param [attributes] additional attributes to add to the span event.
+     */
+    recordException(exception: Error | Object, attributes?: SpanAttributes): this;
+
   }
 
   /**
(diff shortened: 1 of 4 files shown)


rochdev (changes requested):
Why not add this to the tracer instead? This would decouple errors from spans, thus allowing us to eventually change the API so that we can record exceptions with tracing disabled.

BridgeAR:
Superseded

PR #6021, opened 2025-07-03
Remove vendored jsonpath-plus

CarlesDD:
### What does this PR do?
Removes vendored `jsonpath-plus` and use the published package.

### Motivation
The original motivation to vendor `jsonpath-plus` was the need to support Node.js v16 and it was dropped in dd-trace-js v5.

### Additional Notes
See #4785


package.json
```diff
@@ -101,6 +101,7 @@
     "import-in-the-middle": "^1.14.2",
     "istanbul-lib-coverage": "^3.2.2",
     "jest-docblock": "^29.7.0",
+    "jsonpath-plus": "^10.0.0",
```
BridgeAR: Nit
```suggestion
    "jsonpath-plus": "^10.3.0",
```

BridgeAR (approved):

PR #6024, opened 2025-07-03
ESLint: Explicitly load used plugins

watson:
### What does this PR do?

Explicitly list the plugins used in each ESLint rules group.

### Motivation

Makes it more clear which plugins are used where. The reason why it worked before was just a side-effect of a given plugin being loaded in another group. If that group was changed/removed, then it would suddenly break an unrelated group because that group didn't explicitly list the plugins it depended on.


watson:
Dear reviewer: The failing GitLab tests are actually not failing. If you look at [dd-gitlab/default-pipeline](https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-js/-/pipelines/69604017) they are passing. They were failing previously, but after rerunning them they passed. But for some reason they were not removed from the list.

PR #6029, opened 2025-07-04
fix: some guardrail telemetry should only be sent once

watson:
### What does this PR do?

Fixes bug in the guardrail telemetry `hasSeen` function which meant that all telemetry was always sent even though `abort` and `abort.integration` should only be sent once.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


watson:
* **#6029** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/6029?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/6029?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

uurien (approved):
If there is no way to add a test for guardrail/telemetry, LGTM

integration-tests/helpers/index.js
```diff
@@ -58,34 +58,35 @@ let sandbox
...
-    try {
-      assert.deepStrictEqual(msgs.length, 0)
-    } catch (e) {
-      // This console.log is useful for debugging telemetry. Plz don't remove.
-      // eslint-disable-next-line no-console
-      console.error('Expected no telemetry, but got:\n', msgs.map(msg => JSON.stringify(msg[1].points)).join('\n'))
-      throw e
-    }
+    assert.strictEqual(msgs.length, 0, `Expected no telemetry, but got:\n${
+      msgs.map(msg => JSON.stringify(msg[1].points)).join('\n')
```
BridgeAR, line 67: Let us please use util.inspect in the future. JSON.stringify could fail for multiple reasons :)

watson: I didn't want to mess with the output due to the code comment that Bryan made about not removing it becuase it was useful for debugging. But if you like I can - as long as it doesn't introduces linebreaks I guess

BridgeAR: AFAIC the comment would still be adhered. It would just not potentially fail during the JSON.stringify call itself. The reason why it is currently serialized is likely that the default console.error inspection depth is not deep enough and about potentially having it on a single line. Otherwise, that would not even be necessary.

It could be changed to:

`Expected no telemetry, but got:\n${inspect(msgs, { depth: 10, breakLength: Infinity, compact: true })}`

I believe showing the line breaks might actually be even more useful though? I am not sure if this is meant to be consumed by a machine and not a human?

watson: > It would just not potentially fail during the JSON.stringify call itself.

`JSON.stringify` is never going fail here since what `msgs` contains is something that was just decoded from JSON.

> I believe showing the line breaks might actually be even more useful though? I am not sure if this is meant to be consumed by a machine and not a human?

Probably just by a human - it's just sometimes easier to get an overview over some data if a lot of entries are printed one entry per row - it's a poor mans table output.

> The reason why it is currently serialized is likely that the default console.error inspection depth is not deep enough and about potentially having it on a single line. Otherwise, that would not even be necessary.

I don't think it's a matter of depth as these `points` objects are only one level deep. Instead I think it's matter of getting an easy overview as I wrote above regarding it being a poor mans table output.

Regarding printing the entire `msgs` object vs. just `msg[1].points`: I think everything outside the `points` property is irrelevant and will just add noise if logged. So lets keep it to `msg[1].points`.

Here's an example of the current output:

```
      AssertionError [ERR_ASSERTION]: Expected no telemetry, but got:
[{"name":"library_entrypoint.complete","tags":["injection_forced:false"]}]
```

Same, but if using `inspect`:

```
      AssertionError [ERR_ASSERTION]: Expected no telemetry, but got:
[ { name: 'library_entrypoint.complete', tags: [ 'injection_forced:false' ] } ]
```

tlhunter (dismissed):

package.json
```diff
@@ -27,6 +27,8 @@
     "test:eslint-rules": "node eslint-rules/*.test.mjs",
     "test:trace:core": "tap packages/dd-trace/test/*.spec.js \"packages/dd-trace/test/{ci-visibility,datastreams,encode,exporters,opentelemetry,opentracing,plugins,service-naming,standalone,telemetry}/**/*.spec.js\"",
     "test:trace:core:ci": "npm run test:trace:core -- --coverage --nyc-arg=--include=\"packages/dd-trace/src/**/*.js\"",
+    "test:trace:guardrails": "mocha -r \"packages/dd-trace/test/setup/mocha.js\" \"packages/dd-trace/test/guardrails/**/*.spec.js\"",
```
rochdev, line 30: The folder will also need to be added to the exclusions for core tests, otherwise they will run there too.

watson: Where is that specified?

rochdev: Actually it looks like we switched to a whitelist, so never mind.

PR #6030, opened 2025-07-04
ESLint: Scope certain files to other Node.js versions than v18.0.0

watson:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/guardrails/index.js
+++ b/packages/dd-trace/src/guardrails/index.js
@@ -1,7 +1,5 @@
 'use strict'
 
-/* eslint-disable no-var */
-
 var path = require('path')
 var Module = require('module')
 var isTrue = require('./util').isTrue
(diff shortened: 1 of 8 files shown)


packages/dd-trace/src/guardrails/index.js
```diff
@@ -26,9 +24,11 @@ function guard (fn) {
     var resolvedInApp
     var entrypoint = process.argv[1]
     try {
+      // eslint-disable-next-line n/no-unsupported-features/node-builtins
       resolvedInApp = Module.createRequire(entrypoint).resolve('dd-trace')
     } catch (e) {
       // Ignore. If we can't resolve the module, we assume it's not in the app.
+      // TODO: There's also the possibility that this version of Node.js doesn't have Module.createRequire (pre v12.2.0)
```
BridgeAR, line 31: It is fine if that is not supported due to the catch handling that. The important thing is, that it does not error for the user when being called.

watson: It errors with the incorrect assumption though. I was just thinking that maybe at some point we could find a way to continue anyway

BridgeAR: Ah, I misunderstood the comment earlier. Makes sense

PR #6032, opened 2025-07-04
telemetry: fix ssi instrumentation source

BridgeAR:
Init may be called by users as well in case they don't want to programmatically configure the SDK. In that case, SSI would have been reported, while that is obviously incorrect.
The fix now checks for the environment variable used by the agent.

### DIFF
--- a/packages/dd-trace/src/config.js
+++ b/packages/dd-trace/src/config.js
@@ -16,7 +16,7 @@ const { updateConfig } = require('./telemetry')
 const telemetryMetrics = require('./telemetry/metrics')
 const { isInServerlessEnvironment, getIsGCPFunction, getIsAzureFunction } = require('./serverless')
 const {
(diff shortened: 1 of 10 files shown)


packages/dd-trace/src/config.js
```diff
@@ -1004,6 +1004,9 @@ class Config {
     // 0: disabled, 1: logging, 2: garbage collection + logging
     env.spanLeakDebug = maybeInt(DD_TRACE_SPAN_LEAK_DEBUG)
     this._setBoolean(env, 'spanRemoveIntegrationFromService', DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED)
+    if (getEnvironmentVariable('DD_INSTRUMENTATION_INSTALL_TYPE')?.endsWith('single_step')) {
+      this._setString(env, 'instrumentationSource', 'ssi')
+    }
```
watson, line 1009: Please move this up so it's sorted according the config property name (`instrumentationSource`). Also, consider just setting the string directly instead of using the `_setString` helper.

PR #6034, opened 2025-07-04
ci: exercise more of the guardrails code during testing

watson:
### What does this PR do?

This enables logging when running tests for the guardrail code, which in turn also exercise the guardrail logger which would otherwise just be a noop.

### Motivation

The guardrail code is very poorly tested currently.


rochdev (approved):

rochdev:
Worth noting: right now the only guarantee this test aims to provide is that we don't crash. If anything else is not working it's currently out of scope.

PR #6041, opened 2025-07-07
chore: update lru-cache, less lookups, dependabot ignore update

BridgeAR:
This updates the lru-cache to the latest version that supports Node.js 18. The version above is ignored in dependabot. It also ignores path-to-regexp, since that has to stay aligned with the one used in express.js v4.
Express itself is updated as dev dependencies. That way we should
notice earlier in case more things break.
The cache lookups got a bit optimized to not have to check multiple times if an entry exists or not.


package.json
```diff
@@ -138,7 +138,7 @@
     "eslint-plugin-n": "^17.20.0",
     "eslint-plugin-promise": "^7.2.1",
     "eslint-plugin-unicorn": "^59.0.1",
-    "express": "^4.21.2",
+    "express": "^5.1.0",
```
CarlesDD, line 141: In the description of the PR it says `express` is removed from dev dependencies, but it looks like it just got an update. is this intentional?

BridgeAR: I updated the description. I first thought we'd install it as other dependencies, while we actually use it for our test setup itself, so we do need that. Thus, I went ahead and updated it to v5.

.github/workflows/platform.yml
```diff
@@ -296,6 +296,8 @@ jobs:
           version: ${{ matrix.version }}
       - uses: ./.github/actions/install
       - run: yarn add --ignore-scripts mocha@10 # Use older mocha to support old Node.js versions
+      - run: yarn remove express # Remove current express to prevent conflicts
```
BridgeAR: Note to myself: this is probably not needed

docker-compose.yml
simon-id, line 1: what's with this file ? i don't even know what it does ? VCR ? cassettes ? hello ? 2003 ?

IlyasShabi (approved):
LGTM

docker-compose.yml
BridgeAR: It is used by llmobs. I can't tell much about that. I just stumbled upon it and outcommented it as drive-by fix, since my docker containers were failing otherwise.

uurien: I use this file to be able to run plugin tests locally. For example, if I want to run mysql related system tests, I need a mysql installed with the same port/user data that is used in the CI.  To run mysql system tests, I first run `docker compose up mysql` and after that, I run the job plugin tests ci job with the mysql env var.

watson: It seems wrong to just comment it out 🤔 This was added by @sabrenner two weeks ago in PR #5951

sabrenner: yeah we added this for this feature i shared [here](https://dd.slack.com/archives/C025TJ4RZ8X/p1750859740645849) on slack for testing purposes - i think it's fine to have this commented out if it was causing issues (assuming you were running a blanket `docker compose up` and not specifying the container/service, my guess is it would cause issues). for those who want to use this test agent container for local development with the response recorder feature vs the other one might end up here to find it anyways. 

i don't much mind - i think our contributing docs specify to only `docker compose` the service/container you need, which would get around this i think, but ultimately i'm fine either way 😄

BridgeAR: Yes, I started all since I prefer to just have everything up, since I often switch test suites.
I believe keeping this commented is fine.

BridgeAR: I didn't hear anyone blocking, I guess the PR may therefore land as is?

PR #6042, opened 2025-07-07
datastreams: safer schema serialization

BridgeAR:
The original implementation contained unused code and used replace calls after serializing an object to JSON. The strings could have contained parts that the replace call would have matched (while unlikely due to the content). This now directly serializes to JSON and does not have that issue anymore. That should also be faster overall.

### DIFF
--- a/packages/dd-trace/src/datastreams/schemas/schema_builder.js
+++ b/packages/dd-trace/src/datastreams/schemas/schema_builder.js
@@ -13,13 +13,14 @@ class SchemaBuilder {
     this.properties = 0
   }
 
+  // TODO: This is only used in tests. Let's refactor the code and stop exposing the cache.
+  // It might be sufficient to
   static getCache () {
     return CACHE
   }
 
   static getSchemaDefinition (schema) {
-    const noNones = convertToJsonCompatible(schema)
-    const definition = jsonStringify(noNones)
+    const definition = toJSON(schema)
     const id = fnv64(Buffer.from(definition, 'utf8')).toString()
     return new Schema(definition, id)
   }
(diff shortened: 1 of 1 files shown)


tlhunter (dismissed):

packages/dd-trace/src/datastreams/schemas/schema_builder.js
```diff
@@ -92,42 +93,46 @@ class OpenApiComponents {
...
+    return result + '}'
   }
-  return obj
-}
-
-function convertKey (key) {
-  if (key === 'enumValues') {
-    return 'enum'
+  if (typeof value === 'function') {
+    return 'null'
```
wconti27: AVSC and Protobuf schemas do not support functions. This line is probably unnecessary, but then again doesn't really make a difference either way.

wconti27 (dismissed):
Just left a nit, the PR looks good to me, thank you for taking this on!

packages/dd-trace/src/datastreams/schemas/schema_builder.js
watson: If we're sure that value will never be `function`, I think this should be removed. Otherwise it's just going to confuse people in the future.

packages/dd-trace/src/datastreams/schemas/schema_builder.js
```diff
@@ -92,42 +93,46 @@ class OpenApiComponents {
...
+        }
+        result += value[i] == null ? 'null' : toJSON(value[i])
+      }
+      return result + ']'
+    }
+    let result = '{'
+    for (const [key, objectValue] of Object.entries(value)) {
+      if (objectValue != null && typeof key === 'string') {
+        const converted = toJSON(objectValue)
+        if (converted !== undefined) {
```
watson, line 118: I can't see in what scenario `toJSON` will return `undefined`?

watson (commented):
Why bother with the manual JSON stringification in the `toJSON` function and not just go directly to `JSON.stringify`?

packages/dd-trace/src/datastreams/schemas/schema_builder.js
BridgeAR: JSON.stringify(undefined) => undefined

packages/dd-trace/src/datastreams/schemas/schema_builder.js
```diff
@@ -92,42 +92,43 @@ class OpenApiComponents {
...
+function toJSON (value) {
+  // eslint-disable-next-line eslint-rules/eslint-safe-typeof-object
+  if (typeof value === 'object') {
+    if (value === null) {
+      return 'null'
+    }
+    if (Array.isArray(value)) {
+      let result = '['
+      for (let i = 0; i < value.length; i++) {
+        if (i > 0) {
```
watson, line 107: Nit:

```suggestion
        if (i !== 0) {
```

watson (approved):
If you think my nit doesn't change performance, feel free to merge without it (I haven't actually verified what bytecode this gets compiled to)

PR #6050, opened 2025-07-07
use dynamic queue name for aws sqs tests

rochdev:
### What does this PR do?

Use dynamic queue name for AWS SQS tests.

### Motivation

We currently have a lot of flaky tests for SQS, for example:

https://github.com/DataDog/dd-trace-js/actions/runs/16112243292/job/45458174668
https://github.com/DataDog/dd-trace-js/actions/runs/16034849399/job/45244027750

I tried reproducing these locally and wasn't able to. My theory is that there are side-effects between tests which makes them flaky. This has happened with several other message queues in the past, and usually, isolating individual queues/topics has worked to solve the problem, so I am attempting the same here as well. If this works well, I will open additional PRs for the other AWS products.

### DIFF
--- a/packages/datadog-plugin-aws-sdk/test/sqs.spec.js
+++ b/packages/datadog-plugin-aws-sdk/test/sqs.spec.js
@@ -419,7 +419,6 @@ describe('Plugin', () => {
         before(async () => {
           return agent.load('aws-sdk', {
             sqs: {
-              consumer: false,
               dsmEnabled: true
             }
           },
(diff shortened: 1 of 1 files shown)


packages/datadog-plugin-aws-sdk/test/sqs.spec.js
```diff
@@ -500,7 +529,10 @@ describe('Plugin', () => {
         })
 
         if (sqsClientName === 'aws-sdk' && semver.intersects(version, '>=2.3')) {
-          it('Should set pathway hash tag on a span when consuming and promise() was used over a callback',
+          // This test was always failing on its own but for some reason it
+          // passes only because of side-effects from other tests.
+          // TODO: Fix the test to work properly without side-effects.
+          it.skip('Should set pathway hash tag on a span when consuming and promise() was used over a callback',
```
BridgeAR: I think we should fix this before landing

packages/datadog-plugin-aws-sdk/test/sqs.spec.js
```diff
@@ -405,33 +426,41 @@ describe('Plugin', () => {
...
+            'tester',
+            ['direction:in', 'topic:' + queueNameDSM, 'type:sqs'],
+            producerHash
+          ).readBigUInt64BE(0).toString()
+        })
 
-            done()
-          })
+        beforeEach(done => {
+          sqs.createQueue(queueOptionsDsm, (err, res) => err ? done(err) : done())
```
BridgeAR, line 451: Since we touch these anyway: do all versions support promises? If yes, could we change the code to use that instead?

packages/datadog-plugin-aws-sdk/test/sqs.spec.js
rochdev: It's out of scope of this PR since it never worked.

packages/datadog-plugin-aws-sdk/test/sqs.spec.js
rochdev: We are explicitly testing support for the callback API and the promise API separately.

packages/datadog-plugin-aws-sdk/test/sqs.spec.js
rochdev: @wconti27 Any idea why this doesn't work when run standalone? Reading the test it doesn't seem that it could work since it awaits things in the wrong order, but it ends up passing when previous tests are run (presumably because of side-effects of existing in-flight messages). I tried to fix the order, but then the consumer span is never created.

rochdev: @BridgeAR Is this a blocker though? This test never worked and only passed because of previous tests. I'd rather at least have it explicitly skipped with a TODO instead of having a hidden false positive. There is no regression since it never tested what it was meant to test. We can look at it in a separate effort.

packages/datadog-plugin-aws-sdk/test/sqs.spec.js
```diff
@@ -502,28 +530,23 @@ describe('Plugin', () => {
...
-                  }
+                expect(consumeSpanMeta).to.include({
+                  'pathway.hash': expectedConsumerHash
                 })
               })
+
+              await sqs.sendMessage({ MessageBody: 'test DSM', QueueUrl: QueueUrlDsm }).promise()
+              await sqs.receiveMessage({ QueueUrl: QueueUrlDsm }).promise()
+
+              return tracePromise
```
BridgeAR, line 549: Nit: This could cause an unhandled rejection in case the tracePromise is rejected before the receiveMessage ended. It could even fail with sendMessage, while that's unlikely and it would otherwise be more difficult to write the test nicely.

```suggestion
              await Promise.all([
                sqs.receiveMessage({ QueueUrl: QueueUrlDsm }).promise(),
                tracePromise,
              ])
```

PR #6054, opened 2025-07-08
remove get-port usage from inferred proxy test

rochdev:
### What does this PR do?

Remove get-port usage from inferred proxy test.

### Motivation

This was causing flakiness such as https://github.com/DataDog/dd-trace-js/actions/runs/16117264911/job/45473955496

### DIFF
--- a/packages/dd-trace/test/plugins/util/inferred_proxy.spec.js
+++ b/packages/dd-trace/test/plugins/util/inferred_proxy.spec.js
@@ -3,7 +3,6 @@
 require('../../setup/tap')
 
 const agent = require('../agent')
-const getPort = require('get-port')
 const { expect } = require('chai')
 const axios = require('axios')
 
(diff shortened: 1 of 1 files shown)


packages/dd-trace/test/plugins/util/inferred_proxy.spec.js
```diff
@@ -37,14 +35,19 @@ describe('Inferred Proxy Spans', function () {
...
+      appListener = server.listen(0, '127.0.0.1', () => {
+        port = server.address().port
+        resolve()
+      })
+    })
   }
 
   // test cleanup function
-  const cleanupTest = function () {
+  const cleanupTest = async function () {
```
BridgeAR, line 47: I just had a look at the tests themselves. If I am not mistaken, test failures would be silenced due to this cleanup being called in a catch handler. So instead, I think this should be a afterEach hook.

rochdev: Done!

PR #6055, opened 2025-07-08
[test-optimization] Add logs for Playwright test

Mariovido:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


integration-tests/automatic-log-submission.spec.js
```diff
@@ -152,9 +153,17 @@ describe('test visibility automatic log submission', () => {
         })
 
         childProcess.stdout.on('data', (chunk) => {
+          if (name === 'playwright') {
+            // eslint-disable-next-line no-console
+            console.log(chunk.toString())
+          }
```
BridgeAR: The chunks are combined and logged on process exit. Should these really be logged here on top? The same with the spot below.

Mariovido: Yeah it can be done in that way, but as this is something temporary until we fix the flakiness it can be done in this way.

PR #6061, opened 2025-07-08
instrumentation: add hono support

BridgeAR:
This addresses issues that came up with esm and newer hono
versions.
While working on this, a bug surfaced about the resource name.
Either this is an issue in our test setup or we have a major issue
with our esm instrumentation.

Closes #4812

This ports #4812 with a couple of things changed. Newer hono versions use a private dispatcher, so that could not be used. Instead, we instrument fetch.

### DIFF
--- a/packages/datadog-plugin-hono/test/integration-test/client.spec.js
+++ b/packages/datadog-plugin-hono/test/integration-test/client.spec.js
@@ -36,7 +36,7 @@ describe('esm', () => {
 
     it('is instrumented', async () => {
       proc = await spawnPluginIntegrationTestProc(sandbox.folder, 'server.mjs', agent.port)
-      proc.url += 'hello'
+      proc.url += '/hello'
 
       return curlAndAssertMessage(agent, proc, ({ headers, payload }) => {
         assertObjectContains(headers, { host: `127.0.0.1:${agent.port}` })


packages/datadog-instrumentations/src/hono.js
```diff
@@ -0,0 +1,102 @@
...
+function wrapCompose (compose) {
+  return function (middleware, onError, onNotFound) {
+    const instrumentedOnError = (...args) => {
+      const [error, context] = args
+      const req = context.env.incoming
+      errorChannel.publish({ req, error })
+      return onError(...args)
+    }
+
+    const instrumentedMiddlewares = middleware.map(h => {
```
uurien, line 29: I am probably missing something, but don't we need instrumentation for `apm:hono:middleware:enter/next/finish/error`?
which are used here: https://github.com/DataDog/dd-trace-js/blob/bb7d26a48e919d5834be852a5ad74ba1eeb23add/packages/datadog-plugin-router/src/index.js#L20

BridgeAR: I only ported the PR and did not improve the implementation besides making it work with newer versions, fully supporting ESM and by rewriting the tests to use async await and newer test helpers and adding the CI job and a few minor changes.

The question is: do we want to improve this before landing or afterwards.

rochdev: Middleware support is not mandatory, it's only supported by the base class if needed.

PR #6067, opened 2025-07-09
[SVLS-7133] Instrument Azure Service Bus

jcstorms1:
### What does this PR do?

Adds a plugin for the [Azure Service Bus package](https://www.npmjs.com/package/@azure/service-bus)
- Instruments the `sendMessages` method
- Injects context into the message for distributed tracing

Adds instrumentation for the Service Bus Queue trigger in the Azure Functions Plugin
- Extracts context from the message for proper parenting

image

### Motivation

Customers expect full distributed tracing for Azure Functions. Unfortunately [Azure function's output bindings](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-service-bus-output?tabs=python-v2%2Cisolated-process%2Cnodejs-v4%2Cextensionv5&pivots=programming-language-javascript) did not allow us to inject context into the message. This led to the addition of the Azure Service Bus library.
### Plugin Checklist

- [ ] Unit tests.
- [x] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes

Adds azureservicebusemulator and azuresqledge services to run Azure Service Bus locally and in CI for tests
- https://learn.microsoft.com/en-us/azure/service-bus-messaging/test-locally-with-service-bus-emulator?tabs=docker-linux-container


packages/datadog-instrumentations/src/azure-functions.js
```diff
@@ -11,13 +11,17 @@ const azureFunctionsChannel = dc.tracingChannel('datadog:azure:functions:invoke'
...
+  // Http triggers
   shimmer.wrap(app, 'deleteRequest', wrapHandler)
   shimmer.wrap(app, 'http', wrapHandler)
   shimmer.wrap(app, 'get', wrapHandler)
   shimmer.wrap(app, 'patch', wrapHandler)
   shimmer.wrap(app, 'post', wrapHandler)
   shimmer.wrap(app, 'put', wrapHandler)
 
+  // Service Bus triggers
+  shimmer.wrap(app, 'serviceBusQueue', wrapHandler)
```
duncanpharvey, line 23: What about `app.serviceBusTopic`?

packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -35,11 +35,19 @@ class AzureFunctionsPlugin extends TracingPlugin {
         'aas.function.trigger': mapTriggerTag(methodName)
       }
     }, false)
+    if (methodName === 'serviceBusQueue') {
+      const triggerEntity = invocationContext.options.trigger.queueName || invocationContext.options.trigger.topicName
+      span.setTag('messaging.message_id', invocationContext.triggerMetadata.messageId)
+      span.setTag('messaging.operation', 'receive')
+      span.setTag('messaging.system', 'servicebus')
+      span.setTag('messaging.destination.name', triggerEntity)
+      span.setTag('resource.name', 'serviceBusQueueTrigger')
```
duncanpharvey: Resource should be set dynamically to the queue or topic name

packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -35,11 +35,19 @@ class AzureFunctionsPlugin extends TracingPlugin {
         'aas.function.trigger': mapTriggerTag(methodName)
       }
     }, false)
+    if (methodName === 'serviceBusQueue') {
+      const triggerEntity = invocationContext.options.trigger.queueName || invocationContext.options.trigger.topicName
+      span.setTag('messaging.message_id', invocationContext.triggerMetadata.messageId)
+      span.setTag('messaging.operation', 'receive')
+      span.setTag('messaging.system', 'servicebus')
+      span.setTag('messaging.destination.name', triggerEntity)
```
duncanpharvey: You can import the `resource.name` tag key from `ext/tags` https://github.com/DataDog/dd-trace-js/blob/47d3c6312a9f07404fa2aaeb852d4e26c445eb14/ext/tags.js#L6

Also since these `messaging.*` tags are standard tags that could be potentially used for other integrations maybe they should be added to `ext/tags` as well?

packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -35,11 +35,19 @@ class AzureFunctionsPlugin extends TracingPlugin {
...
+      span.setTag('resource.name', 'serviceBusQueueTrigger')
+      span.setTag('span.kind', 'consumer')
```
duncanpharvey: `span.kind` can be imported from `ext/tags` as well
https://github.com/DataDog/dd-trace-js/blob/47d3c6312a9f07404fa2aaeb852d4e26c445eb14/ext/tags.js#L8

packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -11,7 +11,8 @@ const triggerMap = {
   get: 'Http',
   patch: 'Http',
   post: 'Http',
-  put: 'Http'
+  put: 'Http',
+  serviceBusQueue: 'serviceBusQueue'
```
duncanpharvey: This should be `serviceBusQueue: 'ServiceBus'` to follow the pattern in the .NET and Python Tracers

- https://github.com/DataDog/dd-trace-dotnet/blob/49c43b5bc1ca03a6fc5da0ce5077c308ac4cca06/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs#L93
- https://github.com/DataDog/dd-trace-py/blob/c91dd1611c31230cdd7c3a59120caf4c27bdc3de/ddtrace/contrib/internal/azure_functions/patch.py#L95

packages/datadog-plugin-azure-service-bus/src/producer.js
```diff
@@ -0,0 +1,36 @@
...
+    const qualifiedSenderNamespace = sender._sender.audience.replace('sb://', '')
+    const span = this.startSpan({
+      resource: sender.entityPath,
+      type: 'messaging',
+      meta: {
+        component: 'azure-service-bus',
+        'messaging.destination.name': sender.entityPath,
+        'messaging.operation': 'send',
+        'messaging.system': 'servicebus',
+        'network.destination.name': qualifiedSenderNamespace,
```
duncanpharvey, line 20: Consider adding these tag keys to `ext/tags`

packages/datadog-instrumentations/src/azure-service-bus.js
```diff
@@ -0,0 +1,38 @@
...
+
+const producerStartCh = channel('apm:azure-service-bus:send:start')
+const producerErrorCh = channel('apm:azure-service-bus:send:error')
+const producerFinishCh = channel('apm:azure-service-bus:send:finish')
+
+addHook({ name: '@azure/service-bus', versions: ['>=7.9.2'] }, (obj) => {
+  const ServiceBusClient = obj.ServiceBusClient
+  shimmer.wrap(ServiceBusClient.prototype, 'createSender', createSender => function (queueName) {
+    const sender = createSender.apply(this, arguments)
+    shimmer.wrap(sender._sender, 'send', send => function (msg) {
```
duncanpharvey, line 18: When patching [MessageSender.send](https://github.com/Azure/azure-sdk-for-js/blob/e8000a2609939960c43277a6d4ca3e5bce3e8d0d/sdk/servicebus/service-bus/src/core/messageSender.ts#L330) are spans created in the following scenarios?

- Synchronous message send (no await)
- `ServiceBusSender.sendMessages` with a list of messages instead of a single message
- `ServiceBusSender.scheduleMessages`

packages/datadog-instrumentations/src/azure-service-bus.js
```diff
@@ -0,0 +1,38 @@
...
+
+const shimmer = require('../../datadog-shimmer')
+
+const producerStartCh = channel('apm:azure-service-bus:send:start')
+const producerErrorCh = channel('apm:azure-service-bus:send:error')
+const producerFinishCh = channel('apm:azure-service-bus:send:finish')
+
+addHook({ name: '@azure/service-bus', versions: ['>=7.9.2'] }, (obj) => {
+  const ServiceBusClient = obj.ServiceBusClient
+  shimmer.wrap(ServiceBusClient.prototype, 'createSender', createSender => function (queueName) {
```
duncanpharvey: Either a queue or topic can be passed to `createSender`

packages/datadog-plugin-azure-functions/src/index.js
jcstorms1: Why would the azure function trigger be set to the queue or topic name?

duncanpharvey: Ah you're right, I mixed up the producer and the consumer. Since this is the consumer the resource name should be in the form `<trigger type> <function name>`. Trigger type in this case would be `ServiceBus`.

packages/datadog-plugin-azure-service-bus/src/producer.js
jcstorms1: That file doesn't have any other integration specific tags in it and I followed the other messaging plugins as examples. Maybe someone from APM will have an opinion on this.

packages/datadog-plugin-azure-functions/src/index.js
jcstorms1: I commented about this below.

packages/datadog-plugin-azure-functions/src/index.js
jcstorms1: Same thing as your other comments. Other plugins are not using this file. Not sure why though.

packages/datadog-instrumentations/src/azure-service-bus.js
jcstorms1: - Azure service bus does not support synchronous message sending
 - A list did not work, because it was converted into a batch class and encoded 
 ```        if (isServiceBusMessageBatch(messages)) {
            // Case 2: Batch message
            batch = messages;
        }
        else {
            // Case 3: Array of messages
            batch = await this.createMessageBatch(options);
            for (const message of messages) {
                throwIfNotValidServiceBusMessage(message, errorInvalidMessageTypeSingleOrArray);
                if (!batch.tryAddMessage(message, options)) {
```
 - Did not try it, but based on the code I doubt it will work.

Comment (jcstorms1) [reply packages/datadog-instrumentations/src/azure-service-bus.js:18]: Batching and scheduling is non-trivial. We should revisit this later. I'll make cards for us to track them.

File: packages/datadog-instrumentations/src/azure-service-bus.js [hunk 9]
```diff
@@ -0,0 +1,38 @@
...
+const producerErrorCh = channel('apm:azure-service-bus:send:error')
+const producerFinishCh = channel('apm:azure-service-bus:send:finish')
+
+addHook({ name: '@azure/service-bus', versions: ['>=7.9.2'] }, (obj) => {
+  const ServiceBusClient = obj.ServiceBusClient
+  shimmer.wrap(ServiceBusClient.prototype, 'createSender', createSender => function (queueOrTopicName) {
+    const sender = createSender.apply(this, arguments)
+    shimmer.wrap(sender._sender, 'send', send => function (msg) {
+      const ctx = { sender, msg }
+      return producerStartCh.runStores(ctx, () => {
```
rochdev, line 20: Not a blocker, but for a standard promise instrumentation, it would have been possible to use `tracingChannel.tracePromise()` to replace this whole block.

packages/datadog-plugin-azure-functions/src/index.js
rochdev: I personally prefer strings, especially for tests where I want the expectation to be as obvious as possible without having to search for the underlying value.

packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -35,11 +36,19 @@ class AzureFunctionsPlugin extends TracingPlugin {
         'aas.function.trigger': mapTriggerTag(methodName)
       }
     }, false)
+    if (triggerMap[methodName] === 'ServiceBus') {
+      const triggerEntity = invocationContext.options.trigger.queueName || invocationContext.options.trigger.topicName
+      span.setTag('messaging.message_id', invocationContext.triggerMetadata.messageId)
```
rochdev: It would be more efficient to prepare the `meta` object before `startSpan` depending on the trigger, and then pass that directly.

For example:

```js
const meta = getMetaForTrigger(ctx)

const span = this.startSpan(this.operationName(), {
  childOf,
  service: this.serviceName(),
  type: 'serverless',
  meta
}, false)

function getMetaForTrigger({ functionName, methodName, invocationContext }) {
  const meta = {
    'aas.function.name': functionName,
    'aas.function.trigger': mapTriggerTag(methodName)
  }

  if (triggerMap[methodName] === 'ServiceBus') {
    ...
  }

  return meta
}
```

packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -35,11 +36,19 @@ class AzureFunctionsPlugin extends TracingPlugin {
         'aas.function.trigger': mapTriggerTag(methodName)
       }
     }, false)
```
rochdev: Unrelated to this PR, but passing `ctx` here instead of false would remove the need to explicitly assign `currentStore` and `parentStore` to the context at the end of the function.

packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -75,4 +89,13 @@ function mapTriggerTag (methodName) {
   return triggerMap[methodName] || 'Unknown'
 }
 
+function extractTraceContext (tracer, ctx) {
+  switch (String(triggerMap[ctx.methodName])) {
+    case 'Http':
```
rochdev, line 100: Out of scope of this PR but in the future I think it might be worth it to split things as multiple plugins instead of having an ever-growing switch.

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
```diff
@@ -20,8 +20,9 @@ describe('esm', () => {
   withVersions('azure-functions', '@azure/functions', NODE_MAJOR < 20 ? '<4.7.3' : '*', version => {
     before(async function () {
       this.timeout(50000)
-      sandbox = await createSandbox([`@azure/functions@${version}`, 'azure-functions-core-tools@4'], false,
-        ['./packages/datadog-plugin-azure-functions/test/integration-test/fixtures/*'])
+      sandbox = await createSandbox([
+        `@azure/functions@${version}`, 'azure-functions-core-tools@4', '@azure/service-bus@7.9.2'], false,
+      ['./packages/datadog-plugin-azure-functions/test/integration-test/fixtures/*'])
```
rochdev:
```suggestion
      sandbox = await createSandbox([
        `@azure/functions@${version}`,
        'azure-functions-core-tools@4',
        '@azure/service-bus@7.9.2'
      ],
      false,
      ['./packages/datadog-plugin-azure-functions/test/integration-test/fixtures/*'])
```

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
```diff
@@ -54,15 +55,51 @@ describe('esm', () => {
...
         PATH: `${sandbox.folder}/node_modules/azure-functions-core-tools/bin:${process.env.PATH}`
       }
       proc = await spawnPluginIntegrationTestProc(sandbox.folder, 'func', ['start'], agent.port, undefined, envArgs)
 
       return curlAndAssertMessage(agent, 'http://127.0.0.1:7071/api/httptest2', ({ headers, payload }) => {
         assert.strictEqual(payload.length, 2)
-        assert.strictEqual(payload[0][0].parent_id, payload[1][1].span_id)
+        assert.strictEqual(payload[1][0].span_id, payload[1][1].parent_id)
+      })
+    }).timeout(50000)
```
rochdev, line 72: Why 50 seconds? This seems very specific.

rochdev (commented):
Added a few comments but no dealbreakers, just a few things that could be improved.

One thing that I noticed also is that these plugins only have integration tests. These are usually only used to test ESM support right now. Was that done because only ESM is supported? Otherwise the tests could be moved to regular plugin tests which would run a lot faster.

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
jcstorms1: I asked @duncanpharvey about this before. He said it takes a while to start up the azure function runtime and 50 seconds gave enough cushion for the tests to pass reliably.

rochdev: Ok, I guess I was just curious why 50 seconds and not a minute at that point 😄

duncanpharvey: The 50 second timeout was arbitrary

PR #6072, opened 2025-07-10
support iast with fastify

IlyasShabi:
### What does this PR do?

This PR aims to fully support IAST with Fastify

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/datadog-instrumentations/src/fastify.js
```diff
@@ -275,3 +276,23 @@ addHook({ name: 'fastify', versions: ['2'] }, fastify => {
...
+  })
+
+  return Reply
+}
+
+addHook({ name: 'fastify', file: 'lib/reply.js', versions: ['1'] }, wrapReplyHeader)
+
+addHook({ name: 'fastify', file: 'lib/reply.js', versions: ['2'] }, wrapReplyHeader)
+
+addHook({ name: 'fastify', file: 'lib/reply.js', versions: ['>=3'] }, wrapReplyHeader)
```
CarlesDD: I think we can unify these three `addHook`s  since they all are instrumenting the same file with the same hook

IlyasShabi: I wanted to do the same for `wrapFastify` but it's okay also to use `>=1`. wydt ?

packages/dd-trace/src/appsec/iast/index.js
```diff
@@ -89,7 +92,13 @@ function onIncomingHttpRequestEnd (data) {
...
     const iastContext = iastContextFunctions.getIastContext(store, topContext)
     if (iastContext?.rootSpan) {
-      iastResponseEnd.publish(data)
+      const storedHeaders = collectedResponseHeaders.get(data.res) || {}
+
+      iastResponseEnd.publish({ ...data, storedHeaders })
+
+      if (storedHeaders) {
+        collectedResponseHeaders.delete(data.res)
+      }
```
CarlesDD: This conditional is always evaluated as `true`

packages/dd-trace/src/appsec/iast/index.js
```diff
@@ -105,4 +114,13 @@ function onIncomingHttpRequestEnd (data) {
   }
 }
 
+// Response headers are collected here because they are not available in the onIncomingHttpRequestEnd when using Fastify
+function onResponseWriteHeadCollect ({ res, responseHeaders = {} }) {
```
CarlesDD, line 118: Do `responseHeaders` keys come always in lower case?

When [retrieving headers in analyzers](https://github.com/DataDog/dd-trace-js/pull/6072/files#diff-7898738aed765e3d9dafa59774bec49e7fbd2052ea0714d2470857849136df54R32) the code searches by header name in lower case. If the header name from fastify does not come in lower case, they must be lower cased before setting them in the weak map.

CarlesDD: I've checked they come in lower case. I'm resolving this conversation.

packages/datadog-instrumentations/src/fastify.js
simon-id: if you use >=1, the intermediary major versions (2,3) won't be tested. if you want to be completely equivalent you need to do:

```suggestion
addHook({ name: 'fastify', file: 'lib/reply.js', versions: ['1', '2', '>=3'] }, wrapReplyHeader)
```

watson: The fact that the way these ranges are specified in the `versions` option impacts what is being tested has always bothered me. I've wanted to change that for a long time, maybe some day 😅

rochdev: It's already possible to use `>=1` and then define the additional test versions in `externals.json`.

packages/dd-trace/src/appsec/iast/index.js
IlyasShabi: Done

packages/dd-trace/src/appsec/iast/index.js
```diff
@@ -18,11 +18,12 @@ const { IAST_ENABLED_TAG_KEY } = require('./tags')
 const iastTelemetry = require('./telemetry')
 const { enable: enableFsPlugin, disable: disableFsPlugin, IAST_MODULE } = require('../rasp/fs-plugin')
 const securityControls = require('./security-controls')
+const { incomingHttpRequestStart, incomingHttpRequestEnd, responseWriteHead } = require('../channels')
+
+const collectedResponseHeaders = new WeakMap()
 
 // TODO Change to `apm:http:server:request:[start|close]` when the subscription
 //  order of the callbacks can be enforce
```
CarlesDD, line 26: I think this TODO comment should be moved close to the new required channels to keep its meaning.

uurien (approved):

PR #6073, opened 2025-07-10
Fix appsec tests

uurien:
### What does this PR do?

Fixes appsec tests in https://github.com/DataDog/dd-trace-js/pull/5960
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/datadog-instrumentations/src/child_process.js
```diff
@@ -189,7 +186,7 @@ function wrapChildProcessAsyncMethod (ChildProcess, shell = false) {
 
       const context = createContextFromChildProcessInfo(childProcessInfo)
       const abortController = new AbortController()
```
simon-id, line 188: this abortcontroller was never passed to a subscriber no ?

packages/datadog-instrumentations/test/child_process.spec.js
```diff
@@ -58,13 +58,15 @@ describe('child process', () => {
                 const childEmitter = childProcess[methodName]('ls')
 
                 childEmitter.once('close', () => {
-                  expect(start).to.have.been.calledOnceWith({
+                  expect(start).to.have.been.calledOnce
+                  expect(start).to.have.been.calledWithMatch({
```
watson, line 62: Why not this?

```suggestion
                  expect(start).to.have.been.calledOnceWithMatch({
```

uurien: I tried with that first, but it doesn't work:
```
Error: Invalid Chai property: calledOnceWithMatch
```

packages/datadog-instrumentations/src/child_process.js
uurien: > was never passed to a subscriber

This PR is not pointing to master, is pointing to `conti/remove-async-storage-child-process` branch, it is passed to a subscriber in master, but the reason why the tests are failing in @wconti27 PR is that is not passed there. Just fixing it.

simon-id (approved):
oh LGTM, but maybe ask @wconti27 before merging in his branch :+1:

packages/datadog-instrumentations/test/child_process.spec.js
watson: Hmm my best guess is that it's because we're using the chai flavored version of sinon via the `sinon-chai` package. If these tests used pure sinon, it would probably work, eg:

```js
sinon.assert.calledOnceWithMatch(start, {....})
```

uurien: Yes, `sinon.assert.calledOnceWithMatch` works, but tbh I don't want to modify whole test file right now from `expect` to `assert` right now. This is just to unblock other PR.

PR #6075, opened 2025-07-10
API Security -Add support for trace tagging rules

CarlesDD:
### What does this PR do?
This PR adds support for trace-tagging rules, an extension of existing In-App WAF rules. These rules facilitate data collection by adding relevant attributes to the current trace.

Additionally, trace-tagging rules introduce the `keep` field which can be used to indicate whether the current trace should be kept through a change in the sampling priority. Since priority sampling does not rely on events presence anymore, appsec rate limiter has been moved from the `reporter` (where was consulted everytime an attack was reported) to the `waf`, to be consulted every time a WAF result comes with `keep` field set to `true`.

### Motivation
[RFC-1034] API Security: Trace-Tagging Rules

### Plugin Checklist

- [x] Unit tests.

### Additional Notes

[APPSEC-57440](https://datadoghq.atlassian.net/browse/APPSEC-57440)

[System Tests](https://github.com/DataDog/system-tests/pull/4923)

[APPSEC-57440]: https://datadoghq.atlassian.net/browse/APPSEC-57440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dd-trace/src/appsec/reporter.js
```diff
@@ -3,7 +3,6 @@
 const dc = require('dc-polyfill')
 const zlib = require('zlib')
 
-const Limiter = require('../rate_limiter')
```
CarlesDD, line 6: Sampling priority relies now in the keep field returned by the waf and not in the event presence. Due to this, appsec rate limiter has been moved to waf.

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -444,7 +432,7 @@ function reportAttributes (attributes) {
 
   const tags = {}
   for (let [tag, value] of Object.entries(attributes)) {
-    if (!isFingerprintAttribute(tag)) {
+    if (isSchemaAttribute(tag)) {
```
CarlesDD, line 435: Conditional has been flipped in order to only apply the special format (gizp + base64 encoding) to schema attributes (aka derivatives)

CarlesDD:
CI is failing due to a bug on `native-appsec` marshalling process, fixed in [this PR](https://github.com/DataDog/dd-native-appsec-js/pull/145)

uurien (approved):

integration-tests/appsec/trace-tagging.spec.js
```diff
@@ -0,0 +1,95 @@
...
+
+        fastifyRequestReceived = true
+
+        assert.property(payload[0][0].meta, '_dd.appsec.trace.agent')
+        assert.strictEqual(payload[0][0].meta['_dd.appsec.trace.agent'], 'TraceTaggingTest/v1')
+        assert.property(payload[0][0].metrics, '_dd.appsec.trace.integer')
+        assert.strictEqual(payload[0][0].metrics['_dd.appsec.trace.integer'], 1234)
+      }, 30000, 10, true)
+
+      assert.isTrue(fastifyRequestReceived)
```
simon-id, line 95: why these two lines for fastify but not for express ?

packages/dd-trace/src/appsec/waf/index.js
```diff
@@ -99,8 +108,18 @@ function run (data, req, raspRule) {
...
+  const result = wafContext.run(data, raspRule)
+
+  if (result?.keep) {
+    if (limiter.isAllowed()) {
+      const rootSpan = web.root(req)
+      keepTrace(rootSpan, ASM)
+    } else {
+      updateRateLimitedMetric(req)
+    }
+  }
```
simon-id, line 120: i don't understand why this has to live in this file, and not be in the reporter like it was before ? the waf reports its output and state to the reporter, which then takes care of sending it to the backend

simon-id: in my opinion, this shouldn't be here. There should be code in waf_context_wrapper.js that calls a function in reporter.js that executes this code instead.

simon-id: the rate limiter and metrics used to depend on the presence of event=true. Now it depends on the presence of keep=true. That doesn't change anything to the underlying organization of the code no ? just change the field that's used in the condition no ? :shrug:

simon-id: alright we zoomed, and i got convinced enough to keep it this way :+1:

integration-tests/appsec/trace-tagging.spec.js
CarlesDD: This is because, with express, the first message that the fake agent receives is the one that corresponds to `express.request`. However, this does not happen with fastify. This is why the `expectedMessageCount` in `assertMessageReceived` is set to 10 and these additional lines are added to check that the `fastify.request `message has been received and checked. 

It is the same check that is done in this [other integration test](https://github.com/DataDog/dd-trace-js/blob/master/integration-tests/appsec/response-headers.spec.js#L62).

simon-id: sounds good, maybe worth a little comment in the code for our future selves that will forget about this ? :laughing:

packages/dd-trace/test/appsec/waf/index.spec.js
```diff
@@ -122,6 +139,100 @@ describe('WAF Manager', () => {
...
+      it('should handle keep=true with events but rate limiter denying', () => {
+        const result = { keep: true, events: ['attack_event'] }
+        mockWafContext.run.returns(result)
+        limiterStub.isAllowed.returns(false)
+
+        const payload = { persistent: { 'server.io.net.url': 'http://example.com' } }
+        waf.run(payload, req)
+
+        expect(updateRateLimitedMetric).to.have.been.calledOnceWithExactly(req)
+        expect(keepTrace).not.to.have.been.called
```
simon-id: those two tests are very similar to the ones above, the only difference is the presence of events ? is that really neccesary to have specific tests for these cases ?

simon-id:
otherwise lgtm

packages/dd-trace/test/appsec/waf/index.spec.js
CarlesDD: Since the logic for `keepTrace` is unrelated with the presence of events I think its safe to remove these cases.

simon-id (approved):

PR #6076, opened 2025-07-10
add stats and branch/date options to flakiness report

rochdev:
### What does this PR do?

Add stats and branch/date options to flakiness report.

### Motivation

The main reason we want less flakiness is to minimize the percentage of CI runs that disrupt work, so it makes sense to also log that percentage.

### DIFF
--- a/scripts/flakiness.mjs
+++ b/scripts/flakiness.mjs
@@ -113,7 +113,7 @@ await Promise.all(workflows.map(w => checkWorkflowRuns(w)))
 const dateRange = startDate === endDate ? `on ${endDate}` : `from ${startDate} to ${endDate}`
 const logString = `jobs with at least ${OCCURRENCES} occurrences seen ${dateRange} (UTC)*`
 const workflowSuccessRate = +((1 - flakeCount / totalCount) * 100).toFixed(1)
-const pipelineSuccessRate = ((workflowSuccessRate / 100) ** workflows.length * 100).toFixed(1)
+const pipelineSuccessRate = +((workflowSuccessRate / 100) ** workflows.length * 100).toFixed(1)
 const pipelineBadge = pipelineSuccessRate >= 80 ? '🟢' : pipelineSuccessRate >= 70 ? '🟡' : '🔴'
 
 if (Object.keys(flaky).length === 0) {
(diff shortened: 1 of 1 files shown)


scripts/flakiness.mjs
```diff
@@ -95,22 +109,31 @@ async function checkWorkflowJobs (id, page = 1) {
 await Promise.all(workflows.map(w => checkWorkflowRuns(w)))
 
 // TODO: Report this somewhere useful instead.
+
+const dateRange = startDate === endDate ? `on ${endDate}` : `from ${startDate} to ${endDate}`
+const logString = `jobs with at least ${OCCURRENCES} occurrences seen ${dateRange} (UTC)*`
+const workflowSuccessRate = +((1 - flakeCount / totalCount) * 100).toFixed(1)
+const pipelineSuccessRate = +((workflowSuccessRate / 100) ** workflows.length * 100).toFixed(1)
+const pipelineBadge = pipelineSuccessRate >= 80 ? '🟢' : pipelineSuccessRate >= 70 ? '🟡' : '🔴'
```
watson: Nit: I'd move these into the `else` block to better signal what they are used for. But if you put them here on purpose, feel free to ignore me

rochdev: Done!

scripts/flakiness.mjs
```diff
@@ -95,22 +109,32 @@ async function checkWorkflowJobs (id, page = 1) {
...
-    `*No flaky jobs with at least ${OCCURRENCES} occurrences seen in the last ${DAYS > 1 ? `${DAYS} days` : 'day'}*`
-  )
+  console.log(`*No flaky ${logString}`)
 } else {
-  console.log(
-    `*Flaky jobs with at least ${OCCURRENCES} occurrences seen in the last ${DAYS > 1 ? `${DAYS} days` : 'day'}*`
-  )
+  const workflowSuccessRate = +((1 - flakeCount / totalCount) * 100).toFixed(1)
+  const pipelineSuccessRate = +((workflowSuccessRate / 100) ** workflows.length * 100).toFixed(1)
+  const pipelineBadge = pipelineSuccessRate >= 80 ? '🟢' : pipelineSuccessRate >= 70 ? '🟡' : '🔴'
```
BridgeAR, line 121: These numbers still seem relatively low for providing green and yellow badges.

rochdev: I thought I had them at 85 and 75, but yeah the idea is that they are 5-10% above our current target of 75%. Since we've been hovering around 60-65% over the past several months, I think it makes sense to consider being in the 80s as green.

PR #6079, opened 2025-07-11
update confluent kafka tests to use dynamic topic

rochdev:
### What does this PR do?

Update Confluent Kafka tests to use dynamic topic.

### Motivation

Same as #6053 but for Confluent.

### Additional Notes

Instead of using `fromBeginning` everywhere, it would be better to listen to the server events and wait for things like consumer rebalancing before sending messages. However, this would take a significant refactor, almost to the point of rewriting the entire test suite. Simply using `fromBeginning` instead should be sufficient, especially since the driver should be buffering messages, this avoiding race conditions.

### DIFF
--- a/packages/datadog-plugin-confluentinc-kafka-javascript/test/index.spec.js
+++ b/packages/datadog-plugin-confluentinc-kafka-javascript/test/index.spec.js
@@ -360,9 +360,9 @@ describe('Plugin', () => {
               nativeConsumer = new Consumer({
                 'bootstrap.servers': '127.0.0.1:9092',
                 'group.id': groupId,
-                'enable.auto.commit': false, // don't commit my offset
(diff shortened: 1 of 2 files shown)


packages/datadog-plugin-confluentinc-kafka-javascript/test/index.spec.js
```diff
@@ -617,24 +635,22 @@ describe('Plugin', () => {
                     partition,
                     offset: Number(message.offset)
                   }
-                  // Signal that we've processed a message
                   messageProcessedResolve()
                 }
               })
 
-              consumerRunPromise.catch(() => {})
```
BridgeAR, line 625: ❤️

BridgeAR (approved):
LGTM

PR #6081, opened 2025-07-11
push repro

simon-id:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/appsec/rasp/index.js
+++ b/packages/dd-trace/src/appsec/rasp/index.js
@@ -104,6 +104,7 @@ function enable (config) {
 
   process.on('uncaughtExceptionMonitor', handleUncaughtExceptionMonitor)
   expressMiddlewareError.subscribe(blockOnDatadogRaspAbortError)
+  require('dc-polyfill').channel('apm:fastify:middleware:error').subscribe(blockOnDatadogRaspAbortError)
 }
 
 function disable () {
--- a/packages/dd-trace/src/appsec/rasp/utils.js
+++ b/packages/dd-trace/src/appsec/rasp/utils.js
@@ -51,10 +51,13 @@ function handleResult (result, req, res, abortController, config, raspRule) {
   }
 
   if (abortController && !abortOnUncaughtException) {
+    console.log('blockity')
     const blockingAction = getBlockingAction(result?.actions)
 
+ console.log('blockingAction', blockingAction)
+
     // Should block only in express
-    if (blockingAction && rootSpan?.context()._name === 'express.request') {
+    if (blockingAction) { // && rootSpan?.context()._name === 'express.request') {
       const abortError = new DatadogRaspAbortError(req, res, blockingAction, raspRule, ruleTriggered)
       abortController.abort(abortError)
 
--- a/fastifybug.js
+++ b/fastifybug.js
@@ -1,0 +1,26 @@
+process.env.DD_APPSEC_ENABLED = '1'
+process.env.DD_APPSEC_RASP_ENABLED = '1'
+
+const tracer = require('.').init()
+
+//const { statSync } = require('fs')
+
+const fastify = require('fastify')()
+
+fastify.get('/rasp/lfi', async (request, reply) => {
+	const { statSync } = require('fs')
+	let result
+	try {
+	  result = JSON.stringify(statSync(request.query.file))
+	} catch (e) {
+		result = e.toString()
+
+		if (e.name === 'DatadogRaspAbortError') {
+	    	throw e
+		}
+	}
+	return result
+})
+
+
+fastify.listen({ port: 1337 })
--- a/package.json
+++ b/package.json
@@ -122,6 +122,7 @@
     "@opentelemetry/core": "^1.14.0",
     "crypto-randomuuid": "^1.0.0",
     "dc-polyfill": "^0.1.9",
+    "fastify": "^5.4.0",
     "ignore": "^7.0.5",
     "import-in-the-middle": "^1.14.2",
     "istanbul-lib-coverage": "^3.2.2",
--- a/packages/datadog-instrumentations/src/fastify.js
+++ b/packages/datadog-instrumentations/src/fastify.js
@@ -213,7 +213,7 @@ function preParsing (request, reply, payload, done) {
 function wrapSend (send, req) {
   return function sendWithTrace (payload) {
     if (payload instanceof Error) {
-      errorChannel.publish({ req, error: payload })
+      errorChannel.publish({ req, error: payload, res: req.res })
     } else if (canPublishResponsePayload(payload)) {
       const res = getRes(this)
       responsePayloadReadCh.publish({ req, res, body: payload })
@@ -237,7 +237,7 @@ function getRouteConfig (request) {
 
 function publishError (error, req) {
   if (error) {
-    errorChannel.publish({ error, req })
+    errorChannel.publish({ error, req, res: req.res })
   }
 
   return error
--- a/packages/dd-trace/src/appsec/rasp/lfi.js
+++ b/packages/dd-trace/src/appsec/rasp/lfi.js
@@ -98,9 +98,11 @@ function pathToStr (path) {
 
 function shouldAnalyze (path, fs) {
   if (!path) return
+  console.log('path', path)
 
   const notExcludedRootOp = !fs.opExcluded && fs.root
-  return notExcludedRootOp && (isAbsolute(path) || path.includes('../') || shouldAnalyzeURLFile(path, fs))
+  console.log('notExcludedRootOp', notExcludedRootOp)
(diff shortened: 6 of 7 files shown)


packages/dd-trace/src/appsec/rasp/index.js
```diff
@@ -81,15 +85,21 @@ function handleUncaughtExceptionMonitor (error) {
...
   if (!isBlocked(res)) {
-    const blocked = block(req, res, web.root(req), null, blockingAction)
+    const blockFn = isTopLevel ? block : delegateBlock
+    const blocked = blockFn(req, res, web.root(req), null, blockingAction)
     if (ruleTriggered) {
-      updateRaspRuleMatchMetricTags(req, raspRule, true, blocked)
+      // block() returns a bool, and delegateBlock() returns a promise
+      // we use Promise.resolve() to handle both cases
+      Promise.resolve(blocked).then(blocked => {
+        // TODO: bug: this should probably be called for each match even without block
```
uurien: This TODO is not precise. We are calling to the method when the operation is not blocked. The bug is that if the operation is blocked but the request is not (for example: try catched blocked operation), we are not updating the metric neither with `true` nor `false`

packages/dd-trace/src/appsec/rasp/index.js
```diff
@@ -81,15 +85,21 @@ function handleUncaughtExceptionMonitor (error) {
...
   const { req, res, blockingAction, raspRule, ruleTriggered } = abortError
   if (!isBlocked(res)) {
-    const blocked = block(req, res, web.root(req), null, blockingAction)
+    const blockFn = isTopLevel ? block : delegateBlock
+    const blocked = blockFn(req, res, web.root(req), null, blockingAction)
     if (ruleTriggered) {
-      updateRaspRuleMatchMetricTags(req, raspRule, true, blocked)
+      // block() returns a bool, and delegateBlock() returns a promise
+      // we use Promise.resolve() to handle both cases
+      Promise.resolve(blocked).then(blocked => {
```
uurien, line 99: it is only me, or having the same var name with different value instances in the same line is a little weird?

simon-id: well the whole point is that it's basically the same variable, block() returns "blocked" and delegateBlock returns a promise that resolves to "blocked", it's just wrapped in a promise. So the Promise.resolve() is used to handle both unwrapped and wrapped versions, to coalesce them to the same format of the "blocked" variable.

packages/dd-trace/src/appsec/blocking.js
```diff
@@ -127,11 +132,33 @@ function block (req, res, rootSpan, abortController, actionParameters = defaultB
...
+
+function blockDelegate (res) {
+  const delegation = blockDelegations.get(res)
+  if (delegation) {
+    const result = block.apply(this, delegation.args)
+    delegation.resolve(result)
+    return result
+  }
+}
+
```
CarlesDD: Nit: Regarding the functions naming here, both names are quite similar (they are partial anagrams) and `blockDelegate` might benefit from a clearer verb. Myabe something like `registerDelegationBlock` would make the intent of the function a bit more explicit?

Not a blocker, but I think renaming could improve readability here.

packages/dd-trace/src/appsec/rasp/utils.js
```diff
@@ -28,6 +33,12 @@ class DatadogRaspAbortError extends Error {
     this.blockingAction = blockingAction
     this.raspRule = raspRule
     this.ruleTriggered = ruleTriggered
+
+    // hide thees props to not pollute app logs
```
CarlesDD:
```suggestion
    // hide these props to not pollute app logs
```

simon-id: wow good catch hahaha

packages/dd-trace/test/appsec/rasp/utils.spec.js
```diff
@@ -152,6 +152,52 @@ describe('RASP - utils.js', () => {
...
+      web.root.returns(rootSpan)
+
+      utils.handleResult(result, req, res, abortController, config, raspRule)
+
+      sinon.assert.calledOnce(abortController.abort)
+      const abortError = abortController.abort.firstCall.args[0]
+      expect(abortError).to.be.instanceOf(utils.DatadogRaspAbortError)
+      expect(abortError.raspRule).to.equal(raspRule)
+      expect(abortError.blockingAction).to.equal(result.actions.blocking_action)
+    })
```
CarlesDD: Since the only difference between the two tests is the root span name (`express.request` vs `fastify.request`), we could parametrize the test with a loop. Maybe something like this:

```js
['express', 'fastify'].forEach(supportedWebFramework => {
    it(`should create DatadogRaspAbortError when blockingAction is present in ${supportedWebFramework}`, () => {
        const rootSpan = {
            context: sinon.stub().returns({ _name: `${supportedWebFramework}.request` })
        }
        
        {...}
    }
})
```

That way we avoid copy-pasting the exact same logic and make it easier to extend later.

packages/dd-trace/src/appsec/blocking.js
simon-id: `registerBlockDelegation` is better english, wdyt ?

packages/dd-trace/test/appsec/rasp/utils.spec.js
simon-id: alrighty then, but i'm not a fan of loops, i'd rather have a function

simon-id: like this ? [`81afd69` (#6081)](https://github.com/DataDog/dd-trace-js/pull/6081/commits/81afd692f63796e40b37831dac40ec4e1cf11047)

packages/dd-trace/src/appsec/blocking.js
simon-id: also you mean renaming delegateBlock, not blockDelegate right ? because register...() doesn't make sense for blockDelegate, only for delegateBlock.
Here is my proposal:
registerBlockDelegation()
callBlockDelegation()

CarlesDD: Your proposal sounds better 👍

simon-id: like that ? [`fd62980` (#6081)](https://github.com/DataDog/dd-trace-js/pull/6081/commits/fd62980c62a4c81c13ef55e5941103c6dc704bbc)

packages/dd-trace/src/appsec/rasp/index.js
simon-id: wdyt ?
```suggestion
        // TODO: bug: this metric is not called when the raspAbortError is caught by user, or on subsequent blockDelegations

Comment (uurien) [reply packages/dd-trace/src/appsec/rasp/index.js]: It's better, we can go with it.

File: packages/dd-trace/src/appsec/index.js [hunk 6]
```diff
@@ -306,8 +306,13 @@ function onResponseWriteHead ({ req, res, abortController, statusCode, responseH
     storedResponseHeaders.set(req, responseHeaders)
   }
 
+  // TODO: do not call waf if inside block()
+  // if (isBlocking()) {
+  //   return
+  // }
```
uurien, line 312: Is this code comment expected to be here?

packages/dd-trace/src/appsec/rasp/index.js
```diff
@@ -81,15 +85,21 @@ function handleUncaughtExceptionMonitor (error) {
...
   if (!isBlocked(res)) {
-    const blocked = block(req, res, web.root(req), null, blockingAction)
+    const blockFn = isTopLevel ? block : registerBlockDelegation
+    const blocked = blockFn(req, res, web.root(req), null, blockingAction)
     if (ruleTriggered) {
-      updateRaspRuleMatchMetricTags(req, raspRule, true, blocked)
+      // block() returns a bool, and registerBlockDelegation() returns a promise
+      // we use Promise.resolve() to handle both cases
+      Promise.resolve(blocked).then(blocked => {
+        // TODO: bug: this metric is not called when the raspAbortError is caught by user, or on subsequent blockDelegations
```
simon-id:
```suggestion
        // TODO: bug: this metric is not called when the raspAbortError is caught by user
        // or on subsequent blockDelegations
```

uurien (approved):
nit comment added, but lgtm, approving the PR

packages/dd-trace/src/appsec/index.js
simon-id: yes it's a for later todo

PR #6083, opened 2025-07-11
rewrite aws sqs batch test to be deterministic

rochdev:
### What does this PR do?

Rewrite AWS SQS batch test to be deterministic.

### Motivation

A few recent flakes all from this test. I wasn't able to reproduce locally, but my theory is that there was a race condition where one of the consumer span assertions was running before the producer assertion, so `parentId` was not set yet. This is because everything was previously running at the same time and relied on counts to figure out when the test was done. I updated the test to instead do everything in order and deterministically which should solve all flakiness.

### DIFF
--- a/packages/datadog-plugin-aws-sdk/test/sqs.spec.js
+++ b/packages/datadog-plugin-aws-sdk/test/sqs.spec.js
@@ -19,7 +19,7 @@ const getQueueParams = (queueName) => {
   }
 }
 
-describe('Plugin', () => {
+describe.only('Plugin', () => {
   describe('aws-sdk (sqs)', function () {
     setup()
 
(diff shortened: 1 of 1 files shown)


packages/datadog-plugin-aws-sdk/test/sqs.spec.js
```diff
@@ -192,56 +203,45 @@ describe('Plugin', () => {
...
+          await Promise.all([sendPromise, parentPromise])
 
-            expect(parentId).to.be.a('string')
-            expect(span.parent_id.toString()).to.equal(parentId)
-            expect(span.trace_id.toString()).to.equal(traceId)
-            batchChildSpans += 1
-            expect(batchChildSpans).to.equal(3)
-          }, { timeoutMs: 2000 }).then(done, done)
+          async function receiveAndAssertMessage () {
+            const childPromise = agent.assertSomeTraces(traces => {
```
BridgeAR, line 209: Nit: this could use our newer test helper

BridgeAR (approved):
LGTM

PR #6087, opened 2025-07-14
[] [] Fix support for Cucumber 12+

Mariovido:
### What does this PR do?

From `Cucumber 12+`, `Node 18` and `Node 23` are not supported. In order to make it work we must make sure we don't run tests with `Node 18` and `Cucumber 12+` together.

### Motivation

We want to make sure we support the latests versions of the test frameworks we support.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.

### Additional Notes

[Cucumber 12.0.0 Release Notes](https://github.com/cucumber/cucumber-js/releases/tag/v12.0.0)

### DIFF
--- a/.github/workflows/test-optimization.yml
+++ b/.github/workflows/test-optimization.yml
@@ -75,9 +75,8 @@ jobs:
   integration-ci:
     strategy:
       matrix:
-        # TODO: Add cucumber and selenium once cucumber+12 is fixed
         version: [oldest, latest]
-        framework: [jest, mocha]
+        framework: [cucumber, selenium, jest, mocha]
     runs-on: ubuntu-latest
     env:
       DD_SERVICE: dd-trace-js-integration-tests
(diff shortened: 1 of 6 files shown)


anmarchenko (dismissed):

packages/datadog-instrumentations/src/cucumber.js
```diff
@@ -398,9 +399,11 @@ function wrapRun (pl, isLatestVersion) {
         const promise = runStep.apply(this, arguments)
 
         promise.then((result) => {
-          const { status, skipReason, errorMessage } = isLatestVersion
-            ? getStatusFromResultLatest(result)
-            : getStatusFromResult(result)
+          const { status, skipReason, errorMessage } = satisfies(version, '>=12.0.0')
```
calvinbayer: You could do something like
```js
const unwrappedResult = satisfies(version, '>=12.0.0') ? result.result : result
const getStatus = satisfies(version, '>=7.3.0') ? getStatusFromResultLatest : getStatusFromResult

const { status, skipReason, errorMessage } = getStatus(unwrappedResult)
```

to get rid of the nested ternary operator. I find those hard to read

calvinbayer (approved):

packages/datadog-instrumentations/src/cucumber.js
Mariovido: You're right ternary operators make it hard to read. It is already changed :)

calvinbayer (approved):

PR #6088, opened 2025-07-14
Fix fs and rasp/iast problems

uurien:
### What does this PR do?

Fix appsec and iast fs-plugin to work as expected when multiple calls to fs methods are executed without waiting to the callback.

### Motivation

Fix a problem found with fastify + logger and fix the features.

### DIFF
--- a/packages/datadog-instrumentations/src/fs.js
+++ b/packages/datadog-instrumentations/src/fs.js
@@ -319,6 +319,7 @@ function createWrapFunction (prefix = '', override = '') {
               }
             )
           }
+
           finishChannel.runStores(ctx, () => {})
 
           return result


packages/dd-trace/src/appsec/rasp/fs-plugin.js
```diff
@@ -14,25 +14,27 @@ const enabledFor = {
...
+  return store
 }
 
 class AppsecFsPlugin extends Plugin {
   enable () {
-    this.addSub('apm:fs:operation:start', this._onFsOperationStart)
-    this.addSub('apm:fs:operation:finish', this._onFsOperationFinishOrRenderEnd)
-    this.addSub('tracing:datadog:express:response:render:start', this._onResponseRenderStart)
-    this.addSub('tracing:datadog:express:response:render:end', this._onFsOperationFinishOrRenderEnd)
+    this.addBind('apm:fs:operation:start', this._onFsOperationStart)
```
IlyasShabi, line 34: we should bind only on the first one `apm:fs:operation:start` to avoid replacing transform function, then we can listen on others

uurien: In all of them we were doing `enterWith`, we can avoid that using `addBind`. iiuc subscribing to do `enterWith` is almost the same of doing that using `addBind`

packages/dd-trace/test/appsec/rasp/fs-plugin.spec.js
```diff
@@ -225,6 +225,12 @@ describe('AppsecFsPlugin', () => {
 
       it('should clean up store when finishing op', () => {
         let count = 4
+        // TODO Remove this when node 18 is unsupported or dc-polyfill is fixed&updated
+        //  hack to node 18 and early 20.x
+        //  with dc-polyfill addBind is not enough to force a channel.hasSubscribers === true
+        const onStart = () => {}
+        opStartCh.subscribe(onStart)
```
simon-id, line 233: that sounds like a bug in the code that should be fixed in the code, not in the tests, no ?

uurien: I don't think so, at least for this channel.

This is failing because it is a unit test, but in "real life" each time that this is doing `addBind` we will have a subscription to this channel.

But maybe we need something for `tracing:datadog:express:response:render:start`. I'll take a look.

packages/dd-trace/src/appsec/rasp/fs-plugin.js
uurien: I think that it is good with `addBind`. Is the correct way to replace the store, and there is not big difference between `enterWith` and `addBind`, except that we are not modifying parent store with this one, which is better imo.

packages/dd-trace/test/appsec/rasp/fs-plugin.spec.js
uurien: Added something for `tracing:datadog:express:response:render:start` and a test

PR #6089, opened 2025-07-14
Set an upper bound for the number of timeline events gathered

szegedi:
### What does this PR do?
Establishes an upper limit on the number of non-CPU timeline events that can be gathered within a single upload period.

### Motivation
It's a bad idea for it to be unbounded.

### Additional Notes
The upper limit is rather conservative and is heuristically derived from the CPU sampling rate, the upload period, and the number of libuv worker threads (and available parallelism.) E.g. if the CPU sampling rate is 10ms and we upload every 60s, that means at most 6000 CPU samples. We'd like to be able to observe oversubscription on the libuv thread pool. If there are 4 libuv threads, we'd like to be able to observe at least 6000*5 events (events executing on at least 4 threads + 1 waiting for detecting oversubscription). Finally, another 6000 events are allowed for GC events so that gives us `uploadPeriod / samplingPeriod * (threads + 2))` as the maximum number of events we wish to sample. With our defaults of 65s upload period, 99Hz sampling rate, and 4 libuv threads we'll collect at most 38613 events (when more are gathered, events get randomly discarded.) Because we already employ sampling I believe this upper bound is not very limiting in ordinary usage – its purpose is to prevent pathological memory usage runoff.

Jira: [PROF-12127]

[PROF-12127]: https://datadoghq.atlassian.net/browse/PROF-12127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dd-trace/src/profiling/profilers/events.js
```diff
@@ -204,6 +226,29 @@ class EventSerializer {
...
+          // This will cause the samples to no longer be sorted in their array
+          // by their end time. This is fine as the backend has no ordering
+          // expectations.
+          this.samples[replacementIndex] = sample
+        }
+      }
+    }
+  }
+
+  createSample (item) {
```
watson: Shouldn't this be a private method?

packages/dd-trace/src/profiling/profilers/events.js
```diff
@@ -377,28 +443,35 @@ function createPossionProcessSamplingFilter (samplingIntervalMillis) {
...
     this.type = 'events'
-    this.eventSerializer = new EventSerializer()
+    this.maxSamples = getMaxSamples(options)
+    this.eventSerializer = new EventSerializer(this.maxSamples)
 
     const eventHandler = event => this.eventSerializer.addEvent(event)
     const eventFilter = options.timelineSamplingEnabled
       // options.samplingInterval comes in microseconds, we need millis
-      ? createPossionProcessSamplingFilter((options.samplingInterval ?? 1e6 / 99) / 1000)
+      ? createPoissonProcessSamplingFilter((options.samplingInterval ?? 1e6 / 99) / 1000)
```
watson: If I understand correctly, this PR changes `options.samplingInterval` to be milliseconds?

integration-tests/profiler/profiler.spec.js
```diff
@@ -454,6 +454,33 @@ describe('profiler', () => {
...
     })
 
+    it('number of events is limited', async () => {
+      const threads = 1
+      const uploadPeriodSec = 10
+      const samplingPeriodMs = 1000 / 99
+      // Recreates getMaxSamples from packages/dd-trace/src/profiling/profilers/events.js
+      const maxEvents = Math.floor((threads + 2) * uploadPeriodSec * 1000 / samplingPeriodMs)
+
+      // Test will try to force maxEvents + 10 events, we'll assert that only maxEvents are captured.
```
watson: What do you mean by "try"? Isn't it guaranteed to be at least maxEvents + 10 due to the test events being emitted?

watson (commented):
I'm not a fan of the test event code added to `packages/dd-trace/src/profiling/profilers/events.js`. Is there really no good way to make this test without littering the source code with code only meant for testing?

integration-tests/profiler/profiler.spec.js
szegedi: That's unfortunate wording indeed. It's just my inner pessimist allowing for any test having a chance to fail.

packages/dd-trace/src/profiling/profilers/events.js
szegedi: Excellent catch, thanks!

szegedi: I'm now extra glad I made the "possion"->"poisson" typo fix as it drew your attention to this line 😁

packages/dd-trace/src/profiling/profilers/events.js
szegedi: It's in a class that's not exported by the module so I didn't bother, but I made it private – I agree with the general sentiment of using privates where possible. I also made the fields of the exported `EventsProfiler` class private.

packages/dd-trace/src/profiling/profilers/events.js
```diff
@@ -375,52 +431,68 @@ function createPossionProcessSamplingFilter (samplingIntervalMillis) {
...
   constructor (options = {}) {
     this.type = 'events'
-    this.eventSerializer = new EventSerializer()
+    this.#maxSamples = getMaxSamples(options)
+    this.#eventSerializer = new EventSerializer(this.#maxSamples)
 
-    const eventHandler = event => this.eventSerializer.addEvent(event)
+    const eventHandler = event => this.#eventSerializer.addEvent(event)
     const eventFilter = options.timelineSamplingEnabled
       // options.samplingInterval comes in microseconds, we need millis
```
watson: This comment is outdated now and can be removed, right?

```suggestion
```

packages/dd-trace/src/profiling/profilers/events.js
```diff
@@ -375,52 +431,68 @@ function createPossionProcessSamplingFilter (samplingIntervalMillis) {
...
-      ? createPossionProcessSamplingFilter((options.samplingInterval ?? 1e6 / 99) / 1000)
+      ? createPoissonProcessSamplingFilter(getSamplingIntervalMillis(options))
       : _ => true
```
watson: While you're here, I would change this arrow function to not take any arguments. I don't assume that makes any difference, right? I think it makes the code more readable:

```suggestion
      : () => true
```

packages/dd-trace/src/profiling/profilers/events.js
```diff
@@ -204,6 +223,29 @@ class EventSerializer {
...
+      const sample = this.#createSample(item)
+      if (sample !== undefined) {
+        this.samples.push(sample)
+      }
+    } else {
+      // Choose one sample to be dropped. Using rnd(max + 1) - 1 we allow the
+      // current sample too to be the dropped one, with the equal likelihood as
+      // any other sample.
+      const replacementIndex = Math.floor(Math.random() * (this.maxSamples + 1)) - 1
+      if (replacementIndex !== -1) {
```
watson: Do I understand the algorithm correct, that `replacementIndex` will only ever be `-1` if `maxSamples` is `0`? If so wouldn't it be simpler to make the wrapping `else` block into an `else if (this.maxSamples !== 0)` block and then get rid of this `if` block completely?

Then I _think_ you could also simplify the random algorithm to `rnd(max)`, right?

szegedi: No, that's not the intent. The idea is as follows: I have _m_ samples in the array already which is the maximum (`maxSamples` in the code.) I get one more so now I have _m+1_ samples and I need to discard one to get back to _m_ samples. I could just throw away a sample from the array but that favorably treats the new sample – it is not considered to be discarded. It is fairer if the new sample is also a discard candidate so the algorithm doesn't have a bias towards newer samples (I want a random distribution over the entire sampled period.) So I need a random number between _[0, m+1)_  instead of _[0, m)_. The idea is that if the generated random number is in _[0, m)_ I throw away one of samples already in the array at that index. If the number is _m_ though, I throw away the newly arrived sample. 

I found it more elegant to implement this logic by subtracting 1 from the generated random value, so now I get a number that is in _[-1, m)_ and instead of comparing to _m_ I can compare to -1 instead. An alternative but equivalent solution would be:
```javascript
const replacementIndex = Math.floor(Math.random() * (this.maxSamples + 1)) 
      if (replacementIndex !== this.maxSamples) {
```

nsavoire: nit: hum, that does not look right:
the last added event will have a probability of being kept of `1 - 1 / (maxSamples+1)` whereas it should be (for a true random sampling): `maxSamples/n` (where n is the total number of events).
What you need here is reservoir sampling (where `n` is the total number of processed events, including the current one):
```js
const replacementIndex = Math.floor(Math.random() * n)
   if (replacementIndex < this.maxSamples) {
```

szegedi: you're right, of course; not sure what I was thinking with my bespoke probability calculation. Fixed.

nsavoire (approved):

PR #6092, opened 2025-07-14
Update allowed values for config.logInjection

ida613:
### What does this PR do?
ON HOLD while RFC is being updated
Update `config.logInjection` to take string values and deprecate boolean values

### Motivation
[RFC](https://docs.google.com/document/d/1Okl7hQcJgA9NdQ8msSVu5P2cPQOx139K9s9f3C_bTho/edit?usp=sharing)

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/plugins/log_plugin.js
+++ b/packages/dd-trace/src/plugins/log_plugin.js
@@ -2,6 +2,7 @@
 
 const { LOG } = require('../../../../ext/formats')
 const Plugin = require('./plugin')
+const log = require('../log')
 const { storage } = require('../../../datadog-core')
 
 function messageProxy (message, holder) {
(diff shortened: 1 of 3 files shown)


packages/dd-trace/src/plugins/log_plugin.js
```diff
@@ -46,10 +47,17 @@ module.exports = class LogPlugin extends Plugin {
...
+    (config.logInjection === 'all' || config.logInjection === true || config.ciVisAgentlessLogSubmissionEnabled)
   }
 
   configure (config) {
+    if (config.logInjection === true) {
+      log.warn('Boolean value for config.logInjection is deprecated. Use "all" or "structured" instead.')
+    }
+    if (config.logInjection === false) {
+      log.warn('Boolean value for config.logInjection is deprecated. Use "disabled" instead.')
+    }
```
szegedi, line 60: Would it make sense to replace boolean values with equivalent string values as early as possible, e.g. here or whereever this method receives the config object? Then you wouldn't have to worry about boolean values anymore elsewhere in the code.

rochdev: Definitely agree with @szegedi here, I don't see why a string should be enforced. If this really is something we want to enforce though, we need to put this behind a check for `DD_MAJOR` as it's a major change that will need to only be in the next major.

ida613:
closing for now since RFC has been updated

PR #6100, opened 2025-07-15
add explicit dependabot label for use in automation

rochdev:
### What does this PR do?

Add explicit `dependabot` label for use in automation.

### Motivation

Makes it easier to figure out if a PR was open by dependabot in scripts or GitHub automation.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,11 +16,19 @@ updates:
       gh-actions-packages:
         patterns:
           - "*"
+    labels:
+      - dependabot
+      - dependencies
+      - github_actions
   - package-ecosystem: "npm"
     directories:
       - "/"
     schedule:
       interval: "weekly"
+    labels:
+      - dependabot
+      - dependencies
+      - javascript
     ignore:
       - dependency-name: "@types/node"
         # Update the types manually with new Node.js version support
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -16,11 +16,21 @@ updates:
...
+      - semver-patch
   - package-ecosystem: "npm"
     directories:
       - "/"
     schedule:
       interval: "weekly"
+    labels:
+      - dependabot
+      - dependencies
+      - javascript
```
watson, line 32: Wouldn't it make more sense that the label was `npm` in this case?

rochdev: I just used the default labels and added `dependabot` and `semver-patch`.

watson: Where does the default label `javascript` come from then?

rochdev: It's one of the default 2 labels (`dependencies` and `javascript`). When specifying labels, it removes all the default ones so I had to add them back here to keep them.

watson: But why would we want the default `javascript` Dependabot label? It's not adding any value. I think the label `npm` makes a lot more sense.

rochdev: I mean nobody was complaining before and changing this was out of the scope of this PR so I just left the exact same values.

PR #4475, opened 2024-07-05
refactor: move remote_config code to the root of dd-trace

watson:
### What does this PR do?

Move the folder `remote_config` on level up from `packages/dd-trace/src/appsec` to `packages/dd-trace/src`.

### Motivation

The Remote Configuration code was first introduced by the appsec team. Now that more and more teams are using it, it makes sense to move it to a shared location in the root of the `dd-trace` package.

PR #6102, opened 2025-07-15
Fix RC tests

CarlesDD:
### What does this PR do?
Applies the necessary changes so that the remote config tests are run under `test:trace:core` (and therefore in the CI)

### Motivation
After moving the remote config code to the root of the dd-trace (see #4475), the RC tests fell into no man's land, not being executed in the CI.

This PR aims to get RC tested again.


simon-id:
`src/remote_config` but `tests/remote-config` ? i don't like that so much :/

watson:
It would be best if we could reverse the logic, so that all sub-folders were automatically tested, unless specifically added to a disallow list.

PR #6106, opened 2025-07-16
Tag profiles with a sequence number

szegedi:
### What does this PR do?
Adds a `profile_seq` tag to profiles. It starts at 0 for a specific `runtime-id` and increments each time a profile is emitted.

### Motivation
Go and .NET profilers already emit `profile_seq` and Ruby profiler just recently [introduced it](https://github.com/DataDog/dd-trace-rb/pull/4794), so we're bringing it up to feature parity. The feature can be used to e.g. exclude profiles from application startup in analysis, or to easily compare profiles with a set difference of sequence number for e.g. memory leak analysis. It can also be used to detect missing profiles if there are gaps in the sequence number.

### Additional info
Jira: [PROF-12136]

### DIFF
--- a/packages/dd-trace/src/profiling/profiler.js
+++ b/packages/dd-trace/src/profiling/profiler.js
@@ -124,11 +124,12 @@ class Profiler extends EventEmitter {
 
     try {
       const start = new Date()
+      const nearOOMCallback = this._nearOOMExport.bind(this)
       for (const profiler of config.profilers) {
         // TODO: move this out of Profiler when restoring sourcemap support
         profiler.start({
           mapper,
-          nearOOMCallback: this._nearOOMExport.bind(this)
+          nearOOMCallback
         })
         this._logger.debug(`Started ${profiler.type} profiler in ${threadNamePrefix} thread`)
       }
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/profiling/profiler.js
```diff
@@ -50,6 +50,7 @@ class Profiler extends EventEmitter {
     this._timer = undefined
     this._lastStart = undefined
     this._timeoutInterval = undefined
+    this._profile_seq = 0
```
nsavoire: nit:
```suggestion
    this._profileSeq = 0
```

BridgeAR: I think we should start using actual private properties instead.
We already do that in other spots. Maybe it would even be possible to convert a few other of these properties to private properties as well? :)
```suggestion
    this.#profile_seq = 0
```

szegedi: 👍 I figured that since the tag is `profile_seq` it'd be consistent with it, but fair to keep the camel case for the variable.

szegedi: I prefer keeping PRs to the point as much as possible so I'll change this one for now. (Yes, I did hoist a bound function out of the loop in this one as drive-by code gardening but that has much smaller surface area than replacing all these variables with private fields would have.)

szegedi: https://github.com/DataDog/dd-trace-js/pull/6108

PR #6107, opened 2025-07-16
chore(openai): refactor streamed response handling to the tracing plugin

sabrenner:
### What does this PR do?
Moves streamed response coalescing to the tracing plugin instead of the instrumentation. This allows us:

1. To start decoupling business logic from the instrumentation/simplifying it in general
2. Makes it easier to add different streamed implementations for different endpoints going forward (as we plan OpenAI responses soon)

### Motivation
Easier maintenance, developer experience, decoupling.

MLOB-3331

### DIFF
--- a/packages/datadog-plugin-openai/src/stream-helpers.js
+++ b/packages/datadog-plugin-openai/src/stream-helpers.js
@@ -1,0 +1,113 @@
+'use strict'
+
+/**
+ * Combines legacy OpenAI streamed chunks into a single object.
+ * These legacy chunks are returned as buffers instead of individual objects.
+ * @param {readonly Uint8Array[]} chunks
+ * @returns {Array<Record<string, any>>}
+ */
+function convertBuffersToObjects (chunks) {
+  return Buffer
+    .concat(chunks) // combine the buffers
+    .toString() // stringify
+    .split(/(?=data:)/) // split on "data:"
+    .map(chunk => chunk.replaceAll('\n', '').slice(6)) // remove newlines and 'data: ' from the front
+    .slice(0, -1) // remove the last [DONE] message
+    .map(JSON.parse) // parse all of the returned objects
+}
+
+/**
+ * Constructs the entire response from a stream of OpenAI completion chunks,
+ * mainly combining the text choices of each chunk into a single string per choice.
+ * @param {Array<Record<string, any>>} chunks
+ * @param {number} n the number of choices to expect in the response
+ * @returns {Record<string, any>}
+ */
+function constructCompletionResponseFromStreamedChunks (chunks, n) {
+  const body = { ...chunks[0], choices: Array.from({ length: n }) }
+
+  for (const chunk of chunks) {
+    body.usage = chunk.usage
+    for (const choice of chunk.choices) {
+      const choiceIdx = choice.index
+      const oldChoice = body.choices.find(choice => choice?.index === choiceIdx)
+      if (oldChoice) {
+        if (!oldChoice.finish_reason) {
+          oldChoice.finish_reason = choice.finish_reason
+        }
+
+        const text = choice.text
+        if (text) {
+          if (oldChoice.text) {
+            oldChoice.text += text
+          } else {
+            oldChoice.text = text
+          }
+        }
+      } else {
+        body.choices[choiceIdx] = choice
+      }
+    }
+  }
+
+  return body
+}
+
+/**
+ * Constructs the entire response from a stream of OpenAI chat completion chunks,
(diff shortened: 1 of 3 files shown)


packages/datadog-plugin-openai/src/tracing.js
```diff
@@ -48,6 +53,41 @@ class OpenAiTracingPlugin extends TracingPlugin {
...
+      if (!done) return
+
+      let chunks = ctx.chunks
+      if (chunks.length === 0) return
+
+      const firstChunk = chunks[0]
+      // TODO(BridgeAR): It likely depends on the options being passed
+      // through if the stream returns buffers or not. By reading that,
+      // we don't have to do the instanceof check anymore, which is
+      // relatively expensive.
```
sabrenner, line 70: @BridgeAR I saw this TODO while refactoring that I'm guessing is from some cleanup work you were doing and stumbled across this code. 

As far as I know, this is a legacy behavior which OpenAI (circa 4.1.0, they are now on version 5.9.2) did not provide an option for specifying the type of the chunks - it was handled with an intermediary which I believe we cannot wrap. Feel free to let me know what your thoughts are on this 😄

tlhunter (approved):

packages/datadog-plugin-openai/src/stream-helpers.js
```diff
@@ -0,0 +1,113 @@
...
+ * @returns {Record<string, any>}
+ */
+function constructCompletionResponseFromStreamedChunks (chunks, n) {
+  const body = { ...chunks[0], choices: Array.from({ length: n }) }
+
+  for (const chunk of chunks) {
+    body.usage = chunk.usage
+    for (const choice of chunk.choices) {
+      const choiceIdx = choice.index
+      const oldChoice = body.choices.find(choice => choice?.index === choiceIdx)
```
BridgeAR, line 33: How many choices are we likely going to have? Is it more in the low single digits or higher? If it's higher, I would use a map for a faster lookup.

packages/datadog-plugin-openai/src/stream-helpers.js
```diff
@@ -0,0 +1,113 @@
...
+ * @returns {Record<string, any>}
+ */
+function constructChatCompletionResponseFromStreamedChunks (chunks, n) {
+  const body = { ...chunks[0], choices: Array.from({ length: n }) }
+
+  for (const chunk of chunks) {
+    body.usage = chunk.usage
+    for (const choice of chunk.choices) {
+      const choiceIdx = choice.index
+      const oldChoice = body.choices.find(choice => choice?.index === choiceIdx)
```
BridgeAR, line 70: Same as above

packages/datadog-plugin-openai/src/stream-helpers.js
```diff
@@ -0,0 +1,113 @@
...
+}
+
+/**
+ * Constructs the entire response from a stream of OpenAI chat completion chunks,
+ * mainly combining the text choices of each chunk into a single string per choice.
+ * @param {Array<Record<string, any>>} chunks
+ * @param {number} n the number of choices to expect in the response
+ * @returns {Record<string, any>}
+ */
+function constructChatCompletionResponseFromStreamedChunks (chunks, n) {
```
BridgeAR, line 63: We could consolidate these two methods and use a smaller part for the unique handling and just pass through that smaller method. That prevents code duplication :)

packages/datadog-plugin-openai/src/tracing.js
BridgeAR: In that case we should probably remove the comment or change it to explain that? :)

BridgeAR (commented):
I was too late as it seems :D

packages/datadog-plugin-openai/src/stream-helpers.js
sabrenner: i thought about that but just wanted to be explicit in how we're handling each case - if that's all good with you! i thought trying to extract it to pass something like a `onIndexedChunk` function to the shared handler function would hinder readability/initial understanding a bit, but if you feel strongly i can implement it!

packages/datadog-plugin-openai/src/stream-helpers.js
sabrenner: likely lower (n <= 3), so i think this operation should be fine

packages/datadog-plugin-openai/src/tracing.js
sabrenner: yep i can do that 😄

PR #6110, opened 2025-07-17
ci: make telemetry guardrails tests more stable

watson:
### What does this PR do?

Tells the mock telemetry endpoint to expect telemetry in the guardrails tests, so that it will retry up to 10 times if the `telemetry-forwarder.sh` script exits with an `ENOENT` exit code.

### Motivation

An attempt to make this test more stable in CI (though not sure if this actually will help, but it will not hurt either).

### DIFF
--- a/packages/dd-trace/test/guardrails/telemetry.spec.js
+++ b/packages/dd-trace/test/guardrails/telemetry.spec.js
@@ -8,7 +8,7 @@ describe('sendTelemetry', () => {
   let cleanup, sendTelemetry
 
   beforeEach(() => {
-    cleanup = telemetryForwarder()
+    cleanup = telemetryForwarder(true)
     sendTelemetry = proxyquire('../src/guardrails/telemetry', {})
   })
 


packages/dd-trace/test/guardrails/telemetry.spec.js
```diff
@@ -8,7 +8,7 @@ describe('sendTelemetry', () => {
   let cleanup, sendTelemetry
 
   beforeEach(() => {
-    cleanup = telemetryForwarder()
+    cleanup = telemetryForwarder(true)
```
uurien, line 11: This is already defaulted to `true`, am I missing something?

https://github.com/DataDog/dd-trace-js/blob/57efb0af7991982ef47316bcb9701003f5b2954d/integration-tests/helpers/index.js#L243

watson: Ah, you're right 🤦

PR #6115, opened 2025-07-17
remove npm view usage from plugin test version install

rochdev:
### What does this PR do?

Remove `npm view` usage from plugin test version install.

### Motivation

Fix rate limit errors such as https://github.com/DataDog/dd-trace-js/actions/runs/16342472069/job/46168161260#step:4:466

### DIFF
--- a/scripts/install_plugin_modules.js
+++ b/scripts/install_plugin_modules.js
@@ -141,34 +141,6 @@ async function assertPackage (name, version, dependencyVersionRange, external) {
 }
 
 /**
- * @param {object} dependencies
- * @param {string} externalName
- * @param {string} versionRange
- */
-async function addDependencies (dependencies, externalName, versionRange) {
-  const versionList = await getVersionList(externalName)
-  const version = semver.maxSatisfying(versionList, versionRange)
-  const pkgJson = await npmView(`${externalName}@${version}`)
-  for (const { dep, name } of externalDeps.get(externalName)) {
-    // do stuff with dep
-    for (const section of ['devDependencies', 'peerDependencies']) {
-      if (pkgJson[section] && name in pkgJson[section]) {
-        if (dep === externalName) {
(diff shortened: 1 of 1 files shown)


scripts/install_plugin_modules.js
```diff
@@ -144,59 +141,52 @@ async function assertPackage (name, version, dependencyVersionRange, external) {
...
- * @returns {Promise<string[]>}
- */
-async function getVersionList (name) {
-  let list = versionListCache.get(name)
-  if (list) return list
-  list = await npmView(`${name} versions`)
-  versionListCache.set(name, list)
-  return list
-}
+    const externalName = [parent, entry.split('@')[0]].filter(v => v).join('/')
```
watson: If you change `parent` to default to an empty string in the method definition, you can change this logic to use `path.join` instead:


```suggestion
    const externalName = join(parent, entry.split('@')[0])
```

That will make the code much easier to understand and more robust.

scripts/install_plugin_modules.js
```diff
@@ -144,59 +141,52 @@ async function assertPackage (name, version, dependencyVersionRange, external) {
...
+
+    const versionPkgJsonPath = join(folder, 'package.json')
+    const versionPkgJson = require(versionPkgJsonPath)
+
+    for (const { dep, name } of externalDeps.get(externalName)) {
+      const pkgJsonPath = require(folder).pkgJsonPath()
+      const pkgJson = require(pkgJsonPath)
+
+      for (const section of ['devDependencies', 'peerDependencies']) {
+        if (pkgJson[section] && name in pkgJson[section]) {
```
watson: Nit: It seems odd that you're mixing two different ways to detect if a property exists in an object, I'd personally just do:

```suggestion
        if (pkgJson[section]?.[name]) {
```

PR #6116, opened 2025-07-17
starting io removal for langchain

jsimpher:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/datadog-plugin-langchain/src/tracing.js
```diff
@@ -4,15 +4,12 @@ const { MEASURED } = require('../../../ext/tags')
...
-const API_KEY = 'langchain.request.api_key'
 const MODEL = 'langchain.request.model'
 const PROVIDER = 'langchain.request.provider'
 const TYPE = 'langchain.request.type'
 
 const LangChainHandler = require('./handlers/default')
-const LangChainChatModelHandler = require('./handlers/language_models/chat_model')
-const LangChainLLMHandler = require('./handlers/language_models/llm')
-const LangChainChainHandler = require('./handlers/chain')
+const LangChainLanguageModelHandler = require('./handlers/language_models/index')
```
sabrenner: just a js nit: we can do `language_models/index.js` --> `language_models.js` and import it as
```suggestion
const LangChainLanguageModelHandler = require('./handlers/language_models')
```

and it should be the same, to make it a bit cleaner!

packages/datadog-plugin-langchain/src/handlers/default.js
```diff
@@ -10,15 +10,6 @@ class LangChainHandler {
     this.isPromptCompletionSampled = utilities.isPromptCompletionSampled
```
sabrenner: i think we can get rid of this `isPromptCompletionSampled` property, and the above line for `makeUtilites`, then removing the import as well and probably the constructor here. i think it was only used for i/o tagging

sabrenner (approved):
🚢

PR #6124, opened 2025-07-18
Temporarily cap supported `@apollo/server` version <5 

CarlesDD:
### What does this PR do?
The latest version of `@apollo/server` (v5), which was just released, breaks our CI. 
Until a proper fix can be implemented, cap the supported range in appsec test to <5

### DIFF
--- a/packages/dd-trace/test/appsec/graphql.apollo-server.plugin.spec.js
+++ b/packages/dd-trace/test/appsec/graphql.apollo-server.plugin.spec.js
@@ -8,7 +8,7 @@ const {
   graphqlCommonTests
 } = require('./graphq.test-utils')
 
-withVersions('apollo-server', '@apollo/server', '>=4.0.0 <5.0.0', apolloServerVersion => {
+withVersions('apollo-server', '@apollo/server', apolloServerVersion => {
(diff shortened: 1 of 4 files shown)


packages/dd-trace/test/appsec/graphql.apollo-server.plugin.spec.js
```diff
@@ -8,7 +8,7 @@ const {
   graphqlCommonTests
 } = require('./graphq.test-utils')
 
-withVersions('apollo-server', '@apollo/server', apolloServerVersion => {
+withVersions('apollo-server', '@apollo/server', '>=4.0.0 <5.0.0', apolloServerVersion => {
```
uurien: as we are removing the instrumentation for `@apollo/server>=5` is this check necessary? and the same question for the change in `verify-ci-config.js`

CarlesDD: They are not necessary, but I think it is worthwhile to indicate in the test the range of versions for which the test is valid.

watson (dismissed):

uurien (approved):

PR #6132, opened 2025-07-20
update audit workflow to only run when lockfile is changed

rochdev:
### What does this PR do?

Update audit workflow to only run when lockfile is changed.

### Motivation

Otherwise the main branch can start randomly failing when a new vulnerability is detected. We already have the nightly that runs every day which should be enough to catch vulnerabilities early without causing unnecessary noise, blocked PRs and potential code freezes.

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -6,6 +6,8 @@ on:
       - yarn.lock
   push:
     branches: [master]
+    paths:
+      - yarn.lock
   schedule:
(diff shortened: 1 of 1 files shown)


simon-id:
>We already have the nightly that runs every day which should be enough to catch vulnerabilities early

right but who looks at the nightlies ?

watson:
Does the fact that `master` is failing due to audit issues block anybody today? We've already done for PRs what you're doing now for `master`.

PR #6135, opened 2025-07-21
support next 15.4.x on appsec

IlyasShabi:
### What does this PR do?
New Next.js versions 15.4 is breaking our current instrumentation with Appsec, specially `json()`, `text()`, `formData()` and also query string.

### Motivation

At runtime `Dynamic IO` which is introduced by Next.js 14.4 redefines the methods on `NextRequest.prototype` (`json`, `text`, `formData`, `nextUrl`, …).  
Our instrumentation also wraps those methods, but it runs earlier; so Dynamic IO overwrites our wrappers.

The proposed solution is:

1. Wrap the lower level Web `Request.prototype` instead of `NextRequest.prototype`
2. Parse query strings once in `instrument()` helper, removing the dependency on `nextUrl` 

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/datadog-instrumentations/src/next.js
```diff
@@ -280,24 +292,9 @@ addHook({
   versions: ['>=13'],
   file: 'dist/server/web/spec-extension/request.js'
 }, request => {
-  shimmer.wrap(request.NextRequest.prototype, 'nextUrl', function (originalGet) {
```
IlyasShabi, line 283: Move the query string to `instrument()`to have it in one single place

packages/datadog-instrumentations/src/next.js
```diff
@@ -141,6 +141,18 @@ function instrument (req, res, handler, error) {
   requests.add(req)
 
   const ctx = { req, res }
+  // Parse query parameters from request URL
+  if (queryParsedChannel.hasSubscribers && req.url) {
+    const url = new URL(req.url, 'http://n')
```
IlyasShabi: Create an URL based on req.url to get query string

uurien: why is `http://n` required?

watson: Otherwise `new URL` will throw an exception because there's no protocol or hostname

IlyasShabi: `req.url` is only the relative path (`/foo?bar=baz`)  and `new URL()` needs a full URL, so we give it a dummy base.
This have no other effect, do you have a better idea?

packages/datadog-instrumentations/src/next.js
```diff
@@ -141,6 +141,18 @@ function instrument (req, res, handler, error) {
...
+    const query = {}
+    for (const key of url.searchParams.keys()) {
+      if (!query[key]) {
+        query[key] = url.searchParams.getAll(key)
+      }
+    }
```
watson: A better approach is to use the code `querystring` module, so you don't have to generate a complete URL object:

```suggestion
    const query = querystring.parse(req.url)
```

watson (commented):
Any reason why this only focuses on AppSec? Would be be hard to also fix the support for the instrumentation part of the code?

packages/datadog-instrumentations/src/next.js
watson: See my review below

packages/datadog-instrumentations/src/next.js
uurien:
```
> querystring.parse('/foo?bar=baz')
[Object: null prototype] { '/foo?bar': 'baz' }
```
With this approach without `new URL()` we need a regexp (or another step) to clean first param name

watson: Ah sorry, my bad. This would be the correct alternative:

```suggestion
    const { query } = url.parse(req.url, true)
```

But that's of course not 100% spec compliant (and now that I think of it, I guess the `querystring` module isn't either). However, I would still like to improve upon your implementation slightly:

```suggestion
    const { searchParams } = new URL(req.url, 'http://dummy')
    const query = {}
    for (const key of searchParams.keys()) {
      query[key] = searchParams.getAll(key)
    }
```

packages/datadog-instrumentations/src/next.js
```diff
@@ -280,24 +294,9 @@ addHook({
...
-            query[key] = nextUrl.searchParams.getAll(key)
-          }
-        }
-
-        queryParsedChannel.publish({ query })
-      }
-      return nextUrl
-    }
-  })
+  const requestProto = Object.getPrototypeOf(request.NextRequest.prototype)
```
simon-id, line 297: you get the prototype of the prototype ? in this case why mix the syntax between .prototype and .getPrototypeOf() ?

packages/datadog-instrumentations/src/next.js
simon-id: why remove the `if (!query[key]) {` ? it was written like this in the previous instrumentation, probably for a reason no ?

packages/datadog-instrumentations/src/next.js
IlyasShabi: Previous code was using `.prototype` for a reason which allows to get the prototype object where instance methods are defined. Using `getPrototypeOf` in this case is not working because `getPrototypeOf(NextRequest)` gets the parent constructor, not the prototype object where methods live.

In my case I need to get access to global Request prototype which contains `json()` method. Using `.prototype` is not working because `NextRequest.prototype` only contains NextRequest specific methods, not the inherited Request methods.

packages/datadog-instrumentations/src/next.js
watson: @simon-id Yeah that was a mistake on my part

simon-id (approved):

PR #6138, opened 2025-07-21
[test-optimization] [SDTEST-2267] Fix flakiness in cypress tests

Mariovido:
### What does this PR do?

This PR modify the way we create server. Instead of creating it in the `before` we are importing the server.

### Motivation

We want to reduce flakiness of cypress tests.

### Plugin Checklist

- [x] Unit tests.
- [x] Integration tests.

### DIFF
--- a/integration-tests/ci-visibility/web-app-server-without-rum.js
+++ b/integration-tests/ci-visibility/web-app-server-without-rum.js
@@ -1,15 +1,19 @@
 'use strict'
 
-// File to spin an HTTP server that returns an HTML for playwright to visit
 const http = require('http')
 
-module.exports = http.createServer((req, res) => {
-  res.setHeader('Content-Type', 'text/html')
-  res.writeHead(200)
-  res.end(`
-    <!DOCTYPE html>
-    <html>
+const createSimpleServer = () => {
+  return http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'text/html')
+    res.writeHead(200)
+    res.end(`
+      <!DOCTYPE html>
+      <html>
       <div class="hella-world">Hella World</div>
-    </html>
-  `)
-})
+      </html>
+      `)
+  })
+}
+
+// When this file gets imported, it will spin up an HTTP server that returns an HTML for cypress/playwright to visit
+module.exports = createSimpleServer()


integration-tests/ci-visibility/web-app-server-without-rum.js
```diff
@@ -0,0 +1,15 @@
+'use strict'
+
+// File to spin an HTTP server that returns an HTML for playwright to visit
```
calvinbayer: I would personally prefer a well-named function over a comment as the function name will more likely evolve with the code and comments are prone to be overlooked and outdated.

```js
function createServerThatServesStaticHtml () {
    // ...
}

module.exports = createServerThatServesStaticHtml()
``` 

It's interesting to see that this server is created on import of the file rather than by invoking an imported function. I see the pattern for other files too, but would be curios to understand why?

In this case, I would understand adding a comment to the file saying something like `// When importing this file, a HTTP server is created that serves a static HTML document for test automation frameworks to visit`. I would put that on line 2 rather than line 3 as I initially assumed you were adding context to the `const http = require('http')` and couldn't make much sense of the comment in that context.

integration-tests/cypress/cypress.spec.js
```diff
@@ -14,6 +13,7 @@ const {
 } = require('../helpers')
 const { FakeCiVisIntake } = require('../ci-visibility-intake')
 const webAppServer = require('../ci-visibility/web-app-server')
+const secondWebAppServer = require('../ci-visibility/web-app-server-without-rum')
```
calvinbayer, line 16: Why do you think creating the server on import rather than in the `before` hook will improve the test's stability?

Mariovido: This is just a guess, but I believe structuring it this way could help us avoid race conditions in the `before` hook. The concern is that if we create the server within `before`, we might run into issues when trying to retrieve the port at the same stage. Also, in Playwright we follow this pattern for server creation, and it seems to work good without introducing flakiness.
In any case, I am open to hear other ideas on how to solve this :)

integration-tests/ci-visibility/web-app-server-without-rum.js
Mariovido: - Regarding function over comments: I agree with you so I will make the change to use functions naming instead of comments. I did in this way because it was the way it was done in other files.
- In the case of why we should import the file instead of importing a function, I don't know how to answer you this but probably it is because we want to create the server when importing instead of calling the function in a `before`. It is somehow related to this [Comment Thread](https://github.com/DataDog/dd-trace-js/pull/6138#discussion_r2219623659)

integration-tests/cypress/cypress.spec.js
calvinbayer: > This is just a guess, but I believe structuring it this way could help us avoid race conditions in the before hook.

`http.createServer` is a factory for creating a server instance. Also, it's a synchronous function call. How could that possibly cause a race condition when accessing method's of that object?

calvinbayer (approved):

integration-tests/cypress/cypress.spec.js
Mariovido: As I said this this is a guess (I tried to imitate Playwright tests, which seems to not be flaky), but the race condition could be in the `listen`, where we get the port. If this doesn't fix the flakiness we might need to think another way to avoid this.
```js
secondWebAppServer.listen(0, 'localhost', () => {
  secondWebAppPort = secondWebAppServer.address().port
})
```

calvinbayer (approved):

PR #6141, opened 2025-07-22
Update dc-polyfill and remove patches

uurien:
### What does this PR do?

Updates dc-polyfill and removes `// TODO` related with the update

### Motivation

Remove "hacks" and use the latest dc-polyfill.

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     "@opentelemetry/api": "1.8.0",
     "@opentelemetry/core": "^1.14.0",
     "crypto-randomuuid": "^1.0.0",
-    "dc-polyfill": "^0.1.9",
+    "dc-polyfill": "^0.1.10",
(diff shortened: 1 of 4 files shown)


watson:
Is the benchmark regression expected in Node.js 18?

uurien:
> Is the benchmark regression expected in Node.js 18?

I checked the tests and yes, I think it is expected because the tests are executing millions of `log['error|info|debug|…']`, these methods are quite short and fast, and one of the steps in the method is `channel.hasSubscribers` what is exactly what we have modified in `dc-polyfill` to [fix the bug](https://github.com/DataDog/dc-polyfill/pull/19/files), now each time that hasSubscribers is executed in node 18, an extra check is executed and it takes a bit longer.

watson (approved):

PR #6146, opened 2025-07-22
[test-optimization] [SDTEST-2342] Normalize `PR_NUMBER`

Mariovido:
### What does this PR do?

With this change we want to convert the `PR_NUMBER` from a number to a string.

### Motivation

We want to align with the rest of the tracers.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.

### DIFF
--- a/packages/dd-trace/src/plugins/util/ci.js
+++ b/packages/dd-trace/src/plugins/util/ci.js
@@ -86,6 +86,13 @@ function resolveTilde (filePath) {
   return filePath
 }
 
+function normalizeNumber (number) {
+  if (!number) {
+    return number
+  }
+  return number.toString()
+}
+
 function getGitHubEventPayload () {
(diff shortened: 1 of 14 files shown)


packages/dd-trace/src/plugins/util/ci.js
```diff
@@ -86,6 +86,13 @@ function resolveTilde (filePath) {
   return filePath
 }
 
+function normalizeNumber (number) {
+  if (!number) {
```
watson: If `number` is `0` it would not be stringified. Maybe do this instead

```suggestion
  if (typeof number !== 'number') {
```

Mariovido: You're right. Thanks for noticing! Already pushed the change

watson (approved):

PR #6148, opened 2025-07-22
chore: upgrade eslint-plugin-unicorn to v60.0.0

watson:
### What does this PR do?

This PR does two things:
1. Upgrade `eslint-plugin-unicorn` from v59.x to v60.x.
2. Fix any linting rule errors that resulted from the upgrade.

### Motivation

This upgrades a vulnerable dependency, so that we no longer have to use the `resolutions` field in `package.json` to force-upgrade it.

### Additional Notes

The minimum support Node.js version of `eslint-plugin-unicorn` v60 is changed from v18.20.0 is to v20.10.0. However, this shouldn't be an problem, since we run the linter on the currently active LTS version of Node.js (which as of this writing is version 22).

### DIFF
--- a/packages/dd-trace/src/profiling/profilers/events.js
+++ b/packages/dd-trace/src/profiling/profilers/events.js
@@ -375,10 +375,10 @@ function createPossionProcessSamplingFilter (samplingIntervalMillis) {
  * source with a sampling event filter and an event serializer.
  */
 class EventsProfiler {
-  constructor (options = {}) {
-    this.type = 'events'
-    this.eventSerializer = new EventSerializer()
+  type = 'events'
+  eventSerializer = new EventSerializer()
 
+  constructor (options = {}) {
     const eventHandler = event => this.eventSerializer.addEvent(event)
     const eventFilter = options.timelineSamplingEnabled
       // options.samplingInterval comes in microseconds, we need millis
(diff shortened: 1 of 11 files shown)


watson:
* **#6148** <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/6148?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/dd-trace-js/6148?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/dd-trace/src/profiling/profilers/events.js
```diff
@@ -375,10 +375,10 @@ function createPossionProcessSamplingFilter (samplingIntervalMillis) {
  * source with a sampling event filter and an event serializer.
  */
 class EventsProfiler {
-  constructor (options = {}) {
-    this.type = 'events'
-    this.eventSerializer = new EventSerializer()
+  type = 'events'
```
szegedi, line 378: Ouch, this'll give me a merge conflict in #6089.

packages/dd-trace/src/profiling/profilers/wall.js
```diff
@@ -69,8 +69,12 @@ function ensureChannelsActivated () {
 }
 
 class NativeWallProfiler {
+  type = 'wall'
+  _mapper
+  _pprof
```
simon-id, line 74: note to self: do not use field = undefined, otherwise it ends up being a ugly lone label like that

simon-id (approved):
LGTM, it's touching other teams code style but to me it's a reasonable enough change to "impose" everywhere

PR #6154, opened 2025-07-22
first run at bedrock

jsimpher:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/tracing.js
+++ b/packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/tracing.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const BaseAwsSdkPlugin = require('../../base')
-const { parseModelId, extractRequestParams, extractTextAndResponseReason } = require('./utils')
+const { parseModelId } = require('./utils')
 
 const enabledOperations = new Set(['invokeModel'])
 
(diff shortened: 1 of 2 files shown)


packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/tracing.js
```diff
@@ -20,42 +20,19 @@ class BedrockRuntime extends BaseAwsSdkPlugin {
   generateTags (params, operation, response) {
     const { modelProvider, modelName } = parseModelId(params.modelId)
 
-    const requestParams = extractRequestParams(params, modelProvider)
-    const textAndResponseReason = extractTextAndResponseReason(response, modelProvider, modelName)
-
-    const tags = buildTagsFromParams(requestParams, textAndResponseReason, modelProvider, modelName, operation)
+    const tags = buildTagsFromParams(modelProvider, modelName, operation)
```
sabrenner: i think just to clean this up even more, we don't really need this method, we can just define the tags in place
```suggestion
    const tags = {
      'resource.name': operation,
      'aws.bedrock.request.model': modelName,
      'aws.bedrock.request.model_provider': modelProvider.toLowerCase()
    }
```
and then we can git rid of the `buildTagsFromParams` function

sabrenner: oh actually even better, just should be able to return them directly from here, we don't really need to assign them to a `tags` variable

sabrenner (approved):
:shipit:

PR #6172, opened 2025-07-25
Ensure advancement of sampling instant never takes much time

szegedi:
### What does this PR do?
The Poisson process filter has a loop in the code that computes the next sampling instant. This change ensures that the loop only runs for few iterations once every 10ms. It can run for long periods of time if either:
* the next reported event erroneously has such a long duration that its end time is far in the future, or less likely
* the current sampling instant is far in the past (because maybe no events have been reported lately.)

We now guard against both of these conditions:
* we cap the end time of an event at current time to prevent observing future-ending events, and
* if the current sampling instant is far in the past, we first fast-forward it to recent past before generating the next Poisson process element.

### Motivation
We observed that the loop was executing many times in some corner cases for customers. We presume it might've been GC events with erroneously long duration sent to it. 

### Additional Notes
Jira: [PROF-12199]

[PROF-12199]: https://datadoghq.atlassian.net/browse/PROF-12199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dd-trace/src/profiling/profilers/events.js
```diff
@@ -339,35 +342,12 @@ class CompositeEventSource {
 }
 
 function createPossionProcessSamplingFilter (samplingIntervalMillis) {
```
nsavoire: nit: 
```suggestion
function createPoissonProcessSamplingFilter (samplingIntervalMillis) {
```

nsavoire (dismissed):
Nice having unit tests !

packages/dd-trace/src/profiling/profilers/events.js
szegedi: My fingers' muscle memory will apparently never learn this, [exhibit A](https://github.com/DataDog/dd-trace-js/pull/6089/files#diff-7b72eb3a4fd3dbce3d567ab882ca75d313c047a6ff49e202adb4fc71f470c397L341-R369).

nsavoire: You forgot to rename the call site !

PR #6176, opened 2025-07-25
use datadog sca for vulnerability detection

rochdev:
### What does this PR do?

Remove audit workflow in favour of Datadog SCA.

### Motivation

This has several benefits:

- Datadog SCA is a lot more powerful:
    - It collects its data from more sources than a simple `npm audit` and it keeps historical data.
    - It can detect non-npm vulnerabilities.
    - It can also be used for other types of security and code quality issues.
    - It supports alerting us immediately when the vulnerability is detected without waiting for our CI to run.
- Stop blocking PRs for unrelated changes that also touch `package.json` or `yarn.lock`.
    - This is especially true for vulnerabilities that don't have a release with a fix yet.
- Stop broadcasting vulnerabilities publicly while we work on fixing them.
- Allow ignoring specific vulnerabilities, for example if they are known to not affect us and we cannot update because of Node version support.

I kept the workflow only for Dependabot branches, because those branches should auto-merge whenever possible, and without human interaction we need some guardrail in place to block on detection.

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -2,14 +2,8 @@ name: Audit
 
 on:
   pull_request:
-    paths:
-      - yarn.lock
-  push:
-    branches: [master]
-    paths:
-      - yarn.lock
-  schedule:
-    - cron: 0 4 * * *
+    branches:
+      - dependabot/*
 
 concurrency:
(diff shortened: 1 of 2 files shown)


simon-id (commented):
i see you're removing npm audit, but i don't see where you're adding SCA ?

PR #6189, opened 2025-07-28
Migrating Postgres from AsyncResource to Diagnostics Channel - runStores

crysmags:
### What does this PR do?

Migrate Postgres from AsyncResource to Diagnostics Channel - runStores

### Motivation
Async Resource no longer works as expected in Node 24.

### DIFF
--- a/packages/datadog-instrumentations/src/pg.js
+++ b/packages/datadog-instrumentations/src/pg.js
@@ -60,16 +60,15 @@ function wrapQuery (query) {
       abortController,
       stream
     }
-    return startCh.runStores(ctx, () => {
-      const finish = (error, res) => {
-        if (error) {
-          ctx.error = error
-          errorCh.publish(ctx)
-        }
-        ctx.result = res?.rows
-        return finishCh.publish(ctx)
+    const finish = (error, res) => {
+      if (error) {
+        ctx.error = error
+        errorCh.publish(ctx)
       }
-
+      ctx.result = res?.rows
+      return finishCh.publish(ctx)
+    }
+    return startCh.runStores(ctx, () => {
       if (abortController.signal.aborted) {
         const error = abortController.signal.reason || new Error('Aborted')
 
--- a/packages/datadog-instrumentations/src/pg.js
+++ b/packages/datadog-instrumentations/src/pg.js
@@ -54,7 +54,6 @@ function wrapQuery (query) {
     }
     const abortController = new AbortController()
     const ctx = {
-      pgQuery,
       params: this.connectionParameters,
       query: textPropObj,
       processId,
@@ -68,7 +67,7 @@ function wrapQuery (query) {
           errorCh.publish(ctx)
         }
         ctx.result = res?.rows
-        return finishCh.publish(ctx)
+        return finishCh.publish({ result: res?.rows })
       }
 
       if (abortController.signal.aborted) {
@@ -143,7 +142,9 @@ function wrapQuery (query) {
     })
   }
 }
-
+const finish = () => {
+  finishPoolQueryCh.publish()
+}
 function wrapPoolQuery (query) {
   return function () {
     if (!startPoolQueryCh.hasSubscribers) {
@@ -153,13 +154,9 @@ function wrapPoolQuery (query) {
     const pgQuery = arguments[0] !== null && typeof arguments[0] === 'object' ? arguments[0] : { text: arguments[0] }
     const abortController = new AbortController()
 
-    const ctx = { pgQuery, abortController }
+    const ctx = { query: pgQuery, abortController }
 
     return startPoolQueryCh.runStores(ctx, () => {
-      const finish = () => {
-        finishPoolQueryCh.publish(ctx)
-      }
-
       const cb = arguments[arguments.length - 1]
 
       if (abortController.signal.aborted) {
--- a/packages/datadog-instrumentations/src/pg.js
+++ b/packages/datadog-instrumentations/src/pg.js
@@ -143,8 +143,8 @@ function wrapQuery (query) {
     })
   }
 }
-const finish = () => {
-  finishPoolQueryCh.publish()
+const finish = (ctx) => {
+  finishPoolQueryCh.publish(ctx)
 }
 function wrapPoolQuery (query) {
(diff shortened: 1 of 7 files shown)


packages/datadog-instrumentations/src/pg.js
```diff
@@ -55,24 +52,24 @@ function wrapQuery (query) {
...
-      const finish = asyncResource.bind(function (error, res) {
+    const abortController = new AbortController()
+    const ctx = {
+      pgQuery,
+      params: this.connectionParameters,
+      query: textPropObj,
+      processId,
+      abortController,
+      stream
+    }
```
wconti27:
```suggestion
    const ctx = {
      pgQuery,
      params: this.connectionParameters,
      query: textPropObj,
      processId,
      abortController: new AbortController(),
      stream
    }
```
Nit, but you can also instantiate the abort controller in the ctx const.

packages/datadog-instrumentations/src/pg.js
```diff
@@ -152,21 +150,15 @@ function wrapPoolQuery (query) {
...
+    const abortController = new AbortController()
 
-    return asyncResource.runInAsyncScope(() => {
-      const abortController = new AbortController()
-
-      startPoolQueryCh.publish({
-        query: pgQuery,
-        abortController
-      })
+    const ctx = { pgQuery, abortController }
```
wconti27:
```suggestion
    const ctx = { pgQuery, abortController: new AbortController()  }
```
Same nit here.

wconti27 (commented):
LGTM so far, maybe Appsec can help with the failures on their side.

packages/datadog-instrumentations/src/pg.js
crysmags: @wconti27 AbortController is used outside of the context so I would then have to call it from the context, is that better?

packages/datadog-instrumentations/src/pg.js
```diff
@@ -139,35 +136,28 @@ function wrapQuery (query) {
...
+      } catch (error) {
+        ctx.error = error
+        errorCh.publish(ctx)
       }
     })
   }
 }
-
+const finish = () => {
+  finishPoolQueryCh.publish()
```
rochdev: This needs to receive the context, otherwise it may not be able to figure out the current store.

packages/dd-trace/src/appsec/iast/analyzers/sql-injection-analyzer.js
```diff
@@ -16,15 +16,18 @@ class SqlInjectionAnalyzer extends StoredInjectionAnalyzer {
   onConfigure () {
     this.addSub('apm:mysql:query:start', ({ sql }) => this.analyze(sql, undefined, 'MYSQL'))
     this.addSub('datadog:mysql2:outerquery:start', ({ sql }) => this.analyze(sql, undefined, 'MYSQL'))
-    this.addSub('apm:pg:query:start', ({ query }) => this.analyze(query.text, undefined, 'POSTGRES'))
+    this.addSub(
+      'apm:pg:query:start',
+      ({ originalText, query }) => this.analyze(originalText || query.text, undefined, 'POSTGRES')
```
rochdev, line 21: Why is the query not enough here? Even if it gets altered it's only an added comment, that should still work with AppSec?

Also, what is the scenario where we have `query.text` but not `originalText`? If that scenario exists, wouldn't we have the same issue that this is supposed to solve?

cc @uurien

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
```diff
@@ -139,12 +139,12 @@ class TaintTrackingPlugin extends SourceIastPlugin {
   addDatabaseSubscriptions () {
     this.addSub(
       { channelName: 'datadog:sequelize:query:finish', tag: SQL_ROW_VALUE },
-      ({ result }) => this._taintDatabaseResult(result, 'sequelize')
+      ({ result }) => this._taintDatabaseResult(result, 'sequelize', getIastContext(storage('legacy').getStore()))
```
rochdev, line 142: I think we should avoid binding the legacy store from AppSec. The long-term goal is that each product can have their own store and their own plugins with their own bindings, and altering the legacy store bindings in more than one place will make it much more difficult to refactor.

rochdev (changes requested):

packages/dd-trace/src/appsec/iast/analyzers/sql-injection-analyzer.js
uurien: > Why is the query not enough here? Even if it gets altered it's only an added comment, that should still work with AppSec?

In IAST we don't care about the content of the query, we need the right string instance. If we modify a string in the tracer, the taint tracking mechanism lost the new string we can't say that the string is coming from the request. In the moment that plugin modifies the string, the detection stops.

> Also, what is the scenario where we have query.text but not originalText? If that scenario exists, wouldn't we have the same issue that this is supposed to solve?

I wanted to do the fewest changes as possible in pg.js, and there the `originalText` is filled into [this "if"](https://github.com/DataDog/dd-trace-js/pull/6189/files#diff-d568b60c6ac1af79365d9cf6b40668f51f62b562a1841e2acd9b2b4af1cce1caR48). It means that `originalText` could be empty, in that cases we are using `query.text`, it could make sense because IIUC in that cases the plugin is not modifying the original query.

packages/dd-trace/src/appsec/iast/taint-tracking/plugin.js
crysmags: Hey @rochdev this is for Sequelize which we are not changing in this PR, previously `_taintDatabaseResult` had `getIastContext(storage('legacy').getStore())` set in the parameters, I think this is just to keep sequelize the same since we are now passing the `currentStore` to pg in line 147 `({ result, currentStore }) => this._taintDatabaseResult(result, 'pg', getIastContext(currentStore))`

packages/datadog-instrumentations/src/pg.js
crysmags: Done 😃

PR #6190, opened 2025-07-28
add manual checkpoint parameter to consume DSM API

michael-zhao459:
### What does this PR do?

Adds a manualCheckpoint parameter in the public dsm set_consume_checkpoint(), this allows one to turn off the manual_checkpoint:true tag if use of function is not for explicit purpose of manually instrumenting (default remains as manual instrumentation).

### Motivation

Wanting to use this public API in a non manual way in another repo

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/test/datastreams/data_streams_checkpointer.spec.js
+++ b/packages/dd-trace/test/datastreams/data_streams_checkpointer.spec.js
@@ -70,4 +70,21 @@ describe('data streams checkpointer manual api', () => {
 
     expect(DSM_CONTEXT_HEADER in headers).to.equal(true)
   })
+
+  it('should test manual checkpoint behavior', function () {
+    const headers = {}
+    const mockSetCheckpoint = sinon.stub().returns({ hash: Buffer.from([1, 2, 3, 4]) })
+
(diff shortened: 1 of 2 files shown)


packages/dd-trace/test/datastreams/data_streams_checkpointer.spec.js
```diff
@@ -70,4 +70,21 @@ describe('data streams checkpointer manual api', () => {
 
     expect(DSM_CONTEXT_HEADER in headers).to.equal(true)
   })
+
+  it('should test manual checkpoint behavior', function () {
```
piochelepiotr: naming issue here, it's already a test, we should not have `test` in the name.

Maybe:
`it should set manual checkpoint when setConsumeCheckpoint is called without additional parameters`

and add another one:
`it should set an automatic checkpoint when setConsumeCheckpoint is called with manual_checkpoint:false`

piochelepiotr (approved):

PR #6192, opened 2025-07-29
add error message when latest test version cannot be found

rochdev:
### What does this PR do?

Add error message when latest test version cannot be found.

### Motivation

Right now it just sets the value to `undefined` which results in a failure and a cryptic error message if the latest version isn't set.

### DIFF
--- a/scripts/install_plugin_modules.js
+++ b/scripts/install_plugin_modules.js
@@ -112,6 +112,12 @@ async function assertFolder (name, version) {
  * @param {boolean} external
  */
 async function assertPackage (name, version, dependencyVersionRange, external) {
+  if (!latests[name]) {
+    throw new Error(
+      `Latest version for '${name}' needs to be defined in 'packages/dd-trace/test/plugins/versions/package.json'.`
+    )
+  }
+
   const alreadyCapped = dependencyVersionRange.includes('-')
   const cappedVersionRange = external || alreadyCapped
     ? dependencyVersionRange


simon-id (approved):

packages/dd-trace/test/plugins/versions/index.js
```diff
@@ -0,0 +1,26 @@
...
+
+const latests = require('./package.json').dependencies
+
+/**
+ * @param {string} name
+ * @param {string} range
+ */
+function getCappedRange (name, range) {
+  const alreadyCapped = range.split('||').every(r => {
+    return r.includes('-') || r.includes('<') || /^=?\d+\.\d+\.\d+$/.test(r)
```
simon-id: Would that work with ranges like `'4'` or `'4.1'` ?

scripts/install_plugin_modules.js
```diff
@@ -112,12 +112,9 @@ async function assertFolder (name, version) {
...
  */
 async function assertPackage (name, version, dependencyVersionRange, external) {
-  const alreadyCapped = dependencyVersionRange.includes('-')
-  const cappedVersionRange = external || alreadyCapped
-    ? dependencyVersionRange
-    : `${dependencyVersionRange} <=${latests[name]}`
-
-  const dependencies = { [name]: cappedVersionRange }
+  const dependencies = {
+    [name]: getCappedRange(name, dependencyVersionRange)
```
simon-id, line 116: no need to check `external` anymore ?

rochdev: New versions of externals can also fail CI, this was an oversight on my part originally.

packages/dd-trace/test/plugins/versions/index.js
rochdev: `4` isn't capped and `4.1` isn't capped either. However, I just realized that actually even `-` doesn't guarantee a stable range as it could be capped to a future version, so only exact versions should really be considered already capped. I will make the change.

rochdev: Ok I fixed the logic to handle everything that is not an exact range.

simon-id (approved):

PR #6203, opened 2025-07-30
feat: add injection metadata fields to telemetry forwarder

sydney-tung:
### What does this PR do?

This change adds additional fields to the telemetry forwarder used during injection. These fields will be used to add additional context that will be surfaced to the end user in the following Figma:
Screenshot 2025-07-03 at 10 39 35 AM
See [Injection Metadata](https://docs.google.com/document/d/1wydXj2fW0V0jaWU6W9pIbipT_AXD15tmbTtOPXFBEy4/edit?tab=t.f3vpy615c2dp) for more details.

### Motivation

Adding addition information for telemetry forwarder during injection.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


integration-tests/helpers/index.js
```diff
@@ -113,7 +114,10 @@ function assertTelemetryPoints (pid, msgs, expectedTelemetryPoints) {
       runtime_name: 'nodejs',
       runtime_version: process.versions.node,
       tracer_version: require('../../package.json').version,
-      pid: Number(pid)
+      pid: Number(pid),
+      result: telemetryModule.result,
```
tlhunter: It seems weird that these three new fields are camelCase while all the existing fields are snake_case. Does the backend expect this disparity?

packages/dd-trace/src/guardrails/telemetry.js
```diff
@@ -22,7 +31,10 @@ var metadata = {
   runtime_name: 'nodejs',
   runtime_version: process.versions.node,
   tracer_version: tracerVersion,
-  pid: process.pid
+  pid: process.pid,
+  result: module.exports.result,
```
tlhunter: It looks like all that ultimately matters with this data is what is inside of this `metadata` object and not what the module itself is exporting as this object is serialized and later written to stdout (stdin?).

The tests appear to be written to run assertions against what is exported from this file. And telemetry.js is written to update the exported values as well. And while it's true that on these three lines the value from the exports is copied to metadata this copying will only happen once after the above `!process.env.DD_INJECTION_ENABLED` and `typeof telemetryForwarderPath !== 'string' || !fs.existsSync(telemetryForwarderPath)` checks are complete. The four sets of assignments that follow these three lines won't update `metadata`.

I would recommend not modifying exports throughout this file with `module.exports.result = 'error'`, etc, and instead write to `metadata.result` instead.

Since none of the three preceding sets of code ever change the values (they're always 'unknown' no matter what) you could simply replace the three lines here with `result: 'unknown',`, etc.

I think your `mockProc` tests seem to be properly triggering the error conditions so instead of asserting against the export values you'll want to read the JSON from stdin, parse it, then assert the right values that way.

tlhunter (changes requested):

packages/dd-trace/test/guardrails/telemetry.spec.js
```diff
@@ -16,7 +18,7 @@ describe('sendTelemetry', () => {
 
   beforeEach(() => {
     cleanup = telemetryForwarder()
-    sendTelemetry = proxyquire('../src/guardrails/telemetry', {})
```
tlhunter, line 19: This is pretty weird. It looks like the old proxyquire call here was wrong as that file doesn't exist relative to this file, and that you fixed it. And it's not like the location of this test file moved.

Did you find that running this test file would throw errors and that when you fixed the path it started working? If so we might have a bug in our test suite where we're not properly executing this test file...

sydney-tung: oh wait yeah it only worked after I changed the path...

tlhunter: Alright I'm super confused then.

Here are the CI results for a random PR that I found that doesn't contain your changes and I can see that the guardrails tests are running:
https://github.com/DataDog/dd-trace-js/actions/runs/16787347785/job/47541117623?pr=6237

```
> dd-trace@6.0.0-pre test:trace:guardrails
> mocha -r "packages/dd-trace/test/setup/mocha.js" "packages/dd-trace/test/guardrails/**/*.spec.js"



  sendTelemetry
    ✔ should send telemetry (114ms)
    no duplicates
      ✔ should not send `abort` more than once in the same call (105ms)
      ✔ should not send `abort` more than once in a different call (105ms)
      ✔ should not send `abort.integration` more than once if tags are the same in the same call (104ms)
      ✔ should not send `abort.integration` more than once if tags are the same in a different call (105ms)


  5 passing (844ms)

--------------|---------|----------|---------|---------|-------------------
File          | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
--------------|---------|----------|---------|---------|-------------------
All files     |   87.03 |       65 |   35.71 |   88.23 |                   
 log.js       |     100 |    35.71 |       0 |     100 | 20-29             
 telemetry.js |   84.09 |    79.16 |   44.44 |   85.36 | 11,16,53,66,70,74 
 util.js      |     100 |      100 |     100 |     100 |                   
--------------|---------|----------|---------|---------|-------------------
Done in 1.66s.
```

And here are is the test results for your PR:
https://github.com/DataDog/dd-trace-js/actions/runs/16756190382/job/47439327513?pr=6203

```
> dd-trace@6.0.0-pre test:trace:guardrails
> mocha -r "packages/dd-trace/test/setup/mocha.js" "packages/dd-trace/test/guardrails/**/*.spec.js"



  sendTelemetry
    ✔ should send telemetry (108ms)
    no duplicates
      ✔ should not send `abort` more than once in the same call (105ms)
      ✔ should not send `abort` more than once in a different call (105ms)
      ✔ should not send `abort.integration` more than once if tags are the same in the same call (104ms)
      ✔ should not send `abort.integration` more than once if tags are the same in a different call (104ms)
    Error scenarios and metadata
      ✔ should set error metadata when telemetry forwarder fails to spawn
      ✔ should set error metadata when telemetry forwarder exits with non-zero code
      ✔ should set error metadata when writing to telemetry forwarder fails
      ✔ should set success metadata when telemetry forwarder exits successfully


  9 passing (852ms)

--------------|---------|----------|---------|---------|-------------------
File          | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
--------------|---------|----------|---------|---------|-------------------
All files     |   93.93 |     67.5 |   57.14 |   95.23 |                   
 log.js       |     100 |    35.71 |      25 |     100 | 20-29             
 telemetry.js |   92.85 |    83.33 |   66.66 |   94.33 | 11,16,56          
 util.js      |     100 |      100 |     100 |     100 |                   
--------------|---------|----------|---------|---------|-------------------
Done in 1.64s.
```

Based on the location of the files I believe your change is correct:

```
dd-trace-js/packages/dd-trace/test/guardrails/ $ node
Welcome to Node.js v22.18.0.
Type ".help" for more information.
> let proxyquire = require('proxyquire')
undefined
> proxyquire('../src/guardrails/telemetry', {})
Uncaught Error: Cannot find module '../src/guardrails/telemetry'
Require stack:
- <repl>
    at Function._resolveFilename (node:internal/modules/cjs/loader:1365:15)
    at Proxyquire._disableModuleCache (/Users/thomas.hunter/Projects/dd-trace-js/node_modules/proxyquire/lib/proxyquire.js:248:19)
    at Proxyquire._disableCache (/Users/thomas.hunter/Projects/dd-trace-js/node_modules/proxyquire/lib/proxyquire.js:216:15)
    at Proxyquire._withoutCache (/Users/thomas.hunter/Projects/dd-trace-js/node_modules/proxyquire/lib/proxyquire.js:172:27)
    at Proxyquire.load (/Users/thomas.hunter/Projects/dd-trace-js/node_modules/proxyquire/lib/proxyquire.js:136:15) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '<repl>' ]
}
> proxyquire('../../src/guardrails/telemetry', {})
[Function: noop]
```

I see that @watson introduced this test file in commit f97d2cde99bd6a3e36c55bf8c8bc3c8831bd94f3 so I'd like him to take a look. There might be some shenanigans in CI that allow the old code to work?

watson: Just back from PTO. I assume you got this all figured out?

PR #6215, opened 2025-08-01
remove async storage from fastify instrumentation

wconti27:
### What does this PR do?
removes async resource storage usage from fastify instrumentation for Node24 compatibility

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/fastify.js
+++ b/packages/datadog-instrumentations/src/fastify.js
@@ -16,8 +16,7 @@ const finishSetHeaderCh = channel('datadog:fastify:set-header:finish')
 // context management channels
 const preParsingCh = channel('datadog:fastify:pre-parsing:start')
 const preValidationCh = channel('datadog:fastify:pre-validation:start')
-const addHookStartCh = channel('datadog:fastify:add-hook:start')
-const addHookFinishCh = channel('datadog:fastify:add-hook:finish')
+const callbackFinishCh = channel('datadog:fastify:callback:execute')
 
 const parsingContexts = new WeakMap()
 const cookiesPublished = new WeakSet()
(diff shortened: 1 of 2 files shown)


packages/datadog-instrumentations/src/fastify.js
```diff
@@ -51,57 +57,58 @@ function wrapAddHook (addHook) {
...
       try {
         // done callback is always the last argument
         const doneCallback = arguments[arguments.length - 1]
 
         if (typeof doneCallback === 'function') {
-          arguments[arguments.length - 1] = function (err) {
-            publishError(err, req)
-
-            const hasCookies = request.cookies && Object.keys(request.cookies).length > 0
+          arguments[arguments.length - 1] = addHookStartCh.runStores(ctx, () => {
```
rochdev: This hook is not needed as the previous code created the resource right before `doneCallback` was called, so basically the `addHookFinishCh` is enough (although I would also argue the name is confusing as `addHook` is called when the app starts, and the `doneCallback` is called on every request, so it's not happening at the same time as adding a hook at all).

rochdev (changes requested):

rochdev:
Why is this PR marked as `semver-minor` instead of `semver-patch`?

packages/datadog-instrumentations/src/fastify.js
wconti27: Ive removed the start `runStores` call, and changed the naming to be `callback:exec` for the channel. Should be good 👍

wconti27:
> Why is this PR marked as `semver-minor` instead of `semver-patch`?

fixed.

PR #6217, opened 2025-08-01
remove async storage from oracledb

wconti27:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/oracledb.js
+++ b/packages/datadog-instrumentations/src/oracledb.js
@@ -2,8 +2,7 @@
 
 const {
   channel,
-  addHook,
-  AsyncResource
+  addHook
 } = require('./helpers/instrument')
 const shimmer = require('../../datadog-shimmer')
 
@@ -14,15 +13,16 @@ const startChannel = channel('apm:oracledb:query:start')
 const errorChannel = channel('apm:oracledb:query:error')
 const finishChannel = channel('apm:oracledb:query:finish')
 
-function finish (err) {
-  if (err) {
-    errorChannel.publish(err)
+function finish (ctx) {
+  if (ctx.error) {
+    errorChannel.publish(ctx.error)
   }
-  finishChannel.publish()
+  finishChannel.publish(ctx)
 }
 
 addHook({ name: 'oracledb', versions: ['>=5'] }, oracledb => {
   shimmer.wrap(oracledb.Connection.prototype, 'execute', execute => {
+    const ctx = {}
     return function wrappedExecute (dbQuery, ...args) {
       if (!startChannel.hasSubscribers) {
         return execute.apply(this, arguments)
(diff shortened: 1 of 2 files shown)


packages/datadog-instrumentations/src/oracledb.js
```diff
@@ -14,73 +13,78 @@ const startChannel = channel('apm:oracledb:query:start')
...
       }
 
       if (arguments.length && typeof arguments[arguments.length - 1] === 'function') {
         const cb = arguments[arguments.length - 1]
-        const outerAr = new AsyncResource('apm:oracledb:outer-scope')
         arguments[arguments.length - 1] = shimmer.wrapFunction(cb, cb => function wrappedCb (err, result) {
-          finish(err)
-          return outerAr.runInAsyncScope(() => cb.apply(this, arguments))
+          if (err) {
+            errorChannel.publish(err)
```
rochdev: This should be called with `ctx`.

packages/datadog-instrumentations/src/oracledb.js
```diff
@@ -14,73 +13,78 @@ const startChannel = channel('apm:oracledb:query:start')
 const errorChannel = channel('apm:oracledb:query:error')
 const finishChannel = channel('apm:oracledb:query:finish')
 
-function finish (err) {
-  if (err) {
-    errorChannel.publish(err)
+function finish (ctx) {
+  if (ctx.error) {
+    errorChannel.publish(ctx.error)
```
rochdev: Same as below.

packages/datadog-instrumentations/src/oracledb.js
```diff
@@ -14,73 +13,78 @@ const startChannel = channel('apm:oracledb:query:start')
...
+      }
 
-        startChannel.publish({
-          query: dbQuery,
-          connAttrs,
-          dbInstance,
-          port,
-          hostname,
-        })
+      ctx.dbInstance = dbInstance
```
rochdev: `ctx` could be defined here to avoid all the assignments (works because it's still defined before the above function runs even if it's later in code)

packages/datadog-instrumentations/src/oracledb.js
wconti27: fixed.

packages/datadog-instrumentations/src/oracledb.js
wconti27: fixed

packages/datadog-instrumentations/src/oracledb.js
wconti27: fixed.

PR #6228, opened 2025-08-05
Fix service name in metadata

vitkyrka:
### What does this PR do?

Use the correct field which contains the service name when filling the tracer metadata.

### Motivation

Noticed that the field was not set correctly.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/tracer_metadata.js
+++ b/packages/dd-trace/src/tracer_metadata.js
@@ -14,7 +14,7 @@ function storeConfig (config) {
     config.tags['runtime-id'],
     tracerVersion,
     config.hostname,
-    config.server || null,
+    config.service || null,
     config.env || null,
     config.version || null
   )


vitkyrka:
> @BridgeAR While the process is running, we can open and decode the file content to assess what's inside. In fact, that's exactly what the [system test](https://github.com/DataDog/system-tests/blob/main/tests/parametric/test_process_discovery.py) is doing. Does that help clarify your question?

The test was still disabled for nodejs. After enabling it and fixing a bug in the test itself (https://github.com/DataDog/system-tests/pull/5037), it passes with a local build of the tracer with this fix (and fails without it, as expected).

PR #6229, opened 2025-08-05
Fix flaky test

szegedi:
### What does this PR do?
Fixes a flaky test

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,7 @@
   ],
   "dependencies": {
     "@datadog/libdatadog": "0.7.0",
-    "@datadog/native-appsec": "10.0.1",
+    "@datadog/native-appsec": "10.1.0",
     "@datadog/native-iast-taint-tracking": "4.0.0",
     "@datadog/native-metrics": "3.1.1",
     "@datadog/pprof": "5.9.0",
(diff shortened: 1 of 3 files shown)


packages/dd-trace/test/profiling/profilers/poisson.spec.js
```diff
@@ -65,7 +65,8 @@ describe('PoissonProcessSamplingFilter', () => {
       resetInterval: 20,
       now: constantNow
     })
-    const event = { startTime: 0, duration: 100 }
+    assert.strictEqual(callCount, 1)
+    const event = { startTime: 0, duration: Number.POSITIVE_INFINITY }
     assert.doesNotThrow(() => filter.filter(event))
```
BridgeAR, line 70: Independent nit: 

```suggestion
    filter.filter(event)
```

The doesNotThrow API is recommended against, due to not providing any benefit. It just adds a wrapper around the error. In Node.js we have a linter to prevent it's usage. I guess we could do the same here.

simon-id (approved):

packages/dd-trace/test/profiling/profilers/poisson.spec.js
szegedi: It is functionally indeed unnecessary, but I like it as a documentation instrument for the reader of the code. It tells that the writer of the test knows there is a code path that throws an exception and the test is specifically written to ensure the tested scenario avoids that particular code path. 

If I were to remove the `doesNotThrow` call I'd definitely add a source code comment instead.

PR #6233, opened 2025-08-05
[SVLS-7168] Create Synthetic Span for GCP Push Subscriptions (including Cloud Events)

nina9753:
### What does this PR do?
Simplified and optimized Datadog tracing instrumentation for Google Cloud Pub/Sub push messages and Cloud Events across all Node.js web frameworks.

### Motivation
A Synthetic span for the HTTP push post to the Cloud Run service from a pub/sub topic 

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/body-parser.js
+++ b/packages/datadog-instrumentations/src/body-parser.js
@@ -26,6 +26,11 @@ addHook({
   versions: ['>=1.4.0 <1.20.0']
 }, read => {
   return shimmer.wrapFunction(read, read => function (req, res, next) {
+    // Skip body parsing if PubSub/Cloud Event middleware already parsed it
+    if (req._pubsubBodyParsed) {
+      console.log('[DD-TRACE] Skipping body-parser for PubSub/Cloud Event request - already parsed')
+      return next()
+    }
     const nextResource = new AsyncResource('bound-anonymous-fn')
     arguments[2] = nextResource.bind(publishRequestBodyAndNext(req, res, next))
     return read.apply(this, arguments)
@@ -38,6 +43,11 @@ addHook({
   versions: ['>=1.20.0']
 }, read => {
   return shimmer.wrapFunction(read, read => function (req, res, next) {
+    // Skip body parsing if PubSub/Cloud Event middleware already parsed it
+    if (req._pubsubBodyParsed) {
+      console.log('[DD-TRACE] Skipping body-parser for PubSub/Cloud Event request - already parsed')
+      return next()
+    }
     arguments[2] = publishRequestBodyAndNext(req, res, next)
     return read.apply(this, arguments)
   })
--- a/packages/datadog-instrumentations/src/http/server.js
+++ b/packages/datadog-instrumentations/src/http/server.js
@@ -19,6 +19,207 @@ const requestFinishedSet = new WeakSet()
 const httpNames = ['http', 'node:http']
 const httpsNames = ['https', 'node:https']
 
+function handlePubSubOrCloudEvent(req, res, emit, server, originalArgs) {
+  const isCloudEvent = req.headers['content-type']?.includes('application/cloudevents+json') ||
+    req.headers['ce-specversion']
+  const eventType = isCloudEvent ? 'Cloud Event' : 'PubSub push'
+  console.log(`[DD-TRACE] HTTP server handling ${eventType} request (framework-agnostic)`)
+
+  // Get tracer from global reference (avoids circular dependencies)
+  const tracer = global._ddtrace
+  if (!tracer) {
+    console.warn('[DD-TRACE] Tracer not available, skipping PubSub handling')
+    return emit.apply(server, originalArgs)
+  }
+
+  // Collect raw body for PubSub message parsing with error handling
+  const chunks = []
+  const MAX_BODY_SIZE = 10 * 1024 * 1024 // 10MB limit for large Pub/Sub payloads
+  let bodySize = 0
+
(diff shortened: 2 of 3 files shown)


packages/datadog-instrumentations/src/http/server.js
```diff
@@ -19,6 +19,247 @@ const requestFinishedSet = new WeakSet()
...
+  const subscription = json.subscription
+  const attrs = message?.attributes || {}
+  return { message, subscription, attrs }
+}
+
+function createCloudEventSpan (tracer, parent, topicName, projectId, subscription, message, attrs, req) {
+  const spanTags = {
+    component: 'google-cloud-pubsub',
+    'span.kind': 'consumer',
+    'gcloud.project_id': projectId || 'unknown',
```
apiarian-datadog: for my eduction: is this `|| unknown` a standard pattern for span tags? how do we decide between setting the tag to "unknown" vs not setting the tag at all?

packages/datadog-instrumentations/src/http/server.js
```diff
@@ -19,6 +19,247 @@ const requestFinishedSet = new WeakSet()
 const httpNames = ['http', 'node:http']
 const httpsNames = ['https', 'node:https']
 
+function parseCloudEventMessage (json, req) {
+  // Eventarc only uses Binary Content Mode with ce-specversion header
+  // Payload structure: {"message": {...}, "subscription": "..."}
+  const message = json.message || json
+  const attrs = { ...message?.attributes }
+  const subscription = json.subscription || req.headers['ce-subscription'] || 'cloud-event-subscription'
```
nhulston: This `server.js file` is a generic HTTP server instrumentation, so we shouldn't be putting Pub/Sub or Eventarc-specific logic directly in this file. Can you move that logic into a separate file/module?

Maybe you can do something like this (this code was suggested by ChatGPT):
```javascript
addHook({ name: 'http' }, http => {
  shimmer.wrap(http.Server.prototype, 'emit', wrapEmit)

  if (isGCPEnvironment()) {
    require('./pubsub').patch(http)
  }

  return http
})
```

nhulston (changes requested):

packages/datadog-instrumentations/src/http/server.js
nina9753: I didn't do this approach because then both files are hooking the same module:

http/server.js: `addHook({ name: httpNames }, http => { shimmer.wrap(http.Server.prototype, 'emit', wrapEmit) })`
gcp-pubsub-push.js: `addHook({ name: ['http', 'node:http'] }, (http) => { shimmer.wrap(http.Server.prototype, 'emit', wrapEmitForGcp) })`

Both hooks would execute and try to wrap emit on the same prototype. The second wrap overwrites the first and replaces the method, so only the last loaded hook would work. Instead of creating another `addHook({ name: 'http' }, http => {` I can move the logic to another file that will still be referenced in `server.js` addHook this will allow the `wrapEmitForGcp` to extend the other shimmer `wrapEmit`

nhulston: That makes sense. We could still add the hook in `http/server.js`, but delegate as much GCP specific logic as we can to a helper function in a separate file. Might be worth reaching out to #guild-dd-node-js to see if someone there could help with best practices.

nina9753: I reached out in slack (apm-serverless too) for a review and `http/server.js` has a hook already so i just added my code to extend that hook

packages/datadog-instrumentations/src/body-parser.js
```diff
@@ -26,6 +26,10 @@ addHook({
   versions: ['>=1.4.0 <1.20.0']
 }, read => {
   return shimmer.wrapFunction(read, read => function (req, res, next) {
+    // Skip body parsing if PubSub/Cloud Event middleware already parsed it
+    if (req._pubsubBodyParsed) {
```
rochdev: Does this really need to know about PubSub specifically? Wouldn't generically checking if `req.body` exists be enough?

packages/datadog-instrumentations/src/http/server.js
rochdev: The way things typically work is that each integration is split into 2 components:

1) The instrumentation, which is responsible for patching and collecting data, which it can then publish on channels. This part should be generic and contain no code specific to Datadog. I like to think of this as something that should be exposed by the library itself so that we wouldn't have to patch it in the first place, and Datadog specific code would never exist in the library itself (or a built-in module in this case)
2) Plugins, which can then subscribe to the channels exposed by the instrumentation and handle Datadog specific things.

Usually there is only a single instrumentation since its responsibility is generic, and there can be multiple plugins (for example per product, or sometimes hierarchical for more complex integrations). So here, any information not already collected by the `http` instrumentation should be added but only that. Any actual logic should be moved to a new plugin that handles everything else.

packages/datadog-instrumentations/src/http/server.js
rochdev: We don't usually add a fallback when a value is not available. Leaving it undefined will simply not add the tag.

nina9753: I removed `unknown` from the tags

rochdev:
@nina9753 Should this be closed if it's superseded by #6260?

PR #6237, opened 2025-08-06
add supported versions file for integrations

rochdev:
### What does this PR do?

Add supported versions file for integrations.

### Motivation

This makes it easier to have a global view of supported versions when we want to deprecate older versions for major, as well as ensuring that as they are bumped the code doesn't still support an old version. It will also make it easier to report this data to telemetry or to generate documentation automatically.


BridgeAR:
We now have a json and csv file in the repository that does exactly this. Closing as such

PR #6248, opened 2025-08-08
Support Forwarded header for IP extraction

uurien:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/plugins/util/ip_extractor.js
+++ b/packages/dd-trace/src/plugins/util/ip_extractor.js
@@ -3,11 +3,14 @@
 const { BlockList } = require('net')
 const net = require('net')
 
+const FORWARED_HEADER_NAME = 'forwarded'
+
 const ipHeaderList = [
   'x-forwarded-for',
   'x-real-ip',
   'true-client-ip',
   'x-client-ip',
+  FORWARED_HEADER_NAME,
   'forwarded-for',
   'x-cluster-client-ip',
   'fastly-client-ip',
@@ -49,7 +52,8 @@ function extractIp (config, req) {
   let firstPrivateIp
   if (headers) {
     for (const ipHeaderName of ipHeaderList) {
-      const firstIp = findFirstIp(headers[ipHeaderName])
+      const header = headers[ipHeaderName]
+      const firstIp = ipHeaderName === FORWARED_HEADER_NAME ? findFirstIpForwardedFormat(header) : findFirstIp(header)
 
       if (firstIp.public) {
         return firstIp.public
@@ -60,6 +64,23 @@ function extractIp (config, req) {
   }
 
   return firstPrivateIp || req.socket?.remoteAddress
+}
+
+function isPublicIp (chunk, type) {
+  return !privateIPMatcher.check(chunk, type === 6 ? 'ipv6' : 'ipv4')
+}
+
+function extractPublicOrPrivateIp (ip) {
+  const type = net.isIP(ip)
+  if (!type) return {}
+
+  if (isPublicIp(ip, type)) {
+    // it's public, return it immediately
+    return { public: true }
+  }
+
+  // it's private, only save the first one found
+  return { private: true }
 }
 
 function findFirstIp (str) {
@@ -73,20 +94,48 @@ function findFirstIp (str) {
 
     // TODO: strip port and interface data ?
 
-    const type = net.isIP(chunk)
-    if (!type) continue
-
-    if (!privateIPMatcher.check(chunk, type === 6 ? 'ipv6' : 'ipv4')) {
-      // it's public, return it immediately
+    const currentIpInfo = extractPublicOrPrivateIp(chunk)
+    if (currentIpInfo.public) {
       result.public = chunk
       break
     }
 
-    // it's private, only save the first one found
-    if (!result.private) result.private = chunk
+    if (currentIpInfo.private && !result.private) result.private = chunk
   }
 
   return result
+}
+
+const forwardedForRegexp = /for="?\[?(([0-9]+\.)+[0-9]+|[0-9a-f:]*:[0-9a-f]*)/i
+const forwardedByRegexp = /by="?\[?(([0-9]+\.)+[0-9]+|[0-9a-f:]*:[0-9a-f]*)/i
+const forwardedRegexps = [forwardedForRegexp, forwardedByRegexp]
+
+function findFirstIpForwardedFormat (str) {
+  const result = {}
+  if (!str) return result
+
+  const splitted = str.split(',')
+
+  for (const part of splitted) {
+    const chunk = part.trim()
+
+    for (const regex of forwardedRegexps) {
+      const ip = regex.exec(chunk)?.[1]
+
+      const tempResult = extractPublicOrPrivateIp(ip)
+      if (tempResult.public) {
+        result.public = ip
+        break
+      }
+
+      if (tempResult.private && !result.private) result.private = ip
+    }
+
+    if (result.public) break
+  }
+
+  return result
+
 }
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/plugins/util/ip_extractor.js
```diff
@@ -62,6 +66,23 @@ function extractIp (config, req) {
...
+function isPublicIp (chunk, type) {
+  return !privateIPMatcher.check(chunk, type === 6 ? 'ipv6' : 'ipv4')
+}
+
+function extractPublicOrPrivateIp (ip) {
+  const type = net.isIP(ip)
+  if (!type) return {}
+
+  if (isPublicIp(ip, type)) {
+    // it's public, return it immediately
```
CarlesDD: I think the comments in this function come from `findFirstIp` method and they have no sense here.

uurien: You're right.  Failure on reactor 😅

uurien: Moved to the right places.

packages/dd-trace/src/plugins/util/ip_extractor.js
```diff
@@ -62,6 +66,21 @@ function extractIp (config, req) {
   return firstPrivateIp || req.socket?.remoteAddress
 }
 
+function isPublicIp (chunk, type) {
+  return !privateIPMatcher.check(chunk, type === 6 ? 'ipv6' : 'ipv4')
```
simon-id: it's called IP, in the calling function
```suggestion
function isPublicIp (ip, type) {
  return !privateIPMatcher.check(ip, type === 6 ? 'ipv6' : 'ipv4')
```

packages/dd-trace/src/plugins/util/ip_extractor.js
```diff
@@ -62,6 +66,21 @@ function extractIp (config, req) {
...
+}
+
+function extractPublicOrPrivateIp (ip) {
+  const type = net.isIP(ip)
+  if (!type) return {}
```
simon-id: this is a hot path, could we return null/undefined instead, and check for an early exit in findFirstIp like so:
```suggestion
  if (!type) return
```

```javascript
const currentIpInfo = extractPublicOrPrivateIp(chunk)
if (!currentIpInfo) continue
```

packages/dd-trace/src/plugins/util/ip_extractor.js
```diff
@@ -73,17 +92,48 @@ function findFirstIp (str) {
...
+    const currentIpInfo = extractPublicOrPrivateIp(chunk)
+    if (currentIpInfo.public) {
       // it's public, return it immediately
       result.public = chunk
       break
     }
 
     // it's private, only save the first one found
-    if (!result.private) result.private = chunk
+    if (currentIpInfo.private && !result.private) result.private = chunk
```
simon-id: actually even better, if you just put back the net.isIP() back here, you can avoid creating all these objects, and just call "isPublicIP()".
I think this is a hot enough path that we should prioritize speed over not repeating code.
```suggestion
    const type = net.isIP(chunk)
    if (!type) continue

    if (isPublicIp(chunk, type)) {
      // it's public, return it immediately
      result.public = chunk
      break
    }

    // it's private, only save the first one found
    if (!result.private) result.private = chunk
```

packages/dd-trace/src/plugins/util/ip_extractor.js
```diff
@@ -73,17 +92,48 @@ function findFirstIp (str) {
...
+
+      const tempResult = extractPublicOrPrivateIp(ip)
+      if (tempResult.public) {
+        // it's public, return it immediately
+        result.public = ip
+        break
+      }
+
+      // it's private, only save the first one found
+      if (tempResult.private && !result.private) result.private = ip
```
simon-id: same thing here
```suggestion
      const type = net.isIP(chunk)
      if (!type) continue

      if (isPublicIp(ip)) {
        // it's public, return it immediately
        result.public = ip
        break
      }

      // it's private, only save the first one found
      if (!result.private) result.private = ip
```

packages/dd-trace/src/plugins/util/ip_extractor.js
```diff
@@ -73,17 +92,48 @@ function findFirstIp (str) {
...
-    if (!result.private) result.private = chunk
+    if (currentIpInfo.private && !result.private) result.private = chunk
+  }
+
+  return result
+}
+
+const forwardedForRegexp = /for="?\[?(([0-9]+\.)+[0-9]+|[0-9a-f:]*:[0-9a-f]*)/i
+const forwardedByRegexp = /by="?\[?(([0-9]+\.)+[0-9]+|[0-9a-f:]*:[0-9a-f]*)/i
+const forwardedRegexps = [forwardedForRegexp, forwardedByRegexp]
```
simon-id, line 102: I don't understand why you need two regexes and a loop here, why not just combine them ?
```suggestion
const forwardedRegexp = /(?:for|by)="?\[?(([0-9]+\.)+[0-9]+|[0-9a-f:]*:[0-9a-f]*)/i
```

simon-id: also you can optimize the regex by turning the groups you don't care about into non capturing groups like:
`/(?:for|by)="?\[?((?:[0-9]+\.)+[0-9]+|[0-9a-f:]*:[0-9a-f]*)/i`

simon-id (changes requested):
To conclude i think you can greatly simplify this by keeping only one function like so:
```javascript
findFirstIp(header, ipHeaderName === FORWARED_HEADER_NAME)

const forwardedRegex = /(?:for|by)="?\[?((?:[0-9]+\.)+[0-9]+|[0-9a-f:]*:[0-9a-f]*)/i

function findFirstIp (str, isForwardedHeader) {
  const result = {}
  if (!str) return result

  const splitted = str.split(',')

  for (const part of splitted) {
    let chunk = part.trim()

    if (isForwardedHeader) {
        chunk = forwardedRegex.exec(chunk)?.[1] || chunk
    }

    const type = net.isIP(chunk)
    if (!type) continue

    if (isPublicIp(chunk, type)) {
        // it's public, return it immediately
        result.public = chunk
        break
    }

      // it's private, only save the first one found
    if (!result.private) result.private = chunk
  }
}
```

packages/dd-trace/src/plugins/util/ip_extractor.js
uurien: answering the same here: https://github.com/DataDog/dd-trace-js/pull/6248#issuecomment-3173637014
We need either long, complex and unreadable regex or two to have separated `for` and `by`.

packages/dd-trace/src/plugins/util/ip_extractor.js
```diff
@@ -76,7 +84,7 @@ function findFirstIp (str) {
     const type = net.isIP(chunk)
     if (!type) continue
 
-    if (!privateIPMatcher.check(chunk, type === 6 ? 'ipv6' : 'ipv4')) {
+    if (isPublicIp(chunk, type)) {
       // it's public, return it immediately
       result.public = chunk
       break
```
simon-id, line 90: maybe we can do `return result` just like in the new function, to keep some consistnecy

simon-id: maybe we can do return result just like in the new function, to keep some consistnecy
```suggestion
      return result
```

simon-id (approved):

packages/dd-trace/src/plugins/util/ip_extractor.js
uurien: ok yes. I changed in the other because there is a `for` in `for` and we were doing break .... if () { break; }

PR #6249, opened 2025-08-08
test: remove withVersions global

BridgeAR:
This makes sure the withVersions helper is imported in all test files. That helps IDEs to show the description and allows to jump to the implementation.

As drive-by, remove assert.doesNotThrow() and replace those with a comment. The call itself just adds CPU cycles and a comment should be fine as such.


packages/dd-trace/test/profiling/profilers/poisson.spec.js
```diff
@@ -67,7 +67,8 @@ describe('PoissonProcessSamplingFilter', () => {
     })
     assert.strictEqual(callCount, 1)
     const event = { startTime: 0, duration: Number.POSITIVE_INFINITY }
-    assert.doesNotThrow(() => filter.filter(event))
+    // Make sure that filtering events does not throw and calls now() once
```
rochdev, line 70: I guess the net result is the same, but I personally prefer an explicit assertion. The fact that a comment is needed shows that the code is otherwise not clear.

simon-id: Soft agree, the assertion is only useful to make it explicit that this shouldn't throw. I'll survive if we remove them but not sure I'll remember to add comments like that.

simon-id (approved):

packages/dd-trace/test/profiling/profilers/poisson.spec.js
BridgeAR: Any API call should theoretically be able to throw. It feels like we would need to wrap all our code into such things in that case.

rochdev: In those cases something else is asserted. But there are cases where the only thing the test does it to test whether it's throwing, although I guess at that point the test title could just reflect that.

PR #6258, opened 2025-08-11
Endpoints discovery for fastify

IlyasShabi:
### What does this PR do?

Collect endpoints defined using Fastify without relying on trafic [RFC](https://docs.google.com/document/d/1txwuurIiSUWjYX7Xa0let7e49XKW2uhm1djgqjl_gL0/edit?pli=1&tab=t.0#heading=h.8d3o7vtyu1y1)

- [x] Add endpoints discovery config
- [x] Telemetry
- [x] Fastify endpoints collection
- [x] Tests

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/integration-tests/appsec/endpoints-collection.spec.js
+++ b/integration-tests/appsec/endpoints-collection.spec.js
@@ -93,22 +93,19 @@ describe('Endpoints collection', () => {
       const endpointsFound = []
       const isFirstFlags = []
 
-      const telemetryPromise = agent.assertTelemetryReceived(msg => {
-        const { payload } = msg
-        if (payload.request_type === 'app-endpoints') {
-          isFirstFlags.push(Boolean(payload.payload.is_first))
+      const telemetryPromise = agent.assertTelemetryReceived(({ payload }) => {
+        isFirstFlags.push(Boolean(payload.payload.is_first))
 
-          if (payload.payload.endpoints) {
-            payload.payload.endpoints.forEach(endpoint => {
-              endpointsFound.push({
-                method: endpoint.method,
-                path: endpoint.path,
-                type: endpoint.type,
-                operation_name: endpoint.operation_name,
-                resource_name: endpoint.resource_name
-              })
+        if (payload.payload.endpoints) {
+          payload.payload.endpoints.forEach(endpoint => {
+            endpointsFound.push({
+              method: endpoint.method,
+              path: endpoint.path,
+              type: endpoint.type,
+              operation_name: endpoint.operation_name,
+              resource_name: endpoint.resource_name
             })
-          }
+          })
         }
       }, 'app-endpoints', 5_000, 4)
 


integration-tests/appsec/endpoints-collection/fastify.js
```diff
@@ -0,0 +1,52 @@
...
+  url: '/multi-method',
+  handler: async (_, reply) => reply.send('ok')
+})
+
+// Wildcard route
+app.all('/wildcard', async (_, reply) => reply.send('ok'))
+
+// Nested routes with Router
+app.register(async function (fastify) {
+  fastify.put('/nested/:id', async (_, reply) => reply.send('ok'))
```
uurien: The name `fastify` here is confusing, I think that better if we call to it `nestedRoute`, `nestedApp`, `route` or anything else.

```suggestion
app.register(async function (nestedRoute) {
  nestedRoute.put('/nested/:id', async (_, reply) => reply.send('ok'))
```

packages/dd-trace/test/config.spec.js
```diff
@@ -1441,7 +1447,8 @@ describe('Config', () => {
       apmTracingEnabled: true,
       appsec: {
         apiSecurity: {
-          enabled: true
+          enabled: true,
+          endpointCollectionMessageLimit: 150
```
uurien, line 1462: shouldn't you set here also `endpointCollectionEnabled` to false? to test it later

packages/dd-trace/test/config.spec.js
```diff
@@ -1711,7 +1720,9 @@ describe('Config', () => {
     expect(config).to.have.deep.property('appsec', {
       apiSecurity: {
         enabled: true,
-        sampleDelay: 30
+        sampleDelay: 30,
+        endpointCollectionEnabled: true,
+        endpointCollectionMessageLimit: 300
```
uurien: why are you setting here default values and not checking it anywhere? if we don't have experimental options for these properties, we can just ignore this test.

IlyasShabi: We are testing the final config object without setting any value

integration-tests/appsec/endpoints-collection/fastify.js
IlyasShabi: yes it's confusing, I changed router names

packages/dd-trace/src/telemetry/endpoints.js
```diff
@@ -0,0 +1,136 @@
...
+  pendingEndpoints.set(key, { method: method.toUpperCase(), path })
+  scheduleFlush()
+}
+
+function onFastifyRoute (routeData) {
+  if (!routeData) return
+
+  const { routeOptions } = routeData
+  if (!routeOptions) return
+  if (!routeOptions.path) return
```
simon-id:
```suggestion
  const routeOptions = routeData?.routeOptions
  if (!routeOptions?.path) return
```

packages/dd-trace/src/telemetry/endpoints.js
```diff
@@ -0,0 +1,136 @@
...
+
+  const methods = Array.isArray(routeOptions.method) ? routeOptions.method : [routeOptions.method]
+
+  // Check if this is a wildcard route fastify.all()
+  if (methods.length === 8) {
+    recordEndpoint('*', routeOptions.path)
```
simon-id: does the RFC asks you to send the wildcard or did you make it up ?

packages/dd-trace/src/telemetry/endpoints.js
```diff
@@ -0,0 +1,136 @@
...
+}
+
+function scheduleFlush () {
+  if (flushScheduled) return
+  flushScheduled = true
+  setImmediate(flushAndSend).unref()
+}
+
+function recordEndpoint (method, path) {
+  if (!method || !path) return
```
simon-id: not sure this is necesary, you already check path before, and not sure how method could be empty

packages/dd-trace/src/telemetry/endpoints.js
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * Keep track of endpoints that still need to be sent.
+ * Map key is `${METHOD} ${PATH}`, value is { method, path }
+ */
+const pendingEndpoints = new Map()
+let flushScheduled = false
+let isFirstPayload = true
+
+function endpointKey (method, path) {
+  return `${method} ${path}`
```
simon-id: just in case ?
```suggestion
  return `${method.toUpperCase()} ${path}`
```

packages/dd-trace/src/telemetry/endpoints.js
IlyasShabi: This is the list of supported methods listed in the RFC:
```
"method": {
        "type": "string",
         "enum": [
                "GET",
                "POST",
                "PUT",
                "DELETE",
                "PATCH",
                "HEAD",
                "OPTIONS",
                "TRACE",
                "CONNECT",
                "*"
         ]
}
```

simon-id: I count 9 here, so why do you check if it's 8 ?

packages/dd-trace/src/telemetry/endpoints.js
```diff
@@ -0,0 +1,136 @@
...
+function start (_config = {}, _application, _host, getRetryDataFunction, updateRetryDatafunction) {
+  config = _config
+  application = _application
+  host = _host
+  getRetryData = getRetryDataFunction
+  updateRetryData = updateRetryDatafunction
+
+  if (config.appsec?.apiSecurity?.endpointCollectionEnabled) {
+    fastifyRouteCh.subscribe(onFastifyRoute)
+  }
```
simon-id: well if it's not enabled you can quit early no ?
```suggestion
  if (!config.appsec?.apiSecurity?.endpointCollectionEnabled) return

  config = _config
  application = _application
  host = _host
  getRetryData = getRetryDataFunction
  updateRetryData = updateRetryDatafunction

  
  fastifyRouteCh.subscribe(onFastifyRoute)
```

packages/dd-trace/src/telemetry/endpoints.js
```diff
@@ -0,0 +1,136 @@
...
+}
+
+function stop () {
+  if (fastifyRouteCh.hasSubscribers) {
+    fastifyRouteCh.unsubscribe(onFastifyRoute)
+  }
```
simon-id: it's not neccesary to check this anymore. We used to need it because there was a bug in DC, if there was no subscriber, unsubscribe would throw. But it's been fixed now no worries
```suggestion
  fastifyRouteCh.unsubscribe(onFastifyRoute)
```

packages/dd-trace/src/telemetry/endpoints.js
IlyasShabi: Fastify doesn’t accept `'*'`,  so when you do a `app.all()` fastify is defining 8 endpoints for different http methods

simon-id: yes but is this what the RFC mean by `*` ? wouldn't that create an ambiguous situation where the backend cannot tell the difference if there is "connect" (i assume) or not ?

packages/dd-trace/src/config.js
```diff
@@ -483,6 +483,8 @@ class Config {
     defaults.apmTracingEnabled = true
     defaults['appsec.apiSecurity.enabled'] = true
     defaults['appsec.apiSecurity.sampleDelay'] = 30
+    defaults['appsec.apiSecurity.endpointCollectionEnabled'] = true
+    defaults['appsec.apiSecurity.endpointCollectionMessageLimit'] = 300
```
CarlesDD, line 497: Don't forget to add these new confs to [normalization file]

IlyasShabi: Thanks, PR created on other repo

packages/dd-trace/src/telemetry/endpoints.js
IlyasShabi: Is there another meaning for `*`?

simon-id: well if the backend considers * = all nine methods, then this code is invalid right ?

IlyasShabi: When defining a wildcard route `*`, it's important to note that `CONNECT` requests won't be matched, as Fastify does not support this http method.
We should consider whether to use * or explicitly list supported HTTP methods. I can bring this question to the WG for clarification

packages/dd-trace/src/telemetry/endpoints.js
simon-id: @IlyasShabi i see you reverted this ?

IlyasShabi: Done

packages/dd-trace/src/telemetry/endpoints.js
```diff
@@ -0,0 +1,129 @@
...
+  updateRetryData = updateRetryDatafunction
+
+  fastifyRouteCh.subscribe(onFastifyRoute)
+}
+
+function stop () {
+  fastifyRouteCh.unsubscribe(onFastifyRoute)
+
+  pendingEndpoints.clear()
+  flushScheduled = false
```
simon-id, line 117: i don't understand why there is a flushScheduler = false here ?

packages/dd-trace/src/telemetry/endpoints.js
simon-id: Question is still pending unanswered. We will fix it later in this PR if needed: https://github.com/DataDog/dd-trace-js/pull/6266

packages/dd-trace/test/config.spec.js
simon-id, line 1: this is missing 2 asserts in the `updateConfig` variable checks:
- `should initialize with the correct defaults`
- `should initialize from environment variables`

simon-id, line 1: the test `should give priority to the options` is updated completely wrongly. It's supposed to set conflicting values for env var and programmatic configs, and assert it's correctly prioritizing the options. In the current state, it's not doing anything useful.

packages/dd-trace/test/config.spec.js
simon-id: >We are testing the final config object without setting any value

That would be just doing the same as the `should initialize with the correct defaults` test, so there is no point in repeating that here.

>f we don't have experimental options for these properties, we can just ignore this test.

Since apiSecurity is under appsec, and appsec has experimental props because we never bothered to remove it. Then we should also test it for these new props.

integration-tests/appsec/endpoints-collection/fastify.js
```diff
@@ -0,0 +1,52 @@
...
+app.patch('/users/:id', async (_, reply) => reply.send('ok'))
+app.options('/users', async (_, reply) => reply.send('ok'))
+
+// Additional methods
+app.trace('/trace-test', async (_, reply) => reply.send('ok'))
+app.head('/head-test', async (_, reply) => reply.send('ok'))
+
+// Using app.route()
+app.route({
+  method: ['POST'],
```
simon-id: adding coverage
```suggestion
  method: ['POST', 'PUT', 'PATCH'],
```

integration-tests/appsec/endpoints-collection/fastify.js
```diff
@@ -0,0 +1,52 @@
...
+  url: '/multi-method',
+  handler: async (_, reply) => reply.send('ok')
+})
+
+// Wildcard route
+app.all('/wildcard', async (_, reply) => reply.send('ok'))
```
simon-id: adding coverage
```suggestion
app.all('/*', async (_, reply) => reply.send('ok'))
```

integration-tests/appsec/endpoints-collection/fastify.js
```diff
@@ -0,0 +1,52 @@
...
+tracer.init({
+  flushInterval: 0
+})
+
+const fastify = require('fastify')
+const app = fastify()
+
+// Basic routes
+app.get('/users', async (_, reply) => reply.send('ok'))
+app.post('/users', async (_, reply) => reply.send('ok'))
```
simon-id: adding coverage
```suggestion
app.post('/users/', async (_, reply) => reply.send('ok'))
```

integration-tests/appsec/endpoints-collection/fastify.js
```diff
@@ -0,0 +1,52 @@
...
+app.put('/users/:id', async (_, reply) => reply.send('ok'))
+app.delete('/users/:id', async (_, reply) => reply.send('ok'))
+app.patch('/users/:id', async (_, reply) => reply.send('ok'))
```
simon-id: adding coverage
```suggestion
app.patch('/users/:id/:name', async (_, reply) => reply.send('ok'))
```

integration-tests/appsec/endpoints-collection/fastify.js
```diff
@@ -0,0 +1,52 @@
...
+})
+
+const fastify = require('fastify')
+const app = fastify()
+
+// Basic routes
+app.get('/users', async (_, reply) => reply.send('ok'))
+app.post('/users', async (_, reply) => reply.send('ok'))
+app.put('/users/:id', async (_, reply) => reply.send('ok'))
+app.delete('/users/:id', async (_, reply) => reply.send('ok'))
```
simon-id, line 15: adding coverage
```suggestion
app.delete('/example/at/:hour(^\\d{2})h:minute(^\\d{2})m', async (_, reply) => reply.send('ok'))
```

integration-tests/appsec/endpoints-collection/fastify.js
```diff
@@ -0,0 +1,52 @@
...
+app.patch('/users/:id', async (_, reply) => reply.send('ok'))
+app.options('/users', async (_, reply) => reply.send('ok'))
```
simon-id: adding coverage
```suggestion
app.options('/users/:id?', async (_, reply) => reply.send('ok'))
```

integration-tests/appsec/endpoints-collection/fastify.js
```diff
@@ -0,0 +1,52 @@
...
+
+// Additional methods
+app.trace('/trace-test', async (_, reply) => reply.send('ok'))
+app.head('/head-test', async (_, reply) => reply.send('ok'))
+
```
simon-id, line 25: adding coverage
```suggestion

// Custom method
fastify.addHttpMethod('MKCOL', { hasBody: true })
fastify.mkcol('/example/near/:lat-:lng/radius/:r', async (_, reply) => reply.send('ok'))

```

integration-tests/appsec/endpoints-collection/fastify.js
```diff
@@ -0,0 +1,52 @@
...
+// Wildcard route
+app.all('/wildcard', async (_, reply) => reply.send('ok'))
+
+// Nested routes with Router
+app.register(async function (router) {
+  router.put('/nested/:id', async (_, reply) => reply.send('ok'))
+}, { prefix: '/v1' })
+
+// Deeply nested routes
+app.register(async function (router) {
```
simon-id, line 39: adding coverage
```suggestion
app.register(async function (router) {
    router.get('/nested', async (_, reply) => reply.send('ok'))
```

simon-id, line 46: adding coverage

```suggestion
app.register(async function (router) {
  router.get('/nested', async (_, reply) => reply.send('ok'))
```

packages/dd-trace/src/telemetry/endpoints.js
CarlesDD: This approach does not work correctly when customer code adds a custom http method with [addHttpMethod](https://fastify.dev/docs/latest/Reference/Server/#addHttpMethod). In this situation, 8+n methods will be received (the default 8 + n custom methods).

packages/dd-trace/test/config.spec.js
CarlesDD: Assertions have been added to both test cases

CarlesDD: A conflicting set of apiSecurity configs has been set for env vars, making the test meaningful.

integration-tests/appsec/endpoints-collection/fastify.js
CarlesDD: Two new cases have been added.

- One declaring a route for all the available methods (including the ones added with `addHttpMethod`). Route's path name is `/all-routes`
- Another one declaring two routes with wildcard `*`

integration-tests/appsec/endpoints-collection/fastify.js
CarlesDD: The case to cover this is not exactly as the one you proposed, but it works to tests the route with regex case.

packages/dd-trace/test/config.spec.js
CarlesDD: I've reworked the changes to include a config set in experimental to test against it.

packages/dd-trace/src/telemetry/endpoints.js
```diff
@@ -0,0 +1,124 @@
...
+
+  if (isFirstPayload) {
+    isFirstPayload = false
+  }
+
+  // If more endpoints accumulated while sending, schedule another flush.
+  if (pendingEndpoints.size) scheduleFlush()
+}
+
+function start (_config = {}, _application, _host, getRetryDataFunction, updateRetryDatafunction) {
```
CarlesDD:
```suggestion
function start (_config = {}, _application, _host, getRetryDataFunction, updateRetryDataFunction) {
```

packages/dd-trace/src/telemetry/endpoints.js
```diff
@@ -0,0 +1,124 @@
...
+  if (!_config?.appsec?.apiSecurity?.endpointCollectionEnabled) return
+
+  config = _config
+  application = _application
+  host = _host
+  getRetryData = getRetryDataFunction
+  updateRetryData = updateRetryDatafunction
```
CarlesDD:
```suggestion
  updateRetryData = updateRetryDataFunction
```

packages/dd-trace/src/telemetry/endpoints.js
CarlesDD: I guess the reason is to set everything to its initial state. `flushScheduled` could prevent flushing and sending data if it is true after a `start` call. (I'm just trying to bring meaning to this since I'm not the original writer of the code)

integration-tests/appsec/endpoints-collection/fastify.js
```diff
@@ -0,0 +1,64 @@
...
+  router.get('/nested', async (_, reply) => reply.send('ok'))
+  router.register(async function (subRouter) {
+    subRouter.get('/deep', async (_, reply) => reply.send('ok'))
+    subRouter.post('/deep/:id', async (_, reply) => reply.send('ok'))
+  }, { prefix: '/sub' })
+}, { prefix: '/api' })
+
+// Wildcard routes
+app.get('/wildcard/*', async (_, reply) => reply.send('ok'))
+app.get('/*', async (_, reply) => reply.send('ok'))
```
simon-id: i think i made a mistake with my previous suggestion, i was thinking of removing the slash too
```suggestion
app.get('*', async (_, reply) => reply.send('ok'))
```

integration-tests/appsec/endpoints-collection.spec.js
```diff
@@ -0,0 +1,139 @@
...
+      // Check that all expected endpoints were found
+      expectedEndpoints.forEach(expected => {
+        const found = endpointsFound.find(e =>
+          e.method === expected.method && e.path === expected.path
+        )
+        expect(found).to.exist
+        expect(found.type).to.equal('REST')
+        expect(found.operation_name).to.equal('http.request')
+        expect(found.resource_name).to.equal(`${expected.method} ${expected.path}`)
+      })
```
simon-id, line 135:
```suggestion
      // Check that all expected endpoints were found
      expectedEndpoints.forEach(expected => {
        const found = endpointsFound.find(e =>
          e.method === expected.method && e.path === expected.path
        )
        expect(found).to.exist
        expect(found.type).to.equal('REST')
        expect(found.operation_name).to.equal('http.request')
        expect(found.resource_name).to.equal(`${expected.method} ${expected.path}`)
      })

      // check that no additional endpoints were found
      expect(expectedEndpoints.length).to.equal(endpointsFound.length)
```

simon-id: (only difference is the last two lines)

integration-tests/appsec/endpoints-collection/fastify.js
```diff
@@ -0,0 +1,64 @@
...
+
+const start = async () => {
+  await app.listen({ port: 0, host: '127.0.0.1' })
+  const port = app.server.address().port
+  process.send({ port })
+}
+
+start()
```
simon-id: since we have a system that collects the endpoints during the lifetime of the app, should we try to add some delay ?
```suggestion
setTimeout(() => {
    app.get('/later', async (_, reply) => reply.send('ok'))
    start()
}, 5e3)
```

simon-id: actually, what happens when we add a custom method after 5 seconds ? would the app.all() from before be updated ?

simon-id: i just tested it with fastify and if you call addhttpmethod after app.all() it doesn't update the app.all() so the current implementation works :laughing: 
image

packages/dd-trace/src/telemetry/endpoints.js
simon-id: i confused myself wrongly, no problem with it

packages/dd-trace/test/telemetry/endpoints.spec.js
```diff
@@ -0,0 +1,175 @@
...
+      const endpoints = proxyquire('../../src/telemetry/endpoints', {
+        'dc-polyfill': dc
+      })
+
+      const config = { appsec: { apiSecurity: { endpointCollectionEnabled: true } } }
+      endpoints.start(config)
+
+      expect(subscribe).to.have.been.calledOnce
+    })
+  })
```
simon-id: missing a test that check it doesn't subscribe if it's not enabled

CarlesDD: The test has been added

integration-tests/appsec/endpoints-collection.spec.js
CarlesDD: The additional check has been added

integration-tests/appsec/endpoints-collection/fastify.js
CarlesDD: Done

integration-tests/appsec/endpoints-collection/fastify.js
simon-id: added in https://github.com/DataDog/dd-trace-js/pull/6258/commits/5ac48527d3e69fa14394b14630a70829dbab3ebc

simon-id (approved):

integration-tests/appsec/endpoints-collection.spec.js
```diff
@@ -0,0 +1,151 @@
...
+    try {
+      agent = await new FakeAgent().start()
+
+      const expectedEndpoints = getExpectedEndpoints(framework)
+      const endpointsFound = []
+      const isFirstFlags = []
+
+      const telemetryPromise = agent.assertTelemetryReceived(msg => {
+        const { payload } = msg
+        if (payload.request_type === 'app-endpoints') {
```
CarlesDD: I think this conditional is useless, since the check is already done in fake agent:
https://github.com/DataDog/dd-trace-js/blob/master/integration-tests/helpers/fake-agent.js#L214

simon-id: TRUE

simon-id: BAM [`29ef6ef` (#6258)](https://github.com/DataDog/dd-trace-js/pull/6258/commits/29ef6ef1feea7ece950919f04272f018a120f7bf)

simon-id (approved):

PR #6259, opened 2025-08-11
skip dependabot update for major versions of nock nyc and otel

rochdev:
### What does this PR do?

Skip dependabot update for major versions of nock nyc and OTel.

### Motivation

They break our tests when updated:

https://github.com/DataDog/dd-trace-js/pull/6257
https://github.com/DataDog/dd-trace-js/pull/6057
https://github.com/DataDog/dd-trace-js/pull/5999
https://github.com/DataDog/dd-trace-js/pull/5972

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -43,6 +43,12 @@ updates:
         update-types: ["version-update:semver-major"]
         # The path-to-regexp version has to be the same as used in express v4.
         # Consider vendoring it instead.
+      - dependency-name: "nock"
+        # Latest version breaks our tests.
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "nyc"
+        # Latest version breaks our tests.
+        update-types: ["version-update:semver-major"]
       - dependency-name: "path-to-regexp"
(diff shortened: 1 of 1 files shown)


BridgeAR (changes requested):
I didn't add these to the list because we could update these. I don't think we should not receive the update for breaking our test reasons.

I am therefore -0.75 on this

rochdev:
> I didn't add these to the list because we could update these. I don't think we should not receive the update for breaking our test reasons. We should fix the issues instead. Only if that's not possible, then we should skip it.

I agree in principle, but in practice some of those PRs have been sitting there for months, and previous attempts to fix them seemed to lead nowhere. If we're not going to spend the time to fix them anyway, we should at least ignore them from Dependabot otherwise it's just unnecessary noise.

BridgeAR:
Nock was just recently updated to a newer version (while still not being latest). I do still think we should keep the automation for these.

watson:
If we add these to an ignore list, how do we remember to come back later and fix our tests once we do have time? Should we mark it with a `TODO` or?

PR #6260, opened 2025-08-11
[SVLS-7168] Create GCP PubSub Push Subscriptions Plugin

nina9753:
### What does this PR do?
Extends the official PubSub plugin capabilities and optimizes Datadog tracing instrumentation for Google Cloud Pub/Sub push messages and Cloud Events across all Node.js web frameworks.
- Automatic Detection: No configuration needed
- Large Payloads: Supports up to 10MB efficiently
- All Frameworks: Works with Express, Fastify, Hapi, Koa, etc.
- Distributed Tracing: Maintains trace context across services
- Production Ready: Comprehensive error handling

HTTP Request → http/server.js → wrapEmitForGcp → gcp-pubsub-push.js → 
  1. Parse PubSub body
  2. Create PubSub span  
  3. Create new clean request object
  4. Activate span context
  5. Emit to Express with clean request
 
This makes the spans have this hierarchy :
    pubsub.receive (created by gcp-pubsub-push.js)
    └── http.request (manually created as child)
        └── express.request (inherits via scope activation)
            └── express.middleware (inherits via scope activation)
                └── your-business-logic (main.js)

Follow-up PR will be opened to create the synthetic span in the flame graph

### Motivation
An inferred span for the HTTP push post to the Cloud Run service from a pub/sub topic 
[Example full Push Direct Span](https://app.datadoghq.com/apm/traces?query=env%3Ainferred&agg_m=count&agg_m_source=base&agg_t=count&colorLegendSort=time&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&colsTraces=service%2Cresource_name%2C%40duration%2C%40_span.count%2C%40_duration.by_service&fromUser=false&graphType=span_list&historicalData=false&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=time&spanID=931076104363788026&spanType=all&spanViewType=metadata&storage=hot&timeHint=1754931390816&trace=1764a403d998356553d8e428e6b62b9d931076104363788026&traceID=1764a403d998356553d8e428e6b62b9d&traceQuery=&view=spans&start=1754931836845&end=1754932736845&paused=false) of a cloud run service triggering another service using a direct push subscription 
[Example full Eventarc pubsub trigger span](https://ddserverless.datadoghq.com/apm/trace/1265938993667486384?colorLegendSort=time&graphType=service_map&shouldShowLegend=true&spanID=6716787009469975145&timeHint=1754505158413.0085&traceQuery=) of a cloud run service triggering another service using an Eventarc cloud event trigger

### Plugin Checklist

- [x] Unit tests.
- [x] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes
The changes had to be made in server.js because in Push Subscription, the Cloud Run Service receives HTTP Requests. 
- Google Cloud Pub/Sub acts as the HTTP client
- The cloud run service acts as the HTTP server
- GCP topic pushes TO the cloud run service via HTTP POST requests
- Then the cloud run service receives the pushed messages

Additional information can be found in [this doc](https://docs.google.com/document/d/12yVv7aOtQn04gFMixarDVFXUvdqlD6-ApBBxk1IoEB0/edit?pli=1&tab=t.0)

image

### DIFF
--- a/packages/datadog-instrumentations/src/body-parser.js
+++ b/packages/datadog-instrumentations/src/body-parser.js
@@ -26,6 +26,10 @@ addHook({
   versions: ['>=1.4.0 <1.20.0']
 }, read => {
   return shimmer.wrapFunction(read, read => function (req, res, next) {
+    // Skip body parsing if PubSub/Cloud Event middleware already parsed it
+    if (req._pubsubBodyParsed) {
+      return next()
+    }
     const nextResource = new AsyncResource('bound-anonymous-fn')
     arguments[2] = nextResource.bind(publishRequestBodyAndNext(req, res, next))
     return read.apply(this, arguments)
@@ -38,6 +42,10 @@ addHook({
   versions: ['>=1.20.0']
 }, read => {
   return shimmer.wrapFunction(read, read => function (req, res, next) {
+    // Skip body parsing if PubSub/Cloud Event middleware already parsed it
+    if (req._pubsubBodyParsed) {
+      return next()
+    }
     arguments[2] = publishRequestBodyAndNext(req, res, next)
     return read.apply(this, arguments)
   })
--- a/packages/datadog-instrumentations/src/body-parser.js
+++ b/packages/datadog-instrumentations/src/body-parser.js
@@ -26,8 +26,17 @@ addHook({
   versions: ['>=1.4.0 <1.20.0']
 }, read => {
   return shimmer.wrapFunction(read, read => function (req, res, next) {
-    // Skip body parsing if PubSub/Cloud Event middleware already parsed it
-    if (req._pubsubBodyParsed) {
+    // Skip body parsing if body has already been meaningfully parsed by any middleware
+    if (req.body !== undefined && req.body !== null &&
+        ((typeof req.body === 'object' && Object.keys(req.body).length > 0) ||
+         (typeof req.body === 'string' && req.body.length > 0))) {
+      // Still publish the channel so AppSec and IAST can process the body
+      if (bodyParserReadCh.hasSubscribers && req) {
+        const abortController = new AbortController()
+        const body = req.body
+        bodyParserReadCh.publish({ req, res, body, abortController })
+        if (abortController.signal.aborted) return
+      }
       return next()
     }
     const nextResource = new AsyncResource('bound-anonymous-fn')
@@ -42,8 +51,17 @@ addHook({
   versions: ['>=1.20.0']
 }, read => {
   return shimmer.wrapFunction(read, read => function (req, res, next) {
-    // Skip body parsing if PubSub/Cloud Event middleware already parsed it
-    if (req._pubsubBodyParsed) {
+    // Skip body parsing if body has already been meaningfully parsed by any middleware
+    if (req.body !== undefined && req.body !== null &&
+        ((typeof req.body === 'object' && Object.keys(req.body).length > 0) ||
+         (typeof req.body === 'string' && req.body.length > 0))) {
+      // Still publish the channel so AppSec and IAST can process the body
+      if (bodyParserReadCh.hasSubscribers && req) {
+        const abortController = new AbortController()
+        const body = req.body
+        bodyParserReadCh.publish({ req, res, body, abortController })
+        if (abortController.signal.aborted) return
+      }
       return next()
     }
     arguments[2] = publishRequestBodyAndNext(req, res, next)
--- a/packages/datadog-instrumentations/src/shared-channels.js
+++ b/packages/datadog-instrumentations/src/shared-channels.js
@@ -1,0 +1,17 @@
+'use strict'
+
+const { channel } = require('dc-polyfill')
+
+// Shared channel registry to ensure all modules use the same channel instances
+const channels = {}
+
+function getSharedChannel (name) {
+  if (!channels[name]) {
+    channels[name] = channel(name)
+  }
+  return channels[name]
+}
+
+module.exports = {
+  getSharedChannel
+}
--- a/packages/datadog-plugin-google-cloud-pubsub/src/producer.js
+++ b/packages/datadog-plugin-google-cloud-pubsub/src/producer.js
@@ -28,6 +28,11 @@ class GoogleCloudPubsubProducerPlugin extends ProducerPlugin {
         msg.attributes = {}
       }
       this.tracer.inject(span, 'text_map', msg.attributes)
+
+      // Also inject project_id and topic for consumer correlation
+      msg.attributes['gcloud.project_id'] = projectId
+      msg.attributes['pubsub.topic'] = topic
+
       if (this.config.dsmEnabled) {
         const payloadSize = getHeadersSize(msg)
         const dataStreamsContext = this.tracer
--- a/packages/datadog-instrumentations/src/http.js
+++ b/packages/datadog-instrumentations/src/http.js
@@ -1,4 +1,19 @@
 'use strict'
 
+const { getSharedChannel } = require('./shared-channels')
+
+try {
+  // Load the HttpHandler plugin directly to ensure it gets instantiated
+  const HttpHandlerPlugin = require('../../datadog-plugin-google-cloud-pubsub/src/http-handler')
+
+  // Get tracer instance and instantiate the plugin
+  const tracer = require('../../dd-trace')
+  if (tracer && tracer._tracer) {
+    HttpHandlerPlugin(tracer)
+  }
+} catch {
+  // Silently handle plugin loading errors
+}
+
 require('./http/client')
 require('./http/server')
--- a/packages/datadog-instrumentations/src/http.js
+++ b/packages/datadog-instrumentations/src/http.js
@@ -1,7 +1,5 @@
 'use strict'
 
-const { getSharedChannel } = require('./shared-channels')
-
 try {
   // Load the HttpHandler plugin directly to ensure it gets instantiated
   const HttpHandlerPlugin = require('../../datadog-plugin-google-cloud-pubsub/src/http-handler')
@@ -9,7 +7,7 @@ try {
   // Get tracer instance and instantiate the plugin
   const tracer = require('../../dd-trace')
   if (tracer && tracer._tracer) {
-    HttpHandlerPlugin(tracer)
+    new HttpHandlerPlugin(tracer) // eslint-disable-line no-new
   }
 } catch {
   // Silently handle plugin loading errors
--- a/packages/datadog-instrumentations/src/http/server.js
+++ b/packages/datadog-instrumentations/src/http/server.js
@@ -5,6 +5,12 @@ const {
   addHook
 } = require('../helpers/instrument')
 const shimmer = require('../../../datadog-shimmer')
+const { getSharedChannel } = require('../shared-channels')
+
+const httpNames = ['http', 'node:http']
+const httpsNames = ['https', 'node:https']
+
+// Generic HTTP server instrumentation - no product-specific logic
 
 const startServerCh = channel('apm:http:server:request:start')
 const exitServerCh = channel('apm:http:server:request:exit')
@@ -14,14 +20,16 @@ const startWriteHeadCh = channel('apm:http:server:response:writeHead:start')
 const finishSetHeaderCh = channel('datadog:http:server:response:set-header:finish')
 const startSetHeaderCh = channel('datadog:http:server:response:set-header:start')
 
-const requestFinishedSet = new WeakSet()
+// Generic channel for request interception - use shared channel to ensure same instance
+const requestInterceptCh = getSharedChannel('apm:http:server:request:intercept')
 
-const httpNames = ['http', 'node:http']
-const httpsNames = ['https', 'node:https']
+const requestFinishedSet = new WeakSet()
 
 addHook({ name: httpNames }, http => {
   shimmer.wrap(http.ServerResponse.prototype, 'emit', wrapResponseEmit)
+  shimmer.wrap(http.Server.prototype, 'emit', wrapEmitForInterception)
   shimmer.wrap(http.Server.prototype, 'emit', wrapEmit)
+
   shimmer.wrap(http.ServerResponse.prototype, 'writeHead', wrapWriteHead)
   shimmer.wrap(http.ServerResponse.prototype, 'write', wrapWrite)
   shimmer.wrap(http.ServerResponse.prototype, 'end', wrapEnd)
@@ -36,6 +44,7 @@ addHook({ name: httpNames }, http => {
 
 addHook({ name: httpsNames }, http => {
   // http.ServerResponse not present on https
+  shimmer.wrap(http.Server.prototype, 'emit', wrapEmitForInterception)
   shimmer.wrap(http.Server.prototype, 'emit', wrapEmit)
   return http
 })
@@ -54,6 +63,40 @@ function wrapResponseEmit (emit) {
     return emit.apply(this, arguments)
   }
 }
+
+// Generic request interceptor - allows any plugin to intercept requests
+function wrapEmitForInterception (emit) {
+  return function (eventName, req, res) {
+    // Only process 'request' events
+    if (eventName !== 'request') {
+      return emit.apply(this, arguments)
+    }
+
+    // Check if any plugin wants to intercept this request
+    if (requestInterceptCh.hasSubscribers) {
+      const interceptData = {
+        req,
+        res,
+        emit,
+        server: this,
+        originalArgs: arguments,
+        handled: false // Plugin sets this to true if it handles the request
+      }
+
+      // Publish to generic intercept channel - any plugin can subscribe
+      requestInterceptCh.publish(interceptData)
+
+      // If a plugin handled it, don't continue with normal processing
+      if (interceptData.handled) {
+        return true
+      }
+    }
+
+    // No plugin intercepted, continue with normal HTTP processing
+    return emit.apply(this, arguments)
+  }
+}
+
 function wrapEmit (emit) {
   return function (eventName, req, res) {
     if (!startServerCh.hasSubscribers) {
@@ -62,9 +105,12 @@ function wrapEmit (emit) {
 
     if (eventName === 'request') {
       res.req = req
+      if (req._isPubSubPush || req._isCloudEvent) {
+        return emit.apply(this, arguments)
+      }
 
+      // Normal HTTP request processing (not PubSub/Cloud Events)
       const abortController = new AbortController()
-
       startServerCh.publish({ req, res, abortController })
 
       try {
@@ -221,3 +267,6 @@ function wrapEnd (end) {
     return end.apply(this, arguments)
   }
 }
+
+// Export the channel for plugins to use the same instance
+module.exports = { requestInterceptCh }
--- a/packages/datadog-instrumentations/test/google-cloud-pubsub-http-handler.spec.js
+++ b/packages/datadog-instrumentations/test/google-cloud-pubsub-http-handler.spec.js
@@ -1,0 +1,221 @@
+'use strict'
+
+const { expect } = require('chai')
+const sinon = require('sinon')
+const http = require('http')
+const dc = require('dc-polyfill')
+
+describe('Google Cloud Pub/Sub HTTP Handler Plugin', () => {
+  let tracer, server, port
+
+  // Import the HTTP handler plugin
+  const GoogleCloudPubsubHttpHandlerPlugin = require('../../datadog-plugin-google-cloud-pubsub/src/http-handler')
+
+  // Create helper functions for testing
+  const mockTracer = {
+    startSpan: () => ({ setTag: () => {}, finish: () => {} }),
+    extract: () => null,
+    scope: () => ({ activate: (span, cb) => cb() })
+  }
+  const pluginInstance = new GoogleCloudPubsubHttpHandlerPlugin(mockTracer)
+
+  // Extract the methods we want to test
+  const isPubSubRequest = pluginInstance.isPubSubRequest.bind(pluginInstance)
+  const isCloudEventRequest = pluginInstance.isCloudEventRequest.bind(pluginInstance)
+  const processEventRequest = pluginInstance.processPubSubRequest.bind(pluginInstance)
+
+  before(() => {
+    // Set up mock tracer
+    tracer = {
+      startSpan: sinon.stub().returns({
+        context: sinon.stub().returns({
+          parentId: 'parent-123',
+          toSpanId: sinon.stub().returns('span-456'),
+          toTraceId: sinon.stub().returns('trace-789')
+        }),
+        setTag: sinon.stub(),
+        finish: sinon.stub(),
+        finished: false
+      }),
+      extract: sinon.stub().returns({
+        _spanId: 'extracted-parent-123',
+        _traceId: 'extracted-trace-456',
+        toTraceId: sinon.stub().returns('extracted-trace-456')
+      }),
+      inject: sinon.stub(),
+      scope: sinon.stub().returns({
+        activate: sinon.stub().callsArg(1)
+      })
+    }
+
+    global._ddtrace = tracer
+  })
+
+  beforeEach((done) => {
+    server = http.createServer((req, res) => {
+      res.writeHead(200, { 'Content-Type': 'text/plain' })
+      res.end('OK')
+    })
+
+    server.listen(0, () => {
+      port = server.address().port
+      done()
+    })
+  })
+
+  afterEach((done) => {
+    sinon.restore()
+    if (server) {
+      server.close(done)
+    } else {
+      done()
+    }
+  })
+
+  describe('Function Exports', () => {
+    it('should export required functions', () => {
+      expect(typeof isPubSubRequest).to.equal('function')
+      expect(typeof isCloudEventRequest).to.equal('function')
+      expect(typeof processEventRequest).to.equal('function')
+    })
+  })
+
+  after(() => {
+    // Clean up global tracer mock
+    delete global._ddtrace
+  })
+
+  describe('PubSub Push HTTP Request Detection', () => {
+    it('should detect traditional PubSub push requests', (done) => {
+      const payload = JSON.stringify({
+        message: {
+          data: Buffer.from('test message').toString('base64'),
+          messageId: 'test-message-id',
+          attributes: {
+            traceparent: '00-12345678901234567890123456789012-1234567890123456-01'
+          }
+        },
+        subscription: 'projects/test-project/subscriptions/test-subscription'
+      })
+
+      const req = http.request({
+        hostname: 'localhost',
+        port,
+        path: '/',
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'User-Agent': 'APIs-Google; (+https://developers.google.com/webmasters/APIs-Google.html)',
+          'Content-Length': Buffer.byteLength(payload)
+        }
+      }, (res) => {
+        expect(res.statusCode).to.equal(200)
+        done()
+      })
+
+      req.write(payload)
+      req.end()
+    })
+
+    it('should detect Eventarc Cloud Events', (done) => {
+      const payload = JSON.stringify({
+        message: {
+          data: Buffer.from('test message').toString('base64'),
+          messageId: 'test-message-id',
+          attributes: {
+            traceparent: '00-12345678901234567890123456789012-1234567890123456-01'
+          }
+        },
+        subscription: 'projects/test-project/subscriptions/eventarc-sub'
+      })
+
+      const req = http.request({
+        hostname: 'localhost',
+        port,
+        path: '/',
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'ce-specversion': '1.0',
+          'ce-type': 'google.cloud.pubsub.topic.v1.messagePublished',
+          'ce-source': '//pubsub.googleapis.com/projects/test-project/topics/test-topic',
+          'ce-id': 'test-message-id',
+          'Content-Length': Buffer.byteLength(payload)
+        }
+      }, (res) => {
+        expect(res.statusCode).to.equal(200)
+        done()
+      })
+
+      req.write(payload)
+      req.end()
+    })
+
+    it('should not interfere with regular HTTP requests', (done) => {
+      const payload = JSON.stringify({ test: 'data' })
+
+      const req = http.request({
+        hostname: 'localhost',
+        port,
+        path: '/',
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'User-Agent': 'Mozilla/5.0',
+          'Content-Length': Buffer.byteLength(payload)
+        }
+      }, (res) => {
+        expect(res.statusCode).to.equal(200)
+        done()
+      })
+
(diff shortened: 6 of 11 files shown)


packages/datadog-instrumentations/src/body-parser.js
```diff
@@ -26,6 +26,19 @@ addHook({
...
+         (typeof req.body === 'string' && req.body.length > 0))) {
+      // Still publish the channel so AppSec and IAST can process the body
+      if (bodyParserReadCh.hasSubscribers && req) {
+        const abortController = new AbortController()
+        const body = req.body
+        bodyParserReadCh.publish({ req, res, body, abortController })
+        if (abortController.signal.aborted) return
+      }
+      return next()
+    }
```
BridgeAR: I am not sure I understand this change.
I don't think we should ever skip that functionality, since it would also interfere with user's code, if I am not mistaken.

BridgeAR (commented):
This is quite a big PR, if you like, we could set up a sync call to go through the changes to review the PR :)

packages/datadog-instrumentations/src/body-parser.js
nina9753: Great point, I removed this change from the PR. Google also always sends a pubsub message in a JSON format, so for the message attributes, I do not have to worry about parsing the request body

packages/datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription.js
```diff
@@ -0,0 +1,160 @@
...
+    super(...args)
+    const channel = getSharedChannel('apm:http:server:request:intercept')
+    channel.subscribe(this.handleRequestIntercept.bind(this))
+  }
+
+  handleRequestIntercept (interceptData) {
+    const { req } = interceptData
+    const isPubSub = req.method === 'POST' && (
+      req.headers['user-agent']?.includes('APIs-Google') ||
+      Object.keys(req.headers).some(k => k.toLowerCase().startsWith('x-goog-pubsub-'))
```
nina9753: just check for one 
```x-goog-pubsub-subscription-name: "projects/myproject/..."
x-goog-pubsub-message-id: "2070443601311540"
x-goog-pubsub-publish-time: "2021-02-26T19:13:55.749Z"
x-goog-pubsub-ordering-key```

File: packages/datadog-instrumentations/src/shared-channels.js [hunk 3]
```diff
@@ -0,0 +1,17 @@
...
+
+function getSharedChannel (name) {
+  if (!channels[name]) {
+    channels[name] = channel(name)
+  }
+  return channels[name]
+}
+
+module.exports = {
+  getSharedChannel
```
nina9753: use this instead https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-instrumentations/src/helpers/instrument.js

nina9753:
@codex review

packages/datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    const { projectId, topicName } = this._extractProjectTopic(req.headers, subscription)
+    return { message, subscription, attrs: req.headers, projectId, topicName }
+  }
+
+  _createDeliverySpan (messageData, parentContext, tracer) {
+    const { message, subscription, topicName, attrs } = messageData
+    const subscriptionName = subscription.split('/').pop() || subscription
+    const publishStartTime = attrs['x-dd-publish-start-time']
```
tlhunter, line 146: What is `x-dd-publish-start-time`? It kinda looks like an HTTP header.

tlhunter:
We're definitely going to want a test with this.

For example we should simulate the situation where an HTTP request is received and it contains the Google pub sub user agent and headers and makes the appropriate span. Is the intent to only have the google pub sub span or also have an HTTP span with it? And which one is the parent? That's all stuff that should be a part of the test.

packages/datadog-instrumentations/src/google-cloud-pubsub.js
```diff
@@ -6,6 +6,14 @@ const {
 } = require('./helpers/instrument')
 const shimmer = require('../../datadog-shimmer')
 
+// Auto-load push subscription plugin to enable pubsub.delivery spans for push subscriptions
+try {
+  const PushSubscriptionPlugin = require('../../datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription')
+  new PushSubscriptionPlugin(null, {}).configure({})
+} catch {
+  // Push subscription plugin is optional
+}
```
tlhunter: What part of this is expected to throw?

nina9753: The plugin constructor could fail, but I can update this so users can know exactly what/why it failed but the plugin is only required for push subscriptions that have the `--push-no-wrapper-write-metadata` flag. I wrapped the plugin load in a try catch so even if it fails we wont lose traces just the trace context propagation 

```
catch (e) {
  log.debug(`PushSubscriptionPlugin not loaded: ${e.message}`)
}
```

packages/datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription.js
nina9753: Thank you for catching this. It was accidentally moved to the child PR when i refactored this PR

packages/datadog-instrumentations/src/google-cloud-pubsub.js
```diff
@@ -5,6 +5,16 @@ const {
...
+const log = require('../../dd-trace/src/log')
+
+// Auto-load push subscription plugin to enable pubsub.delivery spans for push subscriptions
+try {
+  const PushSubscriptionPlugin = require('../../datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription')
+  new PushSubscriptionPlugin(null, {}).configure({})
+} catch (e) {
+  // Push subscription plugin is optional
+  log.debug(`PushSubscriptionPlugin not loaded: ${e.message}`)
+}
```
tlhunter: It looks like it's not the requiring of the file that can fail but only the class instantiation. For that reason let's hoist the require out of the try/catch. It's better to make try/catch super specific IMO.

```suggestion
const log = require('../../dd-trace/src/log')
const PushSubscriptionPlugin = require('../../datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription')

// Auto-load push subscription plugin to enable pubsub.delivery spans for push subscriptions
try {
  new PushSubscriptionPlugin(null, {}).configure({})
} catch (e) {
  // Push subscription plugin is optional
  log.debug(`PushSubscriptionPlugin not loaded: ${e.message}`)
}
```

packages/datadog-instrumentations/src/google-cloud-pubsub.js
```diff
@@ -151,6 +160,55 @@ addHook({ name: '@google-cloud/pubsub', versions: ['>=1.2'], file: 'build/src/le
   return obj
 })
 
+function injectTraceContext (attributes, pubsub, topicName) {
+  if (attributes['x-datadog-trace-id'] || attributes.traceparent) return
+
+  try {
+    const tracer = require('../../dd-trace')
```
tlhunter: This line should also be hoisted to the top of the file.

packages/datadog-instrumentations/src/google-cloud-pubsub.js
```diff
@@ -151,6 +160,55 @@ addHook({ name: '@google-cloud/pubsub', versions: ['>=1.2'], file: 'build/src/le
...
+    const activeSpan = tracer.scope().active()
+    if (!activeSpan) return
+
+    tracer.inject(activeSpan, 'text_map', attributes)
+
+    const traceIdUpperBits = activeSpan.context()._trace.tags['_dd.p.tid']
+    if (traceIdUpperBits) attributes['_dd.p.tid'] = traceIdUpperBits
+  } catch {
+    // Silently fail - trace context injection is best-effort
```
tlhunter: Let's add a log here as well. Otherwise it would be difficult to diagnose if the feature isn't working for customers who expect it to.

I'd also like to know if you're seeing any failures here during test suite runs. For example if the `activeSpan.context()._trace.tags['_dd.p.tid']` line fails with something like "cannot access _trace of undefined" then the code might better be served looking like activeSpan.context()?._trace.tags['_dd.p.tid']`, and nothing else ever fails, then maybe we can remove the try/catch entirely.

packages/datadog-plugin-google-cloud-pubsub/src/producer.js
```diff
@@ -14,8 +14,9 @@ class GoogleCloudPubsubProducerPlugin extends ProducerPlugin {
 
     const messages = request.messages || []
     const topic = request.topic
+    const topicName = topic.split('/').pop() || topic
     const span = this.startSpan({ // TODO: rename
-      resource: `${api} ${topic}`,
+      resource: `${api} to Topic ${topicName}`,
```
tlhunter, line 75: This looks like it could be a breaking change for customers who are expecting the old resource name.

Is there an RFC or something explaining how the new approach is the correct approach?

packages/datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription.js
```diff
@@ -0,0 +1,110 @@
...
+        'Add --push-no-wrapper-write-metadata to your subscription.'
+      )
+    }
+  }
+
+  _createDeliverySpanAndActivate ({ req, res }) {
+    const messageData = this._parseMessage(req)
+    if (!messageData) return
+
+    const tracer = this.tracer || require('../../dd-trace')
```
tlhunter: This require call should be hoisted to the top of the file

tlhunter: Are you finding that `this.tracer` is sometimes available and sometimes not? If so it might be that a test just isn't providing the tracer when it should be. Ideally we wouldn't need to access the tracer from two sources like this.

packages/datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription.js
```diff
@@ -0,0 +1,110 @@
...
+        'pubsub.message_id': message.messageId,
+        'pubsub.delivery_method': 'push',
+        'pubsub.topic': topicName,
+        service: this.config.service || `${tracer._tracer._service}-pubsub`,
+        '_dd.base_service': tracer._tracer._service,
+        '_dd.serviceoverride.type': 'integration'
+      }
+    })
+
+    span.setTag('resource.name', `Push Subscription ${subscriptionName}`)
```
tlhunter: Does this work?

```suggestion
        '_dd.serviceoverride.type': 'integration',
        'resource.name': `Push Subscription ${subscriptionName}`
      }
    })
```

packages/datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription.js
```diff
@@ -0,0 +1,110 @@
...
+
+    return span
+  }
+
+  _extractProjectTopic (attrs, subscription) {
+    const topicName = attrs['pubsub.topic']
+    const projectId = subscription.match(/projects\/([^\\/]+)\/subscriptions/)
```
tlhunter: Where does the `projectId` get used?

Notably the `String#match()` method either returns an array where the [1] index is the match, or a null if there is no match.

```js
"projects/abcde/subscriptions".match(/projects\/([^\\/]+)\/subscriptions/)
[
  'projects/abcde/subscriptions',
  'abcde',
  index: 0,
  input: 'projects/abcde/subscriptions',
  groups: undefined
]
```

With a name like projectId I'm assuming you would want to return the string at [1].

tlhunter:
As discused on Zoom please add an experimental env var to enable this functionality. Here's a recent example that shows the two config files that need updating and how to access the config in code:

https://github.com/DataDog/dd-trace-js/pull/6733

packages/datadog-plugin-google-cloud-pubsub/src/producer.js
nina9753: There shouldn't be any customers who are using this plugin. This change was just to make the span resource more descriptive. This can be reverted, but I think it would be nice to know if there were any customers before using this plugin, and while this would be a breaking change for monitors, this will also help us identify any legacy users.  We are also sticking the new code behind a feature flag

packages/datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription.js
nina9753: The push subscription plugin initializes differently than consumer/producer plugins and subscribes to channels in its constructor, which happens before this.tracer is set.

packages/dd-trace/src/supported-configurations.json
```diff
@@ -161,6 +161,7 @@
     "DD_RUNTIME_METRICS_GC_ENABLED": ["A"],
     "DD_RUNTIME_METRICS_FLUSH_INTERVAL": ["A"],
     "DD_RUNTIME_METRICS_RUNTIME_ID_ENABLED": ["A"],
+    "DD_SERVERLESS_PUBSUB_ENABLED": ["A"],
```
BridgeAR:
```suggestion
    "DD_TRACE_SERVERLESS_PUBSUB_ENABLED": ["A"],
```

This aligns with other plugin names. I think we should use that.

packages/datadog-plugin-google-cloud-pubsub/src/producer.js
tlhunter: Ah, it might have just looked that way due to how the PRs were stacked.

packages/dd-trace/src/serverless.js
```diff
@@ -13,6 +13,11 @@ function getIsGCPFunction () {
   return isDeprecatedGCPFunction || isNewerGCPFunction
 }
 
+function enableServerlessPubsubSubscription () {
+  const serverlessPubsubSubscriptionEnabled = getEnvironmentVariable('DD_SERVERLESS_PUBSUB_ENABLED')
+  return getEnvironmentVariable('K_SERVICE') !== undefined && serverlessPubsubSubscriptionEnabled === 'true'
```
BridgeAR:
```suggestion
  return getEnvironmentVariable('K_SERVICE') !== undefined && !isFalse(serverlessPubsubSubscriptionEnabled)
```

packages/datadog-plugin-google-cloud-pubsub/test/index.spec.js
```diff
@@ -159,7 +170,7 @@ describe('Plugin', () => {
                 publish(topic, { data: Buffer.from('hello') })
               )
               .then(() => {
-                assert.strictEqual(tracer.scope().active(), firstSpan)
+                expect(tracer.scope().active()).to.equal(firstSpan)
```
BridgeAR: Please do not change any of these :)

Independent changes should be kept out and we actually want to use Node.js assertions.

This applies to multiple entries, not only this one, while I only use a single comment for these.

packages/datadog-instrumentations/src/google-cloud-pubsub-push.js
```diff
@@ -0,0 +1,28 @@
...
+const log = require('../../dd-trace/src/log')
+let initialized = false
+
+function init () {
+  if (initialized || !enableServerlessPubsubSubscription()) {
+    return
+  }
+
+  try {
+    const PushSubscriptionPlugin = require('../../datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription')
```
tlhunter:
```suggestion
const log = require('../../dd-trace/src/log')
const PushSubscriptionPlugin = require('../../datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription')
let initialized = false

function init () {
  if (initialized || !enableServerlessPubsubSubscription()) {
    return
  }

  try {
```

packages/dd-trace/src/supported-configurations.json
nina9753: updated to DD_TRACE_GCP_PUBSUB_PUSH_ENABLED

packages/datadog-plugin-google-cloud-pubsub/src/client.js
```diff
@@ -1,26 +1,48 @@
...
 
+  constructor (...args) {
+    super(...args)
+
+    // Subscribe to context retrieval channel for acknowledge/modifyAckDeadline operations
+    this.addSub('apm:google-cloud-pubsub:message:ack-retrieve', ({ storedContext }) => {
+      if (storedContext) {
+        // Restore the stored context so acknowledge spans are properly linked
+        storage('legacy').enterWith(storedContext)
+      }
```
nina9753: use addBind that returns to  Runstore **future imporvment

packages/datadog-plugin-google-cloud-pubsub/src/producer.js
```diff
@@ -2,42 +2,180 @@
...
+      // Add 128-bit trace ID upper bits if present
+      const traceIdUpperBits = activeSpan.context()._trace.tags['_dd.p.tid']
+      if (traceIdUpperBits) attributes['_dd.p.tid'] = traceIdUpperBits
+
+      // Add metadata
+      if (pubsub) attributes['gcloud.project_id'] = pubsub.projectId
+      if (topicName) attributes['pubsub.topic'] = topicName
+    } catch (err) {
+      // Silently fail - trace context injection is best-effort
+      log.warn(`Error injecting trace context into message attributes: ${err.message}`)
```
BridgeAR: Only using the error message is going to loose information about what is happening and especially the where.

We definitely need the full error, if we want to log this. I am just uncertain if we need to wrap this code in a try / catch in the first place. Could this really ever fail without being a bug in our implementation and we miss tests?

packages/datadog-plugin-google-cloud-pubsub/src/producer.js
```diff
@@ -2,42 +2,180 @@
...
+    const parentData = firstAttrs?.['x-datadog-trace-id'] && firstAttrs['x-datadog-parent-id']
+      ? {
+          traceId: firstAttrs['x-datadog-trace-id'],
+          spanId: firstAttrs['x-datadog-parent-id'],
+          traceIdUpper: firstAttrs['_dd.p.tid'],
+          samplingPriority: firstAttrs['x-datadog-sampling-priority']
+        }
+      : null
+
+    const topicName = topic.split('/').pop() || topic
```
BridgeAR: Nit: it would be faster to do `.lastIndexOf()` + `.slice()`

packages/datadog-plugin-google-cloud-pubsub/src/producer.js
```diff
@@ -2,42 +2,180 @@
...
 
-    for (const msg of messages) {
-      if (!msg.attributes) {
-        msg.attributes = {}
+    const spanCtx = batchSpan.context()
+    const batchTraceId = spanCtx.toTraceId()
+    const batchSpanId = spanCtx.toSpanId()
+    const batchTraceIdUpper = spanCtx._trace.tags['_dd.p.tid']
+    const batchTraceIdHex = BigInt(batchTraceId).toString(16).padStart(16, '0')
+    const batchSpanIdHex = BigInt(batchSpanId).toString(16).padStart(16, '0')
```
BridgeAR:
```suggestion
    const batchTraceIdHex = spanCtx.toTraceId(true)
    const batchSpanIdHex = spanCtx.toSpanId(true)
    const batchTraceIdUpper = spanCtx._trace.tags['_dd.p.tid']
```

We have options to directly get the 128bits :)

packages/datadog-plugin-google-cloud-pubsub/src/producer.js
```diff
@@ -2,42 +2,180 @@
...
+        spanLinkData.map(link => ({
+          trace_id: link.traceId,
+          span_id: link.spanId,
+          flags: link.samplingPriority || 0
+        }))
+      ))
+    }
+
+    messages.forEach((msg, i) => {
+      msg.attributes = msg.attributes || {}
```
BridgeAR: Nit
```suggestion
      msg.attributes ??= {}
```

packages/datadog-plugin-google-cloud-pubsub/src/producer.js
```diff
@@ -2,42 +2,180 @@
...
+
+  bindFinish (ctx) {
+    if (ctx.batchSpan && !ctx.batchSpan._duration) ctx.batchSpan.finish()
+    return super.bindFinish(ctx)
+  }
+
+  bindError (ctx) {
+    if (ctx.error && ctx.batchSpan) {
+      ctx.batchSpan.setTag('error', ctx.error)
+      ctx.batchSpan.finish()
```
BridgeAR: This seems a bit off to me: I would expect the error to just handle the error case and to have the finish call to happen separately.

packages/datadog-plugin-google-cloud-pubsub/src/producer.js
```diff
@@ -2,42 +2,180 @@
...
+  }
+
+  #extractSpanLink (attrs) {
+    if (!attrs?.['x-datadog-trace-id'] || !attrs['x-datadog-parent-id']) return null
+
+    const lowerHex = BigInt(attrs['x-datadog-trace-id']).toString(16).padStart(16, '0')
+    const spanIdHex = BigInt(attrs['x-datadog-parent-id']).toString(16).padStart(16, '0')
+    const traceIdHex = attrs['_dd.p.tid']
+      ? attrs['_dd.p.tid'] + lowerHex
+      : lowerHex.padStart(32, '0')
```
BridgeAR: Should we just create a new ID from the read string and use the toString method of that instead? I believe that would also work

packages/datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription.js
```diff
@@ -0,0 +1,193 @@
...
+      finished = true
+      if (error) {
+        span.setTag('error', error)
+      }
+      span.finish()
+    }
+
+    res.once('finish', () => finishOnce())
+    res.once('close', () => finishOnce())
+    res.once('error', (err) => finishOnce(err))
```
BridgeAR: I think this should actually be part of the instrumentation, not the plugin, since it handles the data

packages/datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription.js
```diff
@@ -0,0 +1,193 @@
...
+      messageId: req.headers['x-goog-pubsub-message-id'],
+      publishTime: req.headers['x-goog-pubsub-publish-time']
+    }
+
+    const topicName = req.headers['pubsub.topic'] || 'push-subscription-topic'
+    return { message, subscription, attrs: req.headers, topicName }
+  }
+
+  #extractContext (messageData) {
+    return this.tracer.extract('text_map', messageData.attrs) ?? undefined
```
BridgeAR: Is the `?? undefined` needed? It would otherwise return potentially `null`, while that seems fine as far as I can tell from the call sites.

packages/datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription.js
```diff
@@ -0,0 +1,193 @@
...
+    return new SpanContext({
+      traceId,
+      spanId: parentId,
+      tags
+    })
+  }
+
+  #createDeliverySpan (messageData, parentContext, linkContext) {
+    const { message, subscription, topicName, attrs } = messageData
+    const subscriptionName = subscription?.split('/').pop() ?? subscription
```
BridgeAR: Nit: .lastIndexOf() + .slice() would be faster

packages/datadog-plugin-http/src/index.js
```diff
@@ -1,16 +1,31 @@
...
+const { enableGCPPubSubPushSubscription } = require('../../dd-trace/src/serverless')
+const log = require('../../dd-trace/src/log')
 
 class HttpPlugin extends CompositePlugin {
   static id = 'http'
   static get plugins () {
-    return {
-      server: HttpServerPlugin,
-      client: HttpClientPlugin
+    const plugins = {}
```
BridgeAR, line 18:
```suggestion
    const plugins = {
      server: HttpServerPlugin,
      client: HttpClientPlugin,
    }
```

And then there is no need for the assignment below.

packages/datadog-plugin-google-cloud-pubsub/src/consumer.js
```diff
@@ -2,30 +2,222 @@
...
+  }
+
+  #reconstructPubSubRequestContext (attrs) {
+    const traceIdLower = attrs['_dd.pubsub_request.trace_id']
+    const spanId = attrs['_dd.pubsub_request.span_id']
+    const traceIdUpper = attrs['_dd.p.tid']
+
+    if (!traceIdLower || !spanId) return null
+
+    try {
```
BridgeAR: I believe this try catch is actually not needed

packages/datadog-plugin-google-cloud-pubsub/src/consumer.js
```diff
@@ -2,30 +2,222 @@
...
+  }
+
   bindStart (ctx) {
     const { message } = ctx
     const subscription = message._subscriber._subscription
-    const topic = subscription.metadata && subscription.metadata.topic
-    const childOf = this.tracer.extract('text_map', message.attributes) || null
+    const topic = (subscription.metadata && subscription.metadata.topic) ||
+                  (message.attributes && message.attributes['pubsub.topic']) ||
+                  (message.attributes && message.attributes['gcloud.project_id']
```
BridgeAR:
```suggestion
    const topic = subscription?.metadata?.topic || (message.attributes &&
                           (message.attributes['pubsub.topic'] || message.attributes['gcloud.project_id']
```

packages/datadog-plugin-google-cloud-pubsub/src/consumer.js
```diff
@@ -2,30 +2,222 @@
...
+                    ? `projects/${message.attributes['gcloud.project_id']}/topics/unknown`
+                    : null)
+
+    const batchRequestTraceId = message.attributes?.['_dd.pubsub_request.trace_id']
+    const batchRequestSpanId = message.attributes?.['_dd.pubsub_request.span_id']
+    const batchSize = message.attributes?.['_dd.batch.size']
+    const batchIndex = message.attributes?.['_dd.batch.index']
+
+    let childOf = this.tracer.extract('text_map', message.attributes) || null
```
BridgeAR: Is the `|| null` needed?

packages/datadog-plugin-google-cloud-pubsub/src/consumer.js
```diff
@@ -2,30 +2,222 @@
...
+
+    const isFirstMessage = batchIndex === '0' || batchIndex === 0
+    if (isFirstMessage && batchRequestSpanId) {
+      const pubsubRequestContext = this.#reconstructPubSubRequestContext(message.attributes)
+      if (pubsubRequestContext) {
+        childOf = pubsubRequestContext
+      }
+    }
+
+    const topicName = topic ? topic.split('/').pop() : subscription.name.split('/').pop()
```
BridgeAR: Nit for perf using .lastIndexOf + .slice

packages/datadog-plugin-google-cloud-pubsub/src/consumer.js
```diff
@@ -2,30 +2,222 @@
...
+      metrics['pubsub.batch.size'] = Number.parseInt(batchSize, 10)
+    }
+    if (batchIndex !== undefined) {
+      metrics['pubsub.batch.message_index'] = Number.parseInt(batchIndex, 10)
+      metrics['pubsub.batch.index'] = Number.parseInt(batchIndex, 10)
+    }
+
+    if (batchSize && batchIndex !== undefined) {
+      const index = Number.parseInt(batchIndex, 10)
+      const size = Number.parseInt(batchSize, 10)
```
BridgeAR: I think we can move this up so that we only have to get the number once. Are these entries always integers? If yes, I rather use `Number(string)` instead of `parseInt` due to the quirks of it.

packages/datadog-plugin-google-cloud-pubsub/src/consumer.js
```diff
@@ -2,30 +2,222 @@
...
+          span_id: pubsubRequestSpanId,
+          flags: 0
+        }]))
+      }
+
+      if (batchSize) {
+        span.setTag('pubsub.batch.size', Number.parseInt(batchSize, 10))
+      }
+      if (batchIndex) {
+        span.setTag('pubsub.batch.index', Number.parseInt(batchIndex, 10))
```
BridgeAR: We can reuse the numbers from above :)

packages/datadog-plugin-google-cloud-pubsub/src/consumer.js
```diff
@@ -2,30 +2,222 @@
...
+// Use WeakMap + WeakRef + FinalizationRegistry pattern for automatic GC cleanup
+// No TTL needed - GC handles cleanup naturally even if acknowledge() never called
+const messageToContext = new WeakMap() // WeakMap<Message, context> - auto-cleanup on GC
+const ackIdToMessage = new Map() // Map<ackId, WeakRef<Message>> - doesn't prevent GC
+
+// FinalizationRegistry cleans up Map entries when Messages are GC'd
+// This handles network failures, crashes, or any case where acknowledge() isn't called
+const ackMapCleanup = new FinalizationRegistry((ackId) => {
+  ackIdToMessage.delete(ackId)
+})
```
BridgeAR: I think this could use a explanation of what is needed for what as I am having trouble following the overall logic / reasoning by just reading the code 😅

packages/datadog-instrumentations/src/google-cloud-pubsub.js
```diff
@@ -124,33 +154,124 @@ addHook({ name: '@google-cloud/pubsub', versions: ['>=1.2'] }, (obj) => {
...
+  const messageContexts = new WeakMap()
 
   shimmer.wrap(LeaseManager.prototype, '_dispense', dispense => function (message) {
-    if (receiveStartCh.hasSubscribers) {
-      ctx.message = message
-      return receiveStartCh.runStores(ctx, dispense, this, ...arguments)
-    }
-    return dispense.apply(this, arguments)
+    const ctx = { message }
+    messageContexts.set(message, ctx)
```
rochdev, line 201: Ideally this would only be needed on the plugin side, but that can be improved in a future iteration.

rochdev (approved):
There are still a few things not ideal with this PR, but at least the structure is now correct, so they can be addressed later:

- The plugin should be responsible for reconnecting, not the instrumentation, so there shouldn't be a WeakMap on the instrumentation side.
- A unit test for memory leaks is not really trustable, we should instead rely on a real-world test.
- Context management in general could be simplified by relying on the plugin system to do most of the work.
- Injection and extraction should be favoured over manual reparenting.
- The plugin shouldn't manually manipulate trace/span IDs as there is no guarantee that the format will stay the same in the future.

BridgeAR (approved):

BridgeAR (approved):

PR #6267, opened 2025-08-12
Eagerly delete elements from the ctx WeakMap

szegedi:
### What does this PR do?
Eagerly deletes elements from a `WeakMap` keyed by callback context objects when we know we're finished with them.

### Motivation
We're seeing some sawtooth pattern RSS usage when timelines are enabled in one of our relenv test apps, and this _might_ be the culprit. It doesn't hurt proactively deleting elements from it.

An alternative implementation would be to bind our data in the context with a private symbol intead of using the weak map. I think the weak map itself has its purpose defeated as the key is also strongly reachable through the value (as "startEvent") in #5673.

### DIFF
--- a/.github/workflows/flakiness.yml
+++ b/.github/workflows/flakiness.yml
@@ -1,6 +1,7 @@
-name: Flakiness Report
+name: '[Flakiness Report]'
 
 on:
+  workflow_dispatch:
   schedule:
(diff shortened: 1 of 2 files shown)


BridgeAR (commented):
This will likely fix the immediate issue while I would favor the alternative approach suggested.

PR #6270, opened 2025-08-12
bug(tracing): fix moleculer client meta propagation

### LINKED ISSUE #6262: [BUG]:  Moleculer js called action context is lost
### Tracer Version(s)

5.24.0 => 5.62.0

### Node.js Version(s)

20.10.0

### Bug Report

Moleculer js has a feature where you can set and send back the meta from the called action to the caller , as said in the [docs](https://moleculer.services/docs/0.13/actions.html#Metadata), but dd trace breaks that feature.

### Reproduction Code

With these actions as an example 
``` 
async first(ctx) {
            await ctx.call("test.second", null, { meta: {
                a: "John"
            }});

            console.log(ctx.meta);
            // Prints: { a: "Doe" }
        },
        second(ctx) {
            // Modify meta
            ctx.meta.a = "Doe";
        }
```

Without datadog tracer we get Doe as the value in first action, but when we use dd-trace we still get John. 

The reason is the way moleculer js implements the context is they add a ctx property to the returned promise (not value) of [the broker call handler result](https://github.com/moleculerjs/moleculer/blob/master/src/service-broker.js#L1246) and then the context of the called action [is merged with the caller's context](https://github.com/moleculerjs/moleculer/blob/master/src/context.js#L291).

In our code, patching this code https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-instrumentations/src/moleculer/client.js#L24 to look like this solves the issue of context being lost, although it is just a hotfix not a proper solution:

```
      promise
        .then(
          result => {
            finishChannel.publish({ broker, ctx })
            return result
          },
          error => {
            errorChannel.publish(error)
            finishChannel.publish({ broker, ctx })
            throw error
          }
        )
        return promise
```

instead of  (original not working code)
```
return promise
        .then(
          result => {
            finishChannel.publish({ broker, ctx })
            return result
          },
          error => {
            errorChannel.publish(error)
            finishChannel.publish({ broker, ctx })
            throw error
          }
        )
```

### Error Logs

_No response_

### Tracer Config

_No response_

### Operating System

_No response_

### Bundling

Unsure

wconti27:
### What does this PR do?
Fixes broken promise chain for moleculer client calls. Adds regression test case. Fixes #6262 

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-plugin-moleculer/test/index.spec.js
+++ b/packages/datadog-plugin-moleculer/test/index.spec.js
@@ -331,6 +331,46 @@ describe('Plugin', () => {
           expect(spanId.toString()).to.equal(parentId.toString())
         })
       })
+      describe('meta propagation', () => {
+        before(() => agent.load('moleculer', {
+          meta: true
+        }))
+
+        before(async () => {
+          const { ServiceBroker } = require(`../../../versions/moleculer@${version}`).get()
+          broker = new ServiceBroker({
+            nodeID: `server-${process.pid}`,
+            logger: false
+          })
+
+          broker.createService({
+            name: 'test',
+            actions: {
+              async first (ctx) {
+                await ctx.call('test.second', null, {
+                  meta: {
+                    a: 'John'
+                  }
+                })
+                expect(ctx.meta.a).to.equal('Doe')
+              },
+              second (ctx) {
+                ctx.meta.a = 'Doe'
+              }
+            }
+          })
+
+          return broker.start()
+        })
+
+        after(() => broker.stop())
+
+        after(() => agent.close({ ritmReset: false }))
+
+        it('should propagate meta from child to parent', done => {
+          broker.call('test.first').then(() => done()).catch(done)
+        })
+      })
     })
   })
 })
--- a/packages/datadog-instrumentations/src/moleculer/client.js
+++ b/packages/datadog-instrumentations/src/moleculer/client.js
@@ -21,7 +21,7 @@ function wrapCall (call) {
       ctx.promiseCtx = promise.ctx
       ctx.broker = broker
 
-      return promise
+      promise
         .then(
           result => {
             finishChannel.publish(ctx)
@@ -34,6 +34,7 @@ function wrapCall (call) {
             throw error
           }
         )
+      return promise
     })
   }
 }


packages/datadog-plugin-moleculer/test/index.spec.js
```diff
@@ -331,6 +331,46 @@ describe('Plugin', () => {
...
+          broker.createService({
+            name: 'test',
+            actions: {
+              async first (ctx) {
+                await ctx.call('test.second', null, {
+                  meta: {
+                    a: 'John'
+                  }
+                })
+                expect(ctx.meta.a).to.equal('Doe')
```
BridgeAR: Nit: please use Node.js assert module instead. We discussed in the guild that we'd like to migrate to it at a later point, so we should stop adding new chai calls :)

packages/datadog-plugin-moleculer/test/index.spec.js
```diff
@@ -331,6 +331,46 @@ describe('Plugin', () => {
...
+              },
+              second (ctx) {
+                ctx.meta.a = 'Doe'
+              }
+            }
+          })
+
+          return broker.start()
```
BridgeAR: Please move this to the actual test. A before hook should never contain the actual assertions.

BridgeAR (commented):
LGTM with the comments being addressed. Thank you for looking into this @wconti27!

packages/datadog-plugin-moleculer/test/index.spec.js
wconti27: done.

packages/datadog-plugin-moleculer/test/index.spec.js
wconti27: done.

packages/datadog-plugin-moleculer/test/index.spec.js
```diff
@@ -177,7 +187,32 @@ describe('Plugin', () => {
...
+              expect(spans[0].meta).to.have.property('moleculer.context.node_id', `server-${process.pid}`)
+              expect(spans[0].meta).to.have.property('moleculer.context.request_id')
+              expect(spans[0].meta).to.have.property('moleculer.context.service', 'error')
+              expect(spans[0].meta).to.have.property('moleculer.namespace', 'multi')
+              expect(spans[0].meta).to.have.property('moleculer.node_id', `server-${process.pid}`)
+              expect(spans[0].metrics).to.have.property('network.destination.port', port)
+            }).then(done, done)
+
+            broker.call('error.error').catch(error => {
+              assert.strictEqual(error.message, 'Invalid number')
```
BridgeAR: If this fails, the error would end up as unhandled rejection. So instead of using assert, we could just create an error and call done(error). While that in itself is also problematic, due to potentially calling done more than once.

That's why I recommend to use promises instead of the callback for done.

packages/datadog-plugin-moleculer/test/index.spec.js
```diff
@@ -331,6 +366,51 @@ describe('Plugin', () => {
...
+                })
+                return ctx.meta.a
+              },
+              second (ctx) {
+                ctx.meta.a = 'Doe'
+              }
+            }
+          })
+
+          return broker.start()
```
BridgeAR, line 410: I would normally move this inside of the test itself, since the part that we want to test is also part of the setup.

I would only keep the creation of the broker in there.

This is a bit opinionated in this case, it just feels cleaner to me.

packages/datadog-plugin-moleculer/test/index.spec.js
```diff
@@ -331,6 +366,51 @@ describe('Plugin', () => {
...
+
+        after(() => agent.close({ ritmReset: false }))
+
+        it('should propagate meta from child to parent', done => {
+          broker.call('test.first')
+            .then(value => {
+              assert.strictEqual(value, 'Doe')
+              done()
+            })
+            .catch(done)
```
BridgeAR:
```suggestion
        it('should propagate meta from child to parent', async () => {
          const value = await broker.call('test.first')
          assert.strictEqual(value, 'Doe')
```

BridgeAR (commented):
Almost LGTM :)

packages/datadog-plugin-moleculer/test/index.spec.js
wconti27: `createService` must be called on the broker before `broker.start()` so we must do it in the test setup.

packages/datadog-plugin-moleculer/test/index.spec.js
wconti27: Makes sense, I fixed this.

packages/datadog-plugin-moleculer/test/index.spec.js
wconti27: Makes sense, I fixed this. 👍

PR #6271, opened 2025-08-12
collect express endpoints

IlyasShabi:
- Instrumenting `express.application.all()` and `route().all()` emits a single canonical `'*'` endpoint because Router’s `.all()` wrapping is unreliable [router.js#L3](https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-instrumentations/src/router.js#L3), so we handle wildcard at the Express layer.
- Router cannot determine mount prefixes for nested routes, so we instrument `express.application.use()` to resolve full paths with prefixes (e.g., `/v1/nested/:id`) when routers are mounted.
- Router only annotates layers with `ddMatchers`; Express reads these matcher strings to join mount segments deterministically without regex parsing.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/integration-tests/appsec/endpoints-collection.spec.js
+++ b/integration-tests/appsec/endpoints-collection.spec.js
@@ -10,7 +10,7 @@ describe('Endpoints collection', () => {
     this.timeout(process.platform === 'win32' ? 90000 : 30000)
 
     sandbox = await createSandbox(
-      ['fastify'],
+      ['express', 'fastify'],
       false
     )
 
@@ -45,12 +45,18 @@ describe('Endpoints collection', () => {
 
       // Nested routes with Router
       { method: 'PUT', path: '/v1/nested/:id' },
+      { method: '*', path: '/v1/nested' },
 
       // Deeply nested routes
       { method: 'GET', path: '/api/sub/deep' },
       { method: 'HEAD', path: '/api/sub/deep' },
       { method: 'POST', path: '/api/sub/deep/:id' },
     ]
+
+    if (framework === 'express') {
+      expectedEndpoints.push({ method: 'CONNECT', path: '/connect-test' })
+      expectedEndpoints.push({ method: '*', path: '/multi-method' })
+    }
 
     return expectedEndpoints
   }
@@ -73,6 +79,9 @@ describe('Endpoints collection', () => {
       const expectedEndpoints = getExpectedEndpoints(framework)
       const endpointsFound = []
       const isFirstFlags = []
+      // Express: 3 messages (adds routes at runtime)
+      // Fastify: 2 messages (routes at startup)
+      const expectedMessageCount = framework === 'express' ? 3 : 2
 
       await agent.assertTelemetryReceived(msg => {
         const { payload } = msg
@@ -91,7 +100,7 @@ describe('Endpoints collection', () => {
             })
           }
         }
-      }, 'app-endpoints', 5_000, 2)
+      }, 'app-endpoints', 10_000, expectedMessageCount)
 
       const trueCount = isFirstFlags.filter(v => v === true).length
       expect(trueCount).to.equal(1)
@@ -111,6 +120,10 @@ describe('Endpoints collection', () => {
       await agent?.stop()
     }
   }
+
+  it('should send express endpoints via telemetry', async () => {
+    await runEndpointTest('express')
+  })
 
   it('should send fastify endpoints via telemetry', async () => {
     await runEndpointTest('fastify')
--- a/packages/datadog-instrumentations/src/express.js
+++ b/packages/datadog-instrumentations/src/express.js
@@ -5,7 +5,10 @@ const shimmer = require('../../datadog-shimmer')
 const { addHook, channel } = require('./helpers/instrument')
 const tracingChannel = require('dc-polyfill').tracingChannel
 
+const METHODS = require('http').METHODS.map(v => v.toLowerCase())
+
 const handleChannel = channel('apm:express:request:handle')
+const routeAddedChannel = channel('apm:express:route:added')
 
 function wrapHandle (handle) {
   return function handleWithTrace (req, res) {
@@ -57,8 +60,121 @@ function wrapResponseRender (render) {
   }
 }
 
+function wrapAppAll (all) {
+  return function wrappedAll (path) {
+    if (routeAddedChannel.hasSubscribers) {
+      routeAddedChannel.publish({
+        method: '*',
+        path
+      })
+    }
+
+    return all.apply(this, arguments)
+  }
+}
+
+// Wrap app.route() to instrument Route object
+function wrapAppRoute (route) {
+  return function wrappedRoute (path) {
+    const routeObj = route.apply(this, arguments)
+
+    if (routeAddedChannel.hasSubscribers && typeof path === 'string') {
+      // Wrap the .all() first
+      if (typeof routeObj.all === 'function') {
+        shimmer.wrap(routeObj, 'all', (original) => function wrappedRouteAll () {
+          routeAddedChannel.publish({
+            method: '*',
+            path
+          })
+
+          return original.apply(this, arguments)
+        })
+      }
+
+      // Wrap each HTTP method
+      METHODS.forEach(method => {
+        if (typeof routeObj[method] === 'function') {
+          shimmer.wrap(routeObj, method, (original) => function wrapMethod () {
+            routeAddedChannel.publish({
+              method,
+              path
+            })
+
+            return original.apply(this, arguments)
+          })
+        }
+      })
+    }
+
+    return routeObj
+  }
+}
+
+function joinPath (base, path) {
+  if (!base || base === '/') return path || '/'
+  if (!path || path === '/') return base
+  return base + path
+}
+
+function wrapAppUse (use) {
+  return function wrappedUse () {
+    if (arguments.length < 2) {
+      return use.apply(this, arguments)
+    }
+
+    // Check if second argument has a stack (likely a router)
+    const mountPath = arguments[0]
+    const router = arguments[1]
+
+    if (typeof mountPath === 'string' && router?.stack?.length && routeAddedChannel.hasSubscribers) {
+      collectRoutesFromRouter(router, mountPath)
+    }
+
+    return use.apply(this, arguments)
+  }
+}
+
+function collectRoutesFromRouter (router, prefix) {
+  if (!router?.stack?.length) return
+
+  router.stack.forEach(layer => {
+    if (layer.route) {
+      // This layer has a direct route
+      const route = layer.route
+      const fullPath = joinPath(prefix, route.path)
+
+      Object.keys(route.methods).forEach(method => {
+        if (route.methods[method] && method !== '_all') {
+          routeAddedChannel.publish({
+            method,
+            path: fullPath
+          })
+        }
+      })
+
+      if (route.methods._all) {
+        routeAddedChannel.publish({
+          method: '*',
+          path: fullPath
+        })
+      }
+    } else if (layer.handle?.stack?.length) {
+      // This layer contains a nested router
+      // Prefer matchers (from router.js) when available to resolve the exact mount path
+      const matchers = layer.ddMatchers
+      const mountPath = (Array.isArray(matchers) && matchers.length) ? matchers[0].path : ''
+
+      const nestedPrefix = joinPath(prefix, mountPath)
+      collectRoutesFromRouter(layer.handle, nestedPrefix)
+    }
+  })
+}
+
 addHook({ name: 'express', versions: ['>=4'] }, express => {
   shimmer.wrap(express.application, 'handle', wrapHandle)
+  shimmer.wrap(express.application, 'all', wrapAppAll)
+  shimmer.wrap(express.application, 'route', wrapAppRoute)
+  shimmer.wrap(express.application, 'use', wrapAppUse)
 
   shimmer.wrap(express.response, 'json', wrapResponseJson)
   shimmer.wrap(express.response, 'jsonp', wrapResponseJson)
--- a/packages/datadog-instrumentations/src/router.js
+++ b/packages/datadog-instrumentations/src/router.js
@@ -80,6 +80,9 @@ function createWrapRouterMethod (name) {
 
       layerMatchers.set(layer, matchers)
 
+      // Save matchers to consume it later on express.js for endpoint discovery
+      layer.ddMatchers = matchers
+
       if (layer.route) {
         METHODS.forEach(method => {
           if (typeof layer.route.stack === 'function') {
--- a/packages/dd-trace/src/telemetry/endpoints.js
+++ b/packages/dd-trace/src/telemetry/endpoints.js
@@ -4,6 +4,7 @@ const dc = require('dc-polyfill')
 const { sendData } = require('./send-data')
 
 const fastifyRouteCh = dc.channel('apm:fastify:route:added')
+const expressRouteCh = dc.channel('apm:express:route:added')
 
 let config
 let application
@@ -16,6 +17,7 @@ let updateRetryData
  * Map key is `${METHOD} ${PATH}`, value is { method, path }
  */
 const pendingEndpoints = new Map()
+const wildcardEndpoints = new Set()
 let flushScheduled = false
 let isFirstPayload = true
 
@@ -50,6 +52,27 @@ function onFastifyRoute (routeData) {
     for (const method of methods) {
       recordEndpoint(method, routeOptions.path)
     }
+  }
+}
+
+function onExpressRoute ({ method, path }) {
+  if (!method || !path) return
+
+  // If wildcard already recorded for this path, skip specific methods
+  if (wildcardEndpoints.has(path)) return
+
+  // If this is a wildcard event, record it and mark path as wildcarded
+  if (method === '*') {
+    wildcardEndpoints.add(path)
+    recordEndpoint('*', path)
+    return
+  }
+
+  recordEndpoint(method, path)
+
+  // Express automatically adds HEAD support for GET routes
+  if (method.toUpperCase() === 'GET') {
+    recordEndpoint('HEAD', path)
   }
 }
 
@@ -113,10 +136,12 @@ function start (_config = {}, _application, _host, getRetryDataFunction, updateR
   updateRetryData = updateRetryDatafunction
 
   fastifyRouteCh.subscribe(onFastifyRoute)
+  expressRouteCh.subscribe(onExpressRoute)
 }
 
 function stop () {
   fastifyRouteCh.unsubscribe(onFastifyRoute)
+  expressRouteCh.unsubscribe(onExpressRoute)
 
   pendingEndpoints.clear()
   flushScheduled = false
--- a/integration-tests/appsec/endpoints-collection/express.js
+++ b/integration-tests/appsec/endpoints-collection/express.js
@@ -1,0 +1,52 @@
+'use strict'
+
+const tracer = require('dd-trace')
+tracer.init({
+  flushInterval: 0
+})
+
+const express = require('express')
+const app = express()
+
+// Basic routes
+app.get('/users', (_, res) => res.send('ok'))
+app.post('/users', (_, res) => res.send('ok'))
+app.put('/users/:id', (_, res) => res.send('ok'))
+app.delete('/users/:id', (_, res) => res.send('ok'))
+app.patch('/users/:id', (_, res) => res.send('ok'))
+app.options('/users', (_, res) => res.send('ok'))
+
+// Additional methods
+app.trace('/trace-test', (_, res) => res.send('ok'))
+app.head('/head-test', (_, res) => res.send('ok'))
+app.connect('/connect-test', (_, res) => res.send('ok'))
+
+// Using app.route()
+app.route('/multi-method')
+  .post((_, res) => res.send('ok'))
+  .all((_, res) => res.send('ok'))
+
+// Wildcard route
+app.all('/wildcard', (_, res) => res.send('ok'))
+
+// Nested routes with Router
+const apiRouter = express.Router()
+apiRouter.put('/nested/:id', (_, res) => res.send('ok'))
+apiRouter.all('/nested', (_, res) => res.send('ok'))
+app.use('/v1', apiRouter)
+
+// Add endpoint during runtime
+setTimeout(() => {
+  // Deeply nested routes
+  const deepRouter = express.Router()
+  const subRouter = express.Router()
+  subRouter.get('/deep', (_, res) => res.send('ok'))
+  subRouter.post('/deep/:id', (_, res) => res.send('ok'))
+  deepRouter.use('/sub', subRouter)
+  app.use('/api', deepRouter)
+}, 1_000)
+
+const server = app.listen(0, '127.0.0.1', () => {
+  const port = server.address().port
+  process.send({ port })
+})
--- a/integration-tests/appsec/endpoints-collection/fastify.js
+++ b/integration-tests/appsec/endpoints-collection/fastify.js
@@ -33,6 +33,7 @@ app.all('/wildcard', async (_, reply) => reply.send('ok'))
 // Nested routes with Router
 app.register(async function (router) {
   router.put('/nested/:id', async (_, reply) => reply.send('ok'))
+  router.all('/nested', async (_, reply) => reply.send('ok'))
 }, { prefix: '/v1' })
 
 // Deeply nested routes
--- a/packages/dd-trace/test/telemetry/endpoints.spec.js
+++ b/packages/dd-trace/test/telemetry/endpoints.spec.js
@@ -9,6 +9,7 @@ const originalSetImmediate = global.setImmediate
 
 describe('endpoints telemetry', () => {
   const fastifyRouteCh = dc.channel('apm:fastify:route:added')
+  const expressRouteCh = dc.channel('apm:express:route:added')
   const application = 'test'
   const host = 'host'
 
@@ -23,7 +24,7 @@ describe('endpoints telemetry', () => {
       const config = { appsec: { apiSecurity: { endpointCollectionEnabled: true } } }
       endpoints.start(config)
 
-      expect(subscribe).to.have.been.calledOnce
+      expect(subscribe).to.have.been.calledTwice
     })
   })
 
@@ -99,6 +100,31 @@ describe('endpoints telemetry', () => {
 
       expect(firstPayload).to.have.property('is_first', true)
       expect(Boolean(secondPayload.is_first)).to.equal(false)
+    })
+
+    it('should record express route and add HEAD for GET', () => {
+      expressRouteCh.publish({ method: 'GET', path: '/test' })
+
+      scheduledCallbacks.forEach(cb => cb())
+
+      expect(sendData).to.have.been.calledOnce
+      const payload = sendData.firstCall.args[4]
+      const resources = payload.endpoints.map(e => e.resource_name)
+      expect(resources).to.include('GET /test')
+      expect(resources).to.include('HEAD /test')
+    })
+
+    it('should record express wildcard and ignore subsequent specific methods for same path', () => {
+      expressRouteCh.publish({ method: '*', path: '/all' })
+      expressRouteCh.publish({ method: 'GET', path: '/all' })
+      expressRouteCh.publish({ method: 'POST', path: '/all' })
+
+      scheduledCallbacks.forEach(cb => cb())
+
+      expect(sendData).to.have.been.calledOnce
+      const payload = sendData.firstCall.args[4]
+      const resources = payload.endpoints.map(e => e.resource_name)
+      expect(resources).to.deep.equal(['* /all'])
     })
 
     it('should record fastify wildcard when all methods provided', () => {


packages/datadog-instrumentations/src/express.js
```diff
@@ -57,8 +60,125 @@ function wrapResponseRender (render) {
...
+            method: '*',
+            path
+          })
+
+          return original.apply(this, arguments)
+        })
+      }
+
+      // Wrap each HTTP method
+      METHODS.forEach(method => {
```
rochdev: Is using `METHODS` needed here? Aren't all methods on the route HTTP methods?

packages/datadog-instrumentations/src/express.js
```diff
@@ -57,8 +60,125 @@ function wrapResponseRender (render) {
...
+
+    if (typeof mountPath === 'string' && router?.stack?.length && routeAddedChannel.hasSubscribers) {
+      collectRoutesFromRouter(router, mountPath)
+    }
+
+    return use.apply(this, arguments)
+  }
+}
+
+function collectRoutesFromRouter (router, prefix) {
```
rochdev: If this operates on the router, doesn't it also apply to the `router` library? Should it be moved over there like most of the Express instrumentation already is?

packages/datadog-instrumentations/src/express.js
```diff
@@ -57,8 +60,125 @@ function wrapResponseRender (render) {
...
+  if (!router?.stack?.length) return
+
+  router.stack.forEach(layer => {
+    if (layer.route) {
+      // This layer has a direct route
+      const route = layer.route
+      const fullPath = joinPath(prefix, route.path)
+
+      Object.keys(route.methods).forEach(method => {
```
rochdev: I think this could be done in just one step with a condition only on the method which would result in simpler code. Same for the similar code above.

packages/datadog-instrumentations/src/express.js
uurien: Most of them, yes, all of them, no. There are methods like: `_handles_method`, `_options`, `dispatch`, `all`:
express4: https://github.com/expressjs/express/blob/4.x/lib/router/route.js#L58
express5 (router lib): https://github.com/pillarjs/router/blob/master/lib/route.js#L54

packages/datadog-instrumentations/src/express.js
uurien: With the current approach, this is designed to use in the final: `app.use(router)` to export all the possible routes in the router. But I can see some uncovered cases with this approach, things like:

```javascript
const router = express.Router()
router.get('/deep1', (req, res) => res.end(''))
const router2 = express.Router()
router2.get('/superdeep', (req, res) => res.end(''))

router.use('/nested', router2)
app.use('/one', router) 
/* 
    /one/deep1 extracted, this is supported 
    /one/nested/superdeep, extracted also here
*/

router.post('/deep2', (req, res) => res.end('')) 
// router updates once the `app.use` is executed is not supported
```

I'm going to try to find a solution, but it doesn't seem trivial as the same router can be used twice for different paths and calculating the whole path from router would be not possible

packages/datadog-instrumentations/src/express.js
```diff
@@ -56,8 +59,105 @@ function wrapResponseRender (render) {
...
+
+      for (const p of paths) {
+        routeAddedChannel.publish({
+          method: '*',
+          path: p instanceof RegExp ? p.toString() : p
+        })
+      }
+    }
+
+    return all.apply(this, arguments)
```
uurien: We usually do this with opposite logic:

```suggestion
    if (!routeAddedChannel.hasSubscribers) return all.apply(this, arguments)

    const paths = Array.isArray(path) ? path : [path]

    for (const p of paths) {
      routeAddedChannel.publish({
        method: '*',
        path: p instanceof RegExp ? p.toString() : p
      })
    }

    return all.apply(this, arguments)
  }
```

packages/datadog-instrumentations/src/router.js
```diff
@@ -80,6 +80,9 @@ function createWrapRouterMethod (name) {
 
       layerMatchers.set(layer, matchers)
 
+      // Save matchers to consume it later on express.js for endpoint discovery
+      layer.ddMatchers = matchers
```
uurien: we shouldn't write properties in libraries objects. We should use a WeakMap<layer, matchers> instead. In this case, as this is to share info with express router, this could be in a external file required by express.js and router.js

integration-tests/appsec/endpoints-collection.spec.js
```diff
@@ -73,16 +55,43 @@ describe('Endpoints collection', () => {
...
+      expectedEndpoints.push({ method: 'HEAD', path: '/all-methods' })
+      expectedEndpoints.push({ method: 'TRACE', path: '/all-methods' })
+      expectedEndpoints.push({ method: 'DELETE', path: '/all-methods' })
+      expectedEndpoints.push({ method: 'OPTIONS', path: '/all-methods' })
+      expectedEndpoints.push({ method: 'PATCH', path: '/all-methods' })
+      expectedEndpoints.push({ method: 'PUT', path: '/all-methods' })
+      expectedEndpoints.push({ method: 'POST', path: '/all-methods' })
+      expectedEndpoints.push({ method: 'GET', path: '/wildcard/*' })
+      expectedEndpoints.push({ method: 'HEAD', path: '/wildcard/*' })
+      // Wildcard routes
```
CarlesDD:
```suggestion
      expectedEndpoints.push({ method: 'HEAD', path: '/wildcard/*' })

      // Wildcard routes
```

integration-tests/appsec/endpoints-collection.spec.js
```diff
@@ -14,7 +14,7 @@ describe('Endpoints collection', () => {
     this.timeout(process.platform === 'win32' ? 90000 : 30000)
 
     sandbox = await createSandbox(
-      ['fastify'],
+      ['express', 'fastify'],
       false
```
CarlesDD: I think the second argument could be omitted since `false` is the default value

integration-tests/appsec/endpoints-collection.spec.js
```diff
@@ -73,16 +52,44 @@ describe('Endpoints collection', () => {
...
-      { method: 'HEAD', path: '/wildcard/*' },
-      { method: 'GET', path: '*' },
-      { method: 'HEAD', path: '*' },
-
       { method: 'GET', path: '/later' },
       { method: 'HEAD', path: '/later' },
     ]
 
+    if (framework === 'fastify') {
+      expectedEndpoints.push({ method: 'OPTIONS', path: '/users/:id?' })
```
simon-id: you can call push once with multiple parameters, or call concat once with one array for all entries

packages/datadog-instrumentations/src/express.js
```diff
@@ -56,8 +66,108 @@ function wrapResponseRender (render) {
...
+      })
+    }
+
+    return routeObj
+  }
+}
+
+function wrapAppUse (use) {
+  return function wrappedUse () {
+    if (arguments.length >= 2) {
```
uurien: `mountPath` is optional, application can do:

```javascript
app.use('/path', function middleware() {})
app.use('/path', function middleware1() {}, function middleware2() {})
app.use(function middleware1() {}, function middleware2() {})
app.use('/path', router)
app.use(router)
// and I think also it could do
app.use('/path', router1, router2)
app.use(router1, router2)
```

docs: https://expressjs.com/en/5x/api.html#app.use

packages/datadog-instrumentations/src/router.js
```diff
@@ -141,6 +151,17 @@ function createWrapRouterMethod (name) {
...
       const router = original.apply(this, arguments)
 
+      // Handle nested router mounting for 'use' method
+      if (original.name === 'use' && arguments.length >= 2) {
+        const [mountPath, nestedRouter] = arguments
+
+        if (typeof mountPath === 'string' && nestedRouter && typeof nestedRouter === 'function') {
+          const parentPath = getRouterMountPath(this)
+          const fullMountPath = joinPath(parentPath, mountPath)
+          setRouterMountPath(nestedRouter, fullMountPath)
```
uurien: Isn't this value totally ignored when `app.use(router)` is used later?

```
router1 // router not used in app use

router1.use('/path', router2)
router2.use('/path2', router3) // this adds <router3, /path/path2> to the Map
// ...
app.use('/root', router1) // this is going to calculate again all the routes in the router1 tree, and fix them because the one saved is invalid
```

If this is usefull only when the app.use has been used, shouldn't this be into the if below?

packages/datadog-instrumentations/src/router.js
```diff
@@ -149,6 +170,34 @@ function createWrapRouterMethod (name) {
         routeAddedChannel.publish({ topOfStackFunc: methodWithTrace, layer: this.stack.at(-1) })
       }
 
+      // Publish only if this router was mounted by app.use() (prevents early '/sub/...')
+      if (expressRouteAddedChannel.hasSubscribers && isAppMounted(this) && this.stack && this.stack.length > offset) {
```
uurien: Can the same router be used in two different routes?
```javascript
myRouter
app.use('/path1', myRouter)
app.use('/path2', myRouter)
myRouter.get('/hi', (req,res)=>{})
```

IlyasShabi: Yes we report both endpoints `/path1/hi` and `/path2/hi`

packages/datadog-instrumentations/src/router.js
IlyasShabi: Fixed

packages/datadog-instrumentations/src/express.js
IlyasShabi: We support last two use cases you mentioned now

packages/datadog-instrumentations/src/express.js
```diff
@@ -3,8 +3,18 @@
...
+  joinPath,
+  getLayerMatchers,
+  normalizeMethodName,
+  markAppMounted
+} = require('./helpers/router-state')
+
+const METHODS = [...require('http').METHODS.map(v => v.toLowerCase()), 'all']
 
 const handleChannel = channel('apm:express:request:handle')
+const routeAddedChannel = channel('apm:express:add:route')
```
simon-id: nit: makes more sense in this syntax
```suggestion
const routeAddedChannel = channel('apm:express:route:add')
```

packages/datadog-instrumentations/src/express.js
```diff
@@ -56,8 +66,115 @@ function wrapResponseRender (render) {
...
+    return all.apply(this, arguments)
+  }
+}
+
+// Wrap app.route() to instrument Route object
+function wrapAppRoute (route) {
+  return function wrappedRoute (path) {
+    const routeObj = route.apply(this, arguments)
+
+    if (routeAddedChannel.hasSubscribers && typeof path === 'string') {
```
simon-id: app.route() doesn't support regex or array paths ?

simon-id: can invert the logic to save one layer of indentation
```suggestion
    if (!routeAddedChannel.hasSubscribers || typeof path !== 'string') return routeObj
```

packages/datadog-instrumentations/src/express.js
```diff
@@ -56,8 +66,115 @@ function wrapResponseRender (render) {
...
+  }
+}
+
+function wrapAppUse (use) {
+  return function wrappedUse () {
+    if (arguments.length === 0) return use.apply(this, arguments)
+
+    // Check if first argument is a mount path or a router/middleware
+    const firstArg = arguments[0]
+    const hasMountPath = typeof firstArg === 'string'
```
simon-id: no regex or array for app.use() ?

packages/datadog-instrumentations/src/express.js
```diff
@@ -56,8 +66,115 @@ function wrapResponseRender (render) {
...
+      const mountPath = typeof layer.path === 'string'
+        ? layer.path
+        : getLayerMatchers(layer)?.[0]?.path || ''
+
+      const nestedPrefix = joinPath(prefix, mountPath)
+      // Set the mount path for the nested router
+      setRouterMountPath(layer.handle, nestedPrefix)
+      markAppMounted(layer.handle)
+      // Recursively collect from nested routers
+      collectRoutesFromRouter(layer.handle, nestedPrefix)
```
simon-id: are the route.path and layer.path in there always string ? can they be other things like regex or arrays ?

simon-id: should we have an infinite recursion protection ?

packages/datadog-instrumentations/src/express.js
uurien: I think this is because the one you are suggesting is already in use for not exactly the same purpose. Maybe we can create some tag name more specific, I dunno, `apm:express:route:defined` maybe?

packages/dd-trace/src/plugins/util/stacktrace.js
```diff
@@ -90,6 +90,7 @@ function parseLine (stack, start, end) {
   [fileName, lineNumber, columnNumber, index] = result
 
   if (isNodeModulesFrame(fileName)) return
+  if (isDDInstrumentationFile(fileName)) return
```
simon-id: why is this needed now suddenly ? or is it just accesory ?

simon-id: do we really want to hide ourselves from stacktraces ? it's a bit misleading no ?

packages/dd-trace/src/telemetry/endpoints.js
```diff
@@ -4,6 +4,7 @@ const dc = require('dc-polyfill')
 const { sendData } = require('./send-data')
 
 const fastifyRouteCh = dc.channel('apm:fastify:route:added')
+const expressRouteCh = dc.channel('apm:express:add:route')
```
simon-id: it's even more clear the name discrepancie here when you look at the current fastify one
```suggestion
const fastifyRouteCh = dc.channel('apm:fastify:route:added')
const expressRouteCh = dc.channel('apm:express:route:added')
```

packages/datadog-instrumentations/src/router.js
```diff
@@ -149,6 +159,45 @@ function createWrapRouterMethod (name) {
...
+        // Handle nested router mounting for 'use' method
+        if (original.name === 'use' && arguments.length >= 2) {
+          const [mountPath, nestedRouter] = arguments
+
+          if (typeof mountPath === 'string' && nestedRouter && typeof nestedRouter === 'function') {
```
simon-id: same question, can mountPath be something else than string ?

packages/datadog-instrumentations/src/router.js
```diff
@@ -149,6 +159,45 @@ function createWrapRouterMethod (name) {
...
+            const parentPath = getRouterMountPath(this)
+            const fullMountPath = joinPath(parentPath, mountPath)
+            setRouterMountPath(nestedRouter, fullMountPath)
+          }
+        }
+
+        const mountPath = getRouterMountPath(this)
+
+        if (mountPath) {
+          const layer = this.stack[this.stack.length - 1]
```
simon-id: can stack be empty array ? if so, accessing -1 index will cause the array to try to get an out of bound value and make the whole object slower for the whole process IIRC

packages/datadog-instrumentations/src/router.js
```diff
@@ -149,6 +159,45 @@ function createWrapRouterMethod (name) {
...
+
+          if (layer?.route) {
+            const route = layer.route
+            const fullPath = joinPath(mountPath, route.path)
+
+            METHODS.forEach(method => {
```
simon-id: is this foreach something we should put in a function since it's used in the express file too ?

packages/datadog-instrumentations/src/router.js
uurien: Can you add a test? Because this is weird, iiuc each time `app.use` is executed, in `router-state.js` we are adding a register to the weakMap `<router, path>`. So the second time app.use is executed, the first path is overridden by the second.

integration-tests/appsec/endpoints-collection.spec.js
```diff
@@ -73,16 +52,57 @@ describe('Endpoints collection', () => {
...
     ]
 
+    if (framework === 'fastify') {
+      expectedEndpoints.push(
+        { method: 'OPTIONS', path: '/users/:id?' },
+
+        // Route with regex - not supported in express5
+        { method: 'DELETE', path: '/regex/:hour(^\\d{2})h:minute(^\\d{2})m' },
+
+        { method: 'OPTIONS', path: '/users/:id?' }, // Added with addHttpMethod
```
simon-id: duplicated line
```suggestion
```

integration-tests/appsec/endpoints-collection.spec.js
```diff
@@ -73,16 +52,57 @@ describe('Endpoints collection', () => {
...
+
+        // All supported methods route
+        { method: 'GET', path: '/all-methods' },
+        { method: 'HEAD', path: '/all-methods' },
+        { method: 'TRACE', path: '/all-methods' },
+        { method: 'DELETE', path: '/all-methods' },
+        { method: 'OPTIONS', path: '/all-methods' },
+        { method: 'PATCH', path: '/all-methods' },
+        { method: 'PUT', path: '/all-methods' },
+        { method: 'POST', path: '/all-methods' },
```
simon-id, line 75: you're missing mkcol no ?

```suggestion
        { method: 'MKCOL', path: '/all-methods' }, // Added with addHttpMethod
```

integration-tests/appsec/endpoints-collection.spec.js
```diff
@@ -73,16 +52,57 @@ describe('Endpoints collection', () => {
...
+        { method: 'GET', path: '/wildcard/*' },
+        { method: 'HEAD', path: '/wildcard/*' },
+
+        // Wildcard routes
+        { method: 'GET', path: '*' },
+        { method: 'HEAD', path: '*' }
```
simon-id: restoring the previous order
```suggestion
        // Wildcard routes
        { method: 'GET', path: '/wildcard/*' },
        { method: 'HEAD', path: '/wildcard/*' },
        { method: 'GET', path: '*' },
        { method: 'HEAD', path: '*' }
```

integration-tests/appsec/endpoints-collection.spec.js
```diff
@@ -35,34 +32,16 @@ describe('Endpoints collection', () => {
...
       { method: 'TRACE', path: '/trace-test' },
       { method: 'HEAD', path: '/head-test' },
 
-      // Custom method
-      { method: 'MKCOL', path: '/example/near/:lat-:lng/radius/:r' },
-
       // Using app.route()
       { method: 'POST', path: '/multi-method' },
       { method: 'PUT', path: '/multi-method' },
       { method: 'PATCH', path: '/multi-method' },
```
simon-id, line 43: are they not supposed to be overriden by the .all() in express ?

IlyasShabi: No they are not supposed to be overridden by `.all()`

packages/datadog-instrumentations/src/express.js
IlyasShabi: Not for app.use()

packages/datadog-instrumentations/src/express.js
IlyasShabi: Fixed with tests

packages/datadog-instrumentations/src/express.js
IlyasShabi: Fixed with tests
Array: https://github.com/DataDog/dd-trace-js/pull/6271/files#diff-2a5147680161266d4943566685afab2ff9983d34a39b78101a7adfdbc838c4f0R34
Regex: https://github.com/DataDog/dd-trace-js/pull/6271/files#diff-2a5147680161266d4943566685afab2ff9983d34a39b78101a7adfdbc838c4f0R31

IlyasShabi: Fixed with tests, now we do not report related endpoints

packages/dd-trace/src/plugins/util/stacktrace.js
IlyasShabi: This is adding a new line to Code origin which is not related to that feature, so we skip our code

packages/datadog-instrumentations/src/express.js
simon-id: image
you sure ?

packages/dd-trace/src/telemetry/endpoints.js
simon-id: now we have fastify "route.added" and express "route.add" :eyes:

packages/datadog-instrumentations/src/router.js
IlyasShabi: The guard `this.stack.length > offset` means the current stack has at least one new entry so we will not index undefined

packages/datadog-instrumentations/src/express.js
IlyasShabi: Screenshot 2025-10-03 at 14 45 20

Same thing with Regex

IlyasShabi: I should add a leading / to make it work, will fix it

packages/datadog-instrumentations/src/router.js
IlyasShabi: I added a test for this use case:
https://github.com/DataDog/dd-trace-js/pull/6271/files#diff-2a5147680161266d4943566685afab2ff9983d34a39b78101a7adfdbc838c4f0R105

packages/datadog-instrumentations/src/express.js
IlyasShabi: Fixed
Array of paths: https://github.com/DataDog/dd-trace-js/pull/6271/files#diff-2a5147680161266d4943566685afab2ff9983d34a39b78101a7adfdbc838c4f0R76
Regex: https://github.com/DataDog/dd-trace-js/pull/6271/files#diff-2a5147680161266d4943566685afab2ff9983d34a39b78101a7adfdbc838c4f0R82

packages/datadog-instrumentations/src/express.js
IlyasShabi: https://github.com/DataDog/dd-trace-js/pull/6271/files#diff-2a5147680161266d4943566685afab2ff9983d34a39b78101a7adfdbc838c4f0R113

packages/datadog-instrumentations/src/router.js
IlyasShabi: Fixed

packages/datadog-instrumentations/src/router.js
IlyasShabi: It's here https://github.com/DataDog/dd-trace-js/pull/6271/files#diff-bb1346844b5542c09adb69ca6c535c756fd5bb3baa4d5c0ba8e131257fecbe3aR139 now

integration-tests/appsec/endpoints-collection.spec.js
IlyasShabi: Merge conflict issue, fixed

integration-tests/appsec/endpoints-collection.spec.js
IlyasShabi: Merge conflict issue, fixed

integration-tests/appsec/endpoints-collection.spec.js
IlyasShabi: Merge conflict issue, fixed

packages/dd-trace/src/telemetry/endpoints.js
IlyasShabi: Fixed

packages/dd-trace/src/plugins/util/stacktrace.js
uurien: @watson could you review the changes in this file? Don't need to approve the whole PR if you are not confident, but a comment with an OK for this file would be appreciated.

Added because new methods are instrumented in express and router and them could add a new stack line between code origin instrumentation and customer code.

packages/datadog-plugin-express/src/code_origin.js
```diff
@@ -19,6 +19,7 @@ class ExpressCodeOriginForSpansPlugin extends Plugin {
     })
 
     this.addSub('apm:express:route:added', ({ topOfStackFunc, layer }) => {
+      if (!layer) return
```
IlyasShabi, line 22: @watson since Im using the same DC channel I added this condition to avoid unnecessary work for this subscriber

packages/datadog-instrumentations/src/helpers/router-helper.js
```diff
@@ -0,0 +1,176 @@
...
+
+  stack.delete(router)
+  return false
+}
+
+function wrapRouteMethodsAndPublish (route, paths, publish) {
+  if (!route || !paths?.length || typeof publish !== 'function') return
+
+  const uniquePaths = [...new Set(paths.filter(Boolean))]
+  if (!uniquePaths.length) return
```
watson: I'm not sure how hot this return path is, but if you wanted to speed it up, you could abort early before creating the array and the Set in the line above, but storing the result of `paths.filter(Boolean)` in a constant and making this check on that instead, so that it happens prior to instantiating the Set and the array.

packages/datadog-instrumentations/src/helpers/router-helper.js
```diff
@@ -0,0 +1,176 @@
...
+      }
+    }
+  }
+
+  stack.delete(router)
+  return false
+}
+
+function wrapRouteMethodsAndPublish (route, paths, publish) {
+  if (!route || !paths?.length || typeof publish !== 'function') return
```
watson: `wrapRouteMethodsAndPublish` is not an exposed API and I don't think there's anywhere where we call it without a `publish` function. Same with `paths`: As far as I can see we always call with the a `paths` object. So why not simply:

```suggestion
  if (!route || !paths.length) return
```

packages/datadog-instrumentations/src/helpers/router-helper.js
```diff
@@ -0,0 +1,176 @@
...
+  const uniquePaths = [...new Set(paths.filter(Boolean))]
+  if (!uniquePaths.length) return
+
+  METHODS.forEach(method => {
+    if (typeof route[method] !== 'function') return
+
+    shimmer.wrap(route, method, (originalMethod) => function wrappedRouteMethod () {
+      for (const path of uniquePaths) {
+        publish({
+          method: normalizeMethodName(method),
```
watson: `method` is constant, so no need to keep normalizing it:

```suggestion
    const normalizedMethod = normalizeMethodName(method)

    shimmer.wrap(route, method, (originalMethod) => function wrappedRouteMethod () {
      for (const path of uniquePaths) {
        publish({
          method: normalizedMethod,
```

Alternatively, if we think it's not likely that the wrapped method is called, we could move it into the body of it to make the normalization lazy, like so:

```suggestion

    shimmer.wrap(route, method, (originalMethod) => function wrappedRouteMethod () {
      const normalizedMethod = normalizeMethodName(method)
      for (const path of uniquePaths) {
        publish({
          method: normalizedMethod,
```

packages/datadog-instrumentations/src/router.js
```diff
@@ -5,6 +5,21 @@ const pathToRegExp = require('path-to-regexp')
 const shimmer = require('../../datadog-shimmer')
 const { addHook, channel } = require('./helpers/instrument')
 
+// Routes registered on routers before they are mounted must emit on the
+// express channel so telemetry treats them like top-level express routes.
+const expressRouteAddedChannel = channel('apm:express:route:added')
```
watson: This channel will be the same as `routeAddedChannel` if `createWrapRouterMethod` is called with `express` as its argument. I find that a bit confusing

packages/datadog-instrumentations/src/router.js
```diff
@@ -149,6 +163,47 @@ function createWrapRouterMethod (name) {
...
+                })
+              })
+            }
+          }
+        }
+
+        const mountPaths = getRouterMountPaths(this)
+
+        if (mountPaths.length) {
+          const layer = this.stack[this.stack.length - 1]
```
watson:
```suggestion
          const layer = this.stack.at(-1)
```

packages/datadog-instrumentations/src/router.js
```diff
@@ -149,6 +163,47 @@ function createWrapRouterMethod (name) {
         routeAddedChannel.publish({ topOfStackFunc: methodWithTrace, layer: this.stack.at(-1) })
       }
 
+      // Publish only if this router was mounted by app.use() (prevents early '/sub/...')
+      if (expressRouteAddedChannel.hasSubscribers && isAppMounted(this) && this.stack && this.stack.length > offset) {
```
watson:
```suggestion
      if (expressRouteAddedChannel.hasSubscribers && isAppMounted(this) && this.stack?.length > offset) {
```

packages/datadog-instrumentations/src/router.js
```diff
@@ -149,6 +163,47 @@ function createWrapRouterMethod (name) {
...
+        // Handle nested router mounting for 'use' method
+        if (original.name === 'use' && arguments.length >= 2) {
+          const { mountPaths, startIdx } = extractMountPaths(arguments)
```
watson: I don't understand this: `arguments` here refer to the arguments given to the `methodWithTrace` callback. The `extractMountPaths` function only picks the first element of `arguments`, meaning `fn`, which we already have a reference to in the `methodWithTrace` function. So why not just change `extractMountPaths` to just take only the first argument? That way you could just call `extractMountPaths(fn)` on this line and `extractMountPaths` didn't have to do `const firstArg = args[0]`.

In general, there's something to be gained refactoring this whole method from using `arguments` to using `...args` as that is more performant. But that is technically outside the scope of this PR, so I'm ok with leaving a larger refactoring a later PR.

packages/datadog-instrumentations/src/helpers/router-helper.js
```diff
@@ -0,0 +1,176 @@
...
+const layerMatchers = new WeakMap() // to store layer matchers
+const appMountedRouters = new WeakSet() // to track routers mounted via app.use()
+
+const METHODS = [...require('http').METHODS.map(v => v.toLowerCase()), 'all']
+
+function joinPath (base, path) {
+  if (!base || base === '/') return path || '/'
+  if (!path || path === '/') return base
+  if (base.endsWith('/') && path.startsWith('/')) return base + path.slice(1)
+  return base + path
```
watson, line 36: Just to be sure, there's no situation where `path` might not start with `/`? If it doesn't, and `base` doesn't end with `/`, we'll have a problem. But if that's never going to happen, I'm fine with ignoring that case here.

packages/datadog-instrumentations/src/router.js
```diff
@@ -149,6 +163,47 @@ function createWrapRouterMethod (name) {
...
+              const nestedRouter = arguments[i]
+
+              if (!nestedRouter || typeof nestedRouter !== 'function') continue
+
+              parentPaths.forEach(parentPath => {
+                mountPaths.forEach(normalizedMountPath => {
+                  const fullMountPath = joinPath(parentPath, normalizedMountPath)
+                  setRouterMountPath(nestedRouter, fullMountPath)
+                })
+              })
```
watson: Nit: Use for...of when possible, there's a slight performance improvement, but obviously in this case it doesn't matter much:

```suggestion
              for (const parentPath of parentPaths) {
                for (const normalizedMountPath of mountPaths) {
                  const fullMountPath = joinPath(parentPath, normalizedMountPath)
                  setRouterMountPath(nestedRouter, fullMountPath)
                }
              }
```

packages/datadog-instrumentations/src/express.js
```diff
@@ -56,8 +69,117 @@ function wrapResponseRender (render) {
...
+        setRouterMountPath(router, normalizedMountPath)
+
+        if (!skipCollection && routeAddedChannel.hasSubscribers) {
+          collectRoutesFromRouter(router, normalizedMountPath)
+        }
+      }
+    }
+
+    return use.apply(this, arguments)
+  }
```
watson: It's better to use `...args` instead of relying on the magic `arguments` variable:

```suggestion
  return function wrappedUse (...args) {
    if (args.length === 0) return use.call(this)

    // Get mount argument and use it to register each router against the exact paths Express will use.
    const { mountPaths, startIdx } = extractMountPaths(args)
    const pathsToRegister = mountPaths.length ? mountPaths : ['/']

    for (let i = startIdx; i < args.length; i++) {
      const router = args[i]

      if (!router || typeof router !== 'function') continue

      markAppMounted(router)

      // Avoid enumerating routes for routers that contain cycles.
      // Express will refuse those at runtime, but collecting them here could loop forever.
      let skipCollection = false
      if (routeAddedChannel.hasSubscribers) {
        skipCollection = hasRouterCycle(router)
      }

      for (const mountPath of pathsToRegister) {
        const normalizedMountPath = mountPath || '/'
        setRouterMountPath(router, normalizedMountPath)

        if (!skipCollection && routeAddedChannel.hasSubscribers) {
          collectRoutesFromRouter(router, normalizedMountPath)
        }
      }
    }

    return use.apply(this, args)
  }
```

packages/datadog-instrumentations/src/helpers/router-helper.js
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// Normalize route definitions coming from Express into a string representation
+function normalizeRoutePath (path) {
+  if (path == null) return null
+  if (path instanceof RegExp) return path.toString()
+  if (typeof path === 'string') return path
```
watson: Nit: swap in order of most likely to happen in the real world (I think a string is more likely than an regex, feel free to ignore if I'm wrong):

```suggestion
  if (typeof path === 'string') return path
  if (path instanceof RegExp) return path.toString()
```

packages/datadog-instrumentations/src/helpers/router-helper.js
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    const normalized = normalizeRoutePath(current)
+
+    if (normalized !== null) {
+      paths.push(normalized)
+    }
+  }
+
+  return paths
```
watson: No reason to implement our own array flatting logic, when we can just use the built-in method:

```suggestion
  const paths = Array.isArray(path) ? path.flat(Infinity) : [path]
  const result = []

  for (const _path of paths) {
    const normalized = normalizeRoutePath(_path)
    if (normalized !== null) {
      result.push(normalized)
    }
  }

  return result
```

An even more performant version would be this one with a fast-path for when `path` isn't an array:

```suggestion
  if (Array.isArray(path) === false) {
    const normalized = normalizeRoutePath(path)
    return normalized !== null ? [normalized] : []
  }

  const paths = path.flat(Infinity)
  const result = []

  for (const _path of paths) {
    const normalized = normalizeRoutePath(_path)
    if (normalized !== null) {
      result.push(normalized)
    }
  }

  return result
```

watson (commented):
Sorry for all the nits and performance improvement ideas - I hope you don't mind 😅

I think this should be ok from a Code Origin point of view. However, I'd like to add a few more tests to the Express Code Origin test file (`packages/datadog-plugin-express/test/code_origin.spec.js`). Do you mind doing that in this PR?

Here's the change I want to make to the file:

```diff
diff --git a/packages/datadog-plugin-express/test/code_origin.spec.js b/packages/datadog-plugin-express/test/code_origin.spec.js
index e23097cc9..1cc3e3f79 100644
--- a/packages/datadog-plugin-express/test/code_origin.spec.js
+++ b/packages/datadog-plugin-express/test/code_origin.spec.js
@@ -102,6 +102,31 @@ describe('Plugin', () => {
               await assertCodeOrigin('/foo/bar', { line, method: 'testCase', type: 'Context' })
             })
 
+            it('should support .route() routes', async function testCase () {
+              app.get('/route_before', (req, res) => res.end())
+              const route = app.route('/foo')
+              const line = getNextLineNumber()
+              route.get((req, res) => {
+                res.end()
+              })
+              app.get('/route_after', (req, res) => res.end())
+
+              await assertCodeOrigin('/foo', { line, method: 'testCase', type: 'Context' })
+            })
+
+            it('should support Router routes', async function testCase () {
+              const router = express.Router()
+              app.get('/route_before', (req, res) => res.end())
+              const line = getNextLineNumber()
+              router.get('/bar', (req, res) => {
+                res.end()
+              })
+              app.use('/foo', router)
+              app.get('/route_after', (req, res) => res.end())
+
+              await assertCodeOrigin('/foo/bar', { line, method: 'testCase', type: 'Context' })
+            })
+
             it('should point to route handler even if passed through a middleware', async function testCase () {
               app.use((req, res, next) => {
                 next()
```

packages/dd-trace/src/telemetry/endpoints.js
```diff
@@ -42,12 +44,32 @@ function onFastifyRoute (routeData) {
...
+  if (wildcardEndpoints.has(path)) return
+
+  // If this is a wildcard event, record it and mark path as wildcarded
+  if (method === '*') {
+    wildcardEndpoints.add(path)
+    recordEndpoint('*', path)
+    return
+  }
+
+  recordEndpoint(method, path)
```
simon-id: avoid some code duplication and imo easier to read
```suggestion
  recordEndpoint(method, path)

  // If this is a wildcard event, record it and mark path as wildcarded
  if (method === '*') {
    wildcardEndpoints.add(path)
    return
  }
```

integration-tests/appsec/endpoints-collection.spec.js
```diff
@@ -35,34 +32,16 @@ describe('Endpoints collection', () => {
...
 
-      // All supported methods route
-      { method: 'GET', path: '/all-methods' },
-      { method: 'HEAD', path: '/all-methods' },
-      { method: 'TRACE', path: '/all-methods' },
-      { method: 'DELETE', path: '/all-methods' },
-      { method: 'OPTIONS', path: '/all-methods' },
-      { method: 'PATCH', path: '/all-methods' },
-      { method: 'PUT', path: '/all-methods' },
-      { method: 'POST', path: '/all-methods' },
```
simon-id, line 63: why not have those in express ? mkcol excluded but why not the rest ?

integration-tests/appsec/endpoints-collection.spec.js
```diff
@@ -35,34 +32,16 @@ describe('Endpoints collection', () => {
       { method: 'PUT', path: '/users/:id' },
       { method: 'DELETE', path: '/users/:id' },
       { method: 'PATCH', path: '/users/:id/:name' },
-      { method: 'OPTIONS', path: '/users/:id?' },
-
-      // Route with regex
-      { method: 'DELETE', path: '/regex/:hour(^\\d{2})h:minute(^\\d{2})m' },
```
simon-id, line 41: we do have routes with regex in express too no ? why not keep this line too ?

packages/datadog-instrumentations/src/helpers/router-helper.js
```diff
@@ -0,0 +1,176 @@
...
+  return base + path
+}
+
+// Normalize route definitions coming from Express into a string representation
+function normalizeRoutePath (path) {
+  if (path == null) return null
+  if (path instanceof RegExp) return path.toString()
+  if (typeof path === 'string') return path
+
+  return String(path)
```
simon-id, line 45: i don't think this is a common case, but if i have a route with `"/.test"` and another route `/.test/`, they will "normalize" to the same string while matching very different URLs right ? i think that's problematic

packages/datadog-instrumentations/src/helpers/router-helper.js
IlyasShabi:
```
const apiRouter = express.Router()
app.use('/v1', apiRouter)
apiRouter.put('/nested/:id', (_, res) => res.send('ok'))
```
Using this example base do not ends with `/` and reported endpoint will be `/v1/nested/:id`

packages/datadog-instrumentations/src/router.js
IlyasShabi: I agree that this is a little bit confusing and need more explanation. With the following example:
```
const api = express.Router()
api.get('/nested', handler)      // defined before mounting
app.use('/api', api)
```

Those routes only pass through the router instrumentation, so we forward them onto `apm:express:route:added` to keep telemetry aware of `/api/nested`. If we rely on `routeAddedChannel` we will not be able to report only the router local piece

packages/datadog-instrumentations/src/helpers/router-helper.js
IlyasShabi: The normalize function will produce different paths for`"/.test"` and `/.test/` there’s no collision

integration-tests/appsec/endpoints-collection.spec.js
IlyasShabi: We have those in express too https://github.com/DataDog/dd-trace-js/blob/d99619323db02c5c524113443a57421e0790e0aa/integration-tests/appsec/endpoints-collection.spec.js#L90 we handle `all()` as with `*`

integration-tests/appsec/endpoints-collection.spec.js
IlyasShabi: It's not supported in Express5

packages/datadog-instrumentations/src/express.js
IlyasShabi: Fixed

packages/datadog-instrumentations/src/router.js
IlyasShabi: I agree with you on replacing `arguments` with `...args` and refactoring of `extractMountPaths` function. I applied suggested changes

packages/datadog-instrumentations/src/helpers/router-helper.js
IlyasShabi: Done

packages/datadog-instrumentations/src/helpers/router-helper.js
watson: But in that example, path starts with `/`, right? I'm asking if there's a risk that path will not start with `/` when at the same time base doesn't end in `/`. Because that's the only edge case the function doesn't handle. But if that will never happen, there's no need to add the code to handle it.

Using your example above, would this be legal (notice the missing `/` prefix in the put route)?

```js
const apiRouter = express.Router()
app.use('/v1', apiRouter)
apiRouter.put('nested/:id', (_, res) => res.send('ok'))
```

packages/datadog-instrumentations/src/express.js
```diff
@@ -56,8 +66,80 @@ function wrapResponseRender (render) {
...
+        const normalizedMountPath = mountPath || '/'
+        setRouterMountPath(router, normalizedMountPath)
+
+        if (!skipCollection && routeAddedChannel.hasSubscribers) {
+          collectRoutesFromRouter(router, normalizedMountPath)
+        }
+      }
+    }
+
+    return use.apply(this, arguments)
```
watson: Missed one:

```suggestion
    return use.apply(this, args)
```

packages/datadog-instrumentations/src/helpers/router-helper.js
```diff
@@ -0,0 +1,217 @@
...
+      const normalizedMethod = normalizeMethodName(method)
+
+      for (const path of uniquePaths) {
+        publish({
+          method: normalizedMethod,
+          path
+        })
+      }
+
+      return originalMethod.apply(this, arguments)
```
watson: Just removing all references to `arguments`...

```suggestion
    shimmer.wrap(route, method, (originalMethod) => function wrappedRouteMethod (...args) {
      const normalizedMethod = normalizeMethodName(method)

      for (const path of uniquePaths) {
        publish({
          method: normalizedMethod,
          path
        })
      }

      return originalMethod.apply(this, args)
```

packages/datadog-instrumentations/src/express.js
```diff
@@ -56,8 +66,80 @@ function wrapResponseRender (render) {
...
+  return function wrappedAll (path) {
+    if (!routeAddedChannel.hasSubscribers) return all.apply(this, arguments)
+
+    const paths = normalizeRoutePaths(path)
+
+    for (const p of paths) {
+      routeAddedChannel.publish({ method: '*', path: p })
+    }
+
+    return all.apply(this, arguments)
```
watson: Just removing all references to `arguments`...

```suggestion
  return function wrappedAll (...args) {
    if (!routeAddedChannel.hasSubscribers) return all.apply(this, args)

    const paths = normalizeRoutePaths(args[0])

    for (const p of paths) {
      routeAddedChannel.publish({ method: '*', path: p })
    }

    return all.apply(this, args)
```

If you think it helps readability to keep the `path` argument, you can do:

```suggestion
  return function wrappedAll (path, ...rest) {
    if (!routeAddedChannel.hasSubscribers) return all.call(this, path, ...rest)

    const paths = normalizeRoutePaths(path)

    for (const p of paths) {
      routeAddedChannel.publish({ method: '*', path: p })
    }

    return all.call(this, path, ...rest)
```

packages/datadog-instrumentations/src/express.js
```diff
@@ -56,8 +66,80 @@ function wrapResponseRender (render) {
...
+}
+
+// Wrap app.route() to instrument Route object
+function wrapAppRoute (route) {
+  return function wrappedRoute (path) {
+    const routeObj = route.apply(this, arguments)
+
+    if (!routeAddedChannel.hasSubscribers) return routeObj
+
+    const paths = normalizeRoutePaths(path)
```
watson: Just removing all references to `arguments`...

Either:

```suggestion
  return function wrappedRoute (...args) {
    const routeObj = route.apply(this, args)

    if (!routeAddedChannel.hasSubscribers) return routeObj

    const paths = normalizeRoutePaths(args[0])
```

Or:

```suggestion
  return function wrappedRoute (path, ...rest) {
    const routeObj = route.call(this, path, ...rest)

    if (!routeAddedChannel.hasSubscribers) return routeObj

    const paths = normalizeRoutePaths(path)
```

packages/datadog-instrumentations/src/router.js
```diff
@@ -149,6 +165,48 @@ function createWrapRouterMethod (name) {
         routeAddedChannel.publish({ topOfStackFunc: methodWithTrace, layer: this.stack.at(-1) })
       }
 
+      // Publish only if this router was mounted by app.use() (prevents early '/sub/...')
+      if (expressRouteAddedChannel.hasSubscribers && isAppMounted(this) && this.stack?.length > offset) {
+        // Handle nested router mounting for 'use' method
+        if (original.name === 'use' && arguments.length >= 2) {
```
watson: I prefer that you refactor this this `methodWithTrace` while you're here to get rid of all the added `arguments` references.

packages/datadog-instrumentations/test/helpers/router-helper.spec.js
```diff
@@ -0,0 +1,334 @@
...
+    })
+  })
+
+  describe('wrapRouteMethodsAndPublish', () => {
+    it('should wrap route methods and publish for each path', () => {
+      const calls = []
+      const published = []
+      const route = {
+        get () {
+          calls.push({ args: Array.from(arguments), context: this })
```
watson:
```suggestion
        get (...args) {
          calls.push({ args, context: this })
```

packages/datadog-instrumentations/src/router.js
watson: Is that because `routeAddedChannel` is created later (inside of `createWrapRouterMethod`) and not when this module is first loaded?

packages/datadog-instrumentations/src/router.js
IlyasShabi: Done

packages/datadog-instrumentations/src/helpers/router-helper.js
IlyasShabi: Yes, that example is legal Express code it will run without errors. However, the route `apiRouter.put('nested/:id', ...)` will never be accessible because Express does not normalize paths without leading slashes.

Before my fix, we were incorrectly reporting `PUT /v1nested/:id` I fixed `joinPath` function so we correctly report nothing for this endpoint since it's not accessible.

packages/datadog-instrumentations/src/router.js
IlyasShabi: Im using `routeAddedChannel` channel now to report the endpoints, In fact I just needed to subscribe to this channel in `endpoints.js`

packages/datadog-instrumentations/src/helpers/router-helper.js
simon-id: ok let me reiterate, imagine this:
```javascript
app.get('/.htaccess/', handler)
app.get(/.htaccess/, handler)
```
Those two routes match different URLs, but will be normalized to the same string.

packages/dd-trace/src/plugins/util/stacktrace.js
simon-id: i still dont understand this line btw

integration-tests/appsec/endpoints-collection.spec.js
simon-id: ah oui my bad

packages/dd-trace/src/plugins/util/stacktrace.js
simon-id: @watson and @IlyasShabi are looking into this because we confirmed with Watson it was incorrect

packages/datadog-instrumentations/src/helpers/router-helper.js
IlyasShabi: As we talked, we can keep it like this

packages/datadog-instrumentations/src/helpers/router-helper.js
```diff
@@ -0,0 +1,239 @@
...
+// Flatten any route definition into an array of normalized path strings.
+function normalizeRoutePaths (path) {
+  if (path == null) return []
+
+  if (Array.isArray(path) === false) {
+    const normalized = normalizeRoutePath(path)
+    return [normalized]
+  }
+
+  const paths = path.flat(Infinity)
```
simon-id, line 96: if path has a circular reference this is an immediate stack overflow, maybe a big number that's not Infinity would be better ? like i don't see a useful case for Infinity

packages/datadog-instrumentations/test/helpers/router-helper.spec.js
simon-id, line 1: too late to change it now, but next time, for new test files we try to use assert instead of expect ;)

integration-tests/appsec/endpoints-collection.spec.js
```diff
@@ -73,16 +52,83 @@ describe('Endpoints collection', () => {
...
       { method: 'GET', path: '/later' },
       { method: 'HEAD', path: '/later' },
     ]
 
+    if (framework === 'fastify') {
+      expectedEndpoints.push(
+        // Route with regex - not supported in express5
+        { method: 'DELETE', path: '/regex/:hour(^\\d{2})h:minute(^\\d{2})m' },
+
+        { method: 'OPTIONS', path: '/users/:id?' }, // Added with addHttpMethod
```
simon-id: wrong copy paste again :p
```suggestion
        { method: 'OPTIONS', path: '/users/:id?' },
```

packages/datadog-instrumentations/src/helpers/router-helper.js
simon-id: something like that maybe ?
```suggestion
  const paths = path.flat(100)
```

packages/datadog-instrumentations/src/helpers/router-helper.js
```diff
@@ -0,0 +1,239 @@
...
+
+  stack.add(router)
+
+  for (const layer of router.stack) {
+    if (!layer?.route && layer?.handle?.stack?.length) {
+      const hasCycle = hasRouterCycle(layer.handle, stack)
+      if (hasCycle) {
+        stack.delete(router)
+        return true
+      }
```
simon-id, line 189: if there is a cyclic reference, you recursively return true until the top level, like an early exit, so why do you need to delete the router form the stack ? the stack is only a temporary variable, so if you know you're exiting the function you can just let the garbage colector do the thing no ?

packages/datadog-instrumentations/src/helpers/router-helper.js
```diff
@@ -0,0 +1,239 @@
...
+  return false
+}
+
+function wrapRouteMethodsAndPublish (route, paths, publish) {
+  if (!route || !paths.length) return
+
+  const filteredPaths = paths.filter(Boolean)
+  if (!filteredPaths.length) return
+
+  const uniquePaths = [...new Set(filteredPaths)]
```
simon-id: you don't need to create an array out of it, since you just do a for of below, just do the for of on the set directly ?

packages/datadog-instrumentations/src/helpers/router-helper.js
```diff
@@ -0,0 +1,239 @@
...
+    if (normalized !== null) {
+      result.push(normalized)
+    }
+  }
+
+  return result
+}
+
+function setRouterMountPath (router, mountPath) {
+  if (!router || typeof mountPath !== 'string') return
```
simon-id, line 109: can this not be array or regex ?

packages/datadog-instrumentations/src/helpers/router-helper.js
IlyasShabi: I ran some tests and found that Express/Router already flattens these and rejects cyclic arrays, so any issues fail before this helper runs.

packages/datadog-instrumentations/src/helpers/router-helper.js
IlyasShabi: mountPath will always be a string thanks to `extractMountPaths` helper that flattens those structures and `normalizeRoutePaths` stringifies them, so by the time we call setRouterMountPath we’re handing it as string

packages/datadog-instrumentations/src/helpers/router-helper.js
IlyasShabi: `stack.delete(router)` here have no effect yes

packages/datadog-instrumentations/src/helpers/router-helper.js
IlyasShabi: Yes agree, done

packages/datadog-plugin-express/test/code_origin.spec.js
```diff
@@ -150,3 +183,18 @@ describe('Plugin', () => {
...
+    '../src/index',
+    '../src/code_origin'
+  ].map(path => require.resolve(path))
+
+  for (const modulePath of modulePaths) {
+    if (require.cache[modulePath]) {
+      delete require.cache[modulePath]
+    }
+  }
+}
```
simon-id:
```suggestion
  const modulePaths = [
    '../../dd-trace/src/plugins/util/stacktrace',
    '../../datadog-code-origin',
    '../src/index',
    '../src/code_origin'
  ].map(path => require.resolve(path))

function clearCodeOriginCaches () {
  for (const modulePath of modulePaths) {
    delete require.cache[modulePath]
  }
}
```

simon-id: I moved the modulePaths outside the function, and no need to check if the path exists in the cache, delete will gracefully continue if the prop doesn't exists

simon-id (dismissed):
LGTM, but i want @watson approval too

packages/datadog-plugin-express/test/code_origin.spec.js
```diff
@@ -2,7 +2,7 @@
 
 const axios = require('axios')
 const { expect } = require('chai')
-const { describe, it, beforeEach, afterEach, before } = require('mocha')
+const { describe, it, beforeEach, afterEach, before, after } = require('mocha')
```
simon-id:
```suggestion
const { describe, it, beforeEach, afterEach, before } = require('mocha')
```

PR #6272, opened 2025-08-12
Removing hook for express v5 router

crysmags:
### What does this PR do?
This PR will remove the hook for Express v5 router. In v5 the router was separated out into its own module and right now we are wrapping it as well as wrapping the router module. This is resulting in double instrumentation for router and thus double spans. This will remove the instrumentation of the router in express and it will fall to just instrumenting the router. 

### Motivation
There is a current escalation where the customer had double spans.

### Plugin Checklist

- [] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/datadog-instrumentations/src/express.js
CarlesDD, line 1: I see what you're trying to do here, but I believe there's an issue with this change.

By removing the `addHook` blocks that scoped the wrapping logic to specific Express versions, we're now applying the same wrapping logic to `express.Router` regardless of whether it's Express 4 or 5. This isn't safe: in Express 5, Router is a constructor and the wrapping should happen on its prototype, while in Express 4 it's not, so wrapping it directly is necessary.

Applying the same wrap to both versions lead to runtime errors, as seen in the [tests](https://github.com/DataDog/dd-trace-js/actions/runs/16918985746/job/47945749937?pr=6272#step:5:2397).

If the goal is to avoid wrapping `express.Router` entirely in Express 5, we should retain the version-specific `addHook` blocks and simply omit the wrapping for Router in the `>=5.0.0` block.

CarlesDD (changes requested):

packages/datadog-instrumentations/src/express.js
crysmags: @CarlesDD I believe I addressed your issue, I have added in an `addHook` block for express v4 with a comment as well. This is will make it more explicit that for v4 we are wrapping the router and in v5 we will fall to the router instrumentation to wrap the calls.

packages/datadog-plugin-express/test/index.spec.js
```diff
@@ -230,6 +230,9 @@ describe('Plugin', () => {
                 const spans = sort(traces[0])
                 const isExpress4 = semver.intersects(version, '<5.0.0')
                 let index = 0
+                const whichMiddleware = semver.intersects(version, '<5.0.0')
+                  ? 'express'
+                  : 'router'
```
uurien: We have this `senver.intersect` 2 lines above.

```suggestion
                const whichMiddleware = isExpress4 ? 'express' : 'router'
```

packages/datadog-plugin-express/src/code_origin.js
```diff
@@ -22,6 +22,16 @@ class ExpressCodeOriginForSpansPlugin extends Plugin {
       if (layerTags.has(layer)) return
       layerTags.set(layer, entryTags(topOfStackFunc))
     })
+    this.addSub('apm:router:middleware:enter', ({ req, layer }) => {
```
uurien, line 26:
```suggestion

    this.addSub('apm:router:middleware:enter', ({ req, layer }) => {
```

packages/dd-trace/src/appsec/rasp/index.js
```diff
@@ -117,6 +118,7 @@ function enable (config) {
 
   expressMiddlewareError.subscribe(blockOnDatadogRaspAbortError)
   fastifyMiddlewareError.subscribe(blockOnDatadogRaspAbortError)
+  routerMiddlewareError.subscribe(blockOnDatadogRaspAbortError)
```
simon-id, line 121: this also needs to be unsubscribed from in disable() a few lines after

simon-id:
The couchbase failure looks unrelated to this PR as it's failing elsewhere too

packages/dd-trace/src/appsec/rasp/index.js
crysmags: thanks for adding the changes!

uurien (approved):

CarlesDD (approved):

PR #6275, opened 2025-08-13
Extended request data collection via rules

uurien:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/appsec/reporter.js
+++ b/packages/dd-trace/src/appsec/reporter.js
@@ -38,6 +38,8 @@ const config = {
 
 const metricsQueue = new Map()
 
+const extendedDataCollectionRequest = new WeakMap()
+
 // following header lists are ordered in the same way the spec orders them, it doesn't matter but it's easier to compare
 const contentHeaderList = [
   'content-length',
@@ -140,7 +142,7 @@ function filterExtendedHeaders (headers, excludedHeaderNames, tagPrefix, limit =
   return result
 }
 
-function getCollectedHeaders (req, res, shouldCollectEventHeaders, storedResponseHeaders = {}) {
+function getCollectedHeaders (req, res, shouldCollectEventHeaders, storedResponseHeaders = {}, extendedDataCollection) {
   // Mandatory
   const mandatoryCollectedHeaders = filterHeaders(req.headers, REQUEST_HEADERS_MAP)
 
@@ -154,7 +156,9 @@ function getCollectedHeaders (req, res, shouldCollectEventHeaders, storedRespons
   const requestEventCollectedHeaders = filterHeaders(req.headers, EVENT_HEADERS_MAP)
   const responseEventCollectedHeaders = filterHeaders(responseHeaders, RESPONSE_HEADERS_MAP)
 
-  if (!config.headersExtendedCollectionEnabled || config.headersRedaction) {
+  // TODO headersExtendedCollectionEnabled and headersRedaction properties should be deprecated to deleted in next major
+  // should be standard if !redaction and no headers enabled
+  if ((!config.headersExtendedCollectionEnabled || config.headersRedaction) && (!extendedDataCollection || extendedDataCollection.redaction)) {
     // Standard collection
     return Object.assign(
       mandatoryCollectedHeaders,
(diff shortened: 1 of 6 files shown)


index.d.ts
simon-id, line 1: should we add a TODO: delete in next major in config.js for those ?

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -163,12 +184,14 @@ function getCollectedHeaders (req, res, shouldCollectEventHeaders, storedRespons
     )
   }
 
+  const maxHeadersCollected = extendedDataCollection?.max_collected_headers || config.maxHeadersCollected
```
simon-id: in case it's zero ?
```suggestion
  const maxHeadersCollected = extendedDataCollection?.max_collected_headers ?? config.maxHeadersCollected
```

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -75,15 +81,27 @@ const requestHeadersList = [
   ...identificationHeaders
 ]
 
+const authenticationHeadersList = [
```
simon-id: could call it "redactedHeadersList" instead, so we know what it will be used for directly ?

packages/dd-trace/test/appsec/extended-data-collection.express.plugin.spec.js
CarlesDD, line 1: I miss a test to check `Authorization` headers (or the ones should be redacted) are correctly redacted

packages/dd-trace/src/appsec/reporter.js
CarlesDD: Nit: it would be useful to have a TODO comment here to remove `config.maxHeadersCollected` in the next major, as there are in the spots where deprectaed config is used.

Another option could be placing the TODO in the `config` object declaration in L32 instead of having multiple TODO comments scattered throughout the coe.

packages/dd-trace/test/appsec/extended-data-collection.express.plugin.spec.js
uurien: ouch, I was pretty sure that I added :-(, Adding them

packages/dd-trace/src/appsec/reporter.js
uurien: I prefer to have multiple todos, so we don't miss any of then in the future.

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -409,7 +439,10 @@ function reportRequestBody (rootSpan, requestBody) {
     const { truncated, value } = truncateRequestBody(requestBody)
     rootSpan.meta_struct['http.request.body'] = value
     if (truncated) {
-      rootSpan.setTag('_dd.appsec.rasp.request_body_size.exceeded', 'true')
+      const sizeExceedTagKey = comesFromRaspAction
+        ? '_dd.appsec.rasp.request_body_size.exceeded' // TODO old metric to delete in a major
+        : '_dd.appsec.request_body_size.exceeded'
+      rootSpan.setTag(sizeExceedTagKey, 'true')
```
IlyasShabi, line 453: When both `config.raspBodyCollection` (RASP attack) and `extendedDataCollection` are true, `reportRequestBody` is called twice:

1. First: sets body + report `_dd.appsec.rasp.request_body_size.exceeded` tag
2. Second call: exits early because `rootSpan.meta_struct['http.request.body']` already exists

So the tag `_dd.appsec.request_body_size.exceeded` is never set when both conditions are met. is this expected?

CarlesDD (approved):
LGTM!

packages/dd-trace/src/appsec/reporter.js
uurien: Yes, I am aware of that. Anyway I can add an `else` clause adding also the other tag. I'm on it

IlyasShabi (dismissed):

PR #6279, opened 2025-08-13
chore: limit dependency version range to prevent them being flagged

BridgeAR:
Dependabot flags these and no matter that no concrete issue exists, it will limit issues being reported.


.github/dependabot.yml
```diff
@@ -87,6 +89,10 @@ updates:
       - dependencies
       - javascript
       - semver-patch
+    ignore:
+      - dependency-name: "express"
```
rochdev, line 107: We don't want to ignore express in test versions, only for ESBuild.

rochdev (approved):

PR #6273, opened 2025-08-12
ci: update github actions to use node latest, not active

tlhunter:
### What does this PR do?
- updates GitHub CI jobs to use the latest (24) Node.js instead of currently active LTS (22)
- some cases already tested latest and active so I kept those the same

### Motivation
- need to ensure we fixed v24 compat

PR #6283, opened 2025-08-14
ci: fix express node v24 test

tlhunter:
### What does this PR do?
- fixes some failing express tests when running on node v24
- for more info see https://github.com/DataDog/dd-trace-js/pull/6273

### Motivation
- node v24 compat is complete
- however these Express tests are having issues due to 11 year old express not working with node v24

All of the Express errors seem to be the same:

```
TypeError: Cannot read properties of undefined (reading 'last-modified')
    at fresh (/home/runner/work/dd-trace-js/dd-trace-js/versions/express@4.3.0/node_modules/fresh/index.js:29:25)
    at IncomingMessage.fresh (/home/runner/work/dd-trace-js/dd-trace-js/versions/express@4.3.0/node_modules/express/lib/request.js:397:12)
```

This seems like it could be an incompatibility with the `fresh` module and node v24 but I don't see any issues on their tracker.

Local testing:

```sh
$ nvm use 22
$ ./node_modules/.bin/mocha -r "packages/dd-trace/test/setup/mocha.js" packages/datadog-plugin-express/test/index.spec.js
# ok
$ nvm use 24
$ ./node_modules/.bin/mocha -r "packages/dd-trace/test/setup/mocha.js" packages/datadog-plugin-express/test/index.spec.js
# not ok
```

The code that fails is in the fresh package, which is called from express. The issue is that the `req.res._headers` is undefined. Those response headers are an internal field. Maybe it's a version mismatch between fresh and express? Locally when I get the failure it's express 4.0.0 and fresh 0.2.2 which both seem pretty old and perhaps of the same era. Or do we clobber the `Response#_headers` somewhere?

My research suggests response _headers was deprecated ages ago so I'm thinking it's an issue with bad package versions getting installed.

### DIFF
--- a/packages/dd-trace/test/appsec/iast/taint-tracking/sources/taint-tracking.express.plugin.spec.js
+++ b/packages/dd-trace/test/appsec/iast/taint-tracking/sources/taint-tracking.express.plugin.spec.js
@@ -84,6 +84,11 @@ describe('Path params sourcing with express', () => {
   let appListener
 
   withVersions('express', 'express', version => {
+    if (semver.intersects(version, '<=4.10.5') && NODE_MAJOR >= 24) {
+      describe.skip(`refusing to run tests as express@${version} is incompatible with Node.js ${NODE_MAJOR}`)
+      return
+    }
+
     const checkParamIsTaintedAndNext = (req, res, next, param, name) => {
       const store = storage('legacy').getStore()
       const iastContext = iastContextFunctions.getIastContext(store)


tlhunter:
Good news!. It looks like Express.js v4.0.0 is just busted:

```sh
npm init -y && npm install express@4.0.0
node app.js &
curl http://localhost:3000
# TypeError: Cannot read properties of undefined (reading &#39;last-modified&#39;)
```

```javascript
# app.js
const express = require('express')
const app = express()
const port = 3000

app.get('/', (req, res) => {
  res.send('Hello World!')
})

app.listen(port, () => {
  console.log(`Example app listening on port ${port}`)
})
```

I'm sure it wasn't busted at release time but that the fresh package became incompatible at a later point.

PR #6287, opened 2025-08-15
fix(llmobs): properly parse DD_TAGS onto span events' tags

sabrenner:
### What does this PR do?
Properly parse DD_TAGS onto the llmobs span events' tags field.

Internally: sets `DD_TAGS` as a private (but accessible) member on the `config` object for the tracer, as the `tags` property gets "polluted" with other tags besides `DD_TAGS` for LLM Observability, we just want to propagate `DD_TAGS` that are set on the config.

### Motivation
We should have been grabbing these the whole time.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/config.js
+++ b/packages/dd-trace/src/config.js
@@ -823,8 +823,12 @@ class Config {
     const env = setHiddenProperty(this, '_env', {})
     setHiddenProperty(this, '_envUnprocessed', {})
 
+    const ddTags = {}
+    tagger.add(ddTags, parseSpaceSeparatedTags(DD_TAGS))
+    setHiddenProperty(this, '_ddTags', ddTags) // usable as a standalone tag set for other products
+
     tagger.add(tags, parseSpaceSeparatedTags(handleOtel(OTEL_RESOURCE_ATTRIBUTES)))
-    tagger.add(tags, parseSpaceSeparatedTags(DD_TAGS))
+    tagger.add(tags, ddTags)
     tagger.add(tags, DD_TRACE_TAGS)
     tagger.add(tags, DD_TRACE_GLOBAL_TAGS)
 
--- a/packages/dd-trace/src/llmobs/span_processor.js
+++ b/packages/dd-trace/src/llmobs/span_processor.js
@@ -177,7 +177,10 @@ class LLMObsSpanProcessor {
   }
 
   _processTags (span, mlApp, sessionId, error) {
+    const ddTags = this._config._ddTags ?? {}
+
     let tags = {
+      ...ddTags,
       version: this._config.version,
       env: this._config.env,
       service: this._config.service,
@@ -187,6 +190,7 @@ class LLMObsSpanProcessor {
       error: Number(!!error) || 0,
       language: 'javascript'
     }
+
     const errType = span.context()._tags[ERROR_TYPE] || error?.name
     if (errType) tags.error_type = errType
     if (sessionId) tags.session_id = sessionId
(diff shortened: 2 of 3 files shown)


Kyle-Verhoog (approved):
good for llmobs!

packages/dd-trace/src/config.js
```diff
@@ -823,8 +823,12 @@ class Config {
     const env = setHiddenProperty(this, '_env', {})
     setHiddenProperty(this, '_envUnprocessed', {})
 
+    const ddTags = {}
+    tagger.add(ddTags, parseSpaceSeparatedTags(DD_TAGS))
+    setHiddenProperty(this, '_ddTags', ddTags) // usable as a standalone tag set for other products
```
BridgeAR: I suggest we use a getter instead of the underscored property. The underscored ones should be prevented in the future.

BridgeAR:
```suggestion
    this.#parsedDdTags = parseSpaceSeparatedTags(DD_TAGS)
```

And in the class

```js
  get parsedDdTags() {
    return this.#parsedDdTags
  }
```

packages/dd-trace/src/llmobs/span_processor.js
```diff
@@ -177,7 +177,10 @@ class LLMObsSpanProcessor {
   }
 
   _processTags (span, mlApp, sessionId, error) {
+    const ddTags = this._config._ddTags ?? {}
+
     let tags = {
+      ...ddTags,
```
BridgeAR:
```suggestion
    let tags = {
      ...this._config.parsedDdTags,
```

No need for the fallback, spreading undefined will be ignored :)

packages/dd-trace/src/config.js
sabrenner: yes i like this a lot better too!

packages/dd-trace/src/llmobs/span_processor.js
sabrenner: ah yes that's right 😅

packages/dd-trace/src/config.js
sabrenner: added in https://github.com/DataDog/dd-trace-js/pull/6287/commits/2e4c353c8a857d55107f5f7ec1d714b6f7a3a74a, altered slightly so just lmk if it's still off 😄

packages/dd-trace/src/config.js
```diff
@@ -823,8 +833,10 @@ class Config {
     const env = setHiddenProperty(this, '_env', {})
     setHiddenProperty(this, '_envUnprocessed', {})
 
+    tagger.add(this.#parsedDdTags, parseSpaceSeparatedTags(DD_TAGS))
```
BridgeAR, line 836: I think we should actually not use `tagger.add()` here. The reason is that it will just add the object to the first argument. So it's like an assignment of a variable.

sabrenner: it seems like `parseSpaceSeparatedTags` [returns a string](https://github.com/DataDog/dd-trace-js/blob/58216feb39022644d4092e590830b4c9a416f77b/packages/dd-trace/src/config.js#L1559-L1564), so i was thinking it still needed the `tagger.add` to actually parse it into an object, but let me know if i'm misunderstanding!

BridgeAR: Hu, you are absolutely correct! Thank you, I missed that it's not directly parsed to an object.

BridgeAR (approved):
LGTM

PR #6288, opened 2025-08-17
improve esbuild integration test stability

rochdev:
### What does this PR do?

Improve esbuild integration test stability by pinning the latest version and bumping the timeout.

### Motivation

Avoid issues related to new versions of esbuild and hopefully prevent further occurrences of timeouts like https://github.com/DataDog/dd-trace-js/actions/runs/16977715214/job/48130733078

### DIFF
--- a/integration-tests/esbuild/index.spec.js
+++ b/integration-tests/esbuild/index.spec.js
@@ -15,19 +15,15 @@ const TEST_DIR = path.join(__dirname, '.')
 process.chdir(TEST_DIR)
 
 // This should switch to our withVersion helper. The order here currently matters.
-const esbuildVersions = ['latest', '0.16.12']
+const latestVersion = require('../../packages/dd-trace/test/plugins/versions/package.json').dependencies.esbuild
+const esbuildVersions = [latestVersion, '0.16.12']
+
+const timeout = 60_000
 
 esbuildVersions.forEach((version) => {
   describe(`esbuild ${version}`, () => {
     before(() => {
-      chproc.execSync('npm install', {
-        timeout: 1000 * 30
-      })
-      if (version !== 'latest') {
-        chproc.execSync(`npm install esbuild@${version}`, {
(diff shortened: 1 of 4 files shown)


integration-tests/esbuild/index.spec.js
```diff
@@ -15,19 +15,15 @@ const TEST_DIR = path.join(__dirname, '.')
...
+const esbuildVersions = [latestVersion, '0.16.12']
+
+const timeout = 60_000
 
 esbuildVersions.forEach((version) => {
   describe(`esbuild ${version}`, () => {
     before(() => {
-      chproc.execSync('npm install', {
-        timeout: 1000 * 30
-      })
```
watson, line 25: Why did we run this `npm install` before?

BridgeAR: It was just a hack for the setup for now. We should instead use our withVersions helper instead. There is already a comment about that at the top.

packages/dd-trace/test/plugins/versions/package.json
```diff
@@ -85,6 +85,7 @@
     "dd-trace-api": "1.0.0",
     "ejs": "3.1.10",
     "elasticsearch": "16.7.3",
+    "esbuild": "0.25.9",
```
BridgeAR, line 88:
```suggestion
```

It is also listed in the other package.json, so this is just a duplication :)

rochdev: The 2 files don't have the same purpose, shouldn't it also be listed here as the auto-updated cap?

BridgeAR: It would not be picked up like that.

rochdev: Ah I see what you mean now, this is because the same issue was addressed in #6279 as well already.

PR #6299, opened 2025-08-19
Fix log injection and applied sampling rule bug

khanayan123:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/integration-tests/log_injection.spec.js
+++ b/integration-tests/log_injection.spec.js
@@ -1,0 +1,71 @@
+'use strict'
+
+const { FakeAgent, createSandbox, spawnProc, curlAndAssertMessage } = require('./helpers')
+const { assert } = require('chai')
+const path = require('path')
+const { USER_KEEP } = require('../ext/priority')
+
+describe('Log Injection', () => {
+  let agent
+  let proc
+  let sandbox
+  let cwd
+  let app
+  let env
+
+  before(async () => {
+    sandbox = await createSandbox(['express', 'winston'])
+    cwd = sandbox.folder
+    app = path.join(cwd, 'log_injection/index.js')
+  })
+
+  after(async () => {
+    await sandbox.remove()
+  })
+
+  beforeEach(async () => {
+    agent = await new FakeAgent().start()
+  })
+
+  describe('log injection with rule based sampling', () => {
+    beforeEach(async () => {
+      agent = await new FakeAgent().start()
+    })
+
+    afterEach(async () => {
+      proc.kill()
+      await agent.stop()
+    })
+
+    it('should correctly apply rule based sampling when log injection is enabled', async () => {
+      env = {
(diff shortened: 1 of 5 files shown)


BridgeAR (commented):
Functionality wise this is LGTM, I believe the overall fix should be implemented somewhere else.

I think we could for example look into setting the resource.name earlier. If we manage that, the bug should also be fixed and it would likely also fix more situations (I could imagine this is also tricky in case other functionality relies on it as well).
This should for example not come with the performance overhead that this currently comes with.

integration-tests/log_injection.spec.js
```diff
@@ -0,0 +1,71 @@
...
+      }
+      proc = await spawnProc(app, { cwd, env, execArgv: [] })
+      const url = proc.url + '/sampled'
+      return curlAndAssertMessage(agent, url, ({ headers, payload }) => {
+        assert.isArray(payload)
+        assert.strictEqual(payload.length, 1)
+        // Bug: previously got USER_REJECT instead of USER_KEEP when log injection is enabled,
+        // meaning resource rules are not applied & instead global sampling is applied
+        // Now gets USER_KEEP because resource rules are applied
+        assert.propertyVal(payload[0][0].metrics, '_sampling_priority_v1', USER_KEEP)
```
BridgeAR: Nit: please change this to use our helper `assertObjectContains` for partial deep equal checks. That way there is additional debugging information in case of a failure and we don't add additional chai usage, which we want to prevent in the future (using Node.js assert instead).

The same applies to the other test.

integration-tests/log_injection.spec.js
```diff
@@ -0,0 +1,71 @@
...
+      }
+      proc = await spawnProc(app, { cwd, env, execArgv: [] })
+      const url = proc.url + '/sampled'
+      return curlAndAssertMessage(agent, url, ({ headers, payload }) => {
+        assert.isArray(payload)
+        assert.strictEqual(payload.length, 1)
+        // Bug: previously got USER_REJECT instead of USER_KEEP when log injection is enabled,
+        // meaning resource rules are not applied & instead global sampling is applied
+        // Now gets USER_KEEP because resource rules are applied
+        assertObjectContains(payload[0][0].metrics, { _sampling_priority_v1: USER_KEEP })
```
BridgeAR:
```suggestion
        // Bug: previously got USER_REJECT instead of USER_KEEP when log injection is enabled,
        // meaning resource rules are not applied & instead global sampling is applied
        // Now gets USER_KEEP because resource rules are applied
        assertObjectContains(payload, [[{ metrics: { _sampling_priority_v1: USER_KEEP }}]])
```

PR #6308, opened 2025-08-19
feat(aws/sqs): add queue url as resource tag

seungleelee:
### What does this PR do?
This PR attaches a `queue_url` tag to AWS SQS spans, which is already available within the tracer since the queue url is a parameter to the SQS operations via AWS SDK.

### Motivation
Part of a larger effort to better link cloud resources within Datadog. Currently the way in which sqs resources are tagged are inconsistent across different trace libraries, and we need a consistent tag key with the same value format to support this linkage.

[Jira Ticket](https://datadoghq.atlassian.net/browse/CLP-1074) and more [context](https://datadoghq.atlassian.net/wiki/x/toGyLwE) around this effort

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-plugin-aws-sdk/src/services/sqs.js
+++ b/packages/datadog-plugin-aws-sdk/src/services/sqs.js
@@ -101,7 +101,8 @@ class Sqs extends BaseAwsSdkPlugin {
     const tags = {
       'resource.name': `${operation} ${params.QueueName || params.QueueUrl}`,
       'aws.sqs.queue_name': params.QueueName || params.QueueUrl,
-      queuename: queueName
+      queuename: queueName,
+      queue_url: params.QueueUrl
     }
 
     switch (operation) {
--- a/packages/datadog-plugin-aws-sdk/test/sqs.spec.js
+++ b/packages/datadog-plugin-aws-sdk/test/sqs.spec.js
@@ -146,7 +146,8 @@ describe('Plugin', () => {
 
             expect(span.resource.startsWith('sendMessage')).to.equal(true)
             expect(span.meta).to.include({
-              queuename: queueName
+              queuename: queueName,
+              queue_url: QueueUrl
             })
 
             parentId = span.span_id.toString()
@@ -196,7 +197,8 @@ describe('Plugin', () => {
 
             expect(span.resource.startsWith('sendMessageBatch')).to.equal(true)
             expect(span.meta).to.include({
-              queuename: queueName
+              queuename: queueName,
+              queue_url: QueueUrl
             })
 
             parentId = span.span_id.toString()
@@ -369,6 +371,7 @@ describe('Plugin', () => {
 
             expect(span.meta).to.include({
               queuename: queueName,
+              queue_url: QueueUrl,
               aws_service: 'SQS',
               region: 'us-east-1'
             })


packages/datadog-plugin-aws-sdk/src/services/sqs.js
```diff
@@ -101,7 +101,8 @@ class Sqs extends BaseAwsSdkPlugin {
     const tags = {
       'resource.name': `${operation} ${params.QueueName || params.QueueUrl}`,
       'aws.sqs.queue_name': params.QueueName || params.QueueUrl,
-      queuename: queueName
+      queuename: queueName,
+      queue_url: params.QueueUrl
```
BridgeAR: Adding this will of course work, I just wonder if we need the queuename in that case. It is in fact the end of the queue url.

Unrelated: That logic above could ideally be changed for performance by the way using:

```js
const lastLashPosition = params.QueueUrl.lastIndexOf('/')
if (lastSlashPosition !== -1) {
  queueName = params.QueueUrl.slice(lastSlashPosition + 1)
}
```

seungleelee: I agree, but I assumed removing an existing tag may be a breaking change, in case people are querying ddsql by the queuename tag or filtering spans with it?

seungleelee: We ended up going a different direction of generating and attaching the ARN instead of exposing the queue url. Updated the description and the tests accordingly

packages/datadog-plugin-aws-sdk/src/util.js
```diff
@@ -84,8 +84,67 @@ const extractPrimaryKeys = (keyNames, keyValuePairs) => {
...
+ */
+const extractQueueMetadata = queueURL => {
+  if (!queueURL) {
+    return null
+  }
+
+  const parts = queueURL.split('/').filter(Boolean)
+
+  // Check if URL has scheme
+  const hasScheme = parts[0] && parts[0].startsWith('http')
```
BridgeAR: Nit: It's just style and now it's consistent a boolean.

```suggestion
  const hasScheme = Boolean(parts[0]?.startsWith('http'))
```

packages/datadog-plugin-aws-sdk/src/util.js
```diff
@@ -84,8 +84,67 @@ const extractPrimaryKeys = (keyNames, keyValuePairs) => {
...
+
+  const accountId = parts[parts.length - 2]
+  const queueName = parts[parts.length - 1]
+  const host = hasScheme ? parts[1] : parts[0]
+
+  let region = 'us-east-1' // Default region if not found in URL
+  const hostParts = host.split('.')
+  if (host.includes('amazon') && hostParts.length > 3) {
+    // sqs.{region}.amazonaws.com or {region}.queue.amazonaws.com
+    region = hostParts[0] === 'sqs' ? hostParts[1] : hostParts[0]
```
BridgeAR:
```suggestion
  if (host.includes('.amazonaws.com')) {
    // sqs.{region}.amazonaws.com or {region}.queue.amazonaws.com
    const startFrom = host.startsWith('sqs.') ? 4 : 0
    const nextDot = host.indexOf('.', startFrom)
    region = host.slice(startFrom, nextDot)
```

Just a suggestion. This prevents allocating a new array and needing to copy the string and is faster therefore. I do not know how frequently we use this method, since it's a fast operation already and is only beneficial in case it is frequently called.

Looking at the examples: could it actually ever happen that the host is not amazonaws.com?

BridgeAR (approved):
I can't say much about the business side, but from the implementation side it's looking good to me. Just left two suggestions.

packages/datadog-plugin-aws-sdk/src/util.js
seungleelee: for all production use cases no, I'm just doing this check for the localstack test cases that are using localhost endpoints. I still wanted to test the arn generation logic within the integration tests, and `us-east-1` is the default region for localstack tests and legacy queues with no region specified

BridgeAR (approved):
LGTM. About the comment with the check always being safe: I can not tell how such URLs might all look like and if it is possible that .amazonaws.com is included or not.

PR #6315, opened 2025-08-20
Fix Appsec Next.js v15.5 plugin test

CarlesDD:
### What does this PR do?
Adds `outputFileTracingRoot` config to generate standalone server file in the expected directory.

### Motivation
Next.js v15.5 started to generate standalone server file in a different directory if this config is not provided, making the Appsec Next.js plugin test to fail.


simon-id (approved):

CarlesDD:
> Needing such change seems like a breaking change in Next.js to me. Could you check if there is already an issue reported and otherwise open such issue? :)

Setting this option explicitly is needed due to a change in the way Next.js compute root dir (here the [commit](https://github.com/vercel/next.js/commit/925f8a9446ca67551324825033febea3accd6419)). They added a warning stating that the _inferred root directory may not be correct_, what make me think they are aware of this situation.

PR #6316, opened 2025-08-20
ci(llmobs): resolve some test flakiness

sabrenner:
### What does this PR do?
Best attempt at CI flakiness for
- an Azure OpenAI test which occasionally timed out, by adding in some fixtures (they still return a 401 response, only because I didn't have the keys set up correctly, but an actual response is not needed for the test)
- the newly added `ai` integration, which seems to randomly time out for whole versions. this had to do with the `agent.close` configuration we were using, and where we were configuring the fixture

### Motivation
no flakiness in ci!!


packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
```diff
@@ -55,8 +55,7 @@ describe('integrations', () => {
       return agent.close({ ritmReset: false, wipe: true })
     })
 
-    // TODO: Remove the range cap once we support openai 5
```
BridgeAR, line 58: ❤️

BridgeAR (approved):
LGTM

PR #6319, opened 2025-08-20
fix(ai): properly handle partial `experimental_telemetry` configurations to wrap the correct tracer

### LINKED ISSUE #6317: [BUG]: AI SDK Autoinstrumentation
### Tracer Version(s)

5.63.3

### Node.js Version(s)

22.17.0

### Bug Report

Auto instrumentation breaks `AI SDK` `generateObject`. I have deployed the application and verified that traces and logs are sent to Datadog, other endpoints in the API works as expected but when `generateObject` is called an error is thrown with the message: 

```json
"err":{"type":"TypeError","message":"Cannot convert undefined or null to object","stack":"TypeError: Cannot convert undefined or null to object\n    at Function.hasOwn (<anonymous>)\n    at wrapTracer (/dd_tracer/node/node_modules/dd-trace/packages/datadog-instrumentations/src/ai.js:45:14)\n    at generateObject (/dd_tracer/node/node_modules/dd-trace/packages/datadog-instrumentations/src/ai.js:102:5)\n    at _.translateProduct (file:///app/apps/pimo/dist/index.js:24:1217)\n    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)"}
```

Running ESM, distroless.

I am using tsup to bundle but have confirmed that `ai` packages are not bundled:

```ts
import { defineConfig } from 'tsup';

export default defineConfig({
  entry: ['src/index.ts'],
  format: ['esm'],
  target: 'node22',
  platform: 'node',
  bundle: true,
  splitting: false,
  minify: true,
  sourcemap: true,
  clean: true,
  external: [
    'ai',
    '@ai-sdk/azure',
    '@ai-sdk/openai',
    'openai',
    /^@ai-sdk\/.*/,
  ],
});
```

### Reproduction Code

```ts
import { createAzure } from '@ai-sdk/azure';
import { generateObject } from 'ai';
import { z } from 'zod';

const azure = createAzure({
  resourceName: process.env.AZURE_OPENAI_RESOURCE_NAME,
  apiKey: process.env.AZURE_OPENAI_API_KEY,
});

const LlmProductSchema = z.object({
  id: z.string(),
  name: z.string(),
  message: z.string().nullable().describe('Product description'),
});

const { object } = await generateObject({
        model: azure("deployment-name"),
        prompt: "hello",
        schema: LlmProductSchema,
        experimental_telemetry: {
          isEnabled: true,
       },
});

### Error Logs

_No response_

### Tracer Config

```ts

const env = process.env.APP_ENV;

if (env === 'test' || env === 'prod') {
  tracer.init({
    service: process.env.DD_SERVICE ?? 'service-name',
    env: process.env.DD_ENV,
    version: process.env.DD_VERSION,
    runtimeMetrics: true,
    logInjection: true,
    tags: { team: 'team' },
    llmobs: {
      mlApp: 'service-name',
    },
  });
}

export default tracer;
```

In index.ts:
```ts
import './lib/tracer';
```

Other env vars are added to container through bicep and Dockerfile. Logs and traces are passed to DD as expected.

### Operating System

 Darwin JQQ756XK7C 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:51 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T8112 arm64

### Bundling

Unsure

sabrenner:
### For reviewers
For those unfamiliar with how the `ai` instrumentation works, there's a longer description in the [original PR](https://github.com/DataDog/dd-trace-js/pull/5858#issue-3131019166), but the TL;DR is that each of the patched functions (`generateText`, for example) can take in an `experimental_telemetry` object, with an `isEnabled` and `tracer` fields, which is an OTel-compatible tracer. To re-use the existing tracing the core AI SDK does with this tracer, we simply patch that tracer, and emit DC events to start/end/add errors to our own spans at the appropriate points.

### What does this PR do?
We previously assumed a not-partial `experimental_telemetry` configuration, meaning the code only handled paths where the configuration didn't exist, or did but had a `tracer` property. This PR corrects that behavior. It makes sure that:

1. If there is a `experimental_telemetry.isEnabled` configuration set to `false`, we don't attempt any patching
2. If there is no `experimental_telemetry` config, we assume that since `dd-trace` is being used that we want to patch the tracer and enable telemetry for the AI core SDK
3. If there is an `experimental_telemetry` config defined, and `experimental_telemetry.isEnabled` is not set to `false`, we'll make sure it's set to `true`, and if there is no `experimental_telemetry.tracer`, set it to the `noopTracer` default, again with the reasoning that since `dd-trace` is being used that we want to patch the tracer and enable telemetry for the AI core SDK.

### Motivation
Fixes #6317 

MLOB-3700

### DIFF
--- a/packages/dd-trace/test/appsec/next/app-dir/next.config.js
+++ b/packages/dd-trace/test/appsec/next/app-dir/next.config.js
@@ -6,7 +6,8 @@ const nextConfig = {
   experimental: {
     appDir: true
   },
-  output: 'standalone'
+  output: 'standalone',
+  outputFileTracingRoot: __dirname
 }
 
 module.exports = nextConfig


packages/datadog-instrumentations/src/ai.js
```diff
@@ -98,7 +98,20 @@ function wrapWithTracer (fn) {
...
+      return fn.apply(this, arguments)
+    }
+
+    if (experimentalTelemetry == null) {
+      options.experimental_telemetry = { isEnabled: true, tracer: noopTracer }
+    } else {
+      experimentalTelemetry.isEnabled = true
+      if (experimentalTelemetry.tracer == null) {
+        experimentalTelemetry.tracer = noopTracer
+      }
```
BridgeAR: Nit
```suggestion
      experimentalTelemetry.tracer ??= noopTracer
```

BridgeAR (dismissed):
LGTM, just left a nit

packages/datadog-instrumentations/src/ai.js
sabrenner: fixed!

BridgeAR (approved):

PR #6321, opened 2025-08-21
fix gc pause metric using millis instead of nanos

rochdev:
### What does this PR do?

Fix GC pause metric using milliseconds instead of nanoseconds.

### Motivation

This made the metric plummet to 0 or close to it as the precision was incorrect.

### DIFF
--- a/packages/dd-trace/src/runtime_metrics/runtime_metrics.js
+++ b/packages/dd-trace/src/runtime_metrics/runtime_metrics.js
@@ -274,9 +274,10 @@ function startGCObserver () {
   gcObserver = new PerformanceObserver(list => {
     for (const entry of list.getEntries()) {
       const type = gcType(entry.detail?.kind || entry.kind)
+      const duration = entry.duration * 1_000_000
 
-      runtimeMetrics.histogram('runtime.node.gc.pause.by.type', entry.duration, `gc_type:${type}`)
-      runtimeMetrics.histogram('runtime.node.gc.pause', entry.duration)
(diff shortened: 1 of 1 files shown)


packages/dd-trace/src/runtime_metrics/runtime_metrics.js
```diff
@@ -274,9 +274,10 @@ function startGCObserver () {
   gcObserver = new PerformanceObserver(list => {
     for (const entry of list.getEntries()) {
       const type = gcType(entry.detail?.kind || entry.kind)
+      const duration = entry.duration * 1_000_000
```
BridgeAR, line 277: The entry is actually returned as milliseconds, not nanoseconds. It is a double with the fractional part representing the higher accuracy.

We do have wrong time metrics in cpu measurements. These are divided with the wrong values.

rochdev: Not sure I understand, if we expect nanoseconds, and we receive milliseconds, wouldn't multiplying them by a million give us the right number? (worth noting I didn't bother flooring because metrics are doubles)

BridgeAR: Ah, we should have better metric names that include the correct type. You're correct!
https://github.com/DataDog/dogweb/blob/prod/integration/node/node_metadata.csv

BridgeAR (approved):
LGTM. I will open a follow-up PR where I fix the event loop utilization and I am trying to come up with tests.

PR #6329, opened 2025-08-21
feat(llmobs): add user llm observability span processor to modify spans

sabrenner:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2557,6 +2557,17 @@ declare namespace tracer {
       annotate (span: tracer.Span | undefined, options: llmobs.AnnotationOptions): void
 
       /**
+       * Register a processor to be called on each LLMObs span.
+       *
+       * This can be used to modify the span before it is sent to LLMObs. For example, you can modify the input/output.
+       * You can also return `null` or `undefined` to omit the span entirely from being sent to LLMObs.
+       *
+       * To deregister the processor, call `llmobs.registerProcessor(null)`
+       * @param processor A function that will be called for each span.
+       */
+      registerProcessor (processor: (span: LLMObservabilitySpan) => LLMObservabilitySpan | undefined | null): void
+
+      /**
        * Submits a custom evaluation metric for a given span ID and trace ID.
        * @param spanContext The span context of the span to submit the evaluation metric for.
        * @param options An object containing the label, metric type, value, and tags of the evaluation metric.
@@ -2567,6 +2578,25 @@ declare namespace tracer {
        * Flushes any remaining spans and evaluation metrics to LLM Observability.
        */
       flush (): void
+    }
+
+    interface LLMObservabilitySpan {
+      /**
+       * The input messages associated with the span.
+       */
+      input: Message[]
+
+      /**
+       * The output messages associated with the span.
+       */
+      output: Message[]
+
+      /**
+       * Get a tag from the span.
+       * @param key The key of the tag to get.
+       * @returns The value of the tag, or `undefined` if the tag does not exist.
+       */
+      getTag (key: string): string | undefined
     }
 
     interface EvaluationOptions {
--- a/packages/dd-trace/src/llmobs/span_processor.js
+++ b/packages/dd-trace/src/llmobs/span_processor.js
@@ -34,9 +34,28 @@ const LLMObsTagger = require('./tagger')
 const tracerVersion = require('../../../../package.json').version
 const logger = require('../log')
 
+const util = require('node:util')
+
+class LLMObservabilitySpan {
+  constructor () {
+    this.input = []
+    this.output = []
+
+    this._tags = {}
+  }
+
+  getTag (key) {
+    return this._tags[key]
+  }
+}
+
 class LLMObsSpanProcessor {
   constructor (config) {
     this._config = config
+  }
+
+  registerProcessor (processor) {
+    this._processor = processor
   }
 
   setWriter (writer) {
@@ -52,6 +71,8 @@ class LLMObsSpanProcessor {
     try {
       const formattedEvent = this.format(span)
       telemetry.incrementLLMObsSpanFinishedCount(span)
+      if (formattedEvent == null) return
+
       this._writer.append(formattedEvent)
     } catch (e) {
       // this should be a rare case
@@ -65,6 +86,9 @@ class LLMObsSpanProcessor {
   }
 
   format (span) {
+    const llmObsSpan = new LLMObservabilitySpan()
+    let inputType, outputType
+
     const spanTags = span.context()._tags
     const mlObsTags = LLMObsTagger.tagMap.get(span)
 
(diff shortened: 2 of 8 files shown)


packages/dd-trace/src/llmobs/span_processor.js
```diff
@@ -191,13 +249,51 @@ class LLMObsSpanProcessor {
...
+          'User span processor must return an instance of an LLMObservabilitySpan or null'
+        )
+      }
+
+      return processedLLMObsSpan
+    } catch (e) {
+      logger.error(`[LLMObs] Error in LLMObs span processor (${util.inspect(processor)}): ${e.message}`)
+      error = true
+    } finally {
+      telemetry.recordLLMObsUserProcessorCalled(error)
```
Kyle-Verhoog, line 294: ty for implementing the telemetry too 🙏

packages/dd-trace/src/llmobs/span_processor.js
```diff
@@ -191,13 +249,51 @@ class LLMObsSpanProcessor {
...
+
+    try {
+      const processedLLMObsSpan = processor(span)
+      if (!processedLLMObsSpan) return null
+
+      if (!(processedLLMObsSpan instanceof LLMObservabilitySpan)) {
+        throw new TypeError(
+          'User span processor must return an instance of an LLMObservabilitySpan or null'
+        )
+      }
```
BridgeAR: This should be outside of the try / catch :)

We know that there is an error, so we can just directly log that and handle it accordingly.
Especially, since we actually only care about the message (is that really what we want? Should we not generally log the whole stack?)

packages/dd-trace/src/llmobs/span_processor.js
```diff
@@ -144,7 +202,7 @@ class LLMObsSpanProcessor {
   // However, we want to protect against circular references or BigInts (unserializable)
   // This function can be reused for other fields if needed
   // Messages, Documents, and Metrics are safeguarded in `llmobs/tagger.js`
-  _addObject (obj, carrier) {
+  #addObject (obj, carrier) {
```
BridgeAR, line 205: ❤️

packages/dd-trace/src/llmobs/span_processor.js
```diff
@@ -34,25 +34,55 @@ const LLMObsTagger = require('./tagger')
...
+  getTag (key) {
+    return this._tags[key]
+  }
+}
+
 class LLMObsSpanProcessor {
+  /** @type {import('../config')} */
+  #config
+
+  /** @type {((span: LLMObservabilitySpan) => LLMObservabilitySpan | null)} */
```
BridgeAR: ❤️

packages/dd-trace/src/llmobs/telemetry.js
```diff
@@ -154,6 +154,11 @@ function recordSubmitEvaluation (options, err, value = 1) {
   llmobsMetrics.count('evals_submitted', tags).inc(value)
 }
 
+function recordLLMObsUserProcessorCalled (error, value = 1) {
+  const tags = { error: Number(!!error) }
```
BridgeAR:
```suggestion
  const tags = { error: error ? 1 : 0 }
```

While we could just directly set it to 1 or 0 when we define this.

sabrenner: oh yeah true - i'll make this fix throughout the file here. we could do it when we pass it in but from a devex perspective i feel its easier to just pass an error in when calling these telemetry methods and normalize here with `error ? 1 : 0`

packages/dd-trace/src/llmobs/span_processor.js
sabrenner: good point. and actually, that would also get caught by this catch [here](https://github.com/DataDog/dd-trace-js/blob/05ee5bba8951d5d1d94d75f093533d60a4d647ea/packages/dd-trace/src/llmobs/span_processor.js#L56-L63)... so i'll `logger.warn` here, set `error = true`, and return `null` (effectively having the same effect without needing to do the expensive error stuff)

docs/test.ts
```diff
@@ -603,6 +603,27 @@ llmobs.enable({
 // manually disable
 llmobs.disable()
 
+// register a processor
+llmobs.registerProcessor((llmobsSpan) => {
```
BridgeAR, line 607: This seems to override a potential former processor without letting the user know that this is the case. If two teams work on the same code and both want such functionality, the might override a processor without knowing.

I would either only allow a single instance and throw for all follow-ups or (my favorite choice) I would just accept multiple and run them in insertion order.

To deregister, I would add a `deregisterProcessor()` method.

docs/test.ts
```diff
@@ -603,6 +603,27 @@ llmobs.enable({
...
+      return {
+        ...input,
+      }
+    })
+  }
+
+  return llmobsSpan
+})
+
+llmobs.registerProcessor(null)
```
BridgeAR: With the comment above, I think an explicit deregister is nicer than passing through `null`. I would expect `null` to actually be a no-op or an error.

packages/dd-trace/src/llmobs/span_processor.js
```diff
@@ -34,25 +34,59 @@ const LLMObsTagger = require('./tagger')
...
+  /** @type {import('./writers/spans')} */
+  #writer
+
   constructor (config) {
-    this._config = config
+    this.#config = config
+  }
+
+  registerProcessor (processor) {
+    if (processor !== null && this.#processor) {
```
BridgeAR:
```suggestion
    if (typeof processor !== 'function') {
      throw new Error('[LLMObs] Processor has to be of type function')
    }
    if (this.#processor) {
```

packages/dd-trace/src/llmobs/span_processor.js
```diff
@@ -118,12 +155,37 @@ class LLMObsSpanProcessor {
 
     const name = mlObsTags[NAME] || span._name
 
+    const tags = this.#getTags(span, mlApp, sessionId, error)
+    llmObsSpan._tags = tags
+
+    const processedSpan = this.#runProcessor(llmObsSpan)
+    if (processedSpan == null) return null
```
BridgeAR: Should we be more rigorous validating the return value?
Right now, we would ignore every return value that is malformed (anything that has no `input` and `output` property).

packages/dd-trace/src/llmobs/span_processor.js
```diff
@@ -191,13 +253,51 @@ class LLMObsSpanProcessor {
...
+  #objectTagsToStringArrayTags (tags) {
     return Object.entries(tags).map(([key, value]) => `${key}:${value ?? ''}`)
   }
+
+  /**
+   * Runs the user span processor, emitting telemetry and adding some guardrails against invalid return types
+   * @param {LLMObservabilitySpan} span
+   * @returns {LLMObservabilitySpan | null}
+   */
+  #runProcessor (span) {
```
BridgeAR, line 273: This will work and I am fine with the implementation.

If I would have implemented it, I would have probably implemented a tad different (as Idea in a future implementation):

I would define the default as a method that just returns the span. If a processor is registered, I would wrap that as callback from a runProcessor method and redefine the internal processor method. That way the code is very isolated and most performant.

packages/dd-trace/src/llmobs/span_processor.js
sabrenner: actually, this is the internal setter for the llmobs span processor - the actual `llmobs.registerProcessor` and `llmobs.deregisterProcessor` both call it under the hood. i can change this name to `setProcessor`, with the idea that this internal API throws if trying to re-define a processor with a non-null function.

i can also create a separate `deregisterProcessor` function in this file as well but i can just do `setProcessor` since maybe that's a bit clearer that it's not the actual API but an internal one

packages/dd-trace/src/llmobs/span_processor.js
sabrenner: yeah, the idea in general is to ignore any span that doesn't have the expected fields, asserted with the 
```javascript
!(processedLLMObsSpan instanceof LLMObservabilitySpan)
```
check in `#runProcessor`. let me know if i'm misunderstanding though 😄

packages/dd-trace/src/llmobs/span_processor.js
sabrenner: i've pushed [375a0bd](https://github.com/DataDog/dd-trace-js/pull/6329/commits/375a0bdd7f56a50353a955fdb2fc5ed70e074b7e) to rename it for clarity

index.d.ts
```diff
@@ -2572,6 +2572,8 @@ declare namespace tracer {
        * This can be used to modify the span before it is sent to LLMObs. For example, you can modify the input/output.
        * You can also return `null` to omit the span entirely from being sent to LLM Observability.
        *
+       * If the return value from the processor is not an instance of `LLMObservabilitySpan` or null, the span will be dropped.
```
BridgeAR: It will in fact also drop the span if `null` is returned. It would just not log an error :)

sabrenner: oooops yes clarified the language here!

BridgeAR (approved):

PR #6330, opened 2025-08-22
[test-optimization] [SDTEST-2548] Fix playwright instrumentation for versions `>=1.55.0`

Mariovido:
### What does this PR do?

With this fix we enable Playwright instrumentation for version `>=1.55.0`

### Motivation

We want to support latests versions of the tracers supported in our product.

### Plugin Checklist

- [x] Unit tests.
- [x] Integration tests.

### DIFF
--- a/packages/datadog-instrumentations/src/playwright.js
+++ b/packages/datadog-instrumentations/src/playwright.js
@@ -148,8 +148,8 @@ function getPlaywrightConfig (playwrightRunner) {
   }
 }
 
-function getRootDir (playwrightRunner) {
-  const config = getPlaywrightConfig(playwrightRunner)
+function getRootDir (playwrightRunner, configArg) {
+  const config = configArg ? configArg.config : getPlaywrightConfig(playwrightRunner)
   if (config.rootDir) {
     return config.rootDir
   }
@@ -162,8 +162,8 @@ function getRootDir (playwrightRunner) {
   return process.cwd()
 }
 
-function getProjectsFromRunner (runner) {
-  const config = getPlaywrightConfig(runner)
+function getProjectsFromRunner (runner, configArg) {
+  const config = configArg || getPlaywrightConfig(runner)
   return config.projects?.map((project) => {
     if (project.project) {
       return project.project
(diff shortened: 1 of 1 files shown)


packages/datadog-instrumentations/src/playwright.js
```diff
@@ -162,8 +162,8 @@ function getRootDir (playwrightRunner) {
   return process.cwd()
 }
 
-function getProjectsFromRunner (runner) {
-  const config = getPlaywrightConfig(runner)
+function getProjectsFromRunner (runner, configArg) {
+  const config = configArg || getPlaywrightConfig(runner)
```
nikita-tkachenko-datadog, line 166: Just to double-check: `configArg ? configArg.config : getPlaywrightConfig(playwrightRunner)` and `configArg || getPlaywrightConfig(runner)` seem inconsistent - one uses `configArg` directly as a replacement for `getPlaywrightConfig`, the other uses `configArg.config`

packages/datadog-instrumentations/src/playwright.js
```diff
@@ -651,6 +652,23 @@ function runnerHook (runnerExport, playwrightVersion) {
...
+  shimmer.wrap(
+    runnerExport.Runner.prototype,
+    'runAllTests',
+    runAllTests => runAllTestsWrapper(runAllTests, playwrightVersion)
+  )
+}
+
+function runnerHookNew (runnerExport, playwrightVersion) {
+  runnerExport = shimmer.wrap(runnerExport, 'runAllTestsWithConfig', function (originalGetter) {
+    const originalFunction = originalGetter.call(this)
```
nikita-tkachenko-datadog, line 668: `runAllTestsWrapper` calls the function it gets as `await runAllTests.apply(this, arguments)` - wouldn't `apply(this, ...)` override whatever binding you do with `call`?
If yes, then it looks like your new hook is essentially identical to the existing one

packages/datadog-instrumentations/src/playwright.js
Mariovido: Because they changed the structure of the config. This `configArg` is the parameter of the function that contains multiple things such as the `projects` or the "real" `config` where the `rootDir` is stored. Previously all of this was stored directly on `playwrightRunner`
- configArg:
  - `projects`
  - `config`
    - `rootDir`

packages/datadog-instrumentations/src/playwright.js
Mariovido: The difference from the previous version is that the function is no longer defined on the Runner's `prototype`, which means we can't instrument it directly anymore. Instead, it has become a **getter**, so we need to override that. To handle this, I first retrieve the original function via the `originalGetter`. Then, I wrap the previous hook function (`runAllTestsWrapper`) in another function that will act as the `getter`

nikita-tkachenko-datadog: Ahh, I confused `call` with `bind`. My JS is getting rusty

PR #6333, opened 2025-08-22
Eric.firth/fix md5 incompatibility with openssl changes issue

ericfirth:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/dd-trace/test/plugins/agent.js
```diff
@@ -69,6 +74,10 @@ function dsmStatsExist (agent, expectedHash, expectedEdgeTags) {
       }
     }
   }
+  if (!hashFound) {
+    const foundHashesStr = Array.from(foundHashes).join(', ')
+    throw new Error(`Hash not found. Expected: ${expectedHash}, Found hashes: [${foundHashesStr}]`)
+  }
   return hashFound
```
BridgeAR: Thank you for improving the error here! That is significantly better! ❤️

We do not need the hashFound variable at all though. It is always going to be false here. I would just remove it, return directly true/false, and serialize the set here with Node.js `util.inspect()`.

```suggestion
  throw new Error(`Hash not found. Expected: ${expectedHash}, Found hashes: ${util.inspect(foundHashes)}`)
```

packages/dd-trace/test/plugins/agent.js
```diff
@@ -43,22 +43,27 @@ function ciVisRequestHandler (request, response) {
...
+                throw new Error(msg)
               }
 
               const expected = expectedEdgeTags.slice().sort()
               const actual = stats.EdgeTags.slice().sort()
 
               for (let i = 0; i < expected.length; i++) {
                 if (expected[i] !== actual[i]) {
-                  return false
+                  throw new Error(`EdgeTags mismatch at index ${i}. Expected "${expected[i]}", got "${actual[i]}"`)
```
BridgeAR: This is all a great improvement over the current implementation! I think, we could just push it even a bit further by using `assert.deepStrictEqual()` from Node.js's assert module. That way also the length check would provide a good output in case of an error.

```js
const expected = expectedEdgeTags.slice().sort()
const actual = stats.EdgeTags.slice().sort()
assert.deepStrictEqual(actual, expected, 'EdgeTags mismatch')
```

packages/dd-trace/src/datastreams/pathway.js
```diff
@@ -17,8 +17,8 @@ const CONTEXT_PROPAGATION_KEY_BASE64 = 'dd-pathway-ctx-base64'
 const logKeys = [CONTEXT_PROPAGATION_KEY, CONTEXT_PROPAGATION_KEY_BASE64]
 
 function shaHash (checkpointString) {
-  const hash = crypto.createHash('md5').update(checkpointString).digest('hex').slice(0, 16)
-  return Buffer.from(hash, 'hex')
+  const hash = crypto.createHash('sha256').update(checkpointString).digest('hex').slice(0, 8)
```
BridgeAR: Is it fine to just switch to sha256? I do not know what requirements we have for the hash and if we may use any hashing algorithm or not.

If changing the algorithm is fine (md5 is of course not secure, while that should not matter for the use case here, as far as I can tell), I would suggest we look into using either [XXH3](https://github.com/Cyan4973/xxHash) or [BLAKE3](https://github.com/BLAKE3-team/BLAKE3). These are much faster.

In addition: if we only care about the first 8 characters, we likely loose some of the hash collision guarantees. Would it potentially make sense to increase the characters we want to look at?

ericfirth: the guarantees we care about are pretty slim. we take the checkpoint string, which is just a bunch of tags concatted together and we want to hash it so that the same string gives the same uint64. collisions should be avoided but isn't the worst thing in the world. I did make a change since i accidentally had made collisions more likely.

At this point I wonder if its worth adding a new dependency in XXH3. Right now we want to fix dd-trace-js asap for users who are experiencing errors.

However, there is an initiative to make DSM on by default and drop collecting the stats at the edge (meaning at the agent level, whereas the tracer is just always on). If we pursue that for node (we are currently testing this on dotnet), I believe the dsm code will have a lot of performance testing to ensure we don't slow down produce/consume calls and that is when it would make sense for us to add XXH3.

packages/dd-trace/test/plugins/agent.js
ericfirth: done

packages/dd-trace/test/plugins/agent.js
ericfirth: done

PR #6335, opened 2025-08-23
[test-optimization] [SDTEST-2548] Fix Cypress tests for versions `>=15.0.0`

Mariovido:
### What does this PR do?

This PR modify tests to be able to be able to support to the latest version of Cypress (`15.0.0`). The main issue is that Cypress stopped support of Node versions (`18` and `23`) on their latest major release.

### Motivation

We want to be able to support latest versions of the frameworks that we support

### Plugin Checklist

- [x] Unit tests.
- [x] Integration tests.

### Additional Notes

[Cypress Release Notes (15.0.0)](https://docs.cypress.io/app/references/changelog#15-0-0)

### DIFF
--- a/integration-tests/cypress/cypress.spec.js
+++ b/integration-tests/cypress/cypress.spec.js
@@ -86,7 +86,8 @@ function shouldTestsRun (type) {
       return version === '6.7.0' && type === 'commonJS'
     }
     if (NODE_MAJOR > 16) {
-      return version === 'latest'
+      // Cypress 15.0.0 has removed support for Node 18
+      return NODE_MAJOR > 18 ? version === 'latest' : version === '14.5.4'
     }
   }
   if (DD_MAJOR === 6) {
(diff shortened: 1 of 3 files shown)


integration-tests/cypress/cypress.spec.js
```diff
@@ -86,15 +86,20 @@ function shouldTestsRun (type) {
       return version === '6.7.0' && type === 'commonJS'
     }
     if (NODE_MAJOR > 16) {
-      return version === 'latest'
+      // Cypress 15.0.0 has removed support for Node 18
+      return NODE_MAJOR > 18 ? version === 'latest' : version === '14.5.4'
     }
```
daniel-mohedano, line 91: tiny nit: i think it'd be easier to follow with just one level of if elses, specially if it keeps growing in the future

daniel-mohedano (approved):

PR #6343, opened 2025-08-25
[test-optimization] [SDTEST-2548] Fix playwright in `v5`

Mariovido:
### What does this PR do?

With this fix we correctly setup the instrumentation for versions lower to `1.38.0` in Playwright.

### Motivation

We want to keep supporting old versions in `v5`

### Plugin Checklist

- [x] Unit tests.
- [x] Integration tests.


BridgeAR (approved):
RSLGTM

PR #6344, opened 2025-08-26
fix: improve runtime metrics

BridgeAR:
The current implementation for runtime.node.event_loop.utilization always calculated a diff from the former entry and replaced the former entry with that diff. That in itself worked for the first two calls, but not afterwards, since we now receive a diff of the diff and that is incorrect.
It is now replaced by directly calculating the utilization instead.

The overall implementation is also a tad faster by removing some calls to native side as well as using process.hrtime.bigint() instead of process.hrtime().

Code for unsupported Node.js versions was also removed as well as adding a comment about the overall metric types and where to find those.

### DIFF
--- a/integration-tests/ci-visibility/test-early-flake-detection/__snapshots__/jest-snapshot.js.snap
+++ b/integration-tests/ci-visibility/test-early-flake-detection/__snapshots__/jest-snapshot.js.snap
@@ -1,3 +1,3 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
 
 exports[`test can do snapshot 1`] = `3`;


packages/dd-trace/src/runtime_metrics/runtime_metrics.js
```diff
@@ -4,86 +4,121 @@
...
 
-    if (config.runtimeMetrics.eventLoop !== false) {
+    if (trackEventLoop) {
       watchers.push('loop')
+    } else {
+      // Prevent the native gc metrics from being tracked. Not passing any
+      // options means all metrics are tracked.
+      // TODO: This is a workaround. We should find a better solution.
+      watchers.push('no-gc')
     }
```
szegedi: This could be written more compactly now as
```
const watchers = trackEventLoop ? ['loop'] : ['no-gc']
```

packages/dd-trace/src/runtime_metrics/runtime_metrics.js
```diff
@@ -4,86 +4,121 @@
...
+const { NODE_MAJOR } = require('../../../../version')
+// TODO: This environment variable may not be changed, since the agent expects a flush every ten seconds.
+// It is only a variable for testing. Think about alternatives.
 const DD_RUNTIME_METRICS_FLUSH_INTERVAL = getEnvironmentVariable('DD_RUNTIME_METRICS_FLUSH_INTERVAL') ?? '10000'
 const INTERVAL = Number.parseInt(DD_RUNTIME_METRICS_FLUSH_INTERVAL, 10)
 
-// Node >=16 has PerformanceObserver with `gc` type, but <16.7 had a critical bug.
-// See: https://github.com/nodejs/node/issues/39548
-const hasGCObserver = NODE_MAJOR >= 18 || (NODE_MAJOR === 16 && NODE_MINOR >= 7)
+const eventLoopDelayResolution = 1
```
szegedi: This will schedule a timer once every millisecond – won't that be too frequent? It'll force all I/O polls to have a timeout of 1ms. (I know this only comes into play when native metrics don't load so it isn't too likely.)

packages/dd-trace/src/runtime_metrics/runtime_metrics.js
```diff
@@ -4,86 +4,121 @@
...
-      log.error('Error starting native metrics', e)
+    } catch (error) {
+      log.error('Error starting native metrics', error)
       nativeMetrics = null
     }
 
     client = new MetricsAggregationClient(new DogStatsDClient(clientConfig))
 
-    time = process.hrtime()
+    lastTime = process.hrtime.bigint()
```
szegedi: I personally prefer using `perf_hooks.performance.now()` – it's specified in millis, but uses fractions (so you get nanos as well), it's an ordinary number and not a bigint, and its zero is the start of the `node` process.

packages/dd-trace/src/runtime_metrics/runtime_metrics.js
```diff
@@ -4,86 +4,121 @@
...
+          // We use these only as fallback values.
+          captureEventLoopDelay()
+        }
         client.flush()
       }, INTERVAL)
     }
 
     interval.unref()
+    // The starttime plus half a second for rounding.
+    startTime = lastTime - BigInt(Math.round(process.uptime() * 1e9)) + 500_000_000n
```
szegedi: If you used `performance.now()` for time values, this could be simply:
```
startTime = 500
```

packages/dd-trace/src/runtime_metrics/runtime_metrics.js
```diff
@@ -160,12 +185,41 @@ function captureMemoryUsage () {
...
+  // node_metadata.csv to be updated for the website.
+  //
+  // client.gauge('runtime.node.mem.arrayBuffers', stats.arrayBuffers)
+}
 
-  stats.external && client.gauge('runtime.node.mem.external', stats.external)
+function captureUptime () {
+  // WARNING: lastTime must be updated in the same interval before this function is called!
+  // This is a faster `process.uptime()`.
+  client.gauge('runtime.node.process.uptime', Number((lastTime - startTime) / 1_000_000_000n))
```
szegedi: Again, with `perf_hooks.performance.now()`, `lastTime` is effectively uptime, so you could do this:
```
client.gauge('runtime.node.process.uptime', lastTime / 1_000)
```
or
```
client.gauge('runtime.node.process.uptime', (lastTime - 500) / 1_000)
```
if you wanted that 500ms offset.

packages/dd-trace/src/runtime_metrics/runtime_metrics.js
BridgeAR: I originally set it to 25 ms. I think that would be fine, while we currently sample way more with the native implementation and 1ms was the nearest implementation to the native one.

packages/dd-trace/src/runtime_metrics/runtime_metrics.js
BridgeAR: I like the suggestion. I will look into how it performs compared to hrtime.

BridgeAR: I changed it to use performance.now(), while it required to mock the API a lot which makes the test weaker. I am not too happy about that but oh well.

packages/dd-trace/src/runtime_metrics/runtime_metrics.js
BridgeAR: I was able to remove it by using performance.now()

packages/dd-trace/src/runtime_metrics/runtime_metrics.js
BridgeAR: I added the offset, while I am not 100% certain which is now correct. It is difficult to tell due to the mocking. I might check again with the former implementation if all values align.

packages/dd-trace/src/runtime_metrics/runtime_metrics.js
BridgeAR: I increased it to 5 ms. Setting it higher would not allow the test to pass anymore without additional changes.

szegedi (approved):

PR #6347, opened 2025-08-27
the loader is broken for openai agents, adds the shim for it as well

gmmeyer:
### What does this PR do?
This adds the OpenAI shim for agents to the ignore list

### Motivation
The loader registration is broken for openai agents and gives this error when you try to use it (I changed this to purely the relative path)

```
@openai/agents-core/dist/config.mjs:8
    return _loadEnv();
    ^

ReferenceError: Cannot access '_loadEnv' before initialization
    at loadEnv (@openai/agents-core/dist/config.mjs:8:5)
    at isEnabled (@openai/agents-core/dist/config.mjs:17:17)
    at get dontLogModelData (@openai/agents-core/dist/config.mjs:41:16)
    at @openai/agents-core/dist/logger.mjs:7:34
    at ModuleJob.run (node:internal/modules/esm/module_job:371:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:669:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)
```

It seems like we are overriding this improperly. You can see this error is coming from here:

https://github.com/openai/openai-agents-js/blob/f075ca780d3ecdb7e1fb737f863ca378f76d3b29/packages/agents-core/src/config.ts#L3-L15

Meaning that somehwere in this file:

https://github.com/openai/openai-agents-js/blob/f075ca780d3ecdb7e1fb737f863ca378f76d3b29/packages/agents-core/src/shims/shims-node.ts#L15

we are removing the loadenv improperly. It also seems like we are ignoring the shims for the other openai package, so I figured this simply needed to be added here as well.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


BridgeAR (commented):
Thank you for the PR! Without looking deeply into it, this seems correct.

Would you mind adding some tests for it as well? That would be very helpful!

Kyle-Verhoog (dismissed):

gmmeyer:
I found a second exclude list and added it to that as well -- we weren't using the initialize.mjs

@BridgeAR I did a little looking, where are the tests for these, I can't find them? happy to add a test

BridgeAR:
> I did a little looking, where are the tests for these, I can't find them? happy to add a test

We'd need a new integration test that uses some of the parts that should be ignored. You can have a look at the tests that were introduced in the PR that introduced the ignore list: https://github.com/DataDog/dd-trace-js/pull/5267/files#diff-f6d5f969d408b29ba2d5f3c988e30ff11cc978c464289d6c1f0866633980b38a :)

So it's pretty much about adding a simple example application of openai that would currently fail to our tests. If it's instrumented without error, and we receive spans, we'd know it works fine.

gmmeyer:
yea I can do this, basically after we installed the openai-agents package the app stopped booting so such a test should be simple enough

gmmeyer:
@BridgeAR @Kyle-Verhoog I added a test, this lib as far as I can tell isn't actually instrumented (except incidentally) so the only real relevant test was whether the app can start with it installed. This was the issue it was preventing loading by monkeypatching. I just added it in the sandbox for the openai test and it fails to load up without my changes, but with them does load properly

PR #6351, opened 2025-08-27
fix ip_extractor algo for Forwarded header

simon-id:
### What does this PR do?
Change the algorithm for ip extraction as the previous implementation was wrong.

- remove usage of the by= directive in the forwarded header
- add support for quotes, brackets, and port stripping in all headers
- add support for special "forwarded" parsing also when custom header is set to "forwarded"


packages/dd-trace/src/plugins/util/ip_extractor.js
```diff
@@ -45,46 +44,66 @@ function extractIp (config, req) {
...
   const result = {}
   if (!str) return result
 
   const splitted = str.split(',')
 
-  for (const part of splitted) {
-    const chunk = part.trim()
+  for (let chunk of splitted) {
+    if (isForwardedHeader) {
+      // find for directive
```
CarlesDD: Nit: avoid confusion with the phrasal verb
```suggestion
      // find "for" directive
```

packages/dd-trace/src/plugins/util/ip_extractor.js
```diff
@@ -97,37 +116,21 @@ function findFirstIp (str) {
...
+    return input.slice(0, colonIndex).trim()
+  }
 
-      // it's private, only save the first one found
-      if (!result.private) result.private = ip
-    }
+  const closeBracketIndex = input.indexOf(']')
+  if (closeBracketIndex !== -1 && input.startsWith('[')) {
+    // treat as ipv6 with brackets
+    return input.slice(1, closeBracketIndex).trim()
```
CarlesDD:
```suggestion
  const closingBracketIndex = input.indexOf(']')
  if (closingBracketIndex !== -1 && input.startsWith('[')) {
    // treat as ipv6 with brackets
    return input.slice(1, closingBracketIndex).trim()
```

uurien (approved):

PR #6354, opened 2025-08-27
fix(tracing): fixes graphql resolver error not returning a then-able result

wconti27:
### What does this PR do?
Fixes error where graphql resolvers were not returning a then-able. This came up as a result of an escalation. Adds a relevant test case as well.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/datadog-plugin-graphql/test/index.spec.js
```diff
@@ -833,6 +853,18 @@ describe('Plugin', () => {
...
+          const source = '{ human { oneTime } }'
+
+          graphql.graphql({ schema, source })
+            .then(result => {
+              expect(result).to.not.have.property('errors')
+              expect(result.data.human.oneTime).to.equal('one-time result')
+              done()
+            })
+            .catch(done)
+        })
```
BridgeAR:
```suggestion
        it('should not re-execute thenables from resolvers', async () => {
          const source = '{ human { oneTime } }'

          const result = await graphql.graphql({ schema, source })
          expect(result).to.not.have.property('errors')
          expect(result.data.human.oneTime).to.equal('one-time result')
        })
```

I guess that should also be fine to use?

packages/datadog-instrumentations/src/graphql.js
```diff
@@ -235,14 +235,20 @@ function callInAsyncScope (fn, thisArg, args, abortController, cb) {
...
+          },
+          err => {
+            cb(err)
+            reject(err)
+          }
+        )
+      })
     }
+    cb(null, result)
     return result
```
BridgeAR, line 250: Should we actually return the result in this case? I would expect we actually should not do that as it could also be detected? I would handle it as it would have worked without our instrumentation (it could be either or).

BridgeAR (commented):
The code seems fine overall, while I would likely remove the return.

I am just very surprised about a thenable to be expected to only ever be called once. That is... unexpected. It is definitely not Promise/A+ spec compliant. Do we know what is used in this case?

packages/datadog-instrumentations/src/graphql.js
```diff
@@ -235,14 +235,20 @@ function callInAsyncScope (fn, thisArg, args, abortController, cb) {
...
+          res => {
+            cb(null, res)
+            resolve(res)
+          },
+          err => {
+            cb(err)
+            reject(err)
+          }
+        )
+      })
```
bengl: Rather than a whole new promise object, can't you just return or throw in the `then` handler?

```suggestion
      return result.then(
        res => {
          cb(null, res)
          return res
        },
        err => {
          cb(err)
          throw err
        }
      )
```

That way you're not creating a whole extra promise.

bengl (changes requested):

packages/datadog-plugin-graphql/test/index.spec.js
wconti27: The case for this was a Mongoose V6 query object that is only expected to be called once. Here's the docs explaining it: https://mongoosejs.com/docs/migrating_to_6.html#duplicate-query-execution.

wconti27: I've updated the tests to use await syntax.

packages/datadog-instrumentations/src/graphql.js
wconti27: Fixed!

packages/datadog-instrumentations/src/graphql.js
wconti27: Why would we not return the result in this case? Is it because an abort may be detected asynchronously?

packages/datadog-instrumentations/src/graphql.js
BridgeAR: @bengl looking at this again: in another library, they used the original promise for state tracking.

This is somewhat conflicting with the mongoose implementation, since it means it is impossible to separate the state tracking and prevent calling the promise only once. I guess that is fine overall, it is just something to keep in mind and I wonder if another library could have a similar behavior.

I also wonder if we should discuss with the mongoose maintainers about being spec compliant.

bengl: Hmm, in that case, should we return the original promise? Like:

```js
result.then(/* .... */)
return result
```

wconti27: @bengl  This is the current day behavior and is what results in the mongoose error.

packages/datadog-instrumentations/src/graphql.js
```diff
@@ -235,14 +235,28 @@ function callInAsyncScope (fn, thisArg, args, abortController, cb) {
...
-        res => cb(null, res),
-        err => cb(err)
+      let promise = result
+
+      // Mongoose queries are thenable, but not true promises. They can't be
+      // resolved multiple times. `.exec()` returns a true promise. We can
+      // identify a mongoose query by the presence of an `exec` function and
+      // its constructor name.
+      if (typeof result.exec === 'function' && result.constructor.name === 'Query') {
+        promise = result.exec()
```
bengl: Does exec() have side effects? Can it be called multiple times?

packages/datadog-instrumentations/src/graphql.js
```diff
@@ -235,14 +235,28 @@ function callInAsyncScope (fn, thisArg, args, abortController, cb) {
...
+      }
+
+      return promise.then(
```
bengl: We don't want to change the return value here, do we?

bengl (changes requested):

packages/datadog-instrumentations/src/graphql.js
wconti27: Fixed to wrap the query object instead of calling `.exec()`

packages/datadog-instrumentations/src/graphql.js
wconti27: nope, reverted

packages/datadog-instrumentations/src/graphql.js
```diff
@@ -235,14 +238,25 @@ function callInAsyncScope (fn, thisArg, args, abortController, cb) {
...
     const result = fn.apply(thisArg, args)
     if (result && typeof result.then === 'function') {
-      // bind callback to this scope
-      result.then(
-        res => cb(null, res),
-        err => cb(err)
+      // Mongoose queries are thenable, but not true promises. We can
+      // identify a mongoose query by the presence of an `exec` function and
+      // its constructor name, then wrap it.
+      if (typeof result.exec === 'function' && result.constructor.name === 'Query') {
```
bengl: Is this enough to know that the object has come from mongoose? I doubt it is. Could we somehow tag it (e.g. by adding it to a WeakSet from within the mongoose integration) so that we know for _sure_ that's what's happening?

bengl (changes requested):

packages/datadog-instrumentations/src/graphql.js
wconti27: Removed any changes related to mongoose, so this isn't a worry anymore

PR #6358, opened 2025-08-28
Topher c/apmsvls 52 fixing payload tagging

TophrC-dd:
### What does this PR do?
This fixes a couple of issues with payload tagging logic.
Currently if using an invalid json path can break tracing. 
Also depending on variable it can turn on the feature despite not intending to. My change adds some error handling code to inform the user when an incorrect path is set when trying to redact payloads. Also I'm adding enhancements to the process that make it so that if the user sets either 
`DD_TRACE_CLOUD_REQUEST_PAYLOAD_TAGGING` or `DD_TRACE_CLOUD_RESPONSE_PAYLOAD_TAGGING`
to `true` it will be the same as effectively setting as `all` the user can also disable the feature using `false`. 

### Motivation
I want to make the use of this feature to be more intuitive along with catching when the user makes a mistake so that they are not questioning why their payloads are redacting or in this case their whole trace disappears. 

### Plugin Checklist
payload-tagging

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/dd-trace/src/config.js
```diff
@@ -189,9 +189,9 @@ function validateNamingVersion (versionString) {
  * @param {string | string[]} input
  */
 function splitJSONPathRules (input) {
-  if (!input) return
+  if (!input || input.toLowerCase() === 'false' || input === '$') return
   if (Array.isArray(input)) return input
-  if (input === 'all') return []
+  if (input === 'all' || input.toLowerCase() === 'true') return []
```
BridgeAR: Using false or true is undocumented behavior. I don't think we should do that. It is clearly documented how it is possible to enable it or disable it for now.

packages/dd-trace/src/config.js
```diff
@@ -189,9 +189,9 @@ function validateNamingVersion (versionString) {
...
   return input.split(',')
```
BridgeAR, line 193: Instead of validating the entries later during runtime, we can already validate the input right away and filter all invalid ones. That applies to both, the array and the string notation.

That way the runtime has less overhead and we would only log a warning once instead of on each usage.

BridgeAR (changes requested):
Thanks for the PR! The overall idea is sound, while we should in my opinion implement a validation in the config and therefore prevent later failures.

That is better for the user in multiple ways (only one log and less overhead).

We'd also need tests for this to land :)

packages/dd-trace/src/config.js
TophrC-dd: Hey @BridgeAR 
That makes a lot of sense. Although another reason why i did made that decision is that invalid JsonPaths also breaks the tracer.  

Would that make it appropriate to log a warning and _then_ remove the jsonpath from list if its invalid?

TophrC-dd: Hey @BridgeAR -- Just to clarify the underlying issue is not with JsonPath it's us potentially trying to access a property that is null from the function parameters. Which comes from the input given at runtime this makes configuration validation impossible. I have implemented an if statement check to ensure that the properties exist within the object. 

I am pretty sure this is the best possible way to implement this. Although a try/catch could also have been used. 

~Topher

BridgeAR: Thanks. The try / catch would have worked, it would have also caused a lot of overhead though. Calculating a stack trace is very expensive

packages/dd-trace/src/config.js
```diff
@@ -181,7 +181,7 @@ function validateNamingVersion (versionString) {
  * @param {string | string[]} input
  */
 function splitJSONPathRules (input) {
-  if (!input) return
+  if (!input || input === '$') return
```
BridgeAR, line 190: Note: this could also be considered a breaking change. I believe with the current state, we can just also handle it as a bugfix.

integration-tests/payload-tagging/index.js
```diff
@@ -0,0 +1,57 @@
...
+  ]
+})
+
+// Create an SQS service object
+const sqs = new AWS.SQS({ apiVersion: '2012-11-05' });
+
+// Define message parameters
+const params = {
+    MessageBody: 'This is a test message from a standalone Node.js application.',
+    QueueUrl: 'https://sqs.us-east-1.amazonaws.com/123456789012/MyTestQueue' // Replace with your SQS queue URL
```
BridgeAR: This should probably be replaced as outlined in the comment.
We could use a similar name as in other tests such as http://127.0.0.1:4566/00000000000000000000/TEST_NAME

joeyzhao2018 (approved):

packages/dd-trace/src/payload-tagging/index.js
```diff
@@ -1,5 +1,6 @@
 'use strict'
 
+
```
BridgeAR:
```suggestion
```

PR #6360, opened 2025-08-29
[test-optimization] [SDTEST-2569] Fix Jest tests with versions `30.1.0+`

Mariovido:
### What does this PR do?

The new version of Jest changed the way a snapshot test should be created. With this PR we update that method to adapt to the new one.

### Motivation

We want to unblock CI.

### Plugin Checklist

- [x] Unit tests.
- [x] Integration tests.


BridgeAR (approved):
RSLGTM

PR #6361, opened 2025-08-29
chore(tracing): fix prisma client tracing for esm and add esm tests

wconti27:
### What does this PR do?
Fixes tracing of prisma client when used with ESM. Previously, we were attempting to replace the `prisma.PrismaClient` with a `TracedPrismaClient`, where we collected data about the configured databases. This works for CommonJS but fails for ESM. The solution is to wrap the `request` method of the objects prototype, and collect the data from there. This PR also adds some ESM tests.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/datadog-instrumentations/src/prisma.js
```diff
@@ -66,23 +66,21 @@ class TracingHelper {
...
-    * This allows us to grab additional information from the PrismaClient such as DB connection strings
-  */
-  class PrismaClient extends prisma.PrismaClient {
-    constructor (...args) {
-      super(...args)
-
-      const datasources = this._engine?.config.inlineDatasources?.db.url?.value
+  // we need to patch the prototype to get db config since this works for ESM and CJS alike.
+  const originalRequest = prisma.PrismaClient.prototype._request
+  prisma.PrismaClient.prototype._request = function () {
```
BridgeAR, line 71: This will run on each request while formerly, we only needed to check this during the construction time of the instance.
I am not sure why replacing the constructor works for CJS but not for ESM. Do you know why that's the case?

If we only ever need the first instance, we could just replace the method again afterwards to the original one, while I wonder if there's another thing that we could hook into that also contains the necessary information.

Another issue I see is if a customer potentially has more than a single Prisma instance. I am not sure if that's even possible. I guess Prisma does not allow that?

bengl: Yeah in ESM, `import-in-the-middle` should handle the ability to replace exported constructors, so this is odd.

I think the multiple instance problem pre-dates this change.

wconti27: So prisma does allow multiple instances, which means we probably should change the code to allow 1 db config block per instantiated client. We could also, once we save the db config, just reset the request method on the instance (not the prototype) to the original request. 

Not sure why the constructor approach wasn't working, though the code for Prisma is bundled so its not easy to look at the src code.

wconti27: Looking at this again, allowing for more than once db instance is a larger change and pretty out of scope for this fix PR, and would require changes in Prisma upstream, since the instrumentation in Prisma does not send any information that would allow us to differentiate between different databases / instances.

bengl: For now, the comment explains why we're doing what we're doing here, so that's fine.

bengl (approved):

PR #6368, opened 2025-09-02
test: add explicit requires in tests

BridgeAR:
This is a first step towards better types and also to be able to easily navigate to that without the IDE needing to pick up globals.


BridgeAR:
I know this is a big PR, I hope it's so easy to review that splitting it in smaller ones is not required. I have a lot of additional code that I would like to open PRs for type changes when this lands. The explicit requires helped identifying issues locally and is something we wanted to do anyway.

packages/datadog-instrumentations/test/url.spec.js
```diff
@@ -72,7 +75,7 @@ names.forEach(name => {
           const OriginalUrl = Object.getPrototypeOf(url.URL)
           const originalUrl = new OriginalUrl('https://www.datadoghq.com')
 
-          assert.isTrue(originalUrl instanceof url.URL)
+          assert.strictEqual(originalUrl instanceof url.URL, true)
```
simon-id:
```suggestion
          assert(originalUrl instanceof url.URL)
```

packages/dd-trace/test/startup-log.spec.js
```diff
@@ -102,7 +105,7 @@ describe('startup logging', () => {
 describe('profiling_enabled', () => {
   it('should be correctly logged', () => {
     [
-      [undefined, false],
+      ['undefined', false],
```
simon-id, line 108: not sure this is supposed to be a string, it's testing that the env var is undefined user-side, not "undefined" os-side no ?

BridgeAR: Every value that is assigned as env is going to be stringified. This helps prevent a type error and makes it clear how the value will look like afterwards. It's the string undefined.

PR #6372, opened 2025-09-03
fix esm hooking on default exports

wconti27:
### What does this PR do?
ESM hooking was not working properly, and was only being done on modulExports.default, but not the main moduleExports. This PR attempts to fix this, and adds a check to prevent double wrapping.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-shimmer/src/shimmer.js
+++ b/packages/datadog-shimmer/src/shimmer.js
@@ -12,7 +12,7 @@ const skipMethods = new Set([
 const skipMethodSize = skipMethods.size
 
 const nonConfigurableModuleExports = new WeakMap()
-// const wrappedFns = new WeakMap()
+const wrappedFns = new WeakMap()
 
 /**
  * Copies properties from the original function to the wrapped function.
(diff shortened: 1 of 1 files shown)


packages/datadog-shimmer/src/shimmer.js
```diff
@@ -12,6 +12,7 @@ const skipMethods = new Set([
 const skipMethodSize = skipMethods.size
 
 const nonConfigurableModuleExports = new WeakMap()
+const wrappedObjects = new WeakMap()
```
bengl: You can use a WeakSet, so you don't need to set a value.

packages/datadog-instrumentations/src/helpers/hook.js
```diff
@@ -41,6 +34,14 @@ function Hook (modules, hookOptions, onrequire) {
...
       moduleExports.default = safeHook(moduleExports.default, moduleName, moduleBaseDir)
     }
+
+    const parts = [moduleBaseDir, moduleName].filter(Boolean)
+    const filename = path.join(...parts)
+
+    if (this._patched[filename]) return moduleExports
+
+    this._patched[filename] = true
+
```
bengl, line 53: You should leave this chunk in safeHook, and just not do the `moduleExports.default` check here, and instead do it inside safeHook, like:

```js
if (moduleExports && moduleExports.default) {
  moduleExports.default = onrequire(moduleExports.default, moduleName, moduleBaseDir, moduleVersion)
}
return onrequire(moduleExports, moduleName, moduleBaseDir, moduleVersion)
```

BridgeAR:
I'm going to close this, since it is actually not going to resolve the issue in full after us looking at the issue in depth together.

We actually have to check for each instrumentation individually and verify the behavior with tests. There are just too many individual setups that can't be generalized fully.

PR #6374, opened 2025-09-03
Added visibility for single and bulked delete/update commands

perhardhernandez:
### What does this PR do?

This PR adds visibility for database statements from delete and update commands (with support for versions below 4.x.x and above). The changes also expose bulk update and delete operations. Tests have been added to verify that delete and update statements are properly exposed, including bulkWrite operations with update and delete commands.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes

update and delete commands are always exposed as an array which can contain one or more statements, hence the adaptations to support arrays and bulkWrite operations.


packages/datadog-plugin-mongodb-core/test/mongodb.spec.js
```diff
@@ -12,6 +12,9 @@ const ddpv = require('mocha/package.json').version
 const withTopologies = fn => {
   withVersions('mongodb-core', 'mongodb', '>=2', (version, moduleName) => {
     describe('using the default topology', () => {
+      before(() => {
+        global.currentMongoVersion = version
+      })
```
bengl: This is unneeded. You can just refer to `version` wherever you're currently referring to `currentMongoVersion`. In general, avoid globals, and if you _do_ need a variable shared between tests, just use a local scoped variable.

packages/datadog-plugin-mongodb-core/test/mongodb.spec.js
```diff
@@ -109,6 +112,138 @@ describe('Plugin', () => {
             collection.insertOne({ a: 1 }, {}, () => {})
           })
 
+          it('should have the statement tag when doing a single delete operation', done => {
+            agent.assertSomeTraces(traces => {
+              const span = traces[0][0]
+              const query = '{"a":1}'
+              const recentResource = `delete test.${collectionName}`
+              const oldResource = `remove test.${collectionName}`
```
bengl: You don't need to set both of these. You only need one of them based on the current version. Instead you can do something like:

```suggestion
              const expectedResource = `${semver.intersects(version, '>=4') ? 'delete', 'remove'} test.${collectionName}`
```

And then just refer to `expectedResource` in you expectation below. You can do this for all the tests that are checking the resource.

bengl: Actually, you're always making that semver check, so you can just do that in a higher-level scope (but inside `withVersions()`, of course), store it as something like `usesDelete` and use that result in all of these tests.

packages/datadog-plugin-mongodb-core/test/mongodb.spec.js
```diff
@@ -109,6 +112,138 @@ describe('Plugin', () => {
             collection.insertOne({ a: 1 }, {}, () => {})
           })
 
+          it('should have the statement tag when doing a single delete operation', done => {
```
bengl: All these tests could probably be rewritten as async functions. Note that the test framework allows for async functions, and will treat rejected promises as failures. You may need to change the order in which things are called (i.e. make the DB operation and _then_ call `agent.assertSomeTraces()`), but that likely doesn't matter.

It's true that most (all?) of this file does _not_ use async functions for tests, and instead does the `.then(done).catch(done)` thing, but that's mostly from copy-pasting, AFAICT.

packages/datadog-plugin-mongodb-core/src/index.js
```diff
@@ -120,6 +136,13 @@ function shouldHide (input) {
 }
 
 function limitDepth (input) {
+  if (Array.isArray(input) && input.length > 0) {
+    for (let i = 0; i < input.length; i++) {
+      const child = input[i]
+      input[i] = limitDepth(child)
```
bengl: Don't need to store `child`. You can just pass `input[i]` to `limitDepth()` directly.

packages/datadog-plugin-mongodb-core/src/index.js
```diff
@@ -90,11 +89,28 @@ function sanitizeBigInt (data) {
   return JSON.stringify(data, (_key, value) => typeof value === 'bigint' ? value.toString() : value)
 }
 
+function extractQuery (statements) {
+  if (statements.length === 1) return statements[0].q
+
+  const extractedQueries = []
+  for (let i = 0; i < statements.length; i++) {
+    const statement = statements[i]
+    extractedQueries.push(statement.q)
```
bengl: Instead of calling `limitDepth()` on what this function returns, you can just call it on every item you add to `extractedQueries`. That way you don't have to loop through them twice (once here, and once in `limitDepth()`), and you don't need to change `limitDepth()`.

packages/datadog-plugin-mongodb-core/test/mongodb.spec.js
```diff
@@ -112,6 +113,165 @@ describe('Plugin', () => {
...
+            collection.deleteMany({ a: 1 }, {}, () => {})
+
+            return agent.assertSomeTraces(traces => {
+              const span = traces[0][0]
+              const query = '{"a":1}'
+              const resource = usesDelete ? 'delete' : 'remove'
+
+              expect(span).to.have.property('resource', resource + ` test.${collectionName}`)
+              expect(span.meta).to.have.property('mongodb.query', query)
+            })
```
BridgeAR: Could we already use our newer helper for these? :)

```suggestion
            const resource = usesDelete ? 'delete' : 'remove'

            return agent.assertFirstTraceSpan({
              resource: resource + ` test.${collectionName}`,
              span: {
                'mongodb.query': '{"a":1}',
              }
            })
```

The same for the other cases.

perhardhernandez: It does seem like an appropriate change, on it!

PR #6375, opened 2025-09-03
fix: send telemetry metadata before process spawn

majorgreys:
### What does this PR do?

Update how telemetry metadata is set and sent.

### Motivation

We introduced telemetry metadata in #6203 but found the metadata was not being passed to the telemetry forwarded as anticipated.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/integration-tests/helpers/index.js
+++ b/integration-tests/helpers/index.js
@@ -131,7 +131,7 @@ function assertTelemetryPoints (pid, msgs, expectedTelemetryPoints) {
     const validResultClasses = ['success', 'incompatible_runtime', 'incompatible_library', 'internal_error', 'unknown']
 
     assert(validResults.includes(actualMetadata.result), `Invalid result: ${actualMetadata.result}`)
-    assert(validResultClasses.includes(actualMetadata.result_class), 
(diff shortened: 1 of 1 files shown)


packages/dd-trace/src/guardrails/telemetry.js
```diff
@@ -62,32 +62,31 @@ function sendTelemetry (name, tags) {
   if (points.length === 0) {
     return
   }
+
+  // Update metadata with provided result metadata
+  var currentMetadata = {}
+  for (var key in metadata) {
```
tlhunter, line 68: I was about to suggest the `...metadata` spread attribute here but then I realized this file needs to work in ancient versions of Node.js.

tlhunter: I see that you're making a clone of `metadata` then copying the incoming attributes to it. Is there a reason to not mutate `metadata` here? Is `metadata` getting sent multiple times from a given process?

tlhunter (approved):

Issue #5324, opened 2025-02-26
[BUG]: Crash when copiying headers with bun


romain-deffains:
### Tracer Version(s)

5.30.0, 5.38.0

### Node.js Version(s)

22.13.1

### Bug Report

We are using ddtrace inside a server using NestJS + Fastify and running it with bun.

When we do a request we always have this error: `Attempting to define property on object that is not extensible.`

By debugging it we found out that the code responsible for it is the `wrapWriteHead` function in [this file](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/plugins/util/web.js).
And more precisely this part: `headers = Object.assign(res.getHeaders(), headers)`

It happens only when headers is empty because Bun defines the return of getHeaders to be `ObjectFreeze({ __proto__: null })` when there is no headers.
So when ddtrace do a `Object.assign` on it, it crashes.

Is this safe to modify the object returned by the `getHeaders` function ? Is copying the headers before modifying it better ?

### Tracer Config

none

### Operating System

Darwin Kernel Version 24.0.0

### Bundling

Unsure

BridgeAR:
@romain-deffains thank you for the report! Bun is just not officially supported. I guess in case you'd open a PR to fix it on your own, it might be merged anyway. You could also reach out to the support to ask for it to be handled.

Related issue: https://github.com/DataDog/dd-trace-js/issues/3749

Issue #3749, opened 2023-10-28
error: Cannot find module "./hooks" from "[object Object]"


diavrank:
Hi, I'm trying to run my app with Bun but I'm getting this error and it seems the root cause is the package of dd-trace. I know this is not your responsibility to fix this kind of issues, however, I wonder if there's any chance you take a look in case you can support the Bun project by fixing the problem from your side or giving feedback about what is the incompatibility issue with this library and Bun.

**Expected behaviour**

Start the Nestjs server successfully with Bun.

NPM script:

"start": "NODE_ENV=localdev nest start --watch --exec \"bun run\"",

**Actual behaviour**

It's throwing the following error when starting a Nest.js app:

```ts
[2:20:07 AM] Starting compilation in watch mode...

[2:20:24 AM] Found 0 errors. Watching for file changes.



error: Cannot find module "./hooks" from "[object Object]"
```

**Steps to reproduce**

1. Create a Nestjs app
2. Install the dependency `ddt-trace` with Bun 1.0.8 
3. Execute bun start command:

Have the start script as:
"start": "NODE_ENV=localdev nest start --watch --exec \"bun run\"",

**Environment**

* **Operation system:** Mac OS Sonoma
* **Node.js version:**  v18.16.1 (But it was run with Bun 1.0.8)
* **Tracer version:**  ^4.12.0" (Additionally, "nestjs-ddtrace": "^3.0.4",)
* **Agent version:** what is this?
* **Relevant library versions:** Bun 1.0.8

diavrank:
Btw, I also reported the same issue in Bun repo https://github.com/oven-sh/bun/issues/6374

Koka:
For anyone looking for a way to enable DD tracing in a Bun runtime - there's a (hacky) way to do so (see gist https://gist.github.com/Koka/f1801289b526eeb5853894cc4e3e747d)

Short summary - I had to monkey patch dogstatsd.CustomMetrics to avoid sending metrics through UDP (not supported by Bun runtime yet) and add a shim BlockList to node net module (blocklist is not supported by Bun runtime yet and shim is built in but detection doesn't play well with Bun)

Using this technique you can run functional APM tracing without any metrics being sent, and without any integrations, so you have to call tracer.trace manually

Jarred-Sumner:
`dd-trace` should work in Bun as of Bun v1.1.6 (current: Bun v1.1.10)

[...]

PR #6386, opened 2025-09-04
fix: make Object.assign compatible with non-extensible headers in Bun runtime

Scoup:
### What does this PR do?
make Object.assign compatible with non-extensible headers in Bun runtime

### Motivation
The `res.getHeaders()` method in Bun returns a non-extensible object, causing
`Object.assign(res.getHeaders(), headers)` to throw `TypeError: Attempting to 
define property on object that is not extensible`.

This fix creates a new extensible object as the target for Object.assign,
ensuring compatibility with both Node.js and Bun runtimes.

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes

Related issues: 
- https://github.com/DataDog/dd-trace-js/issues/5324
- https://github.com/DataDog/dd-trace-js/issues/3749

### DIFF
--- a/packages/dd-trace/src/plugins/util/web.js
+++ b/packages/dd-trace/src/plugins/util/web.js
@@ -387,7 +387,7 @@ const web = {
 
     return function (statusCode, statusMessage, headers) {
       headers = typeof statusMessage === 'string' ? headers : statusMessage
-      headers = Object.assign(res.getHeaders(), headers)
+      headers = Object.assign({}, res.getHeaders(), headers)
 
       if (req.method.toLowerCase() === 'options' && isOriginAllowed(req, headers)) {
         addAllowHeaders(req, res, headers)


packages/dd-trace/src/plugins/util/web.js
```diff
@@ -387,7 +387,7 @@ const web = {
 
     return function (statusCode, statusMessage, headers) {
       headers = typeof statusMessage === 'string' ? headers : statusMessage
-      headers = Object.assign(res.getHeaders(), headers)
+      headers = Object.assign({}, res.getHeaders(), headers)
```
BridgeAR: Is it potentially so that either res.getHeaders() or headers could be empty? If that's the case, we could likely prevent the copy all together? I didn't check the overall code though.

I might be wrong but I believe this will fail our linter. Could you please check that and add an exception if that's true? I think we should add a comment about Bun one way or the other :)

BridgeAR (commented):
I guess this is fine, even if we do not officially support Bun

packages/dd-trace/src/plugins/util/web.js
Scoup: Thanks for the feedback! You're absolutely right about the optimization potential. I've implemented the change to prevent unnecessary object copying when both res.getHeaders() and headers are empty.

BridgeAR: That will actually make it more expensive, since we now have to allocate the arrays for the strings 😅 
Thus, the former implementation was actually better 😇

PR #6388, opened 2025-09-05
Inject git metadata in datadog esbuild plugin

olivwalsh:
### What does this PR do?
Automatically retrieve and inject Git tags (GIT_REPOSITORY_URL and GIT_COMMIT_SHA) in apps bundled with the Datadog esbuild plugin.

### Motivation
Enable Datadog customers using the tracer esbuild plugin to automatically have their traces tagged with Git metadata. This unlocks part of the [Source Code Integration](https://docs.datadoghq.com/integrations/guide/source-code-integration/?tab=githubsaasonprem) features.

### Plugin Checklist
- [x] Unit tests.
- [x] Integration tests.


integration-tests/esbuild/build-and-run.git-tags.js
```diff
@@ -0,0 +1,57 @@
...
+    env: { ...process.env, DD_TRACE_DEBUG: 'true' }
+  })
+  if (stdout.length) {
+    console.log(stdout.toString())
+    const output = stdout.toString()
+    const repositoryURL = output.match(/"_dd\.git\.repository_url":"([^"]+)"/)?.[1]
+    const commitSha = output.match(/"_dd\.git\.commit\.sha":"([^"]+)"/)?.[1]
+    assert.ok(repositoryURL, '_dd.git.repository_url should be present')
+    assert.ok(commitSha, '_dd.git.commit.sha should be present')
+    assert.ok(commitSha.length === 40, 'Git commit sha tag should be valid')
```
p2kmgcl: 🤏 I would use assert.equal here, just for some readability if it fails

integration-tests/esbuild/build-and-run.git-tags.js
```diff
@@ -0,0 +1,57 @@
...
+    'pg-query-stream',
+    'tedious'
+  ]
+}).then(() => {
+  const { status, stdout, stderr } = spawnSync('node', [SCRIPT], {
+    env: { ...process.env, DD_TRACE_DEBUG: 'true' }
+  })
+  if (stdout.length) {
+    console.log(stdout.toString())
+    const output = stdout.toString()
```
p2kmgcl: 🤏 swap these two lines so you do not need to convert the stdout twice 😅

packages/datadog-esbuild/index.js
```diff
@@ -77,6 +111,27 @@ ${build.initialOptions.banner.js}`
     }
   }
 
+  // Get git metadata at build time and add it to the banner for both ESM and CommonJS builds
+  const gitMetadata = getGitMetadata()
+  if (gitMetadata.repositoryURL || gitMetadata.commitSHA) {
+    build.initialOptions.banner ??= {}
+    build.initialOptions.banner.js ??= ''
+
+    build.initialOptions.banner.js = `if (typeof process !== 'undefined' && process.env) {
```
watson: This will also guard against process being anything else than an object:
  
```suggestion
    build.initialOptions.banner.js = `if (typeof process === 'object' && process !== null && process.env !== null && typeof process.env === 'object') {
```

integration-tests/esbuild/index.spec.js
```diff
@@ -87,6 +87,14 @@ esbuildVersions.forEach((version) => {
       })
     })
 
+    it('injects Git metadata into bundled applications', () => {
+      const command = 'node ./build-and-test-git-tags.js'
+      console.log(command)
```
watson, line 92: Did you mean to leave this `console.log` statement in the test?

integration-tests/esbuild/build-and-test-git-tags.js
```diff
@@ -0,0 +1,57 @@
...
+    const commitSha = output.match(/"_dd\.git\.commit\.sha":"([^"]+)"/)?.[1]
+    assert.ok(repositoryURL, '_dd.git.repository_url should be present')
+    assert.ok(commitSha, '_dd.git.commit.sha should be present')
+    assert.equal(commitSha.length, 40, 'Git commit sha tag should be valid')
+  }
+  if (stderr.length) {
+    console.error(stderr.toString())
+  }
+  if (status) {
+    throw new Error('generated script failed to run')
```
watson: Nit: just to help with debugging in case `status` isn't `0`

```suggestion
    throw new Error(`Generated script exited with unexpected exit code: ${status}`)
```

integration-tests/esbuild/build-and-test-git-tags.js
```diff
@@ -0,0 +1,57 @@
...
+    const output = stdout.toString()
+    console.log(output)
+    const repositoryURL = output.match(/"_dd\.git\.repository_url":"([^"]+)"/)?.[1]
+    const commitSha = output.match(/"_dd\.git\.commit\.sha":"([^"]+)"/)?.[1]
+    assert.ok(repositoryURL, '_dd.git.repository_url should be present')
+    assert.ok(commitSha, '_dd.git.commit.sha should be present')
+    assert.equal(commitSha.length, 40, 'Git commit sha tag should be valid')
+  }
+  if (stderr.length) {
+    console.error(stderr.toString())
```
watson: This might be a contrived example, but are we expecting any output to STDERR? If not, there might be an edge case in which `status` is `0` but STDERR stil contains data which should fail the test right?

If so, I'd probably just merge these two if-statements into one. Something like:

```js
if (stderr.length || status) {
  console.error(stderr.toString())
  throw new Error(`Unexpected result when running generated script (exit code: ${status})`)
}
```

integration-tests/esbuild/build-and-test-git-tags.js
```diff
@@ -0,0 +1,57 @@
...
+    'pg-query-stream',
+    'tedious'
+  ]
+}).then(() => {
+  const { status, stdout, stderr } = spawnSync('node', [SCRIPT], {
+    env: { ...process.env, DD_TRACE_DEBUG: 'true' }
+  })
+  if (stdout.length) {
+    const output = stdout.toString()
+    console.log(output)
```
watson: Couldn't we get rid of this `console.log` line?

integration-tests/esbuild/build-and-test-git-tags.js
```diff
@@ -0,0 +1,57 @@
...
+    'better-sqlite3',
+    'sqlite3',
+    'mysql',
+    'oracledb',
+    'pg-query-stream',
+    'tedious'
+  ]
+}).then(() => {
+  const { status, stdout, stderr } = spawnSync('node', [SCRIPT], {
+    env: { ...process.env, DD_TRACE_DEBUG: 'true' }
```
watson: Nit: If you set encoding to be `utf8`, the `stdout` and `stderr` returned here will not be buffers but stings, so you don't need to call `toString()` on them below

```suggestion
    env: { ...process.env, DD_TRACE_DEBUG: 'true' },
    encoding: 'utf8'
```

integration-tests/esbuild/index.spec.js
olivwalsh: Yes for consistency with the other tests, as they log the command as well. Do you think it shouldn't?

integration-tests/esbuild/build-and-test-git-tags.js
olivwalsh: I think that we may have non-fatal errors outputted to STDERR since this test is using the option `DD_TRACE_DEBUG = true`, so I don't think we should throw an error in this case. 
However, I was missing a condition verifying that STDOUT has information outputted, as it is expected when using `DD_TRACE_DEBUG = true`.

olivwalsh: If you think the test should fail if STDERR has any output, should I modify the other tests that are currently not handling this case? https://github.com/DataDog/dd-trace-js/blob/f2f650b5288023f4cb565857dac1ee64d6e29595/integration-tests/esbuild/build-and-test-aws-sdk.js#L26

watson: Ok, if this is what has been done in the other tests, let's just leave it like this for now

watson (commented):
Your changes look good now. But apparently the `tests/docker_ssi/test_docker_ssi.py` test in GitLab is failing. I've verified that even though there's a billion failures in CI, they are all the same error, e.g: https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-js/-/jobs/1131545873

olivwalsh:
@watson thanks for reviewing again. Just updated the branch and now all CI tests are passing 🥳

Issue #6307, opened 2025-08-19
[BUG]: Quarantine causes playwright test to be skipped


zemek:
### Tracer Version(s)

5.63.2

### Node.js Version(s)

18.20.8, 20.19.4

### Bug Report

Marking a test as `Quarantined` in the Flaky Management UI causes playwright tests to be skipped entirely (same behavior as if the test was marked as `Disabled`).

I've attached a minimal reproducible config (a dummy test that is controlled pass/fail via an env var).

You can see in the debug logs provided:
```
"test.status":"skip","test.test_management.is_quarantined":"true"
```

I would expect the test to still be run and report the failure to datadog, but the failure to not block CI.

Notice how the test is just marked as skipped in datadog:
<img width="2369" height="692" alt="Image" src="https://github.com/user-attachments/assets/24ba296e-ed4f-4e73-98b4-46b821e32279" />

### Reproduction Code

[package.json](https://github.com/user-attachments/files/21867685/package.json)

[playwright.config.ts](https://github.com/user-attachments/files/21867712/playwright.config.ts)

[example.spec.ts](https://github.com/user-attachments/files/21867714/example.spec.ts)

### Error Logs

```
DD_TRACE_DEBUG=true DD_SERVICE=jz_test DD_API_KEY=<redacted> DD_CIVISIBILITY_AGENTLESS_ENABLED=true NODE_OPTIONS="-r dd-trace/ci/init" FORCE_TEST_FAILURE=true yarn test
dd-trace is not initialized in a package manager.
yarn run v1.22.22
warning package.json: No license field
$ playwright test
Extract from carrier (datadog, tracecontext, baggage): {"x-datadog-trace-id":"7155915214952172796","x-datadog-parent-id":"0000000000000000"}.
Uploading git history for repository git@github.com:<redacted>
There were 1000 commits since last month.
Flushing 0 metrics via HTTP
Remote settings: {"isCodeCoverageEnabled":false,"isSuitesSkippingEnabled":false,"isItrEnabled":false,"requireGit":false,"isEarlyFlakeDetectionEnabled":true,"earlyFlakeDetectionNumRetries":10,"earlyFlakeDetectionFaultyThreshold":30,"isFlakyTestRetriesEnabled":true,"isDiEnabled":true,"isKnownTestsEnabled":true,"isTestManagementEnabled":true,"testManagementAttemptToFixRetries":20,"isImpactedTestsEnabled":false}
Number of received known tests: 4

[...]

PR #6397, opened 2025-09-08
fix(#6307): quarantine tests are being run but exit with status code 0

calvinbayer:
### What does this PR do?
- If a test is quarantined, the test is executed but if it fails, it should not fail the pipeline (i.e., status should exit with code `0`)
- Previously, the test was skipped

### Motivation
- https://github.com/DataDog/dd-trace-js/issues/6307

### DIFF
--- a/packages/datadog-instrumentations/src/playwright.js
+++ b/packages/datadog-instrumentations/src/playwright.js
@@ -64,6 +64,11 @@ let quarantinedButNotAttemptToFixFqns = new Set()
 let rootDir = ''
 const MINIMUM_SUPPORTED_VERSION_RANGE_EFD = '>=1.38.0' // TODO: remove this once we drop support for v5
 
+function getTestFullyQualifiedName (test) {
+  const fullname = getTestFullname(test)
+  return `${test._requireFile} ${fullname}`
+}
+
 function getTestProperties (test) {
   const testName = getTestFullname(test)
   const testSuite = getTestSuitePath(test._requireFile, rootDir)
@@ -328,8 +333,7 @@ function testEndHandler (test, annotations, testStatus, error, isTimeout, isMain
     return
   }
 
-  const testFullName = getTestFullname(test)
-  const testFqn = `${testSuiteAbsolutePath} ${testFullName}`
+  const testFqn = getTestFullyQualifiedName(test)
   const testStatuses = testsToTestStatuses.get(testFqn) || []
 
   if (testStatuses.length === 0) {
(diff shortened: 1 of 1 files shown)


packages/datadog-instrumentations/src/playwright.js
```diff
@@ -776,8 +785,13 @@ addHook({
           if (testProperties.disabled || testProperties.quarantined) {
             quarantinedOrDisabledTestsAttemptToFix.push(test)
           }
-        } else if (testProperties.disabled || testProperties.quarantined) {
+        } else if (testProperties.disabled) {
           test.expectedStatus = 'skipped'
+        } else if (testProperties.quarantined) {
+          // Do not skip quarantined tests, let them run and overwrite results post-run if they fail
+          const fullname = getTestFullname(test)
```
juan-fernandez: we should probably extract this to a function in this file, as this logic is found in other places

integration-tests/playwright/playwright.spec.js
```diff
@@ -1288,20 +1288,14 @@ versions.forEach((version) => {
               const events = payloads.flatMap(({ payload }) => payload.events)
 
               const testSession = events.find(event => event.type === 'test_session_end').content
-              if (isQuarantining) {
```
juan-fernandez, line 1291: shouldn't we also check that the test session is passing? so that we know that the error is actually suppressed

juan-fernandez: actually, we should modify the test to assert that the test session fails or passes depending on whether we're quarantining or not

calvinbayer: We check the exit code here: https://github.com/DataDog/dd-trace-js/pull/6397/files/e36043bda26333a45225dc1566cbe7e00ddc5990#diff-68b5e7860766c25caad41f6fbfce1920923567d3c88cbf7cf24894ce1bd8e305L1328-L1332

For the test session status: Right now it is reported as failed which is the same behavior as Jest. Let's follow up on that separately in case that's not the desired behavior

packages/datadog-instrumentations/src/playwright.js
calvinbayer: I agree, refactored the code to extract that into a separate function

integration-tests/playwright/playwright.spec.js
calvinbayer: As discussed in Slack, created a separate task to make sure we report the suite/module/session as passed back to Datadog: https://datadoghq.atlassian.net/browse/SDTEST-2611

juan-fernandez (approved):
🚀

Issue #5479, opened 2025-03-25
[BUG]: when using ESM, integrations won't be registered if the hook uses a file


FredericEspiau:
### Tracer Version(s)

5.43.0

### Node.js Version(s)

22.14.0

### Bug Report

Hey and thanks for the hard work,

let's take the example of `express`: https://github.com/DataDog/dd-trace-js/blob/fac89882fd35c07dae6bf91280bb2e7a9d8a0a25/packages/datadog-instrumentations/src/express.js#L104

```js
addHook({
  name: 'express',
  versions: ['4'],
  file: 'lib/middleware/query.js'
}, query => {
  // redacted
})
```

Here, you add a hook for the `express` integration that should work for all version `4.*.*`. It should look for the `lib/middleware/query.js` file.

This hook will work if both the version matches and the file is present.

Unfortunately, there is a difference between `require-in-the-middle` and `import-in-the-middle` that will prevent this from working.

The difference will happen around here: https://github.com/DataDog/dd-trace-js/blob/fac89882fd35c07dae6bf91280bb2e7a9d8a0a25/packages/datadog-instrumentations/src/helpers/register.js#L96

If you use `cjs`, `fullFilename` will be `express/lib/middleware/query.js` and `moduleName` will be `express/lib/middleware/query.js`

If you use `esm`, `fullFilename` will be `express/lib/middleware/query.js` and `moduleName` will be `express`

This doesn't seem to be an issue with `dd-trace-js` but rather with `import-in-the-middle`.

I think the issue is here: https://github.com/nodejs/import-in-the-middle/blob/53a33a9b07799bff815864089a0c072d223df47b/index.js#L147

This looks like an optimisation to prevent sending unrelated files to the callbacks but it prevents lots of files to be sent to the callback, including the ones we need.

If you log the files sent to the `Hook`, you will notice that `require-in-the-middle` sends much more files than `import-in-the-middle`.

If you remove this line `if (!baseDir.endsWith(specifiers.get(filename))) continue` in `import-in-the-middle` it works again.

As I'm really not sure if the issue is with `dd-trace-js` or `import-in-the-middle`, I'm creating an issue both here and there.

Note: this should fix this issue as well: https://github.com/DataDog/dd-trace-js/issues/5092

Related issue in `import-in-the-middle`: https://github.com/nodejs/import-in-the-middle/issues/185

Thanks in advance.

### Bundling

No Bundling

Issue #3500, opened 2023-08-03
Pino and dd-trace not working together


furstenheim-goodnotes:
**Expected behaviour**
Pino should integrate nicely with datadog
**Actual behaviour**
I'm seeing something similar to https://github.com/DataDog/dd-trace-js/issues/2189. Logs of pino do not include the datadog fields

**Steps to reproduce**

I'm using `"dd-trace": "4.11.0"` and `"pino": "8.14.2"`
```
// index.ts
// Import tracer first thing so that pino can be hooked
import tracer from './metrics/Tracer.js';

// import { pino } from "pino";

(async function main () {
    await tracer.trace('Some Trace', async () => {
        console.log(tracer.scope().active())
        const logger = pino({
        } )
        logger.info('Starting service...')    })
})()
```

```
// Tracer.ts
import tracer from 'dd-trace'
// BTW, types are wrong here, it claims that tracer.init does not exist. I've also tried tracer.tracer.init(..) but same result as this bug
// @ts-ignore
tracer.init({
    logInjection: true,
})

// also tried
/*
tracer.tracer.init({
    logInjection: true,
})
*/

// Note that docs https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/nodejs/?tab=containers suggest exporting just tracer but that seems to break
export default tracer.tracer
```

This logs:

```
// DatadogSpan { ... 
// {"level":30,"time":1691057718718,"pid":59863,"hostname":"mm","msg":"Starting service..."}
```

From the logs I can see that the scope of the tracer is defined. I'm even creating the logger inside the scope just for the testing.

**Environment**

* **Operation system:**
Linux
* **Node.js version:**
Node 18
* **Tracer version:**
`"dd-trace": "4.11.0"`
* **Agent version:**
* **Relevant library versions:**
`"pino": "8.14.2"`

furstenheim-goodnotes:
BTW, with winston it does work. So I guess that all the hooks to datadog are in place

zomgbre:
I'm seeing the exact same thing.

ChrisBellew:
I think I know what is broken.

This works
```
import pino from "pino";
```

This does not work for log injection even though the pino logger itself still works.
```
import { pino } from "pino";
```

Don't ask me how many days it took to figure that out :)

tlhunter:
@khanayan123 do you think this could be an ESM bug based on the comment that @ChrisBellew left?

BridgeAR:
[...]

PR #6401, opened 2025-09-08
remove special-casing of default in hook.js

bengl:
### What does this PR do?

See title. Plus:

* appropriate testing
* modifying instrumentations when needed

### Motivation

The previous model of assuming `default === exports` was very incorrect for ESM since they're never the exact same object. We also can't assume they should have the same properties, or even that they're both _objects/functions_, at least not in the general case.

Therefore, we need to be specific about what's being instrumented in _every single instrumentation_.

### Additional Notes

This is obviously in a draft state. The only instrumentation edited so far is `http2`.

This should fix these issues:
Refs: https://github.com/DataDog/dd-trace-js/issues/3500
Refs: https://github.com/DataDog/dd-trace-js/issues/5479

### DIFF
--- a/packages/datadog-instrumentations/src/helpers/hook.js
+++ b/packages/datadog-instrumentations/src/helpers/hook.js
@@ -62,6 +62,7 @@ function Hook (modules, hookOptions, onrequire) {
 Hook.prototype.unhook = function () {
   this._ritmHook.unhook()
   this._iitmHook.unhook()
+  this._patched = Object.create(null)
 }
 
 module.exports = Hook


integration-tests/helpers/index.js
```diff
@@ -381,7 +381,7 @@ function checkSpansForServiceName (spans, name) {
 
 async function spawnPluginIntegrationTestProc (cwd, serverFile, agentPort, stdioHandler, additionalEnvArgs = {}) {
   let env = {
-    NODE_OPTIONS: `--loader=${hookFile}`,
+    NODE_OPTIONS: `--no-warnings --loader=${hookFile}`,
```
BridgeAR: I fear we might cause confusion with deactivating warnings. Is that only needed for a particular Node.js version?

bengl: Yeah I can get rid of that. I just was getting annoyed by the warnings during testing 😆.

packages/datadog-instrumentations/src/fastify.js
```diff
@@ -265,21 +265,21 @@ function canPublishResponsePayload (payload) {
     !ArrayBuffer.isView(payload) // TypedArray
 }
 
-addHook({ name: 'fastify', versions: ['>=3'] }, fastify => {
-  const wrapped = shimmer.wrapFunction(fastify, fastify => wrapFastify(fastify, true))
+addHook({ name: 'fastify', versions: ['>=3'] }, (fastify, _1, _2, isIitm) => {
```
rochdev: As discussed offline, the fact that this is needed is probably due to a bug or something not properly handled in the new implementation. Let's fix it before landing.

packages/datadog-instrumentations/src/helpers/hook.js
```diff
@@ -24,25 +24,31 @@ function Hook (modules, hookOptions, onrequire) {
...
     /**
      * TODO: Find a way to deal with modules that have barrel files, and
-     * add contents after each require, which break our moduleExport caching mechanism 
+     * add contents after each require, which break our moduleExport caching mechanism
      * (example: protobufjs)
      */
     if (newExports &&
       (typeof newExports === 'object' || typeof newExports === 'function') &&
-      (moduleName === 'protobufjs' || !moduleName.includes('protobuf'))) {
+      (moduleName === 'protobufjs' || !moduleName.includes('protobuf')) && !moduleName.includes('aws-sdk')) {
```
bengl: @perhardhernandez Lets put a comment here indicating why these are special-cased. We ought to also be careful with `includes`, and instead specify as precisely as we can.

packages/datadog-instrumentations/src/helpers/hook.js
```diff
@@ -24,25 +24,31 @@ function Hook (modules, hookOptions, onrequire) {
...
+    let defaultWrapResult
     if (
       isIitm &&
       moduleExports.default &&
       (typeof moduleExports.default === 'object' ||
         typeof moduleExports.default === 'function')
     ) {
-      newExports.default = onrequire(moduleExports.default, moduleName, moduleBaseDir, moduleVersion, isIitm)
+      defaultWrapResult = onrequire(moduleExports.default, moduleName, moduleBaseDir, moduleVersion, isIitm)
+      if (moduleName.includes('express')) {
```
bengl: @perhardhernandez Lets put a comment here indicating why this special-cased. We ought to also be careful with includes, and instead specify as precisely as we can.

packages/datadog-plugin-pino/test/integration-test/client.spec.js
```diff
@@ -12,12 +13,18 @@ describe('esm', () => {
...
+  let variants
 
   withVersions('pino', 'pino', version => {
     before(async function () {
       this.timeout(20000)
       sandbox = await createSandbox([`'pino@${version}'`],
         false, ['./packages/datadog-plugin-pino/test/integration-test/*'])
+      variants = varySandbox(sandbox, 'server.mjs', {
+        default: 'import pino from \'pino\'',
+        star: 'import * as modPino from \'pino\'; const pino = (...a) => modPino.default(...a)',
```
BridgeAR: Should this not be:

```suggestion
        star: 'import * as modPino from \'pino\'; const pino = modPino.default',
```

packages/datadog-plugin-pino/test/integration-test/client.spec.js
```diff
@@ -12,12 +13,18 @@ describe('esm', () => {
...
+        destructure: 'import { default as pino } from \'pino\'',
```
BridgeAR: I think we miss one more case that we should also test:

import { pino } from 'pino'

I believe that might not be the case for all modules, but definitely for some.

packages/datadog-plugin-connect/test/integration-test/client.spec.js
```diff
@@ -14,13 +15,18 @@ describe('esm', () => {
...
-  withVersions('connect', 'connect', '>=3', version => {
+  withVersions('connect', 'connect', version => {
     before(async function () {
       this.timeout(20000)
       sandbox = await createSandbox([`'connect@${version}'`], false, [
         './packages/datadog-plugin-connect/test/integration-test/*'])
+      variants = varySandbox(sandbox, 'server.mjs', {
+        default: 'import connect from \'connect\'',
+        star: 'import * as starConnect from \'connect\'; const connect = starConnect.default;',
+        destructure: 'import { default as connect } from \'connect\';'
```
BridgeAR: Since most imports always look like these, should we just have shortcut that accepts the name of the module and that creates these three possible test cases automatically?
That way we only have to define them in more detail in case the default case does not work as expected, which is less frequent.

packages/datadog-plugin-azure-functions/test/integration-test/client.spec.js
```diff
@@ -43,6 +43,10 @@ describe('esm', () => {
       await agent.stop()
     })
 
+    // TODO(bengl): The `varySandbox` helper function isn't well set-up for dealing
+    // with Azure Functions and the way the `func` command expects to find files. I
+    // have manually tested that all the usual import variants work, but really we ought
+    // to figure out a way of automating this.
```
BridgeAR, line 49: We have to memorize to do that again before landing with the latest changes.

packages/datadog-instrumentations/src/helpers/hook.js
```diff
@@ -19,37 +18,46 @@ function Hook (modules, hookOptions, onrequire) {
...
+  const patched = new WeakMap()
   this._patched = Object.create(null)
 
-  const safeHook = (moduleExports, moduleName, moduleBaseDir, moduleVersion) => {
+  const safeHook = (moduleExports, moduleName, moduleBaseDir, moduleVersion, isIitm) => {
     const parts = [moduleBaseDir, moduleName].filter(Boolean)
     const filename = path.join(...parts)
 
-    if (this._patched[filename]) return moduleExports
+    if (this._patched[filename] && patched.has(moduleExports)) return patched.get(moduleExports)
```
BridgeAR: I am uncertain why we use both cases. Would it be possible to get a comment explaining this? :)

packages/datadog-instrumentations/src/cassandra-driver.js
```diff
@@ -45,7 +45,11 @@ addHook({ name: 'cassandra-driver', versions: ['>=3.0.0'] }, cassandra => {
   return cassandra
 })
 
-addHook({ name: 'cassandra-driver', versions: ['>=4.4'] }, cassandra => {
+addHook({ name: 'cassandra-driver', versions: ['>=4.4'] }, (cassandra, _1, _2, isIitm) => {
```
BridgeAR: Before landing, it is probable good to rename _1 and _2 to the actual variable names, just with a leading underscore. That applies to all cases.

rochdev: Since this same check is in multiple places, could it be an option instead? For example:

```js
// Could be another name too, I'm not 100% what this even does.
{ name: '@azure/functions', versions: ['>=4'], patchDefault: false }
```

Then each individual hook doesn't need to know how to handle it and it fixes the weird signature of the function.

packages/datadog-instrumentations/src/helpers/hook.js
rochdev: I'm pretty sure using both doesn't even do anything? What happens if the set is removed?

packages/datadog-instrumentations/src/helpers/hook.js
```diff
@@ -19,37 +18,46 @@ function Hook (modules, hookOptions, onrequire) {
...
+
+    let defaultWrapResult
+    if (
+      isIitm &&
+      moduleExports.default &&
+      (typeof moduleExports.default === 'object' ||
```
rochdev, line 37: Why is this check needed? If a hook is requesting the export it should always be called.

packages/datadog-instrumentations/src/helpers/hook.js
```diff
@@ -19,37 +18,46 @@ function Hook (modules, hookOptions, onrequire) {
...
+    }
+
+    const newExports = onrequire(moduleExports, moduleName, moduleBaseDir, moduleVersion, isIitm)
+
+    if (defaultWrapResult) newExports.default = defaultWrapResult
 
-    this._patched[filename] = true
+    if (newExports &&
+      (typeof newExports === 'object' || typeof newExports === 'function')) {
+      patched.set(moduleExports, newExports)
```
rochdev, line 51: What does this do at this point since this code should never run more than once for an export now (since we switched back to using the filename)

packages/datadog-instrumentations/src/helpers/hook.js
perhardhernandez: Surprisingly we encountered some cases in which the moduleExports was undefined

packages/datadog-instrumentations/src/helpers/hook.js
perhardhernandez: Did we confirm that an export never runs more than once? If so why did we have the following 

`if (this._patched[filename]) return moduleExports`

The idea of keeping both was to make sure we were always returning the patched export.

packages/datadog-instrumentations/src/helpers/hook.js
perhardhernandez: It coming from isIitm does not ensure that .default exist, hence the condition.

packages/datadog-instrumentations/src/helpers/hook.js
perhardhernandez: The idea of keeping both was to ensure that if we ever encounter the same filename twice, we return the patched export, if that's not possible, is this._patched even needed (considering onrequire takes case of filename verification)? And if it is possible, my understanding is that moduleExports always enters our process as the original export object.

packages/datadog-instrumentations/src/cassandra-driver.js
perhardhernandez: The above, was to prevent double wrapping when moduleExports.default and moduleExports share a same prototype, handling it as Roch suggested would work.

packages/datadog-plugin-pino/test/integration-test/client.spec.js
perhardhernandez: That does work as well.

packages/datadog-instrumentations/src/helpers/hook.js
rochdev: Sure, but why the explicit type check? Wouldn't checking just for the presence enough?

packages/datadog-instrumentations/src/cassandra-driver.js
BridgeAR: I am fine either way

packages/datadog-instrumentations/src/helpers/hook.js
perhardhernandez: Express and next have some weird exports that are not of those types, hence the check.

packages/datadog-plugin-pino/test/integration-test/client.spec.js
perhardhernandez: That export is not possible as 'pino'  is not directly exposed.

BridgeAR: I did not check if it's pino in new versions or Node.js that allows it, but it's definitely possible to load pino like that with new versions :)

packages/datadog-instrumentations/src/pino.js
```diff
@@ -73,22 +73,34 @@ function wrapPrettyFactory (prettyFactory) {
   }
 }
 
-addHook({ name: 'pino', versions: ['2 - 3', '4', '>=5 <5.14.0'] }, pino => {
+addHook({ name: 'pino', versions: ['2 - 3', '4'], patchDefault: false }, (pino, _1, _2, isIitm) => {
```
rochdev: Is `isIitm` still needed here? The hook shouldn't have to know how to get the right export.

packages/datadog-plugin-pino/test/integration-test/client.spec.js
perhardhernandez: You are right, it does seem like it's pino version issue. If we want to add into the tests we would need to something like in the pg integration test file.

packages/datadog-instrumentations/src/pino.js
perhardhernandez: Yes, with out it pino tests fail. I'm guessing this is related to some of the weird things that pino do with their exports.

rochdev: Can you look into it a bit more? I'd love to be able to just remove that parameter entirely if we can, and the hook should already be receiving both the export and the default. I wonder if it's because Pino needs the opposite of `patchDefault`?

rochdev (dismissed):

PR #6404, opened 2025-09-09
[test optimization] Do not enable Test Optimization plugins when `isCiVisibility` is false

juan-fernandez:
### What does this PR do?

Do not initialize test optimization plugins if `config.isCiVisibility` is false (that is, if `ci/init` is not used explicitly).

### Motivation
Users should be able to run their test suite with:

```javascript
'use strict'

const tracer = require('dd-trace').init()

tracer.trace('sum.test', { resource: 'sum.test.js' }, () => {
  describe('sum', () => {
    it('should return the sum of two numbers', () => {
      return expect(1 + 2).toBe(3)
    })
  })
})
```

without using Test Optimization at all. 

### Plugin Checklist

- [x] Integration tests.

### DIFF
--- a/integration-tests/ci-visibility/test-optimization-wrong-init.spec.js
+++ b/integration-tests/ci-visibility/test-optimization-wrong-init.spec.js
@@ -1,8 +1,8 @@
 'use strict'
 
+const { once } = require('node:events')
+const assert = require('node:assert')
 const { exec } = require('child_process')
-
-const { assert } = require('chai')
 
 const { createSandbox, getCiVisAgentlessConfig } = require('../helpers')
 const { FakeCiVisIntake } = require('../ci-visibility-intake')
@@ -27,7 +27,7 @@ const testFrameworks = [
     expectedOutput: '1 passed',
     extraTestContext: {
       TEST_DIR: 'ci-visibility/test-optimization-wrong-init/vitest-sum-wrong-init*',
-      NODE_OPTIONS: '--import dd-trace/register.js -r dd-trace/init'
+      NODE_OPTIONS: '--import dd-trace/register.js'
     }
   },
   {
@@ -48,7 +48,7 @@ testFrameworks.forEach(({ testFramework, command, expectedOutput, extraTestConte
     if (NODE_MAJOR <= 18 && testFramework === 'cucumber') return
 
     before(async () => {
-      const testFrameworks = ['jest', 'mocha', 'chai@v4', 'vitest']
+      const testFrameworks = ['jest', 'mocha', 'vitest']
 
       // Remove once we drop support for Node.js@18
       if (NODE_MAJOR > 18) {
@@ -73,7 +73,7 @@ testFrameworks.forEach(({ testFramework, command, expectedOutput, extraTestConte
       await receiver.stop()
     })
 
-    it('does not initialize test optimization plugins if Test Optimization mode is not enabled', (done) => {
+    it('does not initialize test optimization plugins if Test Optimization mode is not enabled', async () => {
       const eventsPromise = receiver
         .gatherPayloadsMaxTimeout(({ url }) => url === '/v0.4/traces', (tracesRequests) => {
           const spans = tracesRequests.flatMap(trace => trace.payload).flatMap(request => request)
@@ -115,15 +115,17 @@ testFrameworks.forEach(({ testFramework, command, expectedOutput, extraTestConte
         processOutput += chunk.toString()
       })
 
-      childProcess.on('exit', () => {
-        assert.include(processOutput,
+      await Promise.all([
+        once(childProcess, 'exit'),
+        eventsPromise
+      ])
+
+      assert.ok(
+        processOutput.includes(
           `Plugin "${testFramework}" is not initialized because Test Optimization mode is not enabled.`
         )
-        assert.include(processOutput, expectedOutput)
-        eventsPromise.then(() => {
-          done()
-        }).catch(done)
-      })
+      )
+      assert.ok(processOutput.includes(expectedOutput))
     })
   })
 })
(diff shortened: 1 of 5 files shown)


packages/datadog-instrumentations/src/cucumber.js
```diff
@@ -747,6 +750,9 @@ function getWrappedRunTestCase (runTestCaseFunction, isNewerCucumberVersion = fa
 
 function getWrappedParseWorkerMessage (parseWorkerMessageFunction, isNewVersion) {
   return function (worker, message) {
+    if (!testSuiteFinishCh.hasSubscribers) {
+      return parseWorkerMessageFunction.apply(this, arguments)
+    }
```
juan-fernandez, line 755: unrelated changes, but disabling the cucumber plugin caused issues in the instrumentation. It's always recommended to check `.hasSubscribers` before doing logic anyway, so this was a good thing

Mariovido (approved):
LGTM

integration-tests/ci-visibility/test-optimization-wrong-init/vitest-sum-wrong-init-test.mjs
```diff
@@ -0,0 +1,13 @@
+import tracer from 'dd-trace'
+import { describe, it, expect } from 'vitest'
+
+tracer.trace('sum.test', { resource: 'sum.test.js' }, () => {
+  describe('sum', () => {
+    it('should return the sum of two numbers', async () => {
+      // we need to give time for the tracer to flush the spans
+      // otherwise the vitest worker process will exit before the spans are flushed
+      await new Promise(resolve => setTimeout(resolve, 2000))
```
calvinbayer: Instead of hardcoding this to 2s, can we reduce the flushInterval of the tracer?
https://docs.datadoghq.com/tracing/trace_collection/library_config/nodejs/#traces

packages/dd-trace/src/plugin_manager.js
```diff
@@ -6,6 +6,15 @@ const plugins = require('./plugins')
 const log = require('./log')
 const { getEnvironmentVariable } = require('../../dd-trace/src/config-helper')
 
+// Test optimization plugins that should only be enabled when isCiVisibility is true
+const TEST_OPTIMIZATION_PLUGINS = new Set([
+  'jest',
+  'vitest',
+  'cucumber',
+  'mocha',
+  'playwright'
```
calvinbayer, line 15: I suppose Cypress is not in this list, because we instrument it through a Cypress plugin and not a tracer plugin?

calvinbayer (approved):

integration-tests/ci-visibility/test-optimization-wrong-init/vitest-sum-wrong-init-test.mjs
juan-fernandez: yeah I considered that but since I'm initializing the library through `NODE_OPTIONS: '--import dd-trace/register.js -r dd-trace/init'` and `flushInterval` is exclusively passed programmatically (no env var), there's no way to do this. 

I could init the library programmatically though. I assumed this wasn't supported in ESM but it seems it is: https://docs.datadoghq.com/tracing/trace_collection/automatic_instrumentation/dd_libraries/nodejs/#esm-applications-only-import-the-loader 

> ECMAScript Modules (ESM) applications require an additional command line argument. Add this argument regardless of how the tracer is otherwise imported and initialized:

packages/dd-trace/src/plugin_manager.js
juan-fernandez: yes that's correct

integration-tests/ci-visibility/test-optimization-wrong-init.spec.js
```diff
@@ -0,0 +1,129 @@
+'use strict'
+
+const { exec } = require('child_process')
+
+const { assert } = require('chai')
```
BridgeAR: Could we use Node.js assert instead? :)

We recently decided in the guild to slowly move to those and introducing more chai assertions makes it harder.

packages/datadog-plugin-mocha/test/index.spec.js
```diff
@@ -109,9 +109,13 @@ describe('Plugin', () => {
...
-      return agent.load(['mocha', 'http'], { service: 'test' }, { experimental: { exporter } }).then(() => {
-        Mocha = require(`../../../versions/mocha@${version}`).get()
-      })
+      return agent.load(
+        ['mocha', 'http'],
+        { service: 'test' },
+        { isCiVisibility: true, experimental: { exporter } })
+        .then(() => {
+          Mocha = require(`../../../versions/mocha@${version}`).get()
+        })
```
BridgeAR: Nit: would you mind using async await where possible? :)

integration-tests/ci-visibility/test-optimization-wrong-init/vitest-sum-wrong-init-test.mjs
juan-fernandez: done in https://github.com/DataDog/dd-trace-js/pull/6404/commits/b67d600779b8913152e545842775454456c6e950

packages/datadog-plugin-mocha/test/index.spec.js
juan-fernandez: done in https://github.com/DataDog/dd-trace-js/pull/6404/commits/b67d600779b8913152e545842775454456c6e950

integration-tests/ci-visibility/test-optimization-wrong-init.spec.js
juan-fernandez: done in https://github.com/DataDog/dd-trace-js/pull/6404/commits/b67d600779b8913152e545842775454456c6e950

packages/dd-trace/src/plugin_manager.js
```diff
@@ -6,6 +6,15 @@ const plugins = require('./plugins')
 const log = require('./log')
 const { getEnvironmentVariable } = require('../../dd-trace/src/config-helper')
 
+// Test optimization plugins that should only be enabled when isCiVisibility is true
+const TEST_OPTIMIZATION_PLUGINS = new Set([
```
rochdev, line 10: Why not have this on the plugins instead of having a static list in the plugin manager itself? Ideally it shouldn't know about plugins. Even better, this could be done in `enable()` so that it doesn't even know about the config option at all.

PR #6408, opened 2025-09-10
[test optimization] Clean up handling of known tests in `jest` 

juan-fernandez:
### What does this PR do?

* Simplify logic in handling of known tests in jest workers.
* Add e2e tests for EFD when jest is working in parallel mode.

### Motivation

There were situations in which we considered every test in a test suite new even though the list of known tests received by the jest worker should've been considered invalid.

### Plugin Checklist

- [x] Integration tests.

### DIFF
--- a/integration-tests/ci-visibility/test/efd-parallel/ci-visibility-test-2.js
+++ b/integration-tests/ci-visibility/test/efd-parallel/ci-visibility-test-2.js
@@ -1,10 +1,10 @@
 'use strict'
 
-const { expect } = require('chai')
+const assert = require('node:assert')
 const sum = require('../sum')
 
 describe('ci visibility 2', () => {
   it('can report tests 2', () => {
-    expect(sum(1, 2)).to.equal(3)
+    assert.strictEqual(sum(1, 2), 3)
   })
 })
--- a/integration-tests/jest/jest.spec.js
+++ b/integration-tests/jest/jest.spec.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const { once } = require('node:events')
 const { fork, exec, execSync } = require('child_process')
 const path = require('path')
 const fs = require('fs')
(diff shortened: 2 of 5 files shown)


integration-tests/ci-visibility/test/efd-parallel/ci-visibility-test-2.js
```diff
@@ -0,0 +1,10 @@
+'use strict'
+
+const { expect } = require('chai')
```
BridgeAR: Please only introduce new tests using `node:assert` :)

I am working on changing the current expect calls, so introducing new makes that harder :D

integration-tests/jest/jest.spec.js
```diff
@@ -2404,6 +2404,141 @@ describe('jest CommonJS', () => {
...
+              TESTS_TO_RUN: 'test/efd-parallel/ci-visibility-test',
+              RUN_IN_PARALLEL: true
+            },
+            stdio: 'inherit'
+          }
+        )
+
+        childProcess.on('exit', () => {
+          eventsPromise.then(() => done()).catch(done)
+        })
```
BridgeAR:
```suggestion
        await Promise.all[
          once(childProcess, 'exit'),
          eventsPromise,
        ])
```

Using async await and the event helper to receive a promise makes sure this is not running into unhandled rejections (in case the eventPromise would reject, it would end up as unhandled rejection).

The same applies in the test above.

integration-tests/ci-visibility/test/efd-parallel/ci-visibility-test-2.js
juan-fernandez: sure. Keep in mind that this is a test that's run _within a test_ to test our instrumentation of the test framework 😄 , so I'm not sure it should follow _our_ testing conventions. For example, we might want to test that our instrumentation works with users using `chai` 😄 . In this particular case it doesn't matter because this is just so that this test file runs both in `mocha` and `jest`

PR #6409, opened 2025-09-10
[test optimization] Clean up handling of known tests in `mocha`

juan-fernandez:
### What does this PR do?

* Check that the response at least includes a `mocha` key before considering it as valid. Otherwise, we might be detecting tests as new even though the response is invalid.

### Motivation

Make sure we err on the side of caution (no new test detected)

### Plugin Checklist

- [x] Integration tests.

### DIFF
--- a/integration-tests/mocha/mocha.spec.js
+++ b/integration-tests/mocha/mocha.spec.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const { once } = require('node:events')
 const { fork, exec, execSync } = require('child_process')
 const path = require('path')
 const fs = require('fs')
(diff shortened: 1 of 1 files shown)


integration-tests/mocha/mocha.spec.js
```diff
@@ -1961,6 +1971,65 @@ describe('mocha CommonJS', function () {
...
+            },
+            stdio: 'inherit'
+          }
+        )
+
+        childProcess.on('exit', () => {
+          eventsPromise.then(() => {
+            done()
+          }).catch(done)
+        })
```
BridgeAR: This could cause an unhandled rejection in case eventsPromise would reject before the child process ends. I suggest to use async / await. I believe this is the case in all these PRs.

PR #6411, opened 2025-09-10
[test optimization] Clean up handling of known tests in `vitest`

juan-fernandez:
### What does this PR do?

* Check that the response at least includes a `vitest` key before considering it as valid. Otherwise, we might be detecting tests as new even though the response is invalid.

### Motivation

Make sure we err on the side of caution (no new test detected)

### Plugin Checklist

- [x] Integration tests.

### DIFF
--- a/packages/datadog-instrumentations/src/vitest.js
+++ b/packages/datadog-instrumentations/src/vitest.js
@@ -51,6 +51,10 @@ function waitForHitProbe () {
   })
 }
 
+function isValidKnownTests (receivedKnownTests) {
+  return !!receivedKnownTests.vitest
+}
+
 function getProvidedContext () {
   try {
     const {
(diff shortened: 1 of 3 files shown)


packages/datadog-instrumentations/src/vitest.js
```diff
@@ -235,28 +239,33 @@ function getSortWrapper (sort, frameworkVersion) {
...
-          try {
-            const workspaceProject = this.ctx.getCoreWorkspaceProject()
-            workspaceProject._provided._ddIsKnownTestsEnabled = isKnownTestsEnabled
-            workspaceProject._provided._ddKnownTests = knownTests.vitest || {}
-            workspaceProject._provided._ddIsEarlyFlakeDetectionEnabled = isEarlyFlakeDetectionEnabled
-            workspaceProject._provided._ddEarlyFlakeDetectionNumRetries = earlyFlakeDetectionNumRetries
-          } catch {
-            log.warn('Could not send known tests to workers so Early Flake Detection will not work.')
-          }
+          isEarlyFlakeDetectionFaulty = true
```
calvinbayer, line 267: Why do we set it to faulty if the known tests are invalid?

calvinbayer (approved):

packages/datadog-instrumentations/src/vitest.js
juan-fernandez: yeah I wondered about that myself. The faulty logic was originally designed for when the backend would return a response that would prompt _too many_ new tests. I can't think of a reason why we wouldn't consider this case the same, as _something_ failed (not necessarily the backend) that would cause too many new tests to be detected

PR #6414, opened 2025-09-10
[SVLS-7168] Create tests for GCP PubSub Push Subscriptions Plugin 

nina9753:
### What does this PR do?
Creates New Test File: `pubsub-push-subscription.spec.js` in relation to the changes and new push subscription plugin created in #6260 #6782 #6415
New Tests in `pubsub-push-subscription.spec.js`:

Should create BOTH `pubsub.delivery `span AND HTTP span in same trace
- Verifies dual-span creation (HTTP + Pub/Sub delivery)
- Validates parent-child relationship (HTTP span is parent)
- Confirms active span in handler is `pubsub.delivery` (not HTTP)
- Checks all required metadata on both spans

Should propagate distributed trace context from producer to push delivery
- Tests Datadog trace context propagation via headers
- Verifies span links for cross-trace correlation
- Validates distributed tracing from producer to consumer

Should add batch metadata to delivery span
- Tests batch size, index, and description tags
- Validates batch request trace/span ID propagation
- Confirms batch metrics are set correctly

Should set service name with -pubsub suffix
- Verifies service name override (test → test-pubsub)
- Validates _dd.base_service and _dd.serviceoverride.type tags

Should NOT create pubsub span for non-push-subscription requests
- Confirms plugin only activates for Google `User-Agent`
- Ensures normal HTTP requests are unaffected

Should NOT create pubsub span when missing required headers

- Validates plugin requires `x-goog-pubsub-message-id` header
- Ensures graceful degradation when headers are missing

Updated Test in `index.spec.js` :
- Updated expectSpanWithDefaults helper to handle consumer spans correctly
- Fixed resource name expectations for consumer spans (Message from topic-name)
- Added flushMinSpans: 1 to ensure spans are exported immediately
- Improved test stability and timing

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/datadog-plugin-google-cloud-pubsub/test/pubsub-push-subscription.spec.js
```diff
@@ -0,0 +1,366 @@
...
+            res.writeHead(200)
+            res.end('OK')
+          })
+        } else {
+          res.writeHead(404)
+          res.end()
+        }
+      })
+
+      appListener = server.listen(0, 'localhost', () => {
```
tlhunter, line 142: Do these tests pass locally on your machine? E.g.

```sh
node packages/datadog-plugin-google-cloud-pubsub/test/pubsub-push-subscription.spec.js
```

Sometimes on macOS it can fail binding on localhost and we'll need to use 127.0.0.1.

nina9753: Yes these test passed locally with `PLUGINS=google-cloud-pubsub SPEC=index yarn test:plugins`

PR #6415, opened 2025-09-10
[SVLS-7168] Create inferred Span and Span links for GCP PubSub

nina9753:
### What does this PR do?
This PR completes the distributed tracing story by adding bidirectional span linking between producer and consumer spans. It enables full trace visualization across the entire message lifecycle: producer → Pub/Sub topic → consumer. The implementation includes intelligent span linking that adapts based on whether messages originated from a single trace or multiple traces.

**Consumer Span Linking (Pull Subscriptions)**
Enhanced the consumer plugin to create span links back to the producer's batch span. When a consumer receives a message, it reconstructs the producer's batch span context from the metadata injected by the producer plugin. This context becomes a span link, creating a visible connection in the APM UI between the consumer operation and the batch that produced it.

**Push Subscription Advanced Linking**
Enhanced the push subscription plugin with conditional span linking based on trace ID comparison.
This also follows the industry standard OpenTelemetry concept, representing causal relationships between spans that aren't parent-child.

`Web (framework) Plugin Integration
`Updated the web framework span creation logic to check AsyncLocalStorage for an active pubsub.delivery span before extracting context from HTTP headers. This ensures web framework spans (Express, Fastify, etc.) correctly parented to the delivery span when handling push subscription requests.

**Enhanced Batch Metadata**
All consumer spans (both pull and push) now include rich batch metadata:
- Total messages in the batch
- This message's position (0-indexed)
- Human-readable description
- Producer batch span trace and span IDs
- These become searchable tags in the APM UI

Followup PR with tests can be found here #6414

### Motivation
With the previous PRs we made it so we get proper trace context propagation from producer to consumer, but the relationship between the producer's batch span and each individual consumer span wasn't visible in the APM UI. This PR adds span links that let you:
- See which messages were part of the same batch
- Trace back from a consumer span to the producer's batch span
- Understand message delivery patterns across services
- Debug issues related to message ordering or batching

### Plugin Checklist

- [x] Unit tests.
- [x] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes

Follow-up PR to #6260
Additional information can be found in [this doc](https://docs.google.com/document/d/12yVv7aOtQn04gFMixarDVFXUvdqlD6-ApBBxk1IoEB0/edit?pli=1&tab=t.0)

image

image

### DIFF
--- a/packages/datadog-plugin-google-cloud-pubsub/src/consumer.js
+++ b/packages/datadog-plugin-google-cloud-pubsub/src/consumer.js
@@ -77,7 +77,6 @@ class GoogleCloudPubsubConsumerPlugin extends ConsumerPlugin {
     }
     if (batchRequestSpanId) {
       meta['pubsub.batch.request_span_id'] = batchRequestSpanId
-      // Also add span link metadata
       meta['_dd.pubsub_request.trace_id'] = batchRequestTraceId
       meta['_dd.pubsub_request.span_id'] = batchRequestSpanId
       if (batchRequestTraceId && batchRequestSpanId) {
@@ -103,7 +102,6 @@ class GoogleCloudPubsubConsumerPlugin extends ConsumerPlugin {
       metrics['pubsub.batch.index'] = Number.parseInt(batchIndex, 10)
     }
 
-    // Add batch description
     if (batchSize && batchIndex !== undefined) {
       const index = Number.parseInt(batchIndex, 10)
       const size = Number.parseInt(batchSize, 10)
--- a/packages/datadog-plugin-google-cloud-pubsub/test/index.spec.js
+++ b/packages/datadog-plugin-google-cloud-pubsub/test/index.spec.js
@@ -282,20 +282,14 @@ describe('Plugin', () => {
             },
             rawExpectedSchema.receive,
             {
-              // Custom selectSpan: look through all traces for a consumer span
-              // (withNamingSchema will check the name matches expected opName)
               selectSpan: (traces) => {
-                // Flatten all spans from all traces
                 for (const trace of traces) {
                   for (const span of trace) {
-                    // Return the first worker-type span (consumer span)
                     if (span.type === 'worker') {
                       return span
                     }
                   }
                 }
-                // If no worker span found, return undefined to trigger retry
-                // (withNamingSchema's assertSomeTraces will keep waiting)
                 return undefined
               }
             }
@@ -431,7 +425,6 @@ describe('Plugin', () => {
             await consume(async () => {
               agent.expectPipelineStats(dsmStats => {
                 let statsPointsReceived = 0
-                // we should have 2 dsm stats points
                 dsmStats.forEach((timeStatsBucket) => {
                   if (timeStatsBucket && timeStatsBucket.Stats) {
                     timeStatsBucket.Stats.forEach((statsBuckets) => {
@@ -453,19 +446,15 @@ describe('Plugin', () => {
         const spanKind = expected.meta?.['span.kind']
 
         if (method === 'publish') {
-          // For publish operations, use the new format: "publish to Topic <topic-name>"
           prefixedResource = `${method} to Topic ${topicName}`
         } else if (spanKind === 'consumer') {
-          // For consumer operations, use the new format: "Message from <topic-name>"
           prefixedResource = `Message from ${topicName}`
         } else if (method) {
-          // For other operations, use the old format: "<method> <full-resource-path>"
           prefixedResource = `${method} ${resource}`
         } else {
           prefixedResource = resource
         }
 
-        // Determine the default operation name based on span kind
         let defaultOpName = 'pubsub.receive'
         if (spanKind === 'consumer') {
           defaultOpName = expectedSchema.receive.opName
--- a/packages/datadog-plugin-google-cloud-pubsub/test/pubsub-push-subscription.spec.js
+++ b/packages/datadog-plugin-google-cloud-pubsub/test/pubsub-push-subscription.spec.js
@@ -1,0 +1,366 @@
+'use strict'
+
+const axios = require('axios')
+const { expect } = require('chai')
+const { describe, it, beforeEach, afterEach, before, after } = require('mocha')
+const agent = require('../../dd-trace/test/plugins/agent')
+
+describe('Push Subscription Plugin', () => {
+  let tracer
+  let appListener
+
+  before(() => {
+    return agent.load(['http', 'google-cloud-pubsub'], { client: false })
+  })
+
+  after(() => {
+    return agent.close({ ritmReset: false })
+  })
+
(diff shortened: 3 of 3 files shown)


packages/datadog-instrumentations/src/google-cloud-pubsub.js
```diff
@@ -6,15 +6,13 @@ const {
...
 const shimmer = require('../../datadog-shimmer')
 const { storage } = require('../../datadog-core')
-const log = require('../../dd-trace/src/log')
 
 // Auto-load push subscription plugin to enable pubsub.delivery spans for push subscriptions
 try {
   const PushSubscriptionPlugin = require('../../datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription')
   new PushSubscriptionPlugin(null, {}).configure({})
-} catch (e) {
+} catch {
```
tlhunter: Was this removed accidentally?

packages/datadog-instrumentations/src/google-cloud-pubsub.js
tlhunter, line 34: Maps like these which correlate to user-operations (such as HTTP requests) are scary because they can cause memory leaks whenever we don't clean up properly.

For that reason we try to use a WeakMap instead. Those hold a "weak reference" to the key and cause the value to be deleted automatically when the key is garbage collected.

However it only works when the key is an object and your Map is using a string as a key. We could theoretically do something where the request is the key but I don't think that is available later when the delete happens?

Can you explain more about the lifecycle of these ack IDs? Maybe we can figure out something that way.

tlhunter:
```suggestion
  if (Message?.prototype?.ack) {
```

packages/datadog-instrumentations/src/google-cloud-pubsub.js
```diff
@@ -270,19 +271,19 @@ function injectTraceContext (attributes, pubsub, topicName) {
...
+        }
+        return publishMessage.apply(this, arguments)
       }
-      return publishMessage.apply(this, arguments)
     })
   }
 
-  // Wrap Topic.publish (legacy API)
-  if (obj.Topic.prototype.publish) {
+  if (typeof obj.Topic.prototype.publish === 'function') {
```
tlhunter, line 330: Did you find that sometimes `obj.Topic.prototype.publish` was a truthy value but wasn't a function?

packages/datadog-plugin-google-cloud-pubsub/src/consumer.js
```diff
@@ -2,30 +2,150 @@
...
+    if (batchSize && batchIndex !== undefined) {
+      const index = Number.parseInt(batchIndex, 10)
+      const size = Number.parseInt(batchSize, 10)
+      meta['pubsub.batch.description'] = `Message ${index + 1} of ${size}`
+    }
 
     const span = this.startSpan({
       childOf,
-      resource: topic,
+      resource: `Message from ${topicName}`,
```
tlhunter, line 113: Like in the previous PR this one also looks like a breaking change.

packages/datadog-plugin-google-cloud-pubsub/src/producer.js
tlhunter, line 119: https://github.com/DataDog/dd-trace-js/pull/6782/files#diff-32e83f63e1f3fd1938530fecdf21e77e942496aee9272d001accbb758a7fc935R119

I'm getting a little confused. Do these two PRs repeat functionality?

packages/datadog-instrumentations/src/google-cloud-pubsub.js
nina9753: The lifecycle has a critical gap that prevents WeakMap usage:
```
// 1. MESSAGE RECEIVED: User gets a Message object
subscription.on('message', (message) => {
  // message.ackId = "abc123" (a string property)
  
  // 2. USER ACKS: We intercept message.ack() and store context
  message.ack()
  // → ackContextMap.set("abc123", context)  // Key is STRING, not object
  
  // 3. MESSAGE OBJECT GOES OUT OF SCOPE: User code finishes
  // The Message object may be GC'd here
  
  // 4. BATCHED ACKNOWLEDGE CALL: SDK batches acks into gRPC call (seconds later)
  pubsubClient.acknowledge({ ackIds: ["abc123", "def456", "ghi789"] })
  // → We only have STRING ackIds, NOT the original Message objects
  // → Need to look up context by string: ackContextMap.get("abc123")
})
```
Between steps 2 and 4, the Message object is gone, but we need to keep the context alive and retrieve it using only the string ackId. The acknowledge gRPC call creates a single `pubsub.request` span for the batch, and we need to ensure that span runs in the right async context. Even though N messages were acked, there's only 1 acknowledge API call. So, because the Message Object is not available during retrieval, we must use a Map. To prevent memory leaks, I added a TTL to handle manual cleanup, a graceful shutdown, and cases where acknowledgment never happens (network failure, crash, etc.).

packages/datadog-plugin-google-cloud-pubsub/src/consumer.js
nina9753: There shouldn't be any users for this currently, but I can always revert to what is in prod. I wanted to make the spans more descriptive for the blog post and product release.

packages/datadog-instrumentations/src/google-cloud-pubsub.js
nina9753: No, I have not found evidence that `obj.Topic.prototype.publish` is ever a truthy non-function value in practice.
The `typeof obj.Topic.prototype.publish === 'function'` check is following the pattern that's used throughout the dd-trace-js codebase for instrumentation you can see other examples in the same file that i did not add myself

packages/datadog-instrumentations/src/google-cloud-pubsub.js
```diff
@@ -30,7 +30,25 @@ const receiveStartCh = channel('apm:google-cloud-pubsub:receive:start')
...
+    if (now - entry.timestamp > ACK_CONTEXT_TTL_MS) {
+      ackContextMap.delete(ackId)
+    }
+  }
+}, 60_000) // Run cleanup every 60 seconds
+
+// Allow process to exit cleanly
+if (ackContextCleanupInterval.unref) {
+  ackContextCleanupInterval.unref()
+}
```
tlhunter, line 51: `.unref()` will always be present (and it returns the timer instance which is nice for method chaining). But I'm pretty concerned about using a timer like this for cleanup since these types of solutions can often cause memory leaks. E.g. with a 60 second timer there could be a saw tooth memory increase with a 60 second period.

```suggestion
}, 60_000).unref() // Run cleanup every 60 seconds
```

PR #6431, opened 2025-09-12
feat(anthropic, llmobs): add support for `@ai-anthropic/sdk`

sabrenner:
### What does this PR do?
Adds support for the `@anthropic-ai/sdk` package. Specifically, the `messages.create` method (both streamed and non-streamed responses), which is also used by `messages.stream` under the hood.

For APM spans, spans only have the `anthropic.request.model` tag.

For LLM Observability, all input and output, as well as metadata, tool calls, tool responses, and token usages are captured.

LLM interactions through `messages.stream` will still show up as `Messages.create`, as that is what's called under the hood, but the response chunks are coalesced the same.

### Motivation
Support for Anthropic, feature parity with our LLM Observability Python auto-instrumentation

### Testing

Shared tests written for the anthropic integration are passing
```
$ TEST_LIBS=nodejs pytest test_anthropic.py
========================================================================================================================================== test session starts ===========================================================================================================================================
platform darwin -- Python 3.13.0, pytest-8.3.3, pluggy-1.5.0 -- **
cachedir: .pytest_cache
sdks: nodejs
rootdir: **
configfile: .pytest.ini
collected 20 items                                                                                                                                                                                                                                                                                       

test_anthropic.py::TestAnthropicApm::test_create[nodejs-test-ml-app-tcp-True] PASSED                                                                                                                                                                                                               [  5%]
test_anthropic.py::TestAnthropicApm::test_create[nodejs-test-ml-app-tcp-False] PASSED                                                                                                                                                                                                              [ 10%]
test_anthropic.py::TestAnthropicApm::test_create_stream_method[nodejs-test-ml-app-tcp] PASSED                                                                                                                                                                                                      [ 15%]
test_anthropic.py::TestAnthropicLlmObs::test_create[nodejs-test-ml-app-tcp-True] PASSED                                                                                                                                                                                                            [ 20%]
test_anthropic.py::TestAnthropicLlmObs::test_create[nodejs-test-ml-app-tcp-False] PASSED                                                                                                                                                                                                           [ 25%]
test_anthropic.py::TestAnthropicLlmObs::test_create_stream_method[nodejs-test-ml-app-tcp] PASSED                                                                                                                                                                                                   [ 30%]
test_anthropic.py::TestAnthropicLlmObs::test_create_content_block[nodejs-test-ml-app-tcp-True] PASSED                                                                                                                                                                                              [ 35%]
test_anthropic.py::TestAnthropicLlmObs::test_create_content_block[nodejs-test-ml-app-tcp-False] PASSED                                                                                                                                                                                             [ 40%]
test_anthropic.py::TestAnthropicLlmObs::test_create_error[nodejs-test-ml-app-tcp-True] PASSED                                                                                                                                                                                                      [ 45%]
test_anthropic.py::TestAnthropicLlmObs::test_create_error[nodejs-test-ml-app-tcp-False] PASSED                                                                                                                                                                                                     [ 50%]
test_anthropic.py::TestAnthropicLlmObs::test_create_multiple_system_prompts[nodejs-test-ml-app-tcp-True] PASSED                                                                                                                                                                                    [ 55%]
test_anthropic.py::TestAnthropicLlmObs::test_create_multiple_system_prompts[nodejs-test-ml-app-tcp-False] PASSED                                                                                                                                                                                   [ 60%]
test_anthropic.py::TestAnthropicLlmObs::test_create_with_tools[nodejs-test-ml-app-tcp-True] PASSED                                                                                                                                                                                                 [ 65%]
test_anthropic.py::TestAnthropicLlmObs::test_create_with_tools[nodejs-test-ml-app-tcp-False] PASSED                                                                                                                                                                                                [ 70%]
test_anthropic.py::TestAnthropicLlmObs::test_create_tool_result[nodejs-test-ml-app-tcp-True] PASSED                                                                                                                                                                                                [ 75%]
test_anthropic.py::TestAnthropicLlmObs::test_create_tool_result[nodejs-test-ml-app-tcp-False] PASSED                                                                                                                                                                                               [ 80%]
test_anthropic.py::TestAnthropicLlmObs::test_create_redact_image_input[nodejs-test-ml-app-tcp-True] PASSED                                                                                                                                                                                         [ 85%]
test_anthropic.py::TestAnthropicLlmObs::test_create_redact_image_input[nodejs-test-ml-app-tcp-False] PASSED                                                                                                                                                                                        [ 90%]
test_anthropic.py::TestAnthropicLlmObs::test_create_prompt_caching[nodejs-test-ml-app-tcp-True] PASSED                                                                                                                                                                                             [ 95%]
test_anthropic.py::TestAnthropicLlmObs::test_create_prompt_caching[nodejs-test-ml-app-tcp-False] PASSED   
```

MLOB-3457

### DIFF
--- a/packages/dd-trace/src/llmobs/plugins/anthropic.js
+++ b/packages/dd-trace/src/llmobs/plugins/anthropic.js
@@ -19,8 +19,9 @@ class AnthropicLLMObsPlugin extends LLMObsPlugin {
     super(...arguments)
 
     this.addSub('apm:anthropic:request:chunk', ({ ctx, chunk, done }) => {
-      const chunks = ctx.chunks ?? (ctx.chunks = [])
-      if (chunk) ctx.chunks.push(chunk)
+      ctx.chunks ??= []
+      const chunks = ctx.chunks
+      if (chunk) chunks.push(chunk)
 
       if (!done) return
 


initialize.mjs
```diff
@@ -36,7 +36,13 @@ ${result.source}`
 const [NODE_MAJOR, NODE_MINOR] = process.versions.node.split('.').map(Number)
 
 const brokenLoaders = NODE_MAJOR === 18 && NODE_MINOR === 0
-const iitmExclusions = [/langsmith/, /openai\/_shims/, /openai\/resources\/chat\/completions\/messages/, /openai\/agents-core\/dist\/shims/]
+const iitmExclusions = [
+  /langsmith/,
+  /openai\/_shims/,
+  /openai\/resources\/chat\/completions\/messages/,
+  /openai\/agents-core\/dist\/shims/,
+  /@anthropic-ai\/sdk\/_shims/
```
Yun-Kim, line 44: dumb question - what is the purpose of this const?

packages/datadog-instrumentations/src/anthropic.js
```diff
@@ -0,0 +1,115 @@
...
+    if (!anthropicTracingChannel.start.hasSubscribers) {
+      return create.apply(this, arguments)
+    }
+
+    const options = arguments[0]
+    const stream = options.stream
+
+    const ctx = { options, resource: 'create' }
+
+    return anthropicTracingChannel.start.runStores(ctx, () => {
```
Yun-Kim, line 46: just to confirm my understanding - this is where we publish the signal to the anthropicTracingChannel to start the span, correct?

packages/datadog-plugin-anthropic/src/tracing.js
```diff
@@ -0,0 +1,30 @@
...
+
+const TracingPlugin = require('../../dd-trace/src/plugins/tracing')
+
+class AnthropicTracingPlugin extends TracingPlugin {
+  static id = 'anthropic'
+  static operation = 'request'
+  static system = 'anthropic'
+  static prefix = 'tracing:apm:anthropic:request'
+
+  bindStart (ctx) {
```
Yun-Kim, line 11: Just for my understanding where does `bindStart` get called?

packages/dd-trace/src/llmobs/plugins/anthropic.js
```diff
@@ -0,0 +1,281 @@
...
+            if (finishReason) response.finish_reason = finishReason
+
+            const { usage } = chunk
+            if (usage) {
+              const responseUsage = response.usage ?? (response.usage = { input_tokens: 0, output_tokens: 0 })
+              responseUsage.output_tokens = usage.output_tokens
+
+              const cacheCreationTokens = usage.cache_creation_input_tokens
+              const cacheReadTokens = usage.cache_read_input_tokens
+              if (cacheCreationTokens) responseUsage.cache_creation_input_tokens = cacheCreationTokens
```
Yun-Kim, line 87: just a note we standardized calling these `cache_write_input_tokens` in our backend/other integrations, not sure if you handle that later down the line

packages/dd-trace/src/llmobs/plugins/anthropic.js
```diff
@@ -0,0 +1,281 @@
...
+    }
+  }
+
+  setLLMObsTags (ctx) {
+    const span = ctx.currentStore?.span
+    if (!span) return
+
+    const { options, result } = ctx
+
+    this.#tagAnthropicInputMessages(span, options)
```
Yun-Kim, line 127: I'm assuming this is standard practice (ensuring these are private methods only called internally?)

packages/dd-trace/src/llmobs/plugins/anthropic.js
```diff
@@ -0,0 +1,281 @@
...
+          let input = block.input
+          if (typeof input === 'string') {
+            input = JSON.parse(input)
+          }
+
+          const toolCall = {
+            name,
+            arguments: input,
+            toolId: id,
+            type
```
Yun-Kim, line 165: what value do we send as `type`?

packages/dd-trace/src/llmobs/plugins/anthropic.js
```diff
@@ -0,0 +1,281 @@
...
+      (inputTokens ?? 0) +
+      (cacheWriteTokens ?? 0) +
+      (cacheReadTokens ?? 0)
+
+    if (outputTokens) metrics.outputTokens = outputTokens
+    const totalTokens = metrics.inputTokens + (outputTokens ?? 0)
+    if (totalTokens) metrics.totalTokens = totalTokens
+
+    if (cacheWriteTokens != null) metrics.cacheWriteTokens = cacheWriteTokens
+    if (cacheReadTokens != null) metrics.cacheReadTokens = cacheReadTokens
```
Yun-Kim, line 257: Just making sure - are we tagging span metrics as `cache_write_input_tokens`? I think our backend span processors depend on these values

initialize.mjs
sabrenner: [import-in-the-middle](https://github.com/nodejs/import-in-the-middle), our dependency for intercepting imports using [ESM](https://nodejs.org/api/esm.html) (tl;dr files where you `import` vs `require`, and some other things), generally intercepts _everything_, and has proven problematic for certain modules or sub-modules. in this case, `@anthropic-ai/sdk/_shims` is one of those modules where, if intercepted by import-in-the-middle, will unintentionally break it, so we just ignore it here. luckily the rest of the `@anthropic-ai/sdk` package is unaffected!

packages/datadog-instrumentations/src/anthropic.js
sabrenner: yeah - `start.runStores` will publish the start event, and run any `bindings` to bind the given context object to the current async frame through [AsyncLocalStorage](https://nodejs.org/api/async_context.html). it's listened to implicitly in the `bindStart` method for the anthropic tracing plugin in the `datadog-plugin-anthropic/tracing.js` file

packages/datadog-plugin-anthropic/src/tracing.js
sabrenner: it's a bit around-ish in how it's handled but the base `TracingPlugin` class will use this method to subscribe to the `start` method for the `tracing:apm:anthropic:request:start` event, specifically the store bindings to make sure the context is entered into the async frame (https://github.com/DataDog/dd-trace-js/pull/6431#discussion_r2392416763)

packages/dd-trace/src/llmobs/plugins/anthropic.js
sabrenner: yep! the `#` prefix is how we're going forward with privatizing methods/members of js classes (are not callable and do not show up on object prototypes, only available within the class itself).

packages/dd-trace/src/llmobs/plugins/anthropic.js
sabrenner: yep! it's coalesced in the `#tagAnthropicUsage` method, and then further in the `LLMObsTagger.tagMetrics` function

```javascript
case 'cacheReadTokens':
  processedKey = CACHE_READ_INPUT_TOKENS_METRIC_KEY
  break
case 'cacheWriteTokens':
  processedKey = CACHE_WRITE_INPUT_TOKENS_METRIC_KEY
  break
```

where

```javascript
CACHE_READ_INPUT_TOKENS_METRIC_KEY: 'cache_read_input_tokens',
CACHE_WRITE_INPUT_TOKENS_METRIC_KEY: 'cache_write_input_tokens',
```

sabrenner: the shared tests we have verify this behavior as well, which this passes 😄

packages/dd-trace/src/llmobs/plugins/anthropic.js
sabrenner: yeah - see https://github.com/DataDog/dd-trace-js/pull/6431#discussion_r2392428341, i just have the camel-case as a js convenience (for us, and for users if/when we document it in our type definitions) but it gets re-mapped in the llmobs tagger

packages/dd-trace/src/llmobs/plugins/anthropic.js
sabrenner: hmm good question, it's whatever was on the block itself - i think for anthropic, something like `tool_use` from looking at some mock responses

packages/datadog-instrumentations/src/anthropic.js
Yun-Kim: gotcha - thanks for confirming!!

packages/dd-trace/src/llmobs/plugins/anthropic.js
Yun-Kim: i did not know you could implicitly set kv pairs like this lol, gotcha

Yun-Kim (approved):

packages/datadog-instrumentations/src/anthropic.js
```diff
@@ -0,0 +1,115 @@
...
+  ctx.result ??= result
+
+  anthropicTracingChannel.asyncEnd.publish(ctx)
+}
+
+const extensions = ['js', 'mjs']
+for (const extension of extensions) {
+  addHook({
+    name: '@anthropic-ai/sdk',
+    file: `resources/messages.${extension}`,
```
tlhunter, line 94: Interesting... I don't think we've had to do this before. Is anthropic unique and ships both .js and .mjs files next to each other?

sabrenner: yeah they do, which is weird, but i have had to do it for some others as well

- [openai](https://github.com/DataDog/dd-trace-js/blob/b6077e7b3341de45cc9418ed2e30391f01175502/packages/datadog-instrumentations/src/openai.js#L201-L203)
- [langchain](https://github.com/DataDog/dd-trace-js/blob/b6077e7b3341de45cc9418ed2e30391f01175502/packages/datadog-instrumentations/src/langchain.js#L27-L29)
- [ai](https://github.com/DataDog/dd-trace-js/blob/b6077e7b3341de45cc9418ed2e30391f01175502/packages/datadog-instrumentations/src/ai.js#L128-L151) (vercel's ai sdk)

i thinnnkk because these libraries are written in TS and just transpile to both, with one or the other being the default? but anyhow, also verifiable by looking at the [installed code via npm](https://www.npmjs.com/package/@anthropic-ai/sdk?activeTab=code) (under `/resources`)

tlhunter: @bengl you might find this interesting. Not sure if it's something worth building an abstraction for since it's not that common.

packages/dd-trace/src/llmobs/plugins/anthropic.js
```diff
@@ -0,0 +1,281 @@
...
+class AnthropicLLMObsPlugin extends LLMObsPlugin {
+  static integration = 'anthropic' // used for llmobs telemetry
+  static id = 'anthropic'
+  static prefix = 'tracing:apm:anthropic:request'
+
+  constructor () {
+    super(...arguments)
+
+    this.addSub('apm:anthropic:request:chunk', ({ ctx, chunk, done }) => {
+      const chunks = ctx.chunks ?? (ctx.chunks = [])
```
tlhunter: I had to read this line a few times and bust out a REPL to confirm understanding.

This isn't as terse but it's more understandable IMO:

```suggestion
      ctx.chunks = ctx.chunks || []
      const chunks = ctx.chunks
```

packages/dd-trace/src/llmobs/plugins/anthropic.js
```diff
@@ -0,0 +1,281 @@
...
+      if (chunk) ctx.chunks.push(chunk)
```
tlhunter:
```suggestion
      if (chunk) chunks.push(chunk)
```

packages/dd-trace/src/llmobs/plugins/anthropic.js
sabrenner: ooh good callout, will simplify this!

tlhunter (approved):

PR #6432, opened 2025-09-15
feat(aws): add partition tag

seungleelee:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-plugin-aws-sdk/src/base.js
+++ b/packages/datadog-plugin-aws-sdk/src/base.js
@@ -115,6 +115,11 @@ class BaseAwsSdkPlugin extends ClientPlugin {
       span.setTag('aws.region', region)
       span.setTag('region', region)
 
+      const partition = getPartition(region)
+      if (partition) {
+        span.setTag('partition', partition)
+      }
+
       if (!this._tracerConfig?._isInServerlessEnvironment()) return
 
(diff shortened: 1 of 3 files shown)


packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -63,6 +63,10 @@ class BaseAwsSdkPlugin extends ClientPlugin {
         'aws.service': awsService,
         component: 'aws-sdk'
       }
+      const partition = getPartition(awsRegion)
+      if (partition) {
+        meta.partition = partition
```
BridgeAR: Nit: it's possible to get the partition before defining meta. That way the partition can be assigned right away and the internal object shape does not need to change anymore.

BridgeAR (commented):
The code is LGTM, I just believe a test is missing for one of the cases. Would it also be possible to add a test case for a non-default region?

packages/datadog-plugin-aws-sdk/src/base.js
seungleelee: updated - also, added a test case to cover all different partition types

PR #6435, opened 2025-09-15
docs: fix default agent hostname in exposed types

watson:
The default agent hostname is the loopback IP address `127.0.0.1`. Not the `localhost` hostname. Updated the exposed types in the `index.d.ts` file to match.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -427,7 +427,7 @@ declare namespace tracer {
 
     /**
      * The address of the trace agent that the tracer will submit to.
-     * @default 'localhost'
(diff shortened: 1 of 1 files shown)


simon-id (approved):
approved because it's indeed the default in config.js, but localhost would be a better default

watson:
> maybe the default should be localhost instead, i usually prefer it

Same here, but if I recall correctly we use `127.0.0.1` because of some specific issue we discovered at some point where `localhost` would fail under certain scenarios. I think @rochdev might be able to fill us in

PR #6436, opened 2025-09-15
fix: ensure default agent hostname is 127.0.0.1 everywhere

watson:
Align the default agent hostname so that it's always `127.0.0.1` instead of `localhost`. The config default was already `127.0.0.1`, but in a few places we would still use `localhost`.

To ease maintenance in the future, instead of hardcoding the default agent hostname/port in each place it's needed, move all the default config values to a shared file that can be required across the code base.


szegedi (approved):

packages/dd-trace/test/ci-visibility/exporters/agent-proxy/agent-proxy.spec.js
```diff
@@ -20,7 +20,7 @@ describe('AgentProxyCiVisibilityExporter', () => {
   const tags = {}
 
   it('should query /info right when it is instantiated', async () => {
-    const scope = nock('http://localhost:8126')
+    const scope = nock('http://127.0.0.1:8126')
```
tylfin, line 23: nit: these were probably found + replaced, but it's probably worth replacing all the strings with a single const

watson: Somehow my local branch got messed up and I've spend 30 minutes trying to restore it now. To speed things up, I'm just going to merge this PR as is, but will clean this up in a follow up PR.

watson: https://github.com/DataDog/dd-trace-js/pull/6462

PR #6437, opened 2025-09-15
[test optimization] Improve logic for base branch detection

juan-fernandez:
### What does this PR do?
Make it so that `checkAndFetchBranch` is not called for multiple possible base branches if one of them is already found.

### Motivation
We don't need to constantly look for base branches if we have already found one.

### Plugin Checklist

- [ ] Unit tests.


juan-fernandez:
it turns out the logic was correct

PR #6438, opened 2025-09-15
test: update prisma test version and fix esm test

BridgeAR:
The ESM tests will now use the regular lower and upper bound as defined in the withVersions range. The former implementation installed prisma separately, which was a workaround to execute the tests. Now, the DB is purged before the tests are run on each run.


packages/datadog-plugin-prisma/test/integration-test/client.spec.js
```diff
@@ -26,46 +28,58 @@ describe('esm', () => {
...
 
-    it('is instrumented', async () => {
+    it('is instrumented', async function () {
+      this.timeout(20000)
       const res = agent.assertMessageReceived(({ headers, payload }) => {
-        assert.propertyVal(headers, 'host', `127.0.0.1:${agent.port}`)
-        assert.isArray(payload)
-        assert.strictEqual(checkSpansForServiceName(payload, 'prisma.client'), true)
-        assertObjectContains(payload[0][0], {
+        assert.strictEqual(headers?.host, `127.0.0.1:${agent.port}`)
```
watson: I prefer writing assertions like this:

```suggestion
        assert.strictEqual(headers.host, `127.0.0.1:${agent.port}`)
```

With the `?`, it would fail with "undefined not equal to...", which could lead you to think that `headers.host` was `undefined` when in fact `headers` was `undefined`. Not having the `?` means it would fail with a much more readable error ("trying to access host on undefined")

tlhunter (dismissed):

packages/datadog-plugin-prisma/test/integration-test/client.spec.js
BridgeAR: Good point. That was actually just a type error showing up. I wonder how to handle that ideally.

tlhunter (approved):

PR #6443, opened 2025-09-15
refactor: only compute the agent url once

watson:
Instead of computing the agent url every time we need it, generate it once in the main config and re-use that throughout the code base.

This PR also changes the `config.url` property to be a string instead of an instance of `URL`.


BridgeAR:
This seems to be superseded by https://github.com/DataDog/dd-trace-js/pull/7359

PR #6446, opened 2025-09-16
test,chore: use less globals and improve linting

BridgeAR:
The sinon, expect, and proxyquire globals were removed. This also fixes a couple more issues around the linter when run locally.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -538,19 +538,12 @@ export default [
   {
     name: 'dd-trace/tests/all',
     files: TEST_FILES,
-    languageOptions: {
-      globals: {
-        sinon: 'readonly',
-        expect: 'readonly',
-        proxyquire: 'readonly',
-      }
-    },
     plugins: {
       mocha: eslintPluginMocha,
       n: eslintPluginN
     },
     rules: {
-      'mocha/max-top-level-suites': 'off',
+      'mocha/max-top-level-suites': ['error', { limit: 1 }],
       'mocha/no-exports': 'off',
       'mocha/no-global-tests': 'off',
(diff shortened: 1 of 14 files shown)


eslint.config.mjs
```diff
@@ -564,6 +557,26 @@ export default [
       'require-await': 'off'
     }
   },
+  {
+    name: 'dd-trace/tests/all',
```
watson: It's best not to have two groups with the same name. Please rename this one to something that signals which files it's covering.

eslint.config.mjs
```diff
@@ -573,7 +586,11 @@ export default [
       'integration-tests/**/*.js',
       'integration-tests/**/*.mjs',
       'packages/*/test/integration-test/**/*.js',
-      'packages/*/test/integration-test/**/*.mjs'
+      'packages/*/test/integration-test/**/*.mjs',
+      // TODO: Move the files in esm-test to integration-test
+      'packages/datadog-plugin-graphql/test/esm-test/**/*.mjs',
+      'packages/dd-trace/test/appsec/**/*.js',
+      'packages/datadog-plugin-jest/test/jest-test.js',
```
watson: This ESLint group is only supposed to contain integration tests. If you want to include these, I think the name should be changed to match which files it contains.

simon-id (commented):
LGTM except for eslint config which i don't know how it works and Watson has asked for changes

eslint.config.mjs
BridgeAR: I understand that. All test-optimization tests are actually just somewhat integration tests.
I renamed it and added another TODO entry that some files should be renamed so that it's clear that it's an integration test.

PR #6447, opened 2025-09-16
chore: add basic tsconfig to check for type errors

BridgeAR:


.github/workflows/project.yml
```diff
@@ -86,8 +86,8 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
       - uses: ./.github/actions/node/latest
       - uses: ./.github/actions/install
-      - run: yarn type:test
-      - run: yarn type:doc
+      - run: npm run type:test
+      - run: npm run type:doc
```
bengl, line 90: Is this necessary? It's inconsistent w/ the rest of our workflows. Maybe do this in a separate PR?

BridgeAR: It is not necessary. I just stumbled over it and we spoke about changing that as well. I understand that it's nicer to separate it out, while I thought we want to change this anyway, so it should not hurt either?

simon-id (commented):
ugh having to pass an argument to promise `resolve()` everywhere just to please typescript does not shine happiness. Is there a way to disable that in the conf or with some ts magic that i'm too untyped to know about ?

BridgeAR:
@simon-id I agree that this is not ideal. Sadly it's a current limitation for which there is neither an option or other possibility to deactivate the rule for that. It is known by the TS team that this behavior is indeed not ideal and they are open for an RFC to change the current behavior, while I guess even if we'd put something together that would be accepted, it won't help us anytime soon.

benchmark/sirun/run-util.js
```diff
@@ -10,7 +10,7 @@ function exec (...args) {
     proc.on('error', reject)
     proc.on('exit', (code) => {
       if (code === 0) {
-        resolve()
+        resolve(undefined)
```
watson: I agree with @simon-id that this isn't nice. In pure TypeScript you'd type your promise like so (notice the `<void>` generic):

```js
new Promise<void>((resolve) => {
  resolve()
})
```

With JSDocs, as we're using, I think you can do the same with:

```js
/** @type {Promise<void>} */
new Promise((resolve) => {
  resolve()
})
```

I would absolutely prefer that instead of calling `resolve` with `undefined`.

packages/dd-trace/src/pkg.js
```diff
@@ -20,11 +20,9 @@ function findPkg () {
...
   const filePath = findUp('package.json', root, directory)
 
-  try {
-    return JSON.parse(fs.readFileSync(filePath, 'utf8'))
-  } catch {
-    return {}
-  }
+  if (filePath === undefined) return {}
+
+  return require(filePath)
```
watson: This might throw, which is a regression compared to the original implementation.

packages/dd-trace/src/rate_limiter.js
```diff
@@ -3,14 +3,25 @@
...
 
 class RateLimiter {
+  /**
+   * @param {number} rateLimit - Allowed units per interval. Negative means unlimited, 0 disables.
+   * @param {'second'|'minute'|'hour'|'day'} [interval='second'] - Time window for the limiter.
+   */
   constructor (rateLimit, interval = 'second') {
-    this._rateLimit = Number.parseInt(rateLimit)
+    this._rateLimit = Number.parseInt(String(rateLimit))
+    // The limiter constructor accepts a token count number and an interval string
```
watson, line 12: Shouldn't this comment be part of the JSDoc instead?

packages/dd-trace/src/rate_limiter.js
```diff
@@ -3,14 +3,25 @@
 const limiter = require('limiter')
 
 class RateLimiter {
+  /**
+   * @param {number} rateLimit - Allowed units per interval. Negative means unlimited, 0 disables.
+   * @param {'second'|'minute'|'hour'|'day'} [interval='second'] - Time window for the limiter.
+   */
   constructor (rateLimit, interval = 'second') {
-    this._rateLimit = Number.parseInt(rateLimit)
+    this._rateLimit = Number.parseInt(String(rateLimit))
```
watson, line 11: With these types, are we not guaranteed that `rateLimit` is a number? At a minimum, I would prefer a type check like this:


```suggestion
    this._rateLimit = typeof rateLimit === 'number' ? rateLimit : Number.parseInt(String(rateLimit))
```

packages/dd-trace/test/plugins/agent.js
```diff
@@ -122,9 +150,10 @@ function addEnvironmentVariablesToHeaders (headers) {
 
   // add the DD environment variables to the header if any exist
   // to send with trace to final agent destination
-  if (ddEnvVars.length > 0) {
-    headers['X-Datadog-Trace-Env-Variables'] = ddEnvVars.join(',')
-  }
+  // if (ddEnvVars.size > 0) {
+  //   // TODO: Should we still do this? It has never worked until now.
+  //   headers['X-Datadog-Trace-Env-Variables'] = [...ddEnvVars].map(([key, value]) => `${key}=${value}`).join(',')
+  // }
```
watson, line 156: Did you mean to comment this out?

packages/dd-trace/test/plugins/agent.js
```diff
@@ -315,12 +374,30 @@ module.exports = {
...
    * @param {Record<string, unknown>} [config]
    * @param {Record<string, unknown>} [tracerConfig={}]
    * @returns Promise<void>
    */
-  async load (pluginName, config, tracerConfig = {}) {
+  /**
+   * Load the plugin on the tracer with an optional config and start a mock agent.
+   *
+   * @overload
+   * @param {String[]} pluginNames - Name or list of names of plugins to load
```
watson, line 382: I'm not sure I'm fully onboard with saying one thing in the types (expects an array), and then explicitly supporting non-arrays as well in the function body 🤔

I do understand that you want to encourage usage the array-type argument, but it still feels weird, and I would prefer this PR fixing all calls to `load()` to use arrays and then remove the fallback in the function-body.

The same goes for the `config` arg.

packages/dd-trace/test/plugins/agent.js
```diff
@@ -536,6 +606,11 @@ module.exports = {
 
   /**
    * Stop the mock agent, reset all expectations and wipe the require cache.
+   *
+   * Defaults:
+   * - ritmReset: true
+   * - wipe: false
```
watson, line 607: Why this addition? Shouldn't this be clear from the types below?

packages/dd-trace/test/profiling/exporters/agent.spec.js
```diff
@@ -207,7 +207,7 @@ describe('exporters/agent', function () {
         app.post('/profiling/v1/input', upload.any(), (req, res) => {
           try {
             verifyRequest(req, profiles, start, end)
-            resolve()
+            resolve(true)
```
watson: Why `true`? The result of the promise isn't used anywhere (same in the other `resolve(true)` in this file)

packages/dd-trace/test/telemetry/dependencies.spec.js
```diff
@@ -13,6 +13,14 @@ const moduleLoadStartChannel = dc.channel('dd-trace:moduleLoadStart')
 const originalSetImmediate = global.setImmediate
 
 describe('dependencies', () => {
+  function setImmediate2 (callback, ...args) {
+    return callback(...args)
+  }
+  setImmediate2.__promisify__ = function (...args) {
```
watson, line 20: Nit:

```suggestion

  setImmediate2.__promisify__ = function (...args) {
```

package.json
```diff
@@ -12,6 +12,7 @@
     "dependencies:dedupe": "yarn-deduplicate yarn.lock",
     "type:doc": "cd docs && yarn && yarn build",
     "type:test": "cd docs && yarn && yarn test",
+    "check:types": "tsc --noEmit -p tsconfig.json",
```
watson: Sort lexically and consider renaming with a `type:` prefix

version.js
```diff
@@ -1,8 +1,10 @@
 'use strict'
 
 var version = require('./package.json').version
-var ddMatches = version.match(/^(\d+)\.(\d+)\.(\d+)/)
-var nodeMatches = process.versions.node.match(/^(\d+)\.(\d+)\.(\d+)/)
+// @ts-ignore
```
watson: Why not `@ts-expect-error`?

packages/dd-trace/src/rate_limiter.js
BridgeAR: I did consider this as well.
It just does the toString call internally otherwise. I think making it explicit is therefore better.

packages/dd-trace/src/rate_limiter.js
BridgeAR: That would be a regression, since we would allow decimal numbers in that case.

packages/dd-trace/test/plugins/agent.js
BridgeAR: Yes, I felt it's the best way to just have the TODO entry and comment it out, since it's definitely broken at the moment and I would like a separate PR that fixes it.

packages/dd-trace/test/plugins/agent.js
BridgeAR: This uses a type overload to make sure the types are narrowed ideally. I think that was overlooked?

packages/dd-trace/test/plugins/agent.js
BridgeAR: That is sadly a downside of the notation with JSDoc: it does not pick up the defaults in an IDE.

benchmark/sirun/run-util.js
BridgeAR: I tried multiple different things and non seem to work.

packages/dd-trace/src/pkg.js
BridgeAR: I added the try back in, keeping the undefined check nevertheless to prevent the overhead

bengl (dismissed):

packages/dd-trace/src/appsec/waf/index.js
```diff
@@ -9,7 +9,38 @@ const { ASM } = require('../../standalone/product')
...
+ *   destroy: () => void,
+ *   updateConfig: (product: string, configId: string, configPath: string, config: object) => void,
+ *   removeConfig: (configPath: string) => void,
+ *   checkAsmDdFallback: () => void,
+ *   run: (data: WafRunPayload, req?: IncomingMessage, raspRule?: string) => WafRunResult | void,
+ *   disposeContext: (req: IncomingMessage) => void,
+ *   WafUpdateError: typeof WafUpdateError
+ * }} WafAPI
+ */
+
```
simon-id: I have to admit i am not interested in maintaining this. also all the inline types are not something i enjoy, at this point let's just write typescript.
We are those definitions not in a jsdoc block on their respective methods ? why are they defined here ?

packages/dd-trace/test/telemetry/dependencies.spec.js
```diff
@@ -13,6 +13,15 @@ const moduleLoadStartChannel = dc.channel('dd-trace:moduleLoadStart')
...
 
 describe('dependencies', () => {
+  function setImmediate2 (callback, ...args) {
+    return callback(...args)
+  }
+
+  setImmediate2.__promisify__ = function (...args) {
+    setImmediate2(() => undefined, ...args)
+    return new Promise((resolve) => resolve(undefined))
+  }
```
simon-id, line 23: i do not understand what this does or why it's needed. Comment ?

simon-id (dismissed):
The changes to the two appsec files feels like i'm being imposed something on my codebase without me understanding what it does or how it works. Either spend some time explaining it to us, or let us set our typing system at our own pace in our own time.

benchmark/sirun/run-util.js
watson: I checked out this PR and got the following change to work in my IDE (I can't inline-suggest it here since it requires changes to part of the code outside this diff):

```diff
-  return new Promise((resolve, reject) => {
+  return /** @type {Promise<void>} */ (new Promise((resolve, reject) => {
     const proc = childProcess.spawn(...args)
     streamAddVersion(proc.stdout)
     proc.on('error', reject)
     proc.on('exit', (code) => {
       if (code === 0) {
-        resolve(undefined)
+        resolve()
       } else {
         reject(new Error('Process exited with non-zero code.'))
       }
     })
-  })
+  }))
```

watson: Something similar should be possible with the other promises

packages/dd-trace/src/rate_limiter.js
watson: Ah yes, forgot `parseInt` did that. Well, in that case, this is much more efficient:

```suggestion
    this._rateLimit = Math.trunc(Number(rateLimit))
```

(The old approach also had a potential bug in that it didn't pass in the base as the 2nd argument to `parseInt`)

packages/dd-trace/src/rate_limiter.js
watson: Ah I actually thought the comment was about the `constructor` function we were currently in, but I see now that it's for the different constructor function that we're calling below. Disregard 👍

packages/dd-trace/test/plugins/agent.js
watson: Why make it part of this PR then? It feels like it should be a separate PR

packages/dd-trace/test/plugins/agent.js
watson: Ah I didn't notice the other JSDoc comment-block above. If I understand `@overload` correctly, you define one base-case without `@overload` and then one or more using `@overload`. Here, both your cases uses `@overload`. The IDE is probably going to understand what you mean, but technically, I think, you should remove one of them - at least that is how I interpret these docs: https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#@overload-support-in-jsdoc

packages/dd-trace/src/appsec/waf/index.js
BridgeAR: We can move them to a ts definition file later. Right now, I am just trying to create the types and we will have to improve them while working on the code over time one way or the other. I suggest we can move them as soon as all the types are actually as expected.

packages/dd-trace/test/plugins/agent.js
BridgeAR: I can undo the changes, if you feel that's better. It is right definitely wrong code that will also not do anything when kept as is. Adding the TODO and out-commenting did seem like a few percent better than what we have :D

packages/dd-trace/src/rate_limiter.js
BridgeAR: This also has a slightly different behavior. parseInt is lenient when it comes to accepting all kinds of inputs, e.g., `123abc` is totally fine for it to return `123`. That would instead of up being `NaN` instead.

We should not have such cases in the code, I just rather not change the behavior at all in this PR.

benchmark/sirun/run-util.js
BridgeAR: This will work for most cases (two are left that would still need to resolve(undefined) or they would need a more complex overall type). I changed in all besides these two.

watson (approved):

packages/dd-trace/test/telemetry/dependencies.spec.js
BridgeAR: This is a custom promisify method to mimic the original setImmediate. It is a Node.js API.

packages/dd-trace/src/appsec/waf/index.js
BridgeAR: I moved the changes out and I will open a separate PR for them when this is merged

integration-tests/helpers/index.js
```diff
@@ -167,7 +193,7 @@ function spawnProc (filename, options = {}, stdioHandler, stderrHandler) {
         if (code !== 0) {
           return reject(new Error(`Process exited with status code ${code}.`))
         }
-        resolve()
+        resolve(undefined)
```
watson: This patch should allow us to not use `resolve(undefined)` her:

```diff
diff --git a/integration-tests/helpers/index.js b/integration-tests/helpers/index.js
index 146c24ec6..e35f46c38 100644
--- a/integration-tests/helpers/index.js
+++ b/integration-tests/helpers/index.js
@@ -172,14 +172,14 @@ function assertTelemetryPoints (pid, msgs, expectedTelemetryPoints) {
  *   standard output of the child process. If not provided, the output will be logged to the console.
  * @param {(data: Buffer) => void} [stderrHandler] - A function that's called with one data argument to handle the
  *   standard error of the child process. If not provided, the error will be logged to the console.
- * @returns {Promise<childProcess.ChildProcess & { url?: string }|undefined>} A promise that resolves when the process
+ * @returns {Promise<childProcess.ChildProcess & { url?: string }|void>} A promise that resolves when the process
  *   is either ready or terminated without an error. If the process is terminated without an error, the promise will
  *   resolve with `undefined`.The returned process will have a `url` property if the process didn't terminate.
  */
 function spawnProc (filename, options = {}, stdioHandler, stderrHandler) {
   const proc = fork(filename, { ...options, stdio: 'pipe' })
 
-  return new Promise((resolve, reject) => {
+  return /** @type {Promise<childProcess.ChildProcess & { url?: string }|void>} */ (new Promise((resolve, reject) => {
     proc
       .on('message', ({ port }) => {
         if (typeof port !== 'number' && typeof port !== 'string') {
@@ -193,7 +193,7 @@ function spawnProc (filename, options = {}, stdioHandler, stderrHandler) {
         if (code !== 0) {
           return reject(new Error(`Process exited with status code ${code}.`))
         }
-        resolve(undefined)
+        resolve()
       })
 
     proc.stdout.on('data', data => {
@@ -211,7 +211,7 @@ function spawnProc (filename, options = {}, stdioHandler, stderrHandler) {
       // eslint-disable-next-line no-console
       if (!options.silent) console.error(data.toString())
     })
-  })
```

integration-tests/opentelemetry.spec.js
```diff
@@ -26,7 +26,7 @@ async function check (agent, proc, timeout, onMessage = () => { }, isMetrics) {
           if (code !== 0) {
             reject(new Error(`Process exited with unexpected status code ${code}.`))
           } else {
-            resolve()
+            resolve(undefined)
```
watson: Also here:

```diff
diff --git a/integration-tests/opentelemetry.spec.js b/integration-tests/opentelemetry.spec.js
index 878ebfd55..802a5e091 100644
--- a/integration-tests/opentelemetry.spec.js
+++ b/integration-tests/opentelemetry.spec.js
@@ -13,7 +13,7 @@ async function check (agent, proc, timeout, onMessage = () => { }, isMetrics) {
 
   const [res] = await Promise.all([
     messageReceiver,
-    new Promise((resolve, reject) => {
+    /** @type {Promise<void>} */ (new Promise((resolve, reject) => {
       const timer = setTimeout(() => {
         reject(new Error('Process timed out'))
       }, timeout)
@@ -26,7 +26,7 @@ async function check (agent, proc, timeout, onMessage = () => { }, isMetrics) {
           if (code !== 0) {
             reject(new Error(`Process exited with unexpected status code ${code}.`))
           } else {
-            resolve(undefined)
+            resolve()
           }
         })
     })
```

packages/dd-trace/test/telemetry/dependencies.spec.js
```diff
@@ -13,6 +13,15 @@ const moduleLoadStartChannel = dc.channel('dd-trace:moduleLoadStart')
 const originalSetImmediate = global.setImmediate
 
 describe('dependencies', () => {
+  function setImmediate2 (callback, ...args) {
+    return callback(...args)
+  }
+
+  setImmediate2.__promisify__ = function (...args) {
+    setImmediate2(() => undefined, ...args)
+    return new Promise((resolve) => resolve(undefined))
```
watson: And here:

```suggestion
    return /** @type {Promise<void>} */ (new Promise((resolve) => resolve()))
```

But maybe even simpler:

```suggestion
    return Promise.resolve()
```

bengl (dismissed):

bengl (approved):

PR #6448, opened 2025-09-16
[PREVIEW] Add Feature Flagging & Experimentations

khanayan123:
### What does this PR do?
Adds Feature Flagging & Experimentation (FFE) module:
  - OpenFeature-compatible provider for flag evaluation
  - Remote Config integration for flag configuration (UFC format via FFE_FLAGS product)
  - Automatic exposure event tracking with buffering/batching to intake via EVP proxy
  - Performance benchmarks for ExposuresWriter operations

### Preview Status
This feature is in preview and not ready for customer use. UI/platform changes for Remote Config delivery are in progress. This PR establishes code infrastructure and wiring only.

## Key Components
  - FlaggingProvider: Wraps @datadog/openfeature-node-server
  - ExposuresWriter: Buffers events (5MB payload limit, 1MB per event)
  - NoopFlaggingProvider: Fallback when disabled
  - Remote Config handler for FFE_FLAGS product which delivers UFC payloads from the UI to the flagging provider
  - Benchmarks: Microbenchmarks for ExposuresWriter operations in benchmark/openfeature.js

### Motivation
Enable feature flag evaluation in Node.js server applications with automatic exposure event collection for experimentation analytics and flag performance monitoring

### Notes
  - Config: DD_EXPERIMENTAL_FLAGGING_PROVIDER_ENABLED=true to enable
  - Dependencies: Requires @datadog/openfeature-node-server@0.1.0-preview.10 and @openfeature/server-sdk@~1.18.0
  - Breaking Changes: None (opt-in feature)
  - Remote Config: handler for FFE_FLAGS product & FFE_FLAG_CONFIGURATION_RULES added for UFC config delivery
  - Type Safety: getConfiguration()/setConfiguration() use any type due to upstream @openfeature/server-sdk type compatibility issues
  - End to end system test PR:
    - add parametric tests for ff&e https://github.com/DataDog/system-tests/pull/5337
    - add FFE capability support for Node.js https://github.com/DataDog/system-tests/pull/5381 - fix for failing CI test (will merge once
  release version is known)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -212,6 +212,13 @@
     "@babel/helper-string-parser" "^7.27.1"
     "@babel/helper-validator-identifier" "^7.27.1"
 
+"@datadog/flagging-core@0.1.0-preview.10":
+  version "0.1.0-preview.10"
+  resolved "https://registry.yarnpkg.com/@datadog/flagging-core/-/flagging-core-0.1.0-preview.10.tgz#a581d391955ec6b7963494cf5312f77c52d2e46f"
+  integrity sha512-E1jRY80xqgL9Hie3vGbQUiK8FXijeN+h9DHXourcPgw+F2fiQ3m1sLTNrjrZTOOpKfhJE/lvahssQpMRtQh1TA==
+  dependencies:
+    spark-md5 "^3.0.2"
+
 "@datadog/libdatadog@0.7.0":
   version "0.7.0"
   resolved "https://registry.yarnpkg.com/@datadog/libdatadog/-/libdatadog-0.7.0.tgz#81e07d3040c628892db697ccd01ae3c4d2a76315"
@@ -238,6 +245,14 @@
   dependencies:
     node-addon-api "^6.1.0"
     node-gyp-build "^3.9.0"
+
+"@datadog/openfeature-node-server@0.1.0-preview.10":
+  version "0.1.0-preview.10"
+  resolved "https://registry.yarnpkg.com/@datadog/openfeature-node-server/-/openfeature-node-server-0.1.0-preview.10.tgz#9ed827d8d35f09132b65c1cd32a82df31daedd70"
+  integrity sha512-btTKlz80Gn2Xj0m2LR9ZKkoB/Wybu4mFa86wPJv3G1IKU61wgPpvdIILDEwDz+lg5gLjxxeJM50r4tP82bEteQ==
+  dependencies:
+    "@datadog/flagging-core" "0.1.0-preview.10"
+    "@openfeature/server-sdk" "~1.18.0"
 
 "@datadog/pprof@5.10.0":
   version "5.10.0"
@@ -681,6 +696,16 @@
     "@octokit/request-error" "^7.0.0"
     "@octokit/webhooks-methods" "^6.0.0"
 
+"@openfeature/core@^1.8.1":
+  version "1.9.1"
+  resolved "https://registry.yarnpkg.com/@openfeature/core/-/core-1.9.1.tgz#9925a04ed0745e92dd7b3793b35cff1ed89d54c1"
+  integrity sha512-YySPtH4s/rKKnHRU0xyFGrqMU8XA+OIPNWDrlEFxE6DCVWCIrxE5YpiB94YD2jMFn6SSdA0cwQ8vLkCkl8lm8A==
+
+"@openfeature/server-sdk@~1.18.0":
+  version "1.18.0"
+  resolved "https://registry.yarnpkg.com/@openfeature/server-sdk/-/server-sdk-1.18.0.tgz#5fed5f1d0900b2535878db18a78295bbaebb997b"
+  integrity sha512-uP8nqEGBS58s3iWXx6d8vnJ6ZVt3DACJL4PWADOAuwIS4xXpID91783e9f6zQ0i1ijQpj3yx+3ZuCB2LfQMUMA==
+
 "@opentelemetry/api-logs@<1.0.0":
   version "0.205.0"
   resolved "https://registry.npmjs.org/@opentelemetry/api-logs/-/api-logs-0.205.0.tgz"
@@ -688,12 +713,12 @@
   dependencies:
     "@opentelemetry/api" "^1.3.0"
 
-"@opentelemetry/api@>=1.0.0 <1.10.0":
+"@opentelemetry/api@>=1.0.0 <1.10.0", "@opentelemetry/api@^1.3.0":
   version "1.9.0"
   resolved "https://registry.yarnpkg.com/@opentelemetry/api/-/api-1.9.0.tgz#d03eba68273dc0f7509e2a3d5cba21eae10379fe"
   integrity sha512-3giAOQvZiH5F9bMlMiv8+GSPMeqg0dbaeo58/0SlA9sxSqZhnUtxzX9/2FzyhS9sWQf5S0GJE0AKBrFqjpeYcg==
 
-"@opentelemetry/core@>=1.14.0 <1.31.0":
+"@opentelemetry/core@1.30.1", "@opentelemetry/core@>=1.14.0 <1.31.0":
   version "1.30.1"
   resolved "https://registry.yarnpkg.com/@opentelemetry/core/-/core-1.30.1.tgz#a0b468bb396358df801881709ea38299fc30ab27"
   integrity sha512-OOCM2C/QIURhJMuKaekP3TRBxBKxG/TWWA0TL2J6nXUtDnuCtccy49LUJF8xPFXMX+0LMcxFpCo8M9cGY1W6rQ==
@@ -705,8 +730,8 @@
   resolved "https://registry.npmjs.org/@opentelemetry/resources/-/resources-1.9.1.tgz"
   integrity sha512-VqBGbnAfubI+l+yrtYxeLyOoL358JK57btPMJDd3TCOV3mV5TNBmzvOfmesM4NeTyXuGJByd3XvOHvFezLn3rQ==
   dependencies:
-    "@opentelemetry/core" "1.30.1"
-    "@opentelemetry/semantic-conventions" "1.28.0"
+    "@opentelemetry/core" "1.9.1"
+    "@opentelemetry/semantic-conventions" "1.9.1"
 
 "@opentelemetry/semantic-conventions@1.28.0":
   version "1.28.0"
@@ -4338,6 +4363,11 @@ source-map@^0.7.4:
   version "0.7.6"
   resolved "https://registry.yarnpkg.com/source-map/-/source-map-0.7.6.tgz#a3658ab87e5b6429c8a1f3ba0083d4c61ca3ef02"
   integrity sha512-i5uvt8C3ikiWeNZSVZNWcfZPItFQOsYTUAOkcUPGd8DqDy1uOUikjt5dG+uRlwyvR108Fb9DOd4GvXfT0N2/uQ==
+
+spark-md5@^3.0.2:
+  version "3.0.2"
+  resolved "https://registry.yarnpkg.com/spark-md5/-/spark-md5-3.0.2.tgz#7952c4a30784347abcee73268e473b9c0167e3fc"
+  integrity sha512-wcFzz9cDfbuqe0FZzfi2or1sgyIrsDwmPwfZC4hiNidPdPINjeUwNfv5kldczoEAcjl9Y1L3SM7Uz2PUEQzxQw==
 
 spawn-wrap@^2.0.0:
   version "2.0.0"


integration-tests/ffe/ffe-remote-config.spec.js
```diff
@@ -0,0 +1,149 @@
...
+        assert.strictEqual(state, ACKNOWLEDGED)
+        assert.notOk(error) // falsy check since error will be an empty string, but that's an implementation detail
+
+        receivedAckUpdate = true
+        endIfDone()
+      })
+
+      // Add UFC config using fixture
+      agent.addRemoteConfig({
+        product: FFE_FLAG_CONFIGURATION_RULES,
```
leoromanovsky: RC team is creating a single product - `FFE_FLAGS`; sorry for the disruption.

packages/dd-trace/src/ffe/noop.js
```diff
@@ -0,0 +1,53 @@
...
+    this._config = {}
+    this.metadata = { name: 'NoopFlaggingProvider' }
+    this.status = 'NOT_READY'
+    this.runsOn = 'server'
+  }
+
+  resolveBooleanEvaluation (flagKey, defaultValue, context, logger) {
+    return Promise.resolve({
+      value: defaultValue,
+      reason: 'DEFAULT'
```
leoromanovsky: `STATIC` might fit better

packages/dd-trace/test/plugins/versions/package.json
```diff
@@ -54,6 +54,10 @@
     "@opentelemetry/instrumentation-express": "0.53.0",
     "@opentelemetry/instrumentation-http": "0.204.0",
     "@opentelemetry/sdk-node": "0.204.0",
+    "@openfeature/server-sdk": "^1.19.0",
+    "@openfeature/core": "^1.9.0",
+    "@datadog/openfeature-node-server": "0.1.0-preview.6",
```
leoromanovsky: 7

packages/dd-trace/src/ffe/writers/exposures.js
```diff
@@ -0,0 +1,112 @@
...
+      },
+      flag: {
+        key: event.flag?.key || event['flag.key']
+      },
+      variant: {
+        key: event.variant?.key || event['variant.key']
+      },
+      subject: {
+        id: event.subject?.id || event['subject.id'],
+        type: event.subject?.type || 'user', // defaults to 'user'
```
leoromanovsky: This is not a safe default; better to throw an exception if missing. Is `user` a special default in APM?

```suggestion
        type: event.subject?.type,
```

leoromanovsky: Look at the schema: https://github.com/DataDog/dd-source/blob/main/domains/evp-workers/apps/exposures-worker/schemas/exposure.json

Only `id` is required - the other keys can be omitted if not provided.

packages/dd-trace/src/proxy.js
```diff
@@ -155,6 +156,18 @@ class Tracer extends NoopProxy {
...
+        if (config.flaggingProvider.enabled) {
+          rc.setProductHandler('FFE_FLAGS', (action, conf) => {
+            // Feed UFC config directly to flagging provider
+            if (action === 'apply' || action === 'modify') {
+              this.flaggingProvider._setConfiguration(conf.flag_configuration)
+            } else if (action === 'unapply') {
+              // For now, just pass empty config
+              this.flaggingProvider._setConfiguration({})
+            }
+          })
```
simon-id, line 169: FYI, the remote config manager doesn't do any merging of received configurations, which means if the backend sends multiple "FFE_FLAGS" (ask whoever is implementing this in the backend to know if it's possible), then each FFE_FLAGS payload will override the previous one, without consolidating them all into one.

Also, line 167 means you'll completely remove the configuration right ? if you have a default/initial config set, you might wanna restore that ?

Just my 2 cents from a RC connoisseur :)

packages/dd-trace/src/proxy.js
```diff
@@ -155,6 +156,18 @@ class Tracer extends NoopProxy {
...
         }
+
+        if (config.flaggingProvider.enabled) {
+          rc.setProductHandler('FFE_FLAGS', (action, conf) => {
+            // Feed UFC config directly to flagging provider
+            if (action === 'apply' || action === 'modify') {
+              this.flaggingProvider._setConfiguration(conf.flag_configuration)
+            } else if (action === 'unapply') {
+              // For now, just pass empty config
+              this.flaggingProvider._setConfiguration({})
```
greghuels: Can you pass `this.flaggingProvider._setConfiguration(undefined)` instead? Passing an empty object would cause an `undefined is not an object` error during evaluation since it expects a "flags" key to exist with an array value.

greghuels: It might be good to cover this with a full integration test that uses the OpenFeature sdk client `@openfeature/server-sdk@~1.18.0` for evaluation

packages/dd-trace/test/plugins/versions/package.json
```diff
@@ -49,6 +50,8 @@
     "@langchain/google-genai": "0.2.18",
     "@langchain/openai": "0.6.13",
     "@node-redis/client": "1.0.6",
+    "@openfeature/core": "^1.9.0",
+    "@openfeature/server-sdk": "~1.18.0",
```
greghuels, line 55: This should also be listed as a peerDependency

greghuels (changes requested):
Main blocker is putting the provider in a state that would cause an error during evaluation

packages/dd-trace/src/proxy.js
khanayan123: @leoromanovsky to confirm: Does the FFE backend send multiple separate FFE_FLAGS configs, or does it always send one consolidated config with all flags? If it sends multiple to represent one full payload, then we will need to implement merging logic

leoromanovsky: A single configuration will have all the flags for an environment; no need to merge, in fact it is not desirable, best to removing existing cache otherwise flags will be stale.

khanayan123: @simon-id  "Also, line 167 means you'll completely remove the configuration right ? if you have a default/initial config set, you might wanna restore that ?"

I don't think there's any current plans to use the 'unapply' action. I think it makes sense to probably remove the handling of it and add it as needed

packages/dd-trace/src/proxy.js
khanayan123: I'am going to remove the handling of that action for now since we don't have a defined spec for it, could add it later as need be

index.d.ts
```diff
@@ -1059,6 +1084,88 @@ declare namespace tracer {
...
+     * @returns Promise resolving to evaluation result with value and reason
+     */
+    resolveObjectEvaluation(flagKey: string, defaultValue: any, context?: any, logger?: any): Promise<{ value: any; reason: string }>;
+
+    /**
+     * Gets the current flag configuration.
+     *
+     * @returns The current configuration object
+     */
+    getConfiguration(): any;
```
greghuels: Using `any` here could cause friction for consumers with strict type analysis checks.

You could do this at the top of the file:
```
import { type DatadogNodeServerProvider } from '@datadog/openfeature-node-server';
type FlaggingConfiguration = ReturnType<DatadogNodeServerProvider['getConfiguration']>;
```

Then, you'd have `getConfiguration(): FlaggingConfiguration`, and `setConfiguration(config: FlaggingConfiguration)`

packages/dd-trace/src/ffe/sdk.js
```diff
@@ -0,0 +1,30 @@
...
+const { DatadogNodeServerProvider } = require('@datadog/openfeature-node-server')
+const { channel } = require('dc-polyfill')
+const log = require('../log')
+const { EXPOSURE_CHANNEL } = require('./constants/constants')
+
+class FlaggingProvider extends DatadogNodeServerProvider {
+  constructor (tracer, config) {
+    // Call parent constructor with required options
+    super({
+      configuration: undefined,
```
greghuels: nit: configuration is no longer a constructor arg. Not an issue for this PR since it's set to undefined.

index.d.ts
bengl: Similarly for the following two methods.

index.d.ts
```diff
@@ -1059,6 +1084,88 @@ declare namespace tracer {
...
+    setConfiguration(config: any): void;
+
+    /**
+     * Internal method to update flag configuration from Remote Config.
+     * This method is called automatically when Remote Config delivers UFC updates.
+     *
+     * @internal
+     * @param ufc Universal Flag Configuration object
+     */
+    _setConfiguration(ufc: any): void;
```
bengl: We probably don't want internal methods in our `.d.ts` file. Can you just have the JSDoc inline?

package.json
```diff
@@ -35,7 +35,7 @@
     "test:instrumentations:misc": "mocha -r 'packages/dd-trace/test/setup/mocha.js' 'packages/datadog-instrumentations/test/*/**/*.spec.js'",
     "test:instrumentations:misc:ci": "nyc --no-clean --include 'packages/datadog-instrumentations/src/**/*.js' -- npm run test:instrumentations:misc",
     "test:core": "tap \"packages/datadog-core/test/**/*.spec.js\"",
-    "test:core:ci": "npm run test:core -- --coverage --nyc-arg=--include=\"packages/datadog-core/src/**/*.js\"",
+  "test:core:ci": "npm run test:core -- --coverage --nyc-arg=--include=\"packages/datadog-core/src/**/*.js\"",
```
bengl: Please maintain the correct indentation.

integration-tests/openfeature/openfeature-exposure-events.spec.js
```diff
@@ -0,0 +1,325 @@
+'use strict'
+
+const { createSandbox, FakeAgent, spawnProc } = require('../helpers')
+const path = require('path')
+const Axios = require('axios')
```
bengl: Let's avoid adding new axios tests. Can you use `fetch` instead?

package.json
watson:
```suggestion
    "test:core:ci": "npm run test:core -- --coverage --nyc-arg=--include=\"packages/datadog-core/src/**/*.js\"",
```

index.d.ts
```diff
@@ -139,6 +139,16 @@ interface Tracer extends opentracing.Tracer {
...
 
+  /**
+   * Feature Flagging Provider (OpenFeature-compatible)
+   *
+   * Extends DatadogNodeServerProvider with Remote Config integration.
+   * Enable with DD_FLAGGING_PROVIDER_ENABLED=true.
+   *
+   * @beta This feature is in preview and not ready for production use
+   */
+  flaggingProvider: tracer.FlaggingProvider;
```
leoromanovsky: The term `openfeature` will be known to users and a familiar export; I noticed in the tracer there is a module exported as `opentelemetry` (https://datadoghq.dev/dd-trace-js/modules.html#opentelemetry).

```
openfeature: tracer.OpenFeatureProvider
```

index.d.ts
```diff
@@ -1059,6 +1084,88 @@ declare namespace tracer {
...
+    /**
+     * Resolves a boolean flag value.
+     *
+     * @param flagKey The key of the flag to evaluate
+     * @param defaultValue The default value to return if evaluation fails
+     * @param context Evaluation context (e.g., user attributes)
+     * @param logger Optional logger instance
+     * @returns Promise resolving to evaluation result with value and reason
+     */
+    resolveBooleanEvaluation(flagKey: string, defaultValue: boolean, context?: any, logger?: any): Promise<{ value: boolean; reason: string }>;
```
leoromanovsky: If `openfeature-core` is a peer dependency, are you able to use the types? Here we are constructing our own API (specifically returning a `string`) reason instead of using an enum from the library or `ResolutionDetails` as the spec says.

leoromanovsky (commented):
Main feedback is on the public API: matching the OF spec in terms of naming and method response type.

package.json
```diff
@@ -148,11 +150,25 @@
     "tlhunter-sorted-set": "^0.1.0",
     "ttl-set": "^1.0.0"
   },
+  "peerDependencies": {
+    "@openfeature/core": "^1.9.0",
+    "@openfeature/server-sdk": "~1.18.0"
+  },
```
leoromanovsky, line 161: @bengl FYI https://openfeature.dev/docs/reference/technologies/server/javascript/#considerations

integration-tests/ffe/app/exposure-events.js
```diff
@@ -0,0 +1,108 @@
...
+const tracer = require('dd-trace')
+tracer.init({
+  flushInterval: 1,
+  service: 'ffe-test-service',
+  version: '1.2.3',
+  env: 'test'
+})
+const express = require('express')
+const { channel } = require('dc-polyfill')
+const { flaggingProvider } = tracer
```
bengl: Note: As discussed offline, we'll need to figure out how this works for `dd-trace-api` users.

packages/dd-trace/src/openfeature/writers/base.js
```diff
@@ -0,0 +1,156 @@
...
+const request = require('../../exporters/common/request')
+const { safeJSONStringify } = require('../../exporters/common/util')
+const { getEnvironmentVariable } = require('../../config-helper')
+const { URL, format } = require('node:url')
+const path = require('node:path')
+
+const log = require('../../log')
+
+class BaseFFEWriter {
+  constructor ({ interval, timeout, config, endpoint, agentUrl, payloadSizeLimit, eventSizeLimit, headers }) {
```
bengl, line 29: In this and all other new files, please add JSDoc types.

packages/dd-trace/src/openfeature/writers/base.js
```diff
@@ -0,0 +1,156 @@
...
+    // Private env vars for testing purposes
```
bengl: Can we do this some way other than environment variables? You're already allowing them to be passed in via the constructor, so maybe you can handle this elsewhere for testing?

packages/dd-trace/src/openfeature/writers/base.js
```diff
@@ -0,0 +1,156 @@
...
+    this._droppedEvents = 0
+  }
+
+  get url () {
+    const baseUrl = this._baseUrl.href
+    const endpoint = this._endpoint
+
+    // Split on protocol separator to preserve it
+    const [protocol, rest] = baseUrl.split('://')
+    return protocol + '://' + path.join(rest, endpoint)
```
bengl: This will be incorrect on Windows, since it'll add a backslash. Is all this necessary? Where is url even used?

packages/dd-trace/src/openfeature/writers/base.js
```diff
@@ -0,0 +1,156 @@
...
+      if (this._droppedEvents > 0) {
+        log.warn(`${this.constructor.name} dropped ${this._droppedEvents} events due to buffer overflow`)
+      }
+    }
+  }
+
+  _getAgentUrl () {
+    const { hostname, port } = this._config
+
+    const overrideOriginEnv = getEnvironmentVariable('_DD_FFE_OVERRIDE_ORIGIN')
```
bengl: Similar to above with environment variables.

packages/dd-trace/src/openfeature/index.js
```diff
@@ -0,0 +1,59 @@
...
+const exposureSubmitCh = channel('ffe:exposure:submit')
+const flushCh = channel('ffe:writers:flush')
+
+let exposuresWriter = null
+
+function _handleExposureSubmit (exposureEvents) {
+  if (!exposuresWriter) return
+
+  const events = Array.isArray(exposureEvents) ? exposureEvents : [exposureEvents]
+  for (const event of events) {
```
bengl: Rather than looping here, it's probably working making `.append()` able to take an array.

packages/dd-trace/src/openfeature/noop.js
```diff
@@ -0,0 +1,55 @@
+'use strict'
+
+const { NOOP_REASON } = require('./constants/constants')
+
+class NoopFlaggingProvider {
```
bengl, line 10: Can you put a comment linking to the FlaggingProvider spec?

packages/dd-trace/src/config.js
```diff
@@ -628,7 +628,8 @@ class Config {
       OTEL_RESOURCE_ATTRIBUTES,
       OTEL_SERVICE_NAME,
       OTEL_TRACES_SAMPLER,
-      OTEL_TRACES_SAMPLER_ARG
+      OTEL_TRACES_SAMPLER_ARG,
+      DD_FLAGGING_PROVIDER_ENABLED
```
bengl: Let's consider making this (and any programmatic config) "experimental", prior to GA.

packages/dd-trace/test/openfeature/writers/exposures.spec.js
```diff
@@ -0,0 +1,384 @@
...
+      }
+    }
+
+    request = sinon.stub().yieldsAsync(null, 'OK', 200)
+
+    config = {
+      site: 'datadoghq.com',
+      hostname: 'localhost',
+      port: 8126,
+      url: new (require('url').URL)('http://localhost:8126'),
```
bengl: URL is a global, isn't it?

bengl (changes requested):
Can you add benchmarks, particularly for the writer?

index.d.ts
khanayan123: I kept any here because any import from @datadog/openfeature-node-server transitively pulls in @openfeature/server-sdk@~1.18.0, which has broken type definitions that fail our yarn type:test

index.d.ts
khanayan123: Cant import the types if I have it as a peer dependency, does it make sense to just add the package as a core dependency?

cc: @bengl

integration-tests/openfeature/app/exposure-events.js
```diff
@@ -0,0 +1,109 @@
...
+// Used to test manual flushing capabilities
+const flushCh = channel('ffe:writers:flush')
+
+OpenFeature.setProvider(openfeature)
+const client = OpenFeature.getClient()
+
+const app = express()
+
+// Used to test remote config polling capabilities
+app.get('/', async (req, res) => {
```
bengl: Why is this async?

integration-tests/openfeature/app/exposure-events.js
```diff
@@ -0,0 +1,109 @@
...
+    res.status(500).json({ error: error.message })
+  }
+})
+
+app.get('/flush', async (req, res) => {
+  flushCh.publish()
+  res.json({ flushed: true })
+})
+
+const server = app.listen(process.env.APP_PORT || 0, (error) => {
```
bengl: Is `server` ever referenced elsewhere? There doesn't seem to be a real need to assign this to a `const` here.

packages/dd-trace/src/openfeature/writers/base.js
```diff
@@ -0,0 +1,190 @@
...
+    if (this._buffer.length === 0) {
+      return
+    }
+    const events = this._buffer
+    this._buffer = []
+    this._bufferSize = 0
+
+    const payload = this._encode(this.makePayload(events))
+
+    log.debug('Encoded payload: %s', safeJSONStringify(payload))
```
bengl: This will stringify the entire payload on every flush, without using the result unless debug is enabled. Please don't stringify if debug is not enabled.

If you're going to keep this log message, please clarify in the log message where this is coming from. "payload" is ambiguous because we have many payloads. It may be good to prefix all the log messages in here with the subsystem name.

packages/dd-trace/src/openfeature/writers/base.js
```diff
@@ -0,0 +1,190 @@
...
+      port: port || 8126
+    }))
+  }
+
+  /**
+   * @private
+   * @returns {Object} constructs HTTP request options
+   */
+  _getOptions () {
+    const options = {
```
bengl: You're re-creating this options object on each flush. Can it be re-used?

packages/dd-trace/src/openfeature/writers/exposures.js
```diff
@@ -0,0 +1,163 @@
...
+
+  /**
+   * Formats exposure events with service context metadata
+   * @param {Array<ExposureEvent>} events - Array of exposure events
+   * @returns {ExposureEventPayload} Formatted payload with service context
+   */
+  makePayload (events) {
+    const formattedEvents = events.map(event => this._formatExposureEvent(event))
+
+    const context = {
```
bengl, line 129: It looks like the context object can be cached, clearing the cache upon config change.

index.d.ts
bengl: Can we not override them?

Also, adding a new API for getting and setting config, even if it's just for one subsystem, expands our config surface (cc @BridgeAR).

packages/dd-trace/src/openfeature/index.js
```diff
@@ -0,0 +1,74 @@
...
+  exposuresWriter?.flush()
+}
+
+module.exports = {
+  /**
+   * Enables the OpenFeature module and sets up FF&E writer and channel subscribers
+   * @param {import('../config')} config - Tracer configuration object
+   * @returns {void}
+   */
+  enable (config) {
```
IlyasShabi: Nit: move enable/disable function outside the module.exports

packages/dd-trace/src/openfeature/writers/base.js
```diff
@@ -0,0 +1,190 @@
...
+      headers: {
+        'Content-Type': 'application/json',
+        ...this._headers
```
IlyasShabi: I think we should change the order to make sure we don't override content-type

index.d.ts
khanayan123: That is true, it is not part of the official provider spec: https://openfeature.dev/docs/reference/concepts/provider/ so we can just not expose it publicly

PR #6449, opened 2025-09-16
chore: make what's possible in wall.js private; sort some fields

szegedi:
### What does this PR do?
make what's possible in `wall.js` private; sort some fields and their usages alphabetically.

### Motivation
Better encapsulation and overall code hygiene.


BridgeAR (approved):
LGTM

PR #6455, opened 2025-09-16
feat(loader-hook)!: add include configuration for import-in-the-middle

rochdev:
### What does this PR do?

Add include configuration to iitm loader hook.

### Motivation

This avoids patching every single module and instead only patches modules we instrument. Since iitm is not compatible with every module, this is needed to avoid issues with unsupported modules.

### DIFF
--- a/loader-hook.mjs
+++ b/loader-hook.mjs
@@ -1,8 +1,6 @@
 import * as iitm from 'import-in-the-middle/hook.mjs'
 import hooks from './packages/datadog-instrumentations/src/helpers/hooks.js'
-import { getEnvironmentVariables } from './packages/dd-trace/src/config-helper.js'
-
-const { DD_IAST_SECURITY_CONTROLS_CONFIGURATION } = getEnvironmentVariables()
+import { getEnvironmentVariable as env } from './packages/dd-trace/src/config-helper.js'
 
 function initialize (data = {}) {
   data.include ??= []
@@ -24,7 +22,7 @@ function addInstrumentations (data) {
 }
 
 function addSecurityControls (data) {
-  const securityControls = (DD_IAST_SECURITY_CONTROLS_CONFIGURATION || '')
+  const securityControls = (env('DD_IAST_SECURITY_CONTROLS_CONFIGURATION') || '')
     .split(';')
     .map(sc => sc.trim().split(':')[2])
     .filter(Boolean)
--- a/loader-hook.mjs
+++ b/loader-hook.mjs
@@ -1,1 +1,29 @@
-export * from 'import-in-the-middle/hook.mjs'
+// export * from 'import-in-the-middle/hook.mjs'
+
+import * as iitm from 'import-in-the-middle/hook.mjs'
+import hooks from './packages/datadog-instrumentations/src/helpers/hooks.js'
+
+const nodules = Object.keys(hooks)
+
+function initialize (data = {}) {
+  data.include ??= []
+  data.exclude ??= []
+  data.exclude.push(
+    /middle/,
+    /langsmith/,
+    /openai\/_shims/,
+    /openai\/resources\/chat\/completions\/messages/,
+    /openai\/agents-core\/dist\/shims/,
+    /@anthropic-ai\/sdk\/_shims/
+  )
+
+  for (const nodule of nodules) {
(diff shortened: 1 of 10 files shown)


BridgeAR:
I am marking this as draft until we find a solution for the appsec code

loader-hook.mjs
```diff
@@ -1 +1,37 @@
-export * from 'import-in-the-middle/hook.mjs'
+import * as iitm from 'import-in-the-middle/hook.mjs'
+import hooks from './packages/datadog-instrumentations/src/helpers/hooks.js'
+import { getEnvironmentVariables } from './packages/dd-trace/src/config-helper.js'
+
+const { DD_IAST_SECURITY_CONTROLS_CONFIGURATION } = getEnvironmentVariables()
```
BridgeAR:
```suggestion
const DD_IAST_SECURITY_CONTROLS_CONFIGURATION = getEnvironmentVariable('DD_IAST_SECURITY_CONTROLS_CONFIGURATION')
```

The other call is relatively expensive.

loader-hook.mjs
```diff
@@ -1 +1,37 @@
...
+
+function initialize (data = {}) {
+  const instrumentations = Object.keys(hooks)
+  const securityControls = (DD_IAST_SECURITY_CONTROLS_CONFIGURATION || '')
```
BridgeAR: I think it would be nice to have the variables local to where they are used next and I would move this down right before the loop.

Same about instrumentations.

loader-hook.mjs
```diff
@@ -1 +1,37 @@
...
+    /openai\/resources\/chat\/completions\/messages/,
+    /openai\/agents-core\/dist\/shims/
+  )
+
+  for (const moduleName of instrumentations) {
+    data.include.push(new RegExp(`node_modules/${moduleName}/(?!node_modules).+`), moduleName)
+  }
+
+  for (const subpath of securityControls) {
+    data.include.push(new RegExp(subpath))
```
BridgeAR: Are the inputs expected to be escaped properly?

rochdev: What issues would you expect from an incorrect input?

loader-hook.mjs
rochdev: The current syntax uses less characters and doesn't repeat the name which is redundant.

rochdev: Why is the other syntax expensive? It seems like something that should only happen once, and also it seems like a foot gun since I would expect it to be cheap.

BridgeAR: The issue is that we manipulate the envs during runtime. So each time we call this, we have to effectively create a copy of all envs. When using the individual getter call, we do not have to copy anything. It is pretty much zero cost.

I agree about the foot gun and think we could add a linter for it not being used outside of config.js.

loader-hook.mjs
BridgeAR: It is just about matching the right things for user input. It is something that seems rather difficult to get right from the API design itself.

It is not really possible to detect if the input should be escaped or not. Thus, the documentation has to make it very clear.

BridgeAR (commented):
This generally looks good to me, while I would appreciate if e.g., @uurien or @simon-id could have a look at the escaping: it is tricky to determine if the user should escape the path or not and we should probably clearly document that. Otherwise it would seemingly work while effectively not matching the anticipated paths.

loader-hook.mjs
rochdev: Ended up aliasing the helper and using it inline since the initialization only happens once.

loader-hook.mjs
rochdev: I escaped the input. The option is meant for something else, so it would always contain a path, not a pattern.

rochdev:
> it is tricky to determine if the user should escape the path or not and we should probably clearly document that

Customers should never escape anything as they are not passing a pattern but a list of local files. This should therefore not be documented as it's not a supported use case for the environment variable. This is also why I don't think escaping should ever be a problem.

rochdev:
I had to revert unrelated changes to the lockfile that were failing, but should be good now.

BridgeAR (dismissed):

juan-fernandez (approved):

PR #6458, opened 2025-09-17
[test optimization] Support `vitest` programmatic API

juan-fernandez:
### What does this PR do?

* Support this API by adding the necessary hook

### Motivation
Allow users to use the `vitest` programmatic API:

```javascript
import { startVitest } from 'vitest/node'

async function runProgrammaticTests () {
  try {
    const vitest = await startVitest('test', [], {
      test: {
        environment: 'node',
      },
      run: true,
      watch: false
    })

    await vitest.close()
  } catch (error) {
    process.exit(1)
  }
}

runProgrammaticTests()
```

https://vitest.dev/advanced/api/#startvitest

### Plugin Checklist

- [x] Integration tests.

### DIFF
--- a/integration-tests/vitest/vitest.spec.js
+++ b/integration-tests/vitest/vitest.spec.js
@@ -375,7 +375,7 @@ versions.forEach((version) => {
           const testSuite = events.find(event => event.type === 'test_suite_end').content
           assert.equal(test.meta[TEST_CODE_OWNERS], JSON.stringify(['@datadog-dd-trace-js']))
           assert.equal(testSuite.meta[TEST_CODE_OWNERS], JSON.stringify(['@datadog-dd-trace-js']))
-        })
+        }, 25000)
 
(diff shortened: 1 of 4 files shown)


integration-tests/vitest/vitest.spec.js
```diff
@@ -375,15 +375,15 @@ versions.forEach((version) => {
           const testSuite = events.find(event => event.type === 'test_suite_end').content
           assert.equal(test.meta[TEST_CODE_OWNERS], JSON.stringify(['@datadog-dd-trace-js']))
           assert.equal(testSuite.meta[TEST_CODE_OWNERS], JSON.stringify(['@datadog-dd-trace-js']))
-        })
+        }, 25000)
```
juan-fernandez, line 378: unrelated change: this test is a bit slow so we need to wait for the data to arrive

PR #6463, opened 2025-09-17
[test optimization] More robust cypress test

juan-fernandez:
### What does this PR do?

* Initialize `testOutput` to `''`, otherwise we might be calling `assert.include` or `assert.notInclude` with undefined, which throws an error before the `try/catch`. 
* Remove the race condition where `'exit'` handler is called before the stdout/stderr streams have finished 

### Motivation
Reduce flakiness

### DIFF
--- a/integration-tests/cypress/cypress.spec.js
+++ b/integration-tests/cypress/cypress.spec.js
@@ -195,7 +195,7 @@ moduleTypes.forEach(({
       })
     }
 
-    it('does not crash if badly init', (done) => {
+    it('does not crash if badly init', async () => {
       const {
         NODE_OPTIONS, // NODE_OPTIONS dd-trace config does not work with cypress
         DD_CIVISIBILITY_AGENTLESS_URL,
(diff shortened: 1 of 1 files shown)


integration-tests/cypress/cypress.spec.js
```diff
@@ -231,22 +230,27 @@ moduleTypes.forEach(({
...
 
-      childProcess.on('exit', () => {
+      await Promise.all([
+        once(childProcess.stdout, 'end'),
+        once(childProcess.stderr, 'end'),
+        once(childProcess, 'exit'),
+        eventsPromise
+      ])
+
+      assert.isFalse(hasReceivedEvents)
```
BridgeAR: Nit: please only use assert methods that are available by Node.js assert as well. That way a later migration becomes easier.

```suggestion
      assert.strictEqual(hasReceivedEvents, false)
```

BridgeAR (approved):
LGTM. Just left a nit as wish

juan-fernandez:
thanks, applied your suggestion @BridgeAR . Could you re-review?

PR #6465, opened 2025-09-18
[WIP] Add support for OpenTelemetry Logs API

mabdinur:
## What does this PR do?

This PR adds OpenTelemetry Logs support to dd-trace-js, enabling users to emit structured log data that integrates with the 

OpenTelemetry ecosystem. The implementation includes:
- OpenTelemetry Logs API Integration: Full compatibility with @opentelemetry/api-logs for creating and using loggers
- OTLP Export: Logs are exported via the OpenTelemetry Protocol (OTLP) over HTTP to any compatible backend
- Batch Processing: Efficient batch processing of log records with configurable batch sizes and timeouts
- Standard Configuration: Uses standard OpenTelemetry environment variables with fallback support
- Protobuf Serialization: Optimized data serialization using Protocol Buffers with JSON fallback

## Motivation

OpenTelemetry has become the standard for observability in the cloud-native ecosystem. While dd-trace-js already provides excellent tracing capabilities, adding OpenTelemetry Logs support enables users to:
- Unified Observability: Use a single library for both tracing and logging with consistent data formats
- Ecosystem Compatibility: Integrate with existing OpenTelemetry infrastructure and tooling
- Standard Compliance: Follow OpenTelemetry specifications for maximum interoperability
- Vendor Flexibility: Export logs to any OTLP-compatible backend (Jaeger, Zipkin, custom collectors, etc.)

This implementation allows users to leverage their existing OpenTelemetry investments while maintaining the ease of use and performance characteristics of dd-trace-js.

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/dd-trace/src/opentelemetry/logs/batch_log_processor.js
```diff
@@ -0,0 +1,138 @@
...
+ * Based on OTLP Protocol v1.7.0.
+ */
+
+const log = require('../../log')
+
+/**
+ * BatchLogRecordProcessor processes log records in batches for efficient export.
+ *
+ * This implementation follows the OpenTelemetry JavaScript SDK BatchLogRecordProcessor:
+ * https://open-telemetry.github.io/opentelemetry-js/classes/_opentelemetry_sdk-logs.BatchLogRecordProcessor.html
```
mabdinur, line 12:
```suggestion
 * https://open-telemetry.github.io/opentelemetry-js/classes/_opentelemetry_sdk-logs.BatchLogRecordProcessor.html
```

packages/dd-trace/src/opentelemetry/logs/logger.js
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+const { SeverityNumber } = require('@opentelemetry/api-logs')
+const { sanitizeAttributes } = require('@opentelemetry/core')
+
+/**
+ * Logger provides methods to emit log records.
+ *
+ * This implementation follows the OpenTelemetry JavaScript SDK Logger:
+ * https://open-telemetry.github.io/opentelemetry-js/classes/_opentelemetry_sdk-logs.Logger.html
```
mabdinur:
```suggestion
 * This implementation follows the OpenTelemetry JavaScript API Logger:
 * https://open-telemetry.github.io/opentelemetry-js/interfaces/_opentelemetry_api-logs.Logger.html
```

packages/dd-trace/src/config_defaults.js
```diff
@@ -119,6 +119,15 @@ module.exports = {
   isTestManagementEnabled: false,
   isImpactedTestsEnabled: false,
   logInjection: true,
+  otelLogsEnabled: false,
+  otelLogsUrl: 'http://localhost:4318/v1/logs',
+  otelLogsHeaders: {},
```
mabdinur:
```suggestion
  otelLogsHeaders: "",
```

packages/dd-trace/src/opentelemetry/logs/logger.js
```diff
@@ -0,0 +1,101 @@
...
+
+  _emitLog (severityText, severityNumber, message, attributes) {
+    this.emit({
+      severityText,
+      severityNumber,
+      body: message,
+      attributes,
+      timestamp: Date.now() * 1_000_000
+    })
+  }
```
mabdinur: Defer defining info/debug/etc. methods

packages/dd-trace/src/opentelemetry/logs/logger.js
```diff
@@ -0,0 +1,101 @@
...
+
+    if (logRecord.attributes) {
+      logRecord.attributes = sanitizeAttributes(logRecord.attributes)
+    }
+
+    this._loggerProvider._processor.onEmit(logRecord, this.instrumentationLibrary, this._getSpanContext(logRecord))
+  }
+
+  debug (message, attributes = {}) {
+    this._emitLog('DEBUG', SeverityNumber.DEBUG, message, attributes)
```
mabdinur: Turn this into private methods/propties

docs/API.md
```diff
@@ -379,6 +379,41 @@ The following attributes are available to override Datadog-specific options:
...
 
+<h3 id="opentelemetry-logs">OpenTelemetry Logs</h3>
+
+dd-trace-js includes experimental support for OpenTelemetry logs, designed as a drop-in replacement for the OpenTelemetry SDK. This support is primarily intended for logging libraries rather than direct user configuration. Enable it by setting `DD_LOGS_OTEL_ENABLED=true` and use the [OpenTelemetry Logs API](https://open-telemetry.github.io/opentelemetry-js/modules/_opentelemetry_api-logs.html) to emit structured log data:
+
+```javascript
+const tracer = require('dd-trace').init()
+const { logs } = require('@opentelemetry/api-logs')
+
+const logger = logs.getLogger('my-service', '1.0.0')
```
mabdinur, line 392: Put this in an express app or something

package.json
```diff
@@ -122,6 +122,7 @@
     "@datadog/wasm-js-rewriter": "4.0.1",
     "@isaacs/ttlcache": "^1.4.1",
     "@opentelemetry/api": ">=1.0.0 <1.10.0",
+    "@opentelemetry/api-logs": "*",
```
mabdinur: Restrict this range

packages/dd-trace/src/config.js
```diff
@@ -649,7 +650,19 @@ class Config {
       OTEL_RESOURCE_ATTRIBUTES,
       OTEL_SERVICE_NAME,
       OTEL_TRACES_SAMPLER,
-      OTEL_TRACES_SAMPLER_ARG
+      OTEL_TRACES_SAMPLER_ARG,
```
mabdinur, line 641: Add these to FPD

packages/dd-trace/src/config.js
```diff
@@ -663,6 +676,32 @@ class Config {
...
+    // OpenTelemetry logs configuration - processed after OTEL_RESOURCE_ATTRIBUTES
+    // Enable logs if DD_LOGS_OTEL_ENABLED is true
+    this._setBoolean(env, 'otelLogsEnabled', isTrue(DD_LOGS_OTEL_ENABLED))
+    // Set OpenTelemetry logs configuration with specific _LOGS_ vars taking precedence over generic _EXPORTERS_ vars
+    // Only set if there's a custom URL, otherwise let calc phase handle the default
+    const customOtelLogsUrl = OTEL_EXPORTER_OTLP_LOGS_ENDPOINT || OTEL_EXPORTERS_OTLP_ENDPOINT
+    if (customOtelLogsUrl) {
+      this._setString(env, 'otelLogsUrl', customOtelLogsUrl)
+    }
+    this._setString(env, 'otelLogsHeaders', OTEL_EXPORTER_OTLP_LOGS_HEADERS || OTEL_EXPORTERS_OTLP_HEADERS)
```
mabdinur: Merge these values

packages/dd-trace/src/config.js
```diff
@@ -663,6 +676,32 @@ class Config {
...
+    // Handle OTLP protocol with grpc warning
+    const requestedProtocol = OTEL_EXPORTER_OTLP_LOGS_PROTOCOL || OTEL_EXPORTER_OTLP_PROTOCOL
+    if (requestedProtocol === 'grpc') {
+      // eslint-disable-next-line no-console
+      console.warn('OTLP gRPC protocol is not supported for logs. ' +
```
mabdinur: Use internal logger here

packages/dd-trace/src/opentelemetry/logs/batch_log_processor.js
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * @class BatchLogRecordProcessor
+ */
+class BatchLogRecordProcessor {
+  /**
+   * Creates a new BatchLogRecordProcessor instance.
+   *
+   * @param {Object} exporter - Log processor for exporting batches to Datadog Agent
```
mabdinur: Avoid using object

packages/dd-trace/src/proxy.js
```diff
@@ -261,6 +265,62 @@ class Tracer extends NoopProxy {
     }
   }
 
+  _initializeOpenTelemetryLogs (config) {
+    try {
+      const { LoggerProvider, BatchLogRecordProcessor, OtlpHttpLogExporter } = require('./opentelemetry/logs')
+      const { logs } = require('@opentelemetry/api-logs')
+      const os = require('os')
```
mabdinur: Lazy assignment, using an outer let. 

```
logs ??= require('@opentelemetry/api-logs').logs
```

packages/dd-trace/src/proxy.js
```diff
@@ -261,6 +265,62 @@ class Tracer extends NoopProxy {
...
+
+      // Build resource attributes
+      const resourceAttributes = {
+        'service.name': config.service,
+        'service.version': config.version,
+        'deployment.environment': config.env
+      }
+
+      // Add all tracer tags (includes DD_TAGS, OTEL_RESOURCE_ATTRIBUTES, DD_TRACE_TAGS, etc.)
+      if (config.tags && Object.keys(config.tags).length > 0) {
```
mabdinur:
```suggestion
      if (config.tags ) {
```

packages/dd-trace/src/proxy.js
```diff
@@ -261,6 +265,62 @@ class Tracer extends NoopProxy {
...
+        resourceAttributes['host.name'] = os.hostname()
+      }
+
+      // Create OTLP exporter using resolved config values
+      const exporter = new OtlpHttpLogExporter(
+        config.otelLogsUrl,
+        config.otelLogsHeaders,
+        config.otelLogsTimeout,
+        config.otelLogsProtocol,
+        resourceAttributes
```
mabdinur: Remove this, this should be passed down by log provider

packages/dd-trace/src/proxy.js
```diff
@@ -261,6 +265,62 @@ class Tracer extends NoopProxy {
     }
   }
 
+  _initializeOpenTelemetryLogs (config) {
+    try {
```
mabdinur: Remove try catch

packages/dd-trace/test/opentelemetry/logs.spec.js
```diff
@@ -0,0 +1,412 @@
...
+
+    it('should handle logger provider shutdown', () => {
+      const { loggerProvider } = setupTracerAndLogger(true)
+
+      expect(loggerProvider).to.exist
+      expect(typeof loggerProvider.shutdown).to.equal('function')
+
+      expect(() => {
+        loggerProvider.shutdown()
+      }).to.not.throw()
```
mabdinur: Useless expect

packages/dd-trace/test/opentelemetry/logs.spec.js
```diff
@@ -0,0 +1,412 @@
...
+    })
+
+    it('should work when disabled', () => {
+      const { logger } = setupTracerAndLogger(false)
+
+      expect(() => {
```
mabdinur: Use assert instead of expect

packages/dd-trace/test/opentelemetry/logs.spec.js
```diff
@@ -0,0 +1,412 @@
...
+        spanContext: () => ({
+          traceId: '1234567890abcdef1234567890abcdef',
+          spanId: '1234567890abcdef'
+        })
+      })
+
+      // Capture log records
+      const logRecords = []
+      loggerProvider._processor = {
+        onEmit: (record, instrumentationLibrary, spanContext) => {
```
mabdinur: Use sinon to do the mocking

packages/dd-trace/test/opentelemetry/logs.spec.js
```diff
@@ -0,0 +1,412 @@
...
+          delete process.env[key]
+        } else {
+          process.env[key] = value
+        }
+      })
+
+      let warningMessage = ''
+      if (expectedWarning) {
+        // eslint-disable-next-line no-console
+        const originalWarn = console.warn
```
mabdinur: We will remove console.warn

packages/dd-trace/test/opentelemetry/logs.spec.js
```diff
@@ -0,0 +1,412 @@
...
+        expect(config.hostname).to.be.a('string')
+        expect(config.hostname.length).to.be.greaterThan(0)
+      })
+    })
+  })
+
+  describe('Telemetry Metrics', () => {
+    // Helper function to test telemetry metrics for a given protocol
+    function testTelemetryMetricsForProtocol (protocol, expectedEncoding) {
+      require('../setup/core') // For sinon-chai
```
mabdinur: Do this at the top

packages/dd-trace/test/opentelemetry/logs.spec.js
```diff
@@ -0,0 +1,412 @@
...
+        body: 'Test message',
+        severityNumber: 9,
+        severityText: 'INFO',
+        attributes: { 'test.attr': 'test-value' },
+        timestamp: Date.now() * 1000000,
+        instrumentationLibrary: { name: 'test-service', version: '1.0.0' }
+      }]
+
+      const result = JSON.parse(transformer.transformLogRecords(logRecords).toString())
+      expect(result).to.have.property('resourceLogs')
```
mabdinur: partial deep equal, use a helper for this (assertObjectContains)

packages/dd-trace/test/opentelemetry/logs.spec.js
```diff
@@ -0,0 +1,412 @@
...
+      process.env.DD_AGENT_HOST = 'default-agent.example.com'
+
+      const config = new Config()
+      expect(config.otelLogsUrl).to.equal('http://default-agent.example.com:4318/v1/logs')
+    })
+
+    it('should use fallback default when no environment variables and no DD_AGENT_HOST', () => {
+      const Config = require('../../src/config')
+
+      delete process.env.OTEL_EXPORTER_OTLP_LOGS_ENDPOINT
```
mabdinur: Move this to before and after each call. DO not delete in test

packages/dd-trace/test/opentelemetry/protobuf.spec.js
```diff
@@ -0,0 +1,205 @@
...
+      {
+        body: { nested: { value: 'object' } },
+        severityNumber: SeverityNumber.INFO
+      }
+    ]
+
+    const result = transformer.transformLogRecords(logRecords)
+
+    expect(result).to.be.instanceOf(Buffer)
+    expect(result.length).to.be.greaterThan(0)
```
mabdinur: Or logRecords length
```suggestion
    expect(result.length).to.be.equal(4)
```

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
```diff
@@ -0,0 +1,151 @@
...
+    this._protocol = protocol
+
+    // Set Content-Type based on protocol
+    const contentType = this._protocol === 'http/json'
+      ? 'application/json'
+      : 'application/x-protobuf'
+
+    this._headers = {
+      'Content-Type': contentType,
+      ...this._parseAdditionalHeaders(headers)
```
mabdinur: We should do the merge of logs and non-logs header configs here

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+    this._timeout = timeout
+    this._transformer = new OtlpTransformer(resource, protocol)
+
+    // Pre-compute telemetry tags for efficiency
+    this._telemetryTags = [
+      `protocol:${this._protocol.startsWith('grpc') ? 'grpc' : 'http'}`,
```
mabdinur:
```suggestion
      'protocol:http',
```

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
```diff
@@ -0,0 +1,151 @@
...
+      resultCallback({ code: 1, error })
+    }
+  }
+
+  _parseAdditionalHeaders (headersString) {
+    if (!headersString || typeof headersString !== 'string') {
+      return {}
+    }
+
+    return Object.fromEntries(
```
mabdinur: This is not efficient

mabdinur: Use existing helpers and avoid intermediate states. Use tagger or an equivalent

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
```diff
@@ -0,0 +1,151 @@
...
+        .split(',')
+        .map(pair => pair.trim().split('='))
+        .filter(([key, value]) => key && value)
+        .map(([key, value]) => [key.trim(), value.trim()])
+    )
+  }
+
+  _sendPayload (payload, resultCallback) {
+    const url = new URL(this._url)
+    const isHttps = url.protocol === 'https:'
```
mabdinur: We do not support https remove this

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
```diff
@@ -0,0 +1,151 @@
...
+    const client = isHttps ? https : http
+
+    const options = {
+      hostname: url.hostname,
+      port: url.port || (isHttps ? 443 : 4318),
```
mabdinur:
```suggestion
      port: url.port,
```

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
```diff
@@ -0,0 +1,151 @@
...
+      path: url.pathname + url.search,
```
mabdinur:
```suggestion
      path: url.pathname,
```

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
```diff
@@ -0,0 +1,151 @@
...
+      resultCallback({ code: 1, error })
+    })
+
+    req.on('timeout', () => {
+      req.destroy()
+      const error = new Error('Request timeout')
+      resultCallback({ code: 1, error })
+    })
+
+    req.write(payload)
```
mabdinur, line 124: Use: undici

packages/dd-trace/src/opentelemetry/logs/protobuf_loader.js
```diff
@@ -0,0 +1,70 @@
...
+let _scopeLogs = null
+let _logRecord = null
+let _severityNumber = null
+
+function loadProtobufDefinitions () {
+  if (_root) {
+    return { _root, _logsService, _resourceLogs, _scopeLogs, _logRecord, _severityNumber }
+  }
+
+  try {
```
mabdinur: Remove try catch

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    // Serialize to protobuf
+    const message = _logsService.create(logsData)
+    const buffer = _logsService.encode(message).finish()
+
+    return buffer
+  }
+
+  _transformToJson (logRecords) {
```
mabdinur: Make this private:

```suggestion
  #transformToJson (logRecords) {
```

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,233 @@
...
+      spanId: this._hexToBytes(logRecord.spanId || '')
+    }
+  }
+
+  _mapSeverityNumber (severityNumber) {
+    const { _severityNumber } = this._getProtobufTypes()
+
+    if (!_severityNumber) {
+      // eslint-disable-next-line no-console
+      console.error('_severityNumber is undefined')
```
mabdinur: Get rid of this

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,233 @@
...
+    map[SeverityNumber.ERROR3] = severityEnum.values.SEVERITY_NUMBER_ERROR3
+    map[SeverityNumber.ERROR4] = severityEnum.values.SEVERITY_NUMBER_ERROR4
+    map[SeverityNumber.FATAL] = severityEnum.values.SEVERITY_NUMBER_FATAL
+    map[SeverityNumber.FATAL2] = severityEnum.values.SEVERITY_NUMBER_FATAL2
+    map[SeverityNumber.FATAL3] = severityEnum.values.SEVERITY_NUMBER_FATAL3
+    map[SeverityNumber.FATAL4] = severityEnum.values.SEVERITY_NUMBER_FATAL4
+    return map
+  }
+
+  _hexToBytes (hexString) {
```
mabdinur:
```suggestion
  _hexToBytes (hexString = '') {
```

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,233 @@
...
+      timeUnixNano: timestamp,
+      observedTimeUnixNano: timestamp,
+      severityNumber: this._mapSeverityNumber(logRecord.severityNumber || SeverityNumber.INFO),
+      severityText: logRecord.severityText || 'INFO',
+      body: this._transformBody(logRecord.body),
+      attributes: this._transformAttributes(logRecord.attributes),
+      droppedAttributesCount: 0,
+      flags: logRecord.flags || 0,
+      traceId: this._hexToBytes(logRecord.traceId || ''),
+      spanId: this._hexToBytes(logRecord.spanId || '')
```
mabdinur:
```suggestion
      traceId: this._hexToBytes(logRecord.traceId),
      spanId: this._hexToBytes(logRecord.spanId)
```

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,233 @@
...
+    map[SeverityNumber.WARN4] = severityEnum.values.SEVERITY_NUMBER_WARN4
+    map[SeverityNumber.ERROR] = severityEnum.values.SEVERITY_NUMBER_ERROR
+    map[SeverityNumber.ERROR2] = severityEnum.values.SEVERITY_NUMBER_ERROR2
+    map[SeverityNumber.ERROR3] = severityEnum.values.SEVERITY_NUMBER_ERROR3
+    map[SeverityNumber.ERROR4] = severityEnum.values.SEVERITY_NUMBER_ERROR4
+    map[SeverityNumber.FATAL] = severityEnum.values.SEVERITY_NUMBER_FATAL
+    map[SeverityNumber.FATAL2] = severityEnum.values.SEVERITY_NUMBER_FATAL2
+    map[SeverityNumber.FATAL3] = severityEnum.values.SEVERITY_NUMBER_FATAL3
+    map[SeverityNumber.FATAL4] = severityEnum.values.SEVERITY_NUMBER_FATAL4
+    return map
```
mabdinur: Init map with these values

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,233 @@
...
+          }))
+        }
+      }
+    }
+    return {
+      stringValue: String(body)
+    }
+  }
+
+  _transformAttributes (attributes) {
```
mabdinur: Add doc strings that link to otel implementations with type defs

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,233 @@
...
+    if (!attributes) {
+      return {}
+    }
+    return Object.entries(attributes).map(([key, value]) => ({
+      key,
+      value: this._transformAnyValue(value)
+    }))
+  }
+
+  _transformAnyValue (value) {
```
mabdinur: Loop in Zach on overhead of serializing otlp payloads, v2 can use msgpack + string table

packages/dd-trace/src/plugin_manager.js
```diff
@@ -179,6 +180,7 @@ module.exports = class PluginManager {
       clientIpHeader,
       ciVisibilityTestSessionName,
       ciVisAgentlessLogSubmissionEnabled,
+      otelLogsEnabled,
```
mabdinur: Do we need to define otelLogsEnabled here? (probably no)

packages/dd-trace/src/opentelemetry/logs/logger.js
```diff
@@ -0,0 +1,101 @@
...
+      return
+    }
+
+    if (!this._loggerProvider._processor) {
+      return
+    }
+
+    if (logRecord.attributes) {
+      logRecord.attributes = sanitizeAttributes(logRecord.attributes)
+    }
```
mabdinur, line 61: Set timestamp here: timestamp: Date.now() * 1_000_000

packages/dd-trace/src/opentelemetry/logs/logger_provider.js
```diff
@@ -0,0 +1,115 @@
...
+   * Creates a new LoggerProvider instance with a single processor for Datadog Agent export.
+   *
+   * @param {Object} [options] - LoggerProvider options
+   * @param {Object} [options.resource] - Resource attributes
+   * @param {Object} [options.resource.attributes] - Resource attribute key-value pairs
+   * @param {Object} [options.processor] - Single LogRecordProcessor instance for exporting logs to Datadog Agent
+   */
+  constructor (options = {}) {
+    this.resource = options.resource
+    this._processor = options.processor || null
```
mabdinur:
```suggestion
    this._processor = options.processor
```

packages/dd-trace/src/opentelemetry/logs/logger.js
```diff
@@ -0,0 +1,101 @@
...
+   * @param {LoggerProvider} loggerProvider - Parent logger provider
+   * @param {Object} [instrumentationLibrary] - Instrumentation library information
+   * @param {string} [instrumentationLibrary.name] - Library name (defaults to 'dd-trace-js')
+   * @param {string} [instrumentationLibrary.version] - Library version (defaults to tracer version)
+   */
+  constructor (loggerProvider, instrumentationLibrary) {
+    this._loggerProvider = loggerProvider
+    this.instrumentationLibrary = {
+      name: instrumentationLibrary?.name || 'dd-trace-js',
+      version: instrumentationLibrary?.version || require('../../../../../package.json').version
```
mabdinur: Move require to top of the file

packages/dd-trace/src/opentelemetry/logs/logger.js
```diff
@@ -0,0 +1,101 @@
...
+      attributes,
+      timestamp: Date.now() * 1_000_000
+    })
+  }
+
+  _getSpanContext (logRecord) {
+    const activeSpan = trace.getSpan(logRecord.context || context.active())
+    if (activeSpan) {
+      const spanContext = activeSpan.spanContext()
+      if (spanContext && spanContext.traceId && spanContext.spanId) {
```
mabdinur:
```suggestion
      if (spanContext?.traceId && spanContext.spanId) {
```

packages/dd-trace/src/config.js
mabdinur: We can do this in a follow up.

packages/dd-trace/test/opentelemetry/protobuf.spec.js
mabdinur: Removed test

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
mabdinur: Skip this for now. It adds too much complexity

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,342 @@
...
+    const grouped = new Map()
+
+    for (const record of logRecords) {
+      const instrumentationLibrary = record.instrumentationLibrary || { name: '', version: '0.0.0' }
+      const key = `${instrumentationLibrary.name}@${instrumentationLibrary.version}`
+
+      if (!grouped.has(key)) {
+        grouped.set(key, [])
+      }
+      grouped.get(key).push(record)
```
BridgeAR: Perf nit

```suggestion
      let group = grouped.get(key)
      if (group === undefined) {
        grouped.set(key, [record])
      } else {
        group.push(record)
      }
```

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,342 @@
...
+    // Delay the loading of protobuf types to reduce startup overhead
+    if (!this.#protobufTypes) {
+      this.#protobufTypes = getProtobufTypes()
+    }
+    return this.#protobufTypes
+  }
+
+  /**
+   * Transforms log records to protobuf format.
+   * @param {Object[]} logRecords - Array of enriched log records to transform
```
BridgeAR: Do we know the exact type by chance? :)

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,342 @@
...
+      }
+      grouped.get(key).push(record)
+    }
+
+    return grouped
+  }
+
+  /**
+   * Gets the protobuf types, loading them lazily to reduce startup overhead.
+   * @returns {Object} Protobuf types object
```
BridgeAR: Do we know what shape that has? :)

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,342 @@
...
+   * @private
+   */
+  #transformToProtobuf (logRecords) {
+    const { _logsService } = this.#getProtobufTypes()
+
+    // Group log records by instrumentation library
+    const groupedRecords = this.#groupByInstrumentationLibrary(logRecords)
+
+    // Create scope logs for each instrumentation library
+    const scopeLogs = [...groupedRecords.entries()].map(([key, records]) => ({
```
BridgeAR: This code is readable, I would just still use a `for ... of` loop instead and allocate an empty array up front that is pushed too. That way there is no need to allocate the array twice.

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,342 @@
...
+    // Serialize to protobuf
+    const message = _logsService.create(logsData)
+    const buffer = _logsService.encode(message).finish()
+
+    return buffer
+  }
+
+  /**
+   * Transforms log records to JSON format.
+   * @param {Object[]} logRecords - Array of enriched log records to transform
```
BridgeAR: Same about the types everywhere

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,342 @@
...
+   * @returns {Buffer} JSON-encoded log records
+   * @private
+   */
+  #transformToJson (logRecords) {
+    // Group log records by instrumentation library
+    const groupedRecords = this.#groupByInstrumentationLibrary(logRecords)
+
+    // Create scope logs for each instrumentation library
+    const scopeLogs = [...groupedRecords.entries()].map(([key, records]) => ({
```
BridgeAR: Same about the array as above

packages/dd-trace/src/opentelemetry/logs/protobuf_loader.js
```diff
@@ -0,0 +1,65 @@
...
+  _logRecord = _root.lookupType('opentelemetry.proto.logs.v1.LogRecord')
+  _severityNumber = _root.lookupEnum('opentelemetry.proto.logs.v1.SeverityNumber')
+
+  return { _root, _logsService, _resourceLogs, _scopeLogs, _logRecord, _severityNumber }
+}
+
+// Lazy load the protobuf definitions
+function getProtobufTypes () {
+  return loadProtobufDefinitions()
+}
```
BridgeAR: Is that not identical to just not calling the method where ever it is used? It seems confusing for me that we have a method that is just w wrapper around the loader.

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
```diff
@@ -0,0 +1,159 @@
...
+      return {}
+    }
+
+    const headers = {}
+    for (const pair of headersString.split(',')) {
+      const [key, value] = pair.trim().split('=')
+      if (key && value) {
+        headers[key.trim()] = value.trim()
+      }
+    }
```
BridgeAR: We could use a single pass algorithm:

```js
function parseHeaders(headersString) {
  const headers = {}
  let key = ''
  let value = ''
  let readingKey = true

  for (const char of headersString) {
    if (readingKey) {
      if (char === '=') {
        readingKey = false
        key = key.trim()
      } else {
        key += char
      }
    } else if (char === ',') {
      value = value.trim()
      if (key && value) {
        headers[] = value
      }
      key = ''
      value = ''
      readingKey = true
    } else {
      value += char
    }
  }

  // Add the last pair if present
  if (!readingKey) {
    value = value.trim()
    if (value) {
      headers[key] = value
    }
  }

  return headers
}
```

packages/dd-trace/src/opentelemetry/logs/batch_log_processor.js
```diff
@@ -0,0 +1,137 @@
...
+      this.#clearTimer()
+
+      this.#export()
+
+      const shutdownPromises = this.exporter ? [this.exporter.shutdown()] : []
+
+      Promise.all(shutdownPromises).then(resolve)
+    })
+
+    return this.#shutdownPromise
```
BridgeAR:
```suggestion
    this.#clearTimer()

    this.#export()

    this.#shutdownPromise = this.exporter ? Promise.all([this.exporter.shutdown()]) : Promise.resolve([])

    return this.#shutdownPromise
```

packages/dd-trace/src/opentelemetry/logs/batch_log_processor.js
```diff
@@ -0,0 +1,137 @@
...
+  forceFlush () {
+    return new Promise((resolve) => {
+      if (this.isShutdown) {
+        resolve()
+        return
+      }
+
+      this.#export()
+      resolve()
+    })
```
BridgeAR: Is this required to be async?

```suggestion
    if (!this.isShutdown) {
      this.#export()
    }

    return Promise.resolve()
```

packages/dd-trace/src/config.js
```diff
@@ -1133,7 +1164,13 @@ class Config {
       calc.testManagementAttemptToFixRetries = maybeInt(DD_TEST_MANAGEMENT_ATTEMPT_TO_FIX_RETRIES) ?? 20
       this._setBoolean(calc, 'isImpactedTestsEnabled', !isFalse(DD_CIVISIBILITY_IMPACTED_TESTS_DETECTION_ENABLED))
     }
+
     calc['dogstatsd.hostname'] = this._getHostname()
+
+    // Compute OTLP logs URL to send payloads to the active Datadog Agent
+    const agentHostname = this._getHostname()
+    calc.otelLogsUrl = `http://${agentHostname}:4318/v1/logs`
```
BridgeAR: Is there port always fix?

BridgeAR (commented):
This locks quite good so far. I left a few suggestions. I suggest we have one more call to go through everything and land it during the call.

packages/dd-trace/src/config.js
mabdinur: The default port is always fixed. The only way to provide a new value is to set OTEL_EXPORTER_OTLP_LOGS_ENDPOINT or OTEL_EXPORTER_OTLP_ENDPOINT configurations

packages/dd-trace/src/opentelemetry/logs/batch_log_processor.js
mabdinur: I thought async would be optimal but honestly it doesn't make sense here. ForceFlush is mainly used in testing (system-tests), I doubt users will ever call this directly. I will make this change

packages/dd-trace/src/opentelemetry/logs/batch_log_processor.js
mabdinur: I also removed `#shutdownPromise`. Since this method is now synchronous it's not needed

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
mabdinur: This is an internal type, that's only initialized in Logger.js. I will define it

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
mabdinur: I consolidated this array with the one above. We can generate scope logs in one place for both json and protobuf.

packages/dd-trace/src/opentelemetry/logs/protobuf_loader.js
mabdinur: I missed this. I refactored getProtobufTypes and now it simply calls loadProtobufDefinitions, I should've just removed it 🙃

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
mabdinur: I used the type from the OpenTelemetry Logs API. We can avoid defining our own internal type.

packages/dd-trace/test/opentelemetry/logs.spec.js
```diff
@@ -0,0 +1,390 @@
...
+      assert.strictEqual(loggerProvider.processor.exporter.transformer.protocol, 'http/protobuf')
+      assert(logMock.getMessage().includes('OTLP gRPC protocol is not supported'))
+
+      logMock.restore()
+    })
+
+    it('configures OTLP endpoint from environment variable', () => {
+      process.env.OTEL_EXPORTER_OTLP_LOGS_ENDPOINT = 'http://custom:4318/v1/logs'
+      const { loggerProvider } = setupTracer()
+      assert.strictEqual(loggerProvider.processor.exporter.url, 'http://custom:4318/v1/logs')
```
mabdinur:
```suggestion
      process.env.OTEL_EXPORTER_OTLP_LOGS_ENDPOINT = 'http://custom:4321/v2/logs'
      const { loggerProvider } = setupTracer()
      assert.strictEqual(loggerProvider.processor.exporter.url, 'http://custom:4321/v2/logs')
```

packages/dd-trace/src/opentelemetry/logs/batch_log_processor.js
```diff
@@ -0,0 +1,129 @@
...
+
+  /**
+   * Forces an immediate flush of all pending log records.
+   * @returns {Promise<void>} Promise that resolves when flush is complete
+   */
+  forceFlush () {
+    if (!this.isShutdown) {
+      this.#export()
+    }
+    return Promise.resolve()
```
mabdinur:
```suggestion
   * @returns {undefined} Promise that resolves when flush is complete
   */
  forceFlush () {
    if (!this.isShutdown) {
      this.#export()
    }
```

packages/dd-trace/src/opentelemetry/logs/batch_log_processor.js
```diff
@@ -0,0 +1,129 @@
...
+   */
+  shutdown () {
+    if (this.isShutdown) {
+      return Promise.resolve()
+    }
+
+    this.isShutdown = true
+    this.#clearTimer()
+    this.#export()
+    return this.exporter ? this.exporter.shutdown() : Promise.resolve()
```
mabdinur:
```suggestion
   * @returns {undefined} Promise that resolves when shutdown is complete
   */
  shutdown () {
    if (this.isShutdown) {
      this.isShutdown = true
      this.#clearTimer()
      this.#export()
      this.exporter.shutdown()
    }
```

packages/dd-trace/src/opentelemetry/logs/batch_log_processor.js
```diff
@@ -0,0 +1,129 @@
...
+  /**
+   * Exports the current batch of log records.
+   * @private
+   */
+  #export () {
+    if (this.#logRecords.length === 0) {
+      return
+    }
+
+    const logRecords = this.#logRecords.splice(0, this.#maxExportBatchSize)
```
mabdinur: Use slice twice
```suggestion
    const logRecords = this.#logRecords.slice(0, this.#maxExportBatchSize)
    this.#logRecords = this.#logRecords.slice(this.#maxExportBatchSize)
    
```

packages/dd-trace/src/opentelemetry/logs/logger.js
```diff
@@ -0,0 +1,89 @@
...
+   * @returns {void}
+   */
+  emit (logRecord) {
+    if (this.loggerProvider.isShutdown) {
+      return
+    }
+
+    if (!this.loggerProvider.processor) {
+      return
+    }
```
mabdinur:
```suggestion
    if (this.loggerProvider.isShutdown || !this.loggerProvider.processor) {
      return
    }
```

packages/dd-trace/src/opentelemetry/logs/logger.js
```diff
@@ -0,0 +1,89 @@
...
+
+    if (logRecord.attributes) {
+      logRecord.attributes = sanitizeAttributes(logRecord.attributes)
+    }
+
+    // Set default timestamp if not provided
+    if (!logRecord.timestamp) {
+      logRecord.timestamp = Date.now() * 1_000_000
```
mabdinur:
```suggestion
      logRecord.timestamp = Number(process.hrtime.bigint())
```

packages/dd-trace/src/opentelemetry/logs/logger_provider.js
```diff
@@ -0,0 +1,142 @@
...
+ * @implements {import('@opentelemetry/api-logs').LoggerProvider}
+ */
+class LoggerProvider {
+  #loggers
+  #contextManager
+
+  /**
+   * Creates a new LoggerProvider instance with a single processor for Datadog Agent export.
+   *
+   * @param {Object} [options] - LoggerProvider options
```
mabdinur, line 29: nit: Use otel type

packages/dd-trace/src/opentelemetry/logs/logger_provider.js
```diff
@@ -0,0 +1,142 @@
...
+    if (this.isShutdown) {
+      return Promise.reject(new Error('LoggerProvider is shutdown'))
+    }
+
+    if (!this.processor) {
+      return Promise.resolve()
+    }
+
+    return this.processor.forceFlush()
+  }
```
mabdinur:
```suggestion
   * @returns {undefined} Promise that resolves when flush is n ssue cncomplete
   */
  forceFlush () {
    if (this.isShutdown) {
      throw new Error('LoggerProvider is shutdown')
    }
    return this.processor?.forceFlush()
  }
```

packages/dd-trace/src/opentelemetry/logs/logger_provider.js
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    this.isShutdown = true
+
+    if (!this.processor) {
+      return Promise.resolve()
+    }
+
+    return this.processor.shutdown()
+  }
```
mabdinur, line 103:
```suggestion
  /**
   * Shuts down the logger provider and all associated processors.
   * @returns {undefined} Promise that resolves when shutdown is complete
   */
  shutdown () {
    if (!this.isShutdown) {
      this.isShutdown = true
      this.processor.shutdown()
    }
  }
```

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
```diff
@@ -0,0 +1,183 @@
...
+
+    const payload = this.transformer.transformLogRecords(logRecords)
+    this.#sendPayload(payload, resultCallback)
+    tracerMetrics.count('otel.log_records', this.#telemetryTags).inc(logRecords.length)
+  }
+
+  /**
+   * Shuts down the exporter.
+   * @returns {Promise<void>} Promise that resolves when shutdown is complete
```
mabdinur:
```suggestion
   * @returns {undefined} Promise that resolves when shutdown is complete
```

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
```diff
@@ -0,0 +1,183 @@
...
+   */
+  shutdown () {
+    return Promise.resolve()
```
mabdinur:
```suggestion
```

packages/dd-trace/src/opentelemetry/logs/protobuf_loader.js
```diff
@@ -0,0 +1,59 @@
...
+ * Reference:
+ * - https://github.com/open-telemetry/opentelemetry-proto (v1.7.0)
+ */
+
+const protobuf = require('protobufjs')
+const path = require('path')
+const fs = require('fs')
+
+let _root = null
+let _logsService = null
```
mabdinur:
```suggestion
let logsService = null
```

mabdinur: Can we do this?

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
mabdinur: Double check

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,387 @@
...
+   * @private
+   */
+  #transformScope (logRecords) {
+    // Group log records by instrumentation library
+    const groupedRecords = this.#groupByInstrumentationScope(logRecords)
+
+    // Create scope logs for each instrumentation library
+    const scopeLogs = []
+
+    for (const [, records] of groupedRecords.entries()) {
```
mabdinur:
```suggestion
    for (const records of groupedRecords.values()) {
```

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,387 @@
...
+  }
+
+  /**
+   * Transforms a single log record to OTLP format.
+   * @param {LogRecord} logRecord - Log record to transform
+   * @returns {Object} OTLP log record object
+   * @private
+   */
+  #transformLogRecord (logRecord) {
+    const timestamp = logRecord.timestamp || Date.now() * 1_000_000
```
mabdinur:
```suggestion
    const timestamp = logRecord.timestamp
```

packages/dd-trace/src/opentelemetry/logs/otlp_transformer.js
```diff
@@ -0,0 +1,387 @@
...
+  }
+
+  /**
+   * Maps OpenTelemetry severity number to protobuf severity number.
+   * @param {number} severityNumber - OpenTelemetry severity number
+   * @returns {number} Protobuf severity number
+   * @private
+   */
+  #mapSeverityNumber (severityNumber) {
+    const { _severityNumber } = this.#getProtobufTypes()
```
mabdinur: Not internal

packages/dd-trace/src/opentelemetry/logs/protobuf_loader.js
```diff
@@ -0,0 +1,59 @@
...
+let _resourceLogs = null
+let _scopeLogs = null
+let _logRecord = null
+let _severityNumber = null
+
+function getProtobufTypes () {
+  if (_root) {
+    return { _root, _logsService, _resourceLogs, _scopeLogs, _logRecord, _severityNumber }
```
mabdinur: Fix return to only return things we need

packages/dd-trace/src/opentelemetry/logs/logger.js
```diff
@@ -0,0 +1,85 @@
...
+      return
+    }
+
+    if (logRecord.attributes) {
+      logRecord.attributes = sanitizeAttributes(logRecord.attributes)
+    }
+
+    // Set default timestamp if not provided
+    if (!logRecord.timestamp) {
+      logRecord.timestamp = Number(process.hrtime.bigint())
```
BridgeAR, line 65:
```suggestion
      // The timestamp is handled in nanoseconds.
      logRecord.timestamp = Number(process.hrtime.bigint())
```

Is it possible to rename the property to include that information?

packages/dd-trace/src/opentelemetry/logs/logger_provider.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  /**
+   * Forces a flush of all pending log records.
+   * @returns {undefined} Promise that resolves when flush is n ssue cncomplete
+   */
+  forceFlush () {
+    if (this.isShutdown) {
+      throw new Error('LoggerProvider is shutdown')
```
BridgeAR: Looking at this again: I believe we should handle it gracefully (not throw an error) and instead just log a warning.

That way we don't interfere with customer code that does something wrong.

@rochdev PTAL

packages/dd-trace/src/opentelemetry/protos/protobuf_loader.js
```diff
@@ -0,0 +1,53 @@
...
+  const protoFiles = [
+    'common.proto',
+    'resource.proto',
+    'logs.proto',
+    'payload.proto'
+  ].map(file => path.join(protoDir, file))
+
+  if (!protoFiles.every(file => fs.existsSync(file))) {
+    throw new Error('Proto files not found')
+  }
```
BridgeAR: Looking at this again: I can not think of a circumstance where this could fail. Should we just remove the check here, since we vendored the files anyway?

BridgeAR (commented):
LGTM. I would just like to get another pair of eyes on it. I think this should be possible to land right afterwards.

packages/dd-trace/src/opentelemetry/logs/logger.js
mabdinur: This type is defined by otel. We can't modify it here: https://github.com/open-telemetry/opentelemetry-js/blob/bfede5741ce2744b2653cdbff633a6dec03f22f7/experimental/packages/api-logs/src/types/LogRecord.ts#L60

packages/dd-trace/src/opentelemetry/protos/protobuf_loader.js
mabdinur: Yup, it can only fail with a refactor but in that scenario the tests will fail. I will clean this up

packages/dd-trace/src/opentelemetry/logs/logger.js
BridgeAR: Makes sense. I thought it might just be good to add the comment for clarity :)

packages/dd-trace/src/proxy.js
```diff
@@ -267,6 +272,58 @@ class Tracer extends NoopProxy {
     }
   }
 
+  #initializeOpenTelemetryLogs (config) {
+    LoggerProvider ??= require('./opentelemetry/logs').LoggerProvider
+    BatchLogRecordProcessor ??= require('./opentelemetry/logs').BatchLogRecordProcessor
+    OtlpHttpLogExporter ??= require('./opentelemetry/logs').OtlpHttpLogExporter
+    logs ??= require('@opentelemetry/api-logs').logs
```
mabdinur:
```suggestion
```

packages/dd-trace/src/proxy.js
```diff
@@ -11,6 +11,7 @@ const nomenclature = require('./service-naming')
 const PluginManager = require('./plugin_manager')
 const NoopDogStatsDClient = require('./noop/dogstatsd')
 const { getEnvironmentVariable } = require('../../dd-trace/src/config-helper')
+let LoggerProvider, BatchLogRecordProcessor, OtlpHttpLogExporter, logs, os
```
mabdinur:
```suggestion
let LoggerProvider, BatchLogRecordProcessor, OtlpHttpLogExporter, os
```

packages/dd-trace/src/config.js
```diff
@@ -649,6 +660,23 @@ class Config {
...
+    if (OTEL_EXPORTER_OTLP_ENDPOINT) {
+      // Only set if there's a custom URL, otherwise let calc phase handle the default
+      this._setString(env, 'otelUrl', OTEL_EXPORTER_OTLP_ENDPOINT)
+    }
+    if (OTEL_EXPORTER_OTLP_ENDPOINT || OTEL_EXPORTER_OTLP_LOGS_ENDPOINT) {
+      this._setString(env, 'otelLogsUrl', OTEL_EXPORTER_OTLP_LOGS_ENDPOINT || env.otelUrl)
+    }
+    this._setString(env, 'otelHeaders', OTEL_EXPORTER_OTLP_HEADERS)
+    this._setString(env, 'otelLogsHeaders', OTEL_EXPORTER_OTLP_LOGS_HEADERS || env.otelHeaders)
+    this._setString(env, 'otelProtocol', OTEL_EXPORTER_OTLP_PROTOCOL)
```
khanayan123, line 676: Do we want to allow for programmatic configuration also?

packages/dd-trace/src/config.js
```diff
@@ -1154,7 +1182,20 @@ class Config {
       calc.testManagementAttemptToFixRetries = maybeInt(DD_TEST_MANAGEMENT_ATTEMPT_TO_FIX_RETRIES) ?? 20
       this._setBoolean(calc, 'isImpactedTestsEnabled', !isFalse(DD_CIVISIBILITY_IMPACTED_TESTS_DETECTION_ENABLED))
     }
+
+    // Disable log injection when OTEL logs are enabled
+    // OTEL logs and DD log injection are mutually exclusive
+    if (this._env.otelLogsEnabled) {
```
khanayan123, line 1190: Could we add a unit test for this scenario?

packages/dd-trace/src/config.js
```diff
@@ -1154,7 +1182,20 @@ class Config {
...
+      this._setBoolean(calc, 'logInjection', false)
+    }
+
     calc['dogstatsd.hostname'] = this._getHostname()
+
+    // Compute OTLP logs URL to send payloads to the active Datadog Agent
+    const agentHostname = this._getHostname()
+    calc.otelLogsUrl = `http://${agentHostname}:4318`
+    calc.otelUrl = `http://${agentHostname}:4318`
```
khanayan123: Could we add port 4318 as a constant?

packages/dd-trace/src/proxy.js
```diff
@@ -267,6 +272,57 @@ class Tracer extends NoopProxy {
     }
   }
 
+  #initializeOpenTelemetryLogs (config) {
```
khanayan123: NIT: could we do the initialization in ./opentelemetry/logs/index.js ?

khanayan123 (commented):
Overall LGTM, just a few small nits. Could we also add an integration test under ./integration-test to cover the client flow?

packages/dd-trace/src/config.js
mabdinur: I am indifferent. We can go with whatever is the best practice for nodejs. In python I did not add a programatic api for configs.

BridgeAR: I think it is better not to do that. I think we should in fact prevent adding programmatic options, if not absolutely having a good reason to do so.

PR #6467, opened 2025-09-18
`esbuild` support for IAST (cjs)

CarlesDD:
### What does this PR do?
Adds `esbuild` support for IAST product in cjs.

This PR adds a new step to `esbuild` dd-trace plugin in order to rewrite with IAST pass every application code file in order to get taint tracking propagation in the output file.

It also adds some adjustments to the process of obtaining the location of detected vulnerabilities in order to report the correct location. It should be noted that this will only be reported correctly if the `esbuild` `sourcemaps` flag is set; otherwise, the location of the first frame of the stack trace will be reported.

### Motivation
Customer request

### Plugin Checklist

- [x] Unit tests.
- [x] Integration tests.

### Additional Notes

[APPSEC-58310](https://datadoghq.atlassian.net/browse/APPSEC-58310)

[APPSEC-58310]: https://datadoghq.atlassian.net/browse/APPSEC-58310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/datadog-esbuild/index.js
+++ b/packages/datadog-esbuild/index.js
@@ -180,10 +180,6 @@ ${build.initialOptions.banner.js}`
       // It is local application code, not an instrumented package
       if (DEBUG) console.log(`APP: ${args.path}`, args)
 
-      const ext = path.extname(args.path).toLowerCase()
-      const isJs = /\.[cm]?[jt]sx?$/.test(ext)
-      if (!isJs && ext) return
-
       return {
         path: fullPathToModule,
         pluginData: {
@@ -282,6 +278,10 @@ ${build.initialOptions.banner.js}`
     }
 
     if (DD_IAST_ENABLED && args.pluginData?.applicationFile) {
+      const ext = path.extname(args.path).toLowerCase()
+      const isJs =/^\.(js|mjs|cjs)$/.test(ext)
+      if (!isJs) return
+
       if (DEBUG) console.log(`REWRITE: ${args.path}`)
       const fileCode = fs.readFileSync(args.path, 'utf8')
       const rewritten = rewriter.rewrite(fileCode, args.path, ['iast'])
--- a/packages/datadog-esbuild/index.js
+++ b/packages/datadog-esbuild/index.js
@@ -178,7 +178,7 @@ ${build.initialOptions.banner.js}`
 
     if (args.path.startsWith('.') && !args.importer.includes('node_modules/')) {
       // It is local application code, not an instrumented package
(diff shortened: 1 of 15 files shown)


packages/datadog-esbuild/index.js
```diff
@@ -62,10 +66,19 @@ function isESMBuild (build) {
 }
 
 module.exports.setup = function (build) {
+  const ddIastEnabled = build.initialOptions.define?.__DD_IAST_ENABLED__ === 'true'
```
CarlesDD: To set whether IAST is enabled or not, the esbuild `define` field has been used to avoid setting env vars on build time, leaving all configuration within the `esbuild` options.

packages/datadog-esbuild/index.js
```diff
@@ -192,12 +218,25 @@ ${build.initialOptions.banner.js}`
...
+      if (ddIastEnabled) {
+        if (DEBUG) console.log(`REWRITE: ${args.path}`)
+        const fileCode = fs.readFileSync(args.path, 'utf8')
+        const rewritten = rewriter.rewrite(fileCode, args.path, ['iast'])
+        return {
+          contents: rewritten.content,
+          loader: 'js',
+          resolveDir: path.dirname(args.path)
+        }
+      }
```
uurien:
```suggestion
    if (ddIastEnabled && args.pluginData?.applicationFile) {
      if (DEBUG) console.log(`REWRITE: ${args.path}`)
      const fileCode = fs.readFileSync(args.path, 'utf8')
      const rewritten = rewriter.rewrite(fileCode, args.path, ['iast'])
      return {
        contents: rewritten.content,
        loader: 'js',
        resolveDir: path.dirname(args.path)
      }
```

packages/datadog-esbuild/index.js
uurien: I can see that we are using env vars to read `DD_TRACE_DEBUG`, why against the usage of `DD_IAST_ENABLED` env var? Is that a bad practice?

CarlesDD: It was an effort to:
- Keep all the config inside esbuild options
- Try to avoid confusion with env vars in build/run time

That being said, I think changing the approach and using the env var will homogenize all the configuration (we are using tracer env var in build time, as you pointed out) and avoid side effects of esbuild define, which acts more as a macro than a config provied.

uurien (approved):

uurien (approved):

PR #6470, opened 2025-09-18
benchmark: make startup-test more stable by using fixed dep list

BridgeAR:
The dependency list is now a fixed subset of our currently installed dependencies. Ideally, that changes to an external set of dependencies that are installed before the benchmark is run. That way the benchmark is independent from what is currently used as actual dependencies.
This is just the first step to overcome regression reports when adding new dependencies.


bengl (approved):

rochdev:
@BridgeAR Isn't adding a new dependency an actual regression on startup time?

PR #6472, opened 2025-09-18
optimizing the test helper

crysmags:
### What does this PR do?
With flakey tests this will optimize how we create the integration test agent.

### DIFF
--- a/integration-tests/helpers/index.js
+++ b/integration-tests/helpers/index.js
@@ -214,12 +214,35 @@ function spawnProc (filename, options = {}, stdioHandler, stderrHandler) {
   }))
 }
 
+async function execHelper (command, options) {
+  /* eslint-disable no-console */
+  try {
+    console.log('Exec START: ', command)
+    await exec(command, options)
+    console.log('Exec SUCCESS: ', command)
+  } catch (error) {
+    console.error('Exec ERROR: ', command, error)
+    if (command.startsWith('yarn')) {
+      try {
+        console.log('Exec RETRY START: ', command)
+        await exec(command, options)
+        console.log('Exec RETRY SUCESS: ', command)
+      } catch (retryError) {
+        console.error('Exec RETRY ERROR', command, retryError)
+      }
+      throw error
+    }
+  }
+  /* eslint-enable no-console */
+}
+
 /**
  * @param {string[]} dependencies
  * @param {boolean} isGitRepo
  * @param {string[]} integrationTestsPaths
  * @param {string} [followUpCommand]
  */
+
 async function createSandbox (dependencies = [], isGitRepo = false,
   integrationTestsPaths = ['./integration-tests/*'], followUpCommand) {
   const cappedDependencies = dependencies.map(dep => {
@@ -241,8 +264,8 @@ async function createSandbox (dependencies = [], isGitRepo = false,
     // yarn-linked into dd-trace and want to run the integration tests against them.
 
     // Link dd-trace to itself, then...
-    await exec('yarn link')
-    await exec('yarn link dd-trace')
+    await execHelper('yarn link')
+    await execHelper('yarn link dd-trace')
     // ... run the tests in the current directory.
     return { folder: path.join(process.cwd(), 'integration-tests'), remove: async () => {} }
   }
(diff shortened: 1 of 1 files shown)


integration-tests/helpers/index.js
```diff
@@ -188,6 +188,23 @@ function spawnProc (filename, options = {}, stdioHandler, stderrHandler) {
   })
 }
 
+async function execHelper (command, options) {
```
rochdev, line 217: This could be called `exec` so that the rest of the code doesn't need to be updated or know about the helper. Not a blocker though.

rochdev: I just realized that this retries _everything_, which shouldn't be the case. The only command that should be able to be flaky is installing from the registry, retrying anything else could hide real issues.

rochdev: What I meant was that only yarn should be retried, but we should still log all commands

integration-tests/helpers/index.js
```diff
@@ -188,6 +188,23 @@ function spawnProc (filename, options = {}, stdioHandler, stderrHandler) {
...
+  try {
+    console.log('Commencing exec execution...', command) // eslint-disable-line no-console
+    await exec(command, options)
+    console.log('Success it ran!') // eslint-disable-line no-console
+  } catch (e) {
+    console.error('We caught, commencing retry, error:', // eslint-disable-line no-console
+      e, 'failed command:', command)
+    try {
+      await exec(command, options)
+      console.log('Success on the second try!') // eslint-disable-line no-console
```
BridgeAR: I guess it would be good to add the command to know what was successful, if multiple commands are run at the same time :) (the async calls could be intertwined).

That is probably not important anymore, since the command should now only be run once.

integration-tests/helpers/index.js
```diff
@@ -426,7 +439,7 @@ function setShouldKill (value) {
   })
 }
 
-// eslint-disable-next-line n/no-unsupported-features/node-builtins
+/* eslint-disable n/no-unsupported-features/node-builtins */
```
BridgeAR: I think that's an independent change we should not do? (Deactivating the rule in general could later on hide things)

BridgeAR (commented):
LGTM with the lintrule changed back (it passes on our main branch, so it should not need any adjustment here).

It just also needs a rebase.

integration-tests/helpers/index.js
```diff
@@ -189,6 +189,46 @@ function spawnProc (filename, options = {}, stdioHandler, stderrHandler) {
...
+    const [execResult, timeoutResult] = results
+
+    if (execResult.status === 'rejected') {
+      throw execResult.reason
+    }
+
+    if (timeoutResult.status === 'rejected') {
+      throw timeoutResult.reason
+    }
+  })
```
BridgeAR: I am not sure if I understand this correct.

If I read the code correct, it will wait for both promises to be settled, no matter how long the command takes and if the timeout already triggered. If it was longer than the timeout and the command succeeded, the command will still be considered failed due to the timeout promise being rejected. Thus, I don't understand why it would not be a `Promise.race`, since I don't understand in what way it is helpful to wait longer until we log the timeout.

crysmags: I used `Promise.allSettled` because when using `Promise.race` if the command timed out I would lose the error if the command reached an error after the timeout period. So I didn't want to lose the context of what happened in the command. Though I removed this and left the original helper function.

integration-tests/helpers/index.js
```diff
@@ -188,6 +188,34 @@ function spawnProc (filename, options = {}, stdioHandler, stderrHandler) {
...
+        throw e
+      }
+    }
+  } else {
+    try {
+      console.log('Commencing exec execution without retrying:', command) // eslint-disable-line no-console
+      await exec(command, options)
+    } catch (e) {
+      console.error('We caught an error on no retry', e) // eslint-disable-line no-console
+      throw e
```
BridgeAR:
```suggestion
  /* eslint-disable no-console */
  try {
    console.log('Exec START:', command)
    await exec(command, options)
    console.log('Exec SUCCESS:, command)
  } catch (error) {
    console.error('Exec ERROR:', command, error)
    if (command.startsWith('yarn')) {
      try {
        console.log('Exec RETRY START:', command)
        await exec(command, options)
        console.log('Exec RETRY SUCESS:', command)
      } catch (retryError) {
        console.error('Exec RETRY ERROR', command, retryError)
      }
      throw error
    }
  }
  /* eslint-enable no-console */
```

BridgeAR (commented):
I just made a suggestion to write it a bit simpler :)

I also wonder if we want to log the time while a command is running. I think that would be useful to see how long each command takes.

I would just start a interval that just logs the current time each second (starting from the time on the first exec is called and clearing the interval when the last one is done).

rochdev (approved):

PR #6475, opened 2025-09-19
change release proposal script exclusion label

rochdev:
### What does this PR do?

Change release proposal script exclusion label.

### Motivation

Support for the `dont-land` labels was removed a while ago in favour of `semver-major`. This turns out to be a problem because it impacts our ability to differentiate major changes by forcing us to use `semver-minor` for major changes. Instead we can switch back to `dont-land` labels when something needs to be excluded (pretty much only the version bump) and we can include semver major changes in minors because they are hidden behind a version conditional.

### DIFF
--- a/scripts/release/proposal.js
+++ b/scripts/release/proposal.js
@@ -66,26 +66,41 @@ try {
 
   pass(`v${releaseLine}.x`)
 
-  const diffCmd = 'branch-diff --user DataDog --repo dd-trace-js --exclude-label=semver-major'
+  // Semver-major changes are behind version conditionals so we can include them
+  // in minors.
+  const diffCmd = `branch-diff --user DataDog --repo dd-trace-js --exclude-label=dont-land-on-v${releaseLine}.x`
 
   start('Determine version increment')
 
   const { DD_MAJOR, DD_MINOR, DD_PATCH } = require('../../version')
-  const lineDiff = capture(`${diffCmd} --markdown=true v${releaseLine}.x ${main}`)
+  const lineDiffWithMajor = capture(`${diffCmd} --markdown=true v${releaseLine}.x ${main}`)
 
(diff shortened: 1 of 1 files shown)


rochdev:
@watson So you would keep `semver-minor` but with an added tag? Not a fan of `v6-only-changes` for a few reasons. First of all I don't think it's quite semantically correct as it doesn't necessarily include only changes for `v6` as you mentioned in your previous comment. Also, I don't think the version number needs to be explicit, major changes will always land in the next major, so maybe something like `semver-major-next`? Or do you think there actually is value in being explicit on the version?

rochdev:
> After the release of v6, we want to be able to land a PR containing changes both targeted v5 and v6.

Isn't that just a normal `semver-patch` or `semver-minor`?

watson:
> > After the release of v6, we want to be able to land a PR containing changes both targeted v5 and v6.

> Isn't that just a normal `semver-patch` or `semver-minor`?

To some extend yes, but that doesn't allow you to automatically know which release notes to update. Just the ones for v5, just the ones for v6, or both 🤷

watson:
> _Note: "landing" here means enabling the change since commits always land everywhere._

Understood 👍 My main concern here is also just about how we mange release notes. I'm aware that we technically merge the PR to all active release lines.

> Only landing on v6.x would mean it's a breaking change.

I don't agree. Say we land a breaking change in the `v6.x` branch prior to the release of v6 to npm. Then if there's a bug in this breaking change after the release of v6 to npm, or if we're adding a feature to that breaking change, it's could come in a PR that only affects the v6 release. And hence, the release notes for the next v5 release shouldn't include any mention of it.

Maybe all of this is more suitable for a Zoom meeting to avoid all this back and forth 😅

BridgeAR:
Marking as draft for the time being.

PR #6477, opened 2025-09-19
ci(llmobs): use plugin test handler for receiving both llm observability and apm spans

sabrenner:
### What does this PR do?
Updates the rest of the LLM Observability tests to use the recently added `useLlmObs` handler to allow getting both LLM Observability and APM span events.

### Motivation
Previously, all of these tests were stubbing the `LLMObsSpanWriter` and intercepting span events there, which isn't really a great practice, and instead we should have been using the same `agent.use`-type callbacks all along.

### DIFF
--- a/.github/actions/install/branch-diff/action.yml
+++ b/.github/actions/install/branch-diff/action.yml
@@ -7,7 +7,7 @@ inputs:
 runs:
   using: composite
   steps:
-    - uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
+    - uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
       with:
         path: ~/.npm
         key: ${{ github.workflow }}-branch-diff-3.1.1
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -31,6 +31,7 @@ updates:
       - "/"
     schedule:
       interval: "weekly"
+    open-pull-requests-limit: 100
     cooldown:
       default-days: 1
       exclude:
--- a/.github/workflows/all-green.yml
+++ b/.github/workflows/all-green.yml
@@ -19,7 +19,7 @@ jobs:
       checks: read
       contents: read
     steps:
-      - uses: wechuli/allcheckspassed@e22f45a4f25f4cf821d1273705ac233355400db1 # v1.2.0
+      - uses: wechuli/allcheckspassed@5b5009ffd707b7c6759574c984bb666034b441eb # v2.0.0
         with:
           retries: 30 # once per minute, some checks take up to 15 min, retries are possible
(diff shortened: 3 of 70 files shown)


packages/dd-trace/test/llmobs/util.js
```diff
@@ -252,14 +236,34 @@ function useLlmObs ({
...
+
+  beforeEach(() => {
+    resetTracesPromises()
   })
 
   after(() => {
-    return agent.close(closeOptions)
+    return agent.close(Object.assign({
+      ritmReset: false
+    }, closeOptions))
```
watson: Spread syntax is preferred over `Object.assign`:

```suggestion
    return agent.close({ ritmReset: false, ...closeOptions })
```

packages/dd-trace/test/llmobs/util.js
```diff
@@ -252,14 +236,34 @@ function useLlmObs ({
...
+    _DD_LLMOBS_FLUSH_INTERVAL: 0
+  })
+
+  before(() => {
+    return agent.load(plugin, {}, Object.assign({
+      llmobs: {
+        mlApp: 'test',
+        agentlessEnabled: false
+      }
+    }, tracerConfigOptions))
```
watson: Spread syntax is preferred over `Object.assign`:

```suggestion
    return agent.load(plugin, {}, {
      llmobs: { mlApp: 'test', agentlessEnabled: false },
      ...tracerConfigOptions
    })
```

packages/dd-trace/test/llmobs/util.js
```diff
@@ -252,14 +236,34 @@ function useLlmObs ({
...
+  })
+
+  beforeEach(() => {
+    resetTracesPromises()
   })
```
watson: Nit:

```suggestion
  beforeEach(resetTracesPromises)
```

packages/dd-trace/test/llmobs/plugins/langchain/index.spec.js
```diff
@@ -80,40 +78,23 @@ describe('integrations', () => {
   }
 
   describe('langchain', () => {
-    before(async () => {
-      sinon.stub(LLMObsSpanWriter.prototype, 'append')
-
+    before(() => {
```
watson: This `describe` block contains two inline `before` hooks + the ones added when `useLlmObs` is called. Is the order important? If not, I recommend merging the two inline into one.

sabrenner: oh nice yeah it looks like i can combine them. it was my assumption previously that they had to be in this order (`iastFilter` for orchestrion stuff before loading the tracer, then loading the tracer in `useLlmObs`, then grabbing the `llmobs` sdk after the tracer is loaded). but, changing it around locally also seems to work!

watson (approved):

PR #6486, opened 2025-09-22
feat(ai_guard): implement AI Guard SDK

manuel-alvarez-alvarez:
### What does this PR do?
This is the first version of the AI Guard JS SDK. It serves as a lightweight wrapper around the AI Guard public REST API, providing a more convenient and javascript oriented interface for interacting with the service.

More information about the API [here](https://github.com/DataDog/dd-source/tree/main/domains/appsec/apps/apis/ai-guard)

### Motivation
We have design partners interested in a javascript SDK to test AI Guard.

### Plugin Checklist

- [x] Unit tests.
- [x] Integration tests.
- [x] Benchmarks.
- [x] TypeScript [tests][2].
- [x] API [documentation][3].
- [x] CI [jobs/workflows][4].

### Additional Notes

[APPSEC-58936](https://datadoghq.atlassian.net/browse/APPSEC-58936)

[APPSEC-58936]: https://datadoghq.atlassian.net/browse/APPSEC-58936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


index.d.ts
```diff
@@ -139,6 +139,11 @@ interface Tracer extends opentracing.Tracer {
    */
   llmobs: tracer.llmobs.LLMObs;
 
+  /**
+   * AI Guard SDK
+   */
+  aiguard: tracer.aiguard.AIGuard;
```
uurien, line 145: The index.d.ts shoud be tested in `docs/test.ts`

index.d.ts
```diff
@@ -867,6 +872,20 @@ declare namespace tracer {
...
     llmobs?: llmobs.LLMObsEnableOptions
+
+    /**
+     * Configuration options for AI Guard
+     */
+    aiguard?: {
+      /**
+       * Set to `true` to enable the SDK.
+       */
+      enabled?: boolean,
```
uurien: Why do we need this property to enable it? This feature is about the sdk, why do we need to manually enable it to be able to use? 
If an application is using the SDK, it is automatically enabled.

packages/dd-trace/src/appsec/ai_guard/client.js
```diff
@@ -0,0 +1,50 @@
+'use strict'
+
+const http = require('http')
+const https = require('https')
+const { URL } = require('url')
+
+function executeRequest (url, headers, body) {
```
uurien: Can you use this method instead? https://github.dev/DataDog/dd-trace-js/blob/c9453261bed089e5b363974b73de0f08f58be740/packages/dd-trace/src/exporters/common/request.js#L36

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,130 @@
...
+      const message = 'AIGuard: requires protocol version 0.4'
+      log.error(message)
+      throw new Error(message)
+    }
+    this._evaluateUrl = `${config.aiguard.endpoint}/evaluate`
+    this._headers = {
+      'DD-API-KEY': config.apiKey,
+      'DD-APPLICATION-KEY': config.appKey,
+    }
+    this._meta = { service: config.service, env: config.env }
```
uurien: An empty line between blocks makes the code more readable.

```suggestion
    super(tracer)

    if (!config.aiguard.endpoint) {
      const message = 'AIGuard: missing endpoint, use env DD_AI_GUARD_ENDPOINT'
      log.error(message)
      throw new Error(message)
    }

    if (!config.apiKey) {
      const message = 'AIGuard: missing api key, use env DD_API_KEY'
      log.error(message)
      throw new Error(message)
    }

    if (!config.appKey) {
      const message = 'AIGuard: missing app key, use env DD_APP_KEY'
      log.error(message)
      throw new Error(message)
    }

    if (config.protocolVersion !== '0.4') {
      const message = 'AIGuard: requires protocol version 0.4'
      log.error(message)
      throw new Error(message)
    }

    this._evaluateUrl = `${config.aiguard.endpoint}/evaluate`
    this._headers = {
      'DD-API-KEY': config.apiKey,
      'DD-APPLICATION-KEY': config.appKey,
    }
    this._meta = { service: config.service, env: config.env }
```

packages/dd-trace/src/appsec/ai_guard/sdk.js
```diff
@@ -0,0 +1,130 @@
...
+    this._headers = {
+      'DD-API-KEY': config.apiKey,
+      'DD-APPLICATION-KEY': config.appKey,
+    }
+    this._meta = { service: config.service, env: config.env }
+  }
+
+  _truncate (messages) {
+    return messages
+  }
```
uurien: why? it is not extended, is it?

index.d.ts
```diff
@@ -1059,6 +1078,170 @@ declare namespace tracer {
...
+      /**
+       * Evaluates a conversation thread.
+       *
+       * @param messages - Array of conversation messages
+       * @param opts - Optional configuration object:
+       *   - `block`: When true, throws an exception if evaluation result is not 'ALLOW'
+       *              and the AI Guard service has blocking mode enabled.
+       * @returns Promise resolving to an Evaluation with the security decision and reasoning
+       * @throws AIGuardAbortError when `opts.block` is true and the evaluation result would block the request
+       * @throws AIGuardClientError when communication with the AI Guard service fails
```
uurien: These are really rejections, not `throws` if a user is doing something like this, catch never is going to be executed, because this would reject, and not throw.

```
try {
  tracer.aiguard.evaluate(...).then(() => { whatever })
} catch (e) {
  // if AIGuardAbortError block everything    
}
```

packages/dd-trace/src/appsec/ai_guard/sdk.js
manuel-alvarez-alvarez: No, the implementation is yet to be done (it requires truncating the messages and sending some telemetry metrics if done)

index.d.ts
manuel-alvarez-alvarez: Changed thanks!

manuel-alvarez-alvarez:
> Added some comments, also seems weird that the code is in `appsec` directory, but the SDK is out of `appsec` namespace.
> 
> IMO, both should be in appsec, or both should be out of appsec.

I've moved it outside appsec, thanks for the feedback!

index.d.ts
manuel-alvarez-alvarez: Done, thanks!

packages/dd-trace/src/appsec/ai_guard/client.js
manuel-alvarez-alvarez: Not quite — that method suppresses the response body whenever the status code indicates an error (<200 or ≥300). In our case, we need to parse the body as JSON since we’re adhering to the JSON:API specification.

packages/dd-trace/src/aiguard/sdk.js
manuel-alvarez-alvarez: Done

packages/dd-trace/src/appsec/ai_guard/sdk.js
manuel-alvarez-alvarez: Done

index.d.ts
manuel-alvarez-alvarez: In the future automatic instrumentations might come (e.g. AI SDK, Langchain, ...), thats the reason for the property.

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,175 @@
...
+      this.cause = opts.cause
+    }
+  }
+}
+
+class AIGuard extends NoopAIGuard {
+  constructor (tracer, config) {
+    super(tracer)
+
+    if (config.protocolVersion !== '0.4') {
```
smola: Is there a config switch to set the protocol? Might be good to mention in the error message if it's user-facing.

smola: Also I wonder whether we want to crash here, or just log an error like "AI Guard requires trace protocol version 0.4 (this can be set with DD...),observability of AI Guard results will not be available."
but not throw?

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,175 @@
...
+      const message = 'AIGuard: requires protocol version 0.4'
+      log.error(message)
+      throw new Error(message)
+    }
+
+    if (!config.aiguard?.endpoint) {
+      const message = 'AIGuard: missing endpoint, use env DD_AI_GUARD_ENDPOINT'
```
smola: I think that this point it's ok to have a default for the endpoint (based on `DD_SITE` to pick up the correct one).

packages/dd-trace/test/aiguard/index.spec.js
```diff
@@ -0,0 +1,283 @@
...
+    { role: 'system', content: 'You are a beautiful AI assistant' },
+    { role: 'user', content: 'What is 2 + 2' },
+    {
+      role: 'assistant',
+      tool_calls: [
+        {
+          id: 'call_1',
+          function: {
+            name: 'calc',
+            arguments: '{ operator: "+", args: [2, 2] }'
```
smola:
```suggestion
            arguments: '{ "operator": "+", "args": [2, 2] }'
```

Technically the above is not JSON (lacks quotes around fields), and so it's not what models like OpenAI would generate.
(Although AI Guard does not _mandate_ that this string is well-formed JSON).

smola (commented):
Public API looks good to me.
Added a few comments, primarily on error handling.

packages/dd-trace/src/aiguard/sdk.js
manuel-alvarez-alvarez: Changed, now only an error message remains, thanks!

packages/dd-trace/test/aiguard/index.spec.js
manuel-alvarez-alvarez: Yep, before I had arguments to be a proper json object instead of a string (that's why the quotes were missing). Changed!

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,172 @@
...
+      log.error('AIGuard: observability of evaluation results requires protocol version 0.4')
+    }
+
+    let endpoint = config.aiguard?.endpoint
+    if (!endpoint) {
+      endpoint = `https://app.${config.site}/api/v2/ai-guard`
+    }
+    this._evaluateUrl = new URL(`${endpoint}/evaluate`)
+
+    if (!config.apiKey || !config.appKey) {
```
IlyasShabi, line 52: This should be before defining the endpoint

manuel-alvarez-alvarez: Changed, thanks!

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,172 @@
...
+    const result = new Array(size)
+    for (let i = 0; i < size; i++) {
+      let message = messages[i]
+      if ('content' in message && message.content.length > this._maxContentSize) {
+        message = { ...message }
+        message.content = message.content.slice(0, this._maxContentSize)
+      }
+      result[i] = message
+    }
+    return result
```
IlyasShabi:
```suggestion
    const result = messages.slice(0, size)
    
    for (let i = 0; i < size; i++) {
      const message = result[i]
      if ('content' in message && message.content.length > this._maxContentSize) {
        result[i] = { ...message, content: message.content.slice(0, max) };
      }
    }
    return result
```

I think this is faster with slice instead of manual copy

packages/dd-trace/src/aiguard/client.js
```diff
@@ -0,0 +1,49 @@
+'use strict'
+
+const http = require('http')
+const https = require('https')
+
+function executeRequest (body, opts) {
```
IlyasShabi: If we are making so many calls to the same host we may need to add an agent with `keep-alive` this is faster since it's gonna reuse same sockets, otherwise dd-trace-js support Node.js 18+ so we can use native Fetch without extra config.
@BridgeAR Can you confirm this?

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,171 @@
...
+          }
+        }
+      }
+    }
+    return null
+  }
+
+  async evaluate (messages, opts) {
+    const { block = false } = opts ?? {}
+    return await this._tracer.trace(AI_GUARD_RESOURCE, {}, async (span) => {
```
IlyasShabi:
```suggestion
    return this._tracer.trace(AI_GUARD_RESOURCE, {}, async (span) => {
```

index.d.ts
```diff
@@ -1059,6 +1093,170 @@ declare namespace tracer {
...
+       *
+       * @param messages - Array of conversation messages
+       * @param opts - Optional configuration object:
+       *   - `block`: When true, throws an exception if evaluation result is not 'ALLOW'
+       *              and the AI Guard service has blocking mode enabled.
+       * @returns Promise resolving to an Evaluation with the security decision and reasoning.
+       *          The promise rejects with AIGuardAbortError when `opts.block` is true and the evaluation result would block the request.
+       *          The promise rejects with AIGuardClientError when communication with the AI Guard service fails.
+       */
+      evaluate(messages: Message[], opts?: { block?: boolean }): Promise<Evaluation>;
```
IlyasShabi: block default is false

packages/dd-trace/src/aiguard/sdk.js
manuel-alvarez-alvarez: Done, thanks!

packages/dd-trace/src/aiguard/sdk.js
manuel-alvarez-alvarez: Changed to return the promise and remove the `async` decl from the method.

index.d.ts
manuel-alvarez-alvarez: Added thanks!

packages/dd-trace/src/aiguard/client.js
manuel-alvarez-alvarez: Just for context: the remote API usually takes about 1–2 seconds to run the evaluation, so extra optimizations here are probably negligible and probably aren’t worth the trouble if they take much work.

index.d.ts
```diff
@@ -593,6 +598,17 @@ declare namespace tracer {
            */
           enabled?: boolean
         }
+      },
+
+      aiguard?: {
```
uurien, line 603: why are these options in experimental and others out of experimental?
If these things are to configure aiguard, I think that we could add these in main `aiguard` config options, out of experimental, when we want to remove things from experimental is always a pain because it is a breaking change.

index.d.ts
```diff
@@ -907,7 +941,7 @@ declare namespace tracer {
     scope?: string,
 
     /**
-     * Custom fields to attach to the user (RBAC, Oauth, etc…).
+     * Custom fields to attach to the user (RBAC, Oauth, etc...).
```
uurien, line 938: accidental change?

packages/dd-trace/src/aiguard/client.js
```diff
@@ -0,0 +1,50 @@
+'use strict'
+
+const http = require('http')
+const https = require('https')
+const { URL } = require('url')
+
+function executeRequest (body, opts) {
+  return new Promise((resolve, reject) => {
+    const url = new URL(opts.url)
+    const transport = url.protocol === 'https:' ? https : http
```
uurien: Have you considered the use of the fetch API? it is much simpler and it is available from node 18? It works to promises, is the same for http and https etc. I think that we are using http in other places because they were implemented to support old node versions.
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

package.json
```diff
@@ -47,6 +49,7 @@
     "test:profiler": "tap \"packages/dd-trace/test/profiling/**/*.spec.js\"",
     "test:profiler:ci": "npm run test:profiler -- --coverage --nyc-arg=--include=\"packages/dd-trace/src/profiling/**/*.js\"",
     "test:integration": "mocha --timeout 60000 -r \"packages/dd-trace/test/setup/core.js\" \"integration-tests/*.spec.js\"",
+    "test:integration:aiguard": "mocha --timeout 60000 -r \"packages/dd-trace/test/setup/core.js\" \"integration-tests/aiguard/*.spec.js\"",
```
uurien, line 53: These test are not configured in any yml to run in the CI, are they?

CODEOWNERS
```diff
@@ -3,6 +3,9 @@
 /packages/dd-trace/src/appsec/ @DataDog/asm-js
 /packages/dd-trace/test/appsec/ @DataDog/asm-js
 
+/packages/dd-trace/src/ai_guard/ @DataDog/asm-js
+/packages/dd-trace/test/ai_guard/ @DataDog/asm-js
```
uurien: This is adding as codeowner of this PR a team where the owner of the PR is not in. We can start like this now, but we should find a better way to do this, maybe creating a new GH team for this?

index.d.ts
manuel-alvarez-alvarez: Moved to `aiguard`, they were in experimental as they might not be in the final version. Thanks for the feedback.

index.d.ts
manuel-alvarez-alvarez: I don't know why the formatter keeps changing it :cry:

package.json
manuel-alvarez-alvarez: What we run in CI is `test:aiguard:ci`, I added an integration job too, thanks for the input

CODEOWNERS
manuel-alvarez-alvarez: Yes, this is a bit the chicken and the egg issue, even though the team was not assigned at least it got reviewed by the team :smile:

packages/dd-trace/src/aiguard/client.js
manuel-alvarez-alvarez: Migrated to fetch, we didn't want to make a lot of changes but seems to be worth it.

CC: @IlyasShabi that also suggested this change

packages/dd-trace/src/aiguard/client.js
```diff
@@ -0,0 +1,35 @@
...
+async function executeRequest (body, opts) {
+  const postData = JSON.stringify(body)
+  const headers = {
+    'Content-Type': 'application/json',
+    'Content-Length': Buffer.byteLength(postData),
+    ...opts.headers
+  }
+
+  const controller = new AbortController()
+  const timeoutId = setTimeout(() => controller.abort(), opts.timeout)
```
uurien: afaik, all the `setTimeout` or `setInterval` should be `.unref` to prevent blocking/delaying any shutdown.

```suggestion
  const timeoutId = setTimeout(() => controller.abort(), opts.timeout)
  timeoutId.unref()
```

index.d.ts
simon-id: I'm reopening this.
Options that might not be in the final version and that might be changed or removed SHOULD go in experimental.
It's very easy to make an alias from experimental to the non-experimental config. But things that might be removed SHOULD NOT got outside of experimental. Ultimately, we should probably have all of aiguard config inside the experimental stuff too no ?
Appsec and iast was in their first implems

manuel-alvarez-alvarez: Moved everything into experimental, we can alias it later. Thanks for the input!

integration-tests/aiguard/api-mock.js
```diff
@@ -0,0 +1,62 @@
+'use strict'
+
+const express = require('express')
+
+// TODO we could try to use cassettes instead of a mock for the AI Guard API
```
sabrenner, line 5: i actually implemented the cassette feature for the testagent, there's likely a small QOL change i can make there for it to be used here but let me know if that's something you're actually interested in using!

packages/dd-trace/src/aiguard/client.js
```diff
@@ -0,0 +1,36 @@
...
+  timeoutId.unref()
```
sabrenner: nit
```suggestion
  const timeoutId = setTimeout(() => controller.abort(), opts.timeout).unref()
```

packages/dd-trace/src/aiguard/client.js
```diff
@@ -0,0 +1,36 @@
+'use strict'
+
+async function executeRequest (body, opts) {
+  const postData = JSON.stringify(body)
```
sabrenner, line 4: because it was of concern for llm observability - any unicode encoding issues we should be concerned about here? `JSON.stringify` will not do this, so if the respective backend cannot unmarshall correctly it could be an issue

packages/dd-trace/src/aiguard/noop.js
```diff
@@ -0,0 +1,15 @@
...
+
+class NoopAIGuard {
+  constructor (noopTracer) {
+    this._tracer = noopTracer
+  }
+
+  evaluate (messages, opts) {
+    return new Promise((resolve, reject) => {
+      resolve({ action: 'ALLOW', reason: 'AI Guard is not enabled' })
+    })
```
sabrenner: nit
```suggestion
    return Promise.resolve({ action: 'ALLOW', reason: 'AI Guard is not enabled' })
```

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,161 @@
...
+class AIGuardAbortError extends Error {
+  constructor (reason) {
+    super(reason)
+    this.name = 'AIGuardAbortError'
+    this.reason = reason
+  }
+}
+
+class AIGuardClientError extends Error {
+  constructor (message, opts = {}) {
```
sabrenner, line 27: nit: any reason we're calling it `message` here but `reason` above?

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,161 @@
...
+    let endpoint = config.experimental.aiguard.endpoint
+    if (!endpoint) {
+      endpoint = `https://app.${config.site}/api/v2/ai-guard`
+    }
+    this._evaluateUrl = `${endpoint}/evaluate`
+
+    this._timeout = config.experimental.aiguard.timeout
+    this._maxMessagesLength = config.experimental.aiguard.maxMessagesLength
+    this._maxContentSize = config.experimental.aiguard.maxContentSize
+    this._meta = { service: config.service, env: config.env }
```
sabrenner: nit: we can probably just read everything once from `config`

```suggestion
    const {
      experimental: {
        aigaurd
      },
      site,
      service,
      env
    } = config
    
    let endpoint = aiguard.endpoint ?? `https://app.${site}/api/v2/ai-guard`
    this._evaluateUrl = `${endpoint}/evaluate`

    this._timeout = aiguard.timeout
    this._maxMessagesLength = aiguard.maxMessagesLength
    this._maxContentSize = aiguard.maxContentSize
    this._meta = { service, env }
```

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  _truncate (messages) {
```
sabrenner: as long as we're not relying on these `_` methods outside of this class, we can change them to actual [private fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements)
```suggestion
  #truncate (messages) {
```

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,161 @@
...
+    const size = Math.min(messages.length, this._maxMessagesLength)
+    const result = messages.slice(-size)
+
+    for (let i = 0; i < size; i++) {
+      const message = result[i]
+      if ('content' in message && message.content.length > this._maxContentSize) {
```
sabrenner: nit
```suggestion
      if (message.content?.length > this._maxContentSize) {
```

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,161 @@
...
+    return result
+  }
+
+  _isToolCall (message) {
+    return 'tool_calls' in message || 'tool_call_id' in message
+  }
+
+  _getToolName (message, history) {
+    // 1. assistant message with tool calls
+    if ('tool_calls' in message) {
```
sabrenner:
```suggestion
    if (message.tool_calls) {
```

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,161 @@
...
+      const message = result[i]
+      if ('content' in message && message.content.length > this._maxContentSize) {
+        result[i] = { ...message, content: message.content.slice(0, this._maxContentSize) }
+      }
+    }
+    return result
+  }
+
+  _isToolCall (message) {
+    return 'tool_calls' in message || 'tool_call_id' in message
```
sabrenner: we can do either 

```javascript
return message.tool_calls || message.tool_call_id
```

which will return a truthy value (not directly a bool), or

```javascript
return Object.hasOwn(message, 'tool_calls') || Object.hasOwn(message, 'tool_call_id')
```

to avoid issues with prototype pollution

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  _getToolName (message, history) {
+    // 1. assistant message with tool calls
+    if ('tool_calls' in message) {
+      const names = message.tool_calls.map((tool) => tool.function.name)
+      return names.length === 0 ? null : names.join(',')
```
sabrenner, line 92: maybe dumb question - if this function is "getToolName", and the tag it ends up getting assigned to is "ai_guard.tool_name", but it's possible we return a comma-separated string of multiple tools, is it better to possibly tag them as "ai_guard.tool_names" instead? just for clarity - i do understand that backend implementation logic probably relies on the singular "ai_guard.tool_name", but just wanna see (also to make sure there's no confusion on my end!)

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,161 @@
...
+    }
+    // 2. assistant message with tool output (search the linked tool call in reverse order)
+    const id = message.tool_call_id
+    for (let i = history.length - 2; i >= 0; i--) {
+      const item = history[i]
+      if ('tool_calls' in item) {
```
sabrenner:
```suggestion
      if (item.tool_calls) {
```

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,161 @@
...
+      }
+      if (response.status !== 200) {
+        throw new AIGuardClientError(
+          `AI Guard service call failed, status ${response.status}`,
+          { errors: response.body?.errors })
+      }
+      let action, reason, blockingEnabled
+      try {
+        const attr = response.body.data.attributes
+        if (!('action' in attr)) {
```
sabrenner:
```suggestion
        if (!Object.hasOwn(attr, "action")) {
```
could probably also just be `!attr.action`

packages/dd-trace/src/config.js
```diff
@@ -859,6 +865,15 @@ class Config {
     this._setBoolean(env, 'trace.nativeSpanEvents', DD_TRACE_NATIVE_SPAN_EVENTS)
     env['vertexai.spanPromptCompletionSampleRate'] = maybeFloat(DD_VERTEXAI_SPAN_PROMPT_COMPLETION_SAMPLE_RATE)
     env['vertexai.spanCharLimit'] = maybeInt(DD_VERTEXAI_SPAN_CHAR_LIMIT)
+    this._setBoolean(env, 'experimental.aiguard.enabled', DD_AI_GUARD_ENABLED)
```
uurien: Remember that, as you are adding new config properties, you have to update https://github.com/DataDog/system-tests/blob/main/utils/telemetry/intake/static/config_norm_rules.json file in the ST and its equivalent in the backend repo.

packages/dd-trace/src/aiguard/sdk.js
smola: `reason` above is a natural language `reason` generated by AI Guard, and we continue that terminology there. While this one is a generic client error message (not from the LLM). I assume that was the difference.

integration-tests/aiguard/api-mock.js
manuel-alvarez-alvarez: So far mocking the API is manageable, but I'll come back to you if we decide to give cassettes a try, thanks!

packages/dd-trace/src/aiguard/sdk.js
manuel-alvarez-alvarez: Yeah, so this is basically an implementation of V2 of the messages format (close to the OpenAI API spec), in V1 it was not possible to have multiple tool calls in a single message, but now it's a reality and we ended up with `ai_guard.tool_name`, in order to change it we should coordinate with backend and frontend (probably we will change it in a separate PR once front and back can support the plural tag).

packages/dd-trace/src/aiguard/sdk.js
manuel-alvarez-alvarez: Fixed thanks for the input!

packages/dd-trace/src/aiguard/sdk.js
manuel-alvarez-alvarez: Changed, thanks!

packages/dd-trace/src/aiguard/client.js
manuel-alvarez-alvarez: I've done some unicode  tests and seems to be properly handled:

image

Any particular cases that I should test explicitly?

CC: @smola I think the backend is handling the encoding correctly, but just to be sure.

packages/dd-trace/src/config.js
manuel-alvarez-alvarez: Thanks I've just created both [PR1](https://github.com/DataDog/dd-go/pull/202551) and [PR2](https://github.com/DataDog/system-tests/pull/5375)

packages/dd-trace/src/aiguard/client.js
sabrenner: nice - no, i don't think any cases in particular, just wanted to call out as we (llmobs) had some issues originally w plain `JSON.stringify`, but all seems good!

CODEOWNERS
simon-id: Do you guys have a github team already ? I'm not happy being codeowner of this :/

packages/dd-trace/test/config.spec.js
```diff
@@ -359,6 +359,11 @@ describe('Config', () => {
     expect(config).to.have.nested.deep.property('tracePropagationStyle.extract', ['datadog', 'tracecontext', 'baggage'])
     expect(config).to.have.nested.deep.property('tracePropagationStyle.inject', ['datadog', 'tracecontext', 'baggage'])
     expect(config).to.have.property('tracing', true)
+    expect(config).to.have.nested.property('experimental.aiguard.enabled', false)
+    expect(config).to.have.nested.property('experimental.aiguard.endpoint', undefined)
+    expect(config).to.have.nested.property('experimental.aiguard.maxMessagesLength', 16)
+    expect(config).to.have.nested.property('experimental.aiguard.maxContentSize', 512 * 1024)
+    expect(config).to.have.nested.property('experimental.aiguard.timeout', 10_000)
```
simon-id: can we have those sorted alphabetically with the rest of the assertions ? basically move them a bit higher next to the other experimental stuff

packages/dd-trace/test/config.spec.js
```diff
@@ -359,6 +359,11 @@ describe('Config', () => {
...
 
     expect(updateConfig).to.be.calledOnce
```
simon-id, line 371: the assertion just below line 369 is missing your default values

packages/dd-trace/test/config.spec.js
```diff
@@ -655,6 +660,12 @@ describe('Config', () => {
     process.env.DD_VERSION = '1.0.0'
     process.env.DD_VERTEXAI_SPAN_CHAR_LIMIT = 50
     process.env.DD_VERTEXAI_SPAN_PROMPT_COMPLETION_SAMPLE_RATE = 0.5
+    process.env.DD_AI_GUARD_ENABLED = 'true'
+    process.env.DD_AI_GUARD_ENDPOINT = 'https://dd.datad0g.com/api/unstable/ai-guard'
+    process.env.DD_AI_GUARD_TIMEOUT = 2000
+    process.env.DD_AI_GUARD_MAX_MESSAGES_LENGTH = 32
+    process.env.DD_AI_GUARD_MAX_CONTENT_SIZE = 1024 * 1024
+    process.env.DD_APP_KEY = 'myAppKey'
```
simon-id: alphabetical order again porfa

packages/dd-trace/test/config.spec.js
```diff
@@ -768,10 +779,17 @@ describe('Config', () => {
...
+    expect(config).to.have.nested.property('experimental.aiguard.endpoint', 'https://dd.datad0g.com/api/unstable/ai-guard')
+    expect(config).to.have.nested.property('experimental.aiguard.maxMessagesLength', 32)
+    expect(config).to.have.nested.property('experimental.aiguard.maxContentSize', 1024 * 1024)
+    expect(config).to.have.nested.property('experimental.aiguard.timeout', 2000)
+    expect(config).to.have.property('appKey', 'myAppKey')
 
     expect(updateConfig).to.be.calledOnce
 
     expect(updateConfig.getCall(0).args[0]).to.deep.include.members([
+
```
simon-id: unnecessary diff
```suggestion
```

packages/dd-trace/test/config.spec.js
```diff
@@ -768,10 +779,17 @@ describe('Config', () => {
     expect(config).to.have.nested.deep.property('tracePropagationStyle.inject', ['b3', 'tracecontext'])
     expect(config).to.have.property('tracing', false)
     expect(config).to.have.property('version', '1.0.0')
+    expect(config).to.have.nested.property('experimental.aiguard.enabled', true)
+    expect(config).to.have.nested.property('experimental.aiguard.endpoint', 'https://dd.datad0g.com/api/unstable/ai-guard')
+    expect(config).to.have.nested.property('experimental.aiguard.maxMessagesLength', 32)
+    expect(config).to.have.nested.property('experimental.aiguard.maxContentSize', 1024 * 1024)
+    expect(config).to.have.nested.property('experimental.aiguard.timeout', 2000)
+    expect(config).to.have.property('appKey', 'myAppKey')
```
simon-id: alphabetical order again

simon-id, line 787: alphabetical order again

CODEOWNERS
```diff
@@ -93,6 +96,7 @@
 # CI
 /.github/workflows/apm-capabilities.yml @DataDog/apm-sdk-capabilities-js
 /.github/workflows/apm-integrations.yml @Datadog/apm-idm-js
+/.github/workflows/aiguardappsec.yml.yml @DataDog/asm-js
```
simon-id: uh ? lmao
```suggestion
/.github/workflows/aiguard.yml @DataDog/asm-js
```

CODEOWNERS
```diff
@@ -3,6 +3,9 @@
 /packages/dd-trace/src/appsec/ @DataDog/asm-js
 /packages/dd-trace/test/appsec/ @DataDog/asm-js
 
+/packages/dd-trace/src/ai_guard/ @DataDog/asm-js
+/packages/dd-trace/test/ai_guard/ @DataDog/asm-js
+
```
simon-id, line 9: could add the integration-tests/aiguard here ?

packages/dd-trace/src/aiguard/client.js
```diff
@@ -0,0 +1,35 @@
...
+
+    const responseBody = await response.json()
+    return {
+      status: response.status,
+      body: responseBody
+    }
+  } catch (error) {
+    clearTimeout(timeoutId)
+    throw error
+  }
```
simon-id: BEHOLD `AbortSignal.timeout()`

```suggestion
  const response = await fetch(opts.url, {
    method: 'POST',
    headers,
    body: postData,
    signal: AbortSignal.timeout(opts.timeout)
  })

  return {
    status: response.status,
    body: await response.json()
  }
```

simon-id, line 32: hold my beer

```suggestion
  const response = await fetch(opts.url, {
    method: 'POST',
    headers,
    body: postData,
    signal: AbortSignal.timeout(opts.timeout)
  })

  const responseBody = await response.json()
  return {
    status: response.status,
    body: responseBody
  }
```

packages/dd-trace/src/aiguard/noop.js
```diff
@@ -0,0 +1,13 @@
+'use strict'
+
+class NoopAIGuard {
+  constructor (noopTracer) {
+    this._tracer = noopTracer
+  }
+
+  evaluate (messages, opts) {
```
simon-id: Not sure the constructor is useful. I think it's only useful if you wanna call _tracer methods in this noop sdk (that would also be noop) ?

```suggestion
  evaluate (messages, opts) {
```

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,161 @@
...
+    }
+  }
+}
+
+class AIGuard extends NoopAIGuard {
+  constructor (tracer, config) {
+    super(tracer)
+
+    if (!config.apiKey || !config.appKey) {
+      throw new Error('AIGuard: missing api and/or app keys, use env DD_API_KEY and DD_APP_KEY')
```
simon-id: This will throw in a potentially unexpected place and probably crash the process. I'm pretty sure our philosophy is to never crash. I'd rather log the error in telemetry/debuglogs and just keep the noop sdk ?

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,161 @@
...
+    }
+    this._headers = {
+      'DD-API-KEY': config.apiKey,
+      'DD-APPLICATION-KEY': config.appKey,
+    }
+
+    let endpoint = config.experimental.aiguard.endpoint
+    if (!endpoint) {
+      endpoint = `https://app.${config.site}/api/v2/ai-guard`
+    }
```
simon-id:
```suggestion
    const endpoint = config.experimental.aiguard.endpoint || `https://app.${config.site}/api/v2/ai-guard`
```

simon-id: actually shouldn't this be set as a default value in config.js instead ? why is there a default value here ?

CODEOWNERS
```diff
@@ -3,6 +3,9 @@
 /packages/dd-trace/src/appsec/ @DataDog/asm-js
 /packages/dd-trace/test/appsec/ @DataDog/asm-js
 
+/packages/dd-trace/src/ai_guard/ @DataDog/asm-js
```
simon-id: folder says aiguard, and here it's ai_guard ?

packages/dd-trace/src/noop/proxy.js
```diff
@@ -4,18 +4,21 @@ const NoopTracer = require('./tracer')
 const NoopAppsecSdk = require('../appsec/sdk/noop')
 const NoopDogStatsDClient = require('./dogstatsd')
 const NoopLLMObsSDK = require('../llmobs/noop')
+const NoopAIGuardSDK = require('../aiguard/noop')
 
 const noop = new NoopTracer()
 const noopAppsec = new NoopAppsecSdk()
 const noopDogStatsDClient = new NoopDogStatsDClient()
 const noopLLMObs = new NoopLLMObsSDK(noop)
+const noopAIGuard = new NoopAIGuardSDK(noop)
```
simon-id: as said in one of the comments above, i don't think passing the nooptracer is useful here
```suggestion
const noopAIGuard = new NoopAIGuardSDK()
```

packages/dd-trace/src/config.js
```diff
@@ -486,11 +486,17 @@ class Config {
     const {
       AWS_LAMBDA_FUNCTION_NAME,
       DD_AGENT_HOST,
+      DD_AI_GUARD_ENABLED,
+      DD_AI_GUARD_ENDPOINT,
+      DD_AI_GUARD_TIMEOUT,
+      DD_AI_GUARD_MAX_MESSAGES_LENGTH,
+      DD_AI_GUARD_MAX_CONTENT_SIZE,
```
simon-id, line 490: we try to keep this in alphabetical order:

```suggestion
      DD_AI_GUARD_ENABLED,
      DD_AI_GUARD_ENDPOINT,
      DD_AI_GUARD_MAX_MESSAGES_LENGTH,
      DD_AI_GUARD_MAX_CONTENT_SIZE,
      DD_AI_GUARD_TIMEOUT,
```

packages/dd-trace/src/config.js
```diff
@@ -859,6 +865,15 @@ class Config {
...
     env['vertexai.spanCharLimit'] = maybeInt(DD_VERTEXAI_SPAN_CHAR_LIMIT)
+    this._setBoolean(env, 'experimental.aiguard.enabled', DD_AI_GUARD_ENABLED)
+    this._setString(env, 'experimental.aiguard.endpoint', DD_AI_GUARD_ENDPOINT)
+    env['experimental.aiguard.timeout'] = maybeInt(DD_AI_GUARD_TIMEOUT)
+    this._envUnprocessed['experimental.aiguard.timeout'] = DD_AI_GUARD_TIMEOUT
+    env['experimental.aiguard.maxMessagesLength'] = maybeInt(DD_AI_GUARD_MAX_MESSAGES_LENGTH)
+    this._envUnprocessed['experimental.aiguard.maxMessagesLength'] = DD_AI_GUARD_MAX_MESSAGES_LENGTH
+    env['experimental.aiguard.maxContentSize'] = maybeInt(DD_AI_GUARD_MAX_CONTENT_SIZE)
+    this._envUnprocessed['experimental.aiguard.maxContentSize'] = DD_AI_GUARD_MAX_CONTENT_SIZE
+    this._setString(env, 'appKey', DD_APP_KEY)
```
simon-id: same thing here, alphabetical order, exprimental should go up with the rest of experimental, and appkey should go all the way up

packages/dd-trace/src/config.js
```diff
@@ -1029,6 +1044,17 @@ class Config {
...
+    this._setBoolean(opts, 'experimental.aiguard.enabled', options.experimental?.aiguard?.enabled)
+    this._setString(opts, 'experimental.aiguard.endpoint', options.experimental?.aiguard?.endpoint)
+    opts['experimental.aiguard.timeout'] = maybeInt(options.experimental?.aiguard?.timeout)
+    this._optsUnprocessed['experimental.aiguard.timeout'] = options.experimental?.aiguard?.timeout
+    opts['experimental.aiguard.maxMessagesLength'] = maybeInt(options.experimental?.aiguard?.maxMessagesLength)
+    this._optsUnprocessed['experimental.aiguard.maxMessagesLength'] = options.experimental?.aiguard?.maxMessagesLength
+    opts['experimental.aiguard.maxContentSize'] = maybeInt(options.experimental?.aiguard?.maxContentSize)
+    this._optsUnprocessed['experimental.aiguard.maxContentSize'] = options.experimental?.aiguard?.maxContentSize
+
+    this._setString(opts, 'appKey', options.appKey)
```
simon-id: same thing, alphabetical order please

packages/dd-trace/test/config.spec.js
manuel-alvarez-alvarez: Changed, thanks!

packages/dd-trace/test/config.spec.js
manuel-alvarez-alvarez: Added, thanks

packages/dd-trace/test/config.spec.js
manuel-alvarez-alvarez: Changed

CODEOWNERS
manuel-alvarez-alvarez: Added them, thanks!

CODEOWNERS
manuel-alvarez-alvarez: Good catch!

packages/dd-trace/src/aiguard/client.js
manuel-alvarez-alvarez: Wow! didn't know about this one, makes the code way cleaner, thanks!

CODEOWNERS
manuel-alvarez-alvarez: Changed, thanks!

packages/dd-trace/src/config.js
manuel-alvarez-alvarez: Changed, thanks!

packages/dd-trace/src/aiguard/sdk.js
manuel-alvarez-alvarez: Changed, I wasn't sure if breaking was OK due to bad config

packages/dd-trace/src/aiguard/noop.js
manuel-alvarez-alvarez: Changed thanks!

packages/dd-trace/src/aiguard/sdk.js
manuel-alvarez-alvarez: This is defined either by config property or computed from the site configuration, is there a better place to add the computation?

packages/dd-trace/test/config.spec.js
```diff
@@ -544,11 +554,17 @@ describe('Config', () => {
   })
 
   it('should initialize from environment variables', () => {
+    process.env.DD_AI_GUARD_ENABLED = 'true'
+    process.env.DD_AI_GUARD_ENDPOINT = 'https://dd.datad0g.com/api/unstable/ai-guard'
+    process.env.DD_AI_GUARD_TIMEOUT = 2000
+    process.env.DD_AI_GUARD_MAX_CONTENT_SIZE = 1024 * 1024
+    process.env.DD_AI_GUARD_MAX_MESSAGES_LENGTH = 32
```
simon-id, line 562:
```suggestion
    process.env.DD_AI_GUARD_ENABLED = 'true'
    process.env.DD_AI_GUARD_ENDPOINT = 'https://dd.datad0g.com/api/unstable/ai-guard'
    process.env.DD_AI_GUARD_MAX_CONTENT_SIZE = 1024 * 1024
    process.env.DD_AI_GUARD_MAX_MESSAGES_LENGTH = 32
    process.env.DD_AI_GUARD_TIMEOUT = 2000
```

packages/dd-trace/test/config.spec.js
simon-id, line 1: line 797: `expect(updateConfig.getCall(0).args[0]).to.deep.include.members([` is missing the env var values too

CODEOWNERS
```diff
@@ -93,6 +97,7 @@
 # CI
 /.github/workflows/apm-capabilities.yml @DataDog/apm-sdk-capabilities-js
 /.github/workflows/apm-integrations.yml @Datadog/apm-idm-js
+/.github/workflows/aiguard.yml.yml @DataDog/asm-js
```
simon-id: it's still wrong hahahha

```suggestion
/.github/workflows/aiguard.yml @DataDog/asm-js
```

packages/dd-trace/src/aiguard/sdk.js
simon-id: Ah I see, it's dependent on other config. In that case we can keep it here :+1:

packages/dd-trace/test/config.spec.js
```diff
@@ -1184,6 +1214,12 @@ describe('Config', () => {
     expect(config).to.have.nested.deep.property('tracePropagationStyle.extract', ['datadog'])
     expect(config).to.have.nested.deep.property('tracePropagationStyle.inject', ['datadog'])
     expect(config).to.have.property('version', '0.1.0')
+    expect(config).to.have.nested.property('experimental.aiguard.enabled', true)
+    expect(config).to.have.nested.property('experimental.aiguard.endpoint', 'https://dd.datad0g.com/api/unstable/ai-guard')
+    expect(config).to.have.nested.property('experimental.aiguard.maxContentSize', 1024 * 1024)
+    expect(config).to.have.nested.property('experimental.aiguard.maxMessagesLength', 32)
+    expect(config).to.have.nested.property('experimental.aiguard.timeout', 2000)
+    expect(config).to.have.property('appKey', 'myAppKey')
```
simon-id: once again, alphabetical order :)

packages/dd-trace/test/config.spec.js
```diff
@@ -1184,6 +1214,12 @@ describe('Config', () => {
...
 
     expect(updateConfig).to.be.calledOnce
```
simon-id, line 1231: and you're also missing the the values right below this line, just like my older comments

packages/dd-trace/test/config.spec.js
simon-id, line 1: you're also missing a change in the `should give priority to the options` test

packages/dd-trace/test/config.spec.js
manuel-alvarez-alvarez: Fixed thanks!

packages/dd-trace/test/config.spec.js
manuel-alvarez-alvarez: Added, thanks for the input

CODEOWNERS
manuel-alvarez-alvarez: My bad! fixed

packages/dd-trace/test/config.spec.js
manuel-alvarez-alvarez: Fixed

packages/dd-trace/test/config.spec.js
manuel-alvarez-alvarez: I included the new assertions

packages/dd-trace/test/config.spec.js
manuel-alvarez-alvarez: Fixed thanks

manuel-alvarez-alvarez (commented):
Applied all PR suggestions

packages/dd-trace/test/config.spec.js
```diff
@@ -809,6 +831,11 @@ describe('Config', () => {
       { name: 'dynamicInstrumentation.redactionExcludedIdentifiers', value: ['a', 'b', 'c'], origin: 'env_var' },
       { name: 'dynamicInstrumentation.uploadIntervalSeconds', value: 0.1, origin: 'env_var' },
       { name: 'env', value: 'test', origin: 'env_var' },
+      { name: 'experimental.aiguard.enabled', value: false, origin: 'default' },
```
simon-id, line 837: missing the app key line a bit higher no ?

packages/dd-trace/test/config.spec.js
simon-id, line 1: Since you added a programmatic configuration for appkey, you need to also include it in the tests `should initialize from the options` and `should give priority to the options` just like the rest

packages/dd-trace/src/config.js
```diff
@@ -943,6 +959,14 @@ class Config {
...
       options.dynamicInstrumentation?.uploadIntervalSeconds
     this._setString(opts, 'env', options.env || tags.env)
+    this._setBoolean(opts, 'experimental.aiguard.enabled', options.experimental?.aiguard?.enabled)
+    this._setString(opts, 'experimental.aiguard.endpoint', options.experimental?.aiguard?.endpoint)
+    opts['experimental.aiguard.timeout'] = maybeInt(options.experimental?.aiguard?.timeout)
+    this._optsUnprocessed['experimental.aiguard.timeout'] = options.experimental?.aiguard?.timeout
+    opts['experimental.aiguard.maxMessagesLength'] = maybeInt(options.experimental?.aiguard?.maxMessagesLength)
+    this._optsUnprocessed['experimental.aiguard.maxMessagesLength'] = options.experimental?.aiguard?.maxMessagesLength
+    opts['experimental.aiguard.maxContentSize'] = maybeInt(options.experimental?.aiguard?.maxContentSize)
+    this._optsUnprocessed['experimental.aiguard.maxContentSize'] = options.experimental?.aiguard?.maxContentSize
```
simon-id, line 966:
```suggestion
    this._setBoolean(opts, 'experimental.aiguard.enabled', options.experimental?.aiguard?.enabled)
    this._setString(opts, 'experimental.aiguard.endpoint', options.experimental?.aiguard?.endpoint)
    opts['experimental.aiguard.maxMessagesLength'] = maybeInt(options.experimental?.aiguard?.maxMessagesLength)
    this._optsUnprocessed['experimental.aiguard.maxMessagesLength'] = options.experimental?.aiguard?.maxMessagesLength
    opts['experimental.aiguard.maxContentSize'] = maybeInt(options.experimental?.aiguard?.maxContentSize)
    this._optsUnprocessed['experimental.aiguard.maxContentSize'] = options.experimental?.aiguard?.maxContentSize
    opts['experimental.aiguard.timeout'] = maybeInt(options.experimental?.aiguard?.timeout)
    this._optsUnprocessed['experimental.aiguard.timeout'] = options.experimental?.aiguard?.timeout
```

packages/dd-trace/test/config.spec.js
manuel-alvarez-alvarez: Yes, I missed it, thanks!

packages/dd-trace/test/config.spec.js
manuel-alvarez-alvarez: I just removed it, as it's a bad idea to allow configuring keys programmatically, only the env is applied (like for `apiKey`)

packages/dd-trace/src/config.js
manuel-alvarez-alvarez: Fixed thanks

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,164 @@
...
+    this._timeout = config.experimental.aiguard.timeout
+    this._maxMessagesLength = config.experimental.aiguard.maxMessagesLength
+    this._maxContentSize = config.experimental.aiguard.maxContentSize
+    this._meta = { service: config.service, env: config.env }
+    this._initialized = true
+  }
+
+  #truncate (messages) {
+    const size = Math.min(messages.length, this._maxMessagesLength)
+    const result = messages.slice(-size)
```
sabrenner, line 73: does it also make sense to grab any system message(s) as part of this for extra context? not sure what was done for python/what is expected for the `evaluate` API in general but maybe makes sense to ensure this is grabbed regardless of if `maxMessagesLength` is set?

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,164 @@
...
+
+    for (let i = 0; i < size; i++) {
+      const message = result[i]
+      if (message.content?.length > this._maxContentSize) {
+        result[i] = { ...message, content: message.content.slice(0, this._maxContentSize) }
```
sabrenner: similarly here, it might make sense to exclude any system message(s) from this to capture full context for any evaluations that happen

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,164 @@
...
+      if (target === 'tool') {
+        const name = this.#getToolName(last, messages)
+        if (name) {
+          span.setTag(AI_GUARD_TOOL_NAME_TAG_KEY, name)
+        }
+      }
+      span.meta_struct = {
+        [AI_GUARD_META_STRUCT_KEY]: {
+          messages: this.#truncate(messages)
+        }
```
sabrenner, line 127: maybe a dumb question but i was grepping in the codebase and saw some other usage of `span.meta_struct`. does it make sense to add on to any other things that might already be there instead of overriding? or does it make sense to just set it as you already do?

```suggestion
      span.meta_struct ??= {}
      span.meta_struct[AI_GUARD_META_STRUCT_KEY] = {
        messages: this.#truncate(messages)
      }
```

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,164 @@
...
+      this.errors = opts.errors
+    }
+    if (opts.cause) {
+      this.cause = opts.cause
+    }
+  }
+}
+
+class AIGuard extends NoopAIGuard {
+  constructor (tracer, config) {
```
sabrenner, line 49: we can also use `#` for all private properties/variables defined here (unless they're accessed in tests, in which case maybe getters make sense?)

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,164 @@
...
+      } catch (e) {
+        throw new AIGuardClientError(`AI Guard service returned unexpected response : ${response.body}`, { cause: e })
+      }
+      span.setTag(AI_GUARD_ACTION_TAG_KEY, action)
+      span.setTag(AI_GUARD_REASON_TAG_KEY, reason)
+      if (block && blockingEnabled && action !== ALLOW) {
+        span.setTag(AI_GUARD_BLOCKED_TAG_KEY, 'true')
+        throw new AIGuardAbortError(reason)
+      }
+      return { action, reason }
```
sabrenner, line 168: (maybe as a follow-up) because exceptions/generating stack traces can be expensive, i wonder if it'd be nice to add some kind of soft-fail configuration (ie `aiguard.softFail`), such that if it's enabled this doesn't throw, but instead returns `block: true` as part of the result along with `{ action, reason }` and let users handle it that way.

i don't have a problem w the exceptions as-is, as i can see it being the users intention to handle errors from `evaluate` as hard exceptions, but still wanted to float the suggestion!

sabrenner (commented):
nice, looks very clean now! i'm good to approve once these comments are resolved/have an answer - i mainly just looked at the sdk file, seeing as Simon looked at a bunch of the config stuff.

also, apologies if some of these comments have come up before and have answers already, feel free to link to past comments!

packages/dd-trace/src/aiguard/sdk.js
manuel-alvarez-alvarez: We only truncate the messages included in the span’s metastruct (to stay within limits), but the API always receives the complete set of messages for evaluation.

packages/dd-trace/src/aiguard/sdk.js
manuel-alvarez-alvarez: Yeah, it makes sense if you are attaching information to an existing span that might already contain `meta_struct` data, in this case we are using a new span so it's OK to fully define the meta_struct (it will always be empty).

packages/dd-trace/src/aiguard/sdk.js
manuel-alvarez-alvarez: We truncate messages from the beginning of the list (dropping the system prompt and early messages in the sp an), but evaluations still use the full context to ensure correctness.

packages/dd-trace/src/aiguard/sdk.js
manuel-alvarez-alvarez: Changed, thanks for the feedback

packages/dd-trace/src/aiguard/sdk.js
manuel-alvarez-alvarez: The evaluate method accepts an optional block parameter:

```ts
evaluate(messages: Message[], opts?: { block?: boolean }): Promise<Evaluation>;
```

By default, `block` is `false`, meaning no exception is thrown—the customer decides how to handle the evaluation.

index.d.ts
```diff
@@ -1059,6 +1087,170 @@ declare namespace tracer {
...
+       *
+       * @param messages - Array of conversation messages
+       * @param opts - Optional configuration object:
+       *   - `block`: When true, throws an exception if evaluation result is not 'ALLOW'
+       *              and the AI Guard service has blocking mode enabled (default: false).
+       * @returns Promise resolving to an Evaluation with the security decision and reasoning.
+       *          The promise rejects with AIGuardAbortError when `opts.block` is true and the evaluation result would block the request.
+       *          The promise rejects with AIGuardClientError when communication with the AI Guard service fails.
+       */
+      evaluate(messages: Message[], opts?: { block?: boolean }): Promise<Evaluation>;
```
uurien: Seems that we have no linter for this file, there is a missing space between method name and `(`
```suggestion
      evaluate (messages: Message[], opts?: { block?: boolean }): Promise<Evaluation>;
```

packages/dd-trace/src/aiguard/sdk.js
sabrenner: ahhh yes i totally missed this, disregard!

packages/dd-trace/src/aiguard/sdk.js
sabrenner: yeah, thanks for the context, i just misread!

packages/dd-trace/src/aiguard/sdk.js
sabrenner: 🤦 i also totally missed this, makes a ton of sense!

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -0,0 +1,173 @@
...
+
+class AIGuard extends NoopAIGuard {
+  #initialized
+  #tracer
+  #headers
+  #evaluateUrl
+  #timeout
+  #maxMessagesLength
+  #maxContentSize
+  #meta
```
sabrenner, line 47: (for a follow up) we can add type hints to these as well! something like

```javascript
/** @type {boolean} */
#initialized

uurien (approved):

PR #6488, opened 2025-09-22
[test optimization] Improve organization of `cypress` integration tests

juan-fernandez:
### What does this PR do?

* Improve flakiness of `cypress` integration tests

### Motivation

* Avoid race conditions where the promise that resolves to the received events resolving before the process' `exit` causes the test to fail. 

### Plugin Checklist
Just check that the tests we have pass.

### DIFF
--- a/integration-tests/esbuild/build-and-test-git-tags.js
+++ b/integration-tests/esbuild/build-and-test-git-tags.js
@@ -1,0 +1,57 @@
+#!/usr/bin/env node
+'use strict'
+
+/* eslint-disable no-console */
+const fs = require('fs')
+const { spawnSync } = require('child_process')
+const assert = require('assert')
+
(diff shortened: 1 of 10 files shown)


integration-tests/cypress/cypress.spec.js
```diff
@@ -2514,38 +2502,41 @@ moduleTypes.forEach(({
...
             )
           })
       })
 
       context('test is new', () => {
-        it('should be retried and marked both as new and modified', (done) => {
-          receiver.setKnownTests({})
+        it('should be retried and marked both as new and modified', async () => {
+          receiver.setKnownTests({
+            cypress: {}
```
juan-fernandez, line 2538: this should've made the test fail before but it didn't, so I'm pretty confused 😆

calvinbayer (approved):

PR #6492, opened 2025-09-22
[test optimization] Update `ci/init` logic not to fail if within a test worker

juan-fernandez:
### What does this PR do?

Change initialization logic:

* If `dd-trace` is running in a test worker, the logic to check agentless and `DD_API_KEY` is skipped altogether. 
* Additionally, telemetry will not be reported from test workers.

### Motivation

Test workers do not communicate to the outside world but through IPC to the main process. Having them fail because `DD_API_KEY` is undefined does not make sense. 

### Plugin Checklist

- [x] Integration tests.

### DIFF
--- a/ci/init.js
+++ b/ci/init.js
@@ -27,7 +27,7 @@ function detectTestWorkerType () {
   if (getEnvironmentVariable('JEST_WORKER_ID')) return 'jest'
   if (getEnvironmentVariable('CUCUMBER_WORKER_ID')) return 'cucumber'
   if (getEnvironmentVariable('MOCHA_WORKER_ID')) return 'mocha'
-  if (getEnvironmentVariable('PLAYWRIGHT_WORKER')) return 'playwright'
+  if (getEnvironmentVariable('DD_PLAYWRIGHT_WORKER')) return 'playwright'
   if (getEnvironmentVariable('TINYPOOL_WORKER_ID')) return 'vitest'
   return null
 }


integration-tests/ci-visibility/automatic-log-submission.spec.js
```diff
@@ -219,18 +220,22 @@ describe('test visibility automatic log submission', () => {
...
         })
+
+        await Promise.all([
+          once(childProcess, 'exit'),
+          once(childProcess.stdout, 'end'),
+          once(childProcess.stderr, 'end'),
+        ])
+
+        assert.include(testOutput, 'Hello simple log!')
+        assert.include(testOutput, 'no automatic log submission will be performed')
```
juan-fernandez, line 238: ℹ️ these changes are just a refactor to use `async` instead of `done`

integration-tests/ci-visibility/test-optimization-startup.spec.js
```diff
@@ -72,11 +79,80 @@ describe('test optimization startup', () => {
...
     })
+
+    await Promise.all([
+      once(childProcess, 'exit'),
+      once(childProcess.stdout, 'end'),
+      once(childProcess.stderr, 'end')
+    ])
+
+    assert.include(processOutput, 'hello!')
+    assert.notInclude(processOutput, 'dd-trace will not be initialized')
```
juan-fernandez, line 156: ℹ️ used the chance to also clean up the test file

ci/init.js
```diff
@@ -6,88 +6,78 @@ const { isTrue, isFalse } = require('../packages/dd-trace/src/util')
...
-    options.experimental = {
-      exporter: 'datadog'
-    }
-  } else {
-    console.error('DD_CIVISIBILITY_AGENTLESS_ENABLED is set, but neither ' +
-      'DD_API_KEY nor DATADOG_API_KEY are set in your environment, so ' +
-      'dd-trace will not be initialized.')
-    shouldInit = false
+if (isTestWorker) {
+  baseOptions.telemetry = { enabled: false }
```
vitor-de-araujo, line 54: Isn't it a problem to not have telemetry for the workers?

juan-fernandez: yeah but right now it's causing more issues than benefits. The telemetry module in the library is not designed for short lived processes that can get killed abruptly. I'd rather disable it now and figure out how to add it back later on

PR #6496, opened 2025-09-23
chore(lint): align naming of file- and directory names

watson:
### What does this PR do?

Add new linting rule, requireing all files and directories (with a few exceptions) to use kebab-case.

To perform the lintig, the project [ls-lint](https://ls-lint.org/) is used. It's added to the `npm run lint` command and will therefore run as part of CI.

### Motivation

No standard existed previously, so the project contained a mixture of different styles. This can make it confusing to parse and to harder to decide what to name a new file or folder.

### Additional Notes

We could also have chosen to use snake case, but most files and folders already used kebab-case, so settling on this standard meant a smaller PR.


index.d.ts
```diff
@@ -170,7 +170,7 @@ interface Plugins {
   "azure-service-bus": tracer.plugins.azure_service_bus;
   "bunyan": tracer.plugins.bunyan;
   "cassandra-driver": tracer.plugins.cassandra_driver;
-  "child_process": tracer.plugins.child_process;
+  "child-process": tracer.plugins.child_process;
```
BridgeAR, line 173: While I can imagine us supporting both notations, this is currently a small breaking change

BridgeAR:
@watson are you likely going to work on this soonish? I would suggest we close it otherwise for now

BridgeAR:
Closing for now. Please feel free to reopen

PR #6502, opened 2025-09-23
Prevent IITM failure when there's another loader hook ahead of it.

bengl:
In cases where there's a loader hook that's loaded before IITM, and that hook does the Node.js-recommended approach for handling CommonJS modules, IITM will break because it imports a CommonJS module immediately. This workaround uses `Module.createRequire()` to bypass the loader hook logic, for now.

### DIFF
--- a/integration-tests/double-loader.spec.js
+++ b/integration-tests/double-loader.spec.js
@@ -1,0 +1,16 @@
+'use strict'
+
+const { execSync } = require('child_process')
+const path = require('path')
+const assert = require('assert')
+const { describe, it } = require('mocha')
+
+describe('double-loader scenario', () => {
+  it('does not fail', () => {
+    const l1 = path.join(__dirname, 'double-loader', 'loader1.mjs')
+    const l2 = path.join(__dirname, '..', 'loader-hook.mjs')
+    assert.doesNotThrow(() => {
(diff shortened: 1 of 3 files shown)


integration-tests/double-loader.spec.js
```diff
@@ -0,0 +1,16 @@
...
+const assert = require('assert')
+const { describe, it } = require('mocha')
+
+describe('double-loader scenario', () => {
+  it('does not fail', () => {
+    const l1 = path.join(__dirname, 'double-loader', 'loader1.mjs')
+    const l2 = path.join(__dirname, '..', 'loader-hook.mjs')
+    assert.doesNotThrow(() => {
+      execSync(`node --no-warnings --loader=${l1} --loader=${l2} -p 0`)
+    })
```
BridgeAR, line 14:
```suggestion
    execSync(`node --no-warnings --loader=${l1} --loader=${l2} -p 0`)
```

See the description of https://nodejs.org/api/assert.html#assertdoesnotthrowfn-error-message

BridgeAR (commented):
Should this land as is? I saw the iitm PR was reverted due to some regressions and we can likely just update iitm fast enough and have this land as test case afterwards.

BridgeAR:
The (alternative) fix landed on iitm directly and was released as v2. Closing as superseded

PR #6504, opened 2025-09-23
test: fix runtime metric flakiness

BridgeAR:
The cpu metric could have a rounding issue. This is fixed by accepting slightly lower values.

### DIFF
--- a/packages/dd-trace/test/runtime_metrics.spec.js
+++ b/packages/dd-trace/test/runtime_metrics.spec.js
@@ -521,7 +521,7 @@ function createGarbage (count = 50) {
               }
               if (metric === 'runtime.node.cpu.total') {
                 assert(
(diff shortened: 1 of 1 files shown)


packages/dd-trace/test/runtime_metrics.spec.js
```diff
@@ -521,7 +521,8 @@ function createGarbage (count = 50) {
               }
               if (metric === 'runtime.node.cpu.total') {
                 assert(
-                  number >= expected && number <= expected + 1,
+                  // Lower numbers can happen due to rounding issues.
```
watson: It's not clear from the comment what you're doing to fix it. I think this should help:

```suggestion
                  // Subtracting 0.02 since lower numbers can happen due to rounding issues.
```

watson (approved):

PR #6505, opened 2025-09-23
[MLOB-3170] add prompt caching metrics for bedrock

ncybul:
This PR adds prompt caching metrics (cached read / write token counts) to spans submitted from the Bedrock integration.

Note: The input token count returned as part of the Bedrock response only includes non-cached tokens. Therefore, we have to adjust this token count to include cached tokens as well (since that is what our backend expects).

## Testing

I ran the following script with `node --inspect-brk bedrock_caching_simple.js` to test these changes:
```
const tracer = require('dd-trace').init({
    llmobs: {
      mlApp: "nicole-test",
      agentlessEnabled: false,
    },
    env: "nicole-test",
  });
  const { llmobs } = tracer;
  
  const AWS = require('@aws-sdk/client-bedrock-runtime');
  const bedrockRuntimeClient = new AWS.BedrockRuntimeClient({
    region: 'us-east-1',
  });
  
  const modelId = 'us.anthropic.claude-sonnet-4-20250514-v1:0';
  const systemPrompt = "You are an expert software developer".repeat(200);
  
  (async () => {
    const request1 = {
      body: JSON.stringify({
        system: systemPrompt,
        temperature: 0.1,
        anthropic_version: 'bedrock-2023-05-31',
        messages: [
          {
            role: 'user',
            content: [
              {
                type: 'text',
                text: 'What are the best practices for API design?',
              },
              {
                type: "text",
                text: "test",
                cache_control: {
                  type: "ephemeral"
                }
              }
            ],
          }
        ],
        max_tokens: 100,
      }),
      contentType: 'application/json',
      accept: 'application/json',
      modelId: modelId
    };
  
    const command1 = new AWS.InvokeModelCommand(request1);
    await bedrockRuntimeClient.send(command1);
})();
```

Running this script once leads to the first request where you can see that the system prompt is being written to the cache. Running the script again leads to the second request where the shared system prompt is read from the cache.

|  | span | Cached Tokens |
|----------|----------|----------|
| first request   | [span](https://app.datadoghq.com/llm/traces?query=%40ml_app%3Anicole-test%20%40event_type%3Aspan%20%40parent_id%3Aundefined&agg_m=count&agg_m_source=base&agg_t=count&fromUser=false&sp=%5B%7B%22sp%22%3A%7B%22width%22%3A%22min%28100%25%2C%20max%28calc%28100%25%20-%20var%28--ui-page-left-offset%29%20-%2016px%29%2C%20900px%29%29%22%7D%2C%22p%22%3A%7B%22eventId%22%3A%22AwAAAZl36hAlMbKDqQAAABhBWmwzNmhBbEFBQ1EzQWVEaV9MdUFBQUEAAAAkMDE5OTc3ZWEtNGYzNi00ZjdiLTg4YmUtOGI4OGFkZjQxYTI0AAASwA%22%7D%2C%22i%22%3A%22llm-obs-panel%22%7D%5D&spanId=4133420068894346222&start=1758652800504&end=1758653700504&paused=false)  | image   |
| second request (hits the cache)   |  [span](https://app.datadoghq.com/llm/traces?query=%40ml_app%3Anicole-test%20%40event_type%3Aspan%20%40parent_id%3Aundefined&agg_m=count&agg_m_source=base&agg_t=count&fromUser=false&sp=%5B%7B%22sp%22%3A%7B%22width%22%3A%22min%28100%25%2C%20max%28calc%28100%25%20-%20var%28--ui-page-left-offset%29%20-%2016px%29%2C%20900px%29%29%22%7D%2C%22p%22%3A%7B%22eventId%22%3A%22AwAAAZl36kjopgzKJQAAABhBWmwzNmtqb0FBREVQMHZvQzk3SEFBQUEAAAAkMDE5OTc3ZWEtZWNjNi00YWY2LTkyMGYtODg0NDg3ZTFlODlkAAAUDw%22%7D%2C%22i%22%3A%22llm-obs-panel%22%7D%5D&spanId=6441125655129925386&start=1758652800504&end=1758653700504&paused=false)  | image   |

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
+++ b/packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
@@ -45,6 +45,8 @@ function extractTextAndResponseReasonFromStream (chunks, modelProvider, modelNam
   let message = ''
   let inputTokens = 0
   let outputTokens = 0
+  let cacheReadTokens = 0
+  let cacheWriteTokens = 0
 
   for (const { chunk: { bytes } } of chunks) {
     const body = JSON.parse(Buffer.from(bytes).toString('utf8'))
@@ -100,6 +102,8 @@ function extractTextAndResponseReasonFromStream (chunks, modelProvider, modelNam
     if (invocationMetrics) {
       inputTokens = invocationMetrics.inputTokenCount
       outputTokens = invocationMetrics.outputTokenCount
+      cacheReadTokens = invocationMetrics.cacheReadInputTokenCount
+      cacheWriteTokens = invocationMetrics.cacheWriteInputTokenCount
     }
   }
 
@@ -107,7 +111,9 @@ function extractTextAndResponseReasonFromStream (chunks, modelProvider, modelNam
     message,
     role: 'assistant',
     inputTokens,
-    outputTokens
+    outputTokens,
+    cacheReadTokens,
+    cacheWriteTokens
   })
 }
 
@@ -118,7 +124,9 @@ class Generation {
     choiceId = '',
     role,
     inputTokens,
-    outputTokens
+    outputTokens,
+    cacheReadTokens,
+    cacheWriteTokens
   } = {}) {
     // stringify message as it could be a single generated message as well as a list of embeddings
     this.message = typeof message === 'string' ? message : JSON.stringify(message) || ''
@@ -127,7 +135,9 @@ class Generation {
     this.role = role
     this.usage = {
       inputTokens,
-      outputTokens
+      outputTokens,
+      cacheReadTokens,
+      cacheWriteTokens
     }
   }
 }
--- a/packages/dd-trace/src/llmobs/plugins/openai.js
+++ b/packages/dd-trace/src/llmobs/plugins/openai.js
@@ -86,8 +86,8 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
 
       const promptTokensDetails = tokenUsage.prompt_tokens_details
       if (promptTokensDetails) {
-        const cachedReadTokens = promptTokensDetails.cached_tokens
-        if (cachedReadTokens) metrics.cachedReadTokens = cachedReadTokens
+        const cacheReadTokens = promptTokensDetails.cached_tokens
+        if (cacheReadTokens) metrics.cacheReadTokens = cacheReadTokens
       }
     }
 
--- a/packages/dd-trace/src/llmobs/plugins/openai.js
+++ b/packages/dd-trace/src/llmobs/plugins/openai.js
@@ -86,8 +86,8 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
 
       const promptTokensDetails = tokenUsage.prompt_tokens_details
       if (promptTokensDetails) {
-        const cachedTokens = promptTokensDetails.cached_tokens
-        if (cachedTokens) metrics.cachedTokens = cachedTokens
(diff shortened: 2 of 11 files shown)


packages/dd-trace/test/llmobs/plugins/aws-sdk/bedrockruntime.spec.js
```diff
@@ -131,6 +135,98 @@ describe('Plugin', () => {
             expect(llmobsSpans[0]).to.deepEqualWithMockValues(expected)
           })
         })
+
+        it('should invoke model and handle cache write tokens', async () => {
```
ncybul, line 144: I had to break the prompt caching tests up into their own separate tests because I was having issues where running both requests in the same test was causing the second request to use the same cassette generated by the first request (I even tried making the prompts more distinct and changing various parameters like temperature but none of that worked). I have yet to figure out why having two separate tests fixed this, but I figured this is better than having no test at all!

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -86,8 +86,8 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
 
       const promptTokensDetails = tokenUsage.prompt_tokens_details
       if (promptTokensDetails) {
-        const cachedTokens = promptTokensDetails.cached_tokens
-        if (cachedTokens) metrics.cachedTokens = cachedTokens
+        const cachedReadTokens = promptTokensDetails.cached_tokens
+        if (cachedReadTokens) metrics.cachedReadTokens = cachedReadTokens
```
```suggestion
        const cacheReadTokens = promptTokensDetails.cached_tokens
        if (cacheReadTokens) metrics.cacheReadTokens = cacheReadTokens
```

i think this is the culprit for the failing openai tests

ncybul: oop just saw those tests were failing, thanks for catching!

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
sabrenner, line 1: there are a couple spots i remember adding for the streamed responses that maybe didn't have their token metrics in the default field at the bottom:
- [amazon models](https://github.com/DataDog/dd-trace-js/blob/12e413ee88d7b0cd8d2f1531dd54fb3f18c9345c/packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js#L56-L57)
- [anthropic models](https://github.com/DataDog/dd-trace-js/blob/12e413ee88d7b0cd8d2f1531dd54fb3f18c9345c/packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js#L76-L77)

evidently the anthropic models are fine bc of the tests u added, but might just be good to quickly double check locally that the amazon model also sends the cached token metrics in the `body['amazon-bedrock-invocationMetrics']` as well!

ncybul: (reiterating what we talked about offline) -- it looks like prompt caching is only supported for the Amazon Nova models, but when I went to verify that it works for that model types, I got the following weird trace, so it does not seem to be supported in general for the Bedrock integration

image

I think this will take it's own investigation / PR, so I made this [ticket](https://datadoghq.atlassian.net/browse/MLOB-3987) and will follow up there.

PR #6509, opened 2025-09-24
[test optimization] Add file system cache for `git` commands

juan-fernandez:
### What does this PR do?

Add a file system cache for `git` operations for test optimization.

### Motivation
Monorepo tools like https://microsoft.github.io/lage/ make it so that running multiple test sessions within the same CI job is easy. `dd-trace` does a lot of operations at the test session level:

* git metadata upload
* git tags extraction
* requests to datadog's backend

Since tools like `lage` will run potentially thousands of test sessions, the `dd-trace` instance in each of these test sessions will repeat these operations. 

This PR makes it so that if you pass `DD_EXPERIMENTAL_TEST_OPT_GIT_CACHE_ENABLED`, a file system cache for `git` operations will be used. 

### Plugin Checklist

- [x] Integration tests.


watson:
Why are you not adding these to `config.js`?

juan-fernandez:
> Why are you not adding these to `config.js`?

no particular reason. Going through `config.js` would make the code harder to read IMO, but it'd be indeed more consistent. We can do that once we're sure this feature is going to be shipped, as right now it's just an experiment

PR #6513, opened 2025-09-24
[WIP] Fixing esbuild + esm

uurien:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/datadog-esbuild/src/utils.js
```diff
@@ -0,0 +1,195 @@
+'use strict'
+
+// Based in import-in-the-middle
```
uurien: What should we write here to indicated that the code is copied and modified from iitm? as IITM is not datadog property anymore I think we should mention it here.

integration-tests/esbuild/esm.integration.spec.js
```diff
@@ -0,0 +1,135 @@
...
+      if (span.type === 'web') {
+        return span
+      }
+    }
+  }
+  throw new Error('web span not found')
+}
+
+esbuildVersions.forEach((version) => {
+  describe('ESM is build and runs as expected in a sandbox', () => {
```
CarlesDD:
```suggestion
  describe('ESM is built and runs as expected in a sandbox', () => {
```

integration-tests/esbuild/esm.integration.spec.js
```diff
@@ -0,0 +1,135 @@
...
+        cwd,
+        env: {
+          DD_TRACE_AGENT_URL: `http://localhost:${agent.port}`
+        },
+        stdio: 'pipe',
+      })
+
+      await Promise.all([
+        agent.assertMessageReceived(({ payload }) => {
+          // http web spans are creaed in bundled application
```
CarlesDD: The typo is present in all the comments
```suggestion
          // http web spans are created in bundled application
```

.github/workflows/platform.yml
```diff
@@ -36,6 +36,18 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: ./.github/actions/node/oldest-maintenance-lts
+      - uses: ./.github/actions/install
+      - run: yarn test:esbuild:ci
+      - uses: ./.github/actions/node/latest
+      - run: yarn test:esbuild:ci
+      - uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
+
+
```
watson: Nit:

```suggestion
```

integration-tests/esbuild/build-and-test-koa.mjs
```diff
@@ -1,12 +1,13 @@
 #!/usr/bin/env node
 
 import fs from 'fs/promises'
-
```
watson, line 4: @BridgeAR I've seen you starting to add these separations between Node.js core imports, npm imports, and local imports. Is this a convention we officially follow, or just your personal preferences?

integration-tests/esbuild/build.esm-hono-output-cjs.mjs
```diff
@@ -0,0 +1,42 @@
...
+  outfile: SCRIPT,
+  minify: false,
+  bundle: true,
+  platform: 'node',
+  target: 'node22',
+  plugins: [ddPlugin],
+  format: 'cjs',
+  external
+}).catch((err) => {
+  process.exit(1)
```
watson: What's the reason for adding this `catch`? If it's not there, I would assume the process would crash with an unhandled rejection error. If it's there in its current form, it will still exit the process with a non-zero exit code, but the error itself will be hidden, which might make debugging harder. But maybe this is run programmatically in which case we just care about the exit code? (same question for all the new `.mjs` files you added in this directory)

integration-tests/esbuild/build-and-test-koa.mjs
```diff
@@ -1,12 +1,13 @@
...
 import * as esbuild from 'esbuild'
 import assert from 'assert'
-
+import versions from '../../version.js'
 import ddPlugin from '../../esbuild.js'
 
+const { NODE_MAJOR } = versions
+
```
watson, line 12: Nit:

```suggestion
import { NODE_MAJOR } from '../../version.js'
import ddPlugin from '../../esbuild.js'

```

packages/datadog-esbuild/src/utils.js
watson: I usually write something along the lines of what you did here. Maybe a bit more format, eg:

```
// The content of this file is copied from the `import-in-the-middle` package with minor modifications (https://www.npmjs.com/package/import-in-the-middle)
```

integration-tests/esbuild/build-and-test-koa.mjs
BridgeAR: I just started doing that recently when I touched a file as personal style.

I think we should have it as a recommendation and later implement a lint rule, if we agree upon it across the repo.

integration-tests/esbuild/build-and-test-koa.mjs
```diff
@@ -27,7 +28,12 @@ try {
   // Verify instrumentation
   const data = await fs.readFile('./outfile.js', 'utf8')
 
-  assert.match(data, /^ {8}package: "koa",$/m, 'Bundle should contain the koa instrumentation')
+  if (NODE_MAJOR >= 22) {
+    // it is resolved as ESM module only in node 22+, becaues the require.resolve accepts conditions in node 22+
+    assert.match(data, /register.*koa.mjs".*"koa"\);$/m, 'Bundle should contain the koa ESM instrumentation')
```
BridgeAR, line 35: I am not sure I understand this properly.

Was it working before for non-ESM esbuilds in these Node.js versions?

And should the test be changed to either only work for ESM and another one only for CJS? I do not remember the details right away for the situation where I wrote the test.

integration-tests/esbuild/esm.integration.spec.js
```diff
@@ -0,0 +1,135 @@
...
+  }
+  throw new Error('web span not found')
+}
+
+esbuildVersions.forEach((version) => {
+  describe('ESM is built and runs as expected in a sandbox', () => {
+    let sandbox, agent, cwd
+
+    before(async () => {
+      sandbox = await createSandbox([`esbuild@${version}`, 'hono', '@hono/node-server'], false, [__dirname])
```
BridgeAR, line 29: Could we use the new vary helper? that way we would also test multiple different import possibilities.

integration-tests/esbuild/esm.integration.spec.js
```diff
@@ -0,0 +1,135 @@
...
+    for (const span of trace) {
+      if (span.type === 'web') {
+        return span
+      }
+    }
+  }
+  throw new Error('web span not found')
+}
+
+esbuildVersions.forEach((version) => {
```
BridgeAR, line 24: I wonder if we rebuild withVersions quite a lot here. Could we use that instead?

packages/datadog-esbuild/src/utils.js
BridgeAR: Could we somehow use IITM directly? We could potentially add a feature there?

I ask, since I fear it might at some point deviate and we do not notice to update the code here accordingly.

packages/dd-trace/test/plugins/versions/package.json
```diff
@@ -88,6 +88,7 @@
     "dd-trace-api": "1.0.0",
     "ejs": "3.1.10",
     "elasticsearch": "16.7.3",
+    "esbuild": "0.25.10",
```
BridgeAR: This is now duplicated in two package.json files. I think we should only have it in one spot.

packages/datadog-esbuild/src/utils.js
uurien: > Could we somehow use IITM directly?

I tried, but the methods that I wanted are private and I couldn't use the methods directly, this PR is using some iitm methods, but the things that are here aren't there.
Anyway, this is copy & modifying some things from iitm.

> We could potentially add a feature there?

tbh, I am far to be confident enough to open a PR in IITM to properly support esbuild

> I ask, since I fear it might at some point deviate and we do not notice to update the code here accordingly.

This definitely could happen in the future, ideally we will totally move to orchestrionjs before it happens.

integration-tests/esbuild/build-and-test-koa.mjs
uurien: I tried it initially, but as `versions.js` is a cjs file, and this is a ESM file, we can't partially import a file.

integration-tests/esbuild/build-and-test-koa.mjs
uurien: Yeah, this is one of the weirdest thing in this PR.

previously, for all node versions, we were using `require.resolve()` to get the path to the file that we want rewrite, this was always returning the ESM file.

Eventually could happen that indirectly two versions of the same file (CJS and ESM) were loaded.

With the new `require.resolve(path, { conditions })` this could be mitigated, **but** that works only in node >= 22. 

This change has caused that this test is loading now the mjs version of koa instead of the cjs one.

integration-tests/esbuild/esm.integration.spec.js
uurien: Just following the same pattern is followed in `integration-tests/esbuild/index.spec.js`

Anyway, the behaviour of `withVersions` is a bit different that what we want here with `createSandbox`:
* `withVersions` requries execution of the services job to download the dependencies in `versions` directory, and they could be required from the repository.
* with `createSandbox`, `npm pack` is executed to create a clean tracer as we will have in npm, and will use that with the libraries that are going to be installed in a temporal directory out of the current tracer directory. This makes the test cleaner and for things related with esbuild, more reliable.

integration-tests/esbuild/esm.integration.spec.js
uurien: what is the new vary helper?

packages/dd-trace/test/plugins/versions/package.json
uurien: True, I modified the test in `index.spec.js` to use this and remove the reference from the other package.json, because this is needed to run `createSandbox`.

integration-tests/esbuild/esm.integration.spec.js
BridgeAR: The `varySandbox` placed in the integration-tests/helper.js file duplicates a test file allowing to test different ESM loading types.

integration-tests/esbuild/build-and-test-koa.mjs
BridgeAR: I see. I guess that makes sense and is fine. Thanks.

packages/datadog-esbuild/index.js
```diff
@@ -249,34 +274,70 @@ ${build.initialOptions.banner.js}`
...
+            internal: data.internal,
+            context: { format: 'module' }
+          })
+
+          const iitmPath = require.resolve('import-in-the-middle/lib/register.js')
+          const toRegister = data.internal ? args.path : pathToFileURL(args.path)
+          contents = `
+import { register } from ${JSON.stringify(iitmPath)}
+import * as namespace from ${JSON.stringify(args.path)}
+// Mimic a Module object (https://tc39.es/ecma262/#sec-module-namespace-objects).
```
tlhunter:
```suggestion
          // Mimic a Module object (https://tc39.es/ecma262/#sec-module-namespace-objects).
          contents = `
import { register } from ${JSON.stringify(iitmPath)}
import * as namespace from ${JSON.stringify(args.path)}
```

Can you hoist the comment out? That way it won't potentially appear many times in the bundle.

tlhunter: Also, can you throw in semicolons? It's a little more defensive and I'm always worried about what kind of code could live in userland.

packages/datadog-esbuild/src/utils.js
```diff
@@ -0,0 +1,198 @@
...
+      setters.set(name, setter)
+    }
+  }
+
+  for (const n of exportNames) {
+    if (n === 'default' && excludeDefault) continue
+
+    if (isStarExportLine(n) === true) {
+      // export * from 'wherever'
+      const [, modFile] = n.split('* from ')
```
tlhunter, line 116: Not sure if it's a big deal but user code could look like this:

```js
import    *   from  \t  'foo'
```

uurien: what is returned by getExport is clean, it doesn't have any `\t` nor `\n`, neither the `'` or `"` of `export * from 'whatever'`.

integration-tests/esbuild/esm.integration.spec.js
uurien: ok, I've checked the method you are mentioning. I think that we can't use it for this use case, because the affected files for this PR are files in modules, like the files in "hono" or "@hono/node-server", the files that we are creating in are just to call to these files.

PR #6515, opened 2025-09-24
test: use promise based fs functions in integration tests helpers

watson:
### What does this PR do?

As the title says...

### Motivation

I'm refactoring a few areas of this file, and this will make that refactoring simpler.

### DIFF
--- a/integration-tests/helpers/index.js
+++ b/integration-tests/helpers/index.js
@@ -5,7 +5,8 @@ const childProcess = require('child_process')
 const { fork, spawn } = childProcess
 const exec = promisify(childProcess.exec)
 const http = require('http')
-const fs = require('fs')
+const { existsSync, readFileSync, unlinkSync } = require('fs')
+const fs = require('fs/promises')
 const { builtinModules } = require('module')
 const os = require('os')
 const path = require('path')
@@ -218,7 +219,7 @@ async function createSandbox (dependencies = [], isGitRepo = false,
   const out = path.join(folder, `dd-trace-${version}.tgz`)
   const allDependencies = [`file:${out}`].concat(cappedDependencies)
 
-  fs.mkdirSync(folder)
+  await fs.mkdir(folder)
   const preferOfflineFlag = process.env.OFFLINE === '1' || process.env.OFFLINE === 'true' ? ' --prefer-offline' : ''
   const addCommand = `yarn add ${allDependencies.join(' ')} --ignore-engines${preferOfflineFlag}`
   const addOptions = { cwd: folder, env: restOfEnv }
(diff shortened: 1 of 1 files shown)


integration-tests/helpers/index.js
```diff
@@ -250,14 +251,14 @@ async function createSandbox (dependencies = [], isGitRepo = false,
 
   if (isGitRepo) {
     await exec('git init', { cwd: folder })
-    fs.writeFileSync(path.join(folder, '.gitignore'), 'node_modules/', { flush: true })
+    await fs.writeFile(path.join(folder, '.gitignore'), 'node_modules/', { flush: true })
     await exec('git config user.email "john@doe.com"', { cwd: folder })
     await exec('git config user.name "John Doe"', { cwd: folder })
     await exec('git config commit.gpgsign false', { cwd: folder })
```
BridgeAR, line 257:
```suggestion
    await exec('git config user.email "john@doe.com"', { cwd: folder })
    await exec('git config user.name "John Doe"', { cwd: folder })
    await exec('git config commit.gpgsign false', { cwd: folder })
    await fs.writeFile(path.join(folder, '.gitignore'), 'node_modules/', { flush: true })
```

I think it would be better to add it afterwards.

We can also combine the git commands to a single child process call. I understand if that's not done in this PR though.

integration-tests/helpers/index.js
```diff
@@ -250,14 +251,14 @@ async function createSandbox (dependencies = [], isGitRepo = false,
...
 
     // Create a unique local bare repo for this test
     const localRemotePath = path.join(folder, '..', `${path.basename(folder)}-remote.git`)
-    if (!fs.existsSync(localRemotePath)) {
+    if (!existsSync(localRemotePath)) {
       await exec(`git init --bare ${localRemotePath}`)
```
BridgeAR, line 262: In theory, this could have a race condition where the call still has the path and when git init is called, it does not exist anymore.

watson: I know, but there's no good way to keep the existing functionality by doing it the correct way, so I specifically left this one to highlight that we're using something that's deprecated instead of switching to `fs.access` or `fs.stat`. 

Ideally we want `git init --bare` to exit with a non-zero exit code if `localRemotePath` already exists, but there's no way to have it do that. It will reset the `HEAD` of the repo at `localRemotePath` to point at the top of the main branch, if it's run on an already existing repo. Otherwise it will leave it alone. So maybe that's ok for our use-case, but I wasn't sure, and hence left it as is.

integration-tests/helpers/index.js
watson: I didn't want to change the functionality of this PR to keep it un-opinionated, but what is the benefit of changing the order here?

BridgeAR: The order does not change anything functionality wise. I would just like to combine the exec calls into one. That is more obvious after changing the order. It's fine though.

BridgeAR (approved):

integration-tests/helpers/index.js
watson: Ah ok, I got confused about your suggested diff because it re-ordered the lines. But yes, we could do it in a single exec call. There's a bunch of places in this file where that could be done, but I'll leave that for now.

PR #6517, opened 2025-09-24
SCI embedding = read Git metadata from `.git` folder

olivwalsh:
### What does this PR do?
Read from `.git` folder for Git metadata if `git.commit.sha` and `git.repository_url` were not found in the environment variables nor in the `git.properties` file.

### Motivation
Enable users copying `.git/` files to their deployment servers to have their traces tagged with Git metadata to leverage source code related features.


packages/dd-trace/test/config.spec.js
```diff
@@ -2436,9 +2439,10 @@ describe('Config', () => {
...
     it('does not crash if git.properties is not available', () => {
       process.env.DD_GIT_PROPERTIES_FILE = '/does/not/exist'
-      const config = new Config({})
-      expect(config).to.have.property('commitSHA', undefined)
-      expect(config).to.have.property('repositoryUrl', undefined)
+      expect(() => {
+        const config = new Config({})
+        expect(config).to.be.an('object')
+      }).to.not.throw()
     })
```
olivwalsh, line 2447: I modified this test as we only want to test that applying the config does not crash if the git.properties file was not found.
Also with the changes introduced in this PR, if the git.properties file is not found, Git tagging might still be applied if information is present in the .git folder.

packages/dd-trace/src/git_properties.js
```diff
@@ -32,4 +34,30 @@ function getGitMetadataFromGitProperties (gitPropertiesString) {
...
-module.exports = { getGitMetadataFromGitProperties, removeUserSensitiveInfo }
+function getGitRepoUrlFromGitConfig (gitConfigContent) {
+  if (!gitConfigContent) {
+    return
+  }
+
+  // Look for [remote "origin"] section and extract the URL
+  const repositoryUrlMatch = gitConfigContent.match(gitRemoteOriginRegex)
+
+  if (repositoryUrlMatch && repositoryUrlMatch.length > 1) {
```
watson:
```suggestion
  if (repositoryUrlMatch?.length > 1) {
```

packages/dd-trace/src/git_properties.js
```diff
@@ -2,6 +2,8 @@
 
 const commitSHARegex = /git\.commit\.sha=([a-f\d]{40})/
 const repositoryUrlRegex = /git\.repository_url=([\w\d:@/.-]+)/
+const gitRemoteOriginRegex = /\[remote\s+"origin"\]\s*\n\s*url\s*=\s*([\w\d:@/.-]+)/
```
watson: The git config file can contain both Windows style and UNIX style line breaks and the content of the file is case-insensitive, so the regex needs to account for that. Additionally, though unlikely, there's no guarantee that the first key/value pair after `[remote "origin"]` is the `url`. I think a safer approach would be to simply parse the entire file like so:

```js
const remoteOriginRegex = /^\[remote\s+"origin"\]/i

function getRemoteOriginURL (gitConfigContent) {
  const lines = gitConfigContent.split('\n')
  let index = 0

  // find the remote origin section
  for (; index < lines.length; index++) {
    const line = lines[index]
    if (line[0] !== '[') continue // fast path
    if (remoteOriginRegex.test(line)) break
  }

  // find the url key/value
  for (; index < lines.length; index++) {
    const line = lines[index]
    if (line[0] === '[') return // abort, section didn't contain a url
    const splitAt = line.indexOf('=')
    if (splitAt === -1) continue
    const key = line.slice(0, splitAt).trim().toLowerCase()
    if (key !== 'url') continue
    return line.slice(splitAt + 1).trim()
  }
}
```

The only thing this algorithm doesn't handle, as far as I can see, is a multi-line value, which technically is allowed by the git config spec. But I can't see how that would make sense for the URL field specifically.

packages/dd-trace/test/config.spec.js
```diff
@@ -35,6 +36,7 @@ describe('Config', () => {
   const BLOCKED_TEMPLATE_GRAPHQL_PATH = require.resolve('./fixtures/config/appsec-blocked-graphql-template.json')
   const BLOCKED_TEMPLATE_GRAPHQL = readFileSync(BLOCKED_TEMPLATE_GRAPHQL_PATH, { encoding: 'utf8' })
   const DD_GIT_PROPERTIES_FILE = require.resolve('./fixtures/config/git.properties')
+  const DD_GIT_FOLDER_PATH = path.join(__dirname, 'fixtures/config/git-folder/')
```
watson: Either do `${__dirname}fixtures/config/git-folder/` in case Windows support isn't important, or use `join` to join each sub-folder. I would always recommend the latter. Mixing the two styles is confusing.

```suggestion
  const DD_GIT_FOLDER_PATH = path.join(__dirname, 'fixtures', 'config', 'git-folder')
```

packages/dd-trace/src/config.js
```diff
@@ -420,6 +424,53 @@ class Config {
...
+        const gitHeadPath = path.join(DD_GIT_FOLDER_PATH, 'HEAD')
+        try {
+          const gitHeadContent = fs.readFileSync(gitHeadPath, 'utf8')
+          if (gitHeadContent) {
+            const gitHeadRef = getGitHeadRef(gitHeadContent)
+            const gitHeadRefPath = path.join(DD_GIT_FOLDER_PATH, gitHeadRef)
+            try {
+              const gitHeadRefContent = fs.readFileSync(gitHeadRefPath, 'utf8')
+              if (gitHeadRefContent) {
+                const gitHeadSha = gitHeadRefContent.trim()
```
watson: For testing and maintainability, I would encapsulate this in a function called `resolveGitHeadSHA` or something like that that I would place in `packages/dd-trace/src/git_properties.js`. This function also needs to handle the case where the `.git/HEAD` file doesn't contain `ref:` but instead a SHA-1 hash, which it can do in case of a detached head.

packages/dd-trace/test/config.spec.js
watson: Nit: Why not move these two inside the `context('sci embedding', () => {...}` block next to `DUMMY_COMMIT_SHA` and `DUMMY_REPOSITORY_URL`?

packages/dd-trace/src/config.js
```diff
@@ -420,6 +424,53 @@ class Config {
           this.repositoryUrl = this.repositoryUrl || repositoryUrl
         }
       }
+      // otherwise, try to read Git metadata from the .git/ folder
+      if (!this.repositoryUrl || !this.commitSHA) {
+        const DD_GIT_FOLDER_PATH = coalesce(
+          getEnvironmentVariable('DD_GIT_FOLDER_PATH'),
+          `${process.cwd()}/.git/`
```
watson:
```suggestion
          path.join(process.cwd(), '.git')
```

packages/dd-trace/src/config.js
```diff
@@ -420,6 +424,53 @@ class Config {
...
+          if (gitConfigContent) {
+            const repositoryUrl = getGitRepoUrlFromGitConfig(gitConfigContent)
+            this.repositoryUrl = this.repositoryUrl || repositoryUrl
+          }
+        } catch (e) {
+          // Only log error if the user has set a .git/ path
+          if (getEnvironmentVariable('DD_GIT_FOLDER_PATH')) {
+            log.error('Error reading git config: %s', gitConfigPath, e)
+          }
+        }
```
watson: Wrapping the code in an if-block, as there's no reason to run it if `this.repositoryUrl` is present:

```suggestion
        if (!this.repositoryUrl) {
          // try to read git config (repository URL)
          const gitConfigPath = path.join(DD_GIT_FOLDER_PATH, 'config')
          try {
            const gitConfigContent = fs.readFileSync(gitConfigPath, 'utf8')
            if (gitConfigContent) {
              this.repositoryUrl = getGitRepoUrlFromGitConfig(gitConfigContent)
            }
          } catch (e) {
            // Only log error if the user has set a .git/ path
            if (getEnvironmentVariable('DD_GIT_FOLDER_PATH')) {
              log.error('Error reading git config: %s', gitConfigPath, e)
            }
          }
        }
```

packages/dd-trace/src/config.js
```diff
@@ -420,6 +424,53 @@ class Config {
...
+                log.error('Error reading git HEAD ref: %s', gitHeadRefPath, e)
+              }
+            }
+          }
+        } catch (e) {
+          // Only log error if the user has set a .git/ path
+          if (getEnvironmentVariable('DD_GIT_FOLDER_PATH')) {
+            log.error('Error reading git HEAD: %s', gitHeadPath, e)
+          }
+        }
```
watson: Same here, no reason to run all this if `this.commitSHA` is present

watson (commented):
Nice work! I found a few edge-cases when reading `.git/config` and `.git/HEAD`. I've added comments on how to potentially deal with them 😃

packages/dd-trace/src/git_properties.js
olivwalsh: Done in https://github.com/DataDog/dd-trace-js/pull/6517/commits/d42131837effdfbf252b22800172fe96adba7ef3

packages/dd-trace/src/config.js
olivwalsh: done in https://github.com/DataDog/dd-trace-js/pull/6517/commits/e876ec7d59da322df604dabf4d8132d8c7081a6d

packages/dd-trace/src/git_properties.js
watson: Thanks for fixing my off by one error with the `index++` 😅

packages/dd-trace/src/git_properties.js
```diff
@@ -32,4 +39,97 @@ function getGitMetadataFromGitProperties (gitPropertiesString) {
...
+  try {
+    const gitHeadContent = fs.readFileSync(gitHeadPath, 'utf8')
+    if (!gitHeadContent) {
+      return
+    }
+
+    const headContent = gitHeadContent.trim()
+
+    // Handle detached head case
+    if (isValidCommitSHA(headContent)) {
```
watson: If you call `test` on the regex, it handles everything for you that's done by `isValidCommitSHA` + it doesn't allocate a match object in memory, so it's more performant as well:

```diff
-    if (isValidCommitSHA(headContent)) {
+    if (commitSHARegex.test(headContent)) {
```

If you do the same below for the other calls to `isValidCommitSHA` functions, you can just get rid of the function alltogether :)

packages/dd-trace/src/git_properties.js
```diff
@@ -32,4 +39,97 @@ function getGitMetadataFromGitProperties (gitPropertiesString) {
...
+  if (!gitHeadContent) {
+    return
+  }
+
+  // Extract the ref after 'ref: '
+  const gitRefMatch = gitHeadContent.match(gitHeadRefRegex)
+
+  if (gitRefMatch?.length > 1) {
+    return gitRefMatch[1]
+  }
```
watson: Nit: This isn't important, but just a small optimization that you can use if you want 😊

```suggestion
  return gitRefMatch?.[1]
```

watson (commented):
Sorry for the delayed review, I just have a tiny change request for performance reasons, but other than that everything looks amazing 🤩

packages/dd-trace/src/config.js
```diff
@@ -378,12 +380,15 @@ class Config {
     }
 
     if (this.gitMetadataEnabled) {
+      // try to read Git metadata from the environment variables
```
khanayan123: NIT: can we encapsulate this logic in a separate method? this constructor is getting quite large

PR #6519, opened 2025-09-24
test: simplify vary test helper

BridgeAR:
This creates an overload for the vary helper to remove the need for an additional argument. It also improves the types and makes sure the entries do not have to be written out in the tests.

### DIFF
--- a/integration-tests/helpers/index.js
+++ b/integration-tests/helpers/index.js
@@ -275,6 +275,18 @@ async function createSandbox (dependencies = [], isGitRepo = false,
 }
 
 /**
+ * @typedef {{ default: string, star: string, destructure: string }} Variants
+ */
+/**
+ * @overload
+ * @param {object} sandbox - A `sandbox` as returned from `createSandbox`
+ * @param {string} filename - The file that will be copied and modified for each variant.
+ * @param {string} bindingName - The binding name that will be use to bind to the packageName.
+ * @param {string} [namedVariant] - The name of the named variant to use.
+ * @param {string} [packageName] - The name of the package. If not provided, the binding name will be used.
+ * @returns {Variants} A map from variant names to resulting filenames
+ */
+/**
  * Creates a bunch of files based on an original file in sandbox. Useful for varying test files
  * without having to create a bunch of them yourself.
  *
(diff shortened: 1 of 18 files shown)


integration-tests/helpers/index.js
```diff
@@ -284,40 +296,44 @@ async function createSandbox (dependencies = [], isGitRepo = false,
...
     }
     writeFileSync(path.join(sandbox.folder, variantFilename), newFileData)
   }
   return variantFilenames
 }
 
+/**
+ * @type {string[]}
+ */
+varySandbox.variants = ['default', 'star', 'destructure']
```
watson: Nit: I prefer using SCREAMING_SNAKE_CASE for constants like this:


```suggestion
varySandbox.VARIANTS = ['default', 'star', 'destructure']
```

Just to signal to anybody who reads the usage of this that it's a constant. But this isn't a dealbreaker for me, just my personal preference.

integration-tests/helpers/index.js
```diff
@@ -284,40 +296,44 @@ async function createSandbox (dependencies = [], isGitRepo = false,
...
- * depending on the parameters passed through.
- * @param {object} bindingName - The binding name that will be use to bind to the packageName.
- * @param {object} namedVariant - The name of the named variant to use.
- * @param {object} packageName - The name of the package.
- * @returns {object} A map from variant names to resulting filenames
+ * @param {Variants} variants - The variants.
+ * @returns {Variants} A map from variant names to resulting filenames
  */
-function varySandbox (sandbox, filename, variants, bindingName, namedVariant, packageName) {
+function varySandbox (sandbox, filename, variants, namedVariant, packageName = variants) {
```
watson, line 302: Would it make sense to rename `variants` to `variantsOrBindingName`, or is that just noise?

packages/datadog-plugin-elasticsearch/test/integration-test/client.spec.js
```diff
@@ -22,7 +22,7 @@ describe('esm', () => {
       this.timeout(20000)
       sandbox = await createSandbox([`'@elastic/elasticsearch@${version}'`], false, [
         './packages/datadog-plugin-elasticsearch/test/integration-test/*'])
-      variants = varySandbox(sandbox, 'server.mjs', null, 'elasticsearch', null, '@elastic/elasticsearch')
+      variants = varySandbox(sandbox, 'server.mjs', 'elasticsearch', undefined, '@elastic/elasticsearch')
```
watson, line 25: Just out of curiosity, why did you change the `namedVariant` argument from `null` to `undefined`? There's no default value here in the function definition, so I don't see the reason

integration-tests/helpers/index.js
```diff
@@ -284,40 +296,44 @@ async function createSandbox (dependencies = [], isGitRepo = false,
...
+        ? `import * as ${bindingName} from '${packageName}'`
+        : `import * as mod${bindingName} from '${packageName}'; const ${bindingName} = mod${bindingName}.default`,
+      destructure: namedVariant
+        ? `import { ${namedVariant} } from '${packageName}'; const ${bindingName} = { ${namedVariant} }`
+        : `import { default as ${bindingName}} from '${packageName}'`
+    }
+  }
+
   const origFileData = readFileSync(path.join(sandbox.folder, filename), 'utf8')
   const [prefix, suffix] = filename.split('.')
```
watson: Nit: I'm aware this hasn't changed since the old implementation, but if `filename` contains multiple periods, this will fail. Not sure if that's ever the case in reality, but doing it like this would be safer:

```diff
-  const [prefix, suffix] = filename.split('.')
+  const { name: prefix, ext: suffix } = path.parse(filename)
   const variantFilenames = /** @type {Variants} */ ({})

   for (const [variant, value] of Object.entries(variants)) {
-    const variantFilename = `${prefix}-${variant}.${suffix}`
+    const variantFilename = `${prefix}-${variant}${suffix}`
```

integration-tests/helpers/index.js
BridgeAR: I wanted to do the same, it is just not going to be picked up properly by the types in that case. I can mark both docs as overloads which would work, but otherwise it won't

packages/datadog-plugin-elasticsearch/test/integration-test/client.spec.js
BridgeAR: I can change it back to `null`, it was just nicer to mark the optional argument in the type, which accepts undefined, not null.

integration-tests/helpers/index.js
watson: Nah, feel free to leave it as is

integration-tests/helpers/index.js
```diff
@@ -284,40 +296,44 @@ async function createSandbox (dependencies = [], isGitRepo = false,
...
   }
-  variants = variants || defaultVariants
-  for (const variant in variants) {
+
+  const origFileData = readFileSync(path.join(sandbox.folder, filename), 'utf8')
+  const { name: prefix, ext: suffix } = path.parse(filename)
+  const variantFilenames = /** @type {Variants} */ ({})
+
+  for (const [variant, value] of Object.entries(variants)) {
     const variantFilename = `${prefix}-${variant}.${suffix}`
```
watson: The `suffix` now contain the `.`, so it shouldn't be added manually here

```suggestion
    const variantFilename = `${prefix}-${variant}${suffix}`
```

BridgeAR: Yes, two dots are indeed not needed :D

PR #6521, opened 2025-09-24
ci: increase dependabot limit for regular PRs to 100

BridgeAR:


.github/dependabot.yml
```diff
@@ -31,6 +31,7 @@ updates:
       - "/"
     schedule:
       interval: "weekly"
+    open-pull-requests-limit: 100
```
rochdev, line 34: This is for a group with semver minor and even semver patch. It should only be applied to majors.

BridgeAR: That does not matter for this, since the group bundles all minor and patch versions of all dev dependencies in a single PR and all prod dependencies in one PR. The only ones that are outliers are the majors.

rochdev (approved):

PR #6538, opened 2025-09-27
chore: update sinon to v21

BridgeAR:
Sinon now mocks more timers and thus, our tests fail in case they rely on some timers to continue to work as before.

This allows to potentially reduce test runtime in more places in the future, while the solution for now is to just limit the mocked APIs to the old ones.

### DIFF
--- a/packages/dd-trace/test/profiling/profiler.spec.js
+++ b/packages/dd-trace/test/profiling/profiler.spec.js
@@ -5,8 +5,6 @@ const { describe, it, beforeEach, afterEach } = require('tap').mocha
 const sinon = require('sinon')
 const proxyquire = require('proxyquire')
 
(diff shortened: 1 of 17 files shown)


packages/dd-trace/test/profiling/profiler.spec.js
```diff
@@ -208,7 +208,7 @@ describe('profiler', function () {
       clock.tick(interval)
 
       await rejected.catch(() => {})
-      await setTimeout(1)
```
simon-id, line 211: the author of the test wrote `clock.tick()` just above, and `setTimeout()` here, so i'm assuming he didn't write `clock.tick()` here for a reason no ? why change it ?

BridgeAR: I am the author of the setTimeout

PR #6541, opened 2025-09-27
chore: update eslint-plugin-mocha to latest

BridgeAR:
It is actually possible to update the dependency, since we already use a ESM config file.

This deactivates some new rules to prevent significant changes.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -524,7 +524,7 @@ export default [
   },
   {
     name: 'mocha/recommended',
-    ...eslintPluginMocha.configs.flat.recommended,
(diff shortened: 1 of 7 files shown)


simon-id (approved):

eslint.config.mjs
```diff
@@ -524,7 +524,7 @@ export default [
   },
   {
     name: 'mocha/recommended',
```
watson, line 526: Every time we upgrade an ESLint plugin from a version that didn't have a `name` property, we need to check if the new one has one. In this case it did, so this line could have been removed. I'm going to remove it in a fixup PR.

PR #6543, opened 2025-09-27
chore: remove get-port dependency completely

BridgeAR:


packages/datadog-plugin-grpc/test/integration-test/server.mjs
```diff
@@ -2,13 +2,12 @@ import 'dd-trace/init.js'
...
 import * as protoLoader from '@grpc/proto-loader'
 import path from 'path'
-import getPort from 'get-port'
 
 const currentDirectoryPath = path.dirname(new URL(import.meta.url).pathname)
 const parentDirectoryPath = path.resolve(currentDirectoryPath, '..')
 
 let server
-const port = await getPort()
+let port
```
watson: Nit: to be consistant with the changes you made in `packages/datadog-plugin-grpc/test/client.spec.js` and `packages/datadog-plugin-grpc/test/server.spec.js`, I would initialize `port` to `0` here as well (or remove it in the other files)

```suggestion
let port = 0
```

packages/datadog-plugin-moleculer/test/index.spec.js
watson, line 1: Nit: I would add a comment above the call to `getPort` indicating that there's a possibility of a race condition. I know this hasn't changed in this PR, but it would help a future reader who might be tempted to copy the `getPort` function, to think twice before doing so.

yarn.lock
watson, line 1: I don't understand why there's all these changes to to the `yarn.lock` file?!? Are some of using using different versions of `yarn` which means that every time one of us runs `yarn` it might make formatting changes to the file unrelated to the primary change being performed?

BridgeAR: Seems like I had some other changes. I cleaned that up and now the independent changes are gone.

watson (dismissed):

watson (approved):

simon-id (approved):

PR #6554, opened 2025-09-29
[test optimization] Fix issue with `jest.mock` and bypassed modules 

juan-fernandez:
### What does this PR do?

* Check if the library has been mocked before bypassing jest's require module. 

### Motivation
If the user was using `jest.mock` to mock libraries that we instrument, such as `winston`, this resulted in the library not actually being mocked:

```javascript
const winston = require('winston')
const logger = require('./logger')

jest.mock('winston', () => ({
  createLogger: jest.fn(() => ({
    info: jest.fn(),
  })),
  format: {
    simple: jest.fn(() => ({}))
  },
  transports: {
    Console: jest.fn()
  }
}))

describe('winston mock test', () => {
  it('should call winston.createLogger when mocked', () => {
    logger.info('test')
    // !! this was not a mock before the change
    expect(winston.createLogger).toHaveBeenCalledTimes(1)
    expect(logger.info).toHaveBeenCalledTimes(1)
  })
```

### Plugin Checklist

- [x] Integration tests.

### DIFF
--- a/packages/datadog-instrumentations/src/jest.js
+++ b/packages/datadog-instrumentations/src/jest.js
@@ -92,6 +92,7 @@ const newTestsTestStatuses = new Map()
 const attemptToFixRetriedTestsStatuses = new Map()
 const wrappedWorkers = new WeakSet()
 const testSuiteMockedFiles = new Map()
+const bypassedMockedModules = new Set()
 
 const BREAKPOINT_HIT_GRACE_PERIOD_MS = 200
 
@@ -1289,6 +1290,9 @@ addHook({
     const suiteFilePath = this._testPath
 
     shimmer.wrap(result, 'mock', mock => function (moduleName) {
+      if (LIBRARIES_BYPASSING_JEST_REQUIRE_ENGINE.has(moduleName)) {
+        bypassedMockedModules.add(moduleName)
+      }
       if (suiteFilePath) {
         const existingMockedFiles = testSuiteMockedFiles.get(suiteFilePath) || []
         const suiteDir = path.dirname(suiteFilePath)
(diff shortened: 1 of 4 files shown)


packages/datadog-instrumentations/src/jest.js
```diff
@@ -1289,6 +1290,9 @@ addHook({
     const suiteFilePath = this._testPath
 
     shimmer.wrap(result, 'mock', mock => function (moduleName) {
+      if (LIBRARIES_BYPASSING_JEST_REQUIRE_ENGINE.has(moduleName)) {
```
BridgeAR, line 1287:
```suggestion
      if (shouldBypassJestRequireEngine(moduleName)) {
```

While we could also remove the helper, since it's probably also clear without it :)

packages/datadog-instrumentations/src/jest.js
```diff
@@ -1289,6 +1290,9 @@ addHook({
...
+        bypassedMockedModules.add(moduleName)
```
BridgeAR: Would it be possible to remove the mock during runtime as well? I could imagine this happening at times.

If that is not possible, should we just remove the entry from the `LIBRARIES_BYPASSING_JEST_REQUIRE_ENGINE` set? That way there is no need for the additional set.

packages/datadog-instrumentations/src/jest.js
juan-fernandez: yeah I'll remove the helper

packages/datadog-instrumentations/src/jest.js
juan-fernandez: yeah good idea. It's not possible to remove the mock as far as I know, at least not without getting into deep and private `jest` properties

packages/datadog-instrumentations/src/jest.js
juan-fernandez: done in https://github.com/DataDog/dd-trace-js/pull/6554/commits/674b50271cfb95fa77fa96b27a898df5db312ced

packages/datadog-instrumentations/src/jest.js
juan-fernandez: done in https://github.com/DataDog/dd-trace-js/pull/6554/commits/674b50271cfb95fa77fa96b27a898df5db312ced

PR #6555, opened 2025-09-29
[test optimization] Test latest release of `jest`

juan-fernandez:
### What does this PR do?
* Make sure latest happy dom changes work with us. 

### Motivation
Latest release of `jest`: https://github.com/jestjs/jest/releases/tag/v30.2.0 broke our integration tests: https://github.com/DataDog/dd-trace-js/actions/runs/18093612861/job/51480494062?pr=6553 from https://github.com/DataDog/dd-trace-js/pull/6553

### Plugin Checklist
Just check that today's tests pass

### DIFF
--- a/packages/datadog-instrumentations/src/jest.js
+++ b/packages/datadog-instrumentations/src/jest.js
@@ -621,9 +621,18 @@ function getWrappedEnvironment (BaseEnvironment, jestVersion) {
 function getTestEnvironment (pkg, jestVersion) {
   if (pkg.default) {
     const wrappedTestEnvironment = getWrappedEnvironment(pkg.default, jestVersion)
-    pkg.default = wrappedTestEnvironment
-    pkg.TestEnvironment = wrappedTestEnvironment
-    return pkg
+    try {
+      pkg.default = wrappedTestEnvironment
+      pkg.TestEnvironment = wrappedTestEnvironment
+      return pkg
+    } catch (e) {
+      log.warn('Failed to assign the wrappedTestEnvironment to the pkg', e)
+      return {
+        ...pkg,
+        default: wrappedTestEnvironment,
+        TestEnvironment: wrappedTestEnvironment
+      }
+    }
   }
   return getWrappedEnvironment(pkg, jestVersion)
 }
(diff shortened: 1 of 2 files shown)


packages/datadog-instrumentations/src/jest.js
```diff
@@ -621,9 +621,18 @@ function getWrappedEnvironment (BaseEnvironment, jestVersion) {
...
     const wrappedTestEnvironment = getWrappedEnvironment(pkg.default, jestVersion)
-    pkg.default = wrappedTestEnvironment
-    pkg.TestEnvironment = wrappedTestEnvironment
-    return pkg
+    try {
+      pkg.default = wrappedTestEnvironment
+      pkg.TestEnvironment = wrappedTestEnvironment
+      return pkg
+    } catch (e) {
+      log.warn('Failed to assign the wrappedTestEnvironment to the pkg', e)
```
BridgeAR: I guess we will soon very frequently run into this case when people update (it being the default to throw from that moment on). Creating an error is always expensive and instead, we could check for the object being frozen or if the property is configurable or not.

I also wonder if it would be possible to use Object.defineProperty instead of the assignment. That might work, depending on the reason for the assignment to fail.

packages/datadog-instrumentations/src/jest.js
```diff
@@ -621,9 +621,18 @@ function getWrappedEnvironment (BaseEnvironment, jestVersion) {
...
+      return {
```
BridgeAR: Should we mimic the original object? I guess we could copy all properties.

packages/datadog-instrumentations/src/jest.js
juan-fernandez: let me try that

packages/datadog-instrumentations/src/jest.js
juan-fernandez: this should be fixed with https://github.com/DataDog/dd-trace-js/pull/6555/commits/a4c596a1da80befab2b4cf90af4bd4adc1026702

packages/datadog-instrumentations/src/jest.js
juan-fernandez: `Object.defineProperty` does not work. Using a `Proxy` seems to work though: https://github.com/DataDog/dd-trace-js/pull/6555/commits/a4c596a1da80befab2b4cf90af4bd4adc1026702

PR #6556, opened 2025-09-29
chore: remove koalas dependency

BridgeAR:
This refactors all of the rest of calesce calls to use native ?? instead of the custom logic. It also fixes small regressions and adds test cases for these with the former change.


packages/datadog-plugin-aws-sdk/src/base.js
```diff
@@ -275,8 +275,7 @@ function normalizeConfig (config, serviceIdentifier) {
     specificConfig.batchPropagationEnabled ??
     getEnvironmentVariable(`DD_TRACE_AWS_SDK_${serviceId}_BATCH_PROPAGATION_ENABLED`) ??
     config.batchPropagationEnabled ??
-    getEnvironmentVariable('DD_TRACE_AWS_SDK_BATCH_PROPAGATION_ENABLED') ??
-    false
+    getEnvironmentVariable('DD_TRACE_AWS_SDK_BATCH_PROPAGATION_ENABLED')
```
BridgeAR, line 278: The `isTrue()` method would catch that automatically.

packages/dd-trace/src/config.js
```diff
@@ -333,7 +332,7 @@ class Config {
     )
 
     const DD_TRACE_CLOUD_PAYLOAD_TAGGING_MAX_DEPTH = maybeInt(
-      getEnvironmentVariable('DD_TRACE_CLOUD_PAYLOAD_TAGGING_MAX_DEPTH'),
+      getEnvironmentVariable('DD_TRACE_CLOUD_PAYLOAD_TAGGING_MAX_DEPTH') ??
```
BridgeAR, line 335: This slipped through earlier

packages/dd-trace/src/log/index.js
```diff
@@ -111,11 +111,11 @@ const log = {
 
   isEnabled (fleetStableConfigValue, localStableConfigValue) {
     return isTrue(
-      (fleetStableConfigValue ??
+      fleetStableConfigValue ??
       getEnvironmentVariable('DD_TRACE_DEBUG') ??
-      getEnvironmentVariable('OTEL_LOG_LEVEL') === 'debug') ||
-      (localStableConfigValue ??
-      config.enabled)
+      (getEnvironmentVariable('OTEL_LOG_LEVEL') === 'debug' || undefined) ??
```
BridgeAR: This also slipped through earlier. The change would have been caused in case the first three parts would be falsy. In that case the other two values should not be checked anymore. That is now the case again.

packages/dd-trace/src/profiling/config.js
```diff
@@ -175,10 +174,10 @@ class Config {
     )
     checkOptionWithSamplingContextAllowed(this.cpuProfilingEnabled, 'CPU profiling')
 
-    this.samplingInterval = coalesce(options.samplingInterval, 1e3 / 99) // 99hz in millis
+    this.samplingInterval = options.samplingInterval || 1e3 / 99 // 99hz in millis
 
-    this.heapSamplingInterval = coalesce(options.heapSamplingInterval,
-      Number(DD_PROFILING_HEAP_SAMPLING_INTERVAL))
+    this.heapSamplingInterval = options.heapSamplingInterval ??
+      (Number(DD_PROFILING_HEAP_SAMPLING_INTERVAL) || 512 * 1024)
```
BridgeAR, line 180: I moved the default here, as it's also done in other places like that

packages/dd-trace/src/profiling/config.js
```diff
@@ -222,8 +221,7 @@ class Config {
 
     const hasExecArg = (arg) => process.execArgv.includes(arg) || String(NODE_OPTIONS).includes(arg)
 
-    this.asyncContextFrameEnabled = isTrue(options.useAsyncContextFrame ??
-      DD_PROFILING_ASYNC_CONTEXT_FRAME_ENABLED ?? false)
+    this.asyncContextFrameEnabled = isTrue(options.useAsyncContextFrame ?? DD_PROFILING_ASYNC_CONTEXT_FRAME_ENABLED)
```
BridgeAR, line 224: No need for the fallback due to the `isTrue()` method

packages/dd-trace/src/profiling/profilers/events.js
```diff
@@ -41,14 +41,9 @@ function labelFromStrStr (stringTable, keyStr, valStr) {
   return labelFromStr(stringTable, stringTable.dedup(keyStr), valStr)
 }
 
-function getSamplingIntervalMillis (options) {
-  return (options.samplingInterval || 1e3 / 99) // 99Hz
```
BridgeAR, line 45: This was already defined in the config earlier

packages/dd-trace/src/profiling/profilers/wall.js
```diff
@@ -137,7 +137,7 @@ class NativeWallProfiler {
     this.#endpointCollectionEnabled = !!options.endpointCollectionEnabled
     this.#flushIntervalMillis = options.flushInterval || 60 * 1e3 // 60 seconds
     this.#logger = options.logger
-    this.#samplingIntervalMicros = (options.samplingInterval || 1e3 / 99) * 1000 // 99hz
+    this.#samplingIntervalMicros = options.samplingInterval * 1000
```
BridgeAR: The default should already be set

packages/dd-trace/src/config.js
```diff
@@ -998,8 +997,7 @@ class Config {
     this._setBoolean(opts, 'runtimeMetricsRuntimeId', options.runtimeMetricsRuntimeId)
     this._setArray(opts, 'sampler.spanSamplingRules', reformatSpanSamplingRules(options.spanSamplingRules))
     this._setUnit(opts, 'sampleRate', options.sampleRate ?? options.ingestion.sampleRate)
-    const ingestion = options.ingestion || {}
-    opts['sampler.rateLimit'] = coalesce(options.rateLimit, ingestion.rateLimit)
+    opts['sampler.rateLimit'] = maybeInt(options.rateLimit ?? options.ingestion.rateLimit)
```
simon-id, line 1000: needs optional chaining to be iso with the previous code no ?
```suggestion
    opts['sampler.rateLimit'] = maybeInt(options.rateLimit ?? options.ingestion?.rateLimit)
```

simon-id, line 1000: needs optional chaining to be iso with the previous code no ?
```suggestion
    opts['sampler.rateLimit'] = maybeInt(options.rateLimit ?? options.ingestion?.rateLimit)
```

BridgeAR: No, the code is actually always assigning an empty object to it and the line above is also directly accessing properties on ingestion. It would fail above if it would be a problem here.

simon-id: so the previous code was unnecessary ye ?

BridgeAR: Yes, feel free to have a look at the code :)

packages/dd-trace/src/log/index.js
szegedi: `(a || undefined) ?? b` isn't equivalent to `a || b` when `a` is guaranteed boolean?

szegedi (approved):
LGTM, with a small nit.

packages/dd-trace/src/log/index.js
BridgeAR: Ha... I should not question my own code too much. Yes, that is correct and I am unsure why I questioned that :D

simon-id (approved):
LGTM but indeed maybe Attila should approve too

PR #6559, opened 2025-09-29
[SVLS-7305] Instrument Azure Event Hubs

jcstorms1:
### What does this PR do?

Adds  support for distributed tracing in certain Azure environments
- Instruments the `EventHubProducerClient` in the Azure Event Hubs SDK
- Instruments the consumer triggers in the Azure Function SDK
- Separates Azure Function tests to isolate each managed service 

New integration for Azure Event Hubs configurable by:
`DD_TRACE_AZUREEVENTHUBS_ENABLED` (default true) enables/disables the integration
`DD_TRACE_AZURE_EVENTHUBS_BATCH_LINKS_ENABLED` (default true) enables/disables Span Linking and TryAdd spans

### Motivation

Azure distributed tracing rollout

### Plugin Checklist

- [x] Unit tests.
- [x] Integration tests.
- [ ] Benchmarks.
- [x] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-plugin-azure-functions/test/integration-test/eventhubs-test/eventhubs.spec.js
+++ b/packages/datadog-plugin-azure-functions/test/integration-test/eventhubs-test/eventhubs.spec.js
@@ -314,7 +314,7 @@ describe('esm', () => {
       })
     }).timeout(60000)
 
-    it('should not contain tryAdd spans or add span links to batches when batch links are disabled', async () => {
+    it('should not create a tryAdd span or add span links to batches when batch links are disabled', async () => {
       const envArgs = {
         PATH: `${sandbox.folder}/node_modules/azure-functions-core-tools/bin:${process.env.PATH}`,
         DD_TRACE_AZURE_EVENTHUBS_BATCH_LINKS_ENABLED: false


.github/workflows/serverless.yml
```diff
@@ -115,10 +147,16 @@ jobs:
               MSSQL_SA_PASSWORD: "Localtestpass1!"
     env:
       PLUGINS: azure-functions
-      SERVICES: azureservicebusemulator,azuresqledge
+      SERVICES: azureservicebusemulator,azuresqledge,azurite,azureeventhubsemulator
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
-      - uses: ./.github/actions/node/active-lts
+      - run: |
+          docker cp \
```
jcstorms1, line 154: We need to add a config file for the Event Hubs emulator, but in github workflows we can't mount the file. The workaround is to use `docker cp` and restart the service. https://github.com/orgs/community/discussions/42127#discussioncomment-7591609

packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -86,6 +93,16 @@ function getMetaForTrigger ({ functionName, methodName, invocationContext }) {
       'resource.name': `ServiceBus ${functionName}`,
       'span.kind': 'consumer'
     }
+  } else if (triggerMap[methodName] === 'EventHubs') {
+    const partitionContext = invocationContext.triggerMetadata.triggerPartitionContext
+    meta = {
+      ...meta,
+      'messaging.destination.name': partitionContext.eventHubName,
+      'messaging.operation': 'receive',
+      'messaging.system': 'eventhub',
```
pablomartinezbernardo:
```suggestion
      'messaging.system': 'eventhubs',
```

The producer is correct though

pablomartinezbernardo (commented):
This comment likely also applies to the python tracer, but I hadn't thought about it until now: If I'm understanding correctly, the integration doesn't cover output bindings, which are actually sent from the host process within .NET. Is that an expected limitation? How will we communicate to customers that they should use the SDK directly rather than using output bindings? Is it not possible to trace output bindings?

packages/datadog-plugin-azure-event-hubs/src/producer.js
```diff
@@ -0,0 +1,78 @@
...
+
+    if (ctx.functionName === 'sendBatch') {
+      const eventData = ctx.eventData
+      const eventDataLength = eventData.length || eventData._context.connection._eventsCount
+      span.setTag('messaging.operation', 'send')
+      span.setTag('messaging.batch.message_count', eventDataLength)
+
+      if (eventData.constructor.name !== 'EventDataBatchImpl' && Array.isArray(eventData)) {
+        eventData.forEach(event => {
+          injectTraceContext(this.tracer, span, event)
```
pablomartinezbernardo, line 52: I'm struggling to follow this along, it looks like when sending an array of EventData the context is injected for each of them always, regardless of batch links setting. Then in `setSpanLinks` regardless of the batch links setting, context is extracted and span links are created.

If that understanding is correct, span links should be created always when sending batches in the form of arrays.

But that seems to contradict the test `should not add span links when they are disabled`. Where is my reasoning wrong? 🤔

packages/datadog-plugin-azure-functions/test/integration-test/eventhubs-test/eventhubs.spec.js
```diff
@@ -0,0 +1,368 @@
...
+    }).timeout(60000)
+
+    it('should not add span links when they are disabled', async () => {
+      const envArgs = {
+        PATH: `${sandbox.folder}/node_modules/azure-functions-core-tools/bin:${process.env.PATH}`,
+        DD_TRACE_AZURE_EVENTHUBS_BATCH_LINKS_ENABLED: false
+      }
+      proc = await spawnPluginIntegrationTestProc(sandbox.folder, 'func', ['start'], agent.port, undefined, envArgs)
+      return curlAndAssertMessage(agent, 'http://127.0.0.1:7071/api/eh2-eventdata', ({ headers, payload }) => {
+        expect(payload[1][0]).to.not.have.property('_dd.span_links')
```
pablomartinezbernardo: Would doing these tests also for batches that are not arrays be useful?

packages/datadog-plugin-azure-event-hubs/src/producer.js
```diff
@@ -0,0 +1,78 @@
...
+      meta: {
+        component: 'azure-event-hubs',
+        'messaging.system': 'eventhubs',
+        'messaging.destination.name': entityPath,
+        'network.destination.name': qualifiedNamespace,
+      }
+    }, ctx)
+
+    if (ctx.functionName === 'tryAdd') {
+      span._spanContext._name = 'azure.eventhubs.create'
```
pablomartinezbernardo, line 31: I believe we have talked in the past about `azure.eventhubs.create` spans not even being created at all when batch links are not enabled.

PR #6561, opened 2025-09-29
add support for test version dependency resolutions

rochdev:
### What does this PR do?

Add support for test version dependency resolutions.

### Motivation

A new version of `is-generator-function` (1.1.1) was released which depends on `generator-function` (2.0.0) which breaks Koa 2.x completely, so we need a way to pin the underlying dependency.

### Additional Notes

Ideally we'd be able to pin all transitive dependencies always through something like a lockfile per test version, but that would require a larger change, so for now this workaround will have to do.

### DIFF
--- a/packages/dd-trace/test/plugins/versions/package.json
+++ b/packages/dd-trace/test/plugins/versions/package.json
@@ -3,9 +3,6 @@
   "version": "1.0.0",
   "license": "BSD-3-Clause",
   "private": true,
-  "resolutions": {
-    "is-generator-function": "1.1.0"
-  },
   "dependencies": {
     "@ai-sdk/openai": "2.0.34",
     "@openai/agents": "0.1.4",


BridgeAR (commented):
Things are working again. I don't think we should pin it therefore.

BridgeAR (commented):
I am not really convinced about adding this while we do not use it. If this comes up again, I think we should look at it in a similar way as we did anyway and adding functionality that is not used is not good for me.

rochdev:
> I am not really convinced about adding this while we do not use it. If this comes up again, I think we should look at it in a similar way as we did anyway and adding functionality that is not used is not good for me.

I agree in principle, but I'd rather have the code already there and ready whenever we need it than not, or having to research this again by someone who doesn't know about this PR. Let's not forget we often get weird situations that can be surprising and sometimes difficult to address, so any amount of readiness already built-in is a good thing, especially when we know it happened in the past.

BridgeAR (changes requested):
I do not think anyone would find this with the current suggestion. If you or me would face a similar situation, we will probably remember either way. Another person won't know about it and won't find it and has to invest the same time due to that.

With that in mind, I do not think we should land this.

rochdev:
> Wouldn't adding it to the resolutions field in packages/dd-trace/test/plugins/versions/package.json fix it?

Unfortunately resolutions don't work at the workspace level. At the end of the day, the right solution is to lock all transitive dependencies, but that's more work that I don't have time to do right now, but I'll try to take time in the following weeks.

rochdev:
Let's revisit this later as a long-term solution would be best.

PR #6563, opened 2025-09-30
Update ldapjs docker image to use bitnamilegacy

simon-id:
old image does not exist anymore


BridgeAR (approved):
I think this is fine as a short term fix, we just have to find another one that is regularly maintained.

simon-id:
@BridgeAR yeah i'm already on it ;)

PR #6566, opened 2025-09-30
[test optimization] Fix `happy-dom` testing issues with latest `jest` release

juan-fernandez:
### What does this PR do?
* Do not run `happy-dom` integration tests for node 18. 
* Remove `happy-dom` plugin tests. 

### Motivation
`happy-dom` released a breaking change for >=19: https://github.com/capricorn86/happy-dom/releases/tag/v19.0.0.

Node 18 is no longer supported for CJS. 

Additionally, I've removed the plugin tests for `happy-dom`: these tests do not reflect the way the library is used and shouldn't have been updated in https://github.com/DataDog/dd-trace-js/pull/6208. These tests should be considered legacy and should eventually be removed.

### DIFF
--- a/integration-tests/jest/jest.spec.js
+++ b/integration-tests/jest/jest.spec.js
@@ -2288,7 +2288,8 @@ describe('jest CommonJS', () => {
       })
     })
     // happy-dom>=19 can only be used with CJS from node 20 and above
-    (NODE_MAJOR < 20 ? it.skip : it)('works with happy-dom', async () => {
+    const happyDomTest = NODE_MAJOR < 20 ? it.skip : it
+    happyDomTest('works with happy-dom', async () => {
       // Tests from ci-visibility/test/ci-visibility-test-2.js will be considered new
       receiver.setKnownTests({
         jest: {


calvinbayer (dismissed):

integration-tests/jest/jest.spec.js
```diff
@@ -2286,78 +2287,80 @@ describe('jest CommonJS', () => {
...
-        }
-      })
-      const NUM_RETRIES_EFD = 3
-      receiver.setSettings({
-        early_flake_detection: {
-          enabled: true,
-          slow_test_retries: {
-            '5s': NUM_RETRIES_EFD
+    // happy-dom>=19 can only be used with CJS from node 20 and above
+    if (NODE_MAJOR >= 20) {
```
BridgeAR: Should we instead just skip the test for those versions? And do we not have a test matrix where older versions of happy-dom would also be tested?

juan-fernandez: > Should we instead just skip the test for those versions? 

sure, I can do that

> And do we not have a test matrix where older versions of happy-dom would also be tested?

not at the moment, no. We don't test the older versions for jest integration tests. This was a conscious decision back then, I think because we were content with the plugin tests testing older versions. But I want to eventually get rid of the plugin tests, so this should be revisited.

integration-tests/jest/jest.spec.js
```diff
@@ -2286,8 +2287,11 @@ describe('jest CommonJS', () => {
         }).catch(done)
       })
     })
-
     it('works with happy-dom', async () => {
+      // happy-dom>=19 can only be used with CJS from node 20 and above
+      if (NODE_MAJOR < 20) {
+        return
+      }
```
BridgeAR:
```suggestion
    // happy-dom>=19 can only be used with CJS from node 20 and above
    (NODE_MAJOR < 20 ? it.skip : it)('works with happy-dom', async () => {
```

juan-fernandez: I thought of this but aren't we trying to move away from `mocha` to `tap`?

BridgeAR: No, @watson looked into updating tap and it is quite a pain overall due to a complete rewrite.

Instead, we decided we will move to Node.js own test-runner from the moment on we drop support for v5 (Node.js 20 has enough features for us)

calvinbayer (approved):

PR #6568, opened 2025-09-30
update dependency blacklist lint rule from recent cleanups

simon-id:
We eradicated these dependencies from our codebase. Never again!

### DIFF
--- a/.github/workflows/appsec.yml
+++ b/.github/workflows/appsec.yml
@@ -56,7 +56,7 @@ jobs:
       PLUGINS: ldapjs
     services:
       openldap:
-        image: bitnami/openldap:latest
+        image: bitnamilegacy/openldap:latest
         ports:
           - '1389:1389'
           - '1636:1636'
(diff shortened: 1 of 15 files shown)


eslint.config.mjs
```diff
@@ -382,6 +382,18 @@ export default [
...
           name: 'semver',
           message: 'Please use semifies instead.'
+        },
+        {
+          name: 'get-port',
+          message: 'Please listen on port 0 instead.'
+        },
+        {
+          name: 'rimraf',
+          message: 'Please use fs.rmSync(path, { recursive: true }) instead.'
```
BridgeAR:
```suggestion
          message: 'Please use fs.rm(path, { recursive: true }) instead.'
```

eslint.config.mjs
```diff
@@ -382,6 +382,18 @@ export default [
...
+        },
+        {
+          name: 'koalas',
+          message: 'Please use ?? instead.'
```
BridgeAR:
```suggestion
          message: 'Please use nullish coalescing operator (??) instead.'
```

BridgeAR (approved):

eslint.config.mjs
```diff
@@ -382,6 +382,18 @@ export default [
...
           name: 'semver',
           message: 'Please use semifies instead.'
+        },
+        {
+          name: 'get-port',
+          message: 'Please listen on port 0 instead.'
+        },
+        {
+          name: 'rimraf',
+          message: 'Please use fs.rm(path, { recursive: true }) instead.'
```
watson, line 392: Technically the equivalent is

```suggestion
          message: 'Please use fs.rm(path, { recursive: true, force: true }) instead.'
```

I've created a follow-up PR: https://github.com/DataDog/dd-trace-js/pull/6592

PR #6571, opened 2025-09-30
feat(llmobs): add `annotationContext` API to annotate auto-generated spans

sabrenner:
### What does this PR do?
Adds an `annotationContext`, similar to our existing API in Python: https://docs.datadoghq.com/llm_observability/instrumentation/sdk?tab=python#annotating-auto-instrumented-spans.

### Motivation
Allow users to annotate auto-instrumented spans within a certain scope/context - brings us more in-parity with the functionality of our Python SDK.

### Additional Notes
The "context" part of the name `annotationContext` might not make as much sense in a Javascript/Node.js context versus Python, so I'm open to name changes before merging!


index.d.ts
```diff
@@ -3099,6 +3108,18 @@ declare namespace tracer {
       tags?: { [key: string]: any }
     }
 
+    interface AnnotationContextOptions {
```
sabrenner, line 3111: note for llmobs reviewers: i'm not adding prompt stuff in here for now since there's no precedent of it in the node.js sdk at all. i'm thinking we can bulk add prompt stuff down the line

packages/dd-trace/src/llmobs/plugins/base.js
```diff
@@ -40,10 +40,14 @@ class LLMObsPlugin extends TracingPlugin {
       telemetry.incrementLLMObsSpanStartCount({ autoinstrumented: true, integration: this.constructor.integration })
 
       ctx.llmobs = {} // initialize context-based namespace
-      llmobsStorage.enterWith({ span })
-      ctx.llmobs.parent = parent
-
-      this._tagger.registerLLMObsSpan(span, { parent, integration: this.constructor.integration, ...registerOptions })
+      llmobsStorage.enterWith({ ...parentStore, span })
+      ctx.llmobs.parent = parentStore
```
Yun-Kim, line 44: I assume we're setting the overall store object rather than just the span because this allows us to keep track of the annotation context?

Yun-Kim (approved):
Logic lgtm, I am happy with whatever name you choose for `annotationContext` that might make more sense for javascript users!

packages/dd-trace/src/llmobs/plugins/base.js
sabrenner: yep!

tlhunter (approved):

PR #6583, opened 2025-10-02
openai responses instrumentation

jordan-wong:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/openai.js
+++ b/packages/datadog-instrumentations/src/openai.js
@@ -23,6 +23,21 @@ const V4_PACKAGE_SHIMS = [
     streamedResponse: true
   },
   {
+    file: 'resources/responses',
+    targetClass: 'Responses',
+    baseResource: 'responses',
+    methods: ['create'],
+    streamedResponse: false
+  },
+  {
+    file: 'resources/responses/responses',
+    targetClass: 'Responses',
+    baseResource: 'responses',
+    methods: ['create'],
+    streamedResponse: false,
+    versions: ['>=4.85.0']
+  },
+  {
     file: 'resources/embeddings',
     targetClass: 'Embeddings',
     baseResource: 'embeddings',
@@ -137,6 +152,24 @@ const V4_PACKAGE_SHIMS = [
   }
 ]
 
+// define and return function to patch over the original function
+function wrapCreate (create) {
+  return function (request) {
+    if (!vertexaiTracingChannel.start.hasSubscribers) {
+      // calls the original function
+      return create.apply(this, arguments)
+    }
+
+    const ctx = {
+      request,
+      instance: this,
+      resource: [this.constructor.name, create.name].join('.')
+    }
+    // am I using the right channel? tracingChannel vs diagnostics channel
+    return ch.tracePromise(create, ctx, this, ...arguments)
+  }
+}
+
 addHook({ name: 'openai', file: 'dist/api.js', versions: ['>=3.0.0 <4'] }, exports => {
   const methodNames = Object.getOwnPropertyNames(exports.OpenAIApi.prototype)
   methodNames.shift() // remove leading 'constructor' method
@@ -159,6 +192,16 @@ addHook({ name: 'openai', file: 'dist/api.js', versions: ['>=3.0.0 <4'] }, expor
 
   return exports
 })
+
+//register patching hooks via addHook
+addHook({ name: 'openai', file: 'resources/responses.js', versions: ['>=4.87.0'] }, exports => {
+  const Responses = exports.OpenAIApi.responses
+  // wrap functions on module exports with shimmer.wrap
+  shimmer.wrap(responses.prototype, 'responses.createResponse', wrapCreate)
+  return exports
+})
+
+
 
 /**
  * For streamed responses, we need to accumulate all of the content in
--- a/packages/datadog-plugin-openai/src/tracing.js
+++ b/packages/datadog-plugin-openai/src/tracing.js
@@ -136,6 +136,10 @@ class OpenAiTracingPlugin extends TracingPlugin {
       case 'createEdit':
         createEditRequestExtraction(tags, payload, openaiStore)
         break
+
+      case 'createResponse':
+        createResponseRequestExtraction(tags, payload, openaiStore)
+        break
     }
 
     span.addTags(tags)
@@ -315,6 +319,10 @@ function normalizeMethodName (methodName) {
     case 'embeddings.create':
       return 'createEmbedding'
 
+    // responses
+    case 'responses.create':
+      return 'createResponse'
+
     // files
     case 'files.create':
       return 'createFile'
@@ -378,6 +386,35 @@ function createEditRequestExtraction (tags, payload, openaiStore) {
   openaiStore.instruction = instruction
 }
 
+function createResponseRequestExtraction (tags, payload, openaiStore) {
+  // Extract model information
+  if (payload.model) {
+    tags['openai.request.model'] = payload.model
+  }
+  
+  // Extract input information
+  if (payload.input) {
+    openaiStore.input = payload.input
+    tags['openai.request.input_length'] = payload.input.length
+  }
+  
+  // Extract reasoning configuration
+  if (payload.reasoning) {
+    if (payload.reasoning.effort) {
+      tags['openai.request.reasoning.effort'] = payload.reasoning.effort
+    }
+    openaiStore.reasoning = payload.reasoning
+  }
+  
+  // Extract background flag
+  if (payload.background !== undefined) {
+    tags['openai.request.background'] = payload.background
+  }
+  
+  // Store the full payload for response extraction
+  openaiStore.responseData = payload
+}
+
 function retrieveModelRequestExtraction (tags, payload) {
   tags['openai.request.id'] = payload.id
 }
@@ -410,6 +447,10 @@ function responseDataExtractionByMethod (methodName, tags, body, openaiStore) {
     case 'createChatCompletion':
     case 'createEdit':
       commonCreateResponseExtraction(tags, body, openaiStore, methodName)
+      break
+
+    case 'createResponse':
+      createResponseResponseExtraction(tags, body, openaiStore)
       break
 
     case 'listFiles':
@@ -513,6 +554,26 @@ function commonCreateResponseExtraction (tags, body, openaiStore, methodName) {
   if (!body.choices) return
 
   openaiStore.choices = body.choices
+}
+
+function createResponseResponseExtraction (tags, body, openaiStore) {
+  // Extract response ID if available
+  if (body.id) {
+    tags['openai.response.id'] = body.id
+  }
+  
+  // Extract status if available
+  if (body.status) {
+    tags['openai.response.status'] = body.status
+  }
+  
+  // Extract model from response if available
+  if (body.model) {
+    tags['openai.response.model'] = body.model
+  }
+  
+  // Store the full response for potential future use
+  openaiStore.response = body
 }
 
 // The server almost always responds with JSON
--- a/packages/dd-trace/src/llmobs/plugins/openai.js
+++ b/packages/dd-trace/src/llmobs/plugins/openai.js
@@ -21,7 +21,7 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
     const methodName = gateResource(normalizeOpenAIResourceName(resource))
     if (!methodName) return // we will not trace all openai methods for llmobs
 
-    const inputs = ctx.args[0] // completion, chat completion, and embeddings take one argument
+    const inputs = ctx.args[0] // completion, chat completion, embeddings, and responses take one argument
     const operation = getOperation(methodName)
     const kind = operation === 'embedding' ? 'embedding' : 'llm'
 
@@ -55,6 +55,8 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
       this._tagChatCompletion(span, inputs, response, error)
     } else if (operation === 'embedding') {
       this._tagEmbedding(span, inputs, response, error)
+    } else if (operation === 'response') {
+      this._tagResponse(span, inputs, response, error)
     }
 
     if (!error) {
@@ -187,6 +189,41 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
 
     this._tagger.tagMetadata(span, metadata)
   }
+
+  _tagResponse (span, inputs, response, error) {
+    const { input, model, reasoning, background, ...parameters } = inputs
+
+    // Create input message format
+    const responseInput = [{ content: input }]
+
+    if (error) {
+      this._tagger.tagLLMIO(span, responseInput, [{ content: '' }])
+      return
+    }
+
+    // Create output message format
+    const responseOutput = [{ content: response.output || '' }]
+
+    this._tagger.tagLLMIO(span, responseInput, responseOutput)
+    console.log('hello params', parameters)
+    
+    // Tag metadata
+    const metadata = Object.entries(parameters).reduce((obj, [key, value]) => {
+      if (!['tools', 'functions'].includes(key)) {
+        obj[key] = value
+      }
+      return obj
+    }, {})
+    
+    if (reasoning) {
+      metadata.reasoning = reasoning
+    }
+    if (background !== undefined) {
+      metadata.background = background
+    }
+
+    this._tagger.tagMetadata(span, metadata)
+  }
 }
 
 // TODO: this will be moved to the APM integration
@@ -203,13 +240,18 @@ function normalizeOpenAIResourceName (resource) {
     // embeddings
     case 'embeddings.create':
       return 'createEmbedding'
+
+    // responses
+    case 'responses.create':
+      return 'createResponse'
+
     default:
       return resource
   }
 }
 
 function gateResource (resource) {
-  return ['createCompletion', 'createChatCompletion', 'createEmbedding'].includes(resource)
+  return ['createCompletion', 'createChatCompletion', 'createEmbedding', 'createResponse'].includes(resource)
     ? resource
     : undefined
 }
@@ -222,6 +264,8 @@ function getOperation (resource) {
       return 'chat'
     case 'createEmbedding':
       return 'embedding'
+    case 'createResponse':
+      return 'response'
     default:
       // should never happen
       return 'unknown'


packages/datadog-instrumentations/src/openai.js
```diff
@@ -22,6 +22,21 @@ const V4_PACKAGE_SHIMS = [
     methods: ['create'],
     streamedResponse: true
   },
+  {
+    file: 'resources/responses',
+    targetClass: 'Responses',
+    baseResource: 'responses',
+    methods: ['create'],
+    streamedResponse: false
+  },
```
sabrenner:
```suggestion
```

packages/datadog-instrumentations/src/openai.js
```diff
@@ -22,6 +22,21 @@ const V4_PACKAGE_SHIMS = [
...
+  {
+    file: 'resources/responses/responses',
+    targetClass: 'Responses',
+    baseResource: 'responses',
+    methods: ['create'],
+    streamedResponse: false,
+    versions: ['>=4.85.0']
```
sabrenner:
```suggestion
    versions: ['>=4.87.0']
```

packages/datadog-instrumentations/src/openai.js
```diff
@@ -22,6 +22,21 @@ const V4_PACKAGE_SHIMS = [
...
+    baseResource: 'responses',
+    methods: ['create'],
+    streamedResponse: false
+  },
+  {
+    file: 'resources/responses/responses',
+    targetClass: 'Responses',
+    baseResource: 'responses',
+    methods: ['create'],
+    streamedResponse: false,
```
sabrenner:
```suggestion
    streamedResponse: true,
```

packages/datadog-instrumentations/src/openai.js
```diff
@@ -137,6 +152,24 @@ const V4_PACKAGE_SHIMS = [
...
+    const ctx = {
+      request,
+      instance: this,
+      resource: [this.constructor.name, create.name].join('.')
+    }
+    // am I using the right channel? tracingChannel vs diagnostics channel
+    return ch.tracePromise(create, ctx, this, ...arguments)
+  }
+}
+
```
sabrenner:
```suggestion
```

packages/datadog-instrumentations/src/openai.js
```diff
@@ -160,6 +193,16 @@ addHook({ name: 'openai', file: 'dist/api.js', versions: ['>=3.0.0 <4'] }, expor
...
+//register patching hooks via addHook
+addHook({ name: 'openai', file: 'resources/responses.js', versions: ['>=4.87.0'] }, exports => {
+  const Responses = exports.OpenAIApi.responses
+  // wrap functions on module exports with shimmer.wrap
+  shimmer.wrap(responses.prototype, 'responses.createResponse', wrapCreate)
+  return exports
+})
+
+
+
```
sabrenner:
```suggestion
```

packages/datadog-plugin-openai/src/tracing.js
sabrenner, line 65: we'll need a branch here to handle the streamed responses result, but we should be able to grab the last chunk like this

```javascript
const lastChunk = chunks[chunks.length - 1].response
```

and then grab usage and output off of that, which should all be combined

packages/datadog-plugin-openai/src/stream-helpers.js
```diff
@@ -107,8 +107,103 @@ function constructChatCompletionResponseFromStreamedChunks (chunks, n) {
...
+        return {
+          ...finalResponse,
+          output: outputText
+        }
+      }
+    }
+    return finalResponse
+  }
+  
+  // If no final response found, fall back to accumulating from deltas and items
```
sabrenner: is there a specific model/response format that doesn't include the aggregated streamed response in the last chunk? i think otherwise we can safely only look at the last chunk and not worry about doing manual aggregating

packages/datadog-plugin-openai/src/tracing.js
```diff
@@ -376,6 +389,35 @@ function createEditRequestExtraction (tags, payload, openaiStore) {
...
+    if (payload.reasoning.effort) {
+      tags['openai.request.reasoning.effort'] = payload.reasoning.effort
+    }
+    openaiStore.reasoning = payload.reasoning
+  }
+  
+  // Extract background flag
+  if (payload.background !== undefined) {
+    tags['openai.request.background'] = payload.background
+  }
```
sabrenner: i think we can remove these blocks and just do model tagging for apm spans

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -75,20 +79,20 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
     const tokenUsage = response.usage
 
     if (tokenUsage) {
-      const inputTokens = tokenUsage.prompt_tokens
-      if (inputTokens) metrics.inputTokens = inputTokens
+      // Responses API uses input_tokens, Chat/Completions use prompt_tokens
```
sabrenner, line 87: with rebasing these changes i think there'll be some conflicts with cached tokens as well, lmk if you need any help/clarifications there

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -191,6 +195,188 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
 
     this._tagger.tagMetadata(span, metadata)
   }
+
+  _tagResponse (span, inputs, response, error) {
```
```suggestion
  #tagResponse (span, inputs, response, error) {
```

just as a general JS practice this is better for private functions (i know the other ones aren't like this, i need to get around to cleaning up the rest of this code 😅)

and then calling the function above would just be

```javascript
this.#tagResponse(...)
```

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -191,6 +195,188 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+              args = {}
+            }
+          }
+          outputMessages.push({
+            role: '',  // Tool calls have empty role in LLMObs
+            toolCalls: [{
+              toolId: item.call_id,
+              name: item.name,
+              arguments: args,
+              type: 'function'
```
```suggestion
              type: item.type
```

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -191,6 +195,188 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+              parsedArgs = {}
+            }
+          }
+          inputMessages.push({
+            role: '',
+            toolCalls: [{
+              toolId: item.call_id,
+              name: item.name,
+              arguments: parsedArgs,
+              type: 'function'
```
```suggestion
              type: item.type
```

packages/datadog-plugin-openai/src/stream-helpers.js
```diff
@@ -107,8 +107,103 @@ function constructChatCompletionResponseFromStreamedChunks (chunks, n) {
...
+      
+      if (outputText) {
+        return {
+          ...finalResponse,
+          output: outputText
+        }
+      }
+    }
+    return finalResponse
+  }
```
sabrenner:
```suggestion
  return finalResponse
```

i think this might be good here - were there any scenarios you found where the last chunk did not have aggregations done properly?

packages/datadog-plugin-openai/src/stream-helpers.js
```diff
@@ -107,8 +107,103 @@ function constructChatCompletionResponseFromStreamedChunks (chunks, n) {
...
+  if (chunks.length === 0) return {}
+  
+  // The responses API streams events with different types:
+  // - response.output_text.delta: incremental text deltas
+  // - response.output_text.done: complete text for a content part
+  // - response.output_item.done: complete output item with role
+  // - response.done/response.incomplete/response.completed: final response with output array and usage
+  
+  // Find the last chunk with a complete response object (status: done, incomplete, or completed)
+  let finalResponse = null
```
sabrenner:
```suggestion
  let finalResponse
```

packages/dd-trace/src/llmobs/span_processor.js
```diff
@@ -224,7 +166,8 @@ class LLMObsSpanProcessor {
           continue
         }
         if (value !== null && typeof value === 'object') {
-          add(value, carrier[key] = {})
+          carrier[key] = Array.isArray(value) ? [] : {}
+          add(value, carrier[key])
```
sabrenner, line 228: was there a specific scenario that was hitting this? fine with this change though!

packages/dd-trace/src/llmobs/tagger.js
```diff
@@ -308,6 +297,23 @@ class LLMObsTagger {
...
+      // Otherwise use conditional tagging which skips empty values
+      let condition
+      if (messageObj.tool_calls && messageObj.tool_calls.length > 0) {
+        // For tool call messages, always include role even if empty
+        messageObj.role = role || ''
+        condition = true
+      } else {
+        condition = this.#tagConditionalString(role, 'Message role', messageObj, 'role')
+      }
+
```
sabrenner: i've realized the shared tests are now a bit out of date with what we do in python - i will fix those and get back to you, i _thinkkkkk_ this stuff here could be removed once i fix up the tests 🙏 i will let you know, but i did confirm that these changes work with the current tests, so thank you!

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -191,6 +195,188 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+      metadata.reasoning_tokens = response.usage.output_tokens_details.reasoning_tokens
+    }
+    
+    // Add reasoning metadata from input parameters
+    if (reasoning) {
+      metadata.reasoning = reasoning
+    }
+    if (background !== undefined) {
+      metadata.background = background
+    }
```
sabrenner: i think we can keep metadata tagging to the request `parameters` and not from the response, and then maybe do a allowlist instead of denylist for what we want to include, ie

```javascript
const metadata = Object.entries(parameters).reduce((obj, [key, value]) => {
      if (['temperature', 'top_p', 'tuncation', ...].includes(key)) {
        obj[key] = value
      }
      return obj
    }, {})
```

lmk if that doesn't make sense!

packages/dd-trace/src/llmobs/span_processor.js
jordan-wong: There was a test case I was failing where an empty array was being passed in the metadata, but it got initialized as an empty Object instead by ths span_processor part. I added a case to  initialize an empty array as an empty array

packages/datadog-plugin-openai/src/stream-helpers.js
```diff
@@ -107,8 +107,38 @@ function constructChatCompletionResponseFromStreamedChunks (chunks, n) {
...
+  let finalResponse
+  for (let i = chunks.length - 1; i >= 0; i--) {
+    const chunk = chunks[i]
+    if (chunk.response && ['done', 'incomplete', 'completed'].includes(chunk.response.status)) {
+      finalResponse = chunk.response
+      return finalResponse
+    }
+  }
+  
+  return finalResponse
```
sabrenner:
```suggestion
  // The responses API streams events with different types:
  // - response.output_text.delta: incremental text deltas
  // - response.output_text.done: complete text for a content part
  // - response.output_item.done: complete output item with role
  // - response.done/response.incomplete/response.completed: final response with output array and usage
  // Find the last chunk with a complete response object (status: done, incomplete, or completed)
  
  for (let i = chunks.length - 1; i >= 0; i--) {
    const chunk = chunks[i]
    if (chunk.response && ['done', 'incomplete', 'completed'].includes(chunk.response.status)) {
      return chunk.response
    }
  }
```

now that we simplified the logic a bit i think we can just return the proper chunk directly, and if there isn't one the function will return `undefined` by default

we could also move the `['done', 'incomplete', 'completed']` array to be a const at the top of the file

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -75,19 +77,30 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+      // For Responses API, always include cache tokens (even if 0)
+      // For Chat API, only include if > 0
+      if (tokenUsage.input_tokens_details) {
+        // Responses API - always include
+        const cacheReadTokens = tokenUsage.input_tokens_details.cached_tokens
+        if (cacheReadTokens !== undefined) metrics.cacheReadTokens = cacheReadTokens
+      } else if (tokenUsage.prompt_tokens_details) {
+        // Chat/Completions API - only include if > 0
+        const cacheReadTokens = tokenUsage.prompt_tokens_details.cached_tokens
+        if (cacheReadTokens !== undefined && cacheReadTokens > 0) {
```
```suggestion
        if (cacheReadTokens) {
```

i think this should be the same, since if `cacheReadTokens` is `0` then it'll still evaluate as false-y.

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -191,6 +204,189 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+    if (Array.isArray(input)) {
+      for (const item of input) {
+        if (item.type === 'function_call') {
+          // Function call: convert to message with tool_calls
+          // Parse arguments if it's a JSON string
+          let parsedArgs = item.arguments
+          if (typeof parsedArgs === 'string') {
+            try {
+              parsedArgs = JSON.parse(parsedArgs)
+            } catch (e) {
```
```suggestion
            } catch {
```

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -191,6 +204,189 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+      outputMessages.push({ role: 'assistant', content: response.output })
+    } else if (Array.isArray(response.output)) {
+      // Array output - process all items to extract reasoning, messages, and tool calls
+      // Non-streaming: array of items (messages, function_calls, or reasoning)
+      for (const item of response.output) {
+        // Handle reasoning type (reasoning responses)
+        if (item.type === 'reasoning') {
+          // Extract reasoning text from summary
+          let reasoningText = ''
+          if (item.summary && Array.isArray(item.summary) && item.summary.length > 0) {
```
```suggestion
          if (Array.isArray(item.summary) && item.summary.length > 0) {
```

`Array.isArray` will also check if non-null/non-undefined!

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -191,6 +204,189 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+            content: reasoningText
+          })
+        } else if (item.type === 'function_call') {
+          // Handle function_call type (responses API tool calls)
+          let args = item.arguments
+          // Parse arguments if it's a JSON string
+          if (typeof args === 'string') {
+            try {
+              args = JSON.parse(args)
+            } catch (e) {
```
```suggestion
            } catch {
```

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -191,6 +204,189 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+        } else {
+          // Handle regular message objects
+          const outputMsg = { role: item.role || 'assistant', content: '' }
+          
+          // Extract content from message
+          if (item.content && Array.isArray(item.content)) {
+            // Content is array of content parts
+            // For responses API, text content has type 'output_text', not 'text'
+            const textParts = item.content
+              .filter(c => c.type === 'text' || c.type === 'output_text')
```
```suggestion
              .filter(c => c.type === 'output_text')
```

based on the comment above, seems we can just filter for `output_text` i think, lmk if i'm misunderstanding tho!

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -191,6 +204,189 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+              .map(c => c.text)
+            outputMsg.content = textParts.join('')
+          } else if (typeof item.content === 'string') {
+            outputMsg.content = item.content
+          }
+          
+          // Extract tool calls if present in message.tool_calls
+          if (item.tool_calls && Array.isArray(item.tool_calls)) {
```
```suggestion
          if (Array.isArray(item.tool_calls)) {
```

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -191,6 +204,189 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+      outputMessages.push({ role: 'assistant', content: response.output_text })
+    } else {
+      // No output
+      outputMessages.push({ role: 'assistant', content: '' })
+    }
+
+    this._tagger.tagLLMIO(span, inputMessages, outputMessages)
+    
+    // Tag metadata - use allowlist approach for request parameters
+    const allowedParamKeys = [
```
sabrenner: we can make this a const at the top of the file

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -191,6 +204,189 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+    
+    // Add fields from response object (convert numbers to floats)
+    if (response.temperature !== undefined) metadata.temperature = parseFloat(response.temperature)
+    if (response.top_p !== undefined) metadata.top_p = parseFloat(response.top_p)
+    if (response.tool_choice !== undefined) metadata.tool_choice = response.tool_choice
+    if (response.truncation !== undefined) metadata.truncation = response.truncation
+    if (response.text !== undefined) metadata.text = response.text
+    if (response.usage?.output_tokens_details?.reasoning_tokens !== undefined) {
+      metadata.reasoning_tokens = response.usage.output_tokens_details.reasoning_tokens
+    }
```
sabrenner: i'm probably misremembering from the shared tests, but i don't think we need to grab anything from the response for metadata tagging, only things from the `parameters` (so maybe we can add `top_p`, `truncation`, etc. to the list defined above)

packages/dd-trace/src/llmobs/tagger.js
```diff
@@ -343,6 +343,23 @@ class LLMObsTagger {
         }
       }
 
+      // Include content if not empty, no tool calls/results, or explicitly provided
+      if (content !== '' || (!messageObj.tool_calls && !messageObj.tool_results) || ('content' in message)) {
```
sabrenner: i think we should be able to get rid of this changed logic here (although I know it'll fail the shared tests 😭). we actually get around this for the newly added anthropic integration by setting content to an empty string: https://github.com/DataDog/dd-trace-js/blob/c2d2d8416312388b5ad6205c14021009006db51e/packages/dd-trace/src/llmobs/plugins/anthropic.js#L178

but i know the shared tests expect no content at all. let's revert this section, and let me see if i can modify the shared tests to account for either empty content/content not included, and we can undo this change, since this touches a path that users could run into as well.

packages/dd-trace/src/llmobs/tagger.js
```diff
@@ -280,11 +280,13 @@ class LLMObsTagger {
...
       }
 
-      const { result, toolId, type } = toolResult
+      const { result, toolId, name = '', type } = toolResult
       const toolResultObj = {}
 
       const condition1 = this.#tagConditionalString(result, 'Tool result', toolResultObj, 'result')
       const condition2 = this.#tagConditionalString(toolId, 'Tool ID', toolResultObj, 'tool_id')
+      // name can be empty string, so always include it
+      toolResultObj.name = name
```
sabrenner:
```suggestion
      if (typeof name === 'string') {
        toolResultObj.name = name
      } else {
        this.#handleFailure(`[LLMObs] Expected tool result name to be a string, instead got "${typeof name}"`)
      }
```

since we can't use tagConfitionalString here (which i should change in the future to only check on null/undefined an not falsy), we should add a small guard that the name is a string (which is what our backend expects)

packages/dd-trace/src/llmobs/plugins/openai.js
sabrenner: disregarding this comment!

packages/datadog-plugin-openai/src/stream-helpers.js
```diff
@@ -107,8 +107,32 @@ function constructChatCompletionResponseFromStreamedChunks (chunks, n) {
...
 }
 
+/**
+ * Constructs the entire response from a stream of OpenAI responses chunks.
+ * The responses API uses event-based streaming with delta chunks.
+ * @param {Array<Record<string, any>>} chunks
+ * @param {number} n (not used for responses API, but kept for consistency)
+ * @returns {Record<string, any>}
+ */
+function constructResponseResponseFromStreamedChunks (chunks, n) {
```
sabrenner: nit: i missed this before but we can get rid of `n` here

packages/datadog-plugin-openai/src/tracing.js
```diff
@@ -59,6 +60,8 @@ class OpenAiTracingPlugin extends TracingPlugin {
         response = constructCompletionResponseFromStreamedChunks(chunks, n)
       } else if (methodName === 'createChatCompletion') {
         response = constructChatCompletionResponseFromStreamedChunks(chunks, n)
+      } else if (methodName === 'createResponse') {
+        response = constructResponseResponseFromStreamedChunks(chunks, n)
```
sabrenner:
```suggestion
        response = constructResponseResponseFromStreamedChunks(chunks)
```

also nit

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -191,6 +211,183 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+              arguments: args,
+              type: item.type
+            }]
+          })
+        } else {
+          // Handle regular message objects
+          const outputMsg = { role: item.role || 'assistant', content: '' }
+
+          // Extract content from message
+          if (item.content && Array.isArray(item.content)) {
```
```suggestion
          if (Array.isArray(item.content)) {
```

nit

packages/dd-trace/src/llmobs/tagger.js
```diff
@@ -343,6 +347,18 @@ class LLMObsTagger {
...
+      let condition
+      if ((messageObj.tool_calls && messageObj.tool_calls.length > 0) ||
+          (messageObj.tool_results && messageObj.tool_results.length > 0)) {
+        // For tool call/result messages, always include role
+        messageObj.role = role || ''
+        condition = true
+      } else {
+        condition = this.#tagConditionalString(role, 'Message role', messageObj, 'role')
+      }
+
```
sabrenner: i forget if i mentioned it before, but with the changes i made to the shared tests i think we should be able to revert the changes from line 328 through 361, but lmk if that's not the case!

sabrenner (commented):
super awesome!! just a couple nits, and i'd say a couple tests locally here but then otherwise it's looking good to go 😄

packages/datadog-plugin-openai/src/stream-helpers.js
```diff
@@ -107,8 +107,31 @@ function constructChatCompletionResponseFromStreamedChunks (chunks, n) {
...
+  // The responses API streams events with different types:
+  // - response.output_text.delta: incremental text deltas
+  // - response.output_text.done: complete text for a content part
+  // - response.output_item.done: complete output item with role
+  // - response.done/response.incomplete/response.completed: final response with output array and usage
+
+  // Find the last chunk with a complete response object (status: done, incomplete, or completed)
+  for (let i = chunks.length - 1; i >= 0; i--) {
+    const chunk = chunks[i]
+    if (chunk.response && ['done', 'incomplete', 'completed'].includes(chunk.response.status)) {
```
BridgeAR: What about:

```suggestion
    if (chunk.response && reponseStatusSet.has(chunk.response.status)) {
```

And outside of the method:

`const responseStatusSet = new Set(['done', 'incomplete', 'completed'])`

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -191,6 +211,183 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+    const metadata = Object.entries(parameters).reduce((obj, [key, value]) => {
+      if (allowedParamKeys.has(key)) {
+        obj[key] = value
+      }
+      return obj
+    }, {})
+
+    // Add fields from response object (convert numbers to floats)
+    if (response.temperature !== undefined) metadata.temperature = Number.parseFloat(response.temperature)
+    if (response.top_p !== undefined) metadata.top_p = Number.parseFloat(response.top_p)
```
BridgeAR:
```suggestion
    if (response.temperature !== undefined) metadata.temperature = Number(response.temperature)
    if (response.top_p !== undefined) metadata.top_p = Number(response.top_p)
```

That would also parse the entry as number, it would just not accept values such as: `1.23a`. That would result in `NaN`, which seems more natural to me.

packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
```diff
@@ -639,6 +639,106 @@ describe('integrations', () => {
...
+            tool_choice: 'auto',
+            truncation: 'disabled',
+            text: { format: { type: 'text' }, verbosity: 'medium' },
+            reasoning_tokens: 0,
+            stream: false
+          },
+          tags: { ml_app: 'test', integration: 'openai' }
+        })
+
+        expect(llmobsSpans[0]).to.deepEqualWithMockValues(expected)
```
sabrenner:
```suggestion
```

packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
```diff
@@ -639,6 +639,106 @@ describe('integrations', () => {
...
+        const stream = await openai.responses.create({
+          model: 'gpt-4o-mini',
+          input: 'Stream this please',
+          max_output_tokens: 50,
+          temperature: 0,
+          stream: true
+        })
+
+        for await (const part of stream) {
+          expect(part).to.have.property('type')
```
sabrenner:
```suggestion
          assert.ok(Object.hasOwn(part, 'type'))
```

packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
```diff
@@ -639,6 +639,106 @@ describe('integrations', () => {
...
+            tool_choice: 'auto',
+            truncation: 'disabled',
+            text: { format: { type: 'text' }, verbosity: 'medium' },
+            reasoning_tokens: 0,
+            stream: true
+          },
+          tags: { ml_app: 'test', integration: 'openai' }
+        })
+
+        expect(llmobsSpans[0]).to.deepEqualWithMockValues(expected)
```
sabrenner:
```suggestion
```

PR #6584, opened 2025-10-02
[test optimization] Make EFD and impacted tests compatible with jest's snapshot testing

juan-fernandez:
### What does this PR do?

* Add compatibility with snapshot testing for `jest` whenever we're retrying tests (early flake detection or attempt to fix)
* Remove added strings to reported test names in `mocha` and `jest` when retrying:
  * in the case of `jest`, they're mostly useless, as in CI no test name is reported (only if it fails), only test files.
  * the issues they were causing in the case of retries were not worth the added benefit.  
  * we'll come up with a better way to report retried tests: most likely as an added report at the end of the execution

### Motivation

* Improve compatibility of retries

### Plugin Checklist

- [x] Integration tests.

### DIFF
--- a/integration-tests/ci-visibility/test-management/test-snapshot-attempt-to-fix-1.js
+++ b/integration-tests/ci-visibility/test-management/test-snapshot-attempt-to-fix-1.js
@@ -4,12 +4,10 @@ let retryCounter = 0
 
 describe('attempt to fix snapshot', () => {
   it('is flaky', () => {
-    retryCounter++
-    const sum = retryCounter > 2 ? 3 : 4
-
     if (process.env.SHOULD_PASS_ALWAYS) {
       expect(3).toMatchSnapshot()
     } else {
+      const sum = ++retryCounter > 2 ? 3 : 4
       expect(sum).toMatchSnapshot()
     }
     expect('a').toMatchSnapshot()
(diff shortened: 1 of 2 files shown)


integration-tests/jest/jest.spec.js
```diff
@@ -2111,11 +2109,17 @@ describe('jest CommonJS', () => {
       })
     })
 
-    it('does not run early flake detection on snapshot tests', (done) => {
+    it('works with snapshot tests', async () => {
```
juan-fernandez, line 2113: let's also add a snapshot test that fails

integration-tests/ci-visibility/test-early-flake-detection/__image_snapshots__/jest-image-snapshot-js-snapshot-can-match-1-snap.png
juan-fernandez, line 1: ℹ️ this is an empty PNG to test [jest-image-snapshot](https://github.com/americanexpress/jest-image-snapshot)

integration-tests/ci-visibility/test-management/__image_snapshots__/test-snapshot-image-js-snapshot-can-match-1-snap.png
juan-fernandez, line 1: ℹ️ this is an empty PNG to test [jest-image-snapshot](https://github.com/americanexpress/jest-image-snapshot)

anmarchenko (approved):

Mariovido (approved):

PR #6597, opened 2025-10-03
Have profilers emit settings and metrics into `event.json`

szegedi:
### What does this PR do?
Profiler uploads contain a file `event.json` that has a section for system info. The contents of system info are often language specific. So far we had the agent exporter set some system-specific values in it, but we want to generalize it a bit so that individual profiler facets (wall, heap, events) can contribute values there as well. The values are typically either metrics or settings.

In this iteration, all three profiler facets contribute their effective settings, and the wall profiler also contributes metrics for tracking the usage of async sample contexts when using `AsyncContextFrame` on Node.js 24+.

Additionally, the wall profiler now emits telemetry about the async sample contexts as well. This telemetry is defined in https://github.com/DataDog/dd-go/pull/190144

### Motivation
Having more information in uploaded profiles about the system configuration + having insight into the use of async sample contexts.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/dd-trace/src/profiling/profiler.js
```diff
@@ -305,7 +325,7 @@ class Profiler extends EventEmitter {
 
     tags.snapshot = snapshotKind
     tags.profile_seq = this.#profileSeq++
-    const exportSpec = { profiles, start, end, tags, endpointCounts }
+    const exportSpec = { profiles, infos, start, end, tags, endpointCounts }
```
nsavoire, line 330: `infos` does not seem used by `export`

packages/dd-trace/src/profiling/profiler.js
```diff
@@ -41,6 +41,16 @@ function findWebSpan (startedSpans, spanId) {
   return false
 }
 
+function processInfo (infos, info, type) {
+  // transpose e.g. info.wall.settings to info.settings.wall. That way
+  // we have functional groupings first, then profiler type.
+  infos.settings[type] = info.settings
+  delete info.settings
```
nsavoire: it seems dangerous to modify the passed object

packages/dd-trace/src/profiling/profiler.js
```diff
@@ -41,6 +41,16 @@ function findWebSpan (startedSpans, spanId) {
...
+  // take over rest of info without transposition
+  if (Object.keys(info).length > 0) {
+    infos[type] = info
```
nsavoire, line 46: No functional grouping here ?

packages/dd-trace/src/profiling/profiler.js
szegedi: `AgentExporter.export` passes it to `this.getEventJSON(exportSpec)`, defined in `EventSerializer` it extends, and that uses it.

packages/dd-trace/src/profiling/profiler.js
szegedi: I thought not, the rest are typically metrics, so they'll end up as 
`wall.totalAsyncContextCount`. Of course, we could embed this into another level, e.g. `metrics` so then it'd become `wall.metrics.totalAsyncContextCount` and we could then transpose that to `metrics.wall.totalAsyncContextCount`. FWIW, we don't even need to do the whole transposition, it just adds complexity; I originally wanted a top-level `settings` key because that's how e.g. Ruby does it, so I wanted to keep it a bit consistent with it:
Screenshot 2025-10-09 at 09 40 23

packages/dd-trace/src/profiling/profiler.js
nsavoire: oh sorry, I missed `event_serializer.js` changes

packages/dd-trace/src/profiling/profiler.js
szegedi: Fair. It happens infrequently enough that it's not necessary to optimize for this.

packages/dd-trace/src/profiling/profiler.js
nsavoire: I agree better be consistent with other langages

szegedi: I actually now reworked it so that settings are constructed by `config.js` and embedded in system info by top-level profiler in`profiler.js` and very little is left in individual profiler facets, just computed settings and metrics, and for those I don't bother with transposition anymore. I think it's much better this way, but I needed you asking to nudge me into rethinking it :-)

nsavoire (dismissed):

PR #6607, opened 2025-10-06
Resolving flakiness in ci

crysmags:
### What does this PR do?
Work in progress


integration-tests/helpers/index.js
```diff
@@ -271,15 +271,18 @@ async function createSandbox (dependencies = [], isGitRepo = false,
...
   const out = path.join(folder, `dd-trace-${version}.tgz`)
-  const allDependencies = [`file:${out}`].concat(cappedDependencies)
 
   await fs.mkdir(folder)
+  await execHelper('command -v bun >/dev/null 2>&1 || npm i -g bun@1.2.23', { env: restOfEnv })
   const preferOfflineFlag = process.env.OFFLINE === '1' || process.env.OFFLINE === 'true' ? ' --prefer-offline' : ''
-  const addCommand = `yarn add ${allDependencies.join(' ')} --ignore-engines${preferOfflineFlag}`
   const addOptions = { cwd: folder, env: restOfEnv }
-  await execHelper(`npm pack --silent --pack-destination ${folder}`, { env: restOfEnv }) // TODO: cache this
+  await execHelper(`bun pm pack --silent --destination ${folder}`, { env: restOfEnv }) // TODO: cache this
```
bengl: If we're going to switch the "packer" for integration tests, it should be whatever we use to publish.

This means either:

1. Use Bun to publish to npm, or
2. Use yarn, for this test helper, since that's what we use for package publishing (afaik? correct me if I'm wrong on that one).

The objective of integration tests is _not_ speed, but accuracy. We should be as close to customer reality as we can here, and using a different tool for packaging is a step further away from that.

integration-tests/helpers/index.js
```diff
@@ -218,18 +216,20 @@ async function execHelper (command, options) {
   /* eslint-disable no-console */
   try {
     console.log('Exec START: ', command)
-    await exec(command, options)
+    await execSync(command, options)
```
bengl: How does `execSync` reduce flakiness?

Also `execSync` is synchronous, by definition, so there's no need to `await` it.

integration-tests/helpers/index.js
bengl, line 279: If we're switching packers in here, we should either switch to whatever we're using at publish time (yarn I think?) or switch our publish job to use Bun.

Overall, I think this is not the best solution specifically for ESM tests, since they really have no business being integration tests in the first place. Let's undo that, first.

bengl (changes requested):

integration-tests/helpers/index.js
```diff
@@ -214,22 +212,24 @@ function spawnProc (filename, options = {}, stdioHandler, stderrHandler) {
...
     console.error('Exec ERROR: ', command, error)
     if (command.startsWith('yarn')) {
       try {
         console.log('Exec RETRY START: ', command)
-        await exec(command, options)
+        execSync(command, options)
         console.log('Exec RETRY SUCESS: ', command)
       } catch (retryError) {
         console.error('Exec RETRY ERROR', command, retryError)
+        throw retryError
```
BridgeAR, line 230: It was intentional how it was written before :)
I would definitely prefer the original error, not the retry error.

crysmags: is this blocking if we log the error on line 222?

crysmags: I merged, but will update this in the next update

PR #6614, opened 2025-10-07
[test optimization] Better names for modified tests

juan-fernandez:
### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


integration-tests/cucumber/cucumber.spec.js
```diff
@@ -2731,45 +2730,46 @@ versions.forEach(version => {
           }
         )
 
-        childProcess.on('exit', (code) => {
-          testAssertionsPromise.then(done).catch(done)
-        })
+        await Promise.all([
```
vitor-de-araujo, line 2733: It doesn't just change names, it also changes some async logic, is this intended to go in this PR?

packages/datadog-instrumentations/src/jest.js
```diff
@@ -197,10 +197,9 @@ function getWrappedEnvironment (BaseEnvironment, jestVersion) {
 
       if (this.isImpactedTestsEnabled) {
         try {
-          const hasImpactedTests = Object.keys(modifiedTests).length > 0
-          this.modifiedTestsForThisSuite = hasImpactedTests
-            ? this.getModifiedTestForThisSuite(modifiedTests)
-            : this.getModifiedTestForThisSuite(this.testEnvironmentOptions._ddModifiedTests)
+          const hasImpactedTests = Object.keys(modifiedFiles).length > 0
+          // TODO: check if `this.testEnvironmentOptions._ddModifiedFiles` is serialized to JSON
```
calvinbayer: Is this TODO supposed to be merged?

integration-tests/cucumber/cucumber.spec.js
juan-fernandez: yeah, this is part of an ongoing effort to migrate tests to `async/await` rather than `done`. I do it whenever I have to modify tests. It should just be more robust 😄 . I should've mentioned it in the PR

packages/datadog-instrumentations/src/jest.js
juan-fernandez: good catch. I don't think so, no

PR #6615, opened 2025-10-07
chore: improve config by making properties private

BridgeAR:
This improves the type situation a lot in the file and makes it clearer what things belong to. It also simplifies the code a bit.

### DIFF
--- a/packages/dd-trace/src/config.js
+++ b/packages/dd-trace/src/config.js
@@ -233,22 +233,36 @@ function reformatSpanSamplingRules (rules) {
   })
 }
 
-const sourcesOrder = [
-  { containerProperty: '_remote', origin: 'remote_config', unprocessedProperty: '_remoteUnprocessed' },
-  { containerProperty: '_options', origin: 'code', unprocessedProperty: '_optsUnprocessed' },
-  { containerProperty: '_fleetStableConfig', origin: 'fleet_stable_config' },
-  { containerProperty: '_env', origin: 'env_var', unprocessedProperty: '_envUnprocessed' },
-  { containerProperty: '_localStableConfig', origin: 'local_stable_config' },
-  { containerProperty: '_calculated', origin: 'calculated' },
-  { containerProperty: '_defaults', origin: 'default' }
-]
-
 class Config {
   /**
    * parsed DD_TAGS, usable as a standalone tag set across products
    * @type {Record<string, string> | undefined}
    */
   #parsedDdTags = {}
+
+  #envUnprocessed = {}
+  #optsUnprocessed = {}
+  #remoteUnprocessed = {}
+  #env = {}
+  #options = {}
+  #remote = {}
+  #defaults = {}
+  #optionsArg = {}
+  #localStableConfig = {}
+  #fleetStableConfig = {}
+  #calculated = {}
+
+  #getSourcesOrder () {
+    return [
+      { container: this.#remote, origin: 'remote_config', unprocessed: this.#remoteUnprocessed },
+      { container: this.#options, origin: 'code', unprocessed: this.#optsUnprocessed },
+      { container: this.#fleetStableConfig, origin: 'fleet_stable_config' },
+      { container: this.#env, origin: 'env_var', unprocessed: this.#envUnprocessed },
+      { container: this.#localStableConfig, origin: 'local_stable_config' },
+      { container: this.#calculated, origin: 'calculated' },
+      { container: this.#defaults, origin: 'default' }
+    ]
+  }
 
   constructor (options = {}) {
     if (!isInServerlessEnvironment()) {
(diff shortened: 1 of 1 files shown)


packages/dd-trace/src/config.js
```diff
@@ -439,18 +453,13 @@ class Config {
...
 
-  // for _merge to work, every config value must have a default value
-  _applyDefaults () {
-    setHiddenProperty(this, '_defaults', defaults)
-  }
-
   _applyLocalStableConfig () {
-    const obj = setHiddenProperty(this, '_localStableConfig', {})
+    const obj = this.#localStableConfig
     this._applyStableConfig(this.stableConfig?.localEntries ?? {}, obj)
```
szegedi:
```suggestion
    this._applyStableConfig(this.stableConfig?.localEntries ?? {}, this.#localStableConfig)
```

packages/dd-trace/src/config.js
```diff
@@ -439,18 +453,13 @@ class Config {
...
   }
 
   _applyFleetStableConfig () {
-    const obj = setHiddenProperty(this, '_fleetStableConfig', {})
+    const obj = this.#fleetStableConfig
     this._applyStableConfig(this.stableConfig?.fleetEntries ?? {}, obj)
```
szegedi:
```suggestion
    this._applyStableConfig(this.stableConfig?.fleetEntries ?? {}, this.#fleetStableConfig)
```

packages/dd-trace/src/config.js
```diff
@@ -233,23 +233,37 @@ function reformatSpanSamplingRules (rules) {
...
+  #env = {}
+  #options = {}
+  #remote = {}
+  #defaults = {}
+  #optionsArg = {}
+  #localStableConfig = {}
+  #fleetStableConfig = {}
+  #calculated = {}
+
+  #getSourcesOrder () {
```
szegedi: nit: a better name for this now would be `getSourcesInOrder` as it no longer just defines ordering, it actually returns the source values in its structure.

packages/dd-trace/src/config.js
BridgeAR: I inlined the whole helper

packages/dd-trace/src/config.js
BridgeAR: I inlined the whole helper

szegedi (dismissed):

packages/dd-trace/src/config.js
```diff
@@ -97,39 +97,49 @@ function validateOtelPropagators (propagators) {
...
+      return true
   }
 }
 
 function checkIfBothOtelAndDdEnvVarSet () {
   for (const [otelEnvVar, ddEnvVar] of Object.entries(otelDdEnvMapping)) {
-    if (ddEnvVar && getEnvironmentVariable(ddEnvVar) && getEnvironmentVariable(otelEnvVar)) {
+    const otelValue = getEnv(otelEnvVar)
+
+    if (ddEnvVar && getEnv(ddEnvVar) && otelValue) {
```
watson, line 136: Nit, to not call `getEnv` unless there's a truthy `otelValue` (feel free to ignore if you don't think it matters):

```suggestion
    if (otelValue && ddEnvVar && getEnv(ddEnvVar)) {
```

packages/dd-trace/src/config.js
```diff
@@ -16,7 +16,7 @@ const telemetryMetrics = require('./telemetry/metrics')
 const { isInServerlessEnvironment, getIsGCPFunction, getIsAzureFunction } = require('./serverless')
 const { ORIGIN_KEY } = require('./constants')
 const { appendRules } = require('./payload-tagging/config')
-const { getEnvironmentVariable, getEnvironmentVariables } = require('./config-helper')
+const { getEnvironmentVariable: getEnv, getEnvironmentVariables } = require('./config-helper')
```
watson, line 19: I see that you do this in a few other files as well, which I understand. But why don't you just rename it in `config-helper.js` if you think it's too long of a name?

watson (approved):

packages/dd-trace/src/config.js
BridgeAR: I thought of that as well. I do think it is nicer to have the long name by default and only use the short version when having many of such calls. That is why I did it like that for now. In one file, I am actually considering to revert it.

watson: For what it's worth, `env` is a pretty commonly known abbreviation in Node.js thanks to `process.env` 🤷

packages/dd-trace/src/config.js
BridgeAR: I'll open a follow-up PR for it soon

packages/dd-trace/src/config.js
BridgeAR: I would keep it for now, while we can get back to it later, if we want to change it.

PR #6617, opened 2025-10-07
fix api docs links redirecting to incorrect urls

rochdev:
### What does this PR do?

Fix API docs links redirecting to incorrect URLs.

### Motivation

At some point the URLs changed, so we implemented a redirect mechanism from the old ones to the new ones. However, it looks like they changed back to the old URLs, presumably because we made some further changes to the types, so we now need to redirect from those new URLs to the original ones.

### DIFF
--- a/docs/add-redirects.sh
+++ b/docs/add-redirects.sh
@@ -10,6 +10,8 @@
 # visit the old URLs to the new URL.
 
 echo "writing redirects..."
+# TODO(2026-10-07): Delete this file and remove from docs/package.json
+# NOTE: Do not add any new entries to this list
 
 declare -a plugins=(
   "aerospike"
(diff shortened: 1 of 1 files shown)


docs/add-redirects.sh
```diff
@@ -1,38 +1,46 @@
...
-# interfaces/export_.plugins.connect.html
-# 
+# interfaces/plugins.amqp10.html
+#
 # This script automatically generates basic HTML files to redirect users who
 # visit the old URLs to the new URL.
 
 echo "writing redirects..."
 
 declare -a plugins=(
```
tlhunter, line 16: Since these redirects are mostly useful for SEO purposes how about we set a timeline on when to delete this file entirely?

Also since it's only useful for integrations that have existed up until now, and not for new integrations which won't ever have any links to their incorrect version, we can add a warning telling people not to add new entries.

```suggestion
# TODO(2026-10-07): Delete this file and remove from docs/package.json
# NOTE: Do not add any new entries to this list

declare -a plugins=(
```

tlhunter (approved):

PR #6622, opened 2025-10-08
remove npm/sandbox operations from fastify esm test

bengl:
### What does this PR do?
Removes `createSandbox` and related operations from the ESM tests for `fastify`.
### Motivation

The operations are unnecessary.


packages/datadog-plugin-fastify/test/integration-test/client.spec.js
```diff
@@ -37,7 +31,7 @@ describe('esm', () => {
     })
 
     it('is instrumented', async () => {
-      proc = await spawnPluginIntegrationTestProc(sandbox.folder, 'server.mjs', agent.port)
+      proc = await spawnPluginIntegrationTestProc(__dirname, 'server.mjs', agent.port, undefined, env)
```
BridgeAR: Nit: I think it would be nice to overload the helper to accept the envs as fourth argument. We can just check if it is a method or not to make that work. Passing through undefined is something I normally try to prevent.

packages/datadog-plugin-fastify/test/integration-test/client.spec.js
```diff
@@ -57,12 +51,12 @@ describe('esm', () => {
...
 
     it('Fastify import fastify is instrumented', async () => {
-      proc = await spawnPluginIntegrationTestProc(sandbox.folder, 'server2.mjs', agent.port)
+      proc = await spawnPluginIntegrationTestProc(__dirname, 'server2.mjs', agent.port, undefined, env)
 
       return curlAndAssertMessage(agent, proc, ({ headers, payload }) => {
         assert.propertyVal(headers, 'host', `127.0.0.1:${agent.port}`)
         assert.isArray(payload)
-        assert.strictEqual(checkSpansForServiceName(payload, 'fastify.request'), true)
+        assert.strictEqual(checkSpansForServiceName(payload, 'fastify.request'), true, undefined, env)
```
BridgeAR: These two arguments seem to not be expected?

```suggestion
        assert.strictEqual(checkSpansForServiceName(payload, 'fastify.request'), true)
```

packages/datadog-plugin-fastify/test/integration-test/client.spec.js
bengl: Fixed 👍

PR #6629, opened 2025-10-08
nightly testing

pabloerhard:
### What does this PR do?
A new manual workflow dispatcher triggers other workflows using the nightly Node version.
Fixes tests that broke with the nightly Node version.
Skips tests for versions that are incompatible with Node nightly.

When triggered any job using the latest version of node gets replaced by the nightly node version.

grpc – Modified version splitting to correctly include versions compatible with Node 25.
next - >=11.0.0 <13 – A dependency uses SlowBuffer, which is deprecated in Node 25.
avsc – Incompatible with version 5.0.0.
postgres – Not triggered by nightly tests because a native dependency cannot be installed with Node 25.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/.github/actions/node/action.yml
+++ b/.github/actions/node/action.yml
@@ -10,6 +10,12 @@ runs:
   steps:
     - uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
       with:
-        node-version: ${{ inputs.version == 'eol' && '16' || inputs.version == 'oldest' && '18' || inputs.version == 'maintenance' && '20' || inputs.version == 'active' && '22' || inputs.version == 'latest' && env.NIGHTLY_VERSION || '24' || inputs.version }}
+        node-version: ${{
+          inputs.version == 'eol' && '16' ||
+          inputs.version == 'oldest' && '18' ||
+          inputs.version == 'maintenance' && '20' ||
+          inputs.version == 'active' && '22' ||
+          inputs.version == 'latest' && (env.LATEST_VERSION || '24') ||
+          inputs.version }}
         check-latest: true
         registry-url: ${{ inputs.registry-url || 'https://registry.npmjs.org' }}
--- a/.github/workflows/apm-capabilities.yml
+++ b/.github/workflows/apm-capabilities.yml
@@ -8,20 +8,19 @@ on:
     - cron: 0 4 * * *
   workflow_dispatch:
     inputs:
-      nightly-version:
-        description: 'Nightly node version to use'
+      latest-version:
+        description: 'Node version to use'
         required: false
-        default: 'false'
         type: string
 
 
 concurrency:
-  group: ${{ inputs.nightly-version && format('{0}-{1}', github.workflow, inputs.nightly-version) || github.workflow }}-${{ github.ref || github.run_id }}"
+  group: ${{ github.workflow }}-${{ inputs.latest-version }}-${{ github.ref || github.run_id }}"
   cancel-in-progress: true
 
 env:
   MOCHA_OPTIONS: ${{ github.ref == 'refs/heads/master' && '--retries 1' || '' }}
-  NIGHTLY_VERSION: ${{ inputs.nightly-version && format('{0}-{1}', inputs.nightly-version, 'nightly') }}
+  LATEST_VERSION: ${{ inputs.latest-version }}
 
 jobs:
   tracing-macos:
--- a/.github/workflows/apm-integrations.yml
+++ b/.github/workflows/apm-integrations.yml
@@ -8,20 +8,19 @@ on:
     - cron: 0 4 * * *
   workflow_dispatch:
     inputs:
-      nightly-version:
-        description: 'Nightly node version to run'
+      latest-version:
+        description: 'Node version to run'
         required: false
-        default: 'false'
         type: string
 
 
 concurrency:
-  group: ${{ inputs.nightly-version == 'true' && format('{0}-{1}', github.workflow, inputs.nightly-version) || github.workflow }}-${{ github.ref || github.run_id }}"
+  group: ${{ github.workflow }}-${{ inputs.latest-version }}-${{ github.ref || github.run_id }}"
   cancel-in-progress: true
 
 env:
   MOCHA_OPTIONS: ${{ github.ref == 'refs/heads/master' && '--retries 1' || '' }}
-  NIGHTLY_VERSION: ${{ inputs.nightly-version && format('{0}-{1}', inputs.nightly-version, 'nightly') }}
+  LATEST_VERSION: ${{ inputs.latest-version }}
 
 # TODO: upstream jobs
 
--- a/.github/workflows/llmobs.yml
+++ b/.github/workflows/llmobs.yml
@@ -8,19 +8,18 @@ on:
     - cron: 0 4 * * *
   workflow_dispatch:
     inputs:
-      nightly-version:
-        description: 'Nightly node version to use'
+      latest-version:
+        description: 'Node version to use'
         required: false
-        default: 'false'
         type: string
 
 concurrency:
-  group: ${{ inputs.nightly-version && format('{0}-{1}', github.workflow, inputs.nightly-version)|| github.workflow }}-${{ github.ref || github.run_id }}"
+  group: ${{ github.workflow }}-${{ inputs.latest-version }}-${{ github.ref || github.run_id }}"
   cancel-in-progress: true
 
 env:
   MOCHA_OPTIONS: ${{ github.ref == 'refs/heads/master' && '--retries 1' || '' }}
-  NIGHTLY_VERSION: ${{ inputs.nightly-version && format('{0}-{1}', inputs.nightly-version, 'nightly') }}
+  LATEST_VERSION: ${{ inputs.latest-version }}
 
 jobs:
   sdk:
--- a/.github/workflows/nightly-dispatch.yml
+++ b/.github/workflows/nightly-dispatch.yml
@@ -32,6 +32,16 @@ on:
         required: false
         default: true
         type: boolean
+      profiling:
+        description: 'Profiling Tests'
+        required: false
+        default: true
+        type: boolean
+      appsec:
+        description: 'AppSec Tests'
+        required: false
+        default: true
+        type: boolean
 
 permissions:
   actions: write
@@ -50,10 +60,12 @@ jobs:
           - llmobs
           - platform
           - serverless
+          - appsec
+          - profiling
     runs-on: ubuntu-latest
     steps:
       - name: Trigger ${{matrix.workflow }} workflow
         if: ${{ inputs[matrix.workflow] }}
-        run: gh workflow run ${{matrix.workflow}}.yml --repo ${{ github.repository }} --ref ${{ env.BRANCH_NAME }} -f nightly-version=${{ inputs.node-version }}
+        run: gh workflow run ${{matrix.workflow}}.yml --repo ${{ github.repository }} --ref ${{ env.BRANCH_NAME }} -f latest-version=${{ inputs.node-version }}
         env:
           GH_TOKEN: ${{ github.token }}
--- a/.github/workflows/serverless.yml
+++ b/.github/workflows/serverless.yml
@@ -8,20 +8,19 @@ on:
     - cron: 0 4 * * *
   workflow_dispatch:
     inputs:
-      nightly-version:
-        description: 'Nightly node version to use'
+      latest-version:
+        description: 'Node version to use'
         required: false
-        default: 'false'
         type: string
 
 
 concurrency:
-  group: ${{ inputs.nightly-version && format('{0}-{1}', github.workflow, inputs.nightly-version) || github.workflow }}-${{ github.ref || github.run_id }}"
+  group: ${{ github.workflow }}-${{ inputs.latest-version }}-${{ github.ref || github.run_id }}"
   cancel-in-progress: true
 
 env:
   MOCHA_OPTIONS: ${{ github.ref == 'refs/heads/master' && '--retries 1' || '' }}
-  NIGHTLY_VERSION: ${{ inputs.nightly-version && format('{0}-{1}', inputs.nightly-version, 'nightly') }}
+  LATEST_VERSION: ${{ inputs.latest-version }}
 
 
 jobs:
--- a/packages/datadog-plugin-avsc/test/index.spec.js
+++ b/packages/datadog-plugin-avsc/test/index.spec.js
@@ -16,7 +16,7 @@ const {
 const sinon = require('sinon')
 const { loadMessage } = require('./helpers')
 const { SchemaBuilder } = require('../../dd-trace/src/datastreams/schemas/schema_builder')
-const { IS_NIGHTLY } = require('../../../version')
+const { NODE_MAJOR } = require('../../../version')
 
 const BASIC_USER_SCHEMA_DEF = JSON.parse(
   fs.readFileSync(path.join(__dirname, 'schemas/expected_user_schema.json'), 'utf8')
@@ -41,7 +41,7 @@ describe('Plugin', () => {
     let mockTime = 0
 
     // avsc version 5.0.0 currently does not support a nodeMajor version greater than major version 24
-    withVersions('avsc', ['avsc'], IS_NIGHTLY ? '>5.0.0' : undefined, (version) => {
+    withVersions('avsc', ['avsc'], NODE_MAJOR >= 25 ? '>5.0.0' : undefined, (version) => {
       before(() => {
         tracer = require('../../dd-trace').init()
         // reset sampled schemas
--- a/packages/datadog-plugin-grpc/test/client.spec.js
+++ b/packages/datadog-plugin-grpc/test/client.spec.js
@@ -12,7 +12,7 @@ const loader = require('../../../versions/@grpc/proto-loader').get()
 const { withNamingSchema, withPeerService, withVersions } = require('../../dd-trace/test/setup/mocha')
 const agent = require('../../dd-trace/test/plugins/agent')
 const { ERROR_MESSAGE, ERROR_TYPE, ERROR_STACK, GRPC_CLIENT_ERROR_STATUSES } = require('../../dd-trace/src/constants')
-const { NODE_MAJOR, IS_NIGHTLY } = require('../../../version')
+const { NODE_MAJOR } = require('../../../version')
 
 const pkgs = NODE_MAJOR > 14 ? ['@grpc/grpc-js'] : ['grpc', '@grpc/grpc-js']
 
@@ -78,7 +78,7 @@ describe('Plugin', () => {
       server.forceShutdown()
     })
 
-    withVersions('grpc', pkgs, IS_NIGHTLY ? '>=1.3.0' : undefined, (version, pkg) => {
+    withVersions('grpc', pkgs, NODE_MAJOR >= 25 && '>=1.3.0', (version, pkg) => {
       for (const clientName in clientBuilders) {
         const buildClient = clientBuilders[clientName]
 
--- a/packages/datadog-plugin-grpc/test/integration-test/client.spec.js
+++ b/packages/datadog-plugin-grpc/test/integration-test/client.spec.js
@@ -8,14 +8,14 @@ const {
 } = require('../../../../integration-tests/helpers')
 const { withVersions } = require('../../../dd-trace/test/setup/mocha')
 const { assert } = require('chai')
-const { IS_NIGHTLY } = require('../../../../version')
+const { NODE_MAJOR } = require('../../../../version')
 
 describe('esm', () => {
   let agent
   let proc
   let sandbox
 
-  withVersions('grpc', '@grpc/grpc-js', IS_NIGHTLY ? '>=1.3.0' : undefined, version => {
+  withVersions('grpc', '@grpc/grpc-js', NODE_MAJOR >= 25 && '>=1.3.0', version => {
     before(async function () {
       this.timeout(60000)
       sandbox = await createSandbox([`'@grpc/grpc-js@${version}'`, '@grpc/proto-loader'], false, [
--- a/packages/datadog-plugin-grpc/test/server.spec.js
+++ b/packages/datadog-plugin-grpc/test/server.spec.js
@@ -11,7 +11,7 @@ const agent = require('../../dd-trace/test/plugins/agent')
 
 const { ERROR_MESSAGE, ERROR_TYPE, ERROR_STACK, GRPC_SERVER_ERROR_STATUSES } = require('../../dd-trace/src/constants')
 
-const { NODE_MAJOR, IS_NIGHTLY } = require('../../../version')
+const { NODE_MAJOR } = require('../../../version')
 const pkgs = NODE_MAJOR > 14 ? ['@grpc/grpc-js'] : ['grpc', '@grpc/grpc-js']
 
 describe('Plugin', () => {
@@ -65,7 +65,7 @@ describe('Plugin', () => {
       server.forceShutdown()
     })
 
-    withVersions('grpc', pkgs, IS_NIGHTLY ? '>=1.3.0' : undefined, (version, pkg) => {
+    withVersions('grpc', pkgs, NODE_MAJOR >= 25 && '>=1.3.0', (version, pkg) => {
       describe('without configuration', () => {
         before(() => {
           return agent.load('grpc', { client: false })
--- a/packages/datadog-plugin-next/test/index.spec.js
+++ b/packages/datadog-plugin-next/test/index.spec.js
@@ -14,20 +14,20 @@ const { writeFileSync, readdirSync } = require('node:fs')
 const { withNamingSchema, withVersions } = require('../../dd-trace/test/setup/mocha')
 const agent = require('../../dd-trace/test/plugins/agent')
 const { rawExpectedSchema } = require('./naming')
-const { IS_NIGHTLY } = require('../../../version')
+const { NODE_MAJOR } = require('../../../version')
 
-const min = IS_NIGHTLY ? '>=13' : '>=11.0'
+const min = NODE_MAJOR >= 25 ? '>=13' : '>=11.1'
 
 describe('Plugin', function () {
   let server
   let port
 
+  // These next versions have a dependency which uses a deprecated node buffer
   describe('next', () => {
-    // These next versions have a dependency which uses a deprecated node buffer
     const satisfiesStandalone = version => satisfies(version, '>=12.0.0')
 
     // TODO: Figure out why 10.x tests are failing.
-    withVersions('next', 'next', `${min}`, version => {
+    withVersions('next', 'next', `${min} <15.4.1`, version => {
       const pkg = require(`../../../versions/next@${version}/node_modules/next/package.json`)
 
       const startServer = ({ withConfig, standalone }, schemaVersion = 'v0', defaultToGlobalService = false) => {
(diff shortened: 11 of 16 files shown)


.github/workflows/apm-capabilities.yml
```diff
@@ -6,13 +6,25 @@ on:
...
+        type: string
+      workflowName:
+        description: 'Workflow Name accessible from inputs'
+        required: false
+        default: 'NightlyCall'
+        type: string
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
+  group: ${{ inputs.runNightly == 'true' && inputs.workflowName || github.workflow }}-${{ github.ref || github.run_id }}"
```
rochdev: Why is this needed? It should be its own workflow run, meaning it should already get a proper unique value.

.github/workflows/apm-integrations.yml
```diff
@@ -677,15 +687,9 @@ jobs:
...
-      - uses: ./.github/actions/node/latest
-      - uses: ./.github/actions/install
-      - run: yarn test:plugins:ci
-      - if: always()
-        uses: ./.github/actions/testagent/logs
+      - uses: ./.github/actions/plugins/test
         with:
-          suffix: plugins-${{ github.job }}-${{ matrix.version }}-${{ matrix.range_clean }}
-      - uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
+          alternativeSuffix: plugins-${{ github.job }}-${{ matrix.range_clean }}
```
rochdev: Why is this needed?

.github/workflows/apm-integrations.yml
```diff
@@ -677,15 +687,9 @@ jobs:
...
-      - uses: ./.github/actions/testagent/start
-      - uses: ./.github/actions/node/latest
-      - uses: ./.github/actions/install
-      - run: yarn test:plugins:ci
-      - if: always()
-        uses: ./.github/actions/testagent/logs
+      - uses: ./.github/actions/plugins/test
         with:
-          suffix: plugins-${{ github.job }}-${{ matrix.version }}-${{ matrix.range_clean }}
-      - uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
```
rochdev, line 800: Why was this removed?

.github/workflows/llmobs.yml
```diff
@@ -6,13 +6,26 @@ on:
     branches: [master]
   schedule:
     - cron: 0 4 * * *
+  workflow_call:
```
rochdev: If the nightly use `workflow_call`, I would expect other triggers to also use that. Otherwise it should instead be `workflow_dispatch` with the nightly running it indirectly. There are trade-offs to both approaches, but we should use only one.

.github/workflows/nightly-dispatch.yml
```diff
@@ -1,4 +1,6 @@
 name: Nightly Testing
+env:
+  GLOBAL_VAR: "This is a global variable"
```
rochdev: Looks like a test leftover?

integration-tests/init.spec.js
```diff
@@ -21,7 +21,9 @@ const telemetryGood = ['complete', 'injection_forced:false']
 
 const { engines } = require('../package.json')
 const supportedRange = engines.node
-const currentVersionIsSupported = semver.satisfies(process.versions.node, supportedRange)
+const isNightly = process.versions.node.includes('nightly')
+const currentNodeVersion = isNightly ? process.versions.node.split('-')[0] : process.versions.node
```
rochdev: Any reason not to always do that instead of based on a condition?

packages/datadog-instrumentations/test/helpers/instrument.spec.js
```diff
@@ -41,7 +38,6 @@ describe('helpers/instrument', () => {
         storage('legacy').run('test2', () => {
           const tested = asyncResource.bind((a, b, c) => {
             expect(storage('legacy').getStore()).to.equal('test1')
-            expect(tested.asyncResource).to.equal(asyncResource)
```
rochdev, line 44: Why is this test removed?

packages/datadog-plugin-avsc/test/index.spec.js
```diff
@@ -41,6 +42,10 @@ describe('Plugin', () => {
     let mockTime = 0
 
     withVersions('avsc', ['avsc'], (version) => {
+      // avsc version 5.0.0 currently does not support a nodeMajor version greater than major version 24
+      if (version === '5.0.0' && nodeMajor >= 25) {
```
rochdev: This can be done with `withVersions('avsc', ['avsc'], '>5.0.0', (version) => {`.

packages/datadog-plugin-fs/test/index.spec.js
```diff
@@ -205,9 +205,9 @@ describe('Plugin', () => {
         describe('promises.open', () => {
           let fd
 
-          afterEach(() => {
-            if (typeof fd === 'number') {
-              realFS.closeSync(fd)
```
rochdev, line 210: Looks like this should work, is there a bug in Node?

packages/datadog-plugin-grpc/test/client.spec.js
```diff
@@ -79,6 +79,9 @@ describe('Plugin', () => {
     })
 
     withVersions('grpc', pkgs, (version, pkg) => {
+      if (!semver.satisfies(version, '>=1.3.0') && nodeMajor >= 25) {
```
rochdev: Similar to above, this is supported by `withVersions`.

packages/datadog-plugin-next/test/index.spec.js
```diff
@@ -14,12 +14,17 @@ const { writeFileSync, readdirSync } = require('node:fs')
...
 const agent = require('../../dd-trace/test/plugins/agent')
 const { rawExpectedSchema } = require('./naming')
+const nodeMajor = parseInt(process.versions.node.split('.')[0])
 
 describe('Plugin', function () {
   let server
   let port
 
   describe('next', () => {
+    // These next versions have a dependency which uses a depricated node buffer
```
rochdev:
```suggestion
    // These next versions have a dependency which uses a deprecated node buffer
```

packages/dd-trace/test/config.spec.js
```diff
@@ -2777,7 +2777,7 @@ describe('Config', () => {
 
     afterEach(() => {
       process.env = env
-      fs.rmdirSync(tempDir, { recursive: true })
```
rochdev, line 2815: Is this a removed API?

packages/datadog-plugin-next/test/integration-test/client.spec.js
```diff
@@ -12,14 +12,19 @@ const { withVersions } = require('../../../dd-trace/test/setup/mocha')
...
   let agent
   let proc
   let sandbox
   let variants
   // match versions tested with unit tests
+  if (process.env.PACKAGE_VERSION_RANGE === '>=11.0.0 <13' && nodeMajor >= 25) {
+    return
+  }
   withVersions('next', 'next', '>=11.1 <15.4.1', version => {
+    // These next versions have a dependency which uses a depricated node buffer
```
rochdev:
```suggestion
    // These next versions have a dependency which uses a deprecated node buffer
```

packages/dd-trace/test/config.spec.js
pabloerhard: Yep, https://nodejs.org/download/nightly/v25.0.0-nightly202510144d396ac9c6/docs/api/deprecations.html#DEP0147

.github/workflows/apm-capabilities.yml
pabloerhard: When a workflow is triggered through workflow dispatch github.workflow is then the name of the parent workflow, which made other jobs cancel out.

.github/workflows/apm-integrations.yml
pabloerhard: When a test has for example, ranges, we opted out of using reusable actions mainly because the suffix was different. Other than that, it used the same steps, so by adding that, we can now use reusable actions in more jobs.

packages/datadog-plugin-fs/test/index.spec.js
pabloerhard: Node 25 no longer does file closing during garbage collection, previously any file that was closed during garbage collection only threw a warning, now it throws an error.

pabloerhard: https://nodejs.org/download/nightly/v25.0.0-nightly202510144d396ac9c6/docs/api/deprecations.html#DEP0137

integration-tests/init.spec.js
pabloerhard: So do you mean something like  
`const currentNodeVersion = process.versions.node.split('-')[0] || process.versions.node`

.github/workflows/apm-integrations.yml
pabloerhard: It was removed to use an already existing action instead.

.github/workflows/llmobs.yml
pabloerhard: The nightly (parent) workflow uses workflow_dispatch, which allows it to be triggered manually. Any other workflow that is triggered by the parent workflow must use workflow_call, which indicates that it can be invoked by another workflow.

If we were to use only workflow_dispatch, we would only be able to run tests one by one, and we would need to specify that they are being run with the nightly Node version each time we trigger them.

packages/datadog-plugin-avsc/test/index.spec.js
pabloerhard: It would need to be something more like this `withVersions('avsc', ['avsc'], nodeMajor >= 25 ? '>5.0.0' : undefined, (version) => {`

packages/datadog-plugin-fs/test/index.spec.js
pabloerhard: We weren't even entering that if condition in node 24 ... node 24 took care of it during garbage collection, node 25 now enforces file handler closing.

.github/workflows/llmobs.yml
rochdev: > Any other workflow that is triggered by the parent workflow must use workflow_call, which indicates that it can be invoked by another workflow.

A workflow can still trigger another `workflow_dispatch` workflow.

integration-tests/init.spec.js
rochdev: Yes basically `const currentNodeVersion = process.versions.node.split('-')[0]`. If there is a dash we want the first part, otherwise we'll get the whole string anyway.

packages/datadog-plugin-avsc/test/index.spec.js
rochdev: Basically. Definitely a nitpick, but I thought I'd share that it can be done in that way.

packages/datadog-plugin-fs/test/index.spec.js
rochdev: Not sure I understand, we're still closing the same file handle in the same place, so what's the difference? I mean the change is fine in any case, I'm just not sure I understand why simply changing how the closing happens would fix the issue.

pabloerhard: So the file was never closing ... because we were never entering that if condition that closes the file, this wasn't an issue in other node versions because the file gets closed by the garbage collector, node 25 no longer does this and it throws an error.

integration-tests/init.spec.js
pabloerhard: Yep, makes sense.

packages/datadog-instrumentations/test/helpers/instrument.spec.js
pabloerhard: This behavior is deprecated on node 25. https://nodejs.org/download/nightly/v25.0.0-nightly20251015170848bc18/docs/api/deprecations.html#DEP0172

packages/datadog-plugin-avsc/test/index.spec.js
pabloerhard: It does look cleaner.

packages/datadog-plugin-avsc/test/index.spec.js
```diff
@@ -40,7 +41,9 @@ describe('Plugin', () => {
     let dateNowStub
     let mockTime = 0
 
-    withVersions('avsc', ['avsc'], (version) => {
+    withVersions('avsc', ['avsc'], nodeMajor >= 25 ? '>5.0.0' : undefined, (version) => {
+      // avsc version 5.0.0 currently does not support a nodeMajor version greater than major version 24
```
rochdev: Nit, but putting the comment above the line it explains would be more readable than below.

packages/datadog-plugin-grpc/test/integration-test/client.spec.js
```diff
@@ -8,13 +8,14 @@ const {
 } = require('../../../../integration-tests/helpers')
 const { withVersions } = require('../../../dd-trace/test/setup/mocha')
 const { assert } = require('chai')
+const nodeMajor = parseInt(process.versions.node.split('.')[0])
```
rochdev: This is already exposed by `const { NODE_MAJOR } = require('../../../../version')`. While both are one-liners, it's easier to find usage using the standardized file.

.github/workflows/serverless.yml
```diff
@@ -6,13 +6,22 @@ on:
     branches: [master]
   schedule:
     - cron: 0 4 * * *
+  workflow_dispatch:
+    inputs:
+      runNightly:
```
rochdev: Passing explicitly which version to use as latest would make the concurrency group easier. It would also allow running the workflow manually with a specific nightly if needed.

.github/workflows/serverless.yml
```diff
@@ -6,13 +6,22 @@ on:
...
+        description: 'Whether to run nightly tests'
+        required: false
+        default: 'false'
+        type: string
+
+env:
```
rochdev: Why was this moved in the file? Seems like an unnecessary change that is adding lines to review.

.github/workflows/nightly-dispatch.yml
```diff
@@ -29,27 +29,54 @@ on:
...
+permissions:
+  actions: write
+  contents: read
+
+env:
+  BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
+
 jobs:
-  nightly:
+  apm-integrations:
```
rochdev: A single job could be defined with a matrix instead. The wouldn't work super well with the `runX` inputs, but I'm still unclear what these do.

.github/workflows/apm-integrations.yml
```diff
@@ -642,9 +637,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version:
-          - 18
-          - latest
```
rochdev, line 759: Why the removal?

rochdev: Same question for Kafka.

rochdev: Looks like it could be to switch to running them in sequence. There was probably a performance reason for splitting them this way. Ideally this change can be avoided, and I think if the latest version is passed as input (as described in another of my comment) it would make this easier. If we can't avoid it, then the change is fine and we can revisit any performance challenges later.

pabloerhard: I only added these changes because I thought using reusable actions was better ... the changes are not needed. If this were done for performance then i can reset them.

packages/datadog-instrumentations/src/grpc/client.js
```diff
@@ -252,9 +252,18 @@ addHook({ name: '@grpc/grpc-js', versions: ['>=1.0.3 <1.1.4'], file: 'build/src/
...
+addHook({ name: '@grpc/grpc-js', versions: ['>=1.1.4 <1.3.0'], file: 'build/src/make-client.js' }, client => {
+  shimmer.wrap(client, 'makeClientConstructor', createWrapMakeClientConstructor(true))
+  shimmer.wrap(client, 'loadPackageDefinition', createWrapLoadPackageDefinition(true))
+
+  return client
+})
+
+addHook({ name: '@grpc/grpc-js', versions: ['>=1.3.0'] }, patch(true))
+
+addHook({ name: '@grpc/grpc-js', versions: ['>=1.3.0'], file: 'build/src/make-client.js' }, client => {
```
BridgeAR: I am not sure I understand the change here. Does this have any impact?

integration-tests/init.spec.js
```diff
@@ -21,7 +21,8 @@ const telemetryGood = ['complete', 'injection_forced:false']
 
 const { engines } = require('../package.json')
 const supportedRange = engines.node
-const currentVersionIsSupported = semver.satisfies(process.versions.node, supportedRange)
+const currentNodeVersion = process.versions.node.split('-')[0] || process.versions.node
```
BridgeAR: We could switch to the version file instead :)

packages/datadog-instrumentations/src/grpc/client.js
pabloerhard: It doesn't, I had added it for some previous changes, but it's not needed anymore. Thank you for catching that!

integration-tests/init.spec.js
pabloerhard: That specific process of version validation has to be done with the full version string and not just the major, unless you prefer doing something like `${NODE_MAJOR}.${NODE_MINOR}.${NODE_PATH}` or ideally also add the full version string into the version file which would result in keep the above logic, but now having the full version available for other files to also use.

Edit: nodeMatches already exposes the full versions as index 0.

.github/workflows/serverless.yml
pabloerhard: This has been removed and will be added on later PR. The objective was to use reusable actions when possible.

.github/actions/node/action.yml
```diff
@@ -10,6 +10,6 @@ runs:
   steps:
     - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
       with:
-        node-version: ${{ inputs.version == 'eol' && '16' || inputs.version == 'oldest' && '18' || inputs.version == 'maintenance' && '20' || inputs.version == 'active' && '22' || inputs.version == 'latest' && '24' || inputs.version }}
+        node-version: ${{ inputs.version == 'eol' && '16' || inputs.version == 'oldest' && '18' || inputs.version == 'maintenance' && '20' || inputs.version == 'active' && '22' || inputs.version == 'latest' && (env.RUN_NIGHTLY && format('{0}-{1}', env.RUN_NIGHTLY, 'nightly') || '24') || inputs.version }}
```
rochdev: The variable should be renamed to match its new behaviour.

rochdev: Also, the formatting wouldn't be needed either here or in the other files if it was something like `latest-versions` instead, which would default to `24` but can be changed to `nightly` or a specific value in other places.

.github/workflows/nightly-dispatch.yml
```diff
@@ -3,53 +3,52 @@ name: Nightly Testing
...
+
+env:
+  BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
+
 jobs:
-  nightly:
+  run-nightly:
+    strategy:
+      matrix:
+        workflow: [apm-integrations, apm-capabilities, llmobs, platform, serverless]
```
rochdev: Nit for now, but since this list will grow let's use the dash syntax instead.

.github/workflows/serverless.yml
rochdev: What I meant is that `env` was below `concurrency` before and was moved before it, which is an unnecessary change and adds noise to the PR.

integration-tests/init.spec.js
```diff
@@ -14,15 +14,14 @@ const fs = require('fs')
 const DD_INJECTION_ENABLED = 'tracing'
 const DD_INJECT_FORCE = 'true'
 const DD_TRACE_DEBUG = 'true'
-
+const { NODE_VERSION } = require('../version')
```
rochdev, line 17:
```suggestion
const { NODE_VERSION } = require('../version')

```

Keeping existing spacing.

integration-tests/init.spec.js
```diff
@@ -14,15 +14,14 @@ const fs = require('fs')
...
 const telemetryAbort = ['abort', 'reason:incompatible_runtime', 'abort.runtime', '']
 const telemetryForced = ['complete', 'injection_forced:true']
 const telemetryGood = ['complete', 'injection_forced:false']
 
 const { engines } = require('../package.json')
 const supportedRange = engines.node
-const currentVersionIsSupported = semver.satisfies(process.versions.node, supportedRange)
-
+const currentVersionIsSupported = semver.satisfies(NODE_VERSION, supportedRange, process.version.node)
```
rochdev:
```suggestion
const currentVersionIsSupported = semver.satisfies(NODE_VERSION, supportedRange, process.version.node)

```

Preserve existing spacing untouched.

rochdev: Also, is it expected that `satisfies` now receives 3 inputs?

packages/datadog-instrumentations/test/helpers/instrument.spec.js
rochdev: I'm a bit worried about removing tests, especially ones for something as fundamental as this. Can it be behind a Node version flag instead? Let's also add a TODO to rework this test with alternatives to AsyncResource.

packages/datadog-plugin-grpc/test/integration-test/client.spec.js
rochdev: Doesn't look like this change landed yet.

packages/datadog-plugin-next/test/integration-test/client.spec.js
```diff
@@ -12,14 +12,19 @@ const { withVersions } = require('../../../dd-trace/test/setup/mocha')
...
   let agent
   let proc
   let sandbox
   let variants
   // match versions tested with unit tests
+  if (process.env.PACKAGE_VERSION_RANGE === '>=11.0.0 <13' && nodeMajor >= 25) {
+    return
+  }
   withVersions('next', 'next', '>=11.1 <15.4.1', version => {
+    // These next versions have a dependency which uses a deprecated node buffer
```
rochdev:
```suggestion
  // These next versions have a dependency which uses a deprecated node buffer
  withVersions('next', 'next', '>=11.1 <15.4.1', version => {
```

Comments are usually above the line they are commenting on, not below.

Also, it looks like an existing comment got pushed up above the new condition, not sure that was expected.

There would also be another way to write this by generating the range, for example:

```js
const min = NODE_MAJOR >= 25 ? '>=13' : '>=11.1'
```

And then use that with `withVersions`:

```js
withVersions('next', 'next', `${min} <15.4.1`
```

The advantage is that the code wouldn't rely on a specific value for `PACKAGE_VERSION_RANGE` which could change in the future.

.github/actions/node/action.yml
```diff
@@ -10,6 +10,12 @@ runs:
   steps:
     - uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
       with:
-        node-version: ${{ inputs.version == 'eol' && '16' || inputs.version == 'oldest' && '18' || inputs.version == 'maintenance' && '20' || inputs.version == 'active' && '22' || inputs.version == 'latest' && '24' || inputs.version }}
+        node-version: ${{
+          inputs.version == 'eol' && '16' ||
+          inputs.version == 'oldest' && '18' ||
+          inputs.version == 'maintenance' && '20' ||
+          inputs.version == 'active' && '22' ||
+          inputs.version == 'latest' && (env.RUN_NIGHTLY || '24') ||
```
rochdev: This should be something like `LATEST_VERSION`.

.github/workflows/apm-capabilities.yml
```diff
@@ -6,13 +6,22 @@ on:
     branches: [master]
   schedule:
     - cron: 0 4 * * *
+  workflow_dispatch:
+    inputs:
+      nightly-version:
```
rochdev: Similar to the above, this should be something like `latest-version`. The fact that a nightly can be passed in is an implementation detail.

.github/workflows/apm-integrations.yml
```diff
@@ -6,13 +6,22 @@ on:
     branches: [master]
   schedule:
     - cron: 0 4 * * *
+  workflow_dispatch:
+    inputs:
+      nightly-version:
+        description: 'Nightly node version to run'
+        required: false
+        default: '24'
```
rochdev: In the current setup, the actual default is provided  by the underlying `node` action, so I would keep this as an empty string (so probably leave it undefined as that's probably the default for strings).

packages/datadog-plugin-grpc/test/client.spec.js
```diff
@@ -78,7 +78,7 @@ describe('Plugin', () => {
       server.forceShutdown()
     })
 
-    withVersions('grpc', pkgs, (version, pkg) => {
+    withVersions('grpc', pkgs, NODE_MAJOR >= 25 ? '>=1.3.0' : undefined, (version, pkg) => {
```
rochdev:
```suggestion
    withVersions('grpc', pkgs, NODE_MAJOR >= 25 && '>=1.3.0', (version, pkg) => {
```

packages/datadog-plugin-grpc/test/integration-test/client.spec.js
```diff
@@ -8,13 +8,14 @@ const {
...
 const { assert } = require('chai')
+const { NODE_MAJOR } = require('../../../../version')
 
 describe('esm', () => {
   let agent
   let proc
   let sandbox
 
-  withVersions('grpc', '@grpc/grpc-js', version => {
+  withVersions('grpc', '@grpc/grpc-js', NODE_MAJOR >= 25 ? '>=1.3.0' : undefined, version => {
```
rochdev:
```suggestion
  withVersions('grpc', '@grpc/grpc-js', NODE_MAJOR >= 25 && '>=1.3.0', version => {
```

packages/datadog-plugin-grpc/test/server.spec.js
```diff
@@ -65,7 +65,7 @@ describe('Plugin', () => {
       server.forceShutdown()
     })
 
-    withVersions('grpc', pkgs, (version, pkg) => {
+    withVersions('grpc', pkgs, NODE_MAJOR >= 25 ? '>=1.3.0' : undefined, (version, pkg) => {
```
rochdev:
```suggestion
    withVersions('grpc', pkgs, NODE_MAJOR >= 25 && '>=1.3.0', (version, pkg) => {
```

packages/datadog-plugin-next/test/index.spec.js
```diff
@@ -14,16 +14,20 @@ const { writeFileSync, readdirSync } = require('node:fs')
...
+const { NODE_MAJOR } = require('../../../version')
+
+const min = NODE_MAJOR >= 25 ? '>=13' : '>=11.1'
 
 describe('Plugin', function () {
   let server
   let port
 
   describe('next', () => {
+    // These next versions have a dependency which uses a deprecated node buffer
```
rochdev: This comment is on the wrong line.

.github/actions/node/action.yml
pabloerhard: I do see the input being named input.latest-version but the actual env defaults to input.latest-version + 'nightly' so that env variable will always reference to a nightly version.

PR #6640, opened 2025-10-10
test: rework esm tests to use insertVersionDep

BridgeAR:
The sandbox usage is very slow and causes tests to time out.

### DIFF
--- a/packages/datadog-plugin-mysql2/test/integration-test/client.spec.js
+++ b/packages/datadog-plugin-mysql2/test/integration-test/client.spec.js
@@ -6,20 +6,28 @@ const { assert } = require('chai')
 
 const {
   FakeAgent,
+  createSandbox,
   checkSpansForServiceName,
   spawnPluginIntegrationTestProc
 } = require('../../../../integration-tests/helpers')
-const { withVersions, insertVersionDep } = require('../../../dd-trace/test/setup/mocha')
+const { withVersions } = require('../../../dd-trace/test/setup/mocha')
+const { assert } = require('chai')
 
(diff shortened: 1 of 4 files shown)


packages/datadog-plugin-axios/test/integration-test/client.spec.js
```diff
@@ -1,27 +1,24 @@
 'use strict'
 
+const { join } = require('node:path')
+
```
bengl, line 4: Non-blocking Nit: unnecessary newline. We usually group all the requires together. Similarly elsewhere.

packages/dd-trace/test/setup/mocha.js
```diff
@@ -337,18 +337,25 @@ function getModulePath (moduleName, version) {
  * @param {string} version - The "version" string, as used with `withVersions`
  */
 function insertVersionDep (dir, pkgName, version) {
-  const nmDir = path.join(dir, 'node_modules')
+  const moduleNameParts = pkgName.split(/[/\\]/)
```
bengl, line 340: Are the changes here in order to deal with scoped pacakges?

BridgeAR: Yes. I know it is not the most elegant way 😅

packages/datadog-plugin-axios/test/integration-test/client.spec.js
BridgeAR: We changed that recently to separate them by type (Node.js modules, npm dependencies, and relative files)

PR #6648, opened 2025-10-13
[APPSEC] Http Downstream request analysis

IlyasShabi:
### What does this PR do?

[APPSEC-58483](https://datadoghq.atlassian.net/browse/APPSEC-58483) API10: Instrumentation & analysis of downstream requests.

We instrument only the `http.get` and `http.request` methods. We do not collect or send the request body to the WAF, as it’s not possible to access it before the first request lines are written.

- [x] Add env vars to configuration files
- [x] Instrument `res.on('data')` and `res.on('end')`
- [x] Instrument `res.read()`
- [x] Collect body chunks only when customer is listening to the response body 
- [x] Subscribe to DC to get body in ssrf.js
- [x] Send request method and headers to the WAF
- [x] Send response status, headers and body to the WAF
- [x] Get response body when only content-type is `application/json` or `application/x-www-form-urlencoded`
- [x] Implement body sampling algorithm
- [x] Report variant on telemetry and span tags
- [x] Unit tests
- [x] Integration tests

### Additional Notes

ST: https://github.com/DataDog/system-tests/pull/5530

[APPSEC-58483]: https://datadoghq.atlassian.net/browse/APPSEC-58483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/datadog-instrumentations/src/http/client.js
```diff
@@ -39,6 +41,162 @@ function normalizeCallback (inputOptions, callback, inputURL) {
...
+
+/**
+ * Resume switches the stream into flowing mode and drains the socket buffers.
+ * Node.js keeps queued data available via read() afterwards, so customer code can
+ * still consume the body later.
+ *
+ * @param {import('http').IncomingMessage} res
+ */
+function autoDrainResponse (res) {
+  if (!res || typeof res.resume !== 'function') {
```
CarlesDD:
```suggestion
  if (typeof res?.resume !== 'function') {
```

packages/dd-trace/src/config/index.js
```diff
@@ -496,6 +496,8 @@ class Config {
       DD_API_SECURITY_SAMPLE_DELAY,
       DD_API_SECURITY_ENDPOINT_COLLECTION_ENABLED,
       DD_API_SECURITY_ENDPOINT_COLLECTION_MESSAGE_LIMIT,
+      DD_API_SECURITY_DOWNSTREAM_REQUEST_BODY_ANALYSIS_SAMPLE_RATE,
+      DD_API_SECURITY_MAX_DOWNSTREAM_REQUEST_BODY_ANALYSIS,
```
CarlesDD: Can these settings only be set by environment variable, excluding programmatic configuration?

IlyasShabi: I asked @Anilm3 same question before, there is no programmatic API accutally

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,252 @@
...
+  const addresses = {}
+
+  if (res.statusCode !== undefined && res.statusCode !== null) {
+    addresses[HTTP_OUTGOING_RESPONSE_STATUS] = String(res.statusCode)
+  }
+
+  const headers = res.headers
+  if (headers && Object.keys(headers).length > 0) {
+    addresses[HTTP_OUTGOING_RESPONSE_HEADERS] = headers
+  }
```
CarlesDD, line 182: Would it make sense to capture trailer headers (`res.trailers`) here so WAF can analyze the full response header set, including end-of-stream metadata?

packages/datadog-instrumentations/src/http/client.js
```diff
@@ -39,6 +41,162 @@ function normalizeCallback (inputOptions, callback, inputURL) {
...
+  // Wrapping read() only taps what the customer already pulled,
+  // so back pressure stays unchanged.
+  const wrapRead = () => {
+    if (!shouldInstrumentData || readWrapped) return
+    if (typeof res.read !== 'function') return
+
+    readWrapped = true
+    originalRead = res.read
+
+    res.read = function wrappedRead () {
```
uurien: why aren't we using shimmer.wrap here?

packages/datadog-instrumentations/src/http/client.js
```diff
@@ -39,6 +41,162 @@ function normalizeCallback (inputOptions, callback, inputURL) {
...
+ *
+ * @param {object} ctx - Instrumentation context
+ * @param {import('http').IncomingMessage} res - The downstream response object.
+ * @returns {{ finalizeIfNeeded: () => void }|null} Cleanup helper used for drain.
+ */
+function setupResponseInstrumentation (ctx, res) {
+  const shouldInstrumentData = responseDataChannel.hasSubscribers
+  const shouldInstrumentFinish = responseFinishChannel.hasSubscribers
+
+  if (!shouldInstrumentData && !shouldInstrumentFinish) {
```
uurien: Does this make sense if we are only subscribed to one of these? 

```suggestion
  if (!shouldInstrumentData || !shouldInstrumentFinish) {
```

packages/datadog-instrumentations/src/http/client.js
```diff
@@ -39,6 +41,162 @@ function normalizeCallback (inputOptions, callback, inputURL) {
...
+      }
+
+      return chunk
+    }
+  }
+
+  onNewListener = (eventName) => {
+    if (eventName === 'data' && shouldInstrumentData && !dataHandler) {
+      dataHandler = publishChunk
+      res.on('data', dataHandler)
```
uurien: possible corner case.
Some application could subscribe to `'data'` get some chunks and after that unsubscribe without waiting to the end.

We'd be subscribed and getting more data that they are.

packages/datadog-instrumentations/src/http/client.js
```diff
@@ -39,6 +41,162 @@ function normalizeCallback (inputOptions, callback, inputURL) {
...
+
+  res.on('newListener', onNewListener)
+
+  // Remove every wrapper/listener we added so the response goes back to normal.
+  const cleanup = () => {
+    if (cleaned) return
+    cleaned = true
+
+    if (dataHandler) {
+      res.removeListener('data', dataHandler)
```
uurien: you are using `on` to subscribe and `removeListener` to unsubscribe, could you choose one? I mean, `on`, `off` combination or `addListener`, `removeListener`.

packages/dd-trace/src/appsec/rasp/ssrf.js
```diff
@@ -25,16 +41,108 @@ function analyzeSsrf (ctx) {
...
+}
+
+/**
+ * Collects outgoing response chunks when body sampling is enabled.
+ * @param {{
+ * ctx: object,
+ * chunk: Buffer|string|Uint8Array,
+ * res: import('http').IncomingMessage}} payload event payload from the channel.
+ */
+function handleResponseData ({ ctx, chunk, res }) {
```
uurien: Not sure if we should be doing this here, or in `client.js`. This whole logic is `http` module dependant, we should try to do module dependant logic in the module instrumentation, this could be different for other lib/frameworks.

I guess that the only reason this is here is because the sampling and the `includeBodies`, could we set a flag from `ctx` here to true to move this logic to `client.js`?

packages/dd-trace/src/appsec/rasp/ssrf.js
```diff
@@ -25,16 +41,108 @@ function analyzeSsrf (ctx) {
...
+  state.done = true
+
+  // Combine collected chunks into a single body (or null if no chunks)
+  let body = null
+  if (state.chunks?.length) {
+    const firstChunk = state.chunks[0]
+    body = typeof firstChunk === 'string'
+      ? state.chunks.join('')
+      : Buffer.concat(state.chunks)
+  }
```
uurien: if we move the logic in `handleResponseData` to client.js, this logic would be also there, and body would be send directly as parameter in the finish channel.

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,252 @@
...
+    maxDownstreamRequestBodyAnalysis,
+    downstreamRequestBodyAnalysisSampleRate
+  } = config.appsec.apiSecurity
+
+  const currentCount = perRequestBodyAnalysisCount.get(req) || 0
+  if (currentCount >= maxDownstreamRequestBodyAnalysis) {
+    return false
+  }
+
+  const samplingRate = Math.min(Math.max(downstreamRequestBodyAnalysisSampleRate, 0), 1)
```
uurien: what are we doing in other appsec use cases where the value in sampling (or other numeric field) is invalid? The value should be between 0 and 1, if someone missunderstood this and set 25 instead of 0.25, this will be translated to 1 and not to 0.25. 
IMHO if the value is incorrect, we should use the default one (maybe loging something)

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,252 @@
...
+
+  const hashed = (globalRequestCounter * KNUTH_FACTOR) % UINT64_MAX
+  const threshold = BigInt(Math.round(samplingRate * Number(UINT64_MAX)))
```
uurien: isn't a bit weird that we are converting BigInt to number to make an operation to convert it againt to BigInt? Aren't we loosing the whole point of using BigInt?

```suggestion
  // replace 1000n with the accuraccy that we want to maintain
  const threshold = BigInt(Math.round(samplingRate * 1000n)) * UINT64_MAX / 1000n
```

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,252 @@
...
+}
+
+/**
+ * Extracts response data for WAF analysis.
+ * @param {import('http').IncomingMessage} res downstream response object.
+ * @param {boolean} includeBody flag describing if the response body should be collected.
+ * @param {Buffer|string|object|null} responseBody response body.
+ * @returns {object} a map of addresses and response data.
+ */
+function extractResponseData (res, includeBody, responseBody) {
```
uurien: as this is called only from `ssrf` with `(res, !!responseBody, responseBody)` parameters, isn't includeBody redundant parameter?

packages/datadog-instrumentations/src/http/client.js
IlyasShabi: we can't use `shimmer.wrap()` here because it's wrapping the the prototype, so all requests will share one wrapper causing ctx collisions. Instead, we directly patch each instance like we do for `req.emit` and `req.setTimeout` to keep ctx isolated.

packages/dd-trace/src/appsec/downstream_requests.js
IlyasShabi: According to Anil's comment on RFC we should not consider trailers for now

packages/datadog-instrumentations/src/http/client.js
IlyasShabi: We keep collecting data even if the customer unsubscribes because we need the complete body for WAF evaluation. Partial bodies can't be parsed properly, making them unusable for the WAF.

packages/dd-trace/src/appsec/rasp/ssrf.js
IlyasShabi: I agree, your suggestion seems cleaner

packages/dd-trace/src/appsec/downstream_requests.js
IlyasShabi: I think this is more for documentation to clarify the usage of this env var. The problem with trying to interpret invalid values is we can't know the customer's intent.
I added a `log.warn` here to monitor what values customers are actually setting, so we can track if this becomes an issue in practice.

packages/dd-trace/src/appsec/rasp/ssrf.js
IlyasShabi: Done

packages/datadog-instrumentations/src/http/client.js
IlyasShabi: Yes agree. Done

packages/dd-trace/src/appsec/downstream_requests.js
IlyasShabi: Yes agree

packages/dd-trace/src/appsec/rasp/ssrf.js
```diff
@@ -25,16 +35,74 @@ function analyzeSsrf (ctx) {
...
-  const res = store?.res
-  handleResult(result, req, res, ctx.abortController, config, raspRule)
+  handleResult(result, req, store?.res, ctx.abortController, config, raspRule)
+
+  downstream.incrementDownstreamAnalysisCount(req)
+
+  // Track body analysis count if we're sampling the response body
+  if (ctx.shouldCollectBody) {
+    downstream.incrementBodyAnalysisCount(req)
+  }
```
simon-id: this could be directly called in shouldSampleBody()

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,323 @@
...
+/**
+ * Determines whether the current downstream request/responses bodies should be sampled for analysis.
+ * @param {import('http').IncomingMessage} req inbound request.
+ * @param {string} outgoingUrl the URL being requested (to check for redirect decisions).
+ * @returns {boolean} true when the downstream response body should be captured.
+ */
+function shouldSampleBody (req, outgoingUrl) {
+  // Check if there's a stored decision from a previous redirect
+  const storedDecision = getRedirectBodyCollectionDecision(req, outgoingUrl)
+  if (!storedDecision) false
```
simon-id: uhm, why does the test pass without the return ?

simon-id: also why does the linter pass ?

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,323 @@
...
+    maxDownstreamRequestBodyAnalysis,
+    downstreamRequestBodyAnalysisSampleRate
+  } = config.appsec.apiSecurity
+
+  const currentCount = bodyAnalysisCount.get(req) || 0
+  if (currentCount >= maxDownstreamRequestBodyAnalysis) {
+    return false
+  }
+
+  const samplingRate = Math.min(Math.max(downstreamRequestBodyAnalysisSampleRate, 0), 1)
```
simon-id: this line should be in enable()

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,323 @@
...
+
+  const isRedirect = res.statusCode >= 300 && res.statusCode < 400
+  const redirectLocation = res.headers?.location || res.headers?.Location || ''
+
+  if (isRedirect && redirectLocation) {
+    // Store the body collection decision for the redirect target
+    storeRedirectBodyCollectionDecision(req, redirectLocation)
+  }
+
+  return { isRedirect, redirectLocation }
```
simon-id: redirectLocation is never used ?

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,323 @@
...
+
+/**
+ * Checks if a response is a redirect
+ * @param {import('http').IncomingMessage} req inbound request.
+ * @param {import('http').IncomingMessage} res downstream response object.
+ * @param {boolean} shouldCollectBody current body collection decision.
+ * @returns {{isRedirect: boolean, redirectLocation: string|null}} redirect information.
+ */
+function handleRedirectResponse (req, res, shouldCollectBody) {
+  if (!shouldCollectBody) return { isRedirect: false, redirectLocation: null }
```
simon-id: this condition should be checked in the caller parent function, not here, to not hide the logic, plus it doesn't make sense with the naming otherwise

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,323 @@
...
+}
+
+/**
+ * Adds tracing telemetry for matches triggered by downstream responses.
+ * @param {import('http').IncomingMessage} req inbound reques.
+ * @param {object} raspRule rule with type and variant
+ */
+function handleResponseTracing (req, raspRule) {
+  updateRaspRuleMatchMetricTags(req, raspRule, false, false)
+}
```
simon-id: is it necessary to hide the telemtry function behind a proxy function ? it just kinda obfuscate what the function does imo

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,323 @@
...
+
+/**
+ * Gets the content-type header value from a case-insensitive headers object.
+ * @param {import('http').IncomingHttpHeaders|object|null} headers response headers object.
+ * @returns {string|null} content-type value
+ */
+function getResponseContentType (headers) {
+  if (!headers) return null
+
+  return headers['content-type'] || headers['Content-Type'] || headers['CONTENT-TYPE'] || null
```
simon-id: [Bizarre](https://www.youtube.com/watch?v=lQK4F1qQlGI)

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,323 @@
...
+    if (mime === 'application/json' || mime === 'text/json') {
+      if (typeof body === 'string') {
+        return JSON.parse(body)
+      }
+
+      if (Buffer.isBuffer(body)) {
+        return JSON.parse(body.toString('utf8'))
+      }
+
+      if (typeof body === 'object' && body !== null) {
```
simon-id:
```suggestion
      if (typeof body === 'object') {
```
unnecessary because of the return null after this if

simon-id: can this really be an object anyway ? maybe it can, but in that case the urlencoded if could also be an object no ? either both can be or both cannot be ?

packages/dd-trace/src/appsec/downstream_requests.js
IlyasShabi: 😱😱😱😱😱

packages/dd-trace/src/appsec/downstream_requests.js
IlyasShabi: Changes applied to Location and Content-Type, Node.js will always return one value

packages/dd-trace/src/appsec/downstream_requests.js
IlyasShabi: The  `if (typeof body === 'object')` is also unnecessary because we will get either a Buffer or string from instrumentation

IlyasShabi: Actually we cannot have a body with type object from [instrumentation](https://github.com/DataDog/dd-trace-js/pull/6648/files#diff-f2ef5bcb789c6faba0e48be5912b4f02a3a7b2ea465b8e6c83ff430c268309a2R154)
we can have only Buffer or string

packages/dd-trace/src/appsec/downstream_requests.js
simon-id: i'm reopning this thread because you didn't reply to the question "why does the tests pass with this bug ?" sounds like missing coverage

packages/dd-trace/src/appsec/addresses.js
```diff
@@ -27,6 +27,11 @@ module.exports = {
   WAF_CONTEXT_PROCESSOR: 'waf.context.processor',
 
   HTTP_OUTGOING_URL: 'server.io.net.url',
+  HTTP_OUTGOING_METHOD: 'server.io.net.request.method',
+  HTTP_OUTGOING_HEADERS: 'server.io.net.request.headers',
+  HTTP_OUTGOING_RESPONSE_STATUS: 'server.io.net.response.status',
+  HTTP_OUTGOING_RESPONSE_HEADERS: 'server.io.net.response.headers',
+  HTTP_OUTGOING_RESPONSE_BODY: 'server.io.net.response.body',
   FS_OPERATION_PATH: 'server.io.fs.file',
```
simon-id, line 36:
```suggestion
  HTTP_OUTGOING_RESPONSE_BODY: 'server.io.net.response.body',

  FS_OPERATION_PATH: 'server.io.fs.file',
```

packages/dd-trace/test/config/index.spec.js
simon-id, line 1: missing asserts in `should initialize with the correct defaults`

packages/dd-trace/src/appsec/rasp/ssrf.js
```diff
@@ -25,16 +36,69 @@ function analyzeSsrf (ctx) {
...
+ *   res: import('http').IncomingMessage,
+ *   body: string|Buffer|null
+ * }} payload event payload from the channel.
+ */
+function handleResponseFinish ({ ctx, res, body }) {
+  if (!res) return
+
+  const store = storage('legacy').getStore()
+  const req = store?.req
+  if (!req) return
```
simon-id: you have req and res living together, but req is the top incoming request, and res is the http client response, maybe worth a comment that here, "req" and "res" are completely separate

packages/dd-trace/src/appsec/rasp/ssrf.js
```diff
@@ -25,16 +36,69 @@ function analyzeSsrf (ctx) {
...
+
+  const evaluateBody = ctx.shouldCollectBody && downstream.handleRedirectResponse(req, res)
+
+  if (evaluateBody) {
+    // Skip body analysis for redirect responses
+    runResponseEvaluation(res, req, null)
```
simon-id:
```suggestion
  // Skip body analysis for redirect responses
  const evaluateBody = ctx.shouldCollectBody && downstream.handleRedirectResponse(req, res) 

  if (evaluateBody) {
    runResponseEvaluation(res, req, null)
```

packages/dd-trace/src/appsec/rasp/ssrf.js
```diff
@@ -25,16 +36,69 @@ function analyzeSsrf (ctx) {
...
+  } else {
+    runResponseEvaluation(res, req, body)
+  }
+}
+
+/**
+ * Evaluates the downstream response and records telemetry.
+ * @param {import('http').IncomingMessage} res outgoing response object.
+ * @param {import('http').IncomingMessage} req originating inbound request.
```
simon-id: either
- outgoing -> incoming
or
- inbound -> outbound
but not a mix of both
```suggestion
 * @param {import('http').IncomingMessage} res outgoing response object.
 * @param {import('http').IncomingMessage} req originating incoming request.
```

packages/dd-trace/src/appsec/rasp/ssrf.js
```diff
@@ -25,16 +36,69 @@ function analyzeSsrf (ctx) {
...
+  if (!Object.keys(responseAddresses).length) return
+
+  const raspRule = { type: RULE_TYPES.SSRF, variant: 'response' }
+  const result = waf.run({ ephemeral: responseAddresses }, req, raspRule)
+
+  const ruleTriggered = !!result?.events?.length
+
+  if (ruleTriggered) {
+    updateRaspRuleMatchMetricTags(req, raspRule, false, false)
+  }
```
simon-id, line 99:
```suggestion
  // TODO: this should be done in the waf functions directly instead of calling it everywhere
  const ruleTriggered = !!result?.events?.length

  if (ruleTriggered) {
    updateRaspRuleMatchMetricTags(req, raspRule, false, false)
  }
```

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,308 @@
...
+  if (samplingRate !== bodyAnalysisSampleRate) {
+    log.warn(
+      'DD_API_SECURITY_DOWNSTREAM_REQUEST_BODY_ANALYSIS_SAMPLE_RATE value is %s and it\'s out of range',
+      bodyAnalysisSampleRate)
+  }
+}
+
+function disable () {
+  config = null
+  globalRequestCounter = 0n
```
simon-id:
```suggestion
  globalRequestCounter = null
```

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,308 @@
...
+/**
+ * Gets the content-type header value from response headers.
+ * @param {import('http').IncomingHttpHeaders|object|null} headers response headers object.
+ * @returns {string} content-type value
+ */
+function getResponseContentType (headers) {
+  if (!headers) return null
+
+  return headers['content-type'] || null
+}
```
simon-id: do you really need a function for that ?
just inline this where it's called no ?
```suggestion
headers?.['content-type']
```

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,308 @@
...
+ */
+function getRedirectBodyCollectionDecision (req, outgoingUrl) {
+  const decisions = redirectBodyCollectionDecisions.get(req)
+  if (!decisions) return false
+
+  const decision = decisions.has(outgoingUrl)
+  if (!decision) return false
+
+  decisions.delete(outgoingUrl)
+  return true
```
simon-id: this is equivalent no ?
```suggestion
  return decisions.delete(outgoingUrl)
```

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,308 @@
...
+  redirectBodyCollectionDecisions = null
+}
+
+/**
+ * Check we have a stored redirect body collection decision for a given URL.
+ * @param {import('http').IncomingMessage} req inbound request.
+ * @param {string} outgoingUrl the URL being requested.
+ * @returns {boolean} the stored decision
+ */
+function getRedirectBodyCollectionDecision (req, outgoingUrl) {
```
simon-id: more like "consumeRedirectBodyCollectionDecision" since you delete it if it exists

packages/dd-trace/src/appsec/downstream_requests.js
IlyasShabi: I added tests more tests for better coverage for `shouldSampleBody`

packages/dd-trace/test/config/index.spec.js
simon-id: @IlyasShabi still missing some of them, read the whole test!

packages/dd-trace/src/appsec/rasp/ssrf.js
```diff
@@ -25,16 +36,66 @@ function analyzeSsrf (ctx) {
...
+  // downstream response object
+  if (!res) return
+
+  const store = storage('legacy').getStore()
+  const originatingRequest = store?.req
+  if (!originatingRequest) return
+
+  // Skip body analysis for redirect responses
+  const evaluateBody = ctx.shouldCollectBody && downstream.handleRedirectResponse(originatingRequest, res)
+  runResponseEvaluation(res, originatingRequest, evaluateBody ? null : body)
```
simon-id: oopsie ?
```suggestion
  runResponseEvaluation(res, originatingRequest, evaluateBody ? body : null)
```

packages/dd-trace/src/appsec/rasp/ssrf.js
simon-id: @IlyasShabi i think you misunderstood my point, you changed the wrong text no ?

packages/dd-trace/src/appsec/rasp/ssrf.js
simon-id: uhm @IlyasShabi , why are the test passing ? this should be breaking the tests no ?

IlyasShabi: Oh shit! thanks for catching this. The tests pass because `evaluateBody` value is wrong. I fixed it and add more tests

packages/dd-trace/src/appsec/rasp/ssrf.js
IlyasShabi: No I changed the right one, but I make it more clear now

packages/dd-trace/test/config/index.spec.js
IlyasShabi: I added more tests

packages/datadog-instrumentations/src/http/client.js
```diff
@@ -39,6 +40,124 @@ function normalizeCallback (inputOptions, callback, inputURL) {
...
+      bodyConsumed = true
+
+      // For 'data' events, add our own listener to collect chunks
+      if (eventName === 'data' && !dataListenerAdded) {
+        dataListenerAdded = true
+        res.on('data', collectChunk)
+      }
+
+      // For 'readable' events, wrap the read() method
+      if (eventName === 'readable' && !originalRead && typeof res.read === 'function') {
```
uurien: when a response is readable, is the event `data` executed? If it is, we could be collecting the data twice.

packages/datadog-instrumentations/src/http/client.js
```diff
@@ -39,6 +40,124 @@ function normalizeCallback (inputOptions, callback, inputURL) {
...
+ * @param {import('http').IncomingMessage} res
+ */
+function resumeResponse (res) {
+  if (!res) return
+  if (typeof res.resume !== 'function') return
+  if (res.destroyed) return
+  if (res.readableFlowing) return
+  if (res.readable === false) return
+
+  res.resume()
```
uurien: why should we resume? In which case can we pause? we are subscribing with the app and wrapping the read method, not calling to it.

packages/dd-trace/src/appsec/rasp/ssrf.js
```diff
@@ -25,16 +36,67 @@ function analyzeSsrf (ctx) {
...
+  // Skip body analysis for redirect responses
+  const evaluateBody = ctx.shouldCollectBody && !downstream.handleRedirectResponse(originatingRequest, res)
+  const responseBody = evaluateBody ? body : null
+  runResponseEvaluation(res, originatingRequest, responseBody)
+}
+
+/**
+ * Evaluates the downstream response and records telemetry.
+ * @param {import('http').IncomingMessage} res incoming response from downstream service.
+ * @param {import('http').IncomingMessage} req originating request.
```
uurien, line 83: Are both `IncomingMessage`? Isn't one of then `OutgoingMessage`?

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,292 @@
...
+let config
+let samplingRate
+let globalRequestCounter
+let bodyAnalysisCount
+let downstreamAnalysisCount
+let redirectBodyCollectionDecisions
+
+function enable (_config) {
+  config = _config
+  globalRequestCounter = 0n
```
uurien: We are never reseting this number, this would grow until infinite. Is that expected? shouldn't we set to 0 again when some max value is raised?

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,292 @@
...
+  if (headers && Object.keys(headers).length > 0) {
+    addresses[HTTP_OUTGOING_HEADERS] = headers
+  }
+
+  return addresses
+}
+
+/**
+ * Checks if a response is a redirect
+ * @param {import('http').IncomingMessage} req outgoing request.
```
uurien: IIUC, this is not the outgoing request, this is the "main" request, the incoming request from the browser or whatever is using the endpoint. This is wrong in multiple places.

packages/dd-trace/src/appsec/rasp/ssrf.js
uurien: I've realized that yes, are the same, one is from the original server incoming message and the other is the downstream request result. sorry

packages/dd-trace/test/appsec/downstream_requests.spec.js
```diff
@@ -0,0 +1,493 @@
...
+        statusCode: 200,
+        headers: {},
+      }
+
+      const isRedirect = downstream.handleRedirectResponse(req, res, true)
+
+      assert.strictEqual(isRedirect, false)
+    })
+
+    it('returns false for redirect without location header', () => {
```
uurien, line 145: test is checking that returns `true`. And the test says false. False seems to be the right one, probably something wrong in the test?

packages/dd-trace/test/appsec/downstream_requests.spec.js
```diff
@@ -0,0 +1,493 @@
...
+      require('sinon').assert.calledWith(span.setTag, '_dd.appsec.downstream_request', 1)
+      require('sinon').assert.calledWith(span.setTag, '_dd.appsec.downstream_request', 2)
+      require('sinon').assert.calledWith(span.setTag, '_dd.appsec.downstream_request', 3)
+      webRootStub.restore()
+    })
+
+    it('does not error when span is null', () => {
+      const webRootStub = require('sinon').stub(web, 'root').returns(null)
+
+      assert.doesNotThrow(() => downstream.incrementDownstreamAnalysisCount(req))
```
uurien: In `AGENTS.md` we have this:

> Never use the `doesNotThrow()` assertion. Instead, execute the method directly.

If it should be respected by ai agents, seems that should be respected also by humans :-) 

```suggestion
      downstream.incrementDownstreamAnalysisCount(req)
```

packages/dd-trace/src/appsec/downstream_requests.js
```diff
@@ -0,0 +1,292 @@
...
+function handleRedirectResponse (req, res) {
+  const isRedirect = res.statusCode >= 300 && res.statusCode < 400
+  const redirectLocation = res.headers?.location || ''
+
+  if (isRedirect && redirectLocation) {
+    // Store the body collection decision for the redirect target
+    storeRedirectBodyCollectionDecision(req, redirectLocation)
+  }
+
+  return isRedirect
```
uurien: Probably what you want is:
```suggestion
  return isRedirect && redirectLocation
```

packages/dd-trace/test/appsec/downstream_requests.spec.js
```diff
@@ -0,0 +1,493 @@
...
+        setTag: require('sinon').stub(),
+      }
+    })
+
+    afterEach(() => {
+      require('sinon').restore()
+    })
+
+    it('increments count and sets metric on span', () => {
+      const webRootStub = require('sinon').stub(web, 'root').returns(span)
```
uurien: why require `sinon` here and not in the file headers?

packages/dd-trace/test/appsec/downstream_requests.integration.spec.js
```diff
@@ -0,0 +1,216 @@
+'use strict'
```
uurien, line 1: shouldn't this file be in the same level as `downstream_requests.spec.js`? This is in `rasp` and the other is not.

packages/dd-trace/test/appsec/downstream_requests.integration.spec.js
```diff
@@ -0,0 +1,216 @@
...
+
+const assert = require('node:assert/strict')
+const path = require('path')
+const Axios = require('axios')
+const { sandboxCwd, useSandbox, FakeAgent, spawnProc } = require('../../../../../integration-tests/helpers')
+
+describe('RASP - downstream request integration', () => {
+  let cwd, appFile
+
+  useSandbox(
```
uurien, line 11: if this is using `useSandbox`, why do you need to have `express.plugin.spec.js`?

packages/dd-trace/test/appsec/downstream_requests.integration.spec.js
```diff
@@ -0,0 +1,216 @@
...
+        const evalVariants = new Set()
+        for (const s of evalSeries) {
+          if (hasTag(s, 'rule_variant:request')) evalVariants.add('request')
+          if (hasTag(s, 'rule_variant:response')) evalVariants.add('response')
+        }
+        assert.strictEqual(evalVariants.has('request'), true, 'rasp.rule.eval should include request variant')
+        assert.strictEqual(evalVariants.has('response'), true, 'rasp.rule.eval should include response variant')
+
+        const matchSeries = series.filter(s => s.metric === 'rasp.rule.match')
+        assert.ok(matchSeries, 'Rasp match eval series should exist')
```
uurien:
```suggestion
        assert.ok(matchSeries, 'Rasp rule match series should exist')
```

packages/dd-trace/src/config/index.js
uurien: This env var is wrong, it changed in the RFC, the good one is `DD_API_SECURITY_DOWNSTREAM_BODY_ANALYSIS_SAMPLE_RATE`
```suggestion
      DD_API_SECURITY_DOWNSTREAM_REQUEST_BODY_ANALYSIS_SAMPLE_RATE,
      DD_API_SECURITY_DOWNSTREAM_BODY_ANALYSIS_SAMPLE_RATE,
```

packages/dd-trace/test/appsec/downstream_requests.spec.js
```diff
@@ -0,0 +1,493 @@
...
+  })
+
+  describe('handleRedirectResponse', () => {
+    it('detects redirect with location header', () => {
+      const res = {
+        statusCode: 302,
+        headers: { location: 'http://example.com/redirect' },
+      }
+
+      const isRedirect = downstream.handleRedirectResponse(req, res, true)
```
CarlesDD: Nit: `handleRedirectResponse` signature has only two params, but here there are tree. It happens in several places in this test suite.

packages/dd-trace/test/appsec/rasp/ssrf.spec.js
```diff
@@ -18,9 +56,27 @@ describe('RASP - ssrf.js', () => {
...
 
+    downstream = {
+      enable: sinon.stub(),
+      disable: sinon.stub(),
+      shouldSampleBody: sinon.stub().returns(true),
+      handleRedirectResponse: sinon.stub().returns(false),
+      extractRequestData: sinon.stub().returns({}),
+      extractResponseData: sinon.stub().returns({}),
+      incrementDownstreamAnalysisCount: sinon.stub(),
+      handleResponseTracing: sinon.stub(),
```
CarlesDD: This stub could be removed

packages/datadog-instrumentations/src/http/client.js
uurien: This example makes me see that yes, the readable and `data` events can be used together, in that case we will be collecting wrong data.

I'll be change this a bit to read only from the first one when more than one listener are applied.

```javascript
const https = require('node:https')

https.get('https://www.datadoghq.com', (res) => {
  let body = ''
  let bodyFromData = ''
  res.on('data', (chunk) => {
    bodyFromData += chunk.toString()
  })
  res.on('readable', () => {
    let chunk
    while ((chunk = res.read()) !== null) {
      body += chunk.toString()
    }
  })
  res.on('end', () => {
    console.log('Status:', res.statusCode)
    console.log('Body length:', body.length)
    console.log('Body from data:', bodyFromData.length)
    console.log('body === bodyFromData', body === bodyFromData)
  })
})
```

Output:
```
Status: 200
Body length: 261709
Body from data: 261709
body === bodyFromData true
```

packages/dd-trace/src/appsec/downstream_requests.js
uurien: wrong description, right type.

packages/datadog-instrumentations/src/http/client.js
uurien: we shouldn't resume, this piece of code is now deleted. we could modify app behavior with this.

PR #1367, opened 2021-05-17
[ci-app] Add JUnit XML Report Upload

juan-fernandez:
### What does this PR do?

Add a step to upload the jUnit XML report from core unit tests to Datadog.

### Motivation
* Dogfood jUnit XML upload.
* Provide visibility into core's unit tests.

PR #6649, opened 2025-10-13
chore (ci): report test result to datadog

cbeauchesne:
### What does this PR do?

* Adds junit report on mocha based tests an system-tests,  then report them to datadog

### Changes

* [x] change `.mocharc.yml` to `.mocharc.js`
* [x] Add multi-reporter and `spec` + `junit` export, with node version in filename (note, the config of mocha is not trivial, I failed to add all conf in `.mocharc.js` and I had to add an extra conf file 😢 )
* [x] Report system-tests to Datadog [fca1709](https://github.com/DataDog/dd-trace-js/pull/6649/commits/fca1709c9f9bb9a91e7386d17ffe2fe6f09f2624)
* [x] Report all tests to Datadog

Extra : adds `.tap/*` in gitignore (not related to this PR, I can remove it if needed)

Good to know : https://github.com/DataDog/dd-trace-js/pull/1367

### Not in this PR 

* tap: There is no version that support both nyc (dropped in `v18` in favor of `c8`)  and junit (introduced in `v18`). The used version is quite old, so it may be a good idea to update it. Though, using two different coverage tool is not an option, so this part must be addressed in a different PR. This current Pr will focus on `mocha` and `system-tests`. Side note, do we want this? `mocha` seems to be the [target](https://github.com/DataDog/dd-trace-js/pull/4728).
* `test:eslint-rules` -> only lint, will see later if needed

### Motivation

[](https://docs.google.com/document/d/1y9imdgzDa3pxvy26DSpi2qMPjBlLAV_G9wk-ct55sUc/edit?pli=1&tab=t.0#heading=h.ihwtku88eswv)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -117,7 +117,6 @@ docs/test.js
 dist
 acmeair-nodejs
 !packages/*/test/**/dist
-!.github/actions/push-junit/dist
 !packages/*/test/**/node_modules
 packages/datadog-plugin-next/test/package.json
 packages/datadog-plugin-next/test/node_modules


.github/actions/push-junit/post.js
```diff
@@ -0,0 +1,97 @@
...
+    await exec.exec(binPath, ['--version']);
+
+    // Build args & env
+    const args = ['junit', 'upload', pattern];
+    if (verbose) args.push('--verbose');
+
+    const env = {
+      ...process.env,
+      DATADOG_API_KEY: apiKey,
+      DATADOG_SITE: site
```
juan-fernandez: let's use `DD_API_KEY` and `DD_SITE`. `DATADOG_` env vars are deprecated and might be removed at a later major version

.github/actions/install/action.yml
```diff
@@ -24,3 +24,9 @@ runs:
     - run: tar -cf node_modules.tar node_modules
       shell: bash
       if: inputs.cache == 'true' && steps.yarn-cache.outputs.cache-hit != 'true'
+    - uses: ./.github/actions/push-junit
+      with:
+        dd-api-key: "wip"  # ${{ secrets.TEST_OPTIMIZATION_API_KEY }}
+      # Convenient way of deploying push to test optim everywhere, without adding this step everywhere
+      # It also ensure that no job will be forgotten, as long as it uses the .github/action/install action
+      # The trick is as the .github/actions/push-junit performs the push as a post step, so after everything else
```
juan-fernandez: is the "post step" requirement why we can't use https://github.com/DataDog/junit-upload-github-action?

.github/actions/push-junit/post.js
cbeauchesne: Addressed in [d046003](https://github.com/DataDog/dd-trace-js/pull/6649/commits/d046003bfd4344c68dfcbcfafe687bb3b9dfadd2)

.github/actions/install/action.yml
cbeauchesne: Seen together on zoom : it'd be great to use it (I wasn't aware of its existence). Unfortunately, if we choose to use the trick of shipping install `actions/install`, we must execute the push inside the `post` step.

cbeauchesne: Actually, composite action does not have access to secrets ... The modification on all call is mandatory. Better to use the official action then.

PR #6653, opened 2025-10-14
[DB plugins] trigger sampling of span before injecting the trace parent comment

na-ji:
### What does this PR do?

This PR mimics the way [dd-trace-py is computing a span sampling priority](https://github.com/DataDog/dd-trace-py/blob/17b60e4cbee7ea3335879a8da09e5576abdd6e59/ddtrace/propagation/_database_monitoring.py#L73-L78) in database plugins before a span ends. 

### Motivation

The database plugins supported by the Database Monitoring require injecting the trace parent as a comment into the query statement. The trace parent to be fully valid, needs to contain the correct value for the "sampled" flag. However, we used to be injecting the trace parent before the decision to sample a span or not had been taken. Because of that, all the trace parents injected had the sampled flag set to 00. This PR fixes this issue by running the sampling decision before injecting the comment when needed.

### Plugin Checklist

- [x] Unit tests.
- [x] Integration tests.
- [x] Benchmarks.
- [ ] TypeScript [tests][2].
- [x] API [documentation][3].
- [x] CI [jobs/workflows][4].

### Additional Notes

Linked PR: https://github.com/DataDog/dd-trace-api-js/pull/14

### DIFF
--- a/packages/datadog-plugin-mysql/test/index.spec.js
+++ b/packages/datadog-plugin-mysql/test/index.spec.js
@@ -468,6 +468,8 @@ describe('Plugin', () => {
           connection.end(() => {
             agent.close({ ritmReset: false }).then(done)
           })
+
+          tracer._tracer.configure({ env: 'tester', sampler: { sampleRate: 1 } })
         })
 
         beforeEach(async () => {
@@ -504,7 +506,7 @@ describe('Plugin', () => {
 
           agent.assertSomeTraces(traces => {
             expect(queryText).to.include('-00\'*/ SELECT 1 + 1 AS solution')
-          }).then(done, done).finally(() => tracer._tracer.configure({ env: 'tester', sampler: { sampleRate: 1 } }))
+          }).then(done, done)
 
           connection.query('SELECT 1 + 1 AS solution', () => {
             queryText = connection._protocol._queue[0].sql
@@ -561,6 +563,8 @@ describe('Plugin', () => {
           pool.end(() => {
             agent.close({ ritmReset: false }).then(done)
           })
+
+          tracer._tracer.configure({ env: 'tester', sampler: { sampleRate: 1 } })
         })
 
         beforeEach(async () => {
@@ -597,7 +601,7 @@ describe('Plugin', () => {
 
           agent.assertSomeTraces(() => {
             expect(queryText).to.include('-00\'*/ SELECT 1 + 1 AS solution')
-          }).then(done, done).finally(() => tracer._tracer.configure({ env: 'tester', sampler: { sampleRate: 1 } }))
+          }).then(done, done)
 
           pool.query('SELECT 1 + 1 AS solution', () => {
             queryText = pool._allConnections[0]._protocol._queue[0].sql
--- a/packages/datadog-plugin-mysql2/test/index.spec.js
+++ b/packages/datadog-plugin-mysql2/test/index.spec.js
@@ -459,6 +459,8 @@ describe('Plugin', () => {
           connection.end(() => {
             agent.close({ ritmReset: false }).then(done)
           })
+
+          tracer._tracer.configure({ env: 'tester', sampler: { sampleRate: 1 } })
         })
 
         beforeEach(async () => {
@@ -499,7 +501,7 @@ describe('Plugin', () => {
             const spanId = traces[0][0].span_id.toString(16).padStart(16, '0')
 
             expect(queryText).to.include(`traceparent='00-${traceId}-${spanId}-00'*/ SELECT 1 + 1 AS solution`)
-          }).then(done, done).finally(() => tracer._tracer.configure({ env: 'tester', sampler: { sampleRate: 1 } }))
+          }).then(done, done)
 
           const connect = connection.query('SELECT 1 + 1 AS solution', () => {
             queryText = connect.sql
@@ -556,6 +558,8 @@ describe('Plugin', () => {
           pool.end(() => {
             agent.close({ ritmReset: false }).then(done)
           })
+
+          tracer._tracer.configure({ env: 'tester', sampler: { sampleRate: 1 } })
         })
 
         beforeEach(async () => {
@@ -596,7 +600,7 @@ describe('Plugin', () => {
             const spanId = traces[0][0].span_id.toString(16).padStart(16, '0')
 
             expect(queryText).to.include(`traceparent='00-${traceId}-${spanId}-00'*/ SELECT 1 + 1 AS solution`)
-          }).then(done, done).finally(() => tracer._tracer.configure({ env: 'tester', sampler: { sampleRate: 1 } }))
+          }).then(done, done)
 
           const queryPool = pool.query('SELECT 1 + 1 AS solution', () => {
             queryText = queryPool.sql


packages/dd-trace/src/opentracing/span.js
```diff
@@ -278,6 +278,10 @@ class DatadogSpan {
     this._processor.process(this)
   }
 
+  sample () {
```
bengl: This adds new API surface area so it would need:

(1) its own test
(2) a corresponding PR against dd-trace-api-js
(3) the appropriate changes in the dd-trace-api integration in this repo

na-ji: Thanks, I wasn't aware of the dd-trace-api process. 
_Maybe_ the [documentation](https://github.com/DataDog/dd-trace-js/blob/019b06661e0af058a276221ad1c36238302284f8/devdocs/adding-apis.md) on it should be linked in the CONTRIBUTING.md doc and added to the PR checklist. 

I opened the api PR here:
https://github.com/DataDog/dd-trace-api-js/pull/14

packages/datadog-plugin-pg/test/index.spec.js
```diff
@@ -685,7 +685,7 @@ describe('Plugin', () => {
 
             expect(queryText).to.equal(
               `/*dddb='postgres',dddbs='post',dde='tester',ddh='127.0.0.1',ddps='test',ddpv='${ddpv}',` +
-              `traceparent='00-${traceId}-${spanId}-00'*/ SELECT $1::text as message`)
+              `traceparent='00-${traceId}-${spanId}-01'*/ SELECT $1::text as message`)
```
crysmags, line 690: I see the expectation now is that all injected comments would have `01` meaning they are all sampled, is that correct? Do we expect all to be commented with `01` or is there a time we expect sampling to be `00`?

na-ji: During test, we do expect them all to be sampled. However, outside of tests, it all depends on the sampling decision. My understanding is that the main factor to the sampling decision is a rate limit. If we produce too many span too fast, a lot of them are going to not be sampled.

packages/dd-trace/src/opentracing/span.js
rochdev: I also think adding an API like that should be independent from this PR. There is already a mechanism in place to sample before propagation, so we could use the same mechanism instead of calling `sample` directly.

na-ji: @rochdev I'm sorry, but which mechanism do you have in mind? The reason I added this API is because I didn't find a way to do it from all the DB plugins.

packages/datadog-plugin-pg/test/index.spec.js
crysmags: I believe customers can also set their sampling. Can we test to the scenario that we expect the span to not be sampled and the appropriate decision is passed in to the comment?

packages/dd-trace/src/opentracing/span.js
rochdev: It's handled by the propagators, which this doesn't use, so it would have to be refactored to a propagator.

See: https://github.com/DataDog/dd-trace-js/blob/18f8a7825a0c17933e0b6822b2ac5f4af8877f20/packages/dd-trace/src/opentracing/tracer.js#L90

na-ji: I'm sorry but I don't see why a propagator would be a better fit in this case. I know that the tracing spec says we should make a sampling decision (if it hasn't happened yet) when propagating trace context to a downstream service. 
However, injecting the DBM comment isn't really "trace" propagation since these comments are not used to continue the trace (create more spans) downstream.
So the goal here is not to propagate the trace, but to tell the agent integration that is going to intercept the query sample/execution plan whether the trace was sampled or not. 
I could rename this API to be more explicit, like the what the [Java tracer did](https://github.com/DataDog/dd-trace-java/blob/d9df78f4c9294456af611bcd9323795d1c8694a3/internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentSpan.java#L183).

rochdev: A propagator is meant to inject the tracing context into a carrier without requiring knowledge or the underlying internals. For example, what if a customer wants to add this data manually? Would they need to know the exact format and how to build it themselves? What if we want to add some more data that is not available through the public API?

All that to say, I think a propagator would still be the right approach for a case like this. An unfortunate limitation in this case however is that the propagation API comes from OpenTracing and only includes a way to update objects, not strings. So unless the query is always on an object it makes this difficult to do without updating the API to support that.

If we were to not use propagators, then that's fine as well, but what we shouldn't do is add a public API based on an internal need. If we want to add a public `sample` method to the span that's fine, but it should be done as a separate PR that only does that. An easy way to do it right now would be to inject either a TraceState directly, or inject to a temporary object and write the TraceState from that object. This would trigger sampling without adding a new API.

na-ji: Following our discussion, I've deleted the public API in [`0e49587` (#6653)](https://github.com/DataDog/dd-trace-js/pull/6653/commits/0e49587f0777565dbd488f7808456148e44b55a5). I've also rebased on master. Thanks for taking the time to explain everything to me.

bengl (dismissed):

packages/datadog-plugin-pg/test/index.spec.js
```diff
@@ -685,7 +686,25 @@ describe('Plugin', () => {
...
+          })
+        })
+
+        it('query text should contain rejected sampling decision in the traceparent', async () => {
+          sinon.stub(tracer._tracer._prioritySampler, 'sample')
```
rochdev: Plugin tests are integration tests, mocking internals should be avoided whenever possible. Can the tracer be reconfigured with different rates instead?

na-ji: I tried at first but I couldn't manage to reconfigure the existing tracer instance with different sampling rates. My conclusion was that I would need to configure it on initialization, which would require a new test suite just to be able to a have a different tracer instance. 
So yes I agree with you, but I would like to see how.

packages/datadog-plugin-mysql/test/index.spec.js
```diff
@@ -491,13 +491,26 @@ describe('Plugin', () => {
...
           })
         })
 
+        it('query text should contain rejected sampling decision in the traceparent', done => {
+          tracer._tracer.configure({ env: 'tester', sampler: { sampleRate: 0 } })
+          let queryText = ''
+
+          agent.assertSomeTraces(traces => {
+            expect(queryText).to.include('-00\'*/ SELECT 1 + 1 AS solution')
+          }).then(done, done).finally(() => tracer._tracer.configure({ env: 'tester', sampler: { sampleRate: 1 } }))
```
rochdev: Resetting should be done in an `afterEach` block and not after an expectation. This is because expectations may never run or an error may be thrown beforehand and then the reset would never happen.

packages/datadog-plugin-mongodb-core/test/mongodb.spec.js
```diff
@@ -614,6 +614,55 @@ describe('Plugin', () => {
...
+
+          return agent.close({ ritmReset: false })
+        })
+
+        beforeEach(async () => {
+          client = await createClient()
+          db = client.db('test')
+          collection = db.collection(collectionName)
+
+          startSpy = sinon.spy(MongodbCorePlugin.prototype, 'start')
```
rochdev, line 637: This spy is not needed, only the behaviour needs to be tested.

rochdev: Never mind, I just saw that there are other occurrences in this file, so fixing this is out of scope of this PR then.

PR #6654, opened 2025-10-14
feat(otlp): Adds support for transforming and exporting metrics

mabdinur:
Adds support for sending metrics via the OTLP protocol.

### Motivation

Allow customers to send OpenTelemetry-compatible metrics from Datadog SDKs.

### Changes

**New Files:**
- `metrics/otlp_http_metric_exporter.js` - HTTP exporter for metrics (extends base class)
- `metrics/otlp_transformer.js` - Transforms OTel metrics to OTLP format
- `test/opentelemetry/metrics.spec.js` - 17 comprehensive tests

**Supported Metric Types:**
- Counter (monotonic)
- UpDownCounter (non-monotonic)
- Histogram (with buckets)
- Gauge (point-in-time values)

**Protocols:**
- `http/protobuf` (default, binary)
- `http/json` (for debugging)

### Testing

- 17 new tests covering all metric types
- Tests validate OTLP structure (resources, scopes, attributes)
- Tests verify both protobuf and JSON protocols

### Example Usage

```javascript
const exporter = new OtlpHttpMetricExporter(
  'http://localhost:4318/v1/metrics',
  '', // headers
  30000, // timeout
  'http/protobuf',
  resource
)

const metrics = [counter, histogram, gauge]
exporter.export(metrics, (result) => {
  console.log('Export result:', result.code)
})
```

N/A

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/opentelemetry/metrics/otlp_transformer.js
+++ b/packages/dd-trace/src/opentelemetry/metrics/otlp_transformer.js
@@ -2,21 +2,11 @@
 
 const OtlpTransformerBase = require('../otlp/otlp_transformer_base')
 const { getProtobufTypes } = require('../otlp/protobuf_loader')
+const { METRIC_TYPES, TEMPORALITY } = require('./constants')
 
-// Get the aggregation temporality enum
 const { protoAggregationTemporality } = getProtobufTypes()
 const AGGREGATION_TEMPORALITY_DELTA = protoAggregationTemporality.values.AGGREGATION_TEMPORALITY_DELTA
 const AGGREGATION_TEMPORALITY_CUMULATIVE = protoAggregationTemporality.values.AGGREGATION_TEMPORALITY_CUMULATIVE
-
-// Metric type constants
-const METRIC_TYPES = {
-  HISTOGRAM: 'histogram',
-  COUNTER: 'counter',
-  UPDOWNCOUNTER: 'updowncounter',
-  OBSERVABLECOUNTER: 'observable-counter',
-  OBSERVABLEUPDOWNCOUNTER: 'observable-updowncounter',
-  GAUGE: 'gauge'
-}
 
 /**
  * OtlpTransformer transforms metrics to OTLP format.
@@ -138,7 +128,7 @@ class OtlpTransformer extends OtlpTransformerBase {
       unit: metric.unit || ''
     }
 
-    const temporality = metric.temporality === 'CUMULATIVE'
+    const temporality = metric.temporality === TEMPORALITY.CUMULATIVE
       ? AGGREGATION_TEMPORALITY_CUMULATIVE
       : AGGREGATION_TEMPORALITY_DELTA
 
@@ -192,7 +182,7 @@ class OtlpTransformer extends OtlpTransformerBase {
       unit: metric.unit || ''
     }
 
-    const temporalityStr = metric.temporality === 'CUMULATIVE'
+    const temporalityStr = metric.temporality === TEMPORALITY.CUMULATIVE
       ? 'AGGREGATION_TEMPORALITY_CUMULATIVE'
       : 'AGGREGATION_TEMPORALITY_DELTA'
 
--- a/packages/dd-trace/src/opentelemetry/metrics/constants.js
+++ b/packages/dd-trace/src/opentelemetry/metrics/constants.js
@@ -1,0 +1,28 @@
+'use strict'
+
+// Metric type constants
+const METRIC_TYPES = {
+  HISTOGRAM: 'histogram',
+  COUNTER: 'counter',
+  UPDOWNCOUNTER: 'updowncounter',
+  OBSERVABLECOUNTER: 'observable-counter',
+  OBSERVABLEUPDOWNCOUNTER: 'observable-updowncounter',
+  GAUGE: 'gauge'
+}
+
+// Temporality constants
+const TEMPORALITY = {
+  DELTA: 'DELTA',
+  CUMULATIVE: 'CUMULATIVE',
+  GAUGE: 'GAUGE',
+  LOWMEMORY: 'LOWMEMORY'
+}
+
+// Default histogram bucket boundaries (in milliseconds for latency metrics)
+const DEFAULT_HISTOGRAM_BUCKETS = [0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10_000]
+
+module.exports = {
+  METRIC_TYPES,
+  TEMPORALITY,
+  DEFAULT_HISTOGRAM_BUCKETS
+}


packages/dd-trace/src/opentelemetry/metrics/otlp_transformer.js
```diff
@@ -0,0 +1,255 @@
...
+   * @private
+   */
+  #transformToProtobuf (metrics) {
+    const { protoMetricsService } = getProtobufTypes()
+
+    const metricsData = {
+      resourceMetrics: [{
+        resource: this._transformResource(),
+        scopeMetrics: this.#transformScope(metrics),
+        schemaUrl: ''
```
mabdinur: We will populate this field in a future PR. For now we just want to make sure the property can be encoded

packages/dd-trace/src/opentelemetry/metrics/otlp_transformer.js
```diff
@@ -0,0 +1,255 @@
...
+   * @param {boolean} isJson - Whether to format for JSON output
+   * @returns {Object[]} Array of scope metric objects
+   * @private
+   */
+  #transformScope (metrics, isJson = false) {
+    const groupedMetrics = this._groupByInstrumentationScope(metrics)
+    const scopeMetrics = []
+
+    for (const [key, metricsInScope] of groupedMetrics) {
+      const [name, version, schemaUrl] = key.split('@')
```
mabdinur: Opentelemetry doesn't allow the `@` character in meter names, metric names, or Attributes. So it's a useful delimiter here

packages/dd-trace/src/opentelemetry/metrics/otlp_transformer.js
```diff
@@ -0,0 +1,255 @@
...
+        dataPoints: metric.data.map(dp => ({
+          attributes: this._transformAttributes(dp.attributes),
+          startTimeUnixNano: dp.startTimeUnixNano,
+          timeUnixNano: dp.timeUnixNano,
+          count: dp.count,
+          sum: dp.sum,
+          bucketCounts: dp.bucketCounts || [],
+          explicitBounds: dp.explicitBounds || [],
+          exemplars: [],
+          flags: 0,
```
mabdinur: These are optional values. Here we are encoding the default values, we may use these fields the future but I can remove them to reduce the size of the payload

packages/dd-trace/src/opentelemetry/metrics/otlp_transformer.js
```diff
@@ -0,0 +1,255 @@
...
+   * @private
+   */
+  #transformMetric (metric) {
+    const result = {
+      name: metric.name,
+      description: metric.description || '',
+      unit: metric.unit || ''
+    }
+
+    if (metric.type === 'histogram') {
```
mabdinur: Should I use an enum here?

packages/dd-trace/src/opentelemetry/metrics/otlp_transformer.js
```diff
@@ -0,0 +1,255 @@
...
+      result.histogram = {
+        dataPoints: metric.data.map(dp => ({
+          attributes: this._attributesToJson(dp.attributes),
+          startTimeUnixNano: String(dp.startTimeUnixNano),
+          timeUnixNano: String(dp.timeUnixNano),
+          count: String(dp.count),
+          sum: dp.sum,
+          bucketCounts: dp.bucketCounts?.map(String) || [],
```
mabdinur:
```suggestion
          startTimeUnixNano: String(dp.startTimeUnixNano),
          timeUnixNano: String(dp.timeUnixNano),
          count: dp.count,
          sum: dp.sum,
          bucketCounts: dp.bucketCounts || [],
```

packages/dd-trace/src/opentelemetry/metrics/otlp_transformer.js
```diff
@@ -0,0 +1,255 @@
...
+      timeUnixNano: String(dataPoint.timeUnixNano)
+    }
+
+    if (dataPoint.startTimeUnixNano) {
+      result.startTimeUnixNano = String(dataPoint.startTimeUnixNano)
+    }
+
+    // Determine if value is int or double
+    if (Number.isInteger(dataPoint.value)) {
+      result.asInt = String(dataPoint.value)
```
mabdinur:
```suggestion
      result.asInt = dataPoint.value
```

packages/dd-trace/src/opentelemetry/otlp/otlp_http_exporter_base.js
```diff
@@ -10,8 +10,6 @@ const tracerMetrics = telemetryMetrics.manager.namespace('tracers')
 /**
  * Base class for OTLP HTTP exporters.
  *
- * This implementation follows the OTLP HTTP specification:
- * https://opentelemetry.io/docs/specs/otlp/#otlphttp
```
BridgeAR, line 14: Is that not the case anymore? :)

packages/dd-trace/src/opentelemetry/metrics/otlp_http_metric_exporter.js
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+/**
+ * OtlpHttpMetricExporter exports metrics via OTLP over HTTP.
+ *
+ * This implementation follows the OTLP HTTP v1.7.0 specification:
+ * https://opentelemetry.io/docs/specs/otlp/#otlphttp
+ *
+ * @class OtlpHttpMetricExporter
+ * @extends OtlpHttpExporterBase
```
BridgeAR: That is fine to add while it is a tad redundant (it should also be picked up automatically)

packages/dd-trace/src/opentelemetry/metrics/otlp_http_metric_exporter.js
```diff
@@ -0,0 +1,69 @@
...
+  constructor (url, headers, timeout, protocol, resource) {
+    super(url, headers, timeout, protocol, '/v1/metrics', 'metrics')
+    this.transformer = new OtlpTransformer(resource, protocol)
+  }
+
+  /**
+   * Exports metrics via OTLP over HTTP.
+   *
+   * @param {Array} metrics - Array of metric data to export
+   * @param {Function} resultCallback - Callback function for export result
```
BridgeAR, line 34: Could we define the return type here as well? :)

packages/dd-trace/src/opentelemetry/metrics/otlp_http_metric_exporter.js
```diff
@@ -0,0 +1,69 @@
...
+    let dataPointCount = 0
+    for (const metric of metrics) {
+      if (metric.data) {
+        dataPointCount += metric.data.length
+      }
+    }
+
+    // Record export attempt with tags
+    const telemetryTags = [...this._getTelemetryTags(), `points:${dataPointCount}`]
+    this._recordTelemetry('otel.metrics_export_attempts', 1, telemetryTags)
```
BridgeAR: Should these methods potentially be public to prevent using underscored methods? Or could we change this to not use inheritance and instead use composition?

packages/dd-trace/src/opentelemetry/metrics/otlp_transformer.js
```diff
@@ -0,0 +1,286 @@
...
+      const scope = {
+        name,
+        version,
+        droppedAttributesCount: 0
+      }
+
+      // Add attributes if present
+      if (attributes && Object.keys(attributes).length > 0) {
+        scope.attributes = isJson ? this._attributesToJson(attributes) : this._transformAttributes(attributes)
+      }
```
BridgeAR: Both of these methods are actually going to create an empty array for an empty object.
So the empty case has the same overhead when calling the method directly. When the object has entries, it would now create an additional array of all keys. To prevent this, we can just get the return type of the method and check if the return is empty or not. That way there is no additional overhead anymore.

```suggestion
      if (attributes) {
        const transformed = isJson ? this._attributesToJson(attributes) : this._transformAttributes(attributes)
        if (transformed.length) {
          scope.attributes = transformed
        }
      }
```

packages/dd-trace/src/opentelemetry/otlp/otlp_http_exporter_base.js
mabdinur: This was a mistake. I meant to remove this from the child class 🤦. I'll update the docs and reduce duplication.

packages/dd-trace/src/opentelemetry/metrics/otlp_http_metric_exporter.js
mabdinur: Removed

packages/dd-trace/src/opentelemetry/metrics/otlp_http_metric_exporter.js
mabdinur: I will make it public

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
```diff
@@ -45,8 +47,8 @@ class OtlpHttpLogExporter extends OtlpHttpExporterBase {
     }
 
     const payload = this.transformer.transformLogRecords(logRecords)
-    this._sendPayload(payload, resultCallback)
-    this._recordTelemetry('otel.log_records', logRecords.length)
+    this.sendPayload(payload, resultCallback)
```
mabdinur, line 50: Look into inverting the subclassing

PR #6656, opened 2025-10-14
[test optimization] Fix test suites being reported with an error message but passed test status

juan-fernandez:
### What does this PR do?

* Add a `ci:jest:test-suite:error` event that fires whenever `requireModuleOrMock` throws. Since at that time we know what the running test suite is, we can attach the error to the test suite. 
* Since this error can happen "off timing", that is, right _after_ the `adapter` function has returned, we need to wait to call  `.finish` on the running test suite span, hence the `process.nextTick`. 

### Motivation

`jest` fails if you try to import a file after a test has finished. This is a jest 30 new behavior introduced in https://github.com/jestjs/jest/pull/14110/. 

```javascript
// will throw with "can't import after env is torn down"
afterAll(() => {
  setTimeout(() => {
     require('./off-timing-import')
  }, 0)
})
it('can do stuff', () => {
   expect(true).toEqual(true)
})
```

```javascript
// will throw with "ReferenceError: You are trying to `import` a file outside of the scope of the test code."
afterAll(() => {
  setTimeout(() => {
     require('./off-timing-import')
  }, 0)
})
it('can do stuff', () => {
   // notice import is async but the test is sync
   import('./off-timing-import')
})
```

Before the changes in this PR, those errors were not caught at all by our instrumentation. This caused confusion in our customers.

### Plugin Checklist

- [x] Integration tests.

### DIFF
--- a/integration-tests/jest/jest.spec.js
+++ b/integration-tests/jest/jest.spec.js
@@ -725,9 +725,9 @@ describe('jest CommonJS', () => {
             'From ci-visibility/jest-bad-import-torn-down/jest-bad-import-test.js'
           )
           // This is the error message that jest should show. We check that we don't mess it up.
-          assert.include(failedTestSuite.content.meta[ERROR_MESSAGE], 'off-timing-import.js')
+          assert.include(failedTestSuite.content.meta[ERROR_MESSAGE], 'off-timing-import')
           assert.include(failedTestSuite.content.meta[ERROR_MESSAGE], 'afterAll')
-          assert.include(failedTestSuite.content.meta[ERROR_MESSAGE], 'process.nextTick')
+          assert.include(failedTestSuite.content.meta[ERROR_MESSAGE], 'nextTick')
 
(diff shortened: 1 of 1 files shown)


integration-tests/jest/jest.spec.js
```diff
@@ -71,7 +71,7 @@ const expectedCoverageFiles = [
   'ci-visibility/test/ci-visibility-test.js',
   'ci-visibility/test/ci-visibility-test-2.js'
 ]
-const runTestsWithCoverageCommand = 'node ./ci-visibility/run-jest.js'
+const runTestsCommand = 'node ./ci-visibility/run-jest.js'
```
juan-fernandez, line 74: ℹ️ unrelated improvement that should make tests faster

juan-fernandez: not the rename ofc 😄 but what it means: we run without code coverage by default now

daniel-mohedano (approved):

PR #6667, opened 2025-10-15
Increase timeout of serverless flaky tests

juan-fernandez:
### What does this PR do?
Increase timeout of flaky tests

### Motivation
Tests are flaking for every PR. 

I realise this does not fix the underlying issue, but if it improves flakiness, we can do it as an interim solution. Otherwise, every team is blocked

⚠️ This should be reverted ASAP, once the root cause is found and fixed

### DIFF
--- a/.github/workflows/serverless.yml
+++ b/.github/workflows/serverless.yml
@@ -87,8 +87,7 @@ jobs:
         uses: ./.github/actions/testagent/logs
         with:
           suffix: plugins-${{ github.job }}-${{ matrix.node-version }}
-      # TODO: uncomment once flakiness is better
-      # - uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
+      - uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
 
   azure-functions:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/serverless.yml
```diff
@@ -81,12 +81,14 @@ jobs:
...
-      - run: yarn test:plugins:ci
+      # TODO: Remove :flaky once flakiness is better
+      - run: yarn test:plugins:ci:flaky
       - if: always()
         uses: ./.github/actions/testagent/logs
         with:
           suffix: plugins-${{ github.job }}-${{ matrix.node-version }}
-      - uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
+      # TODO: uncomment once flakiness is better
+      # - uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
```
bengl: Wait. Codecov itself is flaky?

packages/datadog-plugin-aws-sdk/test/lambda.spec.js
```diff
@@ -15,6 +15,7 @@ const createClientContext = data => Buffer.from(JSON.stringify(data)).toString('
 
 describe('Plugin', () => {
   describe('aws-sdk (lambda)', function () {
+    this.timeout(10000)
```
pabloerhard, line 18: We might want to also add the timeout into the agent.assertSomeTraces

bengl: You mean have it assume the mocha timeout? Seems like a good idea. Let's do that in a future PR.

.github/workflows/serverless.yml
juan-fernandez: `nyc` slows down the execution, making things worse. Codecov itself has nothing to do with the problem

packages/datadog-plugin-aws-sdk/test/lambda.spec.js
pabloerhard: Yes, from what I was able to get some tests were failing because the default timeout for agent.assertSomeTraces is 1000ms so even though the moca timeout is set to 10000 the agent.assert will still "timeout"

juan-fernandez: yes, that's a big issue in our test suite: increasing the timeout of the test does nothing unless the corresponding "agent#wait_for_traces" sees its timeout increase too

PR #6685, opened 2025-10-17
add support for multiple retries to flakiness script

rochdev:
### What does this PR do?

Add support for multiple retries to flakiness script.

### Motivation

Otherwise some flakiness is hidden and it's more difficult to debug when rerunning several times.

### DIFF
--- a/integration-tests/cypress/cypress.spec.js
+++ b/integration-tests/cypress/cypress.spec.js
@@ -2414,18 +2414,25 @@ moduleTypes.forEach(({
           ...restEnvVars
         } = getCiVisEvpProxyConfig(receiver.port)
 
+        const specToRun = 'cypress/e2e/spec.cy.js'
+
         childProcess = exec(
-          testCommand,
+          version === 'latest' ? testCommand : `${testCommand} --spec ${specToRun}`,
           {
             cwd,
             env: {
               ...restEnvVars,
               CYPRESS_BASE_URL: `http://localhost:${webAppPort}`,
-              DD_TEST_SESSION_NAME: 'my-test-session-name'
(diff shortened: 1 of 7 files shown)


scripts/flakiness.mjs
```diff
@@ -89,7 +92,10 @@ async function checkWorkflowJobs (id, page = 1) {
   const { jobs } = response.data
 
   // We've reached the last page and there are no more results.
-  if (jobs.length === 0) return
+  if (jobs.length === 0) {
+    // Check previous attempt to include successive failures.
+    return checkWorkflowJobs(id, attempt - 1, page + 1)
```
watson: I think we should start again from page 1 when checking a different attempt. The requested page might not exist in a different run. Each attempt only contains the jobs objects that were executed in that particular attempt.

```suggestion
    return checkWorkflowJobs(id, attempt - 1)
```

rochdev: Fixed, and I also removed the need for going to the next page when the number of results indicates this is not needed.

scripts/flakiness.mjs
```diff
@@ -68,17 +69,20 @@ async function checkWorkflowRuns (id, page = 1) {
...
   }
 
   promises.push(checkWorkflowRuns(id, page + 1))
 
   return Promise.all(promises)
 }
 
-async function checkWorkflowJobs (id, page = 1) {
+async function checkWorkflowJobs (id, attempt, page = 1) {
+  console.log(id, attempt, page)
```
watson: Did you mean to add this in your last commit?

rochdev: Nope, removed!

PR #6686, opened 2025-10-17
fix(amqplib): prevent crash when basic.get returns no message

khanayan123:
### What does this PR do?
Adds a guard in the amqplib consumer plugin to handle cases where basic.get is called on an empty queue, preventing crashes when fields is undefined

### Motivation
Regression in v5.54.0: when basic.get returns no message (queue empty), amqplib returns false causing fields to be undefined. PR [#5734](https://github.com/DataDog/dd-trace-js/pull/5734/files) removed async storage and try/catch error handling, exposing this as an uncaught exception. Customer reported crashes in production.


packages/datadog-plugin-amqplib/src/consumer.js
```diff
@@ -14,6 +14,8 @@ class AmqplibConsumerPlugin extends ConsumerPlugin {
 
     if (method !== 'basic.deliver' && method !== 'basic.get') return
 
+    if (!fields || !message) return
```
BridgeAR: If I am not mistaken, this prevents the call from being tracked in the first place.

I think instead, we should just be able to handle calls that have no message. Otherwise we loose actual information for the user.

PR #6691, opened 2025-10-17
fix race condition in kinesis tests by swallowing irrelevant errors

rochdev:
### What does this PR do?

Fix race condition in kinesis tests by swallowing irrelevant errors.

### Motivation

Errors from the underlying connection to Kinesis are irrelevant for any tracing test, and they are likely since the stream is destroyed in the `afterEach` hook which could happen before the connection is closed once the test is done.


rochdev:
@BridgeAR Any other concerns? This is a quick fix that resolves the issue entirely.

PR #6693, opened 2025-10-17
switch to bun for integration tests using the sandbox

rochdev:
### What does this PR do?

Switch to bun for integration tests using the sandbox.

### Motivation

Bun is 5-25x faster than npm or yarn at anything.

### Additional Notes

The change is made in a way that doesn't require having Bun preinstalled so that the user doesn't need to know about its existence. It also paves the way for abstracting the package manager used in test from the one used by the user.

@BridgeAR I have validated that the output of `npm pack` and `bun pack` are the same as discussed offline.

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -22,7 +22,7 @@ runs:
     - run: bun install --linker=hoisted --trust || sleep 60 && bun install --linker=hoisted --trust
       shell: bash
       env:
-        DD_IGNORE_ENGINES: 'true'
+        _DD_IGNORE_ENGINES: 'true'
     - run: tar -cf node_modules.tar node_modules
       shell: bash
       if: inputs.cache == 'true' && steps.yarn-cache.outputs.cache-hit != 'true'
--- a/integration-tests/helpers/bun.js
+++ b/integration-tests/helpers/bun.js
@@ -7,7 +7,7 @@ const BUN_INSTALL = join(PROJECT_ROOT, '.bun')
 const BUN = join(PROJECT_ROOT, 'node_modules', '.bin', 'bun')
 
 function withBun (env = process.env) {
-  return { ...env, BUN_INSTALL, DD_IGNORE_ENGINES: true }
+  return { ...env, BUN_INSTALL, _DD_IGNORE_ENGINES: true }
 }
 
 module.exports = { BUN, withBun }
--- a/scripts/preinstall.js
+++ b/scripts/preinstall.js
@@ -19,7 +19,7 @@ const min = Number(requirePackageJson(path.join(__dirname, '..')).engines.node.m
 
 // Most package managers don't support `npm_config_argv`, so we need a custom
 // flag to allow installing dd-trace on unsupported engines.
-const hasIgnoreEngines = process.env.DD_IGNORE_ENGINES === 'true' || npmArgv &&
+const hasIgnoreEngines = process.env._DD_IGNORE_ENGINES === 'true' || npmArgv &&
   npmArgv.original &&
   npmArgv.original.includes('--ignore-engines')
 


bengl (changes requested):
Need to see a test verifying equivalence of both pack algorithms.

rochdev:
> Need to see a test verifying equivalence of both pack algorithms.

@bengl Added a test, please re-review.

.github/actions/node/action.yml
```diff
@@ -13,3 +13,6 @@ runs:
         node-version: ${{ inputs.version == 'eol' && '16' || inputs.version == 'oldest' && '18' || inputs.version == 'maintenance' && '20' || inputs.version == 'active' && '22' || inputs.version == 'latest' && '24' || inputs.version }}
         check-latest: true
         registry-url: ${{ inputs.registry-url || 'https://registry.npmjs.org' }}
+    - uses: oven-sh/setup-bun@v2
```
watson: This should be pinned to a hash for security purposes

.github/workflows/platform.yml
```diff
@@ -17,6 +17,16 @@ env:
 # TODO: upstream jobs
 
 jobs:
+  bun-pack:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: ./.github/actions/node/active-lts
+      - run: mkdir -p npm bun
```
watson:
```suggestion
      - run: mkdir npm bun
```

.github/workflows/platform.yml
```diff
@@ -17,6 +17,16 @@ env:
...
+      - run: FILENAME=$(npm pack) && tar -zxf $FILENAME -C npm
```
watson: Nit: Avoid leaking `FIILENAME` 

```suggestion
      - run: FILENAME=$(npm pack) tar -zxf $FILENAME -C npm
```

But ultimately, why not just this?

```suggestion
      - run: tar -zxf $(npm pack) -C npm
```

.github/workflows/platform.yml
```diff
@@ -17,6 +17,16 @@ env:
...
+      - run: bun pm pack --filename bun.tgz && tar -zxf bun.tgz -C bun
+      - run: diff -r npm bun
```
watson, line 29: This will diff the filenames + their content. Do we really need to also diff the content? We could do this whole operation faster if we just diffed the filenames contained in the two tarballs 🤔 

Then the whole job could just be:

```yaml
- run: NPM_FILENAME=$(npm pack) && bun pm pack --filename bun.tgz && diff <(tar -tf $NPM_FILENAME | sort) <(tar -tf bun.tgz | sort)
```

integration-tests/helpers/bun.js
```diff
@@ -0,0 +1,47 @@
...
+const BUN_BIN = `${BUN_INSTALL}/bin`
+
+let hasCompatibleBun = env => {
+  try {
+    const version = execSync('command -v bun >/dev/null 2>&1 && bun -v', { env }).toString()
+
+    if (!semifies(version, `^${BUN_VERSION}`)) {
+      return false
+    }
+  } catch (e) {
```
watson: I guess our linting rule for this, doesn't cover the `integration-tests` dir

```suggestion
  } catch {
```

integration-tests/helpers/bun.js
```diff
@@ -0,0 +1,47 @@
...
+function withBun (env = process.env) {
+  env = {
+    ...env,
+    PATH: env.PATH ? `${BUN_BIN}:${env.PATH}` : BUN_BIN,
+    BUN_INSTALL
+  }
+
+  if (hasCompatibleBun(env)) return env
+
+  execSync(`curl -fsSL https://bun.com/install | bash -s "bun-v${BUN_VERSION}"`, {
```
watson: Hmm not sure I like this just installing a system wide dependency without confirmation. It's a security concern similar to post-install scripts. We already have prerequisites for other things to be installed on your system when developing, so why not add bun to this list? However, if we want to improve the developer experience, could we either:

1. Bundle the bin in the repo, so it's local to the repo and not system-wide
2. Add a conformation question, asking if users really want to run this install

integration-tests/helpers/index.js
```diff
@@ -271,15 +272,18 @@ async function createSandbox (dependencies = [], isGitRepo = false,
...
   const out = path.join(folder, `dd-trace-${version}.tgz`)
-  const allDependencies = [`file:${out}`].concat(cappedDependencies)
+  const deps = cappedDependencies.join(' ')
 
   await fs.mkdir(folder)
   const preferOfflineFlag = process.env.OFFLINE === '1' || process.env.OFFLINE === 'true' ? ' --prefer-offline' : ''
-  const addCommand = `yarn add ${allDependencies.join(' ')} --ignore-engines${preferOfflineFlag}`
   const addOptions = { cwd: folder, env: restOfEnv }
-  execHelper(`npm pack --silent --pack-destination ${folder}`, { env: restOfEnv }) // TODO: cache this
+  execHelper(`bun pm pack --silent --destination ${folder}`, { env: restOfEnv }) // TODO: cache this
```
watson: I haven't checked if `--silent` is an alias, but the docs state that the correct flag is `--quiet` and it doesn't mention `--silent`. Also added `--gzip-level 0` to speed up packing time:

```suggestion
  execHelper(`bun pm pack --quiet --gzip-level 0 --destination ${folder}`, { env: restOfEnv }) // TODO: cache this
```

.github/workflows/platform.yml
```diff
@@ -17,6 +17,16 @@ env:
...
 
 jobs:
+  bun-pack:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: ./.github/actions/node/active-lts
+      - run: mkdir -p npm bun
+      - run: FILENAME=$(npm pack) && tar -zxf $FILENAME -C npm
+      - run: bun pm pack --filename bun.tgz && tar -zxf bun.tgz -C bun
```
watson: If you want to speed things up, you could do:

```suggestion
      - run: bun pm pack --filename bun.tgz --gzip-level 0 && tar -zxf bun.tgz -C bun
```

integration-tests/helpers/index.js
```diff
@@ -271,15 +272,18 @@ async function createSandbox (dependencies = [], isGitRepo = false,
   }
   const folder = path.join(os.tmpdir(), id().toString())
   const out = path.join(folder, `dd-trace-${version}.tgz`)
-  const allDependencies = [`file:${out}`].concat(cappedDependencies)
+  const deps = cappedDependencies.join(' ')
 
   await fs.mkdir(folder)
   const preferOfflineFlag = process.env.OFFLINE === '1' || process.env.OFFLINE === 'true' ? ' --prefer-offline' : ''
```
watson: The `--prefer-offline` flag from `yarn` isn't supported by `bun`. This is here to support running tests without an internet connection (e.g. while in an airplane). Without it, `yarn` will refuse to even check the local cache if there's no internet connection and will just fail the install.

I unfortunately don't think this can be supported by `bun` 😞

.github/workflows/platform.yml
bengl: I suggest running this in exactly the same manner as we run it elsewhere, to ensure test fidelity.

.github/workflows/platform.yml
bengl: We don't necessarily need to _diff_ the content, but we should at least ensure that they're the same. Maybe a quick hash would be better?

integration-tests/helpers/bun.js
bengl: Yes. Please do not curl-pipe-to-bash in test runner scripts.

I'd prefer an abstraction that uses bun (if a compatible version is available) and then uses npm if bun is not available.

.github/workflows/platform.yml
rochdev: This is actually pretty fast and validates the contents of both folders are the same. Not sure why something else would be better?

integration-tests/helpers/bun.js
bengl: @watson some SSI tests run on very old versions of Node.js (outside our supported range) that don't support an error-less catch.

integration-tests/helpers/bun.js
rochdev: It's not installing a system wide dependency. It installs bun locally in `dd-trace/.bun`. Then the tests run using `PATH="dd-trace/.bun/bin:PATH"` which makes `bun` available as a command.

integration-tests/helpers/index.js
rochdev: > I haven't checked if --silent is an alias, but the docs state that the correct flag is --quiet and it doesn't mention --silent.

Both work, but I guess using the documented flag is better 👍 

> Also added --gzip-level 0 to speed up packing time

I'm amazed that it can be even faster than it already was.

.github/workflows/platform.yml
rochdev: I will add it in both places 👍

integration-tests/helpers/index.js
rochdev: It should be supported. If it's not, there may be something I need to change or there might be a bug.

bengl:
> Installing in this way was the only way I found to install locally only if not present

Right, I'm suggesting to _not_ install locally if not present, and instead instruct devs to install bun if they want this particular speedup.

> and also use the same version in CI

Is that critical? We don't necessarily use the same version of yarn or npm in CI as what we have installed locally.

.github/workflows/platform.yml
rochdev: Taking a look at the job run `diff` takes less than 1 second to run so I think it's fine. Does it do more than we need? Maybe or maybe not, but it's fast enough that it shouldn't matter.

integration-tests/helpers/bun.js
rochdev: Switched to using the npm version of bun, but it was too slow to install so I also switched CI to install our dependencies with bun. Bun installing bun is fast.

rochdev:
@bengl @watson I think I addressed all of your concerns, please take another look.

.github/workflows/test-optimization.yml
```diff
@@ -57,7 +57,7 @@ jobs:
     name: integration-playwright (${{ matrix.playwright-version}}, node-${{ matrix.node-version }})
     runs-on: ubuntu-latest
     container:
-      image: rochdev/playwright-tools@sha256:4903e8078c956b04d3190973732de99eea4e0fd5074188e007c8f6c7af066504 # 1.54.1-2
+      image: rochdev/playwright-tools@sha256:99a8e8df71996aa7797f83e98df50bc550c7c6eaf0b701f19fa4f928f2cf93b2 # 1.54.1-3
```
bengl, line 60: Non-blocker: We need to stop depending on personal images.

.github/workflows/aiguard.yml
```diff
@@ -68,10 +68,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
-      - uses: ./.github/actions/install
       - uses: ./.github/actions/node
         with:
           version: ${{ matrix.version }}
+      - uses: ./.github/actions/install
```
bengl, line 74: Why was the order here changed?

integration-tests/helpers/bun.js
```diff
@@ -0,0 +1,13 @@
+'use strict'
+
+const { join, resolve } = require('path')
+
+const PROJECT_ROOT = resolve(__dirname, '..', '..')
+const BUN_INSTALL = join(PROJECT_ROOT, '.bun')
+const BUN = join(PROJECT_ROOT, 'node_modules', '.bin', 'bun')
```
bengl, line 7: Relying on things being in particular places inside `node_modules` is what gets us into lots of pain. Is there any way we can get this programmatically? For example, with `npx -c 'which bun'` (or the bun equivalent).

integration-tests/helpers/index.js
```diff
@@ -271,15 +272,18 @@ async function createSandbox (dependencies = [], isGitRepo = false,
...
   await fs.mkdir(folder)
-  const preferOfflineFlag = process.env.OFFLINE === '1' || process.env.OFFLINE === 'true' ? ' --prefer-offline' : ''
-  const addCommand = `yarn add ${allDependencies.join(' ')} --ignore-engines${preferOfflineFlag}`
   const addOptions = { cwd: folder, env: restOfEnv }
-  execHelper(`npm pack --silent --pack-destination ${folder}`, { env: restOfEnv }) // TODO: cache this
+  const addFlags = ['--linker=hoisted', '--trust']
+  execHelper(`${BUN} pm pack --quiet --gzip-level 0 --destination ${folder}`, { env: restOfEnv })
 
-  execHelper(addCommand, addOptions)
+  if (process.env.OFFLINE === '1' || process.env.OFFLINE === 'true') {
```
bengl, line 284: Is this documented (or automatically inserted) elsewhere?

.github/workflows/aiguard.yml
rochdev: Because `actions/install` now uses `bun` which is added by `actions/node`.

integration-tests/helpers/bun.js
rochdev: > Is there any way we can get this programmatically? For example, with npx -c 'which bun' (or the bun equivalent).

`npx` is slow, and the `bun` equivalent would have the same discovery issue.

> Relying on things being in particular places inside node_modules is what gets us into lots of pain.

These are 2 different concerns. One is relying on a specific folder structure and the other is just being explicit about a binary location. I don't see why being explicit would be bad here.

integration-tests/helpers/index.js
rochdev: This was moved from above and not added by me.

integration-tests/helpers/bun.js
rochdev: I guess we could look for `bunx` (which would be available in CI) and if it's not available then fallback to `npx` maybe? 🤔 Not sure I prefer that to being explicit, as I think binary locations are pretty standard (although that's more of a gut feeling than something I researched). I wonder if that would actually work with all package managers too if that's the goal. At that point I think I'd just switch back to using the script.

integration-tests/helpers/index.js
rochdev: This seems to work for me with Bun. Can you provide a reproduction where this doesn't work?

rochdev: Actually it seems like Bun default to always prefer offline, so I think it should also work always even without the flag?

integration-tests/helpers/bun.js
bengl: If a package manager can assume that it will always be used to run scripts, then it can put them wherever, really.

That said, that's not really likely, so fine with leaving this as is, and fixing it later if there's a problem.

.github/workflows/platform.yml
watson: Did you also take the time it takes to extract into account? Anyway, this is not a blocker for me, I just didn't like doing things that technically wasn't necessary

integration-tests/helpers/index.js
watson: I can't find the flag documented in the bun docs, so maybe you're right that it just does it by default. If I'm right, and the flag isn't supported (no matter if it's doing it by default or not), we should clean this up and remove the code related to this. Otherwise it will quickly get confusing when reading the source code later.

integration-tests/helpers/index.js
watson: As I wrote in https://github.com/DataDog/dd-trace-js/pull/6693#discussion_r2450466296 I don't think this flag is supported by bun

.github/workflows/platform.yml
rochdev: That's also pretty quick. In any case, this is only a test and it runs in seconds so not sure I understand what the issue is even if it took a minute.

yarn.lock
watson, line 1: There seem to be a lot of changes in this `yarn.lock` file unrelated to the added bun dependency 🤔

integration-tests/helpers/index.js
rochdev: There are quite a few things not documented unfortunately 😅 But I found issues referencing it as a thing. I don't know your use case so it's a bit difficult for me to validate fully, but from a quick test of killing my network connection it was installing just fine. Before removing the flag I'd prefer if you can quickly test it since I believe you added it to begin with.

.github/actions/install/action.yml
```diff
@@ -19,8 +19,10 @@ runs:
       if: inputs.cache == 'true' && steps.yarn-cache.outputs.cache-hit == 'true'
       # Retry in case of server error from registry.
       # Wait 60 seconds to give the registry server time to heal.
-    - run: yarn install --frozen-lockfile --ignore-engines || sleep 60 && yarn install --frozen-lockfile --ignore-engines
+    - run: bun install --linker=hoisted --trust || sleep 60 && bun install --linker=hoisted --trust
```
watson, line 22: The default linker strategy is already `hoisted` as far as I know, so there shouldn't be a need to set it separately?

integration-tests/helpers/index.js
rochdev: I did some further testing and it is supported and works but only with the flag similar to Yarn.

.github/actions/install/action.yml
rochdev: Bun 1.3 switched to `isolated` as the default, so I prefer to set it explicitly so we don't forget when changing version.

yarn.lock
rochdev: Our current `yarn.lock` is broken. Yarn ignores the problem, but if you add a dependency for example you'll see it no longer works, and Bun doesn't ignore the problem. This is true on the main branch as well, we could fix it there too but I figured once this gets merged it will fix the issue as well.

integration-tests/helpers/index.js
watson: Weird... I tested it and you're right. It seems to work offline without the `--prefer-offline` flag. I looked in the source code and some places it says that [it should just work offline by default](https://github.com/oven-sh/bun/blob/89fa0f343945e61d5e4a0077cc7e93a802ed56e7/bench/install/README.md#with-lockfile-offline-mode) with no flag, but other places [the flag is mentioned](https://github.com/oven-sh/bun/blob/89fa0f343945e61d5e4a0077cc7e93a802ed56e7/completions/bun.zsh#L486). So let's just keep it I guess 🤷

yarn.lock
watson: Please explain what you mean by broken?

After switching to your branch locally and running `yarn`, I got these warnings in the output:
image

If I then, immediately after ran `./node_modules/.bin/bun install`, I got this output:
image

It seems to rather be a difference in how `yarn` and `bun` interprets and writes the lock file? So if you switch back and forth from running `yarn` and `bun` they seem to be fighting for the format of the `yarn.lock` file which will become a problem really quick.

After this PR lands, is our recommendation that all devs should use `bun` instead of `yarn` locally?

rochdev: Bun and Yarn are not "fighting" as Bun doesn't ever touch a lockfile it doesn't own. The lockfile in this PR was generated by Yarn. If this is so contentious I can open a separate PR with only the new lockfile instead.

As for why the lockfile is broken to begin with, my only explanation is that it was edited manually.

rochdev:
> Based on my discovery in https://github.com/DataDog/dd-trace-js/pull/6693#discussion_r2450651810, I'll block this PR for the time being, until it can get resolved

The lockfile was generated by Yarn, so I don't see why it would be a blocker for Bun.

rochdev:
@bengl Anything else on your end before this can be approved?

yarn.lock
watson: > Bun and Yarn are not "fighting" as Bun doesn't ever touch a lockfile it doesn't own. The lockfile in this PR was generated by Yarn.

Maybe I'm just confused as to what's going on here. If Bun doesn't use the `yarn.lock` file, what lock file is it using?

> If this is so contentious I can open a separate PR with only the new lockfile instead.

If it's unrelated to this PR, I always prefer to have it as a separate PR. But maybe I don't understand why you bundled it into this PR?

> As for why the lockfile is broken to begin with, my only explanation is that it was edited manually.

Sorry, I still don't see any evidence that it's broken? What symptoms do you see? For me it has worked just fine until now

integration-tests/helpers/index.js
```diff
@@ -11,9 +11,10 @@ const path = require('path')
 const assert = require('assert')
 const FakeAgent = require('./fake-agent')
 const id = require('../../packages/dd-trace/src/id')
-const { version } = require('../../package.json')
 const { getCappedRange } = require('../../packages/dd-trace/test/plugins/versions')
+const { BUN, withBun } = require('./bun')
 
+const sandboxRoot = path.join(os.tmpdir(), id().toString())
```
watson, line 17: Not that it matters from a technical perspective, but did you mean to add the id twice? (you're also adding it again further down in this file when using `sandboxRoot` to create the `folder` string)

integration-tests/helpers/index.js
```diff
@@ -269,17 +270,22 @@ async function createSandbox (dependencies = [], isGitRepo = false,
...
-  const allDependencies = [`file:${out}`].concat(cappedDependencies)
+  const folder = path.join(sandboxRoot, id().toString())
+  const out = path.join(sandboxRoot, 'dd-trace.tgz')
+  const deps = cappedDependencies.concat(`file:${out}`)
 
-  await fs.mkdir(folder)
-  const preferOfflineFlag = process.env.OFFLINE === '1' || process.env.OFFLINE === 'true' ? ' --prefer-offline' : ''
-  const addCommand = `yarn add ${allDependencies.join(' ')} --ignore-engines${preferOfflineFlag}`
+  await fs.mkdir(folder, { recursive: true })
   const addOptions = { cwd: folder, env: restOfEnv }
```
watson, line 278: Tiny nit: This lines confuses me 😅 I know this nit is unrelated to this PR, but if you're changing other things, feel free to either just inline this where it's used below, or at least move the declaration of the constant down above where it's used. For longer than I'd like to admit I thought it was used in the `bun pm pack` command 🫣 Feel free to move the `addFlags` declaration as well 😊

But feel free to ignore this as it's just a cleanup of unrelated code.

watson (changes requested):

watson (approved):

yarn.lock
rochdev: Already discussed these offline but for future reference:

> Maybe I'm just confused as to what's going on here. If Bun doesn't use the yarn.lock file, what lock file is it using?

Bun reads the lockfile but it doesn't write it as it uses its own instead. So the changes in the file were generated by Yarn and not Bun.

> If it's unrelated to this PR, I always prefer to have it as a separate PR. But maybe I don't understand why you bundled it into this PR?

I needed to add the `bun` dependency, but that made unrelated changes to the lockfile that ended up failing the tests. Basically, Yarn tried to fix parts of the broken file at the same time but ended up making things worse. The only way to completely fix everything was to regenerate the lockfile from scratch. I also tried to just remove all the unrelated changes manually but then Bun was unable to use the file in its current state.

> Sorry, I still don't see any evidence that it's broken? What symptoms do you see? For me it has worked just fine until now

Maybe "broken" was a strong word, let's call it internally inconsistent. It was usable by Yarn but not by Bun, and even Yarn was having a hard time with it when adding a new dependency.

PR #6694, opened 2025-10-17
Support all configurations with Declarative Config

mtoffl01:
### What does this PR do?
Previously, only product-enablement configs were supported for Declarative Config (application_monitoring.yaml files). This PR adds support for _all_ env-var configs. Now, all 3 sources -- local file, env vars, and fleet file -- go through the same core logic in new function _applyConfigValues.

The following [smaller] changes were also made as a result of necessary refactoring:
1. Moves processing logic for Otel & DD propagators out of Config constructor and into _applyConfigValues
2. Moves processing logic for DD_API_KEY, DD_APP_KEY, DD_INSTRUMENTATION_INSTALL_* and DD_TRACE_CLOUD_* settings out of constructor and into _applyConfigValues (See test `should support legacy direct-set fields through all stableconfig and env var sources`)
3. Moves logic for determining defaults based on environment (e.g, serverless) into Config constructor

### Motivation
[RFC](https://docs.google.com/document/d/1VXMxdMzrm7HpvVqJENo0u9PNN9qEe-UVioqAMfVJkOI/edit?usp=sharing)
Declarative Config allows APM configuration to be managed at the host level through a YAML-formatted file. This approach is used in the fleet automation workflow, where users can apply configurations to their SDKs via the Fleet UI, or manage them directly through YAML files. Expanding support to all environment-variable-based settings bolsters feature parity and allows greater flexibility for users who prefer or rely on declarative configuration.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/config.js
+++ b/packages/dd-trace/src/config.js
@@ -949,7 +949,6 @@ class Config {
     this.#setBoolean(opts, 'clientIpEnabled', options.clientIpEnabled)
     this.#setString(opts, 'clientIpHeader', options.clientIpHeader?.toLowerCase())
     if (options.cloudPayloadTagging?.request || options.cloudPayloadTagging?.response) {
-      // Auto-enable based on presence of request/response rules
       if (options.cloudPayloadTagging.request) {
         this.#setBoolean(opts, 'cloudPayloadTagging.requestsEnabled', true)
       }
@@ -961,7 +960,6 @@ class Config {
         splitJSONPathRules(options.cloudPayloadTagging.response)
       )
     }
-    // Allow explicit override if provided
     if (options.cloudPayloadTagging?.requestsEnabled !== undefined) {
       this.#setBoolean(opts, 'cloudPayloadTagging.requestsEnabled', options.cloudPayloadTagging.requestsEnabled)
     }


packages/dd-trace/src/config.js
```diff
@@ -442,48 +371,45 @@ class Config {
   // for _merge to work, every config value must have a default value
   _applyDefaults () {
     setHiddenProperty(this, '_defaults', defaults)
+
+    // Runtime-dependent defaults
```
mtoffl01: Fixes: https://github.com/DataDog/dd-trace-js/pull/4261#discussion_r1580513637

packages/dd-trace/src/config-helper.js
```diff
@@ -78,6 +78,7 @@ module.exports = {
    * @returns {string|undefined}
    * @throws {Error} if the configuration is not supported
    */
+  // This method, and callers of this method, need to be updated to check for declarative config sources as well.
   getEnvironmentVariable (name) {
```
mtoffl01, line 84: getEnvironmentVariable is called 150+ times across the codebase. I'm thinking the changes described on line 81 should be done in a separate PR.

packages/dd-trace/src/config_defaults.js
```diff
@@ -19,6 +20,15 @@ const service = getEnv('AWS_LAMBDA_FUNCTION_NAME') ||
...
 module.exports = {
+  apiKey: undefined,
+  appKey: undefined,
+  'installSignature.id': null,
+  'installSignature.time': null,
+  'installSignature.type': null,
+  'cloudPayloadTagging.requestsEnabled': false,
+  'cloudPayloadTagging.responsesEnabled': false,
+  'cloudPayloadTagging.maxDepth': 10,
+  'cloudPayloadTagging.rules': [],
```
BridgeAR: Would you mind adding them in sorted order? They are currently sorted as far as I see it.

packages/dd-trace/src/config.js
```diff
@@ -645,262 +555,324 @@ class Config {
...
 
-    this.#setString(env, 'protocolVersion', DD_TRACE_AGENT_PROTOCOL_VERSION)
-    this.#setString(env, 'queryStringObfuscation', DD_TRACE_OBFUSCATION_QUERY_STRING_REGEXP)
-    this.#setBoolean(env, 'remoteConfig.enabled', DD_REMOTE_CONFIGURATION_ENABLED ?? !this._isInServerlessEnvironment())
-    env['remoteConfig.pollInterval'] = maybeFloat(DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS)
-    this.#envUnprocessed['remoteConfig.pollInterval'] = DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS
-    this.#setBoolean(env, 'reportHostname', DD_TRACE_REPORT_HOSTNAME)
+    this.#setString(target, 'protocolVersion', DD_TRACE_AGENT_PROTOCOL_VERSION)
+    this.#setString(target, 'queryStringObfuscation', DD_TRACE_OBFUSCATION_QUERY_STRING_REGEXP)
+    this.#setBoolean(target, 'remoteConfig.enabled', DD_REMOTE_CONFIGURATION_ENABLED)
```
BridgeAR, line 784: This change makes sense as such, that we expect a user never to define this value if in a serverless environment. If they do however use a single configuration for serverless and non-serverless and they have this set to enabled, it would cause a regression, if I am not mistaken.

packages/dd-trace/src/config.js
```diff
@@ -645,262 +555,324 @@ class Config {
...
-        : 'continue'
-    this.#setBoolean(env, 'tracePropagationStyle.otelPropagators',
-      DD_TRACE_PROPAGATION_STYLE ||
-      DD_TRACE_PROPAGATION_STYLE_INJECT ||
+    target.spanLeakDebug = maybeInt(DD_TRACE_SPAN_LEAK_DEBUG)
+    this.#setBoolean(target, 'spanRemoveIntegrationFromService', DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED)
+    this.#setBoolean(target, 'startupLogs', DD_TRACE_STARTUP_LOGS)
+    this.#setTags(target, 'tags', tags)
+    target.tagsHeaderMaxLength = DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH
+    this.#setBoolean(target, 'telemetry.enabled', DD_INSTRUMENTATION_TELEMETRY_ENABLED)
```
BridgeAR, line 824: Same as with remote config enabled:

This change makes sense as such, that we expect a user never to define this value if in a serverless environment or when using jest. If they do however use a single configuration for serverless and non-serverless and they have this set to enabled, it would cause a regression, if I am not mistaken.

I wonder if we just want to ignore the value if we are in serverless or when using jest. That way it would indeed fall back to the default value, which we can also define as such. In addition, we could use a `unprocessed` object for it, so that we still know the user defined it.

It is just an interesting edge case for how to handle it correct.

packages/dd-trace/src/config.js
```diff
@@ -210,30 +196,33 @@ function remapify (input, mappings) {
...
 
-function propagationStyle (key, option) {
-  // Extract by key if in object-form value
-  if (option !== null && typeof option === 'object' && !Array.isArray(option)) {
-    option = option[key]
+/**
+ * Normalizes propagation style values to a lowercase array.
+ * Handles both string (comma-separated) and array inputs.
+ */
+function normalizePropagationStyle (value) {
```
BridgeAR, line 203: This is much cleaner, thank you for the refactoring! It also helps with knowing what envs are used. Good work :)

packages/dd-trace/src/config.js
```diff
@@ -645,262 +555,324 @@ class Config {
...
+      const stringPropagationBehaviorExtract = String(DD_TRACE_PROPAGATION_BEHAVIOR_EXTRACT)
+      target.tracePropagationBehaviorExtract =
+        VALID_PROPAGATION_BEHAVIOR_EXTRACT.has(stringPropagationBehaviorExtract)
+          ? stringPropagationBehaviorExtract
+          : 'continue'
+    }
+    if (DD_TRACE_PROPAGATION_STYLE !== undefined ||
+        DD_TRACE_PROPAGATION_STYLE_INJECT !== undefined ||
+        DD_TRACE_PROPAGATION_STYLE_EXTRACT !== undefined ||
+        OTEL_PROPAGATORS !== undefined) {
```
BridgeAR, line 862: I guess this is a fix?
If I am not mistaken, this prevents the value from being set if all envs are undefined, while it would have been set before unconditionally.

packages/dd-trace/src/config.js
```diff
@@ -645,262 +555,324 @@ class Config {
...
+      this.#setBoolean(target, 'tracePropagationStyle.otelPropagators',
+        useDdStyle ? false : !!OTEL_PROPAGATORS)
+
+      // Use OTEL_PROPAGATORS if no DD-specific vars are set
+      if (!useDdStyle && OTEL_PROPAGATORS) {
+        const otelStyles = normalizePropagationStyle(OTEL_PROPAGATORS)
+        // Validate OTEL propagators
+        for (const style of otelStyles || []) {
+          if (!VALID_PROPAGATION_STYLES.has(style)) {
+            log.warn('unexpected value for OTEL_PROPAGATORS environment variable')
```
BridgeAR: Should we potentially log the unexpected style as well? I know that is currently also not the case.

packages/dd-trace/src/config.js
```diff
@@ -645,262 +555,324 @@ class Config {
...
+            getCounter('otel.env.invalid', 'DD_TRACE_PROPAGATION_STYLE', 'OTEL_PROPAGATORS').inc()
+            break // Only warn once
```
BridgeAR: That changes the behavior to the former implementation. It seems logical as done now, while I do not know if there was a reason for it to increment the value multiple times before.

We could also gather all styles that are wrong and log them as one message (as addition to my former comment).

packages/dd-trace/test/config.spec.js
```diff
@@ -480,7 +480,7 @@ describe('Config', () => {
         value: '(?:p(?:ass)?w(?:or)?d|pass(?:_?phrase)?|secret|(?:api_?|private_?|public_?|access_?|secret_?)key(?:_?id)?|token|consumer_?(?:id|key|secret)|sign(?:ed|ature)?|auth(?:entication|orization)?)(?:(?:\\s|%20)*(?:=|%3D)[^&]+|(?:"|%22)(?:\\s|%20)*(?::|%3A)(?:\\s|%20)*(?:"|%22)(?:%2[^2]|%[^2]|[^"%])+(?:"|%22))|bearer(?:\\s|%20)+[a-z0-9\\._\\-]+|token(?::|%3A)[a-z0-9]{13}|gh[opsu]_[0-9a-zA-Z]{36}|ey[I-L](?:[\\w=-]|%3D)+\\.ey[I-L](?:[\\w=-]|%3D)+(?:\\.(?:[\\w.+\\/=-]|%3D|%2F|%2B)+)?|[\\-]{5}BEGIN(?:[a-z\\s]|%20)+PRIVATE(?:\\s|%20)KEY[\\-]{5}[^\\-]+[\\-]{5}END(?:[a-z\\s]|%20)+PRIVATE(?:\\s|%20)KEY|ssh-rsa(?:\\s|%20)*(?:[a-z0-9\\/\\.+]|%2F|%5C|%2B){100,}',
         origin: 'default'
       },
-      { name: 'remoteConfig.enabled', value: true, origin: 'env_var' },
+      { name: 'remoteConfig.enabled', value: true, origin: 'default' },
```
BridgeAR, line 483: Thank you for those fixes!

packages/dd-trace/test/config.spec.js
```diff
@@ -2955,6 +2954,126 @@ apm_configuration_default:
       const stableConfig = new Config()
       expect(stableConfig).to.not.have.property('stableConfig')
     })
+    it('should support all extended configs across product areas', () => {
```
BridgeAR, line 2961: Nit: would you mind adding a new line between the test cases? :)

packages/dd-trace/src/config.js
```diff
@@ -452,35 +376,14 @@ class Config {
...
-    this.#setBoolean(obj, 'dynamicInstrumentation.enabled', DD_DYNAMIC_INSTRUMENTATION_ENABLED)
-    this.#setString(obj, 'env', DD_ENV)
-    this.#setBoolean(obj, 'iast.enabled', DD_IAST_ENABLED)
-    this.#setBoolean(obj, 'logInjection', DD_LOGS_INJECTION)
-    const profilingEnabled = normalizeProfilingEnabledValue(DD_PROFILING_ENABLED)
-    this.#setString(obj, 'profiling.enabled', profilingEnabled)
-    this.#setBoolean(obj, 'runtimeMetrics.enabled', DD_RUNTIME_METRICS_ENABLED)
-    this.#setString(obj, 'service', DD_SERVICE)
-    this.#setString(obj, 'version', DD_VERSION)
+    this.#applyConfigValues(config, obj, {})
```
BridgeAR, line 380: The third argument means we will fill that object but never use it, as far as I can tell. Should we potentially just not set the unprocessed values, if we do not pass through a third argument?

mtoffl01: I actually do not understand what the "unprocessedTarget" does. I just used `{}` to maintain existing patterns.

I guess the question is: Does StableConfig need an "unprocessedTarget"? If not, then yes, we could:
1. Not pass a third argument 
2. Modify the logic in applyConfigValues to only set entries on unprocessedTarget if it exists

WDYT?

packages/dd-trace/src/config.js
mtoffl01: With my changes, this logic now runs multiple times. So I need to ensure that a value is only set when it has actually been configured - and **not** when it’s not.

Previously, the value of otelPropagators was determined by the combined "truthiness" of several settings. Even if none of those settings were configured, the logic still produced a value for otelPropagators (e.g, false in this case).

Now, consider a scenario where one of the relevant keys is set through environment variables, resulting in otelPropagators = true based on user configuration. Next, applyConfigValues runs again for fleet stable config, but none of the fleet settings are configured. Under the old logic, this second pass would assign false to otelPropagators, from the higher precedence fleet stable config source - unintentionally overwriting the user’s environment-based configuration. 

So, I updated the logic here - and a few other places in applyConfigValues - to prevent that from happening.

packages/dd-trace/src/config.js
mtoffl01: Sure, why not! The more info the better.

packages/dd-trace/src/config.js
mtoffl01: good catch.

BridgeAR: Do we know if it should increment it multiple times though? Maybe that was a bug before? 😅

packages/dd-trace/src/config.js
BridgeAR: Unprocessed exists because we used to override some values with calculated ones. To still send out the original user defined value, we tracked both cases. That is definitely not ideal and it would be better not to change anything in the first place. That way unprocessed would not be necessary at all.
I believe your changes pretty much resolve most of this already, I just did not check if all cases.

packages/dd-trace/src/config.js
```diff
@@ -646,262 +575,328 @@ class Config {
...
+    this.#setBoolean(target, 'clientIpEnabled', DD_TRACE_CLIENT_IP_ENABLED)
+    this.#setString(target, 'clientIpHeader', DD_TRACE_CLIENT_IP_HEADER?.toLowerCase())
+    if (DD_TRACE_CLOUD_REQUEST_PAYLOAD_TAGGING || DD_TRACE_CLOUD_RESPONSE_PAYLOAD_TAGGING) {
+      if (DD_TRACE_CLOUD_REQUEST_PAYLOAD_TAGGING) {
+        this.#setBoolean(target, 'cloudPayloadTagging.requestsEnabled', true)
+      }
+      if (DD_TRACE_CLOUD_RESPONSE_PAYLOAD_TAGGING) {
+        this.#setBoolean(target, 'cloudPayloadTagging.responsesEnabled', true)
+      }
+      target['cloudPayloadTagging.rules'] = appendRules(
```
BridgeAR, line 666:
```suggestion
      target.cloudPayloadTagging.rules = appendRules(
```

And the object might have to be initialized up front as well as empty object.

packages/dd-trace/src/config.js
```diff
@@ -646,262 +575,328 @@ class Config {
...
+        splitJSONPathRules(DD_TRACE_CLOUD_REQUEST_PAYLOAD_TAGGING),
+        splitJSONPathRules(DD_TRACE_CLOUD_RESPONSE_PAYLOAD_TAGGING)
+      )
+    }
+    if (DD_TRACE_CLOUD_PAYLOAD_TAGGING_MAX_DEPTH) {
+      target['cloudPayloadTagging.maxDepth'] = maybeInt(DD_TRACE_CLOUD_PAYLOAD_TAGGING_MAX_DEPTH)
```
BridgeAR, line 672:
```suggestion
      target.cloudPayloadTagging.maxDepth = maybeInt(DD_TRACE_CLOUD_PAYLOAD_TAGGING_MAX_DEPTH)
```

packages/dd-trace/src/config.js
```diff
@@ -954,6 +949,17 @@ class Config {
     this.#optsUnprocessed['appsec.wafTimeout'] = options.appsec?.wafTimeout
     this.#setBoolean(opts, 'clientIpEnabled', options.clientIpEnabled)
     this.#setString(opts, 'clientIpHeader', options.clientIpHeader?.toLowerCase())
+    if (options.cloudPayloadTagging) {
+      this.#setBoolean(opts, 'cloudPayloadTagging.requestsEnabled', options.cloudPayloadTagging.requestsEnabled)
+      this.#setBoolean(opts, 'cloudPayloadTagging.responsesEnabled', options.cloudPayloadTagging.responsesEnabled)
+      opts['cloudPayloadTagging.maxDepth'] = maybeInt(options.cloudPayloadTagging.maxDepth)
```
BridgeAR:
```suggestion
      opts.cloudPayloadTagging.maxDepth = maybeInt(options.cloudPayloadTagging.maxDepth)
```

packages/dd-trace/src/config.js
```diff
@@ -954,6 +949,17 @@ class Config {
...
+      if (options.cloudPayloadTagging.request || options.cloudPayloadTagging.response) {
+        opts['cloudPayloadTagging.rules'] = appendRules(
```
BridgeAR:
```suggestion
        opts.cloudPayloadTagging.rules = appendRules(
```

BridgeAR (approved):
Great work!

LGTM

PR #6708, opened 2025-10-20
Fix to cookie-parser failing to instrument

pabloerhard:
### What does this PR do?
Fix cookie-parser breaking during ESM instrumentation and added ESM tests.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/src/cookie-parser.js
+++ b/packages/datadog-instrumentations/src/cookie-parser.js
@@ -25,6 +25,7 @@ addHook({
   name: 'cookie-parser',
   versions: ['>=1.0.0']
 }, cookieParser => {
+  // This prevents the non default export from entering the wrapping process
   if (cookieParser.default) return cookieParser
   return shimmer.wrapFunction(cookieParser, cookieParser => function () {
     const cookieMiddleware = cookieParser.apply(this, arguments)
--- a/packages/datadog-plugin-cookie-parser/test/integration-test/client.spec.js
+++ b/packages/datadog-plugin-cookie-parser/test/integration-test/client.spec.js
@@ -15,7 +15,7 @@ withVersions('cookie-parser', 'cookie-parser', version => {
       this.timeout(50000)
       sandbox = await createSandbox([`'cookie-parser@${version}'`, 'express'], false,
         ['./packages/datadog-plugin-cookie-parser/test/integration-test/*'])
-      variants = varySandbox(sandbox, 'server.mjs', 'cookie-parser', undefined, 'cookieParser')
+      variants = varySandbox(sandbox, 'server.mjs', 'cookie-parser', 'cookieParser')
     })
 
     after(async function () {
@@ -28,7 +28,7 @@ withVersions('cookie-parser', 'cookie-parser', version => {
     })
 
     afterEach(async () => {
-      proc && proc.kill()
+      proc?.kill()
       await agent.stop()
     })
 
@@ -36,8 +36,7 @@ withVersions('cookie-parser', 'cookie-parser', version => {
       it(`is instrumented loaded with ${variant}`, async () => {
         proc = await spawnPluginIntegrationTestProc(sandbox.folder, variants[variant], agent.port)
         const response = await curl(proc)
-        const counterValue = response.headers['x-counter']
-        assert.equal(counterValue, '1')
+        assert.equal(response.headers['x-counter'], '1')
       })
     }
   })


packages/datadog-instrumentations/src/cookie-parser.js
```diff
@@ -25,6 +25,7 @@ addHook({
   name: 'cookie-parser',
   versions: ['>=1.0.0']
 }, cookieParser => {
+  if (cookieParser.default) return cookieParser
```
IlyasShabi, line 29: I think that we should also instrument default when present no?
something like:
```
const wrapped = shimmer.wrapFunction(cookieParser, cookieParser...)
cookieParser.default = wrapped
return wrapped
```

packages/datadog-plugin-cookie-parser/test/integration-test/client.spec.js
```diff
@@ -0,0 +1,44 @@
...
+      this.timeout(50000)
+      await sandbox.remove()
+    })
+
+    beforeEach(async () => {
+      agent = await new FakeAgent().start()
+    })
+
+    afterEach(async () => {
+      proc && proc.kill()
```
IlyasShabi:
```suggestion
      proc?.kill()
```

packages/datadog-plugin-cookie-parser/test/integration-test/client.spec.js
```diff
@@ -0,0 +1,44 @@
...
+      await agent.stop()
+    })
+
+    for (const variant of varySandbox.VARIANTS) {
+      it(`is instrumented loaded with ${variant}`, async () => {
```
simon-id, line 36: you're not really testing that it's instrumented right ?
If the instrumentation was completely disabled this test would still pass no ?

packages/datadog-instrumentations/src/cookie-parser.js
BridgeAR: The change is actually a fix for a bug. The reason is CJS / ESM interoperability. ESM loading is always going to have a default, while it is identical in case it is a CJS module.
We should therefore not wrap it again.

packages/datadog-plugin-cookie-parser/test/integration-test/client.spec.js
BridgeAR: This tests that the loading of the module works as before for all ESM variations. That recently regressed.

We could add a comment that this only tests that the module works with instrumentation as smoke test and does not produce own spans.

packages/datadog-plugin-cookie-parser/test/integration-test/client.spec.js
```diff
@@ -0,0 +1,44 @@
...
+
+withVersions('cookie-parser', 'cookie-parser', version => {
+  describe('ESM', () => {
+    let sandbox, variants, proc, agent
+
+    before(async function () {
+      this.timeout(50000)
+      sandbox = await createSandbox([`'cookie-parser@${version}'`, 'express'], false,
+        ['./packages/datadog-plugin-cookie-parser/test/integration-test/*'])
+      variants = varySandbox(sandbox, 'server.mjs', 'cookie-parser', undefined, 'cookieParser')
```
BridgeAR:
```suggestion
      variants = varySandbox(sandbox, 'server.mjs', 'cookie-parser', 'cookieParser')
```

BridgeAR (approved):
LGTM. Adding a comment is likely a good idea to make sure the intent of the test is clear (or change the describe block accordingly)

packages/datadog-plugin-cookie-parser/test/integration-test/client.spec.js
simon-id: so it only test that loading it won't crash, not that it is instrumented, maybe that should just be the name of the test instead ?

BridgeAR: Yes, pretty much. I agree that a different name would be good.

We could pretty much have smoke tests for all ESM instrumentations loading as before and we just have to verify that the instrumentation code is executed (by e.g., checking that the hook is loaded).

BridgeAR: Another possibility is to create a test where the instrumented library is used. But writing such test is a) more difficult and b) mixes different testing goals into one.

packages/datadog-instrumentations/src/cookie-parser.js
pabloerhard: Well what this change is doing is preventing the non default export from entering our wrapping phase. We run the addHook on the default and also the non default export (which contains the .default property) this is ideally checking if the object has the .default property and if so then don't do anything. This was crashing because it was trying to do a .apply on an object and not a function.

packages/datadog-plugin-cookie-parser/test/integration-test/client.spec.js
pabloerhard: It's also testing that the actual wrapping of cookieParser gets applied and executed when using ESM. If the instrumentation was disabled it wouldn't pass considering the cookie parser dc channel would never publish anything making the last assertion fail.

packages/datadog-instrumentations/src/cookie-parser.js
simon-id: a comment to explain this could be nice since both me and ilyas didn't understand the point of this at first :+1:

packages/datadog-plugin-cookie-parser/test/integration-test/client.spec.js
simon-id: Aaaah yes indeed, i read too fast, i didn't see the weird DC feedback loop here :+1: closing the thread

packages/datadog-instrumentations/src/cookie-parser.js
IlyasShabi: Thanks for making this clear. A comment could be nice as Simon said

BridgeAR (approved):

PR #6709, opened 2025-10-21
ci(llmobs): re-work `deepEqualWithMockValues` to use `node:assert`

sabrenner:
### What does this PR do?

Changes the LLM Observability `deepEqualWithMockValues` to use `node:assert` instead of mocha `expect`. Also in doing this, refactor the logic to do bi-directional assertions. The reason this is a bit complicated is because we sometimes want to assert structure with "mock" values in certain places. This is a common practice in our LLM integration testing, especially for related metadata, as shape of objects change often, and we just want to assert that something exists (`MOCK_ANY`), or exists with a certain type (`MOCK_STRING`, `MOCK_OBJECT`, etc.).

### Motivation

Fix the test utility to properly assert and not rely on mocha utilities.

### Follow-up items

There are a few follow-up items to re-enable tests that previously did not fail (incorrectly):

- Fix the `ai` integration for correct tool argument tagging for v5 of the library
- Undefined messages for Cohere models with aws-sdk bedrockruntime
- No output roles for cached token tests for aws-sdk bedrock runtime
- A langchain test seems flaky due to a VCR incompatibility. the test does pass occasionally, so i don't believe it to be an issue with the test. i would like to do a follow-up investigation into how the underlying operation of that test (batched llm calls) is interacting with the testagent's vcr capabilities
- A couple OpenAI errors where input metadata is not recorded correctly when the underlying openai call fails
- Skipped an evaluation metric test which I will quick follow up on re-enabling. Just needs a handler in our local testagent server and an additional method in our llm observability test utilities

Additionally, there are some other tests that need to be updated in a similar cleanup initiative:
- The rest of the LLM Observability SDK tests still use `expect`, and should be changed to use `node:assert`
- The existing "integration" tests should probably be migrated to actual integration tests

### DIFF
--- a/packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
+++ b/packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
@@ -238,6 +238,12 @@ describe('integrations', () => {
 
           for await (const part of stream) {
             assert.ok(part, 'Expected part to be truthy')
+            // last chunk will have no choices, but a usage block instead
+            if (part.choices.length > 0) {
+              assert.ok(part.choices[0].text != null, 'Expected chunk delta to be truthy')
+            } else {
+              assert.ok(part.usage, 'Expected usage to be truthy')
+            }
           }
 
           const { apmSpans, llmobsSpans } = await getEvents()
@@ -290,6 +296,12 @@ describe('integrations', () => {
 
           for await (const part of stream) {
             assert.ok(part, 'Expected part to be truthy')
+            // last chunk will have no choices, but a usage block instead
+            if (part.choices.length > 0) {
+              assert.ok(part.choices[0].delta != null, 'Expected chunk delta to be truthy')
+            } else {
+              assert.ok(part.usage, 'Expected usage to be truthy')
+            }
           }
 
           const { apmSpans, llmobsSpans } = await getEvents()
@@ -349,6 +361,12 @@ describe('integrations', () => {
 
           for await (const part of stream) {
             assert.ok(part, 'Expected part to be truthy')
+            // last chunk will have no choices, but a usage block instead
+            if (part.choices.length > 0) {
+              assert.ok(part.choices[0].delta != null, 'Expected chunk delta to be truthy')
+            } else {
+              assert.ok(part.usage, 'Expected usage to be truthy')
+            }
           }
 
           const { apmSpans, llmobsSpans } = await getEvents()
--- a/packages/dd-trace/test/llmobs/util.js
+++ b/packages/dd-trace/test/llmobs/util.js
@@ -12,7 +12,7 @@ const tracerVersion = require('../../../../package.json').version
 const MOCK_STRING = Symbol('string')
 const MOCK_NUMBER = Symbol('number')
 const MOCK_OBJECT = Symbol('object')
-const MOCK_ANY = Symbol('any')
+const MOCK_NOT_NULLISH = Symbol('not-nullish')
 
 /**
  * @typedef {{
@@ -52,7 +52,7 @@ function assertWithMockValues (actual, expected, key) {
     assert.equal(typeof actual, 'number', `${actualWithName} (${util.inspect(actual)}) is not a number`)
   } else if (expected === MOCK_OBJECT) {
     assert.equal(typeof actual, 'object', `${actualWithName} (${util.inspect(actual)}) is not an object`)
-  } else if (expected === MOCK_ANY) {
+  } else if (expected === MOCK_NOT_NULLISH) {
     assert.ok(actual != null, `${actualWithName} does not exist`)
   } else if (Array.isArray(expected)) {
     assert.ok(Array.isArray(actual), `${actualWithName} (${util.inspect(actual)}) is not an array`)
@@ -62,11 +62,8 @@ function assertWithMockValues (actual, expected, key) {
       `${actualWithName} has different length than expected (${actual.length} !== ${expected.length})`
     )
 
-    const sortedExpected = [...expected.sort()]
-    const sortedActual = [...actual.sort()]
-
-    for (let i = 0; i < sortedExpected.length; i++) {
-      assertWithMockValues(sortedActual[i], sortedExpected[i], `${key}.${i}`)
+    for (let i = 0; i < expected.length; i++) {
+      assertWithMockValues(actual[i], expected[i], `${key}.${i}`)
     }
   } else if (typeof expected === 'object') {
     if (typeof actual !== 'object') {
@@ -82,6 +79,7 @@ function assertWithMockValues (actual, expected, key) {
     }
 
     for (const objKey of expectedKeys) {
+      assert.ok(Object.hasOwn(actual, objKey), `${actualWithName} does not have key ${objKey}`)
       assertWithMockValues(actual[objKey], expected[objKey], `${key}.${objKey}`)
     }
   } else {
@@ -178,7 +176,7 @@ function assertLlmObsSpanEvent (actual, expected = {}) {
     )
   }
 
-  // assert arbitrary objects (mock values)
+  // 1. assert arbitrary objects (mock values)
   const actualMetrics = actual.metrics
   const actualMetadata = actual.meta.metadata
   const actualOutputMessages = actual.meta.output.messages
@@ -197,7 +195,19 @@ function assertLlmObsSpanEvent (actual, expected = {}) {
   assertWithMockValues(actualTraceId, traceId, 'traceId')
   assertWithMockValues(actualMetrics, metrics ?? {}, 'metrics')
   assertWithMockValues(actualMetadata, metadata, 'metadata')
-  assertWithMockValues(actualTags, expectedLLMObsTags({ span, tags, error, errorType: error?.type, sessionId }), 'tags')
+
+  // 1a. sort tags since they might be unordered
+  const expectedTags = expectedLLMObsTags({ span, tags, error, sessionId })
+  const sortedExpectedTags = [...expectedTags.sort()]
+  const sortedActualTags = [...actualTags.sort()]
+  for (let i = 0; i < sortedExpectedTags.length; i++) {
+    assert.equal(
+      sortedActualTags[i],
+      sortedExpectedTags[i],
+      `tags[${i}] does not match expected (${sortedExpectedTags[i]} !== ${sortedActualTags[i]})`
+    )
+  }
+
   if (outputMessages) {
     assertWithMockValues(actualOutputMessages, outputMessages, 'outputMessages')
   } else if (outputDocuments) {
@@ -206,7 +216,7 @@ function assertLlmObsSpanEvent (actual, expected = {}) {
     assertWithMockValues(actualOutputValue, outputValue, 'outputValue')
   }
 
-  // assert deepEqual on everything else
+  // 2. assert deepEqual on everything else
   const expectedMeta = { 'span.kind': spanKind }
 
   if (modelName) expectedMeta.model_name = modelName
@@ -242,7 +252,6 @@ function assertLlmObsSpanEvent (actual, expected = {}) {
 function expectedLLMObsTags ({
   span,
   error,
-  errorType,
   tags,
   sessionId
 }) {
@@ -261,7 +270,7 @@ function expectedLLMObsTags ({
     'language:javascript'
   ]
 
-  if (errorType) spanTags.push(`error_type:${errorType}`)
+  if (error) spanTags.push(`error_type:${span.meta[ERROR_TYPE]}`)
   if (sessionId) spanTags.push(`session_id:${sessionId}`)
 
   for (const [key, value] of Object.entries(tags)) {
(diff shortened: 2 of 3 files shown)


packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
```diff
@@ -243,15 +231,17 @@ describe('integrations', () => {
...
             stream: true,
+            stream_options: {
+              include_usage: true,
+            },
           })
 
           for await (const part of stream) {
-            expect(part).to.have.property('choices')
-            expect(part.choices[0]).to.have.property('text')
+            assert.ok(part, 'Expected part to be truthy')
```
BridgeAR, line 240: Just for me to understand: why is that part changed?
The same question below, so I won't repeat it.

packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
```diff
@@ -514,95 +518,8 @@ describe('integrations', () => {
 
         const { llmobsSpans } = await getEvents()
 
-        expect(llmobsSpans[0]).to.have.property('name', 'DeepSeek.createChatCompletion')
-        expect(llmobsSpans[0].meta).to.have.property('model_provider', 'deepseek')
-      })
-
-      it('submits a completion span with cached token metrics', async () => {
```
BridgeAR, line 521: Is that test really obsolete?

packages/dd-trace/test/llmobs/plugins/ai/index.spec.js
```diff
@@ -53,33 +46,41 @@ describe('Plugin', () => {
...
         outputValue: MOCK_STRING,
-        metadata: {
-          maxTokens: 100,
-          temperature: 0.5,
-          maxSteps: MOCK_NUMBER,
-          maxRetries: MOCK_NUMBER,
-        },
-        tokenMetrics: { input_tokens: MOCK_NUMBER, output_tokens: MOCK_NUMBER, total_tokens: MOCK_NUMBER },
-        tags: { ml_app: 'test', language: 'javascript', integration: 'ai' },
+        metadata: expectedWorkflowMetadata,
```
BridgeAR, line 80: This now tests for less properties. That is probably expected and before these were ignored?

packages/dd-trace/test/llmobs/plugins/ai/index.spec.js
```diff
@@ -53,33 +46,41 @@ describe('Plugin', () => {
...
         spanKind: 'workflow',
         inputValue: 'Hello, OpenAI!',
         outputValue: MOCK_STRING,
-        metadata: {
-          maxTokens: 100,
-          temperature: 0.5,
-          maxSteps: MOCK_NUMBER,
-          maxRetries: MOCK_NUMBER,
-        },
-        tokenMetrics: { input_tokens: MOCK_NUMBER, output_tokens: MOCK_NUMBER, total_tokens: MOCK_NUMBER },
```
BridgeAR, line 78: This does not check for metrics anymore. Is that intentional?

packages/dd-trace/test/llmobs/plugins/ai/index.spec.js
```diff
@@ -162,20 +160,24 @@ describe('Plugin', () => {
...
-        },
-        tags: { ml_app: 'test', language: 'javascript', integration: 'ai' }
-      })
+        tags: { ml_app: 'test', integration: 'ai' }
+      }
 
-      const expectedEmbeddingSpan = expectedLLMObsLLMSpanEvent({
+      if (semifies(realVersion, '>=5.0.0')) {
+        expectedWorkflowSpanEvent.metadata = {
+          maxRetries: MOCK_NUMBER
```
BridgeAR, line 174: maxSteps should not be tested for anymore?

packages/dd-trace/test/llmobs/plugins/ai/index.spec.js
```diff
@@ -431,21 +416,24 @@ describe('Plugin', () => {
...
         spanKind: 'workflow',
         inputValue: 'What is the weather in Tokyo?',
         outputValue: expectedFinalOutput,
-        metadata: {
-          maxSteps: MOCK_NUMBER,
-          maxRetries: MOCK_NUMBER,
-        },
-        tokenMetrics: { input_tokens: MOCK_NUMBER, output_tokens: MOCK_NUMBER, total_tokens: MOCK_NUMBER },
-        tags: { ml_app: 'test', language: 'javascript', integration: 'ai' },
+        metadata: expectedWorkflowMetadata,
```
BridgeAR, line 432: This also tests for less properties. I believe that is fine?

packages/dd-trace/test/llmobs/util.js
```diff
@@ -10,144 +14,229 @@ const MOCK_NUMBER = Symbol('number')
...
+  const actualWithName = key ? `Actual (${key})` : 'Actual'
+
+  if (expected === MOCK_STRING) {
+    assert.equal(typeof actual, 'string', `${actualWithName} (${util.inspect(actual)}) is not a string`)
+  } else if (expected === MOCK_NUMBER) {
+    assert.equal(typeof actual, 'number', `${actualWithName} (${util.inspect(actual)}) is not a number`)
+  } else if (expected === MOCK_OBJECT) {
+    assert.equal(typeof actual, 'object', `${actualWithName} (${util.inspect(actual)}) is not an object`)
+  } else if (expected === MOCK_ANY) {
+    assert.ok(actual != null, `${actualWithName} does not exist`)
```
BridgeAR, line 56: I am good with this, I would just likely rename `MOCK_ANY` to `MOCK_NOT_NULLISH`

packages/dd-trace/test/llmobs/util.js
```diff
@@ -10,144 +14,229 @@ const MOCK_NUMBER = Symbol('number')
...
+  } else if (Array.isArray(expected)) {
+    assert.ok(Array.isArray(actual), `${actualWithName} (${util.inspect(actual)}) is not an array`)
+    assert.equal(
+      actual.length,
+      expected.length,
+      `${actualWithName} has different length than expected (${actual.length} !== ${expected.length})`
+    )
+
+    const sortedExpected = [...expected.sort()]
+    const sortedActual = [...actual.sort()]
```
BridgeAR: This still feels a bit off to me. I would expect numbers to have a semantical meaning in its order. Non-primitive entries will not be sorted anyway.

packages/dd-trace/test/llmobs/util.js
```diff
@@ -10,144 +14,229 @@ const MOCK_NUMBER = Symbol('number')
...
+    const actualKeys = Object.keys(actual)
+    const expectedKeys = Object.keys(expected)
+    if (actualKeys.length !== expectedKeys.length) {
+      assert.fail(
+        `${actualWithName} has different length than expected (${actualKeys.length} !== ${expectedKeys.length})`
+      )
     }
+
+    for (const objKey of expectedKeys) {
+      assertWithMockValues(actual[objKey], expected[objKey], `${key}.${objKey}`)
```
BridgeAR, line 83: If a value is undefined on one side, this could hide errors due to not checking for the property existence.

packages/dd-trace/test/llmobs/sdk/integration.spec.js
```diff
@@ -392,61 +325,57 @@ describe('end to end sdk integration tests', () => {
...
-      const { llmobsSpans } = run(payloadGenerator)
-      expect(llmobsSpans).to.have.lengthOf(6)
+        llmobs.trace({ kind: 'workflow', name: 'afterAnnotationContext' }, () => {})
+      })
+
+      const { llmobsSpans } = await getEvents()
+      assert.equal(llmobsSpans.length, 6)
 
-      expect(llmobsSpans[0].tags).to.not.include('foo:bar')
+      assert.equal(getTag(llmobsSpans[0], 'foo'), undefined)
```
BridgeAR, line 363: This looks like it was `bar` before? The same below.

packages/dd-trace/test/llmobs/sdk/integration.spec.js
```diff
@@ -189,102 +131,94 @@ describe('end to end sdk integration tests', () => {
...
-        ml_app: 'test',
-        timestamp_ms: 1234567890,
-        tags: [`ddtrace.version:${tracerVersion}`, 'ml_app:test']
-      }
-    ]
-
-    check(expected, evaluationMetrics)
 
-    Date.now.restore()
+      // const { spans, llmobsSpans, evaluationMetrics } = run(payloadGenerator)
```
BridgeAR, line 135: What about keeping the code now that the test is skipped?

packages/dd-trace/test/llmobs/plugins/ai/index.spec.js
sabrenner: The extra properties were not being asserted before, I think because we were iterating over the actual keys: https://github.com/DataDog/dd-trace-js/blob/3e68da87721831c8c837d90ecfd6246c96441ef4/packages/dd-trace/test/llmobs/util.js#L16, which would mean extra keys in the expected were never checked. the new function fixes this, however.

packages/dd-trace/test/llmobs/plugins/ai/index.spec.js
sabrenner: yes, similar answer to above in that the metrics should never have been on these spans to begin with (for reference, token metrics should only be on `llm` and `embedding` LLMObs spans, and not the other span kinds).

packages/dd-trace/test/llmobs/plugins/ai/index.spec.js
sabrenner: also a leftover where it was not previously being asserted correctly. the actual embedding call here has no reference to `maxSteps`, and was a bad copy-paste from a different test when writing it originally

packages/dd-trace/test/llmobs/plugins/ai/index.spec.js
sabrenner: yes, this is fine here, because they were never being asserted properly before (as we iterated over `actual` previously)

packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
sabrenner: `stream_options: { include_usage: true }` will make the last chunk undefined for some reason. i'll just add in a check with a comment explaining this in both cases 😄

packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
sabrenner: yeah, the `completions` operation actually doesn't support cached token metrics, all of my tests & googling revealed that it's not recorded

packages/dd-trace/test/llmobs/sdk/integration.spec.js
sabrenner: i didn't want to write another helper in this PR to assert eval metrics (different payload struct), and not having this commented would just leave a bunch of errors for stuff not imported anymore. i do plan on following right up with this, if that sounds good!

packages/dd-trace/test/llmobs/sdk/integration.spec.js
sabrenner: the assertion before was `to.not.include`, so i think asserting that it's `undefined` here is OK, but lmk if i'm misunderstanding!

packages/dd-trace/test/llmobs/util.js
sabrenner: will do!

packages/dd-trace/test/llmobs/util.js
sabrenner: oh yeah i only kept this in here because doing an assert for the tags array wasn't working just with a standard `deepEquals` with the larger diff, so i added below to be asserted with mock values only to sort the array. 

maybe i can just do that one-off though (before the big diff assertion, checking the sorted expected and actual tags separately, and then stripping it out for the final diff check), and then i can take this sorting bit out here. i'll actually try that!

packages/dd-trace/test/llmobs/util.js
sabrenner: ah ok, so maybe adding in some `Object.hasOwn` checks here would be good?

packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
sabrenner: or, rather, the last chunk is **not** undefined, but `part.choices` is, as the last chunk is reserved for usage. i'll make sure that's reflected in the assertions 😄

packages/dd-trace/test/llmobs/util.js
BridgeAR: Yes, that would solve that :)
The hasOwn check would also only have to happen if it is undefined.

packages/dd-trace/test/llmobs/sdk/integration.spec.js
BridgeAR: Thanks, I overlooked the negation :)

BridgeAR (approved):

PR #6712, opened 2025-10-21
IAST ESM support for iitm 2

CarlesDD:
### What does this PR do?
Fixes ESM loader detection for IAST rewriter to support `import-in-the-middle` v2.0.0+.

**Key changes:**
- Sets a global flag (`globalThis.__DD_ESM_LOADER_ACTIVE__`) in `loader-hook.mjs` / `initialize.mjs` to indicate when the ESM loader is active
- Updates `isEsmConfigured()` in `rewriter.js` to check for this flag

**Detection logic (in order):**
1. Check for `--loader` or `--experimental-loader` flags
2. Check for global flag set by `loader-hook.mjs`
3. Return `false` if neither condition is met

### Motivation
`import-in-the-middle` v2.0.0 converted all modules running in loader thread to ESM, breaking the existing detection mechanism.

**The problem:**
- **Before (< v2.0.0):** `hook.js` was loaded via `require()`, so it appeared in `require.cache`
- **After (v2.0.0+):** `hook.mjs` is loaded via ESM imports, so it does **NOT** appear in `require.cache`
- The existing `isEsmConfigured()` function relied on checking `require.cache` for `hook.js`, which failed with v2.0.0+

**The solution:**
Set an explicit global flag when `loader-hook.mjs` is loaded. This works because:
- All ESM loader scenarios (regardless of Node.js version or entry point) eventually load `loader-hook.mjs`
- When using `--loader`/`--experimental-loader` flags, we can detect them directly from process arguments
- The global flag catches cases where the loader is registered programmatically

**Supported scenarios:**

| Scenario | Node.js Version | Detection Method |
|----------|----------------|------------------|
| `--loader dd-trace/loader-hook.mjs` | < v20.6 | `--loader` flag |
| `--import dd-trace/register.js` | ≥ v20.6 | Global flag |
| `--import dd-trace/initialize.mjs` | ≥ v20.6 | Global flag |

### Plugin Checklist

- [ ] Unit tests.
- [x] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes
[ESM applications only: Import the loader](https://docs.datadoghq.com/tracing/trace_collection/automatic_instrumentation/dd_libraries/nodejs/#esm-applications-only-import-the-loader)

### DIFF
--- a/initialize.mjs
+++ b/initialize.mjs
@@ -22,6 +22,10 @@ import {
   getSource as origGetSource,
 } from 'import-in-the-middle/hook.mjs'
 
+// Set global flag to indicate ESM loader is active
+// This is checked by rewriter.js to enable ESM rewriting for IAST
+globalThis.__DD_ESM_LOADER_ACTIVE__ = true
+
 let hasInsertedInit = false
 function insertInit (result) {
   if (!hasInsertedInit) {


initialize.mjs
```diff
@@ -22,6 +22,10 @@ import {
   getSource as origGetSource,
 } from 'import-in-the-middle/hook.mjs'
 
+// Set global flag to indicate ESM loader is active
+// This is checked by rewriter.js to enable ESM rewriting for IAST
+globalThis.__DD_ESM_LOADER_ACTIVE__ = true
```
BridgeAR, line 27: We start to pollute the global space quite a bit. What about using a single symbol that is an object for all dd-trace-js related things we want to add to the global space? That way we limit it to that single symbol and it is actually also less visible.

To make it even less visible, I would not make it enumerable either.

PR #6729, opened 2025-10-23
[test optimization][STV-59] Fix `jest.mock` for multi worker setups

juan-fernandez:
### What does this PR do?

Clean up the previous `message` listener before adding the updated one. I'm not sure why this triggered this specific error, but it seems sensible to me anyway.

### Motivation
Fix issue reported in https://github.com/deecewan/dd-trace-mwe/

```javascript
    Cannot find module 'some-package' from 'jest.setup.js'

    > 1 | jest.mock('some-package', () => {
        |      ^
      2 |   return jest.fn(() => 'Hello, mocked world!')
      3 | });
```

This happened whenever using `jest.mock` on an internal package and multi-worker jest setup. 

### Plugin Checklist

- [x] Integration tests.

### DIFF
--- a/integration-tests/jest/jest.spec.js
+++ b/integration-tests/jest/jest.spec.js
@@ -4700,6 +4700,8 @@ describe('jest CommonJS', () => {
   })
 
   it('does not crash with mocks that are not dependencies', async () => {
+    let testOutput = ''
+
     childProcess = exec(
       runTestsCommand,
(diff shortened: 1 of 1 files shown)


deecewan:
looks like this works great! thanks so much for fixing this.

BridgeAR (approved):

PR #6730, opened 2025-10-23
split sandbox between linked and isolated modes

rochdev:
### What does this PR do?

Split sandbox between linked and isolated modes.

### Motivation

The sandbox was originally created for integration tests with applications mimicking a real app with a real install of dd-trace. But for a lot of our tests, for example ESM tests, we don't really need that level of isolation, and linking to dd-trace is good enough.

### Additional Notes

The main change is in `integration-tests/helpers/index.js`, everything else is just changing all usage of `createSandbox` to either `isolatedSandbox` or `linkedSandbox`.

Supersedes #6640

### DIFF
--- a/integration-tests/helpers/index.js
+++ b/integration-tests/helpers/index.js
@@ -303,11 +303,11 @@ async function createSandbox (dependencies = [], isGitRepo = false,
 
   if (isolated) {
     execHelper(`npm pack --silent --pack-destination ${folder}`, { env: restOfEnv })
-    execHelper(`yarn add ${deps.concat(`file:${out}`).join(' ')}`, addOptions)
+    execHelper(`yarn add ${deps.concat(`file:${out}`).join(' ')} ${addFlags.join(' ')}`, addOptions)
   } else {
     execHelper('yarn link')
     if (deps.length > 0) {
-      execHelper(`yarn add ${deps.join(' ')}`, addOptions)
+      execHelper(`yarn add ${deps.join(' ')} ${addFlags.join(' ')}`, addOptions)
     }
     execHelper('yarn link dd-trace', addOptions)
   }
--- a/integration-tests/helpers/index.js
+++ b/integration-tests/helpers/index.js
@@ -236,15 +236,23 @@ function execHelper (command, options) {
   /* eslint-enable no-console */
 }
 
+async function isolatedSandbox (dependencies, isGitRepo, integrationTestsPaths, followUpCommand) {
+  return createSandbox(dependencies, isGitRepo, integrationTestsPaths, followUpCommand, true)
+}
+
+async function linkedSandbox (dependencies, isGitRepo, integrationTestsPaths, followUpCommand) {
+  return createSandbox(dependencies, isGitRepo, integrationTestsPaths, followUpCommand, false)
+}
+
 /**
  * @param {string[]} dependencies
  * @param {boolean} isGitRepo
  * @param {string[]} integrationTestsPaths
  * @param {string} [followUpCommand]
+ * @param {string} [isolated=true]
  */
-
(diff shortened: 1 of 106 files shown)


packages/dd-trace/test/plugins/versions/package.json
```diff
@@ -66,6 +66,7 @@
     "@prisma/client": "6.17.1",
     "@redis/client": "5.8.3",
     "@smithy/smithy-client": "4.9.0",
+    "@types/node": "18.19.130",
```
BridgeAR, line 69: I am surprised about this addition. I believe that should not have an impact, since it is already listed in our main package.json.

packages/dd-trace/test/plugins/versions/package.json
```diff
@@ -93,6 +94,7 @@
     "couchbase": "4.6.0",
     "cypress": "15.4.0",
     "cypress-fail-fast": "7.1.1",
+    "dc-polyfill": "0.1.10",
```
BridgeAR, line 97: I am not sure about this change. Why do we need to add that to the versions package.json?

integration-tests/helpers/index.js
```diff
@@ -242,9 +242,29 @@ function execHelper (command, options) {
...
+/**
+ * @param {string[]} dependencies
+ * @param {boolean} isGitRepo
+ * @param {string[]} integrationTestsPaths
+ * @param {string} [followUpCommand]
+ * @param {string} [isolated=true]
+ */
 async function createSandbox (dependencies = [], isGitRepo = false,
-  integrationTestsPaths = ['./integration-tests/*'], followUpCommand) {
+  integrationTestsPaths = ['./integration-tests/*'], followUpCommand, isolated = true) {
```
BridgeAR, line 267: We do not really need the default value, since it is explicitely passed through all the time.

I would probably change the arguments order to make the default values arguments at the end.

integration-tests/helpers/index.js
```diff
@@ -271,15 +291,26 @@ async function createSandbox (dependencies = [], isGitRepo = false,
...
+  const addFlags = ['--ignore-engines']
+
+  if (process.env.OFFLINE === '1' || process.env.OFFLINE === 'true') {
+    addFlags.push('--prefer-offline')
+  }
 
-  execHelper(addCommand, addOptions)
+  if (isolated) {
+    execHelper(`npm pack --silent --pack-destination ${folder}`, { env: restOfEnv })
+    execHelper(`yarn add ${deps.concat(`file:${out}`).join(' ')} ${addFlags.join(' ')}`, addOptions)
```
BridgeAR, line 306:
```suggestion
    execHelper(`yarn add ${deps.join(' ')} `file:${out}` ${addFlags.join(' ')}`, addOptions)
```

BridgeAR (commented):
Mostly LGTM. I didn't understand the addition of the modules in the versions package.json though.

packages/dd-trace/test/plugins/versions/package.json
rochdev: It's needed by the test app. It was available before because it got hoisted from dd-trace to the test app node_modules which is no longer the case when linking.

packages/dd-trace/test/plugins/versions/package.json
rochdev: It's needed by the test app. It was available before because it got hoisted from dd-trace to the test app node_modules which is no longer the case when linking.

integration-tests/helpers/index.js
rochdev: It's also used by useSandbox, but I guess it could also pass it explicitly.

BridgeAR:
I am fine with that. If it ever comes up as problem again, let us have another look at it

PR #6733, opened 2025-10-23
collect and propagate process tags in first chunk span

tlhunter:
### What does this PR do?
- [x] collects basic process tags
- [x] tag serializer
- [x] send in first span every time we send a bunch of spans to the agent
- [x] create `DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED=false` to configure
- [ ] send via DSM (requires agent back-propagation work)
- [ ] send via DBM (equires agent back-propagation work)
- [ ] send via Profiling (needs https://github.com/DataDog/libdatadog-nodejs to first update to v24)
- [ ] send via Crash Tracking
- [ ] send via Telemetry
- [ ] send via Dynamic Instrumentation
- [ ] send via Client Trace Statistics
- [ ] send via Remote Config

### Motivation
- assists with service renaming

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/process-tags/index.js
+++ b/packages/dd-trace/src/process-tags/index.js
@@ -30,7 +30,7 @@ module.exports.PROFILING_FIELD_NAME = PROFILING_FIELD_NAME
 module.exports = function getProcessTags () {
   // this list is sorted alphabetically for consistent serialization
   const tags = [
-    // the immediate parent directory name of the entrypoint script, e.g. /foo/bar/baz/banana.js -> baz
+    // the parent directory name of the entrypoint script, e.g. /foo/bar/baz/banana.js -> baz
     ['entrypoint.basedir', ENTRYPOINT_PATH === '' ? undefined : path.basename(path.dirname(ENTRYPOINT_PATH))],
 
     // the entrypoint script filename without the extension, e.g. /foo/bar/baz/banana.js -> banana


packages/dd-trace/src/proxy.js
```diff
@@ -78,6 +79,7 @@ class Tracer extends NoopProxy {
     this.getAllBaggageItems = getAllBaggageItems
     this.removeBaggageItem = removeBaggageItem
     this.removeAllBaggageItems = removeAllBaggageItems
+    this.processTags = getProcessTags()
```
vandonr: I imagine that the fact that this is setup here ensures that we'll only "compute" the tag values once ?

tlhunter: Yeah this is initialization code that only happens once early on in a process.

packages/dd-trace/src/process-tags/index.js
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Sanitize a process tag value
+ *
+ * @param {string} value
+ * @returns {string}
+ */
+function sanitize (value) {
+  return String(value)
+    .toLowerCase()
+    .replaceAll(/[^a-zA-Z0-9/_.-]+/g, '_')
```
dubloom, line 78: It's not written in the RFC, but I think `:` is also an authorized character

packages/dd-trace/src/span_processor.js
```diff
@@ -24,6 +25,10 @@ class SpanProcessor {
 
     this._spanSampler = new SpanSampler(config.sampler)
     this._gitMetadataTagger = new GitMetadataTagger(config)
+
+    this._processTags = config.propagateProcessTags?.enabled
+      ? getProcessTags().serialized
+      : false
```
dubloom, line 31: How much can you expose global constants ? process tags will not change during the lifetime of the program. What we do in python is we exposed a constant (which is None or the value depending of the config) and the other modules can grab this constant (which also prevents recomputing the process tags several times)

packages/dd-trace/src/process-tags/index.js
```diff
@@ -0,0 +1,81 @@
...
+
+    // always script for JavaScript applications
+    ['entrypoint.type', 'script'],
+
+    // last segment of the current working directory, e.g. /foo/bar/baz/ -> baz
+    ['entrypoint.workdir', path.basename(CURRENT_WORKING_DIRECTORY) || undefined],
+
+    // the .name field from the application's package.json
+    ['package.json.name', pkg.name || undefined]
+  ]
```
dubloom, line 47: Can this code throw ? It seems it's not the case but wanted to double check

dubloom (commented):
Small comments !

packages/dd-trace/src/config.js
```diff
@@ -453,6 +453,7 @@ class Config {
       DD_DYNAMIC_INSTRUMENTATION_UPLOAD_INTERVAL_SECONDS,
       DD_ENV,
       DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED,
+      DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED,
```
BridgeAR, line 457: Please add a test case for the env. As soon as that is the case, the CI should complain that the value is not in the supported-configurations.json file, so it would also have to be added there.

I guess I have to rework our linter to pick this up earlier :)

packages/dd-trace/src/process-tags/index.js
tlhunter: It looks like today we decided to disallow `:` characters to not disrupt key/value parsing.

packages/dd-trace/src/span_processor.js
tlhunter: It's easily achievable but is considered a "sin" in Node.js to create globals. However we can essentially have "namespaced globals" by exporting a singleton instance in a file (module) and requiring that file elsewhere.

That said the way I've set up the code we should only ever calculate the process tags a single time for the lifetime of the process so it does benefit from the never-changing behavior of process tags.

packages/dd-trace/src/process-tags/index.js
tlhunter: In all my testing I only found a situation where `require.main.filename` could throw as `require.main` is missing in certain situations where the tracer is used in CI during test runs but I addressed that. Based on where this code is placed it should be running in nearly all of our CI jobs and those aren't failing.

packages/dd-trace/src/config.js
tlhunter: Fixed

packages/dd-trace/src/encode/0.4.js
```diff
@@ -9,8 +9,15 @@ const { getEnvironmentVariable } = require('../config-helper')
...
 const SOFT_LIMIT = 8 * 1024 * 1024 // 8MB
 
-function formatSpan (span, config) {
+function formatSpan (span, config, shouldAddProcessTags = false, processTags = false) {
   span = normalizeSpan(truncateSpan(span, false))
+  
+  // Add process tags to the first span of the first trace in a payload
+  if (shouldAddProcessTags && processTags && span.meta) {
+    const { TRACING_FIELD_NAME } = require('./process-tags')
+    span.meta[TRACING_FIELD_NAME] = processTags
```
bengl: Seems to me that all this should be done in `span_format.js`, no?

packages/dd-trace/src/process-tags/index.js
bengl: Just to verify, has the RFC been updated to reflect everything here?

packages/dd-trace/src/process-tags/index.js
```diff
@@ -0,0 +1,81 @@
...
+// TODO: REMOTE_CONFIG_FIELD_NAME process_tags
+
+// $ cd /foo/bar && node baz/banana.js
+// entrypoint.workdir = bar
+// entrypoint.name = banana
+// entrypoint.type = script
+// entrypoint.basedir = baz
+// package.json.name = <from package.json>
+
+module.exports = function getProcessTags () {
```
bengl, line 30: Should the result of this function be cached here?

tlhunter:
I forgot that I had pushed a half-baked commit

packages/dd-trace/src/process-tags/index.js
tlhunter: It's a property of `SpanProcessor` which is attached to the tracer singleton:

```
packages/dd-trace/src/opentracing/tracer.js
5:const SpanProcessor = require('../span_processor')
34:    this._processor = new SpanProcessor(this._exporter, this._prioritySampler, config)
```

packages/dd-trace/src/encode/0.4.js
tlhunter: I reverted all the changes in this file

packages/dd-trace/src/process-tags/index.js
tlhunter: Yes: https://docs.google.com/document/d/1AFdLUuVk70i0bJd5335-RxqsvwAV9ovAqcO2z5mEMbA/edit?pli=1&tab=t.0#heading=h.t8ah57xwikx1

bengl (approved):

PR #6740, opened 2025-10-24
Always redact exception messages for logs in telemetry

bouwkast:
### What does this PR do?

Always redact exception messages for logs that get sent to telemetry.

### Motivation

The requirements / expectation for sending logs to telemetry is to always ensure that we have constant content (outside of DD / known stack frames).

Previously this would check the frame was ours and if so it would include the exception message, this is not allowed.

Additionally, this would populate the stack trace with multiple frames if there were new lines within the exception message logged.

### Plugin Checklist

### Additional Notes

Please reach out to me for more context - I drafted this with Claude as I'm unfamiliar with dd-trace-js and would like help getting this implemented, but the overall logic here looks correct.


BridgeAR (approved):
LGTM

Issue #6743, opened 2025-10-24
[BUG]: Installs of dd-trace@5.73.0 fail due to peer-dependency violations.


WattyRev:
### Tracer Version(s)

5.73.0

### Node.js Version(s)

22.21.0

### Bug Report

The dd-trace package lists `@openfeature/server-sdk` as a peer dependency at `~1.19.0` ([source](https://github.com/DataDog/dd-trace-js/blob/master/package.json#L163)). dd-trace installs `@datadog/openfeature-node-server@0.1.0-preview.12` ([source](https://github.com/DataDog/dd-trace-js/blob/master/package.json#L128C42-L128C58)), which installs `@openfeature/server-sdk@~1.18.0`. The transient dependency of `@openfeature/server-sdk@~1.18.0` directly conflicts with dd-trace's peer dependency of `@openfeature/server-sdk@~1.19.0`. As a result, running `npm ci`  in node-22 on a project with a package.json listing dd-trace@5.73.0 as a dependency fails with this error:

```
npm error Missing: openapi-types@12.1.3 from lock file
npm error Invalid: lock file's @openfeature/server-sdk@1.18.0 does not satisfy @openfeature/server-sdk@1.19.0
npm error Missing: @openfeature/server-sdk@1.18.0 from lock file

```

dd-trace@5.72.0, does not seem to have this issue.

### Reproduction Code

package.json
```
{
  "name": "demonstration",
  "version": "1.0.0",
  "private": true,
  "main": "dist/index.js",
  "dependencies": {
    "dd-trace": "^5.73.0"
  }
}
```

Run `npm install`
Run `npm ci`

Results in error:
```
npm warn config init.author.name Use `--init-author-name` instead.
npm warn config init.author.email Use `--init-author-email` instead.
npm error code EUSAGE
npm error
npm error `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
npm error
npm error Invalid: lock file's @openfeature/server-sdk@1.18.0 does not satisfy @openfeature/server-sdk@1.19.0
npm error Missing: @openfeature/server-sdk@1.18.0 from lock file
npm error
npm error Clean install a project
npm error
npm error Usage:
npm error npm ci
npm error
npm error Options:
npm error [--install-strategy <hoisted|nested|shallow|linked>] [--legacy-bundling]
[...]

PR #6760, opened 2025-10-27
fix: do not include peer dependencies of a dependency

BridgeAR:
The dependency itself may list it as dependency and we can include the library in our devDependencies for testing purposes.

Fixes: https://github.com/DataDog/dd-trace-js/issues/6743

### DIFF
--- a/packages/datadog-esbuild/index.js
+++ b/packages/datadog-esbuild/index.js
@@ -113,16 +113,6 @@ module.exports.setup = function (build) {
 ${build.initialOptions.banner.js}`
   }
 
-  try {
-    // @openfeature/core is a peer dependency of @openfeature/server-sdk
-    // which is used by @datadog/openfeature-node-server
-    // eslint-disable-next-line n/no-unpublished-require
-    require.resolve('@openfeature/core')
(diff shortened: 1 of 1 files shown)


packages/datadog-esbuild/index.js
```diff
@@ -114,6 +114,9 @@ ${build.initialOptions.banner.js}`
   }
 
   try {
+    // @openfeature/core is a peer dependency of @openfeature/server-sdk
+    // which is used by @datadog/openfeature-node-server
+    // eslint-disable-next-line n/no-unpublished-require
```
BridgeAR: I am uncertain if we should keep this logic or not @uurien.

khanayan123: There was an app sec integration test that uses esbuild that was failing before adding this logic, seems like that test still passes with your changes

uurien (approved):

packages/datadog-esbuild/index.js
uurien: seems that devDependency is good enough for esbuild and also good enough for the package size test. so LGTM

PR #6774, opened 2025-10-28
fix: Restore the try catch in esubild because could fail using yarn

uurien:
### What does this PR do?

Restores the `try catch` removed in a [previous PR] in esbuild plugin to be sure that the build doesn't crash when `@openfeature/core` is not in the dependencies.

Adds an integration test forcing the failure using yarn.

[previous PR]: https://github.com/DataDog/dd-trace-js/pull/6760


BridgeAR (approved):
RSLGTM

PR #6775, opened 2025-10-28
fix missing retry on npm install of wasm-js-rewriter

rochdev:
### What does this PR do?

Fix missing retry on npm install of wasm-js-rewriter.

### Motivation

Installs often fail because of network or registry issues, so we should always wait a bit and retry.

### DIFF
--- a/integration-tests/appsec/iast-esbuild.spec.js
+++ b/integration-tests/appsec/iast-esbuild.spec.js
@@ -2,6 +2,7 @@
 
 const assert = require('node:assert/strict')
 
+const { setTimeout } = require('timers/promises')
 const childProcess = require('child_process')
 const fs = require('fs')
 const path = require('path')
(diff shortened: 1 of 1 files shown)


integration-tests/appsec/iast-esbuild.spec.js
```diff
@@ -12,23 +12,33 @@ const msgpack = require('@msgpack/msgpack')
 const { sandboxCwd, useSandbox, FakeAgent, spawnProc } = require('../helpers')
 
 const exec = promisify(childProcess.exec)
+const retry = async fn => {
+  try {
+    await fn()
+  } catch {
+    await exec('sleep 60')
```
simon-id: process spawn instead of a setTimeout ? damn

rochdev: Ah, yeah it used to be inline, like `npm install && sleep 60 && npm install` and I just split it as-is, I guess I could just use a timer, good catch.

rochdev: Updated!

Issue #6366, opened 2025-09-01
[FEATURE]: Support for cause in error stack


wregis:
### Package Version(s)

5.56.0

### Describe the feature you'd like

When viewing error stack it would be nice to be able to see the complete stack. Currently only the top-level error is shown.

Since [Node.js 16.9](https://nodejs.org/api/errors.html#errorcause) exists the property `cause` for this.

### Additional context

Similar to other languages when an error is partially dealt with and then triggers another error, it is possible to implement such behaviour in Node.js.

PR #6776, opened 2025-10-28
feat: add cause and props to stack output of errors

BridgeAR:
This increases debug information for errors by inspecting the whole error and sending it as stack. That provides a) better output for errors by using util.inspect for the stack instead of the native one. That provides for example better error names. b) This improves the performance for one error in serverless by preventing to create stack frames that are not needed. c) it shows all additional properties that might exist on an error as stack output. That should be helpful for debugging issues.

To limit the output of extra properties to a minimum, only one level is inspected.

Fixes: https://github.com/DataDog/dd-trace-js/issues/6366

### DIFF
--- a/packages/datadog-plugin-http/src/client.js
+++ b/packages/datadog-plugin-http/src/client.js
@@ -1,5 +1,7 @@
 'use strict'
 
+const { URL } = require('url')
+
 const ClientPlugin = require('../../dd-trace/src/plugins/client')
 const { storage } = require('../../datadog-core')
 const tags = require('../../../ext/tags')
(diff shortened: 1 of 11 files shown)


packages/datadog-plugin-fetch/test/index.spec.js
```diff
@@ -226,7 +228,7 @@ describe('Plugin', function () {
           .assertSomeTraces(traces => {
             expect(traces[0][0].meta).to.have.property(ERROR_TYPE, error.name)
             expect(traces[0][0].meta).to.have.property(ERROR_MESSAGE, error.message || error.code)
-            expect(traces[0][0].meta).to.have.property(ERROR_STACK, error.stack)
+            expect(traces[0][0].meta).to.have.property(ERROR_STACK, util.inspect(error, { depth: 0 }))
```
rochdev: What if `error` is not an actual error object?

BridgeAR: See https://github.com/DataDog/dd-trace-js/pull/6776/files#diff-5490d33b84fdf38d10fd4f4460c01af269e82e9a95d9bda16aef99b9353102b7R89-R97

We check for it having a message property or if it is an instance of error to determine if we add the tags.
The object will be serialized as a whole, if it contains a stack property. If it would be a object with a message and a stack property, that would be serialized instead of only showing the stack property. That in itself seems beneficial for users one way or the other. If it does not have a stack property, that is kept as is.

rochdev: If it's not an error we should properly extract the existing stack as it was done before, and only use `util.inspect` for errors since it special cases them to get the stack with additional metadata.

BridgeAR: I discussed it with error tracking and I am going to handle it as separate properties for now.
They said we could have a look how it would be on staging, while it will likely break a couple of existing pretty printing.

So I am marking it as draft for now.

PR #6777, opened 2025-10-28
switch all sandbox usage to new helper

rochdev:
### What does this PR do?

Switch all sandbox usage to new helper.

### Motivation

Managing things like timeouts, retries, etc for the sandbox should not have to be repeated everywhere. This will allow us to make improvements to the sandbox and address flakiness without touching so many files every time.

### DIFF
--- a/integration-tests/esbuild/openfeature.spec.js
+++ b/integration-tests/esbuild/openfeature.spec.js
@@ -3,18 +3,19 @@
 const { execSync } = require('node:child_process')
 const path = require('node:path')
 
-const { FakeAgent, createSandbox } = require('../helpers')
+const { FakeAgent, sandboxCwd, useSandbox } = require('../helpers')
 
 // This should switch to our withVersion helper. The order here currently matters.
 const esbuildVersions = ['latest', '0.16.12']
 
 esbuildVersions.forEach((version) => {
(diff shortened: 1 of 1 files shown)


packages/datadog-plugin-next/test/integration-test/client.spec.js
```diff
@@ -2,38 +2,42 @@
...
       this.timeout(300 * 1000)
-      sandbox = await createSandbox([`'next@${version}'`, 'react@^18.2.0', 'react-dom@^18.2.0'],
-        false, ['./packages/datadog-plugin-next/test/integration-test/*'],
-        'NODE_OPTIONS=--openssl-legacy-provider yarn exec next build')
-      variants = varySandbox(sandbox, 'server.mjs', 'next')
-    })
-
-    after(async () => {
-      await sandbox.remove()
+      execSync('yarn exec next build', {
```
pabloerhard, line 33: Why was this addition needed?

rochdev: It's better to run the command in its own `before` hook so that it's isolated from the sandbox instead of passing it for the sandbox to run. I would like to eventually deprecate passing additional commands to the sandbox.

PR #6780, opened 2025-10-29
test: skip Next.js 16 tests on non-supported Node.js versions

BridgeAR:


packages/dd-trace/test/appsec/extended-data-collection.next.plugin.spec.js
```diff
@@ -36,6 +36,10 @@ describe('extended data collection', () => {
     }
 
     tests.forEach(({ appName, serverPath }) => {
+      if (satisfies(realVersion, '>=16') && NODE_MAJOR < 20) {
```
bengl: Shouldn't this be further up the file, e.g. line 17-20, with the other version checks?

bengl: Or at least right after realVersion is assigned.

BridgeAR: Good point

PR #6782, opened 2025-10-29
[SVLS-7168]Support GCP PubSub Batch Publish

nina9753:
### What does this PR do?
This PR implements comprehensive batch message handling for GCP Pub/Sub, addressing a critical challenge: the Pub/Sub client library automatically batches messages from multiple independent requests, causing async context loss. GCP Pub/Sub has transparent internal batching that happens on the publisher side and internally inside the topic. 

When multiple HTTP requests each call `topic.publishMessage()`, they return immediately to the user. But behind the scenes, the Pub/Sub client queues these messages and N seconds later makes ONE batched gRPC call with ALL the messages together. So by the time the batched gRPC call happens, the original HTTP requests have completed and their async contexts are gone. We're left with either no active span or the wrong span active, leading to miss-parented spans. If we only instrumented at the gRPC level, we'd have no way to determine which trace each message came from.

To handle this we introduced dual-wrapping:
High-Level Wrapping (`Topic.publishMessage / Topic.publish`):
- Wraps the user-facing Topic API that developers call
- Captures trace context immediately when the user calls the API (while their request span is still active)
- Injects this context into message attributes before batching occurs
- Similar to how we handle message.ack() - capture context early before batching

Low-Level Wrapping (`PublisherClient.publish`):
- Wraps the internal gRPC client call that actually sends to Pub/Sub
- Creates the pubsub.request span representing the network operation
- Reads trace contexts from message attributes (stored by high-level wrap)
- Implements span linking when multiple traces are batched together

Producer Plugin Enhancements
- The producer plugin now handles batched messages intelligently:
- Extracts the parent context from the first message's attributes
**- Creates span links to all other messages in the batch (messages 2-N)**
- Injects batch metadata into all messages: batch size, position, and batch span IDs
- Adds proper span link metadata for APM UI visualization
- Supports 128-bit trace IDs for proper linking

Next PR in the batch is #6415

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes
An inferred span for the push subscription HTTP POST request to the Cloud Run service from a pub/sub topic 
[Example full Push Distributed Trace](https://app.datadoghq.com/apm/trace/5857421563878271202?spanID=8398277051735365684&timeHint=1763576006523.0127&trace=58574215638782712026349771102019711151) of a cloud run service triggering another service using a push subscription 

image
image

### DIFF
--- a/packages/datadog-instrumentations/src/google-cloud-pubsub.js
+++ b/packages/datadog-instrumentations/src/google-cloud-pubsub.js
@@ -30,7 +30,25 @@ const receiveStartCh = channel('apm:google-cloud-pubsub:receive:start')
 const receiveFinishCh = channel('apm:google-cloud-pubsub:receive:finish')
 const receiveErrorCh = channel('apm:google-cloud-pubsub:receive:error')
 
+// Bounded map to prevent memory leaks from acks that never complete
 const ackContextMap = new Map()
+const ACK_CONTEXT_MAX_SIZE = 10_000
+const ACK_CONTEXT_TTL_MS = 600_000 // 10 minutes - matches Cloud Run streaming pull default deadline
+
+// Cleanup old entries periodically
+const ackContextCleanupInterval = setInterval(() => {
+  const now = Date.now()
+  for (const [ackId, entry] of ackContextMap.entries()) {
+    if (now - entry.timestamp > ACK_CONTEXT_TTL_MS) {
+      ackContextMap.delete(ackId)
+    }
+  }
+}, 60_000) // Run cleanup every 60 seconds
+
+// Allow process to exit cleanly
+if (ackContextCleanupInterval.unref) {
+  ackContextCleanupInterval.unref()
+}
 
 const publisherMethods = [
   'createTopic',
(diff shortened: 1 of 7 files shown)


packages/datadog-instrumentations/src/google-cloud-pubsub.js
```diff
@@ -133,6 +197,29 @@ addHook({ name: '@google-cloud/pubsub', versions: ['>=1.2'] }, (obj) => {
   return obj
 })
 
+addHook({ name: '@google-cloud/pubsub', versions: ['>=1.2'], file: 'build/src/subscriber.js' }, (obj) => {
+  const Message = obj.Message
+
+  if (Message && Message.prototype && Message.prototype.ack) {
```
tlhunter:
```suggestion
  if (Message?.prototype?.ack) {
```

packages/datadog-plugin-google-cloud-pubsub/src/producer.js
```diff
@@ -9,36 +9,141 @@ class GoogleCloudPubsubProducerPlugin extends ProducerPlugin {
...
+
+  bindError (ctx) {
+    if (ctx.error && ctx.batchSpan) {
+      ctx.batchSpan.setTag('error', ctx.error)
+      ctx.batchSpan.finish()
+    }
+    return ctx.parentStore
+  }
+
+  _extractSpanLink (attrs) {
```
tlhunter, line 119: https://github.com/DataDog/dd-trace-js/pull/6415/files#diff-32e83f63e1f3fd1938530fecdf21e77e942496aee9272d001accbb758a7fc935R119

I'm getting a little confused. Do these two PRs repeat functionality?

nina9753: I just needed to rebase the base branch onto the next child PR. you should only see these changes here now

PR #6783, opened 2025-10-29
APMAPI-1306 feat(otel): add support for otel metrics api via protobuf and json

mabdinur:
### What does this PR do?

Adds full OpenTelemetry Metrics support to dd-trace-js with a custom Meter Provider implementation. Enable with `DD_METRICS_OTEL_ENABLED=true` to export metrics via OTLP protocol.

**Key Features:**
- Full OpenTelemetry Metrics API compliance with all standard instrument types (Counter, UpDownCounter, Histogram, Gauge, and Observable variants)
- OTLP export via `http/protobuf` (default) or `http/json` protocols
- Configurable endpoint, headers, timeout, and export intervals via standard `OTEL_EXPORTER_OTLP_METRICS_*` environment variables
- Periodic metric collection and aggregation with support for DELTA, CUMULATIVE, and LOWMEMORY temporality modes
- Comprehensive test coverage (91.69%) with 39 integration tests

**Configuration:**
- `DD_METRICS_OTEL_ENABLED` - Enable OpenTelemetry metrics (default: `false`)
- `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` - Endpoint URL (default: `http://localhost:4318/v1/metrics`)
- `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` - Protocol: `http/protobuf` or `http/json`
- `OTEL_METRIC_EXPORT_INTERVAL` - Export interval in ms (default: `60000`)
- Additional timeout and header configuration options

### Motivation

Enables customers to use OpenTelemetry Metrics API with dd-trace-js without adding the OpenTelemetry SDK as a dependency. Custom implementation provides better integration with dd-trace-js configurations, avoids vendoring grpc libraries and maintains flexibility.

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
+++ b/packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
@@ -10,7 +10,7 @@ const { stableStringify } = require('../otlp/otlp_transformer_base')
  * @typedef {import('@opentelemetry/api').Attributes} Attributes
  * @typedef {import('@opentelemetry/core').InstrumentationScope} InstrumentationScope
  * @typedef {import('./instruments').Measurement} Measurement
- **/
+ */
 
 /**
  * @typedef {Object} NumberDataPoint
@@ -19,7 +19,7 @@ const { stableStringify } = require('../otlp/otlp_transformer_base')
  * @property {number} timeUnixNano - Timestamp in nanoseconds
  * @property {number} startTimeUnixNano - Start timestamp for cumulative metrics
  * @property {number} value - Metric value
- **/
+ */
 
 /**
  * @typedef {Object} HistogramDataPoint
@@ -33,7 +33,7 @@ const { stableStringify } = require('../otlp/otlp_transformer_base')
  * @property {number} max - Maximum value observed
  * @property {number[]} bucketCounts - Count per histogram bucket
  * @property {number[]} explicitBounds - Histogram bucket boundaries
- **/
+ */
 
 /**
  * @typedef {Object} AggregatedMetricDataPoint
@@ -47,7 +47,7 @@ const { stableStringify } = require('../otlp/otlp_transformer_base')
  * @property {number} max - Maximum value observed
  * @property {number[]} bucketCounts - Count per histogram bucket
  * @property {number[]} explicitBounds - Histogram bucket boundaries
- **/
+ */
 
 /**
  * @typedef {Object} AggregatedMetric
@@ -58,7 +58,7 @@ const { stableStringify } = require('../otlp/otlp_transformer_base')
  * @property {InstrumentationScope} instrumentationScope - Instrumentation scope
  * @property {string} temporality - Temporality from TEMPORALITY constants
  * @property {Map<string, AggregatedMetricDataPoint>} dataPointMap - Map of attribute keys to data points
- **/
+ */
 
 /**
  * PeriodicMetricReader collects and exports metrics at a regular interval.


packages/dd-trace/src/config.js
```diff
@@ -659,6 +667,26 @@ class Config {
...
+    this._setString(env, 'otelMetricsHeaders', OTEL_EXPORTER_OTLP_METRICS_HEADERS || env.otelHeaders)
+    this._setString(env, 'otelMetricsProtocol', OTEL_EXPORTER_OTLP_METRICS_PROTOCOL || env.otelProtocol)
+    env.otelMetricsTimeout = maybeInt(OTEL_EXPORTER_OTLP_METRICS_TIMEOUT) || env.otelTimeout
+    env.otelMetricsExportTimeout = maybeInt(OTEL_METRIC_EXPORT_TIMEOUT)
+    env.otelMetricsExportInterval = maybeInt(OTEL_METRIC_EXPORT_INTERVAL)
+    // Parse temporality preference (default to DELTA for Datadog)
+    if (OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE) {
+      const temporalityPref = OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE.toUpperCase()
+      if (['DELTA', 'CUMULATIVE', 'LOWMEMORY'].includes(temporalityPref)) {
+        this._setString(env, 'otelMetricsTemporalityPreference', temporalityPref)
```
mabdinur: Warn if CUMULATIVE temporality is used

packages/dd-trace/src/config.js
```diff
@@ -1177,9 +1205,10 @@ class Config {
 
     calc['dogstatsd.hostname'] = this._getHostname()
 
-    // Compute OTLP logs URL to send payloads to the active Datadog Agent
+    // Compute OTLP logs and metrics URLs to send payloads to the active Datadog Agent
     const agentHostname = this._getHostname()
     calc.otelLogsUrl = `http://${agentHostname}:${DEFAULT_OTLP_PORT}`
+    calc.otelMetricsUrl = `http://${agentHostname}:${DEFAULT_OTLP_PORT}/v1/metrics`
```
mabdinur, line 1246:
```suggestion
    calc.otelMetricsUrl = `http://${agentHostname}:${DEFAULT_OTLP_PORT}`
```

packages/dd-trace/src/opentelemetry/metrics/index.js
```diff
@@ -0,0 +1,88 @@
...
+ */
+function initializeOpenTelemetryMetrics (config) {
+  // Build resource attributes
+  const resourceAttributes = {
+    'service.name': config.service,
+    'service.version': config.version,
+    'deployment.environment': config.env
+  }
+
+  // Add all tracer tags (includes DD_TAGS, OTEL_RESOURCE_ATTRIBUTES, DD_TRACE_TAGS, etc.)
```
mabdinur: get rid of useless comments

packages/dd-trace/src/opentelemetry/metrics/meter.js
```diff
@@ -0,0 +1,181 @@
...
+      version: instrumentationScope?.version || packageVersion,
+      schemaUrl: instrumentationScope?.schemaUrl || '',
+      attributes: instrumentationScope?.attributes || {}
+    }
+    this.#instruments = new Map()
+  }
+
+  #getOrCreateInstrument (name, type, InstrumentClass, options = {}) {
+    const normalizedName = name.toLowerCase()
+    const key = `${type}:${normalizedName}`
```
mabdinur, line 62: Look into \0

packages/dd-trace/src/opentelemetry/metrics/meter_provider.js
```diff
@@ -0,0 +1,120 @@
...
+    if (this.isShutdown) {
+      return this.#createNoOpMeter()
+    }
+
+    const normalizedName = name.toLowerCase()
+    const meterVersion = version || ''
+    const schemaUrl = options?.schemaUrl || ''
+    const attributes = options?.attributes || {}
+    const attrsKey = JSON.stringify(attributes)
+    const key = `${normalizedName}@${meterVersion}@${schemaUrl}@${attrsKey}`
```
mabdinur: \0

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,357 @@
...
+      if (type === METRIC_TYPES.HISTOGRAM) {
+        this.#aggregateHistogram(metric, value, attributes, attrKey, timestamp, stateKey, cumulativeState)
+      } else if (type === METRIC_TYPES.GAUGE ||
+                 type === METRIC_TYPES.OBSERVABLECOUNTER ||
+                 type === METRIC_TYPES.OBSERVABLEUPDOWNCOUNTER) {
+        // Gauges and observable instruments use last value (observations report current total, not increments)
+        this.#aggregateLastValue(metric, value, attributes, attrKey, timestamp)
+      } else {
+        // Synchronous Counters and UpDownCounters use sum aggregation
+        this.#aggregateSum(metric, value, attributes, attrKey, timestamp, stateKey, cumulativeState)
```
mabdinur: Move this up.

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,357 @@
...
+    const dataPoint = this.#findOrCreateDataPoint(metric, attributes, attrKey, () => ({
+      timeUnixNano: timestamp,
+      value: 0
+    }))
+
+    dataPoint.value = value
+    dataPoint.timeUnixNano = timestamp
+  }
+
+  #aggregateHistogram (metric, value, attributes, attrKey, timestamp, stateKey, cumulativeState) {
```
mabdinur, line 484: https://opentelemetry.io/docs/specs/otel/metrics/data-model/#histogram

packages/dd-trace/src/opentelemetry/metrics/otlp_transformer.js
```diff
@@ -0,0 +1,277 @@
...
+  transformMetrics (metrics) {
+    if (this.protocol === 'http/json') {
+      return this.#transformToJson(metrics)
+    }
+    return this.#transformToProtobuf(metrics)
+  }
+
+  /**
+   * Transforms metrics to protobuf format.
+   * @param {Array} metrics - Array of metrics to transform
```
mabdinur: Better define these types

docs/API.md
```diff
@@ -417,6 +417,64 @@ The Datadog SDK supports many of the configurations supported by the OpenTelemet
...
+- `OTEL_EXPORTER_OTLP_METRICS_HEADERS` - Optional headers in JSON format for metrics (default: `{}`)
+- `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` - OTLP protocol for metrics (default: `http/protobuf`)
+- `OTEL_EXPORTER_OTLP_METRICS_TIMEOUT` - Request timeout in milliseconds for metrics (default: `10000`)
+- `OTEL_METRIC_EXPORT_INTERVAL` - Metric export interval in milliseconds (default: `10000`)
+- `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` - Temporality preference (default: `DELTA`)
+- `OTEL_METRIC_EXPORT_TIMEOUT` - [NOT YET SUPPORTED] Time to export metrics including retries.
+
+Metrics are collected periodically and exported via OTLP over HTTP. The protocol can be configured using `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` or `OTEL_EXPORTER_OTLP_PROTOCOL` environment variables. Supported protocols are `http/protobuf` (default) and `http/json`. All metrics use delta aggregation temporality to match Datadog's data model. For complete OTLP exporter configuration options, see the [OpenTelemetry OTLP Exporter documentation](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/).
+
```
BridgeAR: The text is totally fine, I think it would just be more straight forward in case the individual configuration has all entries listed right away instead of having a separate section with a longer text that describes that.
I would therefore just inline this content into the above variables besides the parts that apply across multiple envs.

packages/dd-trace/src/config.js
```diff
@@ -606,6 +614,26 @@ class Config {
     target.otelLogsTimeout = maybeInt(OTEL_EXPORTER_OTLP_LOGS_TIMEOUT) || target.otelTimeout
     target.otelLogsBatchTimeout = maybeInt(OTEL_BSP_SCHEDULE_DELAY)
     target.otelLogsMaxExportBatchSize = maybeInt(OTEL_BSP_MAX_EXPORT_BATCH_SIZE)
+
+    const otelMetricsExporter = String(OTEL_METRICS_EXPORTER).toLowerCase() !== 'none'
```
BridgeAR: Nit

```suggestion
    const otelMetricsExporter = !OTEL_METRICS_EXPORTER || OTEL_METRICS_EXPORTER.toLowerCase() !== 'none'
```

packages/dd-trace/src/config.js
```diff
@@ -606,6 +614,26 @@ class Config {
...
+    this.#setBoolean(target, 'otelMetricsEnabled', DD_METRICS_OTEL_ENABLED && otelMetricsExporter)
```
BridgeAR, line 629: Do we have some documentation about the none exporter deactivating the metrics?

packages/dd-trace/src/config.js
```diff
@@ -606,6 +614,26 @@ class Config {
...
+    // Set OpenTelemetry metrics configuration with specific _METRICS_ vars
+    // taking precedence over generic _EXPORTERS_ vars
+    if (OTEL_EXPORTER_OTLP_ENDPOINT || OTEL_EXPORTER_OTLP_METRICS_ENDPOINT) {
+      this.#setString(target, 'otelMetricsUrl', OTEL_EXPORTER_OTLP_METRICS_ENDPOINT || target.otelUrl)
+    }
+    this.#setString(target, 'otelMetricsHeaders', OTEL_EXPORTER_OTLP_METRICS_HEADERS || target.otelHeaders)
+    this.#setString(target, 'otelMetricsProtocol', OTEL_EXPORTER_OTLP_METRICS_PROTOCOL || target.otelProtocol)
+    target.otelMetricsTimeout = maybeInt(OTEL_EXPORTER_OTLP_METRICS_TIMEOUT) || target.otelTimeout
+    target.otelMetricsExportTimeout = maybeInt(OTEL_METRIC_EXPORT_TIMEOUT)
+    target.otelMetricsExportInterval = maybeInt(OTEL_METRIC_EXPORT_INTERVAL)
```
BridgeAR: Is zero allowed for any of these values?

packages/dd-trace/src/config.js
```diff
@@ -606,6 +614,26 @@ class Config {
...
     target.otelLogsMaxExportBatchSize = maybeInt(OTEL_BSP_MAX_EXPORT_BATCH_SIZE)
+
+    const otelMetricsExporter = String(OTEL_METRICS_EXPORTER).toLowerCase() !== 'none'
+    this.#setBoolean(target, 'otelMetricsEnabled', DD_METRICS_OTEL_ENABLED && otelMetricsExporter)
+    // Set OpenTelemetry metrics configuration with specific _METRICS_ vars
+    // taking precedence over generic _EXPORTERS_ vars
+    if (OTEL_EXPORTER_OTLP_ENDPOINT || OTEL_EXPORTER_OTLP_METRICS_ENDPOINT) {
+      this.#setString(target, 'otelMetricsUrl', OTEL_EXPORTER_OTLP_METRICS_ENDPOINT || target.otelUrl)
+    }
+    this.#setString(target, 'otelMetricsHeaders', OTEL_EXPORTER_OTLP_METRICS_HEADERS || target.otelHeaders)
```
BridgeAR, line 635: Note: these will mess with telemetry values for now, while that is an issue in lots of places and it will be resolved in another PR where we fix the telemetry (the issue is that the property will be defined by either another property or the env and that can not be differentiated for the telemetry being defined like that).

packages/dd-trace/src/config_defaults.js
```diff
@@ -140,6 +140,14 @@ module.exports = {
...
   otelLogsBatchTimeout: 5000,
   otelLogsMaxExportBatchSize: 512,
+  otelMetricsEnabled: false,
+  otelMetricsUrl: undefined, // Will be computed using agent host
+  otelMetricsHeaders: '',
+  otelMetricsProtocol: 'http/protobuf',
+  otelMetricsTimeout: 10_000,
+  otelMetricsExportTimeout: 7500,
+  otelMetricsExportInterval: 10_000,
+  otelMetricsTemporalityPreference: 'DELTA', // DELTA, CUMULATIVE, or LOWMEMORY
```
BridgeAR, line 151: These values are not documented as far as I know. Should we list them in the API.md?

packages/dd-trace/src/opentelemetry/metrics/instruments.js
```diff
@@ -0,0 +1,147 @@
...
+ * https://open-telemetry.github.io/opentelemetry-js/interfaces/_opentelemetry_api._opentelemetry_api.Instrument.html
+ * @private
+ */
+class Instrument {
+  #baseMetadata
+
+  constructor (name, options, instrumentationScope, reader) {
+    this.name = name
+    this.description = options.description || ''
+    this.unit = options.unit || ''
```
BridgeAR: Nit

```suggestion
    this.description = options.description ?? ''
    this.unit = options.unit ?? ''
```

packages/dd-trace/src/opentelemetry/metrics/instruments.js
```diff
@@ -0,0 +1,147 @@
...
+      name: this.name,
+      description: this.description,
+      unit: this.unit,
+      instrumentationScope: this.instrumentationScope
+    }
+  }
+
+  createMeasurement (type, value, attributes) {
+    return {
+      ...this.#baseMetadata,
```
BridgeAR: It should actually be faster or as fast doing the assignment directly.

```suggestion
  }

  createMeasurement (type, value, attributes) {
    return {
      name: this.name,
      description: this.description,
      unit: this.unit,
      instrumentationScope: this.instrumentationScope,
```

packages/dd-trace/src/opentelemetry/metrics/instruments.js
```diff
@@ -0,0 +1,147 @@
...
+      value,
+      attributes: sanitizeAttributes(attributes),
+      timestamp: Number(process.hrtime.bigint())
+    }
+  }
+}
+
+class Counter extends Instrument {
+  add (value, attributes = {}) {
+    if (value < 0) return
```
BridgeAR, line 77: What if the value is not a number or NaN? Who handles this and where?

That question also applies to the other parts here.

If it is handled somewhere else: would that part not also be able to handle skipping negative values?

Update: I read https://opentelemetry.io/docs/specs/otel/metrics/api/#add and I believe we should actually not check this at all? While I would not know what happens if a negative or invalid value would be passed through.

I am very unhappy with this specification since it seems to be written for strictly typed languages.

packages/dd-trace/src/opentelemetry/metrics/instruments.js
```diff
@@ -0,0 +1,147 @@
...
+ * Base class for observable (asynchronous) instruments.
+ * @private
+ */
+class ObservableInstrument extends Instrument {
+  #callbacks
+  #type
+
+  constructor (name, options, instrumentationScope, reader, type) {
+    super(name, options, instrumentationScope, reader)
+    this.#callbacks = []
```
BridgeAR: Nit

```suggestion
  #callbacks = []
  #type

  constructor (name, options, instrumentationScope, reader, type) {
    super(name, options, instrumentationScope, reader)
```

packages/dd-trace/src/opentelemetry/metrics/instruments.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  addCallback (callback) {
+    if (typeof callback !== 'function') return
+    this.#callbacks.push(callback)
+    this.reader?.registerObservableInstrument(this)
+  }
+
+  removeCallback (callback) {
+    const index = this.#callbacks.indexOf(callback)
```
BridgeAR, line 148: Is it likely to have more than 4 callbacks? If yes, let's use a Set. Otherwise, let us keep the array.

packages/dd-trace/src/opentelemetry/metrics/instruments.js
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * @typedef {import('@opentelemetry/api').Attributes} Attributes
+ * @typedef {import('@opentelemetry/core').InstrumentationScope} InstrumentationScope
+ */
+
+/**
+ * Base class for all metric instruments.
+ *
+ * This implementation follows the OpenTelemetry JavaScript API Instrument interface:
+ * https://open-telemetry.github.io/opentelemetry-js/interfaces/_opentelemetry_api._opentelemetry_api.Instrument.html
```
BridgeAR: This returns a 404

packages/dd-trace/src/opentelemetry/metrics/meter.js
```diff
@@ -0,0 +1,181 @@
+'use strict'
+
+const packageVersion = require('../../../../../package.json').version
```
BridgeAR: Nit

We actually already read that version in `version.js` on the root. Let's read that file, since it could be that this is called when not being cached while the other should definitely be loaded faster.

packages/dd-trace/src/opentelemetry/metrics/meter.js
```diff
@@ -0,0 +1,181 @@
...
+   */
+  constructor (meterProvider, instrumentationScope) {
+    this.meterProvider = meterProvider
+    this.#instrumentationScope = {
+      name: instrumentationScope?.name || 'dd-trace-js',
+      version: instrumentationScope?.version || packageVersion,
+      schemaUrl: instrumentationScope?.schemaUrl || '',
+      attributes: instrumentationScope?.attributes || {}
+    }
+    this.#instruments = new Map()
```
BridgeAR: Nit

```suggestion
  #instruments = new Map()

  /**
   * Creates a new Meter instance.
   *
   * @param {MeterProvider} meterProvider - Parent meter provider
   * @param {InstrumentationScope} instrumentationScope - Instrumentation scope information
   * @param {string} [instrumentationScope.name] - Meter name (defaults to 'dd-trace-js')
   * @param {string} [instrumentationScope.version] - Meter version (defaults to tracer version)
   * @param {string} [instrumentationScope.schemaUrl] - Schema URL
   * @param {Object} [instrumentationScope.attributes] - Attributes for the instrumentation scope
   */
  constructor (meterProvider, instrumentationScope) {
    this.meterProvider = meterProvider
    this.#instrumentationScope = {
      name: instrumentationScope?.name || 'dd-trace-js',
      version: instrumentationScope?.version || packageVersion,
      schemaUrl: instrumentationScope?.schemaUrl || '',
      attributes: instrumentationScope?.attributes || {}
    }
```

packages/dd-trace/src/opentelemetry/metrics/meter.js
```diff
@@ -0,0 +1,181 @@
...
+   * @param {string} [instrumentationScope.schemaUrl] - Schema URL
+   * @param {Object} [instrumentationScope.attributes] - Attributes for the instrumentation scope
+   */
+  constructor (meterProvider, instrumentationScope) {
+    this.meterProvider = meterProvider
+    this.#instrumentationScope = {
+      name: instrumentationScope?.name || 'dd-trace-js',
+      version: instrumentationScope?.version || packageVersion,
+      schemaUrl: instrumentationScope?.schemaUrl || '',
+      attributes: instrumentationScope?.attributes || {}
```
BridgeAR: Nit

```suggestion
  constructor (meterProvider, { name = 'dd-trace-js', version = packageVersion, schemaUrl = '', attributes = {} } = {}) {
    this.meterProvider = meterProvider
    this.#instrumentationScope = {
      name,
      version,
      schemaUrl,
      attributes,
```

packages/dd-trace/src/opentelemetry/metrics/meter.js
```diff
@@ -0,0 +1,181 @@
...
+  #getOrCreateInstrument (name, type, InstrumentClass, options = {}) {
+    const normalizedName = name.toLowerCase()
+    const key = `${type}:${normalizedName}`
+    if (!this.#instruments.has(key)) {
+      const instrument = new InstrumentClass(
+        normalizedName, options, this.#instrumentationScope, this.meterProvider.reader
+      )
+      this.#instruments.set(key, instrument)
+    }
+    return this.#instruments.get(key)
```
BridgeAR:
```suggestion
    let instrument = this.#instruments.get(key)
    if (!instrument) {
      instrument = new InstrumentClass(
        normalizedName, options, this.#instrumentationScope, this.meterProvider.reader
      )
      this.#instruments.set(key, instrument)
    }
    return instrument
```

packages/dd-trace/src/opentelemetry/metrics/meter.js
```diff
@@ -0,0 +1,181 @@
...
+    this.#instrumentationScope = {
+      name: instrumentationScope?.name || 'dd-trace-js',
+      version: instrumentationScope?.version || packageVersion,
+      schemaUrl: instrumentationScope?.schemaUrl || '',
+      attributes: instrumentationScope?.attributes || {}
+    }
+    this.#instruments = new Map()
+  }
+
+  #getOrCreateInstrument (name, type, InstrumentClass, options = {}) {
```
BridgeAR:
```suggestion
  #getOrCreateInstrument (name, type, InstrumentClass, options) {
```

This is defined in all call sites. It could also be changed to remove the default value in all call sites instead.

BridgeAR: Could you also add JSDoc for the private methods? :)

packages/dd-trace/src/opentelemetry/metrics/instruments.js
```diff
@@ -0,0 +1,147 @@
...
+  #callbacks
+  #type
+
+  constructor (name, options, instrumentationScope, reader, type) {
+    super(name, options, instrumentationScope, reader)
+    this.#callbacks = []
+    this.#type = type
+  }
+
+  addCallback (callback) {
```
BridgeAR, line 136: Should these also have JSDoc?

packages/dd-trace/src/opentelemetry/metrics/meter_provider.js
```diff
@@ -0,0 +1,121 @@
...
+   * Creates a new MeterProvider instance with a single reader for Datadog Agent export.
+   *
+   * @param {Object} [options] - MeterProvider options
+   * @param {PeriodicMetricReader} [options.reader] - Single MetricReader instance for
+   *   exporting metrics to Datadog Agent
+   */
+  constructor (options = {}) {
+    this.reader = options.reader
+    this.#meters = new Map()
+    this.#contextManager = new ContextManager()
```
BridgeAR:
```suggestion
  #meters = new Map()
  #contextManager = new ContextManager()

  /**
   * Creates a new MeterProvider instance with a single reader for Datadog Agent export.
   *
   * @param {Object} [options] - MeterProvider options
   * @param {PeriodicMetricReader} [options.reader] - Single MetricReader instance for
   *   exporting metrics to Datadog Agent
   */
  constructor (options = {}) {
    this.reader = options.reader
```

packages/dd-trace/src/opentelemetry/metrics/meter_provider.js
```diff
@@ -0,0 +1,121 @@
...
+   */
+  getMeter (name, version = '', options = {}) {
+    if (this.isShutdown) {
+      return this.#createNoOpMeter()
+    }
+
+    const normalizedName = name.toLowerCase()
+    const meterVersion = version || ''
+    const schemaUrl = options?.schemaUrl || ''
+    const attributes = options?.attributes || {}
```
BridgeAR:
```suggestion
  getMeter (name, version = '', { schemaUrl = '', attributes = {} } = {}) {
    if (this.isShutdown) {
      return this.#createNoOpMeter()
    }

    const normalizedName = name.toLowerCase()
```

packages/dd-trace/src/opentelemetry/metrics/meter_provider.js
```diff
@@ -0,0 +1,121 @@
...
+    const attrsKey = JSON.stringify(attributes)
+    const key = `${normalizedName}@${meterVersion}@${schemaUrl}@${attrsKey}`
+
+    if (!this.#meters.has(key)) {
+      this.#meters.set(key, new Meter(this, { name: normalizedName, version: meterVersion, schemaUrl, attributes }))
+    }
+    return this.#meters.get(key)
```
BridgeAR:
```suggestion
    const key = `${normalizedName}@${version}@${schemaUrl}@${attrsKey}`

    let meter = this.#meters.get(key)
    if (!meter) {
      meter = new Meter(this, { name: normalizedName, version, schemaUrl, attributes })
      this.#meters.set(key, meter)
    }
    return meter
```

packages/dd-trace/src/opentelemetry/otlp/otlp_http_exporter_base.js
```diff
@@ -49,30 +47,21 @@ class OtlpHttpExporterBase {
...
-   */
-  _getTelemetryTags () {
-    return this.#telemetryTags
-  }
-
   /**
    * Records telemetry metrics for exported data.
    * @param {string} metricName - Name of the metric to record
    * @param {number} count - Count to increment
    * @param {Array<string>} [tags] - Optional custom tags (defaults to this exporter's tags)
```
BridgeAR: As far as I can tell this is only used by us internally. As such, I would change that to: additional tags next to the exporters default tags. That way there is no need to expose the default ones and just join them here.

packages/dd-trace/src/opentelemetry/metrics/otlp_transformer.js
```diff
@@ -0,0 +1,280 @@
...
+   * @param {boolean} isJson - Whether to format for JSON output
+   * @returns {Object[]} Array of scope metric objects
+   * @private
+   */
+  #transformScope (metrics, isJson = false) {
+    const groupedMetrics = this.groupByInstrumentationScope(metrics)
+    const scopeMetrics = []
+
+    for (const [, metricsInScope] of groupedMetrics) {
```
BridgeAR:
```suggestion
    for (const metricsInScope of groupedMetrics.values()) {
```

packages/dd-trace/src/opentelemetry/metrics/otlp_transformer.js
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    return scopeMetrics
+  }
+
+  /**
+   * Transforms a single metric to protobuf format.
+   * @private
+   */
+  #transformMetric (metric) {
```
BridgeAR: Could we define an interface for metric?

That also applies to all other methods where we have parameters.

packages/dd-trace/src/opentelemetry/metrics/otlp_transformer.js
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    return result
+  }
+
+  /**
+   * Transforms a single metric to JSON format.
+   * @private
+   */
+  #transformMetricToJson (metric) {
```
BridgeAR: This method could be combined with the transformMetric method. I would just add another argument to the method and there seem to be three different spots where different methods must be called depending on a boolean argument about it being a json or not (one of these would a new method that abstracts the part for METRIC_TYPES.HISTOGRAM).

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,353 @@
...
+   */
+  constructor (exporter, exportInterval, temporalityPreference = 'DELTA') {
+    this.exporter = exporter
+    this.#exportInterval = exportInterval
+    this.#measurements = []
+    this.#observableInstruments = []
+    this.#aggregator = new MetricAggregator(temporalityPreference)
+    this.#timer = null
+    this.#cumulativeState = new Map() // Tracks cumulative values across export cycles
+    this.#lastExportedState = new Map() // Tracks last exported values for delta calculation
```
BridgeAR: Let's move the defaults up to the property assignment where possible :)

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,353 @@
...
+    this.#measurements.push(measurement)
+  }
+
+  /**
+   * Registers an observable instrument for periodic collection.
+   *
+   * @param {ObservableGauge} instrument - The observable instrument to register
+   */
+  registerObservableInstrument (instrument) {
+    if (!this.#observableInstruments.includes(instrument)) {
```
BridgeAR: Are we likely going to have more than 3-4 instruments in a couple of cases? If yes, let's use a set instead.

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,353 @@
...
+
+  #clearTimer () {
+    if (this.#timer) {
+      clearInterval(this.#timer)
+      this.#timer = null
+    }
+  }
+
+  #collectAndExport (callback = () => {}) {
+    const allMeasurements = this.#measurements.splice(0)
```
BridgeAR, line 167: Nit

```suggestion
    const allMeasurements = this.#measurements.slice()
```

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,353 @@
...
+
+    for (const instrument of this.#observableInstruments) {
+      allMeasurements.push(...instrument.collect())
```
BridgeAR: How many entries could we collect in an extreme case?
Could it exceed the number of arguments being able to be handled? About 64k

packages/dd-trace/src/opentelemetry/metrics/instruments.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  removeCallback (callback) {
+    const index = this.#callbacks.indexOf(callback)
+    if (index !== -1) {
+      this.#callbacks.splice(index, 1)
+    }
+  }
+
+  collect () {
```
BridgeAR, line 163: Is this method over called directly by a user or is that only being used by us internally?

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,353 @@
...
+ * MetricAggregator aggregates individual measurements into metric data points.
+ * @private
+ */
+class MetricAggregator {
+  #temporalityPreference
+  #startTime
+
+  constructor (temporalityPreference = TEMPORALITY.DELTA) {
+    this.#temporalityPreference = temporalityPreference
+    this.#startTime = Number(process.hrtime.bigint())
```
BridgeAR:
```suggestion
  #startTime = Number(process.hrtime.bigint())

  constructor (temporalityPreference = TEMPORALITY.DELTA) {
    this.#temporalityPreference = temporalityPreference
```

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,353 @@
...
+          unit,
+          type,
+          instrumentationScope,
+          temporality: this.#getTemporality(type),
+          data: [],
+          dataPointMap: new Map()
+        })
+      }
+
+      const metric = metricsMap.get(metricKey)
```
BridgeAR:
```suggestion
      let metric = metricsMap.get(metricKey)
      if (!metric) {
        metric = {
          name,
          description,
          unit,
          type,
          instrumentationScope,
          temporality: this.#getTemporality(type),
          data: [],
          dataPointMap: new Map()
        }
        metricsMap.set(metricKey, metric)
      }
```

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,353 @@
...
+      } else {
+        this.#aggregateLastValue(metric, value, attributes, attrKey, timestamp)
+      }
+    }
+
+    const metrics = [...metricsMap.values()]
+
+    this.#applyDeltaTemporality(metrics, lastExportedState)
+
+    return metrics
```
BridgeAR: Is this a public return type or is it only used internally?

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,353 @@
...
+        type,
+        instrumentationScope,
+        value,
+        attributes,
+        timestamp
+      } = measurement
+
+      const scopeKey = this.#getScopeKey(instrumentationScope)
+      const metricKey = `${scopeKey}:${name}:${type}`
+      const attrKey = JSON.stringify(attributes)
```
BridgeAR: Using the result of JSON.stringify on an object as key is not stable.

E.g., { a: 1, b: 2 } is not equal to { b: 2, a: 1 }

You can use a library such as https://www.npmjs.com/package/safe-stable-stringify

This should also be checked in other parts of the code where we do similar things.

Please also add test cases for that.

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,353 @@
...
+            dataPoint.sum = currentState.sum - lastState.sum
+            dataPoint.bucketCounts = currentState.bucketCounts.map(
+              (count, idx) => count - (lastState.bucketCounts[idx] || 0)
+            )
+            lastExportedState.set(stateKey, currentState)
+          }
+        }
+      }
+
+      delete metric.dataPointMap
```
BridgeAR: I am uncertain why that property is removed. Could you add a comment? I would also likely just set it to undefined anyway.

packages/dd-trace/src/config.js
mabdinur: Ahh good question. For timeouts `0` corresponds to no timeout (which is handled by http.request). For intervals and batch sizes zero is not allowed and should be ignored. I updated the code to handle this case.

packages/dd-trace/src/config_defaults.js
mabdinur: The config is documented in api docs but it's very high level. I will update the api docs to include this link: https://opentelemetry.io/docs/specs/otel/metrics/sdk_exporters/otlp/#additional-environment-variable-configuration

packages/dd-trace/src/opentelemetry/metrics/instruments.js
mabdinur: Yup, I missed this. Instrument is not a type defined by otel. It only defines metric types (ex: Counter, Gauges, Histrograms, etc.). I fixed the doc strings. For the instruments. They are now all defined.

packages/dd-trace/src/opentelemetry/metrics/meter.js
mabdinur: I changed this line to:

```
const { version } = require('../../../../../package.json')
```

packages/dd-trace/src/config.js
mabdinur: It's documented here but we can definetly do a better job here: https://docs.datadoghq.com/opentelemetry/config/environment_variable_support/ 😅

packages/dd-trace/src/opentelemetry/metrics/instruments.js
mabdinur: We should keep it an array. For the majority of cases there should be one callback per metric (ex: a callback that records the number of active threads at each metric interval).

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
mabdinur: We need splice here to remove all values in `measurements`. `splice` is used to flush the metrics queue (I believe it's roughly equivalent to list.pop operation in python).

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
mabdinur: Great catch. The Opentelemetry SDK [supports](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/) OTEL_BSP_MAX_QUEUE_SIZE (default 2048). I am loading this config in this PR but I was planning to add support for it in a future PR. I can just do it now. It's a small change.

packages/dd-trace/src/opentelemetry/metrics/instruments.js
mabdinur: It can be called by users. Calling this method will allow them to peek/mutate at the current value of a metric. However, it will be mainly used internally. This design follows how the Opentelemetry SDK implements `ObervableResults`: https://github.com/open-telemetry/opentelemetry-js/blob/e460f03ba6ce473e1410ff413236b3fbc8ab36f3/api/src/metrics/ObservableResult.ts#L35

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
mabdinur: Good catch! Using safe-stable-stringify is probably overkill here. Since OpenTelemetry attributes only use string keys and primitive (or array of primitive) values, we don’t need to handle nested objects. I’ve pushed a commit with a simple stableStringify implementation.

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
mabdinur: This field was meant as a temporary store for datapoints to allow quick access.
metric.data is the list that gets encoded and sent in OTLP payloads, while metric.dataPointMap is a helper map for fast lookups within that list.

If a datapoint already exists in the list, we update it; if not, we create a new one.
The idea was to maintain both structures for performance reasons, but it ends up causing more confusion than benefit. I’ll revert this and just use dataPointMap, converting its values to a list right before encoding.

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
mabdinur: The aggregator is an internal property of the PeriodicMetricReader. End user's won't be able to access it. This The return value is used in the PeriodicMetricReader to export metrics.

docs/API.md
mabdinur: Good point. I got rid of this paragraph and inlined it with the configs.

packages/dd-trace/src/config.js
mabdinur: I'll keep an eye on this

packages/dd-trace/src/opentelemetry/metrics/instruments.js
mabdinur: The wording is weird. It says the API will not enforce typing and that it's on the user to supply the right metric types 😅

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
mabdinur: I'll make it a set

packages/dd-trace/src/config.js
BridgeAR: I would not be able to understand that documentation 😅 
So we should probably explain that behavior.

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
BridgeAR: Ah, I misunderstood the intent.

We could split it into two operations that are easier to follow (splice is rarely used).

Just assign the array and reassign an empty array to measurements afterwards. That is also way faster nice no new array is created / the old one is not changed.

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -99,7 +109,27 @@ class PeriodicMetricReader {
     const allMeasurements = this.#measurements.splice(0)
 
     for (const instrument of this.#observableInstruments) {
-      allMeasurements.push(...instrument.collect())
+      if (allMeasurements.length >= this.#maxQueueSize) break
```
BridgeAR: This could only be true in case the allMeasurements is already exceeding the queue size before the instrument iteration starts. I would move it outside therefore.

And can that case be true at all?

packages/dd-trace/src/opentelemetry/otlp/otlp_transformer_base.js
```diff
@@ -6,6 +6,22 @@ const log = require('../../log')
...
+ * Attributes support primitives and arrays of primitives, not deeply nested objects.
+ * @param {Attributes} attributes - Attributes to serialize
+ * @returns {string} Stable string representation
+ */
+function stableStringify (attributes) {
+  if (attributes == null || typeof attributes !== 'object') {
+    return JSON.stringify(attributes)
+  }
+
+  const sortedKeys = Object.keys(attributes).sort()
```
BridgeAR: We should not do this for arrays.

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -103,7 +101,7 @@ class PeriodicMetricReader {
   #collectAndExport (callback = () => {}) {
     const allMeasurements = this.#measurements.splice(0)
 
-    for (const instrument of this.#observableInstruments) {
+    for (const instrument of this.#observableInstruments.values()) {
```
BridgeAR: A set can be iterated directly :)

packages/dd-trace/src/opentelemetry/metrics/constants.js
```diff
@@ -0,0 +1,28 @@
...
+// Temporality constants
+const TEMPORALITY = {
+  DELTA: 'DELTA',
+  CUMULATIVE: 'CUMULATIVE',
+  GAUGE: 'GAUGE',
+  LOWMEMORY: 'LOWMEMORY'
+}
+
+// Default histogram bucket boundaries (in milliseconds for latency metrics)
+const DEFAULT_HISTOGRAM_BUCKETS = [0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10_000]
```
IlyasShabi, line 22: nit: keep same number presentation
```suggestion
const DEFAULT_HISTOGRAM_BUCKETS = [0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1_000, 2_500, 5_000, 7_500, 10_000]
```

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -48,6 +48,7 @@ const { stableStringify } = require('../otlp/otlp_transformer_base')
  * @class PeriodicMetricReader
  */
 class PeriodicMetricReader {
+  MAX_MEASUREMENT_QUEUE_SIZE = 1_048_576 // arbitrary limit to prevent memory exhaustion
```
BridgeAR: Private?

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,513 @@
...
+/**
+ * PeriodicMetricReader collects and exports metrics at a regular interval.
+ *
+ * This implementation follows the OpenTelemetry JavaScript SDK MetricReader pattern:
+ * https://open-telemetry.github.io/opentelemetry-js/classes/_opentelemetry_sdk-metrics.PeriodicExportingMetricReader.html
+ *
+ * @class PeriodicMetricReader
+ */
+class PeriodicMetricReader {
+  MAX_MEASUREMENT_QUEUE_SIZE = 1_048_576 // arbitrary limit to prevent memory exhaustion
```
IlyasShabi: Should be moved to `constants.js` file

packages/dd-trace/src/opentelemetry/metrics/instruments.js
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  /**
+   * Creates a measurement object for recording metric values.
+   * @param {string} type - Metric type from METRIC_TYPES
+   * @param {number} value - Numeric value to record
+   * @param {Attributes} attributes - Key-value pairs for metric dimensions
+   * @returns {Measurement} Measurement object with metadata and timestamp
+   */
+  createMeasurement = (type, value, attributes) => {
```
simon-id: Are you sure you want to use an arrow function here ?

simon-id: You're not creating a class method here, you're creating a closure for every instance of that class

packages/dd-trace/src/config.js
BridgeAR: Ok, these docs sadly do not explain the behavior well. Thus, it would be great to update our docs as well :)

It is just independent of this PR as such.

packages/dd-trace/src/opentelemetry/metrics/instruments.js
simon-id: I've checked in the rest of the code and I don't see a good reason to use an arrow function here

packages/dd-trace/src/config.js
```diff
@@ -1491,6 +1531,16 @@ function maybeFloat (number) {
   return Number.isNaN(parsed) ? undefined : parsed
 }
 
+function nonNegInt (value, envVarName, allowZero = true) {
+  const parsed = maybeInt(value)
+  if (parsed === undefined) return
+  if (parsed < 0 || (parsed === 0 && !allowZero)) {
+    log.warn(`Invalid value ${parsed} for ${envVarName}. Using default value.`)
```
BridgeAR: This is following the concrete wording of https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#duration

However, I believe the intention is to log warnings for invalid values and parsing a non integer to int will be NaN. So I would actually warn for these as well and make a suggestion to the Otel documentation.

packages/dd-trace/src/opentelemetry/metrics/instruments.js
BridgeAR: Let's follow-up with a consolidated view on our side through the Otel-guild.

In my opinion any value that should not be accepted should either be: ignored, log a warning, or throw an error. It should never cause downstream issues that are very difficult to debug. I would therefore keep our current implementation as bare minimum 'validation', even if the doc says we should not.

packages/dd-trace/src/opentelemetry/metrics/meter.js
```diff
@@ -0,0 +1,194 @@
+'use strict'
+
+const { version: packageVersion } = require('../../../../../package.json')
```
BridgeAR:
```suggestion
const { VERSION: packageVersion } = require('../../../../../version')
```

packages/dd-trace/src/opentelemetry/metrics/meter_provider.js
```diff
@@ -0,0 +1,116 @@
...
+   * Gets or creates a meter instance.
+   *
+   * @param {string} name - Meter name (case-insensitive)
+   * @param {string} [version] - Meter version
+   * @param {Object} [options] - Additional options
+   * @param {string} [options.schemaUrl] - Schema URL for the meter
+   * @param {Object} [options.attributes] - Attributes for the instrumentation scope
+   * @returns {Meter} Meter instance
+   */
+  getMeter (name, version = '', { schemaUrl = '', attributes = {} } = {}) {
```
BridgeAR: This is actually different than the spec. The attributes are not defined in the API.

Can we please use the otel/api types? These would have caught that early on.

Could you please check once more if there are any other methods that could also utilize the otel/api types?

packages/dd-trace/src/opentelemetry/otlp/otlp_http_exporter_base.js
```diff
@@ -49,31 +47,21 @@ class OtlpHttpExporterBase {
...
    * @param {number} count - Count to increment
-   * @param {Array<string>} [tags] - Optional custom tags (defaults to this exporter's tags)
+   * @param {Array<string>} [additionalTags] - Optional custom tags (defaults to this exporter's tags)
    * @protected
    */
-  _recordTelemetry (metricName, count, tags) {
-    const telemetryTags = tags || this.#telemetryTags
-    tracerMetrics.count(metricName, telemetryTags).inc(count)
+  recordTelemetry (metricName, count, additionalTags) {
+    tracerMetrics.count(metricName, [...this.telemetryTags, ...additionalTags]).inc(count)
```
BridgeAR:
```suggestion
    tracerMetrics.count(metricName, additionalTags ? [...this.telemetryTags, ...additionalTags] : this.telemetryTags).inc(count)
```

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
```diff
@@ -11,7 +11,7 @@ const OtlpTransformer = require('./otlp_transformer')
 /**
  * OtlpHttpLogExporter exports log records via OTLP over HTTP.
  *
- * This implementation follows the OTLP HTTP specification:
+ * This implementation follows the OTLP HTTP v1.7.0 specification:
```
BridgeAR, line 14: Could you move the `@opentelemetry/resources` dependency to dev dependencies? It is only used as type for us. The same applies to `@opentelemetry/api-logs`.

packages/dd-trace/src/opentelemetry/metrics/instruments.js
mabdinur: Nice catch! I completely missed this. This change snuck into an auto-complete when I was refactoring this class. I'll revert this change.

packages/dd-trace/src/opentelemetry/metrics/meter_provider.js
mabdinur: Yup. I removed the attributes type. The options should be a `MeterOptions`. 
I also replaced the type of the options field in `createCounter/createUpDownCounter/createGauge` from `Object` to `MeterOptions`. 

With these two changes we should be fully compatible with the spec

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
mabdinur: I updated this to:

```
const allMeasurements = this.#measurements
this.#measurements = []
```

packages/dd-trace/src/opentelemetry/otlp/otlp_transformer_base.js
mabdinur: I removed this function. It is no longer needed. `getMeter` does accept attributes

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
mabdinur: Yup. Done

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
mabdinur: I moved this to constants.js and renamed it to DEFAULT_MAX_MEASUREMENT_QUEUE_SIZE. We can make this limit configurable but thats beyond the scope of this PR

packages/dd-trace/src/opentelemetry/otlp/otlp_http_exporter_base.js
mabdinur: I used an if statement instead but the logic should be equivalent

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
mabdinur: I opened this PR to move `opentelemetry/resources`: https://github.com/DataDog/dd-trace-js/pull/6883. 

`api-logs` is used here, we can't make it a dev dependency: https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/opentelemetry/logs/logger_provider.js#L2

packages/dd-trace/src/opentelemetry/otlp/otlp_transformer_base.js
```diff
@@ -61,7 +45,7 @@ class OtlpTransformerBase {
 
     for (const item of items) {
       const instrumentationScope = item.instrumentationScope || { name: '', version: '', schemaUrl: '', attributes: {} }
-      const attrsKey = stableStringify(instrumentationScope.attributes || {})
+      const attrsKey = JSON.stringify(instrumentationScope.attributes || {})
```
BridgeAR: If attributes are an object we must use a stable stringify implementation. Otherwise the keys would be stringified in insertion order and it would not match anymore.

packages/dd-trace/src/config.js
```diff
@@ -1532,9 +1532,9 @@ function maybeFloat (number) {
 }
 
 function nonNegInt (value, envVarName, allowZero = true) {
-  const parsed = maybeInt(value)
-  if (parsed === undefined) return
-  if (parsed < 0 || (parsed === 0 && !allowZero)) {
+  if (value === undefined) return
+  const parsed = Number.parseInt(value)
+  if (Number.isNaN(parsed) || parsed < 0 || (parsed === 0 && !allowZero)) {
```
BridgeAR, line 1538: Should we add a test case for the NaN part being logged?

packages/dd-trace/src/opentelemetry/metrics/constants.js
```diff
@@ -0,0 +1,33 @@
...
+
+// Maximum number of measurements to queue before dropping
+// This is an arbitrary limit to prevent memory exhaustion
+const DEFAULT_MAX_MEASUREMENT_QUEUE_SIZE = 524_288
```
BridgeAR: Should this value not just be the default value for `OTEL_BSP_MAX_QUEUE_SIZE`? That way it is clear that it is used as such. We define theses in `config_defaults.js`. I actually see that we already have a default there. Is it meant to have a separate limit?

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,510 @@
...
+        this.#droppedCount += observableMeasurements.length
+        continue
+      }
+
+      const remainingCapacity = DEFAULT_MAX_MEASUREMENT_QUEUE_SIZE - allMeasurements.length
+
+      if (observableMeasurements.length <= remainingCapacity) {
+        allMeasurements.push(...observableMeasurements)
+      } else {
+        allMeasurements.push(...observableMeasurements.slice(0, remainingCapacity))
```
BridgeAR, line 182: These measurements could exceed the maximum number of arguments (about 2 ** 16). If that is the case, an error is thrown. I am uncertain if it might ever happen that we'd have that many measurements.

packages/dd-trace/src/opentelemetry/otlp/otlp_transformer_base.js
mabdinur: I reintroduced the stableStringify implementation:

```
function stableStringify (attributes) {
  if (attributes == null || typeof attributes !== 'object') {
    return JSON.stringify(attributes)
  }
  // Attributes are sorted by key to ensure consistent serialization regardless of key order.
  // Keys are always strings and values are always strings, numbers, booleans, 
  // or arrays of strings, numbers, or booleans.
  return Object.keys(attributes)
    .sort()
    .map(key => `${key}:${JSON.stringify(attributes[key])}`)
    .join(',')
}
```

We typically expect each metric to have 1–5 attributes. Since Datadog bills based on the number of unique metric + attribute combinations, adding more tags can dramatically increase metric cardinality and therefore costs. To prevent this, users will likely only set a few attributes per metric. So sorting attributes by key will likely add minimal overhead.

mabdinur: Also added a test case to catch this

packages/dd-trace/src/config.js
mabdinur: Done. Also, Number.parseInt(...) partially parses integers, for example, nonNegInt("23213fg") returns 23213. This behavior feels a bit off but it’s consistent with maybeInt, so I think it’s best to leave it as is.

packages/dd-trace/src/opentelemetry/metrics/constants.js
mabdinur: We currently have two limits:

DEFAULT_MAX_MEASUREMENT_QUEUE_SIZE – The maximum number of individual measurements (e.g., counts, gauges) that can be recorded during a metric interval. These measurements are held in memory (not immediately batched) to reduce overhead in the hot path. The limit is 524_288.

OTEL_BSP_MAX_QUEUE_SIZE – The maximum number of aggregated metrics (measurements sharing the same metric name and attributes) that can be sent to the agent within a metrics interval. Metric batching occurs periodically in a separate thread.

This design decouples metric queuing from batching and sending.
Also, we’ll need to ensure that metrics are added to and removed from `PeriodicMetricReader.measurements` in a thread-safe way, I’ll address that in my next commit.

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
mabdinur: DEFAULT_MAX_MEASUREMENT_QUEUE_SIZE is set to 2^19. So `remainingCapacity` will always be less than this value

packages/dd-trace/src/opentelemetry/metrics/constants.js
mabdinur: We could combine these limits and avoid storing individual measurements. This would add some overhead to recording metric values but it could simplify the logic and reduce memory overhead. We could also do this in a follow up.

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
mabdinur: I cleaned this up

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
BridgeAR: Sorry, I wrote a wrong upper bound first. I meant 2 ^ 16.

packages/dd-trace/src/opentelemetry/metrics/constants.js
BridgeAR: Where do we spin up threads? I can not see that in the code.

packages/dd-trace/src/config.js
BridgeAR: I think being stricter for new options is actually better. Just because we were not validating something properly earlier does not mean we have to do the same mistake again in new code :)

packages/dd-trace/src/opentelemetry/otlp/otlp_transformer_base.js
BridgeAR: I believe the attributes are already guaranteed to be an object, no?

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
BridgeAR: I think we should use it as a peer dependency instead of a regular one.

packages/dd-trace/src/opentelemetry/metrics/meter_provider.js
```diff
@@ -0,0 +1,113 @@
...
+      meter = new Meter(this, { name: normalizedName, version, schemaUrl })
+      this.#meters.set(key, meter)
+    }
+    return meter
+  }
+
+  /**
+   * Registers this meter provider as the global provider.
+   */
+  register () {
```
BridgeAR: I just checked the API referenced and it only contains the getMeter method. Also https://opentelemetry.io/docs/specs/otel/metrics/api/ only mentions the getMeter method.

Could these have been added by accident?

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,512 @@
...
+      this.#timer = null
+    }
+  }
+
+  /**
+   * Collects measurements and exports metrics.
+   * @private
+   * @param {Function} [callback] - Called after export completes
+   */
+  #collectAndExport (callback = () => {}) {
```
BridgeAR: The method is never called with arguments, so it should not have the callback defined. I guess it should just be a no-op for the exporter call in the end?

```suggestion
   */
  #collectAndExport () {
```

And the @ private should not be included for JSDoc on private methods :) It will be understood that it's private.

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
BridgeAR: It would just throw an error if the limit would be reached.

packages/dd-trace/src/opentelemetry/metrics/meter_provider.js
mabdinur: This is the pattern used in the OpenTelemetry SDK to ensure providers are registered and the context manager is used. We use the same pattern in the ddtrace  TracerProvider, so I did the same thing for metrics: https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-sdk-trace-node/src/NodeTracerProvider.ts#L96. 

I can remove this for now.

packages/dd-trace/src/opentelemetry/metrics/constants.js
mabdinur: Ahhh I assumed `setInterval` created a thread, that doesn't seem to be the case 😓. I can modify this logic to use threads if you think that's a good path forward

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
mabdinur: I'll update `DEFAULT_MAX_MEASUREMENT_QUEUE_SIZE` to have a limit of 2^16. That should simplify things. We can rewrite this logic if this limit isn't enough.

packages/dd-trace/src/opentelemetry/otlp/otlp_transformer_base.js
mabdinur: Yup they should be. I can get rid of the additional check

packages/dd-trace/src/opentelemetry/logs/otlp_http_log_exporter.js
mabdinur: Yup, I will do that in a separate PR.  We can make both of these peer dependencies:

```
    "@opentelemetry/api-logs": "<1.0.0",
 ```
 
 https://github.com/DataDog/dd-trace-js/pull/6914

File: packages/dd-trace/src/opentelemetry/metrics/constants.js (hunk 58 above)
Comment (BridgeAR) [reply packages/dd-trace/src/opentelemetry/metrics/constants.js]: Creating workers always has an overhead. I am not sure if this is beneficial for these operations. I doubt that for most use cases.

File: packages/dd-trace/test/opentelemetry/metrics.spec.js [hunk 62]
```diff
@@ -0,0 +1,1035 @@
...
+
+      setTimeout(() => { validator(); done() }, 150)
+    })
+
+    it('exports histogram metrics', (done) => {
+      const validator = mockOtlpExport((decoded) => {
+        const histogram = decoded.resourceMetrics[0].scopeMetrics[0].metrics[0]
+        assert.strictEqual(histogram.name, 'duration')
+        assert.strictEqual(histogram.histogram.dataPoints[0].count, 1)
+        assert.strictEqual(histogram.histogram.dataPoints[0].sum, 100)
```
BridgeAR, line 141: Nit: I think checking for all properties of the histogram would be good.

Nit: Ideally, we use our assertObjectContains helper instead of having many individual assertions.

packages/dd-trace/test/opentelemetry/metrics.spec.js
```diff
@@ -0,0 +1,1035 @@
...
+
+      const gauge = meter.createObservableGauge('memory')
+      gauge.addCallback(() => { throw new Error('Callback error') })
+      gauge.addCallback((result) => result.observe(100))
+
+      setTimeout(() => {
+        assert(validated, 'Should have validated an export with all metrics')
+        validator()
+        done()
+      }, 150)
```
BridgeAR, line 191: Instead of using a timeout, it would be possible to just call `done()` from inside of the `validator` method. That way it is simpler and faster.

The same applies for all other methods with that pattern.

It is also more reliable, due to not being dependent on the timing anymore.

Or even better: the `mockOtlpExport` method returns a promise. That way you can trigger the methods sync and just return that promise afterwards. That would be very clean.

packages/dd-trace/test/opentelemetry/metrics.spec.js
```diff
@@ -0,0 +1,1035 @@
...
+        const counter = metrics.find(m => m.sum)
+        const histogram = metrics.find(m => m.histogram)
+        assert(counter, 'Should have counter')
+        assert(histogram, 'Should have histogram')
+        assert.strictEqual(counter.name, 'test')
+        assert.strictEqual(histogram.name, 'test')
+        assert.strictEqual(counter.sum.dataPoints.length, 1, 'Counter should have 1 data point')
+        assert.strictEqual(histogram.histogram.dataPoints.length, 1, 'Histogram should have 1 data point')
+        assert.strictEqual(counter.sum.dataPoints[0].asInt, 5)
+        assert.strictEqual(histogram.histogram.dataPoints[0].sum, 100)
```
BridgeAR, line 584: I know we have lots of code like that, while I would like to start prevent calling so many asserts on a single object. We should just use the partial helper (assertObjectContains, if I am not mistaken)

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,532 @@
...
+
+    const state = cumulativeState.get(stateKey)
+
+    let bucketIndex = DEFAULT_HISTOGRAM_BUCKETS.length
+    for (let i = 0; i < DEFAULT_HISTOGRAM_BUCKETS.length; i++) {
+      if (value <= DEFAULT_HISTOGRAM_BUCKETS[i]) {
+        bucketIndex = i
+        break
+      }
+    }
```
BridgeAR, line 504: Ideally, this would use a binary search.

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,532 @@
...
+        count: 0,
+        sum: 0,
+        min: Infinity,
+        max: -Infinity,
+        bucketCounts: new Array(DEFAULT_HISTOGRAM_BUCKETS.length + 1).fill(0),
+        startTime: metric.temporality === TEMPORALITY.CUMULATIVE ? this.#startTime : timestamp
+      })
+    }
+
+    const state = cumulativeState.get(stateKey)
```
BridgeAR, line 496: Nit

Ideally, it would just get the state if it was defined. If it was not defined, we could combine the handling to just set the first value.

That would be a tad faster.

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,532 @@
...
+      cumulativeState.set(stateKey, {
+        value: 0,
+        startTime: metric.temporality === TEMPORALITY.CUMULATIVE ? this.#startTime : timestamp
+      })
+    }
+
+    const state = cumulativeState.get(stateKey)
+    state.value += value
+
+    const dataPoint = this.#findOrCreateDataPoint(metric, attributes, attrKey, () => ({
```
BridgeAR, line 440: I believe the idea around the method is an performance optimization to prevent the object from being created?

Instead of an object, we are now creating a method that is referencing these parts. I did not benchmark something like that before, I would expect the function creation still to take more time though. The pattern is only useful in case there would be many compute heavy operations.

packages/dd-trace/src/opentelemetry/metrics/periodic_metric_reader.js
```diff
@@ -0,0 +1,532 @@
...
+    state.max = Math.max(state.max, value)
+
+    const dataPoint = this.#findOrCreateDataPoint(metric, attributes, attrKey, () => ({
+      startTimeUnixNano: state.startTime,
+      timeUnixNano: timestamp,
+      count: 0,
+      sum: 0,
+      min: Infinity,
+      max: -Infinity,
+      bucketCounts: new Array(DEFAULT_HISTOGRAM_BUCKETS.length + 1).fill(0),
```
BridgeAR, line 519: This is actually a compute heavy operations (referencing a comment above about findOrCreateDataPoint and the method vs object creation).

Making this lazy is a good idea. To prevent the overhead, I would still create a simple object and just use optional assignment to assign the bucketCounts after the call. That way `dataPoint` will be updated due to the reference and it is cheap.

BridgeAR (approved):
LGTM with a few nits and potential follow-ups :)

PR #6788, opened 2025-10-29
add maximum node version in guardrails

rochdev:
### What does this PR do?

Add maximum Node version in guardrails.

### Motivation

We shouldn't support major versions of Node that don't exist yet in SSI since we automatically inject the library everywhere and if there is any major problem it could crash.

### Additional notes

Since Node 25 is already out at this point, I put the upper range at 26.

### DIFF
--- a/.github/workflows/platform.yml
+++ b/.github/workflows/platform.yml
@@ -461,6 +461,7 @@ jobs:
 
   integration-guardrails-unsupported:
     strategy:
+      fail-fast: false
       matrix:
         version: ['0.8', '0.10', '0.12', '4', '6', '8', '10', '12']
     runs-on: ubuntu-latest
--- a/integration-tests/init.spec.js
+++ b/integration-tests/init.spec.js
@@ -202,38 +202,40 @@ true
       })
     })
 
-    context('when node version is in range of the engines field', () => {
-      useEnv({ NODE_OPTIONS })
+    if (currentVersionIsSupported) {
+      context('when node version is in range of the engines field', () => {
+        useEnv({ NODE_OPTIONS })
 
-      before(() => {
-        const pkg = JSON.parse(pkgStr)
-        pkg.engines.node = '>=0 <1000'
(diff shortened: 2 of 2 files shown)


packages/dd-trace/src/guardrails/index.js
```diff
@@ -14,7 +14,9 @@ function guard (fn) {
   var clobberBailout = false
   var forced = isTrue(process.env.DD_INJECT_FORCE)
   var engines = require('../../../../package.json').engines
-  var minMajor = parseInt(engines.node.replace(/[^0-9]/g, ''))
+  var majors = engines.node.match(/\d+/)
```
BridgeAR: This regexp is not very safe. It is likely that the engines field would be changed at some point and we might e.g., provide support from a minor range on.

rochdev: It works with the current release policy and if that changes we can change this code as well. It's also the same expression used in the preinstall script, the old one was inconsistent but this one is now the same. Please propose a specific regular expression if you don't like this one.

BridgeAR (commented):
This is almost LGTM

The test is now testing the second condition, just not the first condition anymore, as far as I can tell.

Ideally we just verify the edge values and use the current Node.js version as being the only accepted one. We can do that ff we use the following ranges (only testing edge values):

Node.js version e.g., 24.5.3
- => 24.5.3 < 24.5.4 - Accepted patch
- => 24.5.2 < 24.5.3 - Too high patch
- => 24.5.4 < 24.5.5 - Too low patch
- => 24.5 < 24.6 - Accepted minor
- => 24.4 < 24.5 - Too high minor
- => 24.6 < 24.7 - Too low minor
- => 24 < 25 - Accepted major
- => 23 < 24 - Too high major
- => 25 < 26 - Too low major

We could skip patch versions, if we make the range check accept only major and minor versions.

Ideally, we also verify the whole version range to support minor ranges right away.

PR #6796, opened 2025-10-30
fix: made wrapFunction unique to functions

pabloerhard:
### What does this PR do?
This PR limits shimmers.wrapFunction to only be usable for functions. This prevents wrapping behavior from being applied to non-function values, which could previously lead to unexpected behavior and caused regressions with the ESM default export changes  previously applied.

### DIFF
--- a/packages/datadog-instrumentations/src/mariadb.js
+++ b/packages/datadog-instrumentations/src/mariadb.js
@@ -97,7 +97,7 @@ function createWrapQueryCallback (options) {
         arguments[arguments.length - 1] = shimmer.wrapFunction(cb, wrapper)
       } else {
         arguments.length += 1
-        arguments[arguments.length - 1] = wrapper(cb)
+        arguments[arguments.length - 1] = wrapper()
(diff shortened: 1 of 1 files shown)


packages/datadog-instrumentations/src/mariadb.js
```diff
@@ -96,7 +91,14 @@ function createWrapQueryCallback (options) {
...
           ? finishCh.runStores(ctx, cb, this, ...arguments)
           : finishCh.publish(ctx)
-      })
+      }
+
+      if (typeof cb === 'function') {
+        arguments[arguments.length - 1] = shimmer.wrapFunction(cb, wrapper)
+      } else {
+        arguments.length += 1
+        arguments[arguments.length - 1] = wrapper(cb)
```
BridgeAR:
```suggestion
        arguments[arguments.length - 1] = wrapper()
```

BridgeAR (commented):
LGTM with the comment addressed

PR #6797, opened 2025-10-30
[test optimization] Playwright: report tests that did not run

juan-fernandez:
### What does this PR do?

Add a new hook for `lib/reporters/base.js` to wrap `generateSummary`, and report tests that _did not run_, so did not go through the usual instrumentation. 

### Motivation

Playwright projects may have dependencies: https://playwright.dev/docs/test-projects#dependencies. If one of the dependencies fails, the project does not run. This causes the tests **to not run at all**. This means that our instrumentation does not kick in at all (no run test function whatsoever). 

Reporters like JUnit are able to show these skipped tests, so it creates a discrepancy between our instrumentation and user experience with other test reports. This is confusing. 

This PR adds instrumentation to the reporters' base class to grab tests that did not run at all. 

### Plugin Checklist

- [x] Integration tests.

### DIFF
--- a/packages/datadog-instrumentations/src/playwright.js
+++ b/packages/datadog-instrumentations/src/playwright.js
@@ -1262,23 +1262,24 @@ addHook({
 
 function generateSummaryWrapper (generateSummary) {
   return function () {
-    // https://github.com/microsoft/playwright/blob/bf92ffecff6f30a292b53430dbaee0207e0c61ad/packages/playwright/src/reporters/base.ts#L279
-    const didNotRunTests = this.suite.allTests().filter(test =>
-      test.outcome() === 'skipped' && (!test.results.length || test.expectedStatus !== 'skipped')
-    )
-    for (const test of didNotRunTests) {
-      const {
-        _requireFile: testSuiteAbsolutePath,
-        location: { line: testSourceLine },
-      } = test
-      const browserName = getBrowserNameFromProjects(sessionProjects, test)
+    for (const test of this.suite.allTests()) {
+      // https://github.com/microsoft/playwright/blob/bf92ffecff6f30a292b53430dbaee0207e0c61ad/packages/playwright/src/reporters/base.ts#L279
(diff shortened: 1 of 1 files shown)


packages/datadog-instrumentations/src/playwright.js
```diff
@@ -1255,3 +1259,46 @@ addHook({
...
   return workerPackage
 })
+
+function generateSummaryWrapper (generateSummary) {
+  return function () {
+    // https://github.com/microsoft/playwright/blob/bf92ffecff6f30a292b53430dbaee0207e0c61ad/packages/playwright/src/reporters/base.ts#L279
+    const didNotRunTests = this.suite.allTests().filter(test =>
+      test.outcome() === 'skipped' && (!test.results.length || test.expectedStatus !== 'skipped')
+    )
+    for (const test of didNotRunTests) {
```
BridgeAR: Could we not run the logic directly in the same loop? Do we have to filter first and create an intermediate array?

packages/datadog-instrumentations/src/playwright.js
```diff
@@ -1255,3 +1259,46 @@ addHook({
...
+addHook({
+  name: 'playwright',
+  file: 'lib/reporters/base.js',
+  versions: ['>=1.38.0']
+}, (reportersPackage) => {
+  // v1.50.0 changed the name of the base reporter from BaseReporter to TerminalReporter
+  if (reportersPackage.TerminalReporter) {
+    shimmer.wrap(reportersPackage.TerminalReporter.prototype, 'generateSummary', generateSummaryWrapper)
+  }
+  if (reportersPackage.BaseReporter) {
```
BridgeAR:
```suggestion
  } else if (reportersPackage.BaseReporter) {
```

BridgeAR (commented):
LGTM % comment

packages/datadog-instrumentations/src/playwright.js
juan-fernandez: sure

PR #6803, opened 2025-10-30
chore: refactor config class to a singleton

BridgeAR:
This allows to pick up runtime configuration changes in the future as well as allowing to simplify the code in future PRs. One possible change is to remove `configure` and to move the remote configuration to just update the singleton instead.

We can add events to activate and deactivate features as soon as updates are incoming that are not handled by a simple property access.

That way most configurations will be changeable at runtime as soon as the remote configuration handles all entries.

### DIFF
--- a/packages/dd-trace/src/config.js
+++ b/packages/dd-trace/src/config.js
@@ -4,7 +4,6 @@ const fs = require('fs')
 const os = require('os')
 const uuid = require('crypto-randomuuid') // we need to keep the old uuid dep because of cypress
 const { URL } = require('url')
-const util = require('util')
 
 const log = require('./log')
 const tagger = require('./tagger')
@@ -1510,15 +1509,9 @@ function getAgentUrl (url, options) {
 }
 
 let configInstance = null
-let configOptions = null
 function getConfig (options) {
   if (!configInstance) {
-    configOptions = options
     configInstance = new Config(options)
-  }
-  // TODO: Remove before landing
-  if (!util.isDeepStrictEqual(configOptions || {}, options || {})) {
-    throw new Error('Debug: different options since last call to getConfig')
   }
   return configInstance
 }


packages/dd-trace/test/appsec/rasp/ssrf.express.plugin.spec.js
```diff
@@ -68,6 +71,9 @@ describe('RASP - ssrf', () => {
       ['http', 'https'].forEach(protocol => {
         describe(`Test using ${protocol}`, () => {
           it('Should not detect threat', async () => {
+            // Hack to enforce the module to be loaded once before the actual request
+            require(protocol)
+
             app = (req, res) => {
               const clientRequest = require(protocol).get(`${protocol}://${req.query.host}`)
```
uurien: I didn't realize when we were talking, but I think this should work too.

```suggestion
            // Hack to enforce the module to be loaded once before the actual request
            const httpModule = require(protocol)

            app = (req, res) => {
              const clientRequest = httpModule.get(`${protocol}://${req.query.host}`)
```

packages/dd-trace/test/appsec/rasp/sql_injection.mysql2.plugin.spec.js
```diff
@@ -33,7 +33,7 @@ describe('RASP - sql_injection', () => {
             app(req, res)
           })
 
-          appsec.enable(new Config({
+          appsec.enable(getConfig({
```
uurien, line 36: For the most of the tests you are using `getConfigFresh`, why `getConfig` for `sql_injection.[mysql2|pg].plugin.spec.js`?

packages/dd-trace/test/ci-visibility/dynamic-instrumentation/target-app/test-visibility-dynamic-instrumentation-script.js
```diff
@@ -3,12 +3,12 @@
...
 const getDiClient = require('../../../../src/ci-visibility/dynamic-instrumentation')
 const sum = require('./di-dependency')
-const Config = require('../../../../src/config')
+const getConfig = require('../../../../src/config')
 
 // keep process alive
 const intervalId = setInterval(() => {}, 5000)
 
-const diClient = getDiClient(new Config())
+const diClient = getDiClient(getConfig())
```
uurien, line 11: here also, why `getConfig` and not `getConfigFresh`?

packages/dd-trace/test/appsec/rasp/sql_injection.mysql2.plugin.spec.js
BridgeAR: Because this is using the actual config (it is never loaded more than once / the config does not change) without reloading and I thought it is fine to keep it like that.

packages/dd-trace/test/appsec/rasp/ssrf.express.plugin.spec.js
BridgeAR: Yes, I thought it was intentional to do the require call as well though. That is why I wrote it like that. No strong opinion either way.

packages/dd-trace/test/ci-visibility/dynamic-instrumentation/target-app/test-visibility-dynamic-instrumentation-script.js
BridgeAR: See the other similar comment.

packages/dd-trace/test/appsec/rasp/ssrf.express.plugin.spec.js
uurien: it doesn't matter when is loaded the module, what matter is that should be loaded when the instrumentation is already loaded, which is happening in the line 75. 

The reason to have it in the endpoint was to have everything in one line.

packages/dd-trace/src/config.js
```diff
@@ -1507,4 +1507,12 @@ function getAgentUrl (url, options) {
   }
 }
 
-module.exports = Config
+let configInstance = null
+function getConfig (options) {
+  if (!configInstance) {
+    configInstance = new Config(options)
+  }
+  return configInstance
```
szegedi, line 1516: Would it make sense to preserve options passed in the initial call, and check if options passed in subsequent calls are deeply equal, and throw an exception when they aren't? Otherwise we're handing back an object created earlier with different options.

BridgeAR: I can implement that

packages/dd-trace/test/appsec/rasp/ssrf.express.plugin.spec.js
BridgeAR: I implemented the suggestion now

packages/dd-trace/src/config.js
tlhunter: Good idea. That should help prevent confusing during tests. Plus it's not something we'd expect to happen when running normally.

PR #6812, opened 2025-10-31
[test optimization] Support new pools in vitest

juan-fernandez:
### What does this PR do?

Add instrumentation for `ForksPoolWorker` and `ThreadsPoolWorker`, which substitute `tinypool`. 

### Motivation
vitest@4 got rid of tinypool altogether: https://github.com/vitest-dev/vitest/pull/8705 (https://github.com/vitest-dev/vitest/releases/tag/v4.0.0), so we need to re-support IPC for the new pools.

This change hasn't broken our tests because our instrumentation has graceful degradation: if the worker can't find `TINYPOOL_WORKER_ID`, it will report just like the main process, directly to the intake or agent. The instrumentation is more fragible (workers can be shut down with a request in flight) but it still reports data if it can. 

### Plugin Checklist
The already present integration tests should be enough, as they check `test.vitest.pool`. This checks that the new instrumentation is kicking in. 

- [x] Integration tests.


integration-tests/vitest/vitest.spec.js
```diff
@@ -91,134 +91,139 @@ versions.forEach((version) => {
     const poolConfig = ['forks', 'threads']
 
     poolConfig.forEach((poolConfig) => {
-      it(`can run and report tests with pool=${poolConfig}`, (done) => {
-        receiver.gatherPayloadsMaxTimeout(({ url }) => url === '/api/v2/citestcycle', payloads => {
-          const metadataDicts = payloads.flatMap(({ payload }) => payload.metadata)
+      it(`can run and report tests with pool=${poolConfig}`, async () => {
```
juan-fernandez, line 94: no changes, just made it `async`

PR #6813, opened 2025-10-31
fix(ai): fix streamed tool call parsing, tool call arguments in v5

sabrenner:
### What does this PR do?
Fixes two bugs in the `ai` (Vercel AI SDK) integration:
1. Streamed tool calls previously did not have their arguments parsed correctly for all `ai` versions
2. Non-streamed tool calls in v5+ previously were not being parsed correctly

### Motivation
Bug fixes identified as part of #6709, which reworked the assertions to be more accurate, which uncovered this issue.

MLOB-4382


packages/dd-trace/test/llmobs/plugins/ai/index.spec.js
```diff
@@ -629,15 +627,13 @@ describe('Plugin', () => {
...
          *
-         * Usually, this would mean the tool call name is 'toolCall'.
-         *
-         * However, because we used mocked responses, the second time this test is called, the tool call
-         * will have the name 'weather' instead. We just assert that the name exists and is a string to simplify.
+         * Usually, this would mean the tool call name is 'toolCall'. This is a limitation with the older library
+         * versions. In v5+, this is resolved as the tool name is not its index in the tools array, but its actual name.
          */
-        name: MOCK_STRING,
+        name: semifies(realVersion, NODE_MAJOR < 22 ? '<=4.0.2' : '<4.0.2') ? 'toolCall' : 'weather',
```
BridgeAR, line 636: Is this really related to v4.0.2? Or could we just say it is below 5?

sabrenner: i believe it is related to that version, although to be honest i've always been a little confused on why this doesn't work in a matrix of node version and package version, but was apparent when i was testing. as in, it _does_ work with 4.0.2 when using node 24, but does **not** work when using node 22 or lower (hence the check here to make it either strictly less than or equal, or just less than). i didn't dig too deep into the library to see why this was happening.

but, i think it should be fine, since v5 is now the currently release line, we're just supporting v4 as it's not deprecated. maybe in a future major of the tracer we'll just drop support for simplicity though 😅

PR #6814, opened 2025-10-31
chore: add ESM tests for instrumentations missing them

pabloerhard:
### What does this PR do?
Add ESM test instrumentations that didn't have one.

### Motivation
Ensure instrumentations work for ESM and prevent any regression on future ESM work.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/datadog-instrumentations/test/knex.spec.js
+++ b/packages/datadog-instrumentations/test/knex.spec.js
@@ -7,8 +7,6 @@ const agent = require('../../dd-trace/test/plugins/agent')
 const { storage } = require('../../datadog-core')
 const { withVersions } = require('../../dd-trace/test/setup/mocha')
 
-require('../src/knex')
-
 describe('Instrumentation', () => {
   let knex
   let client
@@ -17,6 +15,11 @@ describe('Instrumentation', () => {
   describe('knex', () => {
     withVersions('knex', 'knex', version => {
       describe('without configuration', () => {
+        /**
+         * TODO (Pablo Erhard): Implement a single mechanism to automatically trigger hook loading
+         * across all tests. This should allow test hooks to initialize programmatically
+         * without requiring explicit `agent.load()` calls in every individual test file.
+         */
         before(async () => {
           await agent.load()
         })


packages/datadog-plugin-vm/test/integration-test/client.spec.js
```diff
@@ -0,0 +1,35 @@
+'use strict'
+
+const {
+  useSandbox, sandboxCwd, varySandbox, curl,
+  FakeAgent, spawnPluginIntegrationTestProc
+} = require('../../../../integration-tests/helpers')
+const { assert } = require('chai')
```
BridgeAR: Could you please change these to use Node.js assert instead in all of these files?

```suggestion
const assert = require('node:assert/strict')
```

packages/datadog-plugin-sequelize/test/integration-test/client.spec.js
```diff
@@ -0,0 +1,38 @@
+'use strict'
+
+const {
+  sandboxCwd, useSandbox, varySandbox, curl,
+  FakeAgent, spawnPluginIntegrationTestProc
+} = require('../../../../integration-tests/helpers')
+const { assert } = require('chai')
+const { withVersions } = require('../../../dd-trace/test/setup/mocha')
+
+withVersions('sequelize', 'sequelize', '>=4', version => {
```
BridgeAR: Do we need to limit the version?

The same question about other test files with version limits.

packages/datadog-instrumentations/test/knex.spec.js
```diff
@@ -16,27 +17,36 @@ describe('Instrumentation', () => {
   describe('knex', () => {
     withVersions('knex', 'knex', version => {
       describe('without configuration', () => {
+        before(async () => {
+          await agent.load()
```
BridgeAR, line 24: That should actually not be needed. I guess it is about another test not cleaning up properly?

pabloerhard: With out that the hooks were never executing (patching was not happening), most tests in that folder use this approach to trigger them and also handle dependencies. 

Another approach would be to do  `require('../../dd-trace')` at the top of the file and it would also work.

PR #6837, opened 2025-11-04
[test optimization] Move plugin jest tests to integration tests

juan-fernandez:
### What does this PR do?

* Remove `plugin-jest` job.
* Move `plugin-jest` tests to `integration-jest`.
* Now jest integration tests run for `jest@24.8.0` and `jest@latest` instead of just `latest`.
* Remove `"jest"` entries in `packages/dd-trace/test/plugins/externals.json`, as they are no longer needed for plugin tests.

### Motivation

Similar to #6651 

Integration tests better reflect how test optimization is used. Also, running integration tests with older versions of `jest` is going to increase our coverage. 

### Plugin Checklist
Just check that integration tests pass


anmarchenko (dismissed):

integration-tests/ci-visibility/run-jest.js
```diff
@@ -5,12 +5,15 @@ const jest = require('jest')
 const options = {
   projects: [__dirname],
   testPathIgnorePatterns: ['/node_modules/'],
+  modulePathIgnorePatterns: ['<rootDir>/\\.bun/'],
```
BridgeAR, line 8: I believe the path should instead be added to the gitignore path. I could imagine it is then not needed in here anymore?

juan-fernandez: `bun` will create this folder regardless of gitignore, right? The problem is that `jest` attempts to use this folder for some reason.

PR #6841, opened 2025-11-04
test: reduce esbuild flakiness

BridgeAR:


integration-tests/esbuild/index.spec.js
```diff
@@ -4,55 +4,61 @@
...
       process.chdir(originalDir)
+      execSync('npm remove esbuild', {
+        timeout
+      })
     })
 
-    it('works', () => {
-      console.log('npm run build')
+    it('works', function () {
+      this.timeout(timeout)
```
juan-fernandez: do we need `this.timeout()` calls in every test? I think the `before` one should be enough

PR #6860, opened 2025-11-06
chore(deps): bump @datadog/native-iast-taint-tracking to v4.1.0

CarlesDD:
### What does this PR do?
Bumps `@datadog/native-iast-taint-tracking` to v4.1.0

This new version includes:
- Support for Node.js v25
- Binary size reduction
- New transaction key handling

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
   "dependencies": {
     "@datadog/libdatadog": "0.7.0",
     "@datadog/native-appsec": "10.3.0",
-    "@datadog/native-iast-taint-tracking": "4.0.0",
+    "@datadog/native-iast-taint-tracking": "4.1.0",
     "@datadog/native-metrics": "3.1.1",
     "@datadog/openfeature-node-server": "0.1.0-preview.15",
     "@datadog/pprof": "5.12.0",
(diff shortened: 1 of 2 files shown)


uurien (approved):

BridgeAR:
@CarlesDD do we know what the performance regression is about?

CarlesDD:
The regression observed in the instruction count is due to the introduction of a new system for handling transaction keys in the native taint-tracking module of IAST.

This change is necessary because we identified scenarios where, under certain memory-pressure conditions, the GC may move some of the references we were using as keys. When this happened, the corresponding transaction reference could be lost, potentially filling all the slots in the internal transaction dictionary. In such a situation, IAST will stop detecting vulnerabilities for the affected process.

The benchmarks where the regression appears are configured with oversized scenarios: 100% sampling (while the default is 30%) and a transaction dictionary size of 1000 (while the default is 2). These conditions significantly amplify the regression. Under the default configuration (used by the vast majority of customers in production) the performance impact is close to negligible.

So, while this change introduces a slight overhead in extreme conditions, it improves the robustness of the system and prevents critical failures in vulnerability detection.

PR #6861, opened 2025-11-06
Compute and set http endpoint when route is not available

IlyasShabi:
### What does this PR do?

Implements the `http.endpoint` tag feature for APM Trace Metrics on HTTP service entry spans [APPSEC-58481](https://datadoghq.atlassian.net/browse/APPSEC-58481)

- [x] Added resourceRenamingEnabled configuration flag controlled by `DD_TRACE_RESOURCE_RENAMING_ENABLED` env var, it's also automatically enabled when DD_APPSEC_ENABLED=true unless explicitly disabled.
- [x] Compute `http.endpoint` based on request url and regex to identify and replace parameter types
- [x] Set `http.endpoint`  only when `http.route`  is not available
- [x] Add HTTPMethod and HTTPEndpoint fields to span stats buckets
- [x] Enriched aggregation keys with HTTP method and endpoint

### Motivation
APM tracers currently submit trace metrics to the agent. However, these metrics lack endpoint granularity when http.route is unavailable (when using native http module in Node.js)
This PR implements a fallback mechanism that computes endpoints from raw URLs, enabling endpoint level statistics even when route information is missing.

### Plugin Checklist

- [x] Unit tests.
- [x] Integration tests.
- [x] System tests https://github.com/DataDog/system-tests/pull/5821
- [x] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

[APPSEC-58481]: https://datadoghq.atlassian.net/browse/APPSEC-58481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/test/span_stats.spec.js
+++ b/packages/dd-trace/test/span_stats.spec.js
@@ -136,7 +136,8 @@ describe('SpanAggKey', () => {
       }
     }
     const key = new SpanAggKey(span)
-    assert.strictEqual(key.toString(), 'basic-span,service-name,resource-name,span-type,200,false,POST,/users/{param:int}')
+    assert.strictEqual(
+      key.toString(), 'basic-span,service-name,resource-name,span-type,200,false,POST,/users/{param:int}')
   })
 
   it('should prioritize http.route over http.endpoint', () => {


packages/dd-trace/src/encode/span-stats.js
```diff
@@ -70,6 +70,12 @@ class SpanStatsEncoder extends AgentEncoder {
 
     this._encodeString(bytes, 'TopLevelHits')
     this._encodeLong(bytes, stat.TopLevelHits)
+
+    this._encodeString(bytes, 'HTTPMethod')
+    this._encodeString(bytes, stat.HTTPMethod)
+
+    this._encodeString(bytes, 'HTTPEndpoint')
+    this._encodeString(bytes, stat.HTTPEndpoint)
```
IlyasShabi, line 78: I’m not sure if we should truncate this as we do for `Stats.Type`
The max lengths are defined in [agent](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/traceutil/normalize/normalize.go)

Waiting for an answer in the RFC

packages/dd-trace/src/plugins/util/url.js
```diff
@@ -2,6 +2,121 @@
...
+ * - Handles edge cases like malformed URLs
+ * - Works with relative paths
+ * - Cross tracers compatibility
+ *
+ * @param {string} url
+ * @returns {string} Url path
+ */
+function extractPathFromUrl (url) {
+  if (!url) return '/'
+  const match = url.match(PATH_REGEX)
```
uurien, line 74: why this approach instead using `new URL().pathname`?

packages/dd-trace/src/config.js
```diff
@@ -1208,6 +1212,15 @@ class Config {
     this.#setBoolean(calc, 'isGitUploadEnabled',
       calc.isIntelligentTestRunnerEnabled && !isFalse(getEnv('DD_CIVISIBILITY_GIT_UPLOAD_ENABLED')))
 
+    // Enable resourceRenamingEnabled when appsec is enabled and only
+    // if DD_TRACE_RESOURCE_RENAMING_ENABLED is not explicitly set
```
CarlesDD, line 1262: Since `resourceRenamingEnabled` is only computed here, enabling AppSec via Remote Configuration at runtime will not update this value. As a result, `resourceRenamingEnabled` will remain disabled even if AppSec becomes active later. Is this expected?

packages/datadog-plugin-http/test/http_endpoint.spec.js
```diff
@@ -0,0 +1,113 @@
...
+            })
+        })
+
+        afterEach(() => {
+          appListener && appListener.close()
+          return agent.close({ ritmReset: false })
+        })
+
+        beforeEach(() => {
+          require('../../dd-trace') // Load tracer
```
uurien: why do you need to require de tracer without initializing it?

packages/dd-trace/test/config.spec.js
```diff
@@ -3090,6 +3090,89 @@ rules:
...
+      }
+      if (originalAppsecEnabled !== undefined) {
+        process.env.DD_APPSEC_ENABLED = originalAppsecEnabled
+      } else {
+        delete process.env.DD_APPSEC_ENABLED
+      }
+    })
+
+    it('should be false by default', () => {
+      const config = new Config()
```
uurien: Update from `master` branch and change these tests, because `new Config` is not available anymore.

packages/dd-trace/src/plugins/util/url.js
IlyasShabi: Using `new URL().pathname` is valid option too but it could fail with a malformed URL. RFC warning:

> but beware, as some urls may fail to be parsed by the most strict tools

I used the regex because:
- Compatibility with other tracers
- Gracefully handle any URL
- Faster than creating new URLs

packages/dd-trace/src/config.js
IlyasShabi: This is a valid question that I asked Christophe about it. In fact we could avoid it right now since it's complicated and we only have 1% customer who are enabling appsec via RC, and also because a service will report different kind of aggregation depending on wether appsec is enabled/disabled by RC during its lifetime,  which could lead to weird data in the end. (as Christophe said)

packages/dd-trace/src/plugins/util/url.js
```diff
@@ -2,6 +2,121 @@
...
+
+  const protocol = getProtocol(req)
+  return `${protocol}://${req.headers.host}${req.originalUrl || req.url}`
+}
+
+function getProtocol (req) {
+  if (req.socket && req.socket.encrypted) return 'https'
+  if (req.connection && req.connection.encrypted) return 'https'
+
+  return 'http'
```
CarlesDD:
```suggestion
  return (req.socket?.encrypted || req.connection?.encrypted) ? 'https' : 'http'
```

packages/dd-trace/src/plugins/util/url.js
```diff
@@ -2,6 +2,121 @@
...
+ * - Cross tracers compatibility
+ *
+ * @param {string} url
+ * @returns {string} Url path
+ */
+function extractPathFromUrl (url) {
+  if (!url) return '/'
+  const match = url.match(PATH_REGEX)
+
+  return match && match.groups && match.groups.path ? match.groups.path : '/'
```
CarlesDD:
```suggestion
  return match?.groups?.path ? match.groups.path : '/'
```

uurien: so.. why not just:
```suggestion
  return match?.groups?.path || '/'
```

packages/dd-trace/src/plugins/util/url.js
```diff
@@ -2,6 +2,118 @@
 
 const { URL } = require('url')
 
+const HTTP2_HEADER_AUTHORITY = ':authority'
+const HTTP2_HEADER_SCHEME = ':scheme'
+const HTTP2_HEADER_PATH = ':path'
+
+const PATH_REGEX = /^(?:[a-z]+:\/\/(?:[^?/]+))?(?<path>\/[^?]*)(?:(\?).*)?$/
+
+const INT_SEGMENT = /^[1-9][0-9]+$/ // Integer of size at least 2
```
bengl: Isn't this more like "Positive integer of value at least 10"?

packages/dd-trace/src/plugins/util/url.js
```diff
@@ -2,6 +2,118 @@
...
+  return match?.groups?.path || '/'
+}
+
+/**
+ * Calculate http.endpoint from URL path
+ *
+ * @param {string} url
+ * @returns {string} The normalized endpoint
+ */
+function calculateHttpEndpoint (url) {
```
bengl, line 85: This whole function seems like a lot of string processing. Can you add a benchmark for this?

IlyasShabi: Good point! done https://github.com/DataDog/dd-trace-js/pull/6861/files#diff-b754d035821154b01ab2b0d5c8429c31814948e7ebc238be7cb4cec4d6fd9003R206

packages/dd-trace/src/plugins/util/url.js
IlyasShabi: Yes indeed! All comments in this PR are from the RFC to match the spec. I added `(>=10)` for more clarification

rochdev (approved):

bengl (approved):

PR #6862, opened 2025-11-06
chore: add appsec packages to dependabot ignore list

simon-id:
### What does this PR do?
Add AppSec native addons to the dependabot ignore list

### Motivation
We always release these package manually, so we don't need any reminder. If they get a new release it's because we're already working on it


BridgeAR (commented):
I think the original intention behind the exclude pattern was to actually add them to the ignore list. Please therefore move the other libraries as well :)

PR #6889, opened 2025-11-11
chore: add types to the debugger logger

watson:
### What does this PR do?

Add types to the logger used in the debugger worker thread.

### Motivation

Just a fly-by change to make the linter happy.

### DIFF
--- a/integration-tests/debugger/template.spec.js
+++ b/integration-tests/debugger/template.spec.js
@@ -1,8 +1,11 @@
 'use strict'
 
 const { assert } = require('chai')
+const semver = require('semver')
 const { setup } = require('./utils')
 const { NODE_MAJOR } = require('../../version')
+
+const NODE_24_11_1_OR_LATER = semver.gte(process.version, '24.11.1')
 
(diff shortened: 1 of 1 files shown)


packages/dd-trace/src/debugger/devtools_client/log.js
```diff
@@ -1,10 +1,15 @@
 'use strict'
 
+/** @typedef {'error'|'warn'|'info'|'debug'} Level */
+/** @typedef {(...args: any[]) => void} LogFn */
```
BridgeAR: Could we use unknown as args type? Or use a generic?

watson: I'm confused, why isn't `void` the correct return value?

BridgeAR: Sorry, I mistyped. I meant the arguments.

watson: Updated to use `unknown[]`

PR #6890, opened 2025-11-11
chore(lint): fix a few minor local ESLint errors

watson:
### What does this PR do?

- Add the root `.bun` directory to the global ESLint ignore-list. This is used by Bun as a cache and might appear locally on the developer machine if they run Bun. It's already in `.gitignore`.
- Disable two places in the integration tests where ESLint was complaining about `n/no-extraneous-require` where `dd-trace` was being required. Not sure why this wasn't blowing up in CI?!?

### Motivation

Make the local dev-experience better.

### Additional Notes

I'm very curious why the two integration test issues wasn't a problem in CI, but only locally.

### DIFF
--- a/integration-tests/appsec/iast-esbuild/app.js
+++ b/integration-tests/appsec/iast-esbuild/app.js
@@ -1,6 +1,6 @@
 'use strict'
 
-require('dd-trace').init()
+require('dd-trace').init() // eslint-disable-line n/no-extraneous-require
 
 const express = require('express')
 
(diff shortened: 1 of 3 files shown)


integration-tests/appsec/iast-esbuild/app.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-require('dd-trace').init()
+require('dd-trace').init() // eslint-disable-line n/no-extraneous-require
```
BridgeAR, line 3: These actually fail on the CI (they are not reported there)

We have more local reports, I did not identify the root cause of the difference so far either.

watson: 🤦 annoying... I'll close update the PR to just concern itself with `.bun`.

PR #6891, opened 2025-11-11
chore: improve instrumentation hooks

BridgeAR:
This improves our instrumentations in multiple ways:

- Add JSDoc to many methods that it is easier to work with.
- Fix multiple type errors.
- Remove code that was only used in testing such as `unhook()`.
- Fix small issues in esbuild instrumentation where a module might not return a moduleExports object.
- Improve code to run slightly faster (these are not hot code paths).
- Automatically instruments unprefixed Node.js modules when defining a instrumentation for a module. This removes the need for much special handling (duplicating the instrumentation, using the symbol to detect already instrumented calls, etc.).
- Fix using an array as file for some modules. This did not cause any issues, but it was not correct.
- Removed unused isIitm arguments in instrumentations.

### DIFF
--- a/packages/datadog-instrumentations/src/helpers/register.js
+++ b/packages/datadog-instrumentations/src/helpers/register.js
@@ -140,7 +140,7 @@ for (const name of names) {
         }
 
         if (matchVersion(version, versions)) {
-          allInstrumentations[instrumentationFileName] ||= false
+          allInstrumentations[instrumentationFileName] = true
 
           try {
             loadChannel.publish({ name })


packages/datadog-instrumentations/src/helpers/instrument.js
```diff
@@ -24,23 +43,19 @@ exports.tracingChannel = function (name) {
 
 /**
  * @param {object} args
- * @param {string|string[]} args.name module name
- * @param {string[]} args.versions array of semver range strings
+ * @param {string} args.name module name
```
simon-id, line 46: why remove the possibility to have multiple modules on the same hook ?

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -38,23 +38,39 @@ if (!disabledInstrumentations.has('process')) {
...
 const allInstrumentations = {}
+/** @type {Set<string>} */
+const instrumentedNodeModules = new Set()
+
+for (const name of names) {
+  if (name.startsWith('node:')) {
+    // Add all unprefixed node modules to the instrumentations list.
+    names.push(name.slice(5))
+    // Always disable prefixed and unprefixed node modules if one is disabled.
+    // TODO: Activate & write a regression test for this.
```
simon-id, line 58: _μηδ’ ἀναβάλλεσθαι ἔς τ’ αὔριον ἔς τε ἔνηφι_

BridgeAR: I wanted to do that in a separate PR as it changes behavior

packages/datadog-instrumentations/src/helpers/instrument.js
BridgeAR: This is not how it worked. It created multiple hook entries, each with an individual name but the same hook method.
That was to overcome handling the `node:` prefix issue which is now automatically handled in all cases.

simon-id: >This is not how it worked. It created multiple hook entries, each with an individual name but the same hook method.

yeah this is what i mean. So it's just not used anywhere anymore ? i guess we can remove it but could be nice to keep the code similar enough that we can reimplement it later if we have a new usecase for this syntax :+1:

packages/datadog-instrumentations/src/helpers/register.js
simon-id: aaah, you meant write a test for the code in the comment under, i see :+1:

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -109,65 +124,55 @@ for (const packageName of names) {
...
 
         if (matchVersion(version, versions)) {
-          allInstrumentations[instrumentationFileName] = true
-
-          // Check if the hook already has a set moduleExport
-          if (hook[HOOK_SYMBOL].has(moduleExports)) {
-            namesAndSuccesses[`${name}@${version}`] = true
-            return moduleExports
-          }
+          allInstrumentations[instrumentationFileName] ||= false
```
simon-id: so this variable used to be set to true even when it failed, was that causing a bug ? so now you moved the = true at the end to prevent that right ?

BridgeAR:
Closing as a follow-up PR includes these changes

PR #6897, opened 2025-11-12
feat(debugger): add snapshot time budget

watson:
### What does this PR do?

- Enforces a per-snapshot time budget for Dynamic Instrumentation debugger snapshots.
- Default budget is 15ms; configurable via:
  - Env var: `DD_DYNAMIC_INSTRUMENTATION_CAPTURE_TIMEOUT_MS`
  - Programmatic: `tracer.init({ dynamicInstrumentation: { captureTimeoutMs }})`
- When the budget is exceeded, remaining values that are not already resolved or in the process of being resolved (awaiting a `session.post` promise), are marked with `notCapturedReason: 'timeout'`.
- Enforce max concurrency when resolving object properties (previously objects would have their properties resolved in parallel, now we enforce a max concurrency of two)
- Adds unit and integration tests to validate partial snapshots under tight budgets.

### Motivation

Prevent snapshot collection from exceeding acceptable latency and resource usage in complex/large object graphs.

### Additional Notes

~Depends on https://github.com/DataDog/system-tests/pull/5705~

Not covered by this PR:
- A single CDP round trip can exceed the time budget significantly, e.g. calling `Runtime.getProperties` on an array with a million elements will take much much longer than 15ms. I'll address this in follow up PRs (https://github.com/DataDog/dd-trace-js/pull/6912).


packages/dd-trace/src/supported-configurations.json
```diff
@@ -62,6 +62,7 @@
     "DD_DBM_PROPAGATION_MODE": ["A"],
     "DD_DOGSTATSD_HOST": ["A"],
     "DD_DOGSTATSD_PORT": ["A"],
+    "DD_DYNAMIC_INSTRUMENTATION_CAPTURE_TIMEOUT_MS": ["A"],
```
tylfin, line 65: Should we map this to the older value for backwards compatibility and system-tests sake?

tylfin (dismissed):

packages/dd-trace/src/supported-configurations.json
watson: A different environment variable name? Not sure I understand?

tylfin: Yeah, I mean mapping `DD_DYNAMIC_INSTRUMENTATION_CAPTURE_TIMEOUT` to `DD_DYNAMIC_INSTRUMENTATION_CAPTURE_TIMEOUT_MS` but keeping it internal

tylfin: We talked about it in Slack, the other tracers don't have a well-defined value for this, so there's no point to make it backwards compatible

watson: Result of Slack discussion: The other language tracers don't currently align on this name either, so there's no benefit in trying to support the name `DD_DYNAMIC_INSTRUMENTATION_CAPTURE_TIMEOUT` as well just for backwards compatibility with tracer configs from other languages

packages/dd-trace/src/debugger/devtools_client/snapshot/processor.js
```diff
@@ -195,6 +201,9 @@ function toMap (type, pairs, maxLength) {
       // This can be skipped and we can go directly to its children, of which
       // there will always be exactly two, the first containing the key, and the
       // second containing the value of this entry of the Map.
+      if (Object.hasOwn(value, timeBudgetSym)) {
```
shatzi: I think we can reduce the snapshot size by simply filter those items out and put noCaptureReason: 'timeout' on the map/collection object itself. same as size capture limit.

watson: I agree. I didn't like this bit either. I took a second look and found a better implementation that addresses your concern.

integration-tests/debugger/utils.js
```diff
@@ -53,14 +53,16 @@ function setup ({ env, testApp, testAppSource, dependencies, silent, stdioHandle
...
+  async function triggerBreakpoint (url) {
     let triggered = false
-    t.agent.on('debugger-diagnostics', ({ payload }) => {
-      payload.forEach((event) => {
-        if (!triggered && event.debugger.diagnostics.status === 'INSTALLED') {
-          triggered = true
-          t.axios.get(url)
-        }
+    return new Promise((resolve, reject) => {
+      t.agent.on('debugger-diagnostics', ({ payload }) => {
```
BridgeAR, line 59: Could this trigger multiple times? If yes, do we expect it to be in any of the payloads or always in the first?
I would normally switch to `once`.

integration-tests/debugger/utils.js
```diff
@@ -53,14 +53,16 @@ function setup ({ env, testApp, testAppSource, dependencies, silent, stdioHandle
...
+        payload.forEach((event) => {
+          if (!triggered && event.debugger.diagnostics.status === 'INSTALLED') {
+            triggered = true
```
BridgeAR, line 62: Depending on if we only ever care about the very first event, we could use some to not have to iterate through all payloads. That way the `triggered` variable would not be needed.

integration-tests/debugger/utils.js
```diff
@@ -53,14 +53,16 @@ function setup ({ env, testApp, testAppSource, dependencies, silent, stdioHandle
   }
 
   // Trigger the breakpoint once probe is successfully installed
-  function triggerBreakpoint (url) {
+  async function triggerBreakpoint (url) {
```
BridgeAR, line 56: This actually adds a single tick without any other benefit (besides marking the method as an async one).

Could you add JSDoc as well? :)

packages/dd-trace/src/debugger/devtools_client/index.js
```diff
@@ -167,7 +167,13 @@ session.on('Debugger.paused', async ({ params }) => {
   // TODO: Create unique states for each affected probe based on that probes unique `capture` settings (DEBUG-2863)
   const processLocalState = numberOfProbesWithSnapshots !== 0 && await getLocalStateForCallFrame(
     params.callFrames[0],
-    { maxReferenceDepth, maxCollectionSize, maxFieldCount, maxLength }
+    {
+      maxReferenceDepth,
+      maxCollectionSize,
+      maxFieldCount,
+      maxLength,
+      deadlineNs: start + BigInt(config.dynamicInstrumentation.captureTimeoutMs ?? 10) * 1_000_000n
```
BridgeAR: I believe the default is already defined in the config. We could also create the BigInt in the config directly (and also define it as default). That seems best in my opinion.

```suggestion
      deadlineNs: start + config.dynamicInstrumentation.captureTimeoutMs * 1_000_000n
```

packages/dd-trace/src/debugger/devtools_client/snapshot/index.js
```diff
@@ -17,26 +19,44 @@ function returnError () {
...
-      rawState.push(...await getRuntimeObject(
-        scope.object.objectId,
-        { maxReferenceDepth, maxCollectionSize, maxFieldCount }
-      ))
-    }))
+    for (const scope of callFrame.scopeChain) {
+      if (opts.deadlineReached === true) break // TODO: Variables in scope are silently dropped: Not the best UX
+      if (scope.type === 'global') continue // The global scope is too noisy
+      // eslint-disable-next-line no-await-in-loop
+      rawState.push(...await getRuntimeObject(scope.object.objectId, opts))
```
BridgeAR: Is there a reason that this should be run sequential? If yes, I think it would be good to add a comment why.

packages/dd-trace/src/debugger/devtools_client/snapshot/collector.js
```diff
@@ -189,3 +233,9 @@ function removeNonEnumerableProperties (props) {
     }
   }
 }
+
+function overBudget (opts) {
+  if (opts.deadlineReached) return true
```
BridgeAR: I believe it should be impossible for this to be true, if not defined by a user, which it should not?

packages/dd-trace/src/debugger/devtools_client/snapshot/collector.js
```diff
@@ -10,6 +10,28 @@ module.exports = {
   getRuntimeObject: getObject
 }
 
+/**
+ * @typedef {Object} GetObjectOptions
+ * @property {Object} maxReferenceDepth - The maximum depth of the object to traverse
+ * @property {number} maxCollectionSize - The maximum size of a collection to include in the snapshot
+ * @property {number} maxFieldCount - The maximum number of properties on an object to include in the snapshot
+ * @property {bigint} deadlineNs - The deadline in nanoseconds compared to `process.hrtime.bigint()`
+ * @property {boolean} [deadlineReached=false] - Whether the deadline has been reached. Should not be set by the
```
BridgeAR: If it should not be set by a user, let us not document it here? And if I understand it correct, it is just a way to transport the information to other code parts. Is there any other way to transport that information?

packages/dd-trace/src/debugger/devtools_client/snapshot/collector.js
```diff
@@ -43,26 +65,48 @@ async function traverseGetPropertiesResult (props, opts, depth) {
...
     }
   }
 
-  if (promises.length) {
-    await Promise.all(promises)
+  if (work.length) {
+    // Iterate over the work in chunks of 2. The closer to 1, the less we'll overshoot the deadline, but the longer it
+    // takes to complete. `2` seems to be the best compromise.
+    // Anecdotally, on my machine, with no deadline, a concurrency of `1` takes twice as long as a concurrency of `2`.
+    // From thereon, there's no real measureable savings with a higher concurrency.
```
BridgeAR, line 97: While I understand the reasoning, it is a very machine specific measurement that could easily change on a different one.

I did work on an auto scaling feature at some point to achieve something like this where all machine resources should be used depending on the machine load. Defining this number as 2 will likely be lower than what a couple of servers could handle.

packages/dd-trace/src/debugger/devtools_client/snapshot/processor.js
```diff
@@ -74,11 +74,13 @@ function getPropertyValueRaw (prop, maxLength) {
 }
 
 function getObjectValue (obj, maxLength) {
+  const timeBudgetReached = Object.hasOwn(obj, timeBudgetSym)
```
BridgeAR: We could just check for existence with not `undefined` instead of `hasOwn()`. That should be faster.

packages/dd-trace/src/debugger/devtools_client/snapshot/processor.js
```diff
@@ -209,8 +216,10 @@ function toMap (type, pairs, maxLength) {
   return result
 }
 
-function toSet (type, values, maxLength) {
+function toSet (type, values, maxLength, timeBudgetReached) {
+  if (timeBudgetReached === true) return notCapturedTimeBudget(type)
   if (values === undefined) return notCapturedDepth(type)
+  if (values.every(({ value }) => Object.hasOwn(value, timeBudgetSym))) return notCapturedTimeBudget(type)
```
BridgeAR: This is actually handled different for arrays and objects. Is that intentional? I would have expected to work the same in all of these cases.

integration-tests/debugger/utils.js
watson: It will trigger many times. But we don't know for sure which time the event we're looking for will be emitted

integration-tests/debugger/utils.js
watson: This is test code, so I optimize for readability rather than performance. But I can add the JSDoc

packages/dd-trace/src/debugger/devtools_client/snapshot/collector.js
watson: It can be set to `true` a few lines below. So when `overBudget` is called again later, I want to not incur the overhead of calling hrtime again. The reason why `overBudget` can be called again even though we're already over budget, is because some CDP requests might already be queued up before the budget was used up. Once those resolve, code will be executed that call `overBudget` again.

packages/dd-trace/src/debugger/devtools_client/snapshot/index.js
watson: Yes, otherwise we'll queue up a lot of CDP requests that all need to complete even if deadline has been reached, which means we risk overshooting the deadline significantly. Plus, there's no performance improvements gained by running it in parallel since the Chrome DevTools Protocol pushes everything over a serial socket anyway. That's also why I changed the `Promise.all` code inside of the `collector.js` file to only ever run two CDP requests in parallel. Anything higher than that would not have a measurable performance improvement. Anything lower would however take twice as long to run. I assume that's because there's some savings to be gained internally in CDP if a little work is queued up.

packages/dd-trace/src/debugger/devtools_client/snapshot/collector.js
BridgeAR: Maybe worth a comment? But not mandatory.

integration-tests/debugger/utils.js
watson: I ended up adding types to the entirety of the debugger test utils, and the changes have become quite extensive, so I'll add those in follow-up PR to keep this one clean

integration-tests/debugger/utils.js
watson: I can use `some` if you think it's important? But out of all the debugger integration tests, there's currently only a single one where the `payload` array has two elements - the rest of the time there's only a single element.

But in any case, the `triggered` boolean will be needed if `INSTALLED` is ever emitted twice, which is possible (though I can't remember off the top of my head if we have a test that does that).

packages/dd-trace/src/debugger/devtools_client/index.js
watson: Do you mean convert `captureTimeoutMs` to nanoseconds in the config? So we still take `captureTimeoutMs` as the `init` input, but compute a separate `captureTimeoutNs` at config time, which we reference here?

packages/dd-trace/src/debugger/devtools_client/snapshot/collector.js
watson: If we don't document it here, TypeScript will complain when we reference it, right?

But yes, it's just a way to transport the information out to the caller, so that when they pass in the same `opts` object again later in a subsequent call to `getRuntimeObject`, this information is retained, so that we don't have to re-compute it. I agree this is not a good programming pattern, but I felt it was the one that had the best performance for what we wanted. But I'd be happy to consider alternatives 😊

Another approach I considered was a context object that could be passed around. But since we already pass the `opts` object around all over code, I felt it was easier to just re-use that. But maybe we could make it more obvious by nesting it under a `ctx` property. So instead of `deadlineReached` being directly on the root of the `opts` object, it could be `ctx.deadlineReached`.

packages/dd-trace/src/debugger/devtools_client/snapshot/collector.js
watson: The Chrome DevTools Protocol (CDP) pushes everything over a serial socket, so there shouldn't actually be anything gained making this code run in parallel. I am hypothesising that the reason we see a large improvement going from a concurrency of `1` to `2`, is because CDP has an internal buffer which is optimizing some things as long as it's never empty.

packages/dd-trace/src/debugger/devtools_client/snapshot/processor.js
watson: Yeah, it's a bit ugly. It's because sets and maps has some internal fields that we need to traverse before getting to the actual content of the sets and maps. Arrays doesn't have this extra internal structure. I've tried to explain it in the large comment below and in a similar comment in the `toMap` function.

When the time budget is used up before we get a chance to traverse this internal structure, we end up in a situation where its only partially resolved, and hence there's no `properties` property, and so we need to either re-write the code expecting the `properties` property to exist, or bail before we get to that (which is what I opted for here).

packages/dd-trace/src/debugger/devtools_client/snapshot/processor.js
watson: Fixed 👍

packages/dd-trace/src/debugger/devtools_client/snapshot/collector.js
watson: I fixed this in the latest commit - both the JSDoc and added the `ctx` property

packages/dd-trace/src/debugger/devtools_client/index.js
watson: We unfortunately can't store BigInts in the config because those can't be serialized. We rely on them being serialized, both to send them to instrumentation telemetry and to the debugger worker thread

watson: I found an alternative: After the millisecond value is serialized and sent to the debugger worker thread, I convert it to a nanosecond BigInt once on worker thread startup. From thereon I re-use that value inside the worker thread. I've updated the PR with that approach.

At the same time I also changed the default value from being 10ms to 15ms. Through testing, I found that 10ms was just a bit too agressive of a budget.

integration-tests/debugger/utils.js
watson: I will refactor the `triggerBreakpoint` in a separate PR to address your comment

packages/dd-trace/src/debugger/devtools_client/index.js
BridgeAR: Would a budget of e.g., 25 ms or even 50 ms not also still be in a fine range?

That way we are probably safer when it comes to not hiding information the user might be interested in by default? Especially on a slow machine or in case of a heavy workload for the machine.

watson: Maybe. I was going for worst case scenario. The debugger code already has the following limits build in:
- A single probe can't take a snapshot more than once a seconds
- The entire process can have more than one probe active at the same time, so we allow for up to 25 snapshots to be taken per second across all probes

So in a worst case scenario where you have 25 probes all in hot code paths, the main thread will be paused for 25*(15ms + 2ms overhead) = approx. 0.5 sec/sec

PR #6899, opened 2025-11-12
chore: introduce jsdoc linter

BridgeAR:
The linter ensures that our added JSDoc is using valid types that can be picked up by the IDE. A couple of rules are either manually fixed or are deactivated that would be too strict for now. Activate more rules over time.

The linter is actually producing warnings instead of errors. This is helpful to differentiate the JSDoc issues from other linter errors. It will still fail the CI due to not accepting warnings.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -983,12 +983,7 @@
   resolved "https://registry.yarnpkg.com/@types/yoga-layout/-/yoga-layout-1.9.2.tgz#efaf9e991a7390dc081a0b679185979a83a9639a"
   integrity sha512-S9q47ByT2pPvD65IvrWp7qppVMpk9WGMbVq9wbWZOHg6tnXSD4vyhao6nOSBwwfDdV2p3Kx9evA9vI+XWTfDvw==
 
-"@typescript-eslint/types@^8.46.0":
-  version "8.46.4"
-  resolved "https://registry.yarnpkg.com/@typescript-eslint/types/-/types-8.46.4.tgz#38022bfda051be80e4120eeefcd2b6e3e630a69b"
-  integrity sha512-USjyxm3gQEePdUwJBFjjGNG18xY9A2grDVGuk7/9AkjIF1L+ZrVnwR5VAU5JXtUnBL/Nwt3H31KlRDaksnM7/w==
-
-"@typescript-eslint/types@^8.47.0":
+"@typescript-eslint/types@^8.46.0", "@typescript-eslint/types@^8.47.0":
   version "8.47.0"
   resolved "https://registry.yarnpkg.com/@typescript-eslint/types/-/types-8.47.0.tgz#c7fc9b6642d03505f447a8392934b9d1850de5af"
   integrity sha512-nHAE6bMKsizhA2uuYZbEbmp5z2UpffNrPEqiKIeN7VsV6UY/roxanWfoRrf6x/k9+Obf+GQdkm0nPU+vnMXo9A==


eslint.config.mjs
```diff
@@ -35,6 +36,7 @@ export default [
   {
     name: 'dd-trace/global-ignore',
     ignores: [
+      '**/.bun', // Ignore autogenerated bun files
```
BridgeAR, line 39: This is an independent fix. Running the linter with these entries there fails the linter and that makes it hard locally.

integration-tests/ci-visibility/unskippable-test/test-unskippable.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jsdoc/valid-types, jsdoc/check-tag-names */
 /**
  * @datadog {"unskippable": true}
```
BridgeAR, line 4: @juan-fernandez I guess we support this notation regularly?

JSDoc linter will complain it is not a valid type and I guess if we support this notation, customers could also run into a similar situation where it is difficult to make this value accepted in JSDocEslintPlugin. Could we wrap it in an object notation? That way it would be regular legit syntax.

packages/dd-trace/src/lambda/runtime/ritm.js
```diff
@@ -53,8 +53,7 @@ function _extractModuleNameAndHandlerPath (handler) {
   const FUNCTION_EXPR = /^([^.]*)\.(.*)$/
   const match = handler.match(FUNCTION_EXPR)
   if (!match || match.length !== 3) {
-    // Malformed Handler Name
-    return // TODO: throw error
+    throw new Error(`Malformed handler name: ${handler}`)
```
BridgeAR, line 56: The way we use this API would have always caused an error, since we destructure the return value.
Thus, throwing an error seems better than the current behavior.

eslint.config.mjs
```diff
@@ -56,6 +58,7 @@ export default [
     ]
   },
   { name: '@eslint/js/recommended', ...eslintPluginJs.configs.recommended },
+  { name: 'jsdoc/recommended', ...eslintPluginJSDoc.configs['flat/recommended'] },
```
watson: We only add the names for the configs that doesn't ship with one. In that past ESLint configs didn't have names. Now they do, so almost all newer ESLint configs come with one. I usually load it up in the REPL and type `eslintPluginJSDoc.configs['flat/recommended'].name` to check. So you can simplify this line to:

```suggestion
  eslintPluginJSDoc.configs['flat/recommended'],
```

packages/dd-trace/test/debugger/devtools_client/condition-test-cases.js
```diff
@@ -845,7 +848,7 @@ function createClassWithOverloadedMethodInPrototypeChain (Builtin, propName) {
  *  collectionOperations: TestCase[],
  *  membershipAndMatching: TestCase[],
  *  typeAndDefinitionChecks: TestCase[]
- * }} */
+  }} */
```
watson: I think this looks nicer

```suggestion
 * }} 
 */
```

packages/dd-trace/test/debugger/devtools_client/breakpoints.spec.js
```diff
@@ -463,8 +463,9 @@ describe('breakpoints', function () {
  * @param {object} [config] Optional configuration object.
  * @param {string} [config.id='probe-1'] The probe ID.
  * @param {number} [config.version=1] The probe version.
- * @param {object} [config.where={ sourceFile: 'test.js', lines: ['10'] }] The location information.
- * @param {object} [config.when={ json: { eq: [{ ref: 'foo' }, 42] }, dsl: 'foo = 42' }] The condition for the probe.
+ * @param {object} [config.where = { sourceFile: 'test.js', lines: ['10'] }] The location information.
+ * @param {object} [config.when] The condition for the probe.
+ *  { json: { eq: [{ ref: 'foo' }, 42] }, dsl: 'foo = 42' } by default.
```
watson: In all our other JSDoc comments where we wrap a line, we always indent by two spaces on the 2nd line:

```suggestion
 *   { json: { eq: [{ ref: 'foo' }, 42] }, dsl: 'foo = 42' } by default.
```

BridgeAR: We could likely find a rule for that as well 😆

eslint.config.mjs
BridgeAR: I also removed it from @eslint/js/recommended. I just didn't check that closely.

packages/dd-trace/src/plugins/plugin.js
```diff
@@ -118,18 +118,16 @@ module.exports = class Plugin {
...
      */
-    const wrappedHandler = function () {
+    const wrappedHandler = (...args) => {
       try {
-        return handler.apply(this, arguments)
-      } catch (e) {
-        logger.error('Error in plugin handler:', e)
-        logger.info('Disabling plugin: %s', plugin.id)
-        plugin.configure(false)
+        return handler(...args)
```
watson: I can't immediately see if `handler` will be bound to the correct object here compared to previously. It might be ok, but should be verified by someone more familiar with this part of the code

BridgeAR: This is indeed a regression. While looking into it, I found a bug here, so I went ahead and opened a separate PR to fix that: https://github.com/DataDog/dd-trace-js/pull/6947

I will push a local fix to also handle the instance correct, while I only added the test in the other PR.

watson (approved):

PR #6901, opened 2025-11-12
fix(debugger): abort on unknown Debugger.paused reason

watson:
### What does this PR do?

Terminate the debugger in case of unexpected `Debugger.paused` reason.

### Motivation

When the `Debugger.paused` event is emitted in the debugger implementation, it's expected to be because a breakpoint is hit. In that case the events `params.reason` will be set to `other`.
    
However, if the inspected thread experiences an OOM event, `Debugger.paused` will also be emitted with `params.reason` set to `OOM`. There might also be other situations where `params.reason` is different from `other`. In all these situations we should abort and close the debugger.
    
In the case of an OOM event in the inspected thread, exiting the debugging thread, will also stop the debugging session, and this will allow the inspected thread to properly exit with an OOM error. If we didn't do this we would essentially swallow the OOM error.

### DIFF
--- a/packages/dd-trace/src/debugger/devtools_client/index.js
+++ b/packages/dd-trace/src/debugger/devtools_client/index.js
@@ -59,9 +59,8 @@ session.on('Debugger.paused', async ({ params }) => {
   const start = process.hrtime.bigint()
 
   if (params.reason !== 'other') {
-    log.error(`[debugger:devtools_client] Unexpected Debugger.paused reason: ${params.reason}`)
-    // It's ok to use process.exit here: This will just exit the worker thread, not the main process.
-    process.exit(1) // eslint-disable-line unicorn/no-process-exit
+    // This error should not be caught, and should exit the worker thread, effectively stopping the debugging session
+    throw new Error(`Unexpected Debugger.paused reason: ${params.reason}`)
   }
 
   let maxReferenceDepth, maxCollectionSize, maxFieldCount, maxLength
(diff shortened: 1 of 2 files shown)


tylfin (dismissed):

packages/dd-trace/src/debugger/devtools_client/index.js
```diff
@@ -58,6 +58,12 @@ if (SUPPORT_ARRAY_BUFFER_RESIZE) {
 session.on('Debugger.paused', async ({ params }) => {
   const start = process.hrtime.bigint()
 
+  if (params.reason !== 'other') {
+    log.error(`[debugger:devtools_client] Unexpected Debugger.paused reason: ${params.reason}`)
```
BridgeAR: This could be an async operation. Calling process.exit will therefore potentially prevent the log from being printed. I guess the solution is nevertheless still better than what we have right now.

BridgeAR (dismissed):

packages/dd-trace/src/debugger/devtools_client/index.js
watson: Yeah, I was thinking about that but didn't really have a good alternative. My initial implementation was throwing an error that would then not be caught and then exit the thread while still printing the thrown error. But that potentially introduces a race condition as the thrown error doesn't exit the process synchronously if I'm not mistaken. And in case of an OOM error in the parent thread, I wanted to quickly exit so that the parent thread could be allowed to crash.

watson: I updated the code to throw instead of calling `process.exit`

BridgeAR (approved):

PR #6904, opened 2025-11-12
fix: Propagate otel span kind to span.kind tag 

mabdinur:
### What does this PR do?
Propagates opentelemetry's span kind into datadog's `span.kind` tag.

### Motivation
Since opentelemetry tags always have span kinds defined (and if the host project doesn't specify it then it defaults to INTERNAL) then let's propagate that span kind to DD's `span.kind` tag so that it is properly visible in the UI. Otherwise we are forcing clients to add a redundant addAttribute if they want to retain the span kind after the trace is sent to datadog.

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


bm1549 (approved):
Approving this to unblock merge as this was cleared by @BridgeAR as part of the community-opened PR

> The PR itself is looking fine taking a glimpse, it mainly needs to be ported and to check if anything needs adjustment running the tests.

BridgeAR (approved):

PR #6908, opened 2025-11-13
fix(express): 2 crashes when `router[method]()` is used with no handler

IlyasShabi:
### What does this PR do?
Fixes a bug in Router instrumentation where calling HTTP methods like `router.bind('/')` without handlers causes unexpected errors due to changed argument passing behavior.

### Motivation

The PR #6271 (endpoint collection feature) changed how arguments are passed to Express route methods. This changed Express's behavior when methods are called without handlers and this will impact getting [handlers in 
  router](https://github.com/expressjs/express/blob/4.x/lib/router/route.js#L208)

Given the example `router.bind('/')` this is how router will register handlers:

```
original.call(this, fn, ...otherArgs) // handlers are [undefined] - Router validates undefined as a handler and throws

original.apply(this, arguments) // handlers are [] - nothing to do
```

Edit:

Even with the first solution we still encounter another bug related to router `stack` which is undefined in some early Express 4 versions (latest one not concerned). We had to fix this by adding optional chaining only when needed

See [SCP-950](https://datadoghq.atlassian.net/browse/SCP-950) for more details

[SCP-950]: https://datadoghq.atlassian.net/browse/SCP-950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BridgeAR (changes requested):
Please add a regression test

simon-id (dismissed):

simon-id (approved):

PR #6909, opened 2025-11-13
chore(ci): merge release workflow into one file

simon-id:
### What does this PR do?
This merges the v3, v4, v5, and dev publish workflow into one single file, which is a requirement to be able to use OIDC publish.

### Motivation
npm is deleting every classic token by the end of the week, so we won't be able to release if we don't migrate to OIDC.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,7 +6,7 @@ on:
       - v3.x
       - v4.x
       - v5.x
-      - workflow_dispatch
+  workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref || github.run_id }}


CarlesDD (dismissed):

.github/workflows/release.yml
```diff
@@ -0,0 +1,166 @@
+name: Releases
+
+on:
+  push:
+    branches:
+      - v3.x
+      - v4.x
+      - v5.x
+      - workflow_dispatch
```
uurien:
```
on: workflow_dispatch
```
is not on branch workflow_dispatch.

uurien: I guess it should be something like:
```
on:
  push:
    branches:
      - v3.x
      - v4.x
      - v5.x
  workflow_dispatch
```

simon-id: good point :laughing:

simon-id: apparently it needs to have a colon at the end of the line too don't ask me why

simon-id: https://github.com/DataDog/dd-trace-js/pull/6909/commits/94c58cbd4cd65f7b4bd2d94dfd204d36dd32bb02

.github/workflows/release.yml
```diff
@@ -0,0 +1,166 @@
...
+          git tag --force dev
+          git push origin :refs/tags/dev
+          git push origin --tags
+
+  status:
+    needs: [
+      'publish-v3',
+      'publish-v4',
+      'publish-latest',
+    ]
```
uurien, line 154: isn't this going to be waiting forever, because waits for these 3 events?

https://docs.github.com/es/actions/reference/workflows-and-actions/workflow-syntax#example-requiring-successful-dependent-jobs


> and job3 waits for both job1 and job2 to complete.

simon-id: we're not exactly sure that this works, but we'll try.
The logic is, it waits for the jobs to be either successful or skipped, and only execute, if one of them is succesful

uurien: a ok, you are adding the `always()` in the next line, ok, lets see if it works.

PR #6910, opened 2025-11-13
fix: always send telemetry for not instrumented code

BridgeAR:
This makes sure not instrumented code will always send out the
proper telemetry. Formerly, it would rely on any part of the file
being instrumented. Now, we properly check by version again. To
guarantee it is only logging it once, it is send with a timeout or
before the service ends in case that happens earlier.

It also fixes DD_TRACE_DISABLED_INSTRUMENTATIONS to work for
prefixed and unprefixed Node automatically in case either is
deactivated. Before, both would have to be deactivated.

In addition, it simplifies code, adds some JSDoc and makes sure
the Node.js version is added to Node.js modules in the telemetry /
failures.

This also fixes a couple of tests that were not properly checking
the behavior as some state was captured in the instrumentation.

This is based on another PR.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -231,6 +231,7 @@
 /integration-tests/bun/ @DataDog/lang-platform-js
 /integration-tests/init.spec.js @DataDog/lang-platform-js
 /integration-tests/package-guardrails.spec.js @DataDog/lang-platform-js
+/integration-tests/package-guardrails/flush.js @DataDog/lang-platform-js
 /integration-tests/startup.spec.js @DataDog/lang-platform-js
 
 /packages/datadog-core @DataDog/lang-platform-js


packages/datadog-esbuild/index.js
```diff
@@ -173,8 +184,8 @@ ${build.initialOptions.banner.js}`
 
     // TODO: Should this also check for namespace === 'file'?
     if (!modulesOfInterest.has(args.path) &&
-        args.path.startsWith('@') &&
-        !args.importer.includes('node_modules/')) {
+      args.path.startsWith('@') &&
+      !args.importer.includes('node_modules/')) {
```
```suggestion
        args.path.startsWith('@') &&
        !args.importer.includes('node_modules/')) {
```

packages/datadog-instrumentations/src/helpers/bundler-register.js
```diff
@@ -22,45 +23,73 @@ if (!dc.unsubscribe) {
...
-    if (!matchVersion(payload.version, versions)) continue
+  for (const { file, versions, hook } of instrumentation) {
+    if (payload.path !== filename(name, file) || !matchVersion(payload.version, versions)) {
+      continue
+    }
 
     try {
       loadChannel.publish({ name, version: payload.version, file })
-      payload.module = hook(payload.module, payload.version)
+      payload.module = hook(payload.module, payload.version) ?? payload.module
```
BridgeAR, line 114: Note: we might have to pass through if iitm is used or not for prisma

packages/datadog-instrumentations/src/helpers/instrument.js
```diff
@@ -34,24 +53,19 @@ exports.getHooks = function getHooks (names) {
...
- * @param {string|string[]} args.name module name
- * @param {string[]} args.versions array of semver range strings
+ * @param {string} args.name module name
+ * @param {string[]} [args.versions] array of semver range strings
  * @param {string} [args.file='index.js'] path to file within package to instrument
  * @param {string} [args.filePattern] pattern to match files within package to instrument
- * @param {boolean} [args.patchDefault] whether to patch the default export
- * @param {(moduleExports: unknown, version: string) => unknown} hook
+ * @param {boolean} [args.patchDefault=true] whether to patch the default export
+ * @param {import('./instrumentations').Hook} hook
```
BridgeAR:
```suggestion
 * @param {(moduleExports: unknown, version: string, isIitm?: boolean) => unknown
```

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -65,173 +87,120 @@ for (const packageName of names) {
     hook = hook.fn
   }
 
-  // get the instrumentation file name to save all hooked versions
-  const instrumentationFileName = parseHookInstrumentationFileName(packageName)
-
-  Hook([packageName], hookOptions, (moduleExports, moduleName, moduleBaseDir, moduleVersion, isIitm) => {
-    moduleName = moduleName.replace(pathSepExpr, '/')
+  Hook([name], hookOptions, (moduleExports, moduleName, moduleBaseDir, moduleVersion, isIitm) => {
+    if (timeout === undefined) {
```
BridgeAR:
```suggestion
    if (timeout === undefined) {
      // TODO: Replace with read on first flush
```

eslint.config.mjs
```diff
@@ -269,7 +269,8 @@ export default [
       }],
       'import/no-useless-path-segments': 'error',
       'import/no-webpack-loader-syntax': 'error',
-      'jsdoc/check-tag-names': ['error', { definedTags: ['datadog'] }],
+      'jsdoc/check-param-names': ['warn', { disableMissingParamChecks: true }],
+      'jsdoc/check-tag-names': ['warn', { definedTags: ['datadog'] }],
```
BridgeAR:
```suggestion
      'jsdoc/check-param-names': ['error', { disableMissingParamChecks: true }],
      'jsdoc/check-tag-names': ['error', { definedTags: ['datadog'] }],
```

packages/datadog-esbuild/index.js
```diff
@@ -173,8 +184,8 @@ ${build.initialOptions.banner.js}`
 
     // TODO: Should this also check for namespace === 'file'?
     if (!modulesOfInterest.has(args.path) &&
-        args.path.startsWith('@') &&
-        !args.importer.includes('node_modules/')) {
+       args.path.startsWith('@') &&
+       !args.importer.includes('node_modules/')) {
```
```suggestion
        args.path.startsWith('@') &&
        !args.importer.includes('node_modules/')) {
```

packages/datadog-instrumentations/src/helpers/register.js
```diff
@@ -65,173 +85,113 @@ for (const packageName of names) {
...
 
     return moduleExports
   })
 }
 
-function matchVersion (version, ranges) {
-  return !version || !ranges || ranges.some(range => satisfies(version, range))
-}
+globalThis[Symbol.for('dd-trace')]?.beforeExitHandlers.add(logAbortedIntegrations)
+channel('dd-trace:exporter:first-flush').subscribe(logAbortedIntegrations)
```
BridgeAR, line 158: Using a DC channel will work, while I wonder why we not just use a plain method that we export.

This cuts through all layers, I am just not sure if I find this elegant or not ideal 😆
Maybe someone else might want to chime in? @bengl @rochdev @watson

packages/datadog-esbuild/index.js
pabloerhard: I had added that change :/ adding it back.

packages/datadog-instrumentations/src/helpers/register.js
pabloerhard: Do we want to refactor this? Before conflicts keep growing

BridgeAR: Suggestion after speaking with @bengl: let's use the way that requires the minimum changes for now since this provides actual value, no matter if it is implemented one way or the other.

bengl: Let's keep this as is for now, and refactor later if/when we decide to cull the single-use diagnostics channel pattern.

PR #6912, opened 2025-11-13
feat(debugger): add special handling for very large collections/objects

watson:
### What does this PR do?

When a snapshot is being collected as part of a Dynamic Instrumentation/Live Debugger line probe, the following new logic has been added:

- Skip enumerating collections with more than 500 elements/entries. Mark with `notCapturedReason: 'Large collection with too many elements (skip threshold: 500)'` and record size.
- Collect object with more then 500 properties the first time a probe encounters them but flag the probe to prevent future snapshot attempts for that probe.

#### Error reporting

For the case where an object with more than 500 properties was detected, an evaluation error will be added to the probe result, letting the user know that no future snapshots will be collected for this version of the probe within the current process. This evaluation error will also be added to all future probe results of this version of the probe within the current process.

### Motivation

Handle edge cases where an object has a lot of properties, or where collections (arrays/sets/maps) has a lot of elements.
    
In those cases, a single Chrome DevTools Protocol `Debugger.getProperties` request will take a long time to serialize their properties/elements, which can siginificantly overshoot the time budget (as ongoing CDP requests are allowed to complete before aborting).
    
### Additional Notes

The implementation in this PR has the following downsides:

- The threshold of 500 was chosen by looking at avg. CDP request times on my own machine, but might not be the right number on other hardware.
- A user might see a significant overhead the first time a probe tries to snapshot a very large object with a lot of properties on the same level. However, it's a one time overhead (only reappearing if the Node.js process is restarted and the probe is updated/re-initialized).
- A snapshot will never contain any entries/elements from collections that has a size larger than 500 entries/elements.

### DIFF
--- a/integration-tests/debugger/snapshot-time-budget.spec.js
+++ b/integration-tests/debugger/snapshot-time-budget.spec.js
@@ -87,8 +87,8 @@ describe('Dynamic Instrumentation', function () {
             // Trigger the same breakpoint again directly
             const result2 = await t.axios.get(breakpoint.url)
             assert.ok(
-              result2.data.paused <= 5,
-              `expected thread to be paused <=5ms, but was paused for ~${result2.data.paused}ms`
+              result2.data.paused <= 50,
+              `expected thread to be paused <=50ms, but was paused for ~${result2.data.paused}ms`
             )
 
             await secondPayloadReceived


packages/dd-trace/test/debugger/devtools_client/snapshot/target-code/max-collection-size.js
```diff
@@ -1,15 +1,20 @@
 'use strict'
 
+const { LARGE_OBJECT_SKIP_THRESHOLD } = require('../../../../../src/debugger/devtools_client/snapshot/constants')
+
+// `LARGE_SIZE` is larger the default maxCollectionSize, but lower than the large object skip threshold, after which
```
tylfin: typo?

```suggestion
// `LARGE_SIZE` is larger than the default maxCollectionSize, but lower than the large object skip threshold, after which
```

packages/dd-trace/src/debugger/devtools_client/index.js
```diff
@@ -215,16 +216,23 @@ session.on('Debugger.paused', async ({ params }) => {
...
 
     if (probe.captureSnapshot) {
-      const state = processLocalState()
-      if (state instanceof Error) {
-        snapshot.captureError = state.message
-      } else if (state) {
-        snapshot.captures = {
-          lines: { [probe.location.lines[0]]: { locals: state } }
-        }
+      if (captureErrors.length > 0) {
```
IlyasShabi: I think it's safe if we init `captureErrors` with `[]` or use `captureErrors?.length` to avoid undefined behavior when `numberOfProbesWithSnapshots === 0` (if this possible)

packages/dd-trace/src/debugger/devtools_client/snapshot/index.js
```diff
@@ -50,38 +46,38 @@ async function getLocalStateForCallFrame (
...
+        'Future snapshots for existing probes in this location will be skipped until the Node.js process is restarted',
+        { cause: err } // TODO: The cause is not used by the backend
+      ))
     }
-  } catch (err) {
-    // TODO: We might be able to get part of the scope chain.
-    // Consider if we could set errors just for the part of the scope chain that throws during collection.
-    log.error('[debugger:devtools_client] Error getting local state for call frame', err)
-    return returnError
+    if (opts.ctx.deadlineReached === true) break // TODO: Bad UX; Variables in remaining scopes are silently dropped
```
IlyasShabi: This should be in a finally block no?

IlyasShabi: Q: why not adding an error to `captureErrors` when we reach the deadline?

watson: > This should be in a finally block no?

Since we're not re-throwing inside of the `catch` block, having no `finally` block vs having a `finally` block is functionally equivalent: The first line of code after the end of the `catch` block is going to be executed after either the `try` or the `catch` block successfully ends.

> Q: why not adding an error to `captureErrors` when we reach the deadline?

If the calling code sees any errors in the `captureErrors` array they will be treated as permanent errors and future capturing of snapshots for this probe will be disabled. We don't want that, as reaching the deadline is not seen as a permanent error, but just as a normal stop-gab to keep the overhead low. So we don't want to add an error to the `captureErrors` array in case we reach the deadline.

packages/dd-trace/src/debugger/devtools_client/index.js
watson: Good call! I fixed it in the latest commit 👍

packages/dd-trace/src/debugger/devtools_client/snapshot/collector.js
```diff
@@ -51,6 +54,13 @@ async function collectObjectProperties (objectId, opts, depth = 0, collection =
   } else if (result.length > opts.maxFieldCount) {
     // Trim the number of properties on the object if there's too many.
     const size = result.length
+    if (size > LARGE_OBJECT_SKIP_THRESHOLD) {
+      opts.ctx.captureErrors.push(new Error(
```
tylfin, line 58: Out of curiosity, do these errors get bubbled to customers? It feels like this should fall into a "warning" category since there isn't really an action for them to take unless we want to disable probes that hit too many of these conditions

watson: They are added as evaluation errors to the probe result. Both to the current probe result and to all subsequent probe results of the current probe version, as future invocations of this probe will have snapshot collection disabled. In the future we need a better way to communicate errors like these to the backend, but right now it's either as evaluation errors or as a error on the diagnostics endpoint. The latter would not work as it would be cleared the a few milliseconds later as the probe result is sent.

Issue #6882, opened 2025-11-10
[BUG]: @opentelemetry/api version compatibility issue: TracerProvider produces no-op spans with versions >1.4.1


norelina:
### Tracer Version(s)

5.76.0

### Node.js Version(s)

22.16

### Bug Report

When using dd-trace (v5.76.0) with @opentelemetry/api versions newer than 1.4.1 (e.g., 1.9.0), the TracerProvider registration appears to succeed, but all custom spans created via the OpenTelemetry API become no-ops with trace IDs showing as all zeros (00000000000000000000000000000000).

**Environment**

- dd-trace version: 5.76.0
- @opentelemetry/api version: 1.9.0 (fails), 1.4.1 (works)
- Node.js version: 22.16.0
- Platform: Windows/Linux
- Runtime: Node.js with TypeScript

**Expected Behavior**
According to [Datadog Node.js Custom Instrumentation using the OpenTelemetry API docs](https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/nodejs/otel/#adding-span-tags) when registering dd-trace's TracerProvider with the OpenTelemetry API, custom spans created using trace.getTracer() should:

- Be properly instrumented and sent to Datadog
- Have valid trace IDs that correlate with dd-trace's native instrumentation
- Maintain proper parent-child relationships with auto-instrumented spans

**Actual Behavior**
When using @opentelemetry/api@1.9.0:

- TracerProvider registration succeeds without errors
- trace.getTracer() returns a tracer instance
- tracer.startActiveSpan() appears to work
- However, all spans are no-ops with trace IDs of 00000000000000000000000000000000
- No trace data is sent to Datadog for these custom spans
- Auto-instrumentation via dd-trace continues to work normally

### Reproduction Code

**Prerequisites**
- Datadog agent running on default port: 8126
 
**Set up package.json with @opentelemetry/api @1.9.0:**
```
{
  "name": "testapp",
  "version": "1.0.0",
  "description": "Test app for dd-trace with OpenTelemetry tracing",
  "main": "dist/index.js",
  "scripts": {
    "build": "tsc",
    "start": "npx tsx app.ts",
    "clean": "rm -rf dist"
  },
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "@types/express": "^5.0.5",
    "@types/node": "^24.9.2",
    "tsx": "^4.20.6",
    "typescript": "^5.9.3"
  },
  "dependencies": {
[...]

PR #6914, opened 2025-11-14
chore(otel): make otel api a peer dependency

mabdinur:
Resolves: https://github.com/DataDog/dd-trace-js/issues/6882

### What does this PR do?

Converts `@opentelemetry/api` and `@opentelemetry/api-logs` from regular dependencies to peer dependencies, and bumps the tested OpenTelemetry API version from v1.8.0 to v1.9.0.

### Motivation

When users install a different version of `@opentelemetry/api` than what dd-trace bundles, npm can create two separate instances of the module. This breaks the OpenTelemetry singleton pattern - dd-trace sets the tracer/logger provider delegate on one instance while user code imports from the other. The result is custom spans becoming no-ops with all-zero trace IDs and log records not being captured

Making these peer dependencies ensures only one instance exists in the dependency tree, which is the standard pattern for singleton modules like OpenTelemetry API.

### Risk 

Converting Opentelemetry API has the potential to break applications that expect the dd-trace-js to bundle opentelemetry interfaces. This could be considered a breaking change.

- [x] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [x] API [documentation][3].
- [ ] CI [jobs/workflows][4].


BridgeAR (commented):
This will need the yarn lock file being updated (just running `yarn` is enough).

package.json
```diff
@@ -161,13 +159,19 @@
...
+    "@opentelemetry/api-logs": "<1.0.0"
+  },
   "devDependencies": {
     "@babel/helpers": "^7.28.4",
     "@eslint/eslintrc": "^3.3.1",
     "@eslint/js": "^9.39.0",
     "@msgpack/msgpack": "^3.1.2",
     "@openfeature/core": "^1.9.0",
-    "@openfeature/server-sdk": "~1.20.0",
+    "@openfeature/server-sdk": "^1.20.0",
```
BridgeAR: This is an independent change that breaks the yarn lock file lookup.

mabdinur:
```suggestion
    "@openfeature/server-sdk": "~1.20.0",
```

LICENSE-3rdparty.csv
```diff
@@ -7,8 +7,6 @@ require,@datadog/openfeature-node-server,Apache license 2.0,Copyright 2024 Datad
 require,@datadog/pprof,Apache license 2.0,Copyright 2019 Google Inc.
 require,@datadog/sketches-js,Apache license 2.0,Copyright 2020 Datadog Inc.
 require,@datadog/wasm-js-rewriter,Apache license 2.0,Copyright 2018 Datadog Inc.
-require,@opentelemetry/api,Apache license 2.0,Copyright OpenTelemetry Authors
-require,@opentelemetry/api-logs,Apache license 2.0,Copyright OpenTelemetry Authors
```
BridgeAR: I believe we just have to move these to the dev dependency section instead of removing them :)

package.json
```diff
@@ -161,13 +159,19 @@
...
+  },
   "devDependencies": {
     "@babel/helpers": "^7.28.4",
     "@eslint/eslintrc": "^3.3.1",
     "@eslint/js": "^9.39.0",
     "@msgpack/msgpack": "^3.1.2",
     "@openfeature/core": "^1.9.0",
     "@openfeature/server-sdk": "~1.20.0",
+    "@opentelemetry/api": "^1.9.0",
+    "@opentelemetry/api-logs": "^0.208.0",
```
BridgeAR: Please use the identical range as the peer dependency ones. Otherwise we might test something else than the user.

integration-tests/opentelemetry.spec.js
```diff
@@ -55,7 +55,7 @@ describe('opentelemetry', () => {
   let cwd
   const timeout = 5000
   const dependencies = [
-    '@opentelemetry/api@1.8.0',
+    '@opentelemetry/api@1.9.0',
```
BridgeAR: ❤️  Good catch

scripts/check_licenses.js
```diff
@@ -6,8 +6,9 @@ const readline = require('readline')
 const pkg = require(path.join(__dirname, '..', '/package.json'))
 
 const filePath = path.join(__dirname, '..', '/LICENSE-3rdparty.csv')
-const deps = new Set(Object.keys(pkg.dependencies || {}))
-const devDeps = new Set(Object.keys(pkg.devDependencies || {}))
+const peerDeps = new Set(Object.keys(pkg.peerDependencies || {}))
+const deps = new Set(Object.keys(pkg.dependencies || {}).filter(dep => !peerDeps.has(dep)))
+const devDeps = new Set(Object.keys(pkg.devDependencies || {}).filter(dep => !peerDeps.has(dep)))
```
BridgeAR: I do not think we should filter anything here. Was there a specific reason why that's done?

docs/API.md
```diff
@@ -364,7 +364,17 @@ The following tags are available to override Datadog-specific options:
 
 <h2 id="opentelemetry-api">OpenTelemetry Compatibility</h2>
 
-This library is OpenTelemetry compliant. Use the [OpenTelemetry API](https://opentelemetry.io/docs/instrumentation/js/) and the Datadog Tracer (dd-trace) library to measure execution times for specific pieces of code. In the following example, a Datadog TracerProvider is registered with @opentelemetry/api:
+This library is OpenTelemetry compliant. The OpenTelemetry API packages (`@opentelemetry/api` and `@opentelemetry/api-logs`) are peer dependencies and must be installed separately.
```
BridgeAR, line 386: Could we add a warning in case a user tries to use any of the otel APIs with a missing dependency and a potential fix?

@rochdev or @bengl PTAL. I am uncertain in what way it is fine to log the warning or not.

BridgeAR (changes requested):

LICENSE-3rdparty.csv
mabdinur: Ahhh I didn't realize we documented licenses for dev dependencies

scripts/check_licenses.js
mabdinur: Yes, there was a linting error since I removed the licenses for the opentelemetry api package. So I made this change to avoid checking licenses for peer dependencies 😆 😰. 

I should've dug a bit deeper and noticed that we DO set licenses on peer dependencies (ex: openfeatures).

package.json
```diff
@@ -141,16 +141,29 @@
     "@datadog/openfeature-node-server": "^0.3.3",
     "@datadog/pprof": "5.13.3",
     "@datadog/wasm-js-rewriter": "5.0.1",
```
mabdinur, line 153:
```suggestion
    "@datadog/wasm-js-rewriter": "5.0.1"
  },
```

docs/API.md
mabdinur: We log a warning here and return a no-op tracer provider: https://github.com/DataDog/dd-trace-js/pull/6914/changes#diff-3bd963f551b045268418623fb12bfa5864bf3422b8895bf1bd0f6185c6e1f524. 

Since we don't expose the MeterProvider and LogsProvider, just logging a warning is enough.

package.json
```diff
@@ -140,19 +140,41 @@
...
     "@datadog/openfeature-node-server": "^0.3.3",
+<<<<<<< munir/make-otel-api-a-peer-dependency
+    "@datadog/pprof": "5.13.3",
+    "@datadog/wasm-js-rewriter": "5.0.1"
+  },
+  "peerDependencies": {
+=======
     "@datadog/pprof": "5.13.4",
     "@datadog/wasm-js-rewriter": "5.0.1",
+>>>>>>> master
```
mabdinur:
```suggestion
    "@datadog/pprof": "5.13.4",
    "@datadog/wasm-js-rewriter": "5.0.1"
  },
  "peerDependencies": {
```

package.json
```diff
@@ -140,19 +140,41 @@
...
     "@openfeature/core": "^1.8.1",
     "@openfeature/server-sdk": "~1.20.0",
+<<<<<<< munir/make-otel-api-a-peer-dependency
+    "@opentelemetry/api": ">=1.0.0 <1.10.0",
+    "@opentelemetry/api-logs": "<1.0.0",
+    "@opentelemetry/core": ">=1.14.0 <1.31.0",
+    "@opentelemetry/resources": ">=1.0.0 <1.31.0",
+=======
     "@snyk/github-codeowners": "^1.1.0",
+>>>>>>> master
```
mabdinur:
```suggestion
    "@opentelemetry/api": ">=1.0.0 <1.10.0",
    "@opentelemetry/api-logs": "<1.0.0",
    "@opentelemetry/core": ">=1.14.0 <1.31.0",
    "@opentelemetry/resources": ">=1.0.0 <1.31.0",
    "@snyk/github-codeowners": "^1.1.0",
```

integration-tests/opentelemetry.spec.js
```diff
@@ -58,7 +58,7 @@ describe('opentelemetry', function () {
   let cwd = /** @type {string} */ ('')
   const timeout = 5000
   const dependencies = [
-    '@opentelemetry/api@1.8.0',
+    '@opentelemetry/api@1.9.0',
```
BridgeAR:
```suggestion
    '@opentelemetry/api',
```

This way we know when it breaks with an update :)

packages/dd-trace/src/opentelemetry/noop/tracer_provider.js
```diff
@@ -0,0 +1,49 @@
...
+const noopSpan = {
+  spanContext: () => ({}),
+  setAttribute: () => {},
+  setAttributes: () => {},
+  addEvent: () => {},
+  updateName: () => {},
+  setStatus: () => {},
+  end: () => {},
+  isRecording: () => false,
+}
```
BridgeAR: Would it be possible to prevent creating the noop tracer provider somehow so that this will just never be accessed anywhere in case it is not there?
That would be nicer, since the noop one otherwise needs to stay in sync and that is likely to deviate soon.

package.json
```diff
@@ -153,7 +153,15 @@
     "@datadog/wasm-js-rewriter": "5.0.1",
     "@opentelemetry/api": ">=1.0.0 <1.10.0",
     "@opentelemetry/api-logs": "<1.0.0",
-    "oxc-parser": "^0.121.0"
+    "oxc-parser":  "^0.118.0"
```
BridgeAR:
```suggestion
    "oxc-parser": "^0.121.0"
```

packages/dd-trace/src/opentelemetry/noop/tracer_provider.js
mabdinur: Yup, we can raise an error if TracerProvider is accessed when one isn't registered. A noop TracerProvider can allow tracing to fail silently. It better if we are noisy about it.

packages/dd-trace/test/opentelemetry/tracer_provider.spec.js
```diff
@@ -75,3 +76,27 @@ describe('OTel TracerProvider', () => {
...
+    assert.throws(() => new NoopClass(), { message: expectedMessage })
+  })
+
+  it('should throw when instantiated via noop proxy', () => {
+    const NoopProxy = proxyquire.noPreserveCache()('../../src/noop/proxy', {
+      '../opentelemetry/check_peer_deps': noopCheckPeerDeps,
+    })
+    const getter = Object.getOwnPropertyDescriptor(NoopProxy.prototype, 'TracerProvider').get
+    const NoopClass = getter.call({})
+    assert.throws(() => new NoopClass(), { message: expectedMessage })
```
BridgeAR: Could you please change these tests to ones how a user could actually reach this point? These are directly calling in internals in a way that a user would not be. With that, we are uncertain if this is indeed the triggered behavior.

packages/dd-trace/src/proxy.js
```diff
@@ -382,7 +384,18 @@ class Tracer extends NoopProxy {
...
+    if (checkOpenTelemetryAPIDeps()) {
+      return require('./opentelemetry/tracer_provider')
+    }
+    return class {
+      constructor () {
+        throw new Error(
+          '@opentelemetry/api is required to use TracerProvider. ' +
+          'Install it with: npm install @opentelemetry/api'
+        )
+      }
```
BridgeAR, line 397: I like this, it is a neat way of doing this! I am just unsure if we really want to throw or if logging a warning here would be best instead and to gracefully handle it otherwise. I am unsure, if that is an option. Handling it gracefully would probably require the noop tracer again?

I would have to know how the user could reach this to decide on what to do.

mabdinur: The main documented usage is https://github.com/DataDog/dd-trace-js/blob/de1327bc462ed2f94ecb224e1d4f44b2ee7e49af/docs/API.md?plain=1#L391:      
 
``` 
  const tracer = require('dd-trace').init()                                                                           
  const tracerProvider = new tracer.TracerProvider()        
  tracerProvider.register()     
```                                                                                      
   
IMO we should deprecate this. Instead, dd-trace should auto-register its TracerProvider during .init() when       @opentelemetry/api is present (similar to how we handle otel logs and metrics). Then users could then just do:  

```                                                                                                            
  require('dd-trace').init()                                
  // dd-trace auto-registers — use the standard OTel API from here 
  const { trace } = require("@opentelemetry/api");                                                   
  const provider = trace.getTracerProvider()   
```
                                                                                                 
This keeps dd-trace's OTel surface minimal and aligns with how the OTel ecosystem expects providers to be set up.

We can do this in a future PR

BridgeAR: Conceptually, I like that. The question I have is: how can we do that without providing loading overhead for people not using Otel? I guess it could work as a hook: the hook is our activation trigger. @rochdev what do you think?

That aside: what shall we do short term with this? Our tracer should normally handle things gracefully and this would not be the case anymore. I think this is fine as a major version, I am not yet sure about before. @rochdev @bengl opinions?
We would have to document that change though.

mabdinur:
Closing to reduce noise. Will reopen when this change is in a better place

PR #6917, opened 2025-11-14
fix: Azure Service Bus batching memory leak

jcstorms1:
### What does this PR do?

- Fixes a bug with Azure Service Bus batching where adding multiple messages resulted in a memory leak. 
- Switches to a weak map for storing spanContexts instead of using Azure's built in map

### Motivation

Subsequent calls of `tryAddMessage` on a batch results in a memory leak and the amount of spans increase unintentionally. The example below was 5 calls to the endpoint. There should be only 2 spans for `azure.servicebus.create` but the continuous repatching of the method resulted in 10 spans.

image

Also, we should not be using a built in map from Azure's sender to store our span contexts. This may lead to conflicts with their tracing.

### Plugin Checklist

- [ ] Unit tests.
- [x] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/datadog-plugin-azure-service-bus/src/producer.js
```diff
@@ -36,7 +37,11 @@ class AzureServiceBusProducerPlugin extends ProducerPlugin {
       }
 
       if (batchLinksAreEnabled()) {
-        ctx.batch._spanContexts.push(span.context())
+        if (spanContexts.has(ctx.batch)) {
+          spanContexts.get(ctx.batch).push(span.context())
+        } else {
+          spanContexts.set(ctx.batch, [span.context()])
+        }
```
watson: You can save one lookup like this:

```suggestion
        const spanContext = spanContexts.get(ctx.batch)
        if (spanContext) {
          spanContext.push(span.context())
        } else {
          spanContexts.set(ctx.batch, [span.context()])
        }
```

packages/datadog-plugin-azure-service-bus/src/producer.js
```diff
@@ -47,7 +52,8 @@ class AzureServiceBusProducerPlugin extends ProducerPlugin {
       if (isBatch) {
         span.setTag('messaging.batch.message_count', messages.count)
         if (batchLinksAreEnabled()) {
-          messages._spanContexts.forEach(spanContext => {
+          const contexts = spanContexts.get(messages) || []
+          contexts.forEach(spanContext => {
             span.addLink(spanContext)
           })
```
watson: We always prefer `for...of` instead of `forEach`:

```suggestion
          const contexts = spanContexts.get(messages)
          if (contexts) {
            for (const spanContext of contexts) {
              span.addLink(spanContext)
            }
          }
```

packages/datadog-instrumentations/src/azure-service-bus.js
```diff
@@ -8,57 +8,63 @@ const shimmer = require('../../datadog-shimmer')
 const dc = require('dc-polyfill')
 
 const producerCh = dc.tracingChannel('apm:azure-service-bus:send')
-
-addHook({ name: '@azure/service-bus', versions: ['>=7.9.2'], patchDefault: false }, (obj) => {
+const isItDefault = new WeakSet()
+addHook({ name: '@azure/service-bus', versions: ['>=7.9.2'] }, (obj) => {
```
watson, line 13: Re-add the blank line from before

```suggestion

addHook({ name: '@azure/service-bus', versions: ['>=7.9.2'] }, (obj) => {
```

watson (commented):
I only reviewed this from a performance and style point of view, but didn't test the business logic.

bengl (approved):

PR #6925, opened 2025-11-17
[test optimization] feat: support multiple docblocks

BridgeAR:
This allows other comments to be above the actual datadog docblock. This is also a requested feature in jest-docblock.

Refs: https://github.com/jestjs/jest/issues/12573

I limited the number of docblocks to read to up to 10, just in case of a weird file.

### DIFF
--- a/packages/datadog-plugin-jest/src/util.js
+++ b/packages/datadog-plugin-jest/src/util.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const { readFileSync } = require('fs')
-const { parse, extract } = require('jest-docblock')
+const { parse } = require('jest-docblock')
 
 const { getTestSuitePath } = require('../../dd-trace/src/plugins/util/test')
 const log = require('../../dd-trace/src/log')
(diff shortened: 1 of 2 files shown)


packages/datadog-plugin-jest/src/util.js
```diff
@@ -61,29 +61,59 @@ function getJestTestName (test, shouldStripSeed = false) {
...
+    const match = re.exec(testSource)
+    if (!match) break
+    const comment = match[1]
+
+    let docblocks
+    try {
+      docblocks = parse(comment)
+    } catch {
+      // Skip unparsable comment and continue scanning
+      if (commentsChecked++ >= 10) {
```
juan-fernandez, line 89: nitpick: could we extract the `10` to a variable like `MAX_NUMBER_OF_COMMENTS_TO_CHECK` or something like that? Just so that this is easier to read

BridgeAR: I will have to rebase the JSDoc PR and include it there

PR #6930, opened 2025-11-17
allow providing a custom timeout to openfeature datadog provider

leoromanovsky:
### Motivation

Timeout is hard-coded to 30 seconds; during this period a customers' nodejs server can be blocked from starting up. Those with more strict startup requirements need to be able to set a lower value.

This learning came from dogfooding in the eppo.cloud backend product.

### Plugin Checklist

- [x] Unit tests.
- [x] Integration tests.
- [ ] Benchmarks.
- [x] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/openfeature/flagging_provider.js
+++ b/packages/dd-trace/src/openfeature/flagging_provider.js
@@ -24,7 +24,8 @@ class FlaggingProvider extends DatadogNodeServerProvider {
     this._tracer = tracer
     this._config = config
 
-    log.debug(this.constructor.name + ' created with timeout: ' + config.experimental.flaggingProvider.initializationTimeoutMs + 'ms')
+    log.debug(this.constructor.name + ' created with timeout: ' +
+      config.experimental.flaggingProvider.initializationTimeoutMs + 'ms')
   }
 
   /**


packages/dd-trace/test/openfeature/flagging_provider_timeout.spec.js
```diff
@@ -173,4 +179,81 @@ describe('FlaggingProvider Initialization Timeout', () => {
     expect(readyEventSpy).to.have.been.calledOnce
     expect(provider.getConfiguration()).to.equal(ufc)
   })
+
+  describe('custom timeout configuration', () => {
+    it('should use custom timeout when specified in config', async () => {
```
typotter, line 184: test looks familiar... ;)

packages/dd-trace/test/openfeature/flagging_provider_timeout.spec.js
```diff
@@ -173,4 +179,81 @@ describe('FlaggingProvider Initialization Timeout', () => {
...
+      await clock.tickAsync(4900)
+
+      // Should still be initializing
+      expect(provider.initController.isInitializing()).to.be.true
+
+      // Advance by another 200ms to trigger the 5-second timeout
+      await clock.tickAsync(200)
+
+      // Wait for promise to settle
+      await initPromise.catch(() => {})
```
typotter, line 217: should we fail the test if the promise resolves successfully? or does that happen intrinsically because we're catching here?

index.d.ts
```diff
@@ -654,6 +654,13 @@ declare namespace tracer {
          * @default false
          */
         enabled?: boolean
+        /**
+         * Timeout in milliseconds for OpenFeature provider initialization.
+         * If configuration is not received within this time, initialization fails.
```
typotter, line 659: Not sure if these docs are the right place; the openfeature SDK can recover from the error state to `PROVIDER_READY` when a config is set.

typotter (approved):
LGTM from my end; nothing blocking just some curiosities

packages/dd-trace/test/openfeature/flagging_provider_timeout.spec.js
leoromanovsky: Yea the LLM originally made many duplicate tests, stuff I felt was already covered with existing ones. I hoped this one is canonical for the feature being developed here.

And yes, you left us a great example.

PR #6932, opened 2025-11-17
fix(openai, llmobs): update openai integration consistency

sabrenner:
### What does this PR do?
Makes the following changes to be more consistent with our Python integration:
1. Update the model name of the LLM Observability span event to be the response model. This should not break anything cost-metrics-wise but just provide more detail for the user in the DD LLM Observability UI.
2. If prompt token details are present for chat completion spans, **always** record them, even if they are 0.
3. Record metadata even if there is an error for an operation. For `responses`, since we record input and output metadata, we make sure to always record input metadata.
4. Explicitly names the keys and default values for `reasoning` roles from OpenAI Responses operations

In addition, there are some subtle tagger changes (internal API):
1. One of `content`, `toolCalls`, or `toolResults` is needed for the `llm` message object, defaulting to empty `content` if none are provided for backwards compatibility, but still erroring if one of them exists but is not the expected type.
2. Default `role` to an empty string if it does not exist. Still errors if it does exist and is not a string.

Lastly, all of these changes are then reflected in tests.

### Motivation
Consistency with our Python OpenAI auto-instrumentation. Part of an initiative introducing shared testing into system tests.

### Additional Testing
These changes are further verified with a local system tests branch that introduces the shared tests from a different repository.

```
system-tests git:(sabrenner/add-openai-llmobs-tests) ✗ ./run.sh -S INTEGRATION_FRAMEWORKS -L nodejs --weblog openai-js@6.0.0 tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs -vv
Build framework test container...
Build complete
============================================================================================================================= test context =============================================================================================================================
Scenario: INTEGRATION_FRAMEWORKS
Logs folder: ./logs_integration_frameworks
Library: nodejs@6.0.0-pre
========================================================================================================================= test session starts ==========================================================================================================================
collected 20 items                                                                                                                                                                                                                                                     



tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_chat_completion[True] PASSED                                                                                                                                                                 [  5%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_chat_completion[False] PASSED                                                                                                                                                                [ 10%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_chat_completion_error[True] PASSED                                                                                                                                                           [ 15%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_chat_completion_error[False] PASSED                                                                                                                                                          [ 20%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_completion PASSED                                                                                                                                                                            [ 25%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_completion_error PASSED                                                                                                                                                                      [ 30%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_embedding PASSED                                                                                                                                                                             [ 35%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_embedding_error PASSED                                                                                                                                                                       [ 40%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_chat_completion_tool_call[True] XFAIL (missing_feature (Node.js LLM Observability OpenAI integration does not submit tool definitions))                                                      [ 45%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_chat_completion_tool_call[False] XFAIL (missing_feature (Node.js LLM Observability OpenAI integration does not submit tool definitions))                                                     [ 50%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_responses_create[True] PASSED                                                                                                                                                                [ 55%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_responses_create[False] PASSED                                                                                                                                                               [ 60%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_responses_create_error[True] PASSED                                                                                                                                                          [ 65%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_responses_create_error[False] PASSED                                                                                                                                                         [ 70%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_responses_create_tool_call[True] XFAIL (missing_feature (Node.js LLM Observability OpenAI integration does not submit tool definitions))                                                     [ 75%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_responses_create_tool_call[False] XFAIL (missing_feature (Node.js LLM Observability OpenAI integration does not submit tool definitions))                                                    [ 80%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_responses_create_reasoning[True] PASSED                                                                                                                                                      [ 85%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_responses_create_reasoning[False] PASSED                                                                                                                                                     [ 90%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_responses_create_tool_input[True] PASSED                                                                                                                                                     [ 95%]
tests/integration_frameworks/test_openai.py::TestOpenAiLlmObs::test_responses_create_tool_input[False] PASSED                                                                                                                                                    [100%]

---------------------------------------------------------------------------------- generated xml file: /Users/sam.brenner/dd/system-tests/logs_integration_frameworks/reportJunit.xml ----------------------------------------------------------------------------------
============================================================================================================== 16 passed, 4 xfailed in 125.60s (0:02:05) ===============================================================================================================
```

### DIFF
--- a/packages/datadog-instrumentations/src/azure-service-bus.js
+++ b/packages/datadog-instrumentations/src/azure-service-bus.js
@@ -8,57 +8,64 @@ const shimmer = require('../../datadog-shimmer')
 const dc = require('dc-polyfill')
 
 const producerCh = dc.tracingChannel('apm:azure-service-bus:send')
+const isItDefault = new WeakSet()
 
-addHook({ name: '@azure/service-bus', versions: ['>=7.9.2'], patchDefault: false }, (obj) => {
+addHook({ name: '@azure/service-bus', versions: ['>=7.9.2'] }, (obj) => {
   const ServiceBusClient = obj.ServiceBusClient
-  let didItShim = false
   shimmer.wrap(ServiceBusClient.prototype, 'createSender',
     createSender => function (queueOrTopicName) {
       const sender = createSender.apply(this, arguments)
-      if (didItShim) return sender
       const senderPrototype = sender.constructor.prototype
       const senderSenderPrototype = sender._sender.constructor.prototype
-      shimmer.wrap(senderPrototype, 'scheduleMessages', scheduleMessages =>
-        function (msg, scheduledEnqueueTimeUtc) {
-          const functionName = scheduleMessages.name
+
+      if (!isItDefault.has(senderPrototype)) {
+        isItDefault.add(senderPrototype)
+
+        shimmer.wrap(senderPrototype, 'scheduleMessages', scheduleMessages =>
+          function (msg, scheduledEnqueueTimeUtc) {
+            const functionName = scheduleMessages.name
+            const config = this._context.config
+            const entityPath = this._entityPath
+            return producerCh.tracePromise(
+              scheduleMessages,
+              { config, entityPath, functionName, msg, scheduledEnqueueTimeUtc },
+              this, ...arguments
+            )
+          })
+
+        shimmer.wrap(senderPrototype, 'createMessageBatch', createMessageBatch => async function () {
+          const batch = await createMessageBatch.apply(this, arguments)
+          shimmer.wrap(batch, 'tryAddMessage', tryAddMessage => function (msg) {
+            const functionName = tryAddMessage.name
+            const config = this._context.config
+            return producerCh.tracePromise(
+              tryAddMessage, { config, functionName, batch, msg }, this, ...arguments)
+          })
+          return batch
+        })
+      }
+
+      if (!isItDefault.has(senderSenderPrototype)) {
+        isItDefault.add(senderSenderPrototype)
+
+        shimmer.wrap(senderSenderPrototype, 'send', send => function (msg) {
+          const functionName = send.name
           const config = this._context.config
-          const entityPath = this._entityPath
+          const entityPath = this.entityPath
           return producerCh.tracePromise(
-            scheduleMessages,
-            { config, entityPath, functionName, msg, scheduledEnqueueTimeUtc },
-            this, ...arguments
+            send, { config, entityPath, functionName, msg }, this, ...arguments
           )
         })
 
-      shimmer.wrap(senderPrototype, 'createMessageBatch', createMessageBatch => async function () {
-        const batch = await createMessageBatch.apply(this, arguments)
-        shimmer.wrap(batch.constructor.prototype, 'tryAddMessage', tryAddMessage => function (msg) {
-          const functionName = tryAddMessage.name
+        shimmer.wrap(senderSenderPrototype, 'sendBatch', sendBatch => function (msg) {
+          const functionName = sendBatch.name
           const config = this._context.config
+          const entityPath = this.entityPath
           return producerCh.tracePromise(
-            tryAddMessage, { config, functionName, batch, msg }, this, ...arguments)
+            sendBatch, { config, entityPath, functionName, msg }, this, ...arguments
+          )
         })
-        return batch
-      })
-
-      shimmer.wrap(senderSenderPrototype, 'send', send => function (msg) {
-        const functionName = send.name
-        const config = this._context.config
-        const entityPath = this.entityPath
-        return producerCh.tracePromise(
-          send, { config, entityPath, functionName, msg }, this, ...arguments
-        )
-      })
-
-      shimmer.wrap(senderSenderPrototype, 'sendBatch', sendBatch => function (msg) {
-        const functionName = sendBatch.name
-        const config = this._context.config
-        const entityPath = this.entityPath
-        return producerCh.tracePromise(
-          sendBatch, { config, entityPath, functionName, msg }, this, ...arguments
-        )
-      })
-      didItShim = true
+      }
       return sender
     })
   return obj
--- a/packages/datadog-plugin-azure-service-bus/src/producer.js
+++ b/packages/datadog-plugin-azure-service-bus/src/producer.js
@@ -2,6 +2,7 @@
 
 const { getEnvironmentVariable } = require('../../dd-trace/src/config-helper')
 const ProducerPlugin = require('../../dd-trace/src/plugins/producer')
+const spanContexts = new WeakMap()
 
 class AzureServiceBusProducerPlugin extends ProducerPlugin {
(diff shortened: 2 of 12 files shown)


packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -105,7 +111,7 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
       } else if (tokenUsage.prompt_tokens_details) {
         // Chat/Completions API - only include if > 0
         const cacheReadTokens = tokenUsage.prompt_tokens_details.cached_tokens
-        if (cacheReadTokens) {
+        if (cacheReadTokens != null) {
```
Yun-Kim, line 114: do we care about if it's undefined?

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -159,6 +165,16 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
   _tagChatCompletion (span, inputs, response, error) {
     const { messages, model, ...parameters } = inputs
 
+    const metadata = Object.entries(parameters).reduce((obj, [key, value]) => {
```
Yun-Kim, line 168: We're moving this here to ensure input metadata is recorded even in case of error right?

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -369,24 +380,19 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
-    if (response.temperature !== undefined) metadata.temperature = Number(response.temperature)
-    if (response.top_p !== undefined) metadata.top_p = Number(response.top_p)
-    if (response.tool_choice !== undefined) metadata.tool_choice = response.tool_choice
-    if (response.truncation !== undefined) metadata.truncation = response.truncation
-    if (response.text !== undefined) metadata.text = response.text
+    if (response.temperature !== undefined) outputMetadata.temperature = Number(response.temperature)
+    if (response.top_p !== undefined) outputMetadata.top_p = Number(response.top_p)
+    if (response.tool_choice !== undefined) outputMetadata.tool_choice = response.tool_choice
+    if (response.truncation !== undefined) outputMetadata.truncation = response.truncation
+    if (response.text !== undefined) outputMetadata.text = response.text
```
Yun-Kim, line 390: Dumb question - are these present on the response output? I thought these are typically input parameters in the request.

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -369,24 +380,19 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
     if (response.usage?.output_tokens_details?.reasoning_tokens !== undefined) {
-      metadata.reasoning_tokens = response.usage.output_tokens_details.reasoning_tokens
+      outputMetadata.reasoning_tokens = response.usage.output_tokens_details.reasoning_tokens
```
Yun-Kim, line 392: Can we make this a `reasoning_output_tokens` token metric instead of putting on the metadata?

packages/dd-trace/src/llmobs/plugins/openai.js
sabrenner: `value != null` checks for both `null` and `undefined`! using `!==` will check for only the exact match (ie `null !== undefined` is `true`, but `null != undefined` is `false`)

```
> null != undefined
false
> null !== undefined
true
```

packages/dd-trace/src/llmobs/plugins/openai.js
sabrenner: yep! before it was happening at the end, even when we might've returned early bc of an error. we still wanna capture input metadata/params even if there's an error

packages/dd-trace/src/llmobs/plugins/openai.js
sabrenner: yea it's a bit werid. if the user doesn't set them, the default values used by the model are returned (at least for OpenAI responses). we do this in our Python auto-instrumentation somehow too as the shared tests i wrote that we wrote this Node.js integration against asserted metadata parameter values that weren't specified in the input

packages/dd-trace/src/llmobs/plugins/openai.js
sabrenner: we currently don't do this for our Python integration, so maybe we can make a ticket for it and then update shared tests/implement in both language integrations? system-tests should have openai llmobs tests by next week 😄

packages/dd-trace/src/llmobs/plugins/openai.js
Yun-Kim: 😭 i love javascript

packages/dd-trace/src/llmobs/plugins/openai.js
Yun-Kim: okay good point, thanks for confirming

packages/dd-trace/src/llmobs/plugins/openai.js
Yun-Kim: Okay, let's just omit reasoning tokens for now in the metadata and we can follow up on this in a future PR with shared tests

sabrenner: i think they unfortunately need to stay for consistency w python (else in the shared/system tests we gotta do some weird checking for only python or only nodejs). gonna keep them for now if that's ok but yes i'll make a ticket to swap them over to metrics instead!

Yun-Kim: rip i didn't realize we also emit reasoning tokens in metadata in python 💀 yeah go for it

PR #6933, opened 2025-11-17
Crysmags/gemini rebase

crysmags:
### What does this PR do?
This is to add support for Google's GenAi

### DIFF
--- a/docs/test.ts
+++ b/docs/test.ts
@@ -332,6 +332,7 @@ tracer.use('fetch', httpClientOptions);
 tracer.use('generic-pool');
 tracer.use('google-cloud-pubsub');
 tracer.use('google-cloud-vertexai');
+tracer.use('google-genai');
 tracer.use('graphql');
 tracer.use('graphql', graphqlOptions);
 tracer.use('graphql', { variables: ['foo', 'bar'] });
--- a/index.d.ts
+++ b/index.d.ts
@@ -207,6 +207,7 @@ interface Plugins {
   "generic-pool": tracer.plugins.generic_pool;
   "google-cloud-pubsub": tracer.plugins.google_cloud_pubsub;
   "google-cloud-vertexai": tracer.plugins.google_cloud_vertexai;
+  "google-genai": tracer.plugins.google_genai;
   "graphql": tracer.plugins.graphql;
   "grpc": tracer.plugins.grpc;
   "hapi": tracer.plugins.hapi;
@@ -1843,23 +1844,29 @@ declare namespace tracer {
      * [@google-cloud/pubsub](https://github.com/googleapis/nodejs-pubsub) module.
      */
     interface google_cloud_pubsub extends Integration {}
-
+    
     /**
      * This plugin automatically instruments the
      * [@google-cloud/vertexai](https://github.com/googleapis/nodejs-vertexai) module.
-     */
-    interface google_cloud_vertexai extends Integration {}
+    */
+   interface google_cloud_vertexai extends Integration {}
 
-    /** @hidden */
-    interface ExecutionArgs {
-      schema: any,
-      document: any,
-      rootValue?: any,
-      contextValue?: any,
-      variableValues?: any,
-      operationName?: string,
-      fieldResolver?: any,
-      typeResolver?: any,
+   /**
+    * This plugin automatically instruments the
+    * [@google-genai](https://github.com/googleapis/js-genai) module.
+    */
+   interface google_genai extends Integration {}
+
+   /** @hidden */
+   interface ExecutionArgs {
+     schema: any,
+     document: any,
+     rootValue?: any,
+     contextValue?: any,
+     variableValues?: any,
+     operationName?: string,
+     fieldResolver?: any,
+     typeResolver?: any,
     }
 
     /**


packages/datadog-instrumentations/src/genai.js
```diff
@@ -0,0 +1,125 @@
...
+
+      const inputs = args[0]
+      const normalizedName = normalizeMethodName(method)
+
+      const ctx = {
+        methodName: normalizedName,
+        inputs,
+        args,
+        model: inputs?.model || 'unknown'
+      }
```
sabrenner: for simplicity, i think we can move these to the subscriber side and pass in 
```javascript
const ctx = { args }
```
as the context, since it looks like everything can be grabbed from `args`

packages/datadog-instrumentations/src/google-genai.js
```diff
@@ -0,0 +1,125 @@
...
+        args,
+        model: inputs?.model || 'unknown'
+      }
+
+      return genaiTracingChannel.start.runStores(ctx, () => {
+        let result
+        try {
+          result = original.apply(this, arguments)
+        } catch (error) {
+          finish(ctx, null, error)
```
sabrenner, line 27:
```suggestion
          finish(ctx, null, error)
          throw error
```

i think this needs to happen too but lmk if i'm wrong!

packages/datadog-instrumentations/src/genai.js
```diff
@@ -0,0 +1,125 @@
...
+      return genaiTracingChannel.start.runStores(ctx, () => {
+        let result
+        try {
+          result = original.apply(this, arguments)
+        } catch (error) {
+          finish(ctx, null, error)
+        }
+
+        return result.then(response => {
+          if (method === 'generateContentStream') {
```
sabrenner: i'm fine with this if it's the only way to identify it but i wonder if we can just check something like
```javascript
if (response[Symbol.asyncIterator) { ... }
```

as i think `generateContent` and `embedContent` can never return an async iterator, so this should be a good way of identifying the streamed response without a hardcoded check for the method name. but lmk if this does not work!

packages/datadog-instrumentations/src/genai.js
```diff
@@ -0,0 +1,125 @@
...
+            shimmer.wrap(response, Symbol.asyncIterator, iterator => wrapStreamIterator(iterator, ctx))
+          } else {
+            finish(ctx, response, null)
+            genaiTracingChannel.end.publish(ctx)
```
sabrenner: let's move this up to before the `return result.then(...)` as i think that represents the actual synchronous end, but lmk if that breaks anything

packages/datadog-instrumentations/src/genai.js
```diff
@@ -0,0 +1,125 @@
...
+              shimmer.wrap(this.models, 'generateContentStream', wrapGenerateContent('generateContentStream'))
+            }
+            if (this.models.embedContent) {
+              shimmer.wrap(this.models, 'embedContent', wrapGenerateContent('embedContent'))
+            }
+          }
+        }
+      }
+    })
+  }
```
sabrenner:
```suggestion
  if (!exports.GoogleGenAI) return exports
  
  shimmer.wrap(exports, 'GoogleGenAI', GoogleGenAI => {
      return class extends GoogleGenAI {
        constructor (...args) {
          super(...args)

          // Wrap the models property after it's created
          if (this.models) {
            if (this.models.generateContent) {
              shimmer.wrap(this.models, 'generateContent', wrapGenerateContent('generateContent'))
            }
            if (this.models.generateContentStream) {
              shimmer.wrap(this.models, 'generateContentStream', wrapGenerateContent('generateContentStream'))
            }
            if (this.models.embedContent) {
              shimmer.wrap(this.models, 'embedContent', wrapGenerateContent('embedContent'))
            }
          }
        }
      }
    })
```

just a personal nit for some negative spacing to reduce indents 😅

sabrenner: also, maybe we should leave a small comment as to why we had to do instance patching here (for future maintainers)

packages/datadog-plugin-google-genai/src/tracing.js
```diff
@@ -0,0 +1,49 @@
...
+  asyncEnd (ctx) {
+    if (ctx.result) {
+      ctx.currentStore.span.setTag('google_genai.response.model', ctx.result.modelVersion || ctx.inputs?.model)
+    }
+    if (ctx.currentStore.span) {
+      ctx.currentStore.span.finish()
+    }
+  }
+
+  end (ctx) {
```
sabrenner: i dont think we need `end` for the tracing plugin, just `asyncEnd`

packages/datadog-plugin-google-genai/src/tracing.js
```diff
@@ -0,0 +1,49 @@
...
+    return ctx.currentStore
+  }
+
+  asyncEnd (ctx) {
+    if (ctx.result) {
+      ctx.currentStore.span.setTag('google_genai.response.model', ctx.result.modelVersion || ctx.inputs?.model)
+    }
+    if (ctx.currentStore.span) {
+      ctx.currentStore.span.finish()
+    }
```
sabrenner, line 36:
```suggestion
    const { span } = ctx.currentStore
    if (!span) return

    if (ctx.result) {
      span.setTag('google_genai.response.model', ctx.result.modelVersion || ctx.inputs?.model)
    }

    span.finish()
```

nit

packages/dd-trace/src/llmobs/plugins/genai.js
```diff
@@ -0,0 +1,518 @@
...
+      const metrics = this._extractMetrics(response)
+      this._tagger.tagMetrics(span, metrics)
+    }
+  }
+
+  // ============================================================================
+  // Streaming Utilities
+  // ============================================================================
+
+  _aggregateStreamingChunks (chunks) {
```
sabrenner:
```suggestion
  #aggregateStreamingChunks (chunks) {
```

i think we're trying to do this notation more than the legacy `_` notation throughout the codebase (same for other private methods too)

packages/dd-trace/src/llmobs/plugins/genai.js
```diff
@@ -0,0 +1,518 @@
...
+    // Subscribe to streaming chunk events
+    this.addSub('apm:google:genai:request:chunk', ({ ctx, chunk, done }) => {
+      ctx.isStreaming = true
+      ctx.chunks = ctx.chunks || []
+
+      if (chunk) ctx.chunks.push(chunk)
+      if (!done) return
+
+      // Aggregate streaming chunks into a single response
+      ctx.result = this._aggregateStreamingChunks(ctx.chunks)
```
sabrenner: nice!

packages/dd-trace/src/llmobs/plugins/genai.js
```diff
@@ -0,0 +1,518 @@
...
+  // Public API Methods
+  // ============================================================================
+
+  getLLMObsSpanRegisterOptions (ctx) {
+    const methodName = ctx.methodName
+    if (!methodName) return
+
+    const inputs = ctx.inputs
+    const operation = getOperation(methodName)
+    const kind = operation === 'embedding' ? 'embedding' : 'llm'
```
sabrenner:
```suggestion
    const kind = operation
```

looks like this took some inspiration from the openai plugin which is it's own whole mess 😭 but i think here we can just do this

packages/dd-trace/src/llmobs/plugins/genai.js
```diff
@@ -0,0 +1,518 @@
...
+    return metadata
+  }
+}
+
+// ============================================================================
+// Module-level Utilities
+// ============================================================================
+
+function getOperation (methodName) {
+  if (!methodName) return 'unknown'
```
sabrenner: is there a case where `methodName` is empty or undefined? i think we can just 
```javascript
return methodName.includes('embed') ? 'embedding' : 'llm'
```
to also use it directly for span kind (instead of `chat`)

packages/dd-trace/src/llmobs/plugins/genai.js
```diff
@@ -0,0 +1,518 @@
...
+    if (error) {
+      this._tagger.tagLLMIO(span, inputMessages, [{ content: '' }])
+      return
+    }
+
+    const outputMessages = this._formatOutputMessages(response, isStreaming)
+    this._tagger.tagLLMIO(span, inputMessages, outputMessages)
+
+    const metadata = this._extractMetadata(config)
+    this._tagger.tagMetadata(span, metadata)
```
sabrenner: let's move this above the `if (error) { ... }` early return so that we can capture metadata even if there is an error

packages/dd-trace/src/llmobs/plugins/genai.js
```diff
@@ -0,0 +1,518 @@
...
+    const embeddingInput = this._formatEmbeddingInput(inputs.contents)
+
+    if (error) {
+      this._tagger.tagEmbeddingIO(span, embeddingInput)
+      return
+    }
+
+    const embeddingOutput = this._formatEmbeddingOutput(response)
+    this._tagger.tagEmbeddingIO(span, embeddingInput, embeddingOutput)
+    this._tagger.tagMetadata(span, {})
```
sabrenner: we can get rid of this call i think, since it's setting an empty metadata object (`{}`)

packages/dd-trace/src/llmobs/plugins/genai/index.js
```diff
@@ -0,0 +1,113 @@
...
+  extractMetadata,
+  aggregateStreamingChunks,
+  formatInputMessages,
+  formatEmbeddingInput,
+  formatOutputMessages,
+  formatEmbeddingOutput
+} = require('./util')
+
+class GenAiLLMObsPlugin extends LLMObsPlugin {
+  static id = 'genai'
```
sabrenner: i think this should be the same id as the other plugin
```suggestion
  static id = 'google-genai'
```
but we're good to keep the `integration` one as just `genai` below since it's used purely for telemetry.

packages/dd-trace/src/llmobs/plugins/genai/index.js
```diff
@@ -0,0 +1,113 @@
...
+      if (!done) return
+
+      // Aggregate streaming chunks into a single response
+      ctx.result = aggregateStreamingChunks(ctx.chunks)
+    })
+  }
+
+  // ============================================================================
+  // Public API Methods
+  // ============================================================================
```
sabrenner: i do think we're good to get rid of the comments like this

packages/dd-trace/src/llmobs/plugins/genai/index.js
```diff
@@ -0,0 +1,113 @@
...
+    const inputs = args[0]
+    const operation = getOperation(methodName)
+    const kind = operation
+
+    return {
+      modelProvider: 'google',
+      modelName: inputs.model,
+      kind,
+      name: 'google_genai.request'
+    }
```
sabrenner, line 48:
```suggestion
    const operation = getOperation(methodName)

    return {
      modelProvider: 'google',
      modelName: inputs.model,
      kind: operation,
      name: 'google_genai.request'
    }
```

packages/dd-trace/src/supported-configurations.json
sabrenner, line 1: i think technically only one of these is needed now (the second one maybe?), but i don't think it hurts to have both here should we allow both in the future

packages/dd-trace/src/llmobs/plugins/genai/util.js
```diff
@@ -0,0 +1,499 @@
...
+
+/**
+ * Extract metadata from config
+ * @param {object} config
+ * @returns {object}
+ */
+function extractMetadata (config) {
+  if (!config) return {}
+
+  const fieldMap = {
```
sabrenner, line 133: we can make this a constant at the top of the file

packages/dd-trace/src/llmobs/plugins/genai/util.js
```diff
@@ -0,0 +1,499 @@
...
+  }))
+
+  const textParts = extractTextParts(parts)
+  const content = textParts.length > 0 ? textParts.join('\n') : undefined
+
+  return {
+    role,
+    ...(content && { content }),
+    toolCalls
+  }
```
sabrenner:
```suggestion
  const message = { role, toolCalls }
  if (content) message.content = content

  return  message
```

just to be more explicit about it, took me a sec to figure out what was going on with `...(content && { content })` 😅

packages/dd-trace/src/llmobs/plugins/genai/util.js
```diff
@@ -0,0 +1,499 @@
...
+  aggregateStreamingChunks,
+  formatContentObject,
+  formatInputMessages,
+  formatEmbeddingInput,
+  formatNonStreamingCandidate,
+  formatStreamingOutput,
+  formatNonStreamingOutput,
+  formatOutputMessages,
+  formatEmbeddingOutput
+}
```
sabrenner, line 486: from the `index.js` file, it looks like we only ever use

```
getOperation
extractMetrics
extractMetadata
aggregateStreamingChunks
formatInputMessages
formatEmbeddingInput
formatOutputMessages
formatEmbeddingOutput
```

so we can probably trim down what we export from this file to just that.

otherwise the contents of this file lgtm!

sabrenner (commented):
all looks good to me just some last couple cleanup comments/making some things constants at the top of files. otherwise good to approve after!

sabrenner (approved):
🚢

PR #6935, opened 2025-11-17
fix: failing appsec tests after iitm bump to v2.0.0

pabloerhard:
### What does this PR do?
This PR aims to fix appsec tests failing after bumping iitm to version 2.0.0

### Motivation
iitm version 2.0.0 includes changes that include files being rewritten from CommonJS to ESM, which caused the conditions in the isEsmConfigured function to fail. This happened because the files are no longer keys in the require.cache. This PR adds the flag --import dd-trace/initialize.mjs as an early return statement to bypass the issues caused by the iitm change.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
```diff
@@ -184,8 +184,11 @@ function enableRewriter (telemetryVerbosity) {
 }
 
 function isEsmConfigured () {
-  const hasLoaderArg = isFlagPresent('--loader') || isFlagPresent('--experimental-loader')
-  if (hasLoaderArg) return true
+  const hasLoaderOrImportArg = (isFlagPresent('--loader') ||
+    isFlagPresent('--experimental-loader') ||
+    isFlagPresent('--import dd-trace/initialize.mjs'))
```
BridgeAR: Should it not just check for the following?

```suggestion
    isFlagPresent('--import'))
```

Our instrumentation might not even be used like that and other imports might be there as well.

packages/datadog-esbuild/src/utils.js
```diff
@@ -7,9 +7,12 @@ const path = require('node:path')
 const { NODE_MAJOR, NODE_MINOR } = require('../../../version.js')
 
 const getExportsImporting = (url) => import(url).then(Object.keys)
-const getExports = NODE_MAJOR >= 20 || (NODE_MAJOR === 18 && NODE_MINOR >= 19)
-  ? require('import-in-the-middle/lib/get-exports.js')
-  : getExportsImporting
+
+const isModernNode = NODE_MAJOR >= 20 || (NODE_MAJOR === 18 && NODE_MINOR >= 19)
```
BridgeAR: Could you check if this is still needed when loading iitm with ESM? I am not sure what was introduced at that version (the special handling could use a comment).

BridgeAR (commented):
LGTM with the import being more general. The other part is optional.

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
uurien: or else, `isFlagPresent('dd-trace/initialize.mjs')` because the method `isFlagPresent` checks in `NODE_OPTIONS` (current approach works there) but also in `process.execArgv`, and here, if an app is starting with `node --import dd-trace/initialize.mjs` the value in `process.execArgv` is going to be `['--import', 'dd-trace/initialize.mjs']`

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
uurien: and I think we should remove this block, because `hooks.js` doesn't exist in iitm 2.0

uurien (dismissed):

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
pabloerhard: True, i had removed them but added them back in a git reset ... forgot to remove them again.

packages/datadog-esbuild/src/utils.js
pabloerhard: Yep it is still needed and something identical is done in the iitm library. 
https://github.com/nodejs/import-in-the-middle/blob/main/create-hook.mjs#L23

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
pabloerhard: Yep, I think in this case it revolves around whether we want the check to be unique to initialize.mjs or more general to --import.

pabloerhard: If we were to just check for 'dd-trace/initialize.mjs' wouldn't it not cover something like the following: 
node --import dd-trace/register.js --require dd-trace/init app.js ?

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
```diff
@@ -184,14 +184,9 @@ function enableRewriter (telemetryVerbosity) {
...
-  if (hasLoaderArg) return true
-
-  // Fast path for common case when enabled
-  if (require.cache[`${process.cwd()}/node_modules/import-in-the-middle/hook.js`]) {
-    return true
-  }
-  return Object.keys(require.cache).some(file => file.endsWith('import-in-the-middle/hook.js'))
+  return (isFlagPresent('--loader') ||
+    isFlagPresent('--experimental-loader') ||
+    isFlagPresent('--import'))
```
bengl: The presence of `--import` alone is insufficient to to say that we're correctly configured for ESM.

bengl (dismissed):

packages/dd-trace/src/appsec/iast/taint-tracking/rewriter.js
pabloerhard: Could the same be said for the presence of --loader? (general doubt)

pabloerhard: Should checking for the presence of dd-trace/initialize.mjs or dd-trace/register.js suffice it?

pabloerhard (commented):
I added dd-trace/initialize.mjs and dd-trace/register.js as flags to consider when checking whether ESM is configured, since these are the options provided in Datadog’s documentation. ESM does not directly expose the module cache, so a direct alternative to require.cache does not currently exist. I added this general comment so we can further discuss any other options or concerns, as the two open comments relate to this change.
@bengl @BridgeAR @uurien

packages/datadog-esbuild/src/utils.js
```diff
@@ -7,9 +7,11 @@ const path = require('node:path')
 const { NODE_MAJOR, NODE_MINOR } = require('../../../version.js')
 
 const getExportsImporting = (url) => import(url).then(Object.keys)
-const getExports = NODE_MAJOR >= 20 || (NODE_MAJOR === 18 && NODE_MINOR >= 19)
-  ? require('import-in-the-middle/lib/get-exports.js')
-  : getExportsImporting
+
+const getExports = async () => {
```
uurien: why are you moving this logic to `processModule` instead of doing all the logic in `getExports` method?

package.json
```diff
@@ -140,7 +140,7 @@
     "dc-polyfill": "^0.1.10",
     "escape-string-regexp": "^5.0.0",
     "ignore": "^7.0.5",
-    "import-in-the-middle": "^1.14.2",
+    "import-in-the-middle": "2.0.0",
```
uurien: why is this pinned? afaik we only pin our packages (`@datadog/xxx`)
```suggestion
    "import-in-the-middle": "^2.0.0",
```

BridgeAR (commented):
LGTM % @uurien 's comment. It would remove the overhead of checking for that more than once (which will not show up anywhere, but it is still nice).

packages/datadog-esbuild/src/utils.js
```diff
@@ -7,9 +7,22 @@ const path = require('node:path')
...
+  return getExportsModulePromise
+}
+
+const getExports = async (srcUrl, context, getSource) => {
+  if (NODE_MAJOR >= 20 || (NODE_MAJOR === 18 && NODE_MINOR >= 19)) {
+    const mod = await loadGetExportsModule()
+    return mod.getExports(srcUrl, context, getSource)
+  }
+  return getExportsImporting(srcUrl, context, getSource)
+}
```
BridgeAR: This does a runtime check each time the method is called. Before, it would create the proper method depending on the version, never doing the runtime check.

Could you please change that back to the former behavior? :) 

```suggestion
const getExports = NODE_MAJOR >= 20 || (NODE_MAJOR === 18 && NODE_MINOR >= 19
  ? async (srcUrl, context, getSource) => {
    const mod = await loadGetExportsModule()
    return mod.getExports(srcUrl, context, getSource)
  }
  : getExportsImporting
```

packages/datadog-esbuild/src/utils.js
```diff
@@ -80,6 +92,7 @@ async function processModule ({ path, internal, context, excludeDefault }) {
   } else {
     srcUrl = pathToFileURL(path)
     exportNames = await getExports(srcUrl, context, getSource)
+    process._rawDebug('This is getExports ', exportNames)
```
BridgeAR:
```suggestion
```

BridgeAR (commented):
LGTM with the debug log removed

PR #6940, opened 2025-11-18
ci: set benchmarking jobs to be interruptible outside master

igoragoli:
### What does this PR do?

- Set `interruptible: true` on benchmarking jobs outside `master` as they become redundant on new runs.
    - Do the same for serverless benchmarks, since recent benchmarks still become redundant on new runs inside development branches.
- Change macrobenchmark runner type to `runner:apm-k8s-m7i-metal` since they're also run on every branch.

### Motivation

[#incident-45572](https://dd.enterprise.slack.com/archives/C09S0H9M3B9) and [APMSP-2369](https://datadoghq.atlassian.net/browse/APMSP-2369).

Why set jobs to interruptible? Because by doing so, we prevent redundant runs from overloading our APM runner capacity.

Why only outside of `master`? Because benchmarks on `master` are never redundant: we need every run on master to collect results and maintain a continuous performance history.

### Additional Notes

[APMSP-2369]: https://datadoghq.atlassian.net/browse/APMSP-2369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitlab/macrobenchmarks.yml
+++ b/.gitlab/macrobenchmarks.yml
@@ -11,7 +11,7 @@ include:
     - if: $CI_COMMIT_BRANCH == 'master'
       interruptible: false
     - interruptible: true
-  tags: ["runner:apm-k8s-m7i-metal"]
+  tags: ["runner:apm-k8s-same-cpu"]
   needs: []
   timeout: 15m # TODO: Fix worker queueing and reduce this.
   image: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/benchmarking-platform:js-hapi


duncanista (approved):
Looks good from Serverless!

.gitlab/macrobenchmarks.yml
```diff
@@ -6,12 +6,13 @@ include:
...
   stage: macrobenchmarks
+  when: always
   rules:
-    - when: always
-    - when: manual
-  tags: ["runner:apm-k8s-same-cpu"]
+    - if: $CI_COMMIT_BRANCH == 'master'
+      interruptible: false
+    - interruptible: true
+  tags: ["runner:apm-k8s-m7i-metal"]
```
ddyurchenko: Please revert, we need to keep runner:apm-k8s-same-cpu for macrobenchmarks.

BridgeAR:
```suggestion
  tags: ["runner:apm-k8s-same-cpu"]
```

PR #6941, opened 2025-11-18
feat(openai): instrument openai reusable prompts

PROFeNoM:
### What does this PR do?

Adds prompt tracking for OpenAI reusable prompts with support for mixed input types (text, images, files).

**Key Changes:**
- Extracts prompt metadata (id, version, variables) from request arguments
- `normalizePromptVariables()`: Extracts values from OpenAI SDK objects
- `extractChatTemplateFromInstructions()`: Generates templates using `{{variable_name}}` when values are available, falls back to `[image]`/`[file]` when stripped

👉 **Design details:** See [Prompt Tracking with Mixed Input Types](https://docs.google.com/document/d/1qLgC45fU0ZSKhxucHFnLXydBM5pa_i74--9RVBzxgcA/edit?usp=sharing)

### Motivation

Instruments OpenAI's Reusable Prompts feature for LLM Observability. Implementation is consistent with dd-trace-py ([#15359](https://github.com/DataDog/dd-trace-py/pull/15359)).

### Testing

- Template extraction and variable normalization for all input types
- Edge cases: overlapping values, partial matches, special characters, empty values
- Mixed input types with `include` parameter for URL preservation

**Test Prompts** (on Datadog Staging OpenAI dashboard):
- [overlapping_values](https://platform.openai.com/chat/edit?prompt=pmpt_6911a8b8f7648197b39bd62127a696910d4a05830d5ba1e6&version=1)
- [partial_word_match](https://platform.openai.com/chat/edit?prompt=pmpt_6911a954c8988190a82b11560faa47cd0d6629899573dd8f&version=2)
- [special_characters](https://platform.openai.com/chat/edit?prompt=pmpt_6911a99a3eec81959d5f2e408a2654380b2b15731a51f191&version=2)
- [mixed_input_types](https://platform.openai.com/chat/edit?prompt=pmpt_69201db75c4c81959c01ea6987ab023c070192cd2843dec0&version=2)

### Plugin Checklist

- [x] Unit tests
- [x] Integration tests
- [ ] Benchmarks
- [ ] TypeScript definitions
- [ ] TypeScript tests
- [ ] API documentation
- [ ] CI jobs/workflows

### DIFF
--- a/packages/dd-trace/src/llmobs/plugins/openai/index.js
+++ b/packages/dd-trace/src/llmobs/plugins/openai/index.js
@@ -295,7 +295,7 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
           inputMessages.push({ role: item.role, content: item.content })
         }
       }
-    } else if (input) {
+    } else {
       // Simple string input
       inputMessages.push({ role: 'user', content: input })
     }


packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -380,6 +406,33 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+              ? value.text
+              : value
+          }
+
+          this._tagger._setTag(span, '_ml_obs.meta.input.prompt', {
+            id,
+            version,
+            variables: normalizedVariables,
+            chat_template: chatTemplate
+          })
```
sabrenner: i think for right now this is ok - ideally we'd have a `this._tagger.tagPrompt` method that would do validation like [we do in python.](https://github.com/DataDog/dd-trace-py/blob/2cc3c8b749084b51895095a5cf78ba35dad1fe02/ddtrace/llmobs/_utils.py#L51)

i actually have another branch that does this but has no pr. unless you wanna implement it here i think it's better left out of scope for now and we can clean this block up to use something like `this._tagger.tagPrompt` in a future pr that does all of the validation, wdyt?

packages/dd-trace/src/llmobs/util.js
```diff
@@ -174,9 +174,63 @@ function spanHasError (span) {
   return !!(tags.error || tags['error.type'])
 }
 
+// Extracts chat templates from OpenAI response instructions by replacing variable values with placeholders
+function extractChatTemplateFromInstructions (instructions, variables) {
```
sabrenner: what i think we can do with this util instead is give it directly to openai - we can make `plugins/openai.js` into a directory `plugins/openai`, with a `index.js` file that has the current contents of `openai.js`, and then an additional `utils.js` file which has this function, just to give some separation of this generic util file and one specifically for openai utils. wdyt?

packages/dd-trace/src/llmobs/util.js
```diff
@@ -174,9 +174,63 @@ function spanHasError (span) {
...
+  if (!instructions || !Array.isArray(instructions)) return []
```
sabrenner:
```suggestion
  if (!Array.isArray(instructions)) return []
```

`Array.isArray` also checks for existence 😄

packages/dd-trace/src/llmobs/util.js
```diff
@@ -174,9 +174,63 @@ function spanHasError (span) {
...
+  // Sort values by length (longest first) to handle overlapping values correctly
+  const sortedValues = Object.keys(valueToPlaceholder).sort((a, b) => b.length - a.length)
+
+  // Process each instruction
+  for (const instruction of instructions) {
+    const role = instruction.role
+    if (!role) continue
+
+    const contentItems = instruction.content
+    if (!contentItems || !Array.isArray(contentItems)) continue
```
sabrenner:
```suggestion
    if (!Array.isArray(contentItems)) continue
```

ditto

packages/dd-trace/src/llmobs/util.js
```diff
@@ -174,9 +174,63 @@ function spanHasError (span) {
...
+
+    // Collect text parts from content items
+    const textParts = []
+    for (const contentItem of contentItems) {
+      const text = contentItem.text
+      if (text) {
+        textParts.push(String(text))
```
sabrenner: just wanna check what we expect `text` could be. if it's always non-object, and just primitives like numbers, strings, or booleans, for example, this should be fine

```
> String('foo')
'foo'
> String(5)
'5'
> String(true)
'true'
```

but if it's an object it doesn't look as nice

```
> String({ text: 'foo' })
'[object Object]'
```

we might want to use `JSON.stringify` if we expect it to be an object, and probably try-catch it (as JSON cannot serialize circular objects, or objects with `bigint`s). but lmk, if it's just expected to be primitives (here and above), then `String` is ok!

packages/dd-trace/test/llmobs/util.spec.js
```diff
@@ -177,4 +178,102 @@ describe('util', () => {
       expect(spanHasError(span)).to.equal(true)
     })
   })
+
+  describe('extractChatTemplateFromInstructions', () => {
```
sabrenner: if we move this utility to an openai folder, and we instead test the functionality/edge cases as part of `openaiv4.spec.js`, i think we're good to get rid of these tests 😄

packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
sabrenner, line 1: do the `prompt.id`s that exist in these tests (ie `pmpt_6911a8b8f7648197b39bd62127a696910d4a05830d5ba1e6`) already exist? i was wondering where we were getting the ids from 😅 totally good if so, maybe just a small comment explaining so

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -231,10 +233,34 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+            // Extract text from content items (e.g., { type: 'input_text', text: '...' })
+            const textParts = item.content
+              .filter(c => c.type === 'input_text' && c.text)
+              .map(c => c.text)
+            content = textParts.join('')
+          } else if (typeof item.content === 'string') {
+            content = item.content
+          }
+
+          if (content || role === 'developer') {
```
sabrenner: if for some reason `content` is not set and is still an empty string, and the role is not `developer`, we will not push the message at all, and then skip all other branches. is this intentional? or do we want to do a `content != null` check. if this is intentional then all good, just wanted to check!

packages/dd-trace/src/llmobs/plugins/openai.js
```diff
@@ -380,6 +406,33 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
 
     this._tagger.tagLLMIO(span, inputMessages, outputMessages)
 
+    // Handle prompt tracking for reusable prompts
+    if (response && response.prompt && typeof response.prompt === 'object') {
+      const { id, version, variables } = response.prompt
+      if (id && version && variables) {
+        // Extract chat template from response instructions
+        const instructions = response.instructions
+        if (instructions && Array.isArray(instructions)) {
```
```suggestion
        if (Array.isArray(instructions)) {
```

packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
PROFeNoM: You're right., I forgot to mention it.

The reusable prompts can be found on `Datadog Staging` on the OpenAI dashboard:
- [overlapping_values](https://platform.openai.com/chat/edit?prompt=pmpt_6911a8b8f7648197b39bd62127a696910d4a05830d5ba1e6&version=1)
- [partial_word_match](https://platform.openai.com/chat/edit?prompt=pmpt_6911a954c8988190a82b11560faa47cd0d6629899573dd8f&version=2)
- [special_characters](https://platform.openai.com/chat/edit?prompt=pmpt_6911a99a3eec81959d5f2e408a2654380b2b15731a51f191&version=2)

I've added a one-line comment but perhaps we can make it even more informative.

packages/dd-trace/src/llmobs/util.js
PROFeNoM: Ok, changed the structure :+1:

packages/dd-trace/src/llmobs/plugins/openai.js
PROFeNoM: I was wondering this as well when writing it (since I was mostly translating from python to js). I agree this should be deferred to a follow-up PR :+1:

packages/dd-trace/src/llmobs/util.js
PROFeNoM: `text` is always a string primitive according to [OpenAI's SDK](https://github.com/openai/openai-node/blob/c9a4d6886cf36d6a7abb5df8d7da1b007d2ae89d/src/resources/responses/responses.ts#L3505-L3515):
```ts
interface ResponseInputText {
  text: string;  // Always a string
  type: 'input_text';
}
```

In the updated code (`.../plugins/openai/utils.js`), I changed it so that (line 46) we directly push `contentItem.text` without `String()` conversion since it's guaranteed to be a string. I'm only using it when creating the map key (line 29) `String(valueStr)` since `valueStr` can be either a plain string or extracted from a `ResponseInputText` object, and we need to ensure it's a string for use as an object key. Perhaps we can even go without it tbh. I'm just trying to be defense here. But in any case, I believe it is *not* supposed to be an object :pray:

packages/dd-trace/src/llmobs/plugins/openai.js
PROFeNoM: big mb there. No role ought to be special-cased. Empty content should be skipped regardless of role (hence aligning with the Python implementation).

packages/dd-trace/src/llmobs/util.js
PROFeNoM: I've also added a bit more information (in `packages/dd-trace/src/llmobs/plugins/openai/utils.js`) about the types of the variables because I am already getting lost and don't want to re-spend time on searching for the types a third time lol 😭

packages/dd-trace/src/llmobs/plugins/openai/index.js
```diff
@@ -380,6 +406,33 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+        const instructions = response.instructions
+        if (Array.isArray(instructions)) {
+          const chatTemplate = extractChatTemplateFromInstructions(instructions, variables)
+
+          // Normalize variables - OpenAI returns objects with 'text' field, we need simple key-value pairs
+          const normalizedVariables = {}
+          for (const [key, value] of Object.entries(variables)) {
+            normalizedVariables[key] = (value && typeof value === 'object' && value.text !== undefined)
+              ? value.text
+              : value
```
PROFeNoM: I need to rework that part i forgot about something (i.e., image or file inputs)

packages/dd-trace/src/llmobs/plugins/openai/index.js
```diff
@@ -231,10 +233,33 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
       inputMessages.push({ role: 'system', content: inputs.instructions })
     }
 
+    // For reusable prompts, use response.instructions if no explicit input is provided
+    if (!input && inputs.prompt && response && response.instructions) {
```
sabrenner:
```suggestion
    if (!input && inputs.prompt && response?.instructions) {
```

packages/dd-trace/src/llmobs/plugins/openai/index.js
```diff
@@ -270,7 +295,7 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
           inputMessages.push({ role: item.role, content: item.content })
         }
       }
-    } else {
+    } else if (input) {
```
sabrenner: was this breaking anything with shared tests? i think if now for whatever reason if the user sets an input message as an empty string we won't attach it as a message, which maybe leaves an observability hole. but lmk!

packages/dd-trace/src/llmobs/plugins/openai/index.js
```diff
@@ -380,6 +405,21 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
 
     this._tagger.tagLLMIO(span, inputMessages, outputMessages)
 
+    // Handle prompt tracking for reusable prompts
+    if (inputs.prompt && response && response.prompt) {
```
sabrenner:
```suggestion
    if (inputs.prompt && response?.prompt) {
```

packages/dd-trace/src/llmobs/plugins/openai/index.js
```diff
@@ -380,6 +405,21 @@ class OpenAiLLMObsPlugin extends LLMObsPlugin {
...
+      const { id, version } = response.prompt // ResponsePrompt
+      if (id && version) {
+        const normalizedVariables = normalizePromptVariables(inputs.prompt.variables)
+        const chatTemplate = extractChatTemplateFromInstructions(response.instructions, normalizedVariables)
+        this._tagger._setTag(span, '_ml_obs.meta.input.prompt', {
```
sabrenner, line 415: i don't think i've made a ticket yet, but feel free to make one and add a TODO here that we'll follow up on turning this into an actual tagger API for future prompts work!

packages/dd-trace/src/llmobs/plugins/openai/utils.js
```diff
@@ -0,0 +1,118 @@
...
+  const normalized = {}
+  for (const [key, value] of Object.entries(variables)) {
+    let normalizedValue = value
+    if (value && typeof value === 'object') {
+      if (value.text !== undefined) { // ResponseInputText
+        normalizedValue = value.text
+      } else if (value.type === 'input_image') { // ResponseInputImage
+        normalizedValue = value.image_url || value.file_id || '[image]'
+      } else if (value.type === 'input_file') { // ResponseInputFile
+        normalizedValue = value.file_url || value.file_id || value.filename || '[file]'
```
sabrenner: just checking if we're fine with empty strings for any of these values, otherwise we'll want to use `??` in place of `||` (nullish coalescing vs false-ish coalescing)

packages/dd-trace/src/llmobs/plugins/openai/utils.js
```diff
@@ -0,0 +1,118 @@
...
+ * @param {Object<string, string|Object>} variables - From ResponsePrompt.variables
+ * @returns {Object<string, string>} Normalized variables with simple string values
+ */
+function normalizePromptVariables (variables) {
+  if (!variables) return {}
+
+  const normalized = {}
+  for (const [key, value] of Object.entries(variables)) {
+    let normalizedValue = value
+    if (value && typeof value === 'object') {
```
sabrenner: should we be using the above `extractTextFromContentItem` here? they look very similar, maybe something like
```javascript
if (value && typeof value === 'object') {
  const textValueFromObject = extractTextFromContentItem(value)
  if (textValueFromObject != null) normalizedValue = textValueFromObject
}
```

packages/dd-trace/src/llmobs/plugins/openai/utils.js
```diff
@@ -0,0 +1,118 @@
...
+  if (!Array.isArray(instructions) || !variables) return []
+
+  const chatTemplate = []
+
+  // Build map of values to placeholders - exclude fallback markers so they remain as-is
+  const valueToPlaceholder = {}
+  for (const [varName, varValue] of Object.entries(variables)) {
+    const valueStr = varValue ? String(varValue) : ''
+    // Exclude fallback markers - they should remain as [image]/[file] in the template
+    if (valueStr && valueStr !== '[image]' && valueStr !== '[file]') {
```
sabrenner: maybe we can make these `'[image']` and `'[file]'` constants at the top to use throughout the file, wdyt?

packages/dd-trace/src/llmobs/plugins/openai/utils.js
```diff
@@ -0,0 +1,118 @@
...
+ */
+function extractChatTemplateFromInstructions (instructions, variables) {
+  if (!Array.isArray(instructions) || !variables) return []
+
+  const chatTemplate = []
+
+  // Build map of values to placeholders - exclude fallback markers so they remain as-is
+  const valueToPlaceholder = {}
+  for (const [varName, varValue] of Object.entries(variables)) {
+    const valueStr = varValue ? String(varValue) : ''
```
sabrenner: maybe we can do this check/normalization in the `normalizePromptVariables` method since it pertains more there (although fine to keep the `{{...}}` part here)

packages/dd-trace/src/llmobs/plugins/openai/utils.js
```diff
@@ -0,0 +1,118 @@
...
+      .map(extractTextFromContentItem)
+      .filter(Boolean)
+
+    if (textParts.length === 0) continue
+
+    // Combine text and replace variable values with placeholders (longest first)
+    let fullText = textParts.join('')
+    for (const valueStr of sortedValues) {
+      const placeholder = valueToPlaceholder[valueStr]
+      const escapedValue = valueStr.replaceAll(/[.*+?^${}()|[\]\\]/g, String.raw`\$&`)
```
sabrenner: the `/[.*+?^${}()|[\]\\]/g` part can be a constant at the top of the file i think

packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
```diff
@@ -755,6 +755,245 @@ describe('integrations', () => {
           tags: { ml_app: 'test', integration: 'openai' }
         })
       })
+
+      // Helper function to verify prompt tracking in span events
+      // Note: Prompt IDs (pmpt_*) are real reusable prompts created on "Datadog Staging" OpenAI's dashboard for testing
+      function assertPromptTracking (spanEvent, expectedPrompt, expectedInputMessages) {
```
sabrenner: we can hoist this function to the top of the file outside of the testing blocks i think

sabrenner: or maybe better, put it in the `util.js` file in this test directory

packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
sabrenner: thank you!

packages/dd-trace/test/llmobs/plugins/openai/openaiv4.spec.js
```diff
@@ -755,6 +755,245 @@ describe('integrations', () => {
...
+        // Verify prompt metadata
+        assert(spanEvent.meta.input.prompt, 'Prompt metadata should be present')
+        const prompt = spanEvent.meta.input.prompt
+        assert.strictEqual(prompt.id, expectedPrompt.id)
+        assert.strictEqual(prompt.version, expectedPrompt.version)
+        assert.deepStrictEqual(prompt.variables, expectedPrompt.variables)
+        assert.deepStrictEqual(prompt.chat_template, expectedPrompt.chat_template)
+      }
+
+      it('submits a response span with prompt tracking - overlapping values', async function () {
```
sabrenner: another thing we can consider for this is grouping all of these tests under a `describe('prompts', ...)` section and we can do a global skip. this way we can reduce the `this.skip()` calls and then to run all the prompts test a developer only needs to add `.only` to the `describe` block, with `describe.only(...)`

```javascript
describe('prompts', function () {
  beforeEach(function () {
    if (semifies(realVersion, '<4.87.0')) {
      this.skip()
    }
  })

  // all of the prompts tests
})

File: packages/dd-trace/src/llmobs/plugins/openai/index.js (hunk 11 above)
Comment (PROFeNoM) [reply packages/dd-trace/src/llmobs/plugins/openai/index.js]: It wasn't breaking anything, no. It's just that since I looked more deeply into the Python implementation _before_, I had seen that in [_openai_parse_input_response_messages](https://github.com/DataDog/dd-trace-py/blob/9ca9bc41d263394162f18b1203c865460f03c754/ddtrace/llmobs/_integrations/utils.py#L569-L570), empty inputs were ignored rather than creating a message with empty content. 

Tbh, I can change it back to as it was before. I don't really care.

File: packages/dd-trace/src/llmobs/plugins/openai/utils.js (hunk 14 above)
Comment (PROFeNoM) [reply packages/dd-trace/src/llmobs/plugins/openai/utils.js]: If `image_url: ''` is provided, we want to fall through to `file_id` or ultimately `[image]`... so `||` is right, no? Unless I misread your message 😅 Like, if the value on the left side is falsy, then I want to return the right side.

File: packages/dd-trace/src/llmobs/plugins/openai/utils.js (hunk 15 above)
Comment (PROFeNoM) [reply packages/dd-trace/src/llmobs/plugins/openai/utils.js]: Absolutely right + this allows for much idiomatic code in the end

File: packages/dd-trace/src/llmobs/plugins/openai/index.js (hunk 11 above)
Comment (sabrenner) [reply packages/dd-trace/src/llmobs/plugins/openai/index.js]: ~~yea just to be safe let's change it back!~~ i misread your message, let's keep it as you have it now for consistency 👍

File: packages/dd-trace/src/llmobs/plugins/openai/utils.js (hunk 14 above)
Comment (sabrenner) [reply packages/dd-trace/src/llmobs/plugins/openai/utils.js]: oh man i think i must've been asleep while looking at this part, you're correct haha. disregard!

File: packages/dd-trace/src/llmobs/plugins/openai/index.js (hunk 11 above)
Comment (PROFeNoM) [reply packages/dd-trace/src/llmobs/plugins/openai/index.js]: ok :+1:

```suggestion
    } else {
```

sabrenner (approved):

PR #6948, opened 2025-11-19
add env var to configure ffe init time

leoromanovsky:
## Motivation

Follow-up to the OpenFeature flagging provider feature. The provider was initially configurable only via the programmatic API. This PR adds environment variable support so users can enable and configure the flagging provider without code changes.

## Changes

- Add `DD_FLAGGING_PROVIDER_ENABLED` and `DD_EXPERIMENTAL_FLAGGING_PROVIDER_ENABLED` env vars to enable the flagging provider
- Add `DD_FLAGGING_PROVIDER_INITIALIZATION_TIMEOUT_MS` and `DD_EXPERIMENTAL_FLAGGING_PROVIDER_INITIALIZATION_TIMEOUT_MS` env vars to configure initialization timeout
- Update TypeScript definitions to document the env vars
- Add tests for env var configuration

## Decisions

- Both experimental (`DD_EXPERIMENTAL_*`) and non-experimental (`DD_*`) env var names are supported for consistency with other configuration options
- Experimental env vars take precedence when both are set
- Programmatic config values override env vars (standard behavior)


packages/dd-trace/src/config/supported-configurations.json
```diff
@@ -73,11 +73,13 @@
     "DD_ENV": ["A"],
     "DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED": ["A"],
     "DD_EXPERIMENTAL_FLAGGING_PROVIDER_ENABLED": ["A"],
+    "DD_EXPERIMENTAL_FLAGGING_PROVIDER_INITIALIZATION_TIMEOUT_MS": ["A"],
     "DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED": ["A"],
     "DD_EXPERIMENTAL_TEST_OPT_GIT_CACHE_ENABLED": ["A"],
     "DD_EXPERIMENTAL_TEST_OPT_GIT_CACHE_DIR": ["A"],
     "DD_EXTERNAL_ENV": ["A"],
     "DD_FLAGGING_PROVIDER_ENABLED": ["A"],
+    "DD_FLAGGING_PROVIDER_INITIALIZATION_TIMEOUT_MS": ["A"],
```
BridgeAR: I am unsure why two envs are added, one named experimental and one non experimental.

I think we should only use one and since it is experimental, I would remove the other one for now.


```suggestion
```

BridgeAR (changes requested):
LGTM in case we use a single env.
I do not know how stable things are, so I would also be fine to use the `non experimental` name. While it is a new case for me and I would have to look up what rules we define for us about that (if we have some).

PR #6949, opened 2025-11-19
fix(llmobs): make sure all sdk methods are on the no-op sdk as well

sabrenner:
### What does this PR do?
Ensures recently added methods (`annotationContext`, `registerProcessor`, and `deregisterProcessor`) are on the no-op sdk, and are properly tested.

### Motivation
The LLM Observability SDK should not crash with these functions when it is disabled.


sabrenner:
@BridgeAR yeah i actually had that locally but thought that it felt too gimmicky somehow 😅 but i can add that back in as well!

PR #6958, opened 2025-11-21
vendor bundled non-datadog dependencies

rochdev:
### What does this PR do?

Vendor bundled versions of non-Datadog dependencies.

### Motivation

As we start supporting more and more non-server environments, our package size becomes increasingly problematic. Other than native addons (which are not included in non-server environments anyway), most of it comes from our dependencies, and most of their size comes from files that they package but never actually use. But even the code they do use is often unoptimized as it's not minified and is spread across many files. All of this not only makes their package bigger but also slower to load. By including vendored versions of our dependencies, we can make sure all of them are optimized in the way we want. This also has several other benefits:

* Determinism: Right now what we test is not necessarily what the user installs. Even our tests themselves are inconsistent as unit tests use the lockfile, but not integration tests since they do a real install. This also means that once all dependencies are vendored, we could even skip the install completely in tests.
* Speed: Installing dd-trace and all of its dependencies can take several seconds because of the resolution stage and the hundred or so additional transitive dependencies that have to be downloaded. With vendors the resolution is instant and no additional download is needed.
* Isolation: Once all dependencies are included (right now dependencies under `@datadog` are not included) then the tracer becomes self-contained. Once packed, it can be unpacked and already fully functional without any additional install. This also has the benefit of avoiding side-effects when merging `node_modules`, for example with the Lambda extension.
* Consistency: This is already basically how we ship the SDK in Serverless and SSI environments, but today it's done with brittle hacks in the builds of each target instead of upstream in the library. Other languages also already vendor their dependencies (Java, .NET, etc).

### Additional Notes

A next step would be to make the native addons a separate package so that they don't need to be removed when repackaging in a non-server environment and would instead simply not be installed in the first place.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -92,37 +92,6 @@ updates:
           - "minor"
           - "patch"
 
-  # OTEL in our supported ranges
-  - package-ecosystem: "npm"
-    directories:
-      - "/"
-    schedule:
-      interval: "weekly"
-    open-pull-requests-limit: 1
-    cooldown:
-      default-days: 5
-    # Widen range across major versions, if possible. Ignore / manually update otherwise.
-    versioning-strategy: "widen"
-    labels:
-      - dependabot
-      - dependencies
-      - javascript
-      - semver-patch
-      - OTEL
-    ignore:
-      - dependency-name: "@opentelemetry/core"
-        # 2.0.0 onwards only supports Node.js 18.19.0 and above
-        update-types: ["version-update:semver-major"]
-      - dependency-name: "@opentelemetry/resources"
-        # 2.0.0 onwards only supports Node.js 18.19.0 and above
-        update-types: ["version-update:semver-major"]
-    allow:
-      - dependency-name: "@opentelemetry/*"
-    groups:
-      otel-dependencies:
-        patterns:
-          - "*"
-
   # Vendored dependencies
   - package-ecosystem: "npm"
     directories:


simon-id:
Uuuh, i am ok with vendoring, but should we really commit obfuscated code like that ? I don't feel comfortable with that for security reasons, anyone can just hide shit in there, and pretend it's just an update of the lib. I'd prefere if this was done by a script in the CI or something, before publishing the package.

cbeauchesne:
> Did we ever get an answer from GitHub as to what is going on?

Unfortunately no. At some point we'll need more data to prove them the impact this issue have. But TBH, it happened a little bit less those past weeks.

eslint.config.mjs
```diff
@@ -355,6 +355,8 @@ export default [
         dynamicImports: 'always-multiline'
       }],
       'import/no-extraneous-dependencies': 'error',
+      'n/no-extraneous-require': 'off',
+      'n/no-unpublished-require': 'off',
```
BridgeAR: I think we should still use these, if we can. We could look at the options such as https://github.com/raycohen/eslint-plugin-n/blob/master/docs/rules/no-extraneous-require.md

integration-tests/playwright/playwright.spec.js
```diff
@@ -297,6 +297,8 @@ versions.forEach((version) => {
           stdio: 'inherit'
         }
       )
+      childProcess.stdout.pipe(process.stdout)
+      childProcess.stderr.pipe(process.stderr)
```
BridgeAR:
```suggestion
```

package.json
```diff
@@ -115,59 +115,37 @@
...
     "@datadog/native-appsec": "10.3.0",
     "@datadog/native-iast-taint-tracking": "4.1.0",
     "@datadog/native-metrics": "3.1.1",
-    "@datadog/openfeature-node-server": "^0.2.0",
     "@datadog/pprof": "5.12.0",
-    "@datadog/sketches-js": "2.1.1",
     "@datadog/wasm-js-rewriter": "5.0.1",
-    "@isaacs/ttlcache": "^2.0.1",
+    "@openfeature/core": "^1.8.1",
+    "@openfeature/server-sdk": "~1.20.0",
```
BridgeAR: Let us keep these out and use @datadog/openfeature-node-server in our regular dependencies for now until we find a solution for these.

packages/datadog-esbuild/index.js
```diff
@@ -245,29 +244,33 @@ ${build.initialOptions.banner.js}`
...
+            full: fullPathToModule,
+            raw: args.path,
+            pkgOfInterest: true,
+            kind: args.kind,
+            internal,
+            isESM
+          }
         }
+      } catch {
+        // Skip vendored dependencies which never have a `package.json`.
```
BridgeAR: This looks confusing for me. I think we should not fail. I understand that the package.json is not found of the dependency, we just still have to be sure the code will be bundled.

packages/datadog-instrumentations/src/graphql.js
```diff
@@ -360,3 +360,21 @@ addHook({ name: 'graphql', file: 'validation/validate.js', versions: ['>=0.10']
 
   return validate
 })
+
+addHook({ name: 'graphql', file: 'language/printer.js', versions: ['>=0.10'] }, printer => {
+  // HACK: It's possible `graphql` is loaded before `@apollo/gateway` so we
+  //       can't use a channel as the latter plugin would load after the publish
+  //       happened. Not sure how to handle this so for now use a global.
+  globalThis._dd_graphql_printer = printer
```
BridgeAR: I believe we already have an object using a symbol that exposes a dd-trace-js specific object and we should just use that for all additional properties instead of adding it to the global.

Let us prevent further global pollution. If it is not there, let us add it.

packages/datadog-plugin-graphql/src/execute.js
```diff
@@ -66,7 +66,7 @@ function addVariableTags (config, span, variableValues) {
   span.addTags(tags)
 }
 
-function getSignature (document, operationName, operationType, calculate) {
+function getSignature (document, operationName, operationType, calculate, self) {
```
BridgeAR: Unused? Please check for an eslint rule to detect this.

packages/datadog-esbuild/index.js
tlhunter: This might be fine but it's we should definitely log what happens:

```js
} catch (err) {
  if (err.code === 'MODULE_NOT_FOUND') { // is this what you expect?
    if (DEBUG) console.log(`skipping package.json lookup presumably for a vendored package but not entirely sure lol`)
  } else {
    throw err
  }
}
```

.github/dependabot.yml
```diff
@@ -54,11 +54,45 @@ updates:
...
+          - "minor"
+          - "patch"
+  - package-ecosystem: "npm"
+    directories:
+      - "/vendor"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 100
+    cooldown:
+      default-days: 1
```
watson: Should this be increased to 5?

.github/workflows/platform.yml
```diff
@@ -35,6 +35,18 @@ jobs:
...
+  bundle-validate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+      - uses: ./.github/actions/node/active-lts
+      # Running `yarn` also automatically runs `rspack` as a postinstall script.
+      - run: yarn --frozen-lockfile
+        working-directory: vendor
+      # Licenses are pulled from internet and not deterministic so revert them.
+      - run: git checkout packages/node_modules/**/*/LICENSE
```
watson: I don't understand this step. Are we on purpose reverting any license change that the vendored dependency have made to their own license, and if so why?

eslint.config.mjs
```diff
@@ -7,12 +7,15 @@ import eslintPluginMocha from 'eslint-plugin-mocha'
 import eslintPluginN from 'eslint-plugin-n'
 import eslintPluginPromise from 'eslint-plugin-promise'
 import eslintPluginUnicorn from 'eslint-plugin-unicorn'
+import { readFileSync } from 'fs'
```
watson, line 10: Nit: I would prefer to put this in a group by itself that contains Node.js core modules - but I know this is technically unrelated to this PR

eslint.config.mjs
```diff
@@ -7,12 +7,15 @@ import eslintPluginMocha from 'eslint-plugin-mocha'
...
 import globals from 'globals'
 
 import eslintProcessEnv from './eslint-rules/eslint-process-env.mjs'
 import eslintEnvAliases from './eslint-rules/eslint-env-aliases.mjs'
 import eslintSafeTypeOfObject from './eslint-rules/eslint-safe-typeof-object.mjs'
 
+const { dependencies } = JSON.parse(readFileSync('./vendor/package.json', 'utf8'))
```
watson, line 17: Why not this, just as you do in `vendor/rspack.config.js`?

```suggestion
const { dependencies } = require('./vendor/package.json')
```

watson: Looking at how this is used below, I'd actually extend it to be simply:

```suggestion
const allowModules = Object.keys(require('./vendor/package.json').dependencies)
```

And then update the references below

loader-hook.mjs
```diff
@@ -33,7 +33,7 @@ function addSecurityControls (data) {
     .map(sc => sc.trim())
 
   for (const subpath of securityControls) {
-    data.include.push(new RegExp(regexpEscape(subpath)))
+    data.include.push(new RegExp(regexpEscape.default(subpath)))
```
watson: Maybe I'm not understanding this change, but wouldn't it be better to update the import to import the default into `regexpEscape` 🤔

package.json
```diff
@@ -115,58 +115,37 @@
     "packages/*/index.js",
     "packages/*/lib/**/*",
     "packages/*/src/**/*",
+    "packages/node_modules/**/*.js",
+    "packages/node_modules/**/*.ts",
```
watson: Why are we adding `.ts` files to the package?

scripts/check_licenses.js
```diff
@@ -55,12 +55,31 @@ function getProdDeps () {
...
+  const stdout = execSync('yarn list --production --json', {
+    encoding: 'utf8',
+    stdio: ['ignore', 'pipe', 'inherit'],
+    cwd
+  })
+
+  for (const line of stdout.split('\n')) {
+    if (!line) continue
+    const parsed = JSON.parse(line)
+    if (parsed.type === 'tree' && parsed.data && Array.isArray(parsed.data.trees)) {
```
watson:
```suggestion
    if (parsed.type === 'tree' && Array.isArray(parsed.data?.trees)) {
```

scripts/check_licenses.js
```diff
@@ -55,12 +55,31 @@ function getProdDeps () {
     }
   }
 
+  addProdDeps(deps, process.cwd())
```
watson, line 44: Isn't this line duplicating what's above? If so, I think the lines above should be deleted

.github/dependabot.yml
rochdev: This was copied from the same file. Changing this value is out of scope of this PR, unless the other value also changes upstream.

.github/workflows/platform.yml
rochdev: I thought the license was pulled from GitHub and was thus not deterministic as I've seen one of the licenses change for seemingly no reason at some point, but it might have been a version change that I didn't realize, especially since I can't find any evidence of external HTTP calls in the library. I can revert this line and address it if we see any weirdness again in the future.

eslint.config.mjs
rochdev: Yeah I'd rather not make more changes than needed, it's already a pretty large PR.

eslint.config.mjs
rochdev: Using require in ESM can be a bit complicated 😅 It doesn't "just work", I'd have to create an instance of it through `Module`. I figured I'd just stick to ESM best practices instead.

loader-hook.mjs
rochdev: Wouldn't that just move the `.default` to the top? It's not ESM so it doesn't have an actual default export.

package.json
rochdev: It's explained in a comment in `rspack.config.js`, but basically we expose types from some of our dependencies in our own type definition file, so we need to include their types in that scenario.

scripts/check_licenses.js
rochdev: Ah yeah I copied them but forgot to delete them it seems.

eslint.config.mjs
watson: Oh, didn't see this was ESM 🤦 Then this should be the normal way to do it right?

```js
import { dependencies } from './package.json' assert { type: 'json' };
```

loader-hook.mjs
watson: Yes, but I would say that's the expected way to do it, here you're leaking the module "internals" into the code instead of keeping it at the import stage

package.json
watson: I'm probably blind, but I can't find any comment about this in there. Is it specific to `*.d.ts` files and if so, could we scope this to be that as well? If not, we'll just keep it as is.

.github/workflows/platform.yml
watson: Do you mean when you run yarn install, that the `LICENSE` file is then pulled from GitHub? That's not the case, it's just a file like any other in the tarball downloaded from the npm registry. So I wouldn't recommend that we manipulate what the tarball contains like this.

loader-hook.mjs
watson: But it's not a deal breaker for me, just a style thing that I think we should avoid

packages/datadog-plugin-graphql/src/execute.js
```diff
@@ -21,7 +21,7 @@ class GraphQLExecutePlugin extends TracingPlugin {
 
     const span = this.startSpan(this.operationName(), {
       service: this.config.service || this.serviceName(),
-      resource: getSignature(document, name, type, this.config.signature),
+      resource: getSignature(document, name, type, this.config.signature, this),
```
BridgeAR:
```suggestion
      resource: getSignature(document, name, type, this.config.signature),
```

eslint.config.mjs
rochdev: That doesn't work because ESM doesn't consider a property of a default export to be a named export.

loader-hook.mjs
rochdev: What I mean is it would still be in code after the import, but I get what you're saying, at least it's at the top.

package.json
```diff
@@ -115,58 +115,37 @@
     "packages/*/index.js",
     "packages/*/lib/**/*",
     "packages/*/src/**/*",
+    "packages/node_modules/**/*.js",
+    "packages/node_modules/**/*.d.ts",
+    "packages/node_modules/**/*.wasm",
+    "packages/node_modules/**/LICENSE",
```
watson, line 121: We should make sure that LICENSE files that has an extension (e.g. `LICENSE.md` are also included)

rochdev: The plugin is configured to always export `LICENSE` regardless of the source.

packages/datadog-esbuild/index.js
rochdev: Added logging!

BridgeAR (approved):

PR #6951, opened 2025-11-20
test: rewrite assertion to use Node.js assert

BridgeAR:
All of these changes are done by a script including a lint:fix run

Manual changes came afterwards to fix things. These have a separate commit.

PR #6966, opened 2025-11-22
fix(debugger): address TS errors

watson:
### What does this PR do?

Address a large majority of the TS type errors in the debugger code. The only actual change to the runtime logic is the following:
    
When adding a new breakpoint to a file that has a source map, we now validate that the generated position can be identified. If not, we abort and don't add the breakpoint (which most likely would have failed anyway). An error will be sent to the debugger diagnostics endpoint if that happens.
    
Another minor, but hopefully invisible change, is that all private fields in the `JSONBuffer` class are now properly private instead of just being prefixed with an underscore.
    
The only debugger TS errors not addressed in this PR are the following:
- The ones already addressed in https://github.com/DataDog/dd-trace-js/pull/6965
- The ones which will disappear once https://github.com/DataDog/dd-trace-js/pull/6951 lands
    
This PR also adds a `tsconfig.debugger.json` file, which can be used to check the types of the debugger code:
    
```
./node_modules/.bin/tsc --noEmit -p tsconfig.debugger.json
```

### Motivation

Once this PR and the two other mentioned above lands, we can run a TS check as part of CI for all the debugger code.

### DIFF
--- a/integration-tests/debugger/basic.spec.js
+++ b/integration-tests/debugger/basic.spec.js
@@ -124,7 +124,11 @@ describe('Dynamic Instrumentation', function () {
             const expected = expectedPayloads.shift()
             assertObjectContains(event, expected)
             assertUUID(event.debugger.diagnostics.runtimeId)
-            if (event.debugger.diagnostics.status === 'INSTALLED') triggers.shift()()
+            if (event.debugger.diagnostics.status === 'INSTALLED') {
+              const trigger = triggers.shift()
+              assert.ok(trigger, 'expecting a trigger function to be defined')
+              trigger()
+            }
             endIfDone()
           })
         })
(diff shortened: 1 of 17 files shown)


watson:
* **#6966** <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-js/6966?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-js/6966?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

BridgeAR (approved):
I am uncertain why TS is not picking up the correct entries so that `ts-ignore` must be used. This is likely worth investigating at some point.

LGTM for now

PR #6982, opened 2025-11-24
Add class ConfigEnvSources to merge stable config and environment variables

mtoffl01:
### Reviewer tips
The majority of code changes are just replacing `getEnvironmentVariable` with `getValueFromEnvSources`. Feel free to gloss over those changes once you get the idea from a few, and instead focus on:
- packages/dd-trace/src/config-helper.js
- packages/dd-trace-/src/config.js
- packages/dd-trace-/src/profiling/...

### Background

Currently, stable config files and environment variables are only processed when the Config class is instantiated, which happens in tracer.init(). This limits us from earlier access: code that runs before tracer.init() can't access stable config values.
This PR allows us to access stable config values before tracer.init(), when we peek at DD_TRACE env vars.
It also deduplicates configuration sourcing in profiler.

### What does this PR do?

Introduces `getValueFromEnvSources` API in config-helper.js that resolves configuration from local stable config, environment variables, and fleet stable config in ascending priority order, before the `Config` singleton is initialized. This method can be used instead of `getEnvironmentVariable` when we want to see whether particular DD_TRACE configs have been set.

#### Key additions:
**config-helper**:
- getValueFromEnvSources(): Queries for the value at specified config in fleet stable config, env var, and local stable config input, in that order
- Comprehensive tests for the new config-helper functionality

**profiler**
- Profiling config now consumes tracer Config fields: profiling/config.js reads options passed down from the main Config singleton, instead of re-reading them directly. For "DD_" configs that are not yet processed by the Config singleton, we use getValueFromEnvSources in order to resolve them from stable config inputs in addition to process.env
- Profiling tests updated to mirror runtime behavior: profiling/config.spec.js and profiling/profiler.spec.js now construct profiling Config with a realistic agent URL (and hostname/port when relevant), matching what the tracer Config would provide at runtime, so tests exercise the same code paths the profiler uses in production.

### Motivation

All environment variables must be set-able via stable config files: [RFC](https://docs.google.com/document/d/1VXMxdMzrm7HpvVqJENo0u9PNN9qEe-UVioqAMfVJkOI/edit?tab=t.0#heading=h.rq5ego28zgx8)

### Note

Not all calls to getEnvironmentVariable were replaced by getResolvedEnv. In some cases, we are really looking for an environment variable that is a heuristic for a particular type of environment. Stable config is not relevant in these cases.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/.github/workflows/dependabot-automation.yml
+++ b/.github/workflows/dependabot-automation.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     types:
       - opened
+      - synchronize
 
 env:
   GROUPS: '["dev-minor-and-patch-dependencies", "gh-actions-packages", "test-versions"]'
@@ -38,35 +39,150 @@ jobs:
           PR_URL: ${{ github.event.pull_request.html_url }}
           GH_TOKEN: ${{ steps.octo-sts.outputs.token }}
 
-  vendor:
+  vendor-build:
     if: github.event.pull_request.user.login == 'dependabot[bot]'
     runs-on: ubuntu-latest
+    # Security: this job checks out and runs code from the PR (vendoring build),
+    # so it is intentionally restricted to read-only permissions and produces a
+    # patch artifact instead of pushing directly.
     permissions:
-      id-token: write
-      contents: write
+      contents: read
+      pull-requests: read
+    outputs:
+      has_changes: ${{ steps.diff.outputs.has_changes }}
+      is_vendor_group: ${{ steps.ctx.outputs.is_vendor_group }}
     steps:
       - name: Dependabot metadata
         id: metadata
         uses: dependabot/fetch-metadata@21025c705c08248db411dc16f3619e6b5f9ea21a # 2.5.0
-      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
-        if: steps.metadata.outputs.dependency-group == 'vendor-minor-and-patch-dependencies'
-      - run: yarn
-        working-directory: ./vendor
-        if: steps.metadata.outputs.dependency-group == 'vendor-minor-and-patch-dependencies'
-      - name: Create commits
-        id: create-commits
+      - name: Compute vendor context
+        id: ctx
         run: |
+          set -euo pipefail
+
+          echo "is_vendor_group=${{ steps.metadata.outputs.dependency-group == 'vendor-minor-and-patch-dependencies' }}" >> $GITHUB_OUTPUT
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        if: steps.ctx.outputs.is_vendor_group == 'true'
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 1
+          persist-credentials: false
+      - name: Restore trusted vendoring scripts
+        if: steps.ctx.outputs.is_vendor_group == 'true'
+        run: |
+          git fetch --no-tags --depth=1 origin "${{ github.event.pull_request.base.sha }}"
+          git checkout "${{ github.event.pull_request.base.sha }}" -- vendor/rspack.js vendor/rspack.config.js
+      - name: Install vendoring deps (no lifecycle scripts)
+        if: steps.ctx.outputs.is_vendor_group == 'true'
+        run: yarn --ignore-scripts --frozen-lockfile --non-interactive
+        working-directory: ./vendor
+      - name: Build vendored bundles (trusted script)
+        if: steps.ctx.outputs.is_vendor_group == 'true'
+        run: node rspack
+        working-directory: ./vendor
+      - name: Create patch (restricted paths only)
+        id: diff
+        run: |
+          set -euo pipefail
+
+          if [ "${{ steps.ctx.outputs.is_vendor_group }}" != "true" ]; then
+            echo "has_changes=false" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+
+          if git diff --quiet; then
+            echo "has_changes=false" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+
+          allowed_prefix_1="vendor/dist/"
+          allowed_file_1="vendor/package.json"
+          allowed_file_2="vendor/yarn.lock"
+
+          bad=0
+          while IFS= read -r file; do
+            case "$file" in
+              "$allowed_file_1" | "$allowed_file_2" | "$allowed_prefix_1"*)
+                ;;
+              *)
+                echo "Unexpected changed path: $file"
+                bad=1
+                ;;
+            esac
+          done < <(git diff --name-only)
+
+          if [ "$bad" -ne 0 ]; then
+            echo "Refusing to proceed: unexpected paths changed during vendoring."
+            exit 1
+          fi
+
+          git diff --binary --no-color > vendor.patch
+          echo "has_changes=true" >> $GITHUB_OUTPUT
+      - uses: actions/upload-artifact@ea165f8a6f3f9f5b76b7b70ee0fdd3b5d7d3b6a2 # v4.6.2
+        if: steps.diff.outputs.has_changes == 'true'
+        with:
+          name: vendor-patch
+          path: vendor.patch
+          if-no-files-found: error
+
+  vendor-push:
+    if: github.event.pull_request.user.login == 'dependabot[bot]' && needs.vendor-build.outputs.is_vendor_group == 'true' && needs.vendor-build.outputs.has_changes == 'true'
+    runs-on: ubuntu-latest
+    needs: vendor-build
+    # Security: this job has write permissions but never runs installs/builds.
+    # It only applies the vetted patch artifact and pushes a single commit.
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.sha }}
+          persist-credentials: false
+      - uses: actions/download-artifact@65c5b1180b77f85ea54a67a3d1f4d5f1e2e50bd8 # v4.2.0
+        with:
+          name: vendor-patch
+      - name: Apply patch
+        run: git apply --whitespace=nowarn vendor.patch
+      - name: Validate changed paths
+        run: |
+          set -euo pipefail
+
+          allowed_prefix_1="vendor/dist/"
+          allowed_file_1="vendor/package.json"
+          allowed_file_2="vendor/yarn.lock"
+
+          bad=0
+          while IFS= read -r file; do
+            case "$file" in
+              "$allowed_file_1" | "$allowed_file_2" | "$allowed_prefix_1"*)
+                ;;
+              *)
+                echo "Unexpected changed path after applying patch: $file"
+                bad=1
+                ;;
+            esac
+          done < <(git diff --name-only)
+
+          if [ "$bad" -ne 0 ]; then
+            echo "Refusing to proceed: unexpected paths changed."
+            exit 1
+          fi
+      - name: Create commit
+        id: create-commit
+        run: |
+          set -euo pipefail
+
           git config user.name github-actions
           git config user.email github-actions@github.com
+
           git add -A
           git commit -m "update vendored dependencies with new versions"
-
-          echo "commit=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
-        if: steps.metadata.outputs.dependency-group == 'vendor-minor-and-patch-dependencies'
-      - name: Push commits
+          echo "commits=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
+      - name: Push commit
         uses: DataDog/commit-headless@583489e08d78037e7fa256c14adf998d5463f6a0 # action/v2.0.2
         with:
-          branch: ${{ github.ref_name }}
+          branch: ${{ github.event.pull_request.head.ref }}
           command: push
-          commits: "${{ steps.create-commits.outputs.commits }}"
-        if: steps.metadata.outputs.dependency-group == 'vendor-minor-and-patch-dependencies'
+          commits: "${{ steps.create-commit.outputs.commits }}"
--- a/packages/datadog-instrumentations/src/pg.js
+++ b/packages/datadog-instrumentations/src/pg.js
@@ -112,8 +112,10 @@ function wrapQuery (query) {
       arguments[0] = pgQuery
 
       const retval = query.apply(this, arguments)
-      const queryQueue = this._queryQueue || this.queryQueue
-      const activeQuery = Object.hasOwn(this, '_activeQuery') ? this._activeQuery : this.activeQuery
+
+      const deperecated = Object.hasOwn(this, '_activeQuery')
+      const queryQueue = deperecated ? this._queryQueue : this.queryQueue
+      const activeQuery = deperecated ? this._activeQuery : this.activeQuery
 
       const newQuery = queryQueue.at(-1) || activeQuery
 
--- a/packages/datadog-plugin-body-parser/test/integration-test/client.spec.js
+++ b/packages/datadog-plugin-body-parser/test/integration-test/client.spec.js
@@ -1,0 +1,39 @@
+'use strict'
+
+const assert = require('node:assert/strict')
+const axios = require('axios')
+const {
+  sandboxCwd, useSandbox, varySandbox,
+  FakeAgent, spawnPluginIntegrationTestProc
+} = require('../../../../integration-tests/helpers')
+const { withVersions } = require('../../../dd-trace/test/setup/mocha')
+
+withVersions('body-parser', 'body-parser', version => {
+  describe('ESM', () => {
+    let variants, proc, agent
+
+    useSandbox([`'body-parser@${version}'`, 'express'], false,
+      ['./packages/datadog-plugin-body-parser/test/integration-test/*'])
+
+    before(function () {
+      variants = varySandbox('server.mjs', 'bodyParser', undefined, 'body-parser')
+    })
+
+    beforeEach(async () => {
+      agent = await new FakeAgent().start()
+    })
+
+    afterEach(async () => {
+      proc?.kill()
+      await agent.stop()
+    })
+
+    for (const variant of varySandbox.VARIANTS) {
+      it(`is instrumented ${variant}`, async () => {
+        proc = await spawnPluginIntegrationTestProc(sandboxCwd(), variants[variant], agent.port)
+        const response = await axios.post(`${proc.url}/`, { key: 'value' })
+        assert.equal(response.headers['x-counter'], '1')
+      })
+    }
+  })
+})
--- a/packages/datadog-plugin-body-parser/test/integration-test/server.mjs
+++ b/packages/datadog-plugin-body-parser/test/integration-test/server.mjs
@@ -1,0 +1,22 @@
+import 'dd-trace/init.js'
+import express from 'express'
+import bodyParser from 'body-parser'
+import dc from 'dc-polyfill'
+const bodyParserReadCh = dc.channel('datadog:body-parser:read:finish')
+
+let counter = 0
+bodyParserReadCh.subscribe(() => {
+  counter += 1
+})
+const app = express()
+
+app.use(bodyParser.json())
+app.post('/', (req, res) => {
+  res.setHeader('X-Counter', counter)
+  res.end('hello, world\n')
+})
+
+const server = app.listen(0, () => {
+  const port = server.address().port
+  process.send({ port })
+})
--- a/packages/datadog-plugin-pg/test/index.spec.js
+++ b/packages/datadog-plugin-pg/test/index.spec.js
@@ -27,7 +27,7 @@ describe('Plugin', () => {
   let tracer
 
   describe('pg', () => {
-    withVersions('pg', 'pg', version => {
+    withVersions('pg', 'pg', (version) => {
       beforeEach(() => {
         tracer = require('../../dd-trace')
       })
@@ -509,6 +509,8 @@ describe('Plugin', () => {
 
           client.connect(err => done(err))
 
+          const queryQueueName = Object.hasOwn(client, '_queryQueue') ? '_queryQueue' : 'queryQueue'
+
           client.query('SELECT $1::text as message', ['Hello world!'], (err, result) => {
             if (err) return done(err)
 
@@ -516,9 +518,9 @@ describe('Plugin', () => {
               if (err) return done(err)
             })
           })
-          if (client.queryQueue[0] !== undefined) {
+          if (client[queryQueueName][0]) {
             try {
-              assert.strictEqual(client.queryQueue[0].text,
+              assert.strictEqual(client[queryQueueName][0].text,
                 '/*dddb=\'postgres\',dddbs=\'serviced\',dde=\'tester\',ddh=\'127.0.0.1\',ddps=\'test\',' +
                 `ddpv='${ddpv}'*/ SELECT $1::text as message`)
             } catch (e) {
@@ -566,6 +568,8 @@ describe('Plugin', () => {
         })
 
         it('DBM propagation should handle special characters', done => {
+          const queryQueueName = Object.hasOwn(clientDBM, '_queryQueue') ? '_queryQueue' : 'queryQueue'
+
           clientDBM.query('SELECT $1::text as message', ['Hello world!'], (err, result) => {
             if (err) return done(err)
 
@@ -574,9 +578,9 @@ describe('Plugin', () => {
             })
           })
 
-          if (clientDBM.queryQueue[0] !== undefined) {
+          if (clientDBM[queryQueueName][0]) {
             try {
-              assert.strictEqual(clientDBM.queryQueue[0].text,
+              assert.strictEqual(clientDBM[queryQueueName][0].text,
                 '/*dddb=\'postgres\',dddbs=\'~!%40%23%24%25%5E%26*()_%2B%7C%3F%3F%2F%3C%3E\',dde=\'tester\',' +
                 `ddh='127.0.0.1',ddps='test',ddpv='${ddpv}'*/ SELECT $1::text as message`)
               done()
@@ -682,6 +686,7 @@ describe('Plugin', () => {
 
       describe('DBM propagation enabled with full should handle query config objects', () => {
         const tracer = require('../../dd-trace')
+        let queryQueueName
 
         before(() => {
           return agent.load('pg')
@@ -702,6 +707,9 @@ describe('Plugin', () => {
             password: 'postgres',
             database: 'postgres'
           })
+
+          queryQueueName = Object.hasOwn(client, '_queryQueue') ? '_queryQueue' : 'queryQueue'
+
           client.connect(err => done(err))
         })
 
@@ -717,7 +725,7 @@ describe('Plugin', () => {
           }
 
           const queryPromise = client.query(query, ['Hello world!'])
-          const queryText = client.queryQueue[0].text
+          const queryText = client[queryQueueName][0].text
 
           await queryPromise
 
@@ -739,7 +747,7 @@ describe('Plugin', () => {
           }
 
           const queryPromise = client.query(query, ['Hello world!'])
-          const queryText = client.queryQueue[0].text
+          const queryText = client[queryQueueName][0].text
 
           await queryPromise
 
@@ -770,7 +778,7 @@ describe('Plugin', () => {
           client.query(query, ['Hello world!'], (err) => {
             done(err)
           })
-          assert.strictEqual(client.queryQueue[0].text,
+          assert.strictEqual(client[queryQueueName][0].text,
             `/*dddb='postgres',dddbs='post',dde='tester',ddh='127.0.0.1',ddps='test',ddpv='${ddpv}'` +
             '*/ SELECT $1::text as message'
           )
@@ -785,7 +793,7 @@ describe('Plugin', () => {
           client.query(query, ['Hello world!'], async (err) => {
             done(err)
           })
-          assert.strictEqual(client.queryQueue[0].text,
+          assert.strictEqual(client[queryQueueName][0].text,
             `/*dddb='postgres',dddbs='post',dde='tester',ddh='127.0.0.1',ddps='test',ddpv='${ddpv}'` +
             '*/ SELECT $1::text as message')
         })
@@ -809,7 +817,7 @@ describe('Plugin', () => {
           client.query(query, ['Goodbye'], (err) => {
             done(err)
           })
-          assert.strictEqual(client.queryQueue[0].text,
+          assert.strictEqual(client[queryQueueName][0].text,
             `/*dddb='postgres',dddbs='post',dde='tester',ddh='127.0.0.1',ddps='test',ddpv='${ddpv}'` +
             '*/ SELECT $1::text as greeting')
         })
@@ -844,9 +852,11 @@ describe('Plugin', () => {
         })
 
         it('should append comment in query text', async () => {
+          const queryQueueName = Object.hasOwn(client, '_queryQueue') ? '_queryQueue' : 'queryQueue'
+
           const queryPromise = client.query('SELECT $1::text as message', ['Hello world!'])
 
-          assert.strictEqual(client.queryQueue[0].text,
+          assert.strictEqual(client[queryQueueName][0].text,
             'SELECT $1::text as message /*dddb=\'postgres\',dddbs=\'serviced\',dde=\'tester\',' +
               `ddh='127.0.0.1',ddps='test',ddpv='${ddpv}'*/`
           )
--- a/packages/dd-trace/test/plugins/plugin-structure.spec.js
+++ b/packages/dd-trace/test/plugins/plugin-structure.spec.js
@@ -37,6 +37,7 @@ const missingPlugins = [
   'datadog-plugin-pug', // pug does not produce spans
   'datadog-plugin-vm', // vm does not produce spans
   'datadog-plugin-sequelize', // sequelize does not produce spans
+  'datadog-plugin-body-parser', // body-parser does not produce spans
 ]
 
 // instrumentations that do not have a hook, but are still instrumented


packages/dd-trace/src/config_stable.js
```diff
@@ -17,6 +17,7 @@ class StableConfig {
       return
     }
 
+    // cache the file reads
```
mtoffl01: Tried this here: https://github.com/DataDog/dd-trace-js/pull/7035/files
TLDR not sure it's worth it

packages/dd-trace/src/ci-visibility/exporters/test-worker/index.js
```diff
@@ -26,6 +26,7 @@ function getInterprocessTraceCode () {
   if (getEnvironmentVariable('DD_PLAYWRIGHT_WORKER')) {
     return PLAYWRIGHT_WORKER_TRACE_PAYLOAD_CODE
   }
```
mtoffl01: I'm not sure what this env var does

packages/dd-trace/src/config-helper.js
```diff
@@ -95,5 +94,31 @@ module.exports = {
...
+   * Returns the value stored at the given name, assumed to be in environment variable format,
+   * from the provided source. Falls back to aliases if the canonical name is not set.
+   * Used for stable config sources.
+   *
+   * @param {string} name Environment variable name
+   * @param {object} source Environment source
+   * @returns {string|undefined}
+   * @throws {Error} if the configuration is not supported
+   */
+  getValueFromEnvSource (name, source) {
```
mtoffl01: For reviewers: I created this helper function in here, even though it is only used by ConfigEnvSources, because:
- this class already has access to `supportedConfigurations` and `aliases`
- process.env can only be accessed from inside config-helpers.js (and I agree with that!)

BridgeAR: I think we could just change `getEnvironmentVariable()` to accept a source as second argument that falls back to process.env, if not defined.

That way there is less code duplication :)

packages/dd-trace/src/config-env-sources.js
```diff
@@ -0,0 +1,112 @@
...
+  }
+  return configEnvSourcesInstance
+}
+
+/**
+ * Reset the singleton instance (useful for testing)
+ */
+function resetConfigEnvSources () {
+  configEnvSourcesInstance = null
+}
```
BridgeAR: I think we can just use proxyquire to achieve the same instead of writing dedicated production code to make things testable :)

packages/dd-trace/src/config-env-sources.js
```diff
@@ -0,0 +1,112 @@
...
+      const StableConfig = require('./config_stable')
+      const instance = new StableConfig()
+      return {
+        localEntries: instance.localEntries ?? {},
+        fleetEntries: instance.fleetEntries ?? {},
+        warnings: instance.warnings ?? []
+      }
+    } catch {
+      // Stable config is optional, continue without it
+      return null
```
BridgeAR: Is there a reason this could fail?

packages/dd-trace/src/config-env-sources.js
```diff
@@ -0,0 +1,112 @@
...
+    return getValueFromEnvSource(name, config.localStableConfig)
+  }
+}
+
+module.exports = {
+  ConfigEnvSources,
+  createConfigEnvSources,
+  getConfigEnvSources,
+  resetConfigEnvSources,
+  getResolvedEnv
```
BridgeAR:
```suggestion
  getResolvedEnv: isInServerlessEnvironment() ? getEnvironmentVariable : getResolvedEnv
```

That way it is a direct env lookup in serverless and the special handling is not needed in the class above.

We can in fact also simplify the class to not exist (which I would recommend).

packages/dd-trace/src/config-helper.js
BridgeAR: After looking at it more, I think it would in fact be best to handle it as follows:

1. Add the second argument. Also to get all envs.
2. Move the resolving of stable config in here.
3. If no argument is passed through, resolve all sources in correct order.
4. If a source is passed through, only check that one.

That way almost no code change would be needed and we achieve the behavior we like.

packages/dd-trace/src/config/supported-configurations.json
```diff
@@ -93,6 +93,7 @@
     "DD_GIT_TAG": ["A"],
     "DD_GIT_PULL_REQUEST_BASE_BRANCH": ["A"],
     "DD_GIT_PULL_REQUEST_BASE_BRANCH_SHA": ["A"],
+    "DD_GIT_COMMIT_HEAD_SHA": ["A"],
```
BridgeAR, line 96: I know this is actually independent and fixing a former bug. Could you add a test case / adjust a test for it nevertheless? That way we would not regress :)

packages/dd-trace/src/config-helper.js
```diff
@@ -34,7 +34,80 @@ for (const deprecation of Object.keys(deprecations)) {
...
+let hasLoadedStableConfig = false
+
+// Test-only helper to reset stable config state between test cases.
+// Not intended for production use.
+function _resetStableConfigForTesting () {
+  localStableConfig = {}
+  fleetStableConfig = {}
+  stableConfigWarnings = []
+  hasLoadedStableConfig = false
+}
```
BridgeAR: We can use proxyquire to achieve the same without having to expose a test helper :)

That way we do not write testing code in our production code.

packages/dd-trace/src/profiling/config.js
```diff
@@ -40,40 +39,32 @@ class Config {
...
       DD_PROFILING_TIMELINE_ENABLED,
       DD_PROFILING_UPLOAD_PERIOD,
       DD_PROFILING_UPLOAD_TIMEOUT,
       DD_PROFILING_V8_PROFILER_BUG_WORKAROUND,
       DD_PROFILING_WALLTIME_ENABLED,
-      DD_SERVICE,
       DD_TAGS,
-      DD_TRACE_AGENT_PORT,
-      DD_TRACE_AGENT_URL,
-      DD_VERSION,
```
BridgeAR, line 53: This is a great cleanup!

packages/dd-trace/src/plugin_manager.js
```diff
@@ -5,6 +5,7 @@ const { isFalse, isTrue, normalizePluginEnvName } = require('./util')
 const plugins = require('./plugins')
 const log = require('./log')
 const { getEnvironmentVariable } = require('../../dd-trace/src/config-helper')
+const { getValueFromEnvSources } = require('./config-helper')
```
watson: These two should be merged

packages/dd-trace/src/plugin_manager.js
```diff
@@ -94,7 +95,8 @@ module.exports = class PluginManager {
       this._pluginsByName[name] = new Plugin(this._tracer, this._tracerConfig)
     }
     const pluginConfig = this._configsByName[name] || {
-      enabled: this._tracerConfig.plugins !== false && (!Plugin.experimental || isTrue(getEnvEnabled(Plugin)))
+      enabled: this._tracerConfig.plugins !== false &&
+        (!Plugin.experimental || isTrue(getEnabled(Plugin)))
```
watson, line 99: Was this change intentional?

packages/dd-trace/src/plugins/util/ci.js
```diff
@@ -691,9 +691,9 @@ module.exports = {
     if (env.CODEBUILD_INITIATOR?.startsWith('codepipeline/')) {
       const {
         CODEBUILD_BUILD_ARN,
-        DD_ACTION_EXECUTION_ID,
-        DD_PIPELINE_EXECUTION_ID
       } = env
```
watson: Nit: I would personally write this as:

```js
const { CODEBUILD_BUILD_ARN } = env
```

packages/dd-trace/src/profiling/config.js
```diff
@@ -16,16 +16,65 @@ const { getIsAzureFunction } = require('../serverless')
 const { tagger } = require('./tagger')
 const { isFalse, isTrue } = require('../util')
 const { getAzureTagsFromMetadata, getAzureAppMetadata, getAzureFunctionMetadata } = require('../azure_metadata')
-const { getEnvironmentVariables } = require('../config-helper')
-const defaults = require('../config_defaults')
+const { getEnvironmentVariable, getValueFromEnvSources } = require('../config-helper')
+
+function getProfilingEnvValues () {
```
watson: Nit: I usually put utility functions at the end of the file. That way the important code in the file is at the top, which makes reading it easier

packages/dd-trace/src/profiling/config.js
```diff
@@ -16,16 +16,65 @@ const { getIsAzureFunction } = require('../serverless')
...
+  return {
+    DD_INTERNAL_PROFILING_TIMELINE_SAMPLING_ENABLED:
+    getValueFromEnvSources('DD_INTERNAL_PROFILING_TIMELINE_SAMPLING_ENABLED'),
```
watson:
```suggestion
      getValueFromEnvSources('DD_INTERNAL_PROFILING_TIMELINE_SAMPLING_ENABLED'),
```

packages/dd-trace/src/telemetry/endpoints.js
```diff
@@ -141,7 +201,6 @@ function stop () {
 
   pendingEndpoints.clear()
   flushScheduled = false
-  config = application = host = getRetryData = updateRetryData = null
```
watson, line 146: I was concerned that not clearing `getRetryData` and `updateRetryData` might keep a closure alive which would hinder GC in cleaning up something. But as far as I can see these two functions are also used elsewhere, so not clearing them here should be ok.

packages/dd-trace/test/config.spec.js
```diff
@@ -12,7 +14,10 @@ const path = require('node:path')
 require('./setup/core')
 
 const { GRPC_CLIENT_ERROR_STATUSES, GRPC_SERVER_ERROR_STATUSES } = require('../src/constants')
-const { getEnvironmentVariable, getEnvironmentVariables } = require('../src/config-helper')
+const {
+  getEnvironmentVariable,
+  getEnvironmentVariables,
+} = require('../src/config-helper')
```
watson: This change seems unnecessary?

packages/dd-trace/test/profiling/config.spec.js
```diff
@@ -527,7 +547,9 @@ describe('config', () => {
         } else {
           process.env.DD_PROFILING_ASYNC_CONTEXT_FRAME_ENABLED = '0'
           try {
-            const config = new Config({})
+            const config = new Config({
+              url: 'http://127.0.0.1:8126'
```
watson, line 553: For completeness, do you want to also add the comment from below here?

```suggestion
              // In production this comes from the tracer Config singleton; we mimic it here.
              url: 'http://127.0.0.1:8126'
```

packages/dd-trace/src/config/stable.js
```diff
@@ -70,32 +66,24 @@ class StableConfig {
...
-      case 'linux':
-        localConfigPath = '/etc/datadog-agent/application_monitoring.yaml'
-        fleetConfigPath = '/etc/datadog-agent/managed/datadog-agent/stable/application_monitoring.yaml'
-        break
+  #getStableConfigPaths () {
+    // TODO(BridgeAR): Remove these environment variables once we have a proper way to test the stable config.
+    // Allow overriding the paths for testing
+    let localConfigPath = getEnvironmentVariable('DD_TEST_LOCAL_CONFIG_PATH')
+    let fleetConfigPath = getEnvironmentVariable('DD_TEST_FLEET_CONFIG_PATH')
+    switch (os.platform()) {
```
watson, line 74: If both `localConfigPath` and `fleetConfigPath` are set, we could return early and skip the switch statement. Do you think it's worth it?

packages/dd-trace/src/config/stable.js
```diff
@@ -59,7 +55,7 @@ class StableConfig {
     }
   }
 
-  _readConfigFromPath (path) {
+  #readConfigFromPath (path) {
```
watson, line 58: This function doesn't need to return an empty string any more if it enters the `catch` block as the place where we read the function return value just aborts if the return value is falsy

packages/dd-trace/src/config-helper.js
```diff
@@ -72,28 +125,47 @@ module.exports = {
     return configs
   },
 
+  getEnvironmentVariable (name) {
+    if ((name.startsWith('DD_') || name.startsWith('OTEL_') || aliasToCanonical[name]) &&
+      !supportedConfigurations[name]) {
+      throw new Error(`Missing ${name} env/configuration in "supported-configurations.json" file.`)
+    }
```
watson: Could we pull this guard out into a helper function as it's repeated in `getValueFromEnvSources`

packages/dd-trace/src/config-helper.js
```diff
@@ -34,7 +34,59 @@ for (const deprecation of Object.keys(deprecations)) {
...
+let fleetStableConfig
+let stableConfigWarnings
+let stableConfigLoaded = false
+
+function loadStableConfig () {
+  stableConfigLoaded = true
+
+  // Lazy require to avoid circular dependency at module load time.
+  const { isInServerlessEnvironment } = require('./serverless')
+  if (isInServerlessEnvironment && isInServerlessEnvironment()) {
```
watson: When is `isInServerlessEnvironment` ever not present?!?

```suggestion
  if (isInServerlessEnvironment()) {
```

packages/dd-trace/src/config-helper.js
```diff
@@ -34,7 +34,59 @@ for (const deprecation of Object.keys(deprecations)) {
...
+   * @returns {{ localStableConfig: object, fleetStableConfig: object, stableConfigWarnings: string[] }}
+   */
+  getStableConfigSources () {
+    if (!stableConfigLoaded) {
+      loadStableConfig()
+    }
+    return {
+      localStableConfig: localStableConfig ?? {},
+      fleetStableConfig: fleetStableConfig ?? {},
+      stableConfigWarnings: stableConfigWarnings ?? []
```
watson: As far as I can see these 3 variables will never be `undefined`, so this could be written as:

```suggestion
      localStableConfig,
      fleetStableConfig,
      stableConfigWarnings
```

packages/dd-trace/src/config/helper.js
```diff
@@ -72,28 +125,47 @@ module.exports = {
...
+    if (!stableConfigLoaded) {
+      loadStableConfig()
+    }
+
+    if (fleetStableConfig !== undefined) {
+      const fromFleet = getValueFromSource(name, fleetStableConfig)
+      if (fromFleet !== undefined) {
+        return fromFleet
       }
     }
```
watson, line 161: As far as I can see this variable will never be `undefined`, so this could be written as:

```suggestion
    const fromFleet = getValueFromSource(name, fleetStableConfig)
    if (fromFleet !== undefined) {
      return fromFleet
    }
```

packages/dd-trace/src/config/helper.js
```diff
@@ -72,28 +125,47 @@ module.exports = {
...
-  }
+
+    const fromEnv = getValueFromSource(name, process.env)
+    if (fromEnv !== undefined) {
+      return fromEnv
+    }
+
+    if (localStableConfig !== undefined) {
+      return getValueFromSource(name, localStableConfig)
+    }
```
watson, line 170: As far as I can see this variable will never be `undefined`, so this could be written as:

```suggestion
    return getValueFromSource(name, localStableConfig)
```

packages/dd-trace/src/telemetry/telemetry.js
```diff
@@ -95,13 +189,11 @@ function getProducts (config) {
       enabled: profilingEnabledToBoolean(config.profiling.enabled)
     }
   }
-  if (errors.profilingError) {
-    products.profiler.error = errors.profilingError
-    errors.profilingError = {}
-  }
```
watson, line 105: It's not clear to me why this change was made?

packages/dd-trace/src/telemetry/telemetry.js
```diff
@@ -159,50 +257,55 @@ function createAppObject (config) {
...
-    return {
-      hostname: os.hostname(),
-      os: osName,
-      architecture: os.arch(),
-      kernel_version: os.version(),
-      kernel_release: os.release(),
-      kernel_name: osName
-    }
+  const base = {
+    hostname: os.hostname(), // TODO is this enough?
```
watson: Was this comment originally meant for if returning only `hostname` and `os` was enough, or is it specifically related to `os.hostname()`? It seems inconsistent.

packages/dd-trace/src/telemetry/telemetry.js
BridgeAR: It was a cleanup when I looked into this code.
It is never defined by any code, so it will never be reached and is dead code.

packages/dd-trace/src/telemetry/telemetry.js
BridgeAR: It is specific to `os.hostname()`. I will update the comment to clarify that

packages/dd-trace/src/plugin_manager.js
BridgeAR: You mean to use two lines instead of one? I find it nicer to read like that, while I have no strong opinion.

packages/dd-trace/src/config/stable.js
BridgeAR: While that is true, it will be addressed anyway by the comment that the envs should never be used (the comment changed to a TODO in a later commit than the reviewed change here)

packages/dd-trace/src/config/stable.js
BridgeAR: It would still be needed if one call alone fails and the other passes, since we pass these variables to `get_configuration` later. We actually do not need the `toString()` calls though.
I am uncertain if `get_configuration()` accepts undefined, I will check for that.

I checked, it is not accepted.

packages/dd-trace/src/config-helper.js
BridgeAR: Good catch

packages/dd-trace/src/config/helper.js
BridgeAR: If it is in serverless, it would be undefined.

packages/dd-trace/src/config/helper.js
BridgeAR: If it is in serverless, it would be undefined

packages/dd-trace/src/config.js
```diff
@@ -277,8 +277,12 @@ class Config {
   constructor (options = {}) {
     if (!isInServerlessEnvironment()) {
       // Bail out early if we're in a serverless environment, stable config isn't supported
```
watson: Nit: This comment is misleading. I suggest changing it while we're here anyway. Either remove it completely as it's not really adding anything not clearly stated by the code itself, or change it to:

```suggestion
      // Stable config is not available in a serverless environment
```

packages/dd-trace/src/config-helper.js
watson: As you said in some of your other comments, this is not strictly true. In case of a serverless environment, these will be undefined. However, in the code we never actually call this function in that case, so it should work without the `undefined` check. However, the JSDoc `@returns` is now technically wrong I guess. But it's all a mess because if you guard the call `getStableConfigSources` with a check for if we're not in serverless, then they are guaranteed to be objects, so 🤷

BridgeAR: Yes, I just checked the overall calls and we guard against calling this in config.js by first checking if the environment is not serverless.

I guess it is a minimal risk to not have the fallback in case a refactoring would ever change that in a weird way. Otherwise this should be good as is. That also applies to the JSDoc as such 😅 

I am fine either way. What is your preferred way of handling this?

packages/dd-trace/src/config.js
BridgeAR:
```suggestion
```

packages/dd-trace/src/exporters/agent/writer.js
```diff
@@ -79,23 +82,15 @@ function makeRequest (version, data, count, url, headers, lookup, needsStartupLo
       'X-Datadog-Trace-Count': String(count),
       'Datadog-Meta-Lang': 'nodejs',
       'Datadog-Meta-Lang-Version': process.version,
-      'Datadog-Meta-Lang-Interpreter': process.jsEngine || 'v8'
+      'Datadog-Meta-Lang-Interpreter': process.versions.bun ? 'JavaScriptCore' : 'V8'
```
BridgeAR: Note: I believe this is just a value we read, while it could theoretically have an impact on dashboards that look for the lowercase version (most cases would be case insensitive anyway though).

packages/dd-trace/src/exporters/agent/writer.js
```diff
@@ -26,7 +28,10 @@ class AgentWriter extends BaseWriter {
     runtimeMetrics.increment(`${METRIC_PREFIX}.requests`, true)
 
     const { _headers, _lookup, _protocolVersion, _url } = this
-    makeRequest(_protocolVersion, data, count, _url, _headers, _lookup, true, (err, res, status) => {
+    makeRequest(_protocolVersion, data, count, _url, _headers, _lookup, (err, res, status) => {
+      // Note that logging will only happen once, regardless of how many times this is called.
+      startupLog(status !== 404 && status !== 200 ? { status, message: err?.message ?? inspect(err) } : undefined)
```
BridgeAR, line 33: Note: I am uncertain about the fallback (it should never happen that an error has no message, but just in theory). We would receive the full stacktrace in case it's still a regular error, which would be likely.

packages/dd-trace/src/startup-log.js
```diff
@@ -42,8 +34,8 @@ function startupLog ({ agentError } = {}) {
   if (agentError) {
     warn('DATADOG TRACER DIAGNOSTIC - Agent Error: ' + agentError.message)
     errors.agentError = {
-      code: agentError.code ?? '',
-      message: `Agent Error:${agentError.message}`
+      code: agentError.status,
```
BridgeAR, line 36: I am actually uncertain, if this is definitely correct. I have to verify that again.

packages/dd-trace/src/telemetry/telemetry.js
```diff
@@ -124,11 +221,10 @@ function appStarted (config) {
...
     app.install_signature = installSignature
   }
-  // TODO: add app.error with correct error codes
-  // if (errors.agentError) {
-  //   app.error = errors.agentError
-  //   errors.agentError = {}
-  // }
+  if (errors.agentError) {
+    app.error = errors.agentError
+    errors.agentError = undefined
```
BridgeAR, line 228: I am unsure why this was never activated before. I am going to try to verify that before landing.

packages/dd-trace/src/telemetry/telemetry.js
```diff
@@ -162,50 +262,55 @@ function createAppObject (config) {
...
   }
 
-  if (osName === 'Windows_NT') {
-    return {
-      hostname: os.hostname(),
-      os: osName,
-      architecture: os.arch(),
-      os_version: os.version()
-    }
+  if (os.platform() === 'win32') {
```
BridgeAR, line 273: Note: This is actually a bug fix.

packages/dd-trace/src/config-helper.js
watson: Feel free to keep it as is

packages/dd-trace/src/config/stable.js
```diff
@@ -59,43 +55,35 @@ class StableConfig {
...
-  _getStableConfigPaths () {
-    let localConfigPath = ''
-    let fleetConfigPath = ''
-    switch (os.type().toLowerCase()) {
-      case 'linux':
-        localConfigPath = '/etc/datadog-agent/application_monitoring.yaml'
-        fleetConfigPath = '/etc/datadog-agent/managed/datadog-agent/stable/application_monitoring.yaml'
-        break
+  #getStableConfigPaths () {
+    // TODO(BridgeAR): Remove these environment variables once we have a proper way to test the stable config.
```
bengl, line 70: As much as we'd like to avoid mocking .... would a filesystem mock suffice here? I think it's sufficiently low level and not mocking dd-trace functionality that it might be appropriate here.

BridgeAR: Yes, that would work. I just did not want to make even more changes 😅

packages/dd-trace/src/ci-visibility/exporters/test-worker/index.js
```diff
@@ -23,13 +23,13 @@ function getInterprocessTraceCode () {
...
   if (getEnvironmentVariable('TINYPOOL_WORKER_ID')) {
     return VITEST_WORKER_TRACE_PAYLOAD_CODE
   }
-  if (getEnvironmentVariable('DD_VITEST_WORKER')) {
+  if (getValueFromEnvSources('DD_VITEST_WORKER')) {
     return VITEST_WORKER_TRACE_PAYLOAD_CODE
   }
   return null
```
simon-id, line 35: I'd like eyes of @juan-fernandez on the test-viz changes, some of the env vars here are DD_ and some are not. Should we really change it to geTValueFromEnvSources ? Is having a mix of `getEnvironementVariable()` and `getValueFromEnvSources()` the behavior we want ?

simon-id: Maybe having this mix of behaviors depending on the env var name in this function will cause unexpected behavior to customers or even to our own devs.

packages/datadog-instrumentations/src/otel-sdk-trace.js
```diff
@@ -2,12 +2,12 @@
...
-const { getEnvironmentVariable } = require('../../dd-trace/src/config-helper')
+const { getValueFromEnvSources } = require('../../dd-trace/src/config-helper')
 const { addHook } = require('./helpers/instrument')
 
-const otelSdkEnabled = getEnvironmentVariable('DD_TRACE_OTEL_ENABLED') ||
-getEnvironmentVariable('OTEL_SDK_DISABLED')
-  ? !getEnvironmentVariable('OTEL_SDK_DISABLED')
+const otelSdkEnabled = getValueFromEnvSources('DD_TRACE_OTEL_ENABLED') ||
+getValueFromEnvSources('OTEL_SDK_DISABLED')
+  ? !getValueFromEnvSources('OTEL_SDK_DISABLED')
```
simon-id, line 10: OTEL_ env vars should be included too ? I thought it was only for DD_ env vars ?

packages/dd-trace/src/ritm.js
```diff
@@ -71,9 +72,10 @@ function Hook (modules, options, onrequire) {
     let name, basedir, hooks
     // return known patched modules immediately
     if (cache[filename]) {
+      const externalCacheEntry = require.cache[filename]
       // require.cache was potentially altered externally
-      if (require.cache[filename] && require.cache[filename].exports !== cache[filename].original) {
-        return require.cache[filename].exports
+      if (externalCacheEntry && externalCacheEntry.exports !== cache[filename].original) {
```
simon-id, line 77:
```suggestion
      if (externalCacheEntry?.exports !== cache[filename].original) {
```

packages/datadog-instrumentations/src/otel-sdk-trace.js
BridgeAR: OTEL_ envs are things we are also documenting as being supported or not.

Now this is an interesting case from two perspectives and I will discuss that internally a bit: should we handle them in stable config? I believe so, because we want our customers to define everything that is relevant for our tracer to run inside of such a file. Or should we exclude those because we only want DD_ envs there?
We could actually consider reading any value from stable config, I have to look into our RFCs for what was defined for that.

packages/dd-trace/src/ci-visibility/exporters/test-worker/index.js
BridgeAR: I am going to check, what our RFCs say about stable config and non DD_ envs. It would only depend on that.

packages/dd-trace/src/ritm.js
BridgeAR: I think style things should be addressed with linter rule activations in a separate PR, while a rule for optional chaining requires type information to work properly (there are a lot of nuances that could break with it).

packages/dd-trace/src/profiling/config.js
```diff
@@ -18,14 +17,16 @@ const SpaceProfiler = require('./profilers/space')
...
 const { tagger } = require('./tagger')
-
 class Config {
   constructor (options = {}) {
+    // TODO: Remove entries that were already resolved in config.
+    // For the others, move them over to config.
+    const AWS_LAMBDA_FUNCTION_NAME = getEnvironmentVariable('AWS_LAMBDA_FUNCTION_NAME')
+    const NODE_OPTIONS = getEnvironmentVariable('NODE_OPTIONS')
+
+    // TODO: Move initialization of these values to packages/dd-trace/src/config.js, and just read from config
```
watson: Non-blocking nit:

```suggestion
    // TODO: Move initialization of these values to packages/dd-trace/src/config/index.js, and just read from config
```

packages/dd-trace/src/profiling/config.js
```diff
@@ -40,40 +41,37 @@ class Config {
...
+      DD_TAGS
+    } = getProfilingEnvValues()
+
     const host = os.hostname()
-    const version = options.version ?? DD_VERSION
     // Must be longer than one minute so pad with five seconds
     const flushInterval = options.interval ?? (Number(DD_PROFILING_UPLOAD_PERIOD) * 1000 || 65 * 1000)
     const uploadTimeout = options.uploadTimeout ?? (Number(DD_PROFILING_UPLOAD_TIMEOUT) || 60 * 1000)
-    const sourceMap = options.sourceMap ?? DD_PROFILING_SOURCE_MAP ?? true
+    const sourceMap = options.sourceMap
```
watson: Non-blocking nit: Just read directly from `options.sourceMap` the one place where the `sourceMap` const is used.

packages/dd-trace/src/profiling/config.js
```diff
@@ -40,40 +41,37 @@ class Config {
...
     const pprofPrefix = options.pprofPrefix ?? DD_PROFILING_PPROF_PREFIX ?? ''
 
-    this.service = service
-    this.env = env
+    this.service = options.service || 'node'
```
watson, line 59: Do we really need this fallback to `node`? I wish we could fall back to the service value from config

packages/dd-trace/src/profiling/profiler.js
```diff
@@ -67,9 +68,57 @@ class Profiler extends EventEmitter {
...
+      commitSHA,
+      injectionEnabled
+    } = config
+    const { enabled, sourceMap, exporters } = config.profiling
+    const { heartbeatInterval } = config.telemetry
+
+    const logger = {
+      debug (message) { log.debug(message) },
+      info (message) { log.info(message) },
+      warn (message) { log.warn(message) },
```
watson, line 93: Not sure I understand the need for a wrapped custom logger to be used within `packages/dd-trace/src/profiling/profiler.js`. I understand we need pass one into `@datadog/pprof` and maybe that module has special requirements, but everywhere in this module, I would assume we could just use the regular logger?

But if it's required, why only support a single argument in `debug`, `info`, and `warn`, when we support multiple in `error`? I see several places in this code where we're using template strings for logging, where we could use printf style logs for more performant logging.

If it's out of scope for this PR to fix this, at least add a TODO.

packages/dd-trace/src/telemetry/telemetry.js
watson: Any update on this?

packages/dd-trace/test/config/helper.spec.js
watson, line 1: This file should be moved to `packages/dd-trace/test/config/helper.spec.js`

packages/dd-trace/src/telemetry/telemetry.js
watson, line 1: This is a quite significant change and it's hard for me to properly review it. As far as I can see, we only have one dedicated test for telemetry: `integration-tests/telemetry.spec.js` - but it doesn't test very much. It would be good with all these changes if we could extend it a bit in this PR.

packages/dd-trace/src/telemetry/telemetry.js
BridgeAR: @bengl and me discussed it and the RFCs gone missing. We are going to activate it and see.

packages/dd-trace/src/startup-log.js
BridgeAR: Acceptable risk

packages/dd-trace/src/exporters/agent/writer.js
BridgeAR: Acceptable risk

packages/dd-trace/src/exporters/agent/writer.js
BridgeAR: Note: go back to lowercase

packages/dd-trace/src/telemetry/telemetry.js
```diff
@@ -159,50 +259,55 @@ function createAppObject (config) {
...
-    return {
-      hostname: os.hostname(),
-      os: osName,
-      architecture: os.arch(),
-      kernel_version: os.version(),
-      kernel_release: os.release(),
-      kernel_name: osName
-    }
+  const base = {
+    hostname: os.hostname(), // TODO is os.hostname() enough?
```
bengl: You can remove this todo

bengl (approved):
LGTM except changes we discussed verbally.

packages/dd-trace/src/profiling/config.js
BridgeAR: I believe that is fine, just out of scope

packages/dd-trace/src/telemetry/telemetry.js
BridgeAR: I went through the changes with @bengl. Mostly it is types. There are a few changes around the host and the error cases and the repeated setTimeout to setInterval. I will work more with telemetry soon and add additional tests there, while this should be safe.

packages/dd-trace/src/ci-visibility/exporters/test-worker/index.js
BridgeAR: So, these envs are all environment defined that are in test optimization. These should never be user defined. As such, we should never read them from anywhere else than from process.env.
That applies to AFAIK all non DD_ envs we access.

PR #6991, opened 2025-11-25
fix(aws-sdk): do not patch client config deserialization multiple times

### LINKED ISSUE #6985: [BUG]: Infinite recursion in `aws-sdk` instrumentation on dd-trace 5.80.0
### Tracer Version(s)

5.80.0

### Node.js Version(s)

24.11.1

### Bug Report

Hey,

After upgrading from dd-trace 5.79.0 to **5.80.0**, we are getting a crash caused by **infinite recursion** inside the AWS SDK v3 instrumentation. The stack repeatedly shows the wrapper calling itself:

```
RangeError: Maximum call stack size exceeded
at AwsJson1_0Protocol.deserializeResponse (.../datadog-instrumentations/src/aws-sdk.js:46:24)
at AwsJson1_0Protocol.deserializeResponse (... same ...)
...
```

When we disable the integration:

```
DD_TRACE_DISABLED_PLUGINS=aws-sdk
```

the issue disappears. All SQS calls work normally.

It looks like the change introduced https://github.com/DataDog/dd-trace-js/pull/6911/files is causing recursion.

### Environment

* Node 24.11.1
* dd-trace 5.80.0
* @aws-sdk/client-sqs 3.936

Thanks!

sabrenner:
### What does this PR do?
Makes sure we do not patch `client.config.protocol.deserializeResponse` multiple times, as `shimmer.wrap` does not check if the target object's method has already been patched.

Perhaps we should patch the base protocol's prototype, but looking at the [source code](https://github.com/smithy-lang/smithy-typescript/blob/4313e2f5dbd759d0f5a3ff355d45d64e33d4583d/packages/core/src/submodules/protocols/HttpProtocol.ts), it seems all of the base classes are either abstract or just type interfaces, and/or the `deserializeResponse` method is also abstract. But, maybe we can just wrap the one that `BedrockRuntime` uses, as it's the one this change was introduced for.

### Motivation
Fixes #6985 

### Testing
I created an integration test in order to set a super small max call stack size, and was able to reproduce the issue from the linked issue locally:

```
RangeError: Maximum call stack size exceeded
  Deserialization error: to see the raw response, inspect the hidden field {error}.$response on this object.
    at deref (/private/var/folders/94/lnn2crkx7914c2835zmwrwvh0000gq/T/7e02ad1fc7294706/1cc97b417deae3ae/node_modules/@smithy/core/dist-cjs/submodules/schema/index.js:8:16)
    at NormalizedSchema.of (/private/var/folders/94/lnn2crkx7914c2835zmwrwvh0000gq/T/7e02ad1fc7294706/1cc97b417deae3ae/node_modules/@smithy/core/dist-cjs/submodules/schema/index.js:307:20)
    at AwsJson1_0Protocol.deserializeResponse (/private/var/folders/94/lnn2crkx7914c2835zmwrwvh0000gq/T/7e02ad1fc7294706/1cc97b417deae3ae/node_modules/@smithy/core/dist-cjs/submodules/protocols/index.js:466:44)
    at AwsJson1_0Protocol.deserializeResponse (/private/var/folders/94/lnn2crkx7914c2835zmwrwvh0000gq/T/7e02ad1fc7294706/1cc97b417deae3ae/node_modules/dd-trace/packages/datadog-instrumentations/src/aws-sdk.js:46:24)
    at AwsJson1_0Protocol.deserializeResponse (/private/var/folders/94/lnn2crkx7914c2835zmwrwvh0000gq/T/7e02ad1fc7294706/1cc97b417deae3ae/node_modules/dd-trace/packages/datadog-instrumentations/src/aws-sdk.js:46:24)
    at AwsJson1_0Protocol.deserializeResponse (/private/var/folders/94/lnn2crkx7914c2835zmwrwvh0000gq/T/7e02ad1fc7294706/1cc97b417deae3ae/node_modules/dd-trace/packages/datadog-instrumentations/src/aws-sdk.js:46:24)
    at AwsJson1_0Protocol.deserializeResponse (/private/var/folders/94/lnn2crkx7914c2835zmwrwvh0000gq/T/7e02ad1fc7294706/1cc97b417deae3ae/node_modules/dd-trace/packages/datadog-instrumentations/src/aws-sdk.js:46:24)
    at AwsJson1_0Protocol.deserializeResponse (/private/var/folders/94/lnn2crkx7914c2835zmwrwvh0000gq/T/7e02ad1fc7294706/1cc97b417deae3ae/node_modules/dd-trace/packages/datadog-instrumentations/src/aws-sdk.js:46:24)
    at AwsJson1_0Protocol.deserializeResponse (/private/var/folders/94/lnn2crkx7914c2835zmwrwvh0000gq/T/7e02ad1fc7294706/1cc97b417deae3ae/node_modules/dd-trace/packages/datadog-instrumentations/src/aws-sdk.js:46:24)
    at AwsJson1_0Protocol.deserializeResponse (/private/var/folders/94/lnn2crkx7914c2835zmwrwvh0000gq/T/7e02ad1fc7294706/1cc97b417deae3ae/node_modules/dd-trace/packages/datadog-instrumentations/src/aws-sdk.js:46:24) {
  '$metadata': {
    httpStatusCode: 200,
    requestId: 'a769fd06-6640-4205-9c3f-2561a886b8a9',
    extendedRequestId: undefined,
    cfId: undefined,
    attempts: 1,
    totalRetryDelay: 0
  }
}

Node.js v22.17.0
```

### DIFF
--- a/packages/datadog-plugin-aws-sdk/test/integration-test/sqs.spec.js
+++ b/packages/datadog-plugin-aws-sdk/test/integration-test/sqs.spec.js
@@ -11,7 +11,7 @@ const { withVersions } = require('../../../dd-trace/test/setup/mocha')
 
 const assert = require('node:assert')
 
-describe.only('recursion regression test', () => {
+describe('recursion regression test', () => {
   let agent
   let proc
 


packages/datadog-instrumentations/src/aws-sdk.js
```diff
@@ -67,12 +67,17 @@ function wrapSmithySend (send) {
...
+      this.config?.protocol?.deserializeResponse &&
+      !Object.hasOwn(this.config.protocol, Symbol.for('dd.wrapped'))
+    ) {
       shimmer.wrap(
         this.config.protocol,
         'deserializeResponse',
         deserializeResponse => wrapDeserialize(deserializeResponse, channelSuffix, 2)
       )
+
+      Object.defineProperty(this.config.protocol, Symbol.for('dd.wrapped'), { value: true })
```
BridgeAR: This will work fine, if the objects are not frozen. I personally would just use a WeakSet instead. I also wonder about the method being called multiple times.

BridgeAR (commented):
Do you know why this method would be wrapped multiple times while the main part would not cause the same problem? I am confused about that part.

packages/datadog-instrumentations/src/aws-sdk.js
sabrenner: It's being called multiple times because this wrapping is happening per `client.send(command)` call, which can happen many times over a service's lifespan. So we in effect end up with the function calling its own wrapper, not through infinite recursion so-to-say, as in theory the stack call would have a fixed size if unbounded, it's just that the stack size grows with the number of `client.send` calls made. I had thought `shimmer.wrap` did some checks for if something was already wrapped, but as it doesn't, I went with this approach. Let me know if that makes sense. 

I do think I'll try and convert to a WeakSet though 😄

PR #6992, opened 2025-11-25
Fix a bug with tryAddMessage being changed to async which causes messages to be dropped

dwilson6:
### What does this PR do?

Fixes a bug in the azure-service-bus instrumentation that breaks the createMessageBatch tryAddMessage method by changing it from a sync method that returns a boolean to an async method returning a promise.

### Motivation

Ran into this bug which was causing some code to drop service bus messages when building batches due to returning a promise which is truthy even when the resolved promise would return false but the consuming code is synchronous.

### Plugin Checklist

- [ ] Unit tests.
- [x] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


jcstorms1:
Hey @dwilson6 thank you for letting us know about this and for all the work you did. We require commits to be signed so I needed to remake this PR. The new PR is [here](https://github.com/DataDog/dd-trace-js/pull/7098) and I'm going to close this one. Thanks again!

PR #6999, opened 2025-11-26
test: rewrite assertion to use Node.js assert

BridgeAR:
Please check the individual commis for details.

The first two are a script that is mostly auto generated to transform the assertions.

The third commit is the rewrite with that script.
The fourth commit is the auto fixes by eslint.
The fifth commit includes all manual changes on top. I had quite a few of these, since the script used a few rewrites in ways that were not correct.

I suggest to take a glimpse at the script, than to have a glimpse at the auto changed things individually per commit and to have a closer look at all manual changes. Yes, they are way more than I wanted, while we still have more assertions to change and more chai assertions are added as long as we do not change these entries soon.


BridgeAR:
@watson I just added the PR description and updated the code.

@simon-id as @watson outlined, it is mostly auto generated, while this contains quite a lot of manual changes as well.
I added suggestions how to review this best in the PR description.

integration-tests/cucumber/cucumber.spec.js
```diff
@@ -366,19 +373,21 @@ describe(`cucumber@${version} commonJS`, () => {
                 assert.ok(metrics[DD_HOST_CPU_COUNT] != null)
               })
 
-              assert.includeMembers(testEvents.map(test => test.content.resource), [
+              assert.deepStrictEqual(testEvents.map(test => test.content.resource).sort(), [
```
pabloerhard, line 376: Is there an equivalent of includeMembers which does not involve having to use deepStrictEqual (saw in some place that asserObjectContains is used)? Just curious what would happen if a bigger array needed to be asserted.

pabloerhard (approved):
Went quickly over the script and looked into all the manual changes (along some of the full files), LGTM.

integration-tests/cucumber/cucumber.spec.js
BridgeAR: `assertObjectContains()` stands for `assert.partialDeepStrictEqual()` and is the equivalent, while it is a tad stricter (the order matters and duplicates are not ignored).

So if you want to match a subset, that is perfect.

pabloerhard (dismissed):

Issue #1750, opened 2022-01-12
NextJS page based resource.name


Meemaw:
It would be great if `NextJS` did per page based `resource.name` tagging (or allow customising of it).

We currently have to hack this by overriding it ourselves doing something like this in our middleware:
```js
const root = req._datadog.span
// Set resource name to path pattern
root.setTag("resource.name", route.pattern)
```

This allows us to get a very powerful grouping per resource/page that would look something like this:

<img width="1507" alt="Screenshot 2022-01-12 at 12 02 38" src="https://user-images.githubusercontent.com/8524109/149128429-10fbf216-c650-4486-9ea2-788981db795d.png">

rochdev:
This should already be the case. Are you looking at `http.request` or `next.request` on the service page? If `next.request` is selected you should see the routes for Next specifically as they don't bubble up to the root span. You can also set it as the primary operation if you're mostly interested in Next metadata.

Meemaw:
<img width="958" alt="Screenshot 2022-01-14 at 16 41 38" src="https://user-images.githubusercontent.com/8524109/149543026-97b04a6f-0ccb-40b2-8fa6-ecc84aac452b.png">

Seems the `resource.name` for `next.request` is just `GET`, which kinda correlated to this code: https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-plugin-next/src/index.js#L101

rochdev:
This is just the default, and the resource name is later updated according to the page. It's possible it's not working for you case as NextJS has a complex router internally, so it's possible there are cases not handled properly. Can you provide a reproduction snippet that I could try?

rochdev:
@Meemaw Are you still having this issue?

Meemaw:
Yes

rochdev:
One thing I'm wondering is if the tracer is initialized early enough, as load order issues can cause some of the metadata to be missing. Generally speaking the fact that the span is named `next.request` should indicate that it's loaded early enough, but for Next specifically it can be complicated to load first since there is no way to do that in code, so maybe something to look into. The only way for NextJS to load the tracer before all other imports is to use `NODE_OPTIONS='-r dd-trace/init'`. If this is already how you are initializing the tracer then there is definitely a bug and we would need a reproduction. If the tracer is initialized programmatically then I would try switching to this approach instead which should fix the issue.

Eronarn:
[...]

PR #7000, opened 2025-11-26
Fix: Update resource name when child span is active in store for Next.js

crysmags:
### What does this PR do?
When OpenTelemetry TracerProvider was enabled, Next.js routes showed only HTTP method as resource name ("GET") instead of full path ("GET /api/hello").

### Motivation
This is related to https://github.com/DataDog/dd-trace-js/issues/1750 and escalation https://datadoghq.atlassian.net/browse/APMS-17880. 

Root Cause
The Next.js plugin builds resource names in two stages:
`bindStart()` - Creates HTTP span with method-only resource name
`pageLoad()` - Updates with full path
The plugin stored requests in a WeakMap using span objects as keys: `this._requests.set(span, req)`.
When TracerProvider enabled, OTel creates child spans (e.g., "resolve page components") that become active in async context between `bindStart` and `pageLoad`. The `pageLoad` method would then retrieve a different span object (the child) from storage, causing WeakMap lookup to fail since it requires an exact object identity (===). 
This is a change from an object-based WeakMap to ID-based Map with parent span fallback in order to handle the case where a child span is created and the context storage has changed. There is a cleanup in `finish` that will prevent memory leaks since we are now using a Map.

### DIFF
--- a/packages/datadog-plugin-next/src/index.js
+++ b/packages/datadog-plugin-next/src/index.js
@@ -10,10 +10,10 @@ const errorPages = new Set(['/404', '/500', '/_error', '/_not-found', '/_not-fou
 
 class NextPlugin extends ServerPlugin {
   static id = 'next'
+  #requestsBySpanId = new WeakMap()
 
   constructor (...args) {
     super(...args)
-    this._requestsBySpanId = new Map()
     this.addSub('apm:next:page:load', message => this.pageLoad(message))
   }
 
@@ -36,8 +36,8 @@ class NextPlugin extends ServerPlugin {
     analyticsSampler.sample(span, this.config.measured, true)
 
     // Store request by span ID to handle cases where child spans are activated
-    const spanId = span.context().toSpanId()
-    this._requestsBySpanId.set(spanId, req)
+    const spanId = span.context()._spanId
+    this.#requestsBySpanId.set(spanId, req)
 
     return { ...store, span }
   }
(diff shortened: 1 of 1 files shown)


packages/datadog-plugin-next/src/index.js
IlyasShabi, line 1: Can you add a test for this please?

crysmags: @IlyasShabi added a test that checks if the request object is still returned when handling child spans🫡

packages/datadog-plugin-next/src/index.js
```diff
@@ -13,7 +13,7 @@ class NextPlugin extends ServerPlugin {
 
   constructor (...args) {
     super(...args)
-    this._requests = new WeakMap()
+    this._requestsBySpanId = new Map()
```
BridgeAR: Please make this a private property instead of using the underscore :)

packages/datadog-plugin-next/src/index.js
```diff
@@ -89,7 +96,22 @@ class NextPlugin extends ServerPlugin {
...
     const span = store.span
-    const req = this._requests.get(span)
+
+    const spanId = span.context().toSpanId()
+
+    // Convert parent ID from hex to decimal to match storage format.
+    let parentSpanId = null
+    if (span.context()._parentId) {
+      parentSpanId = span.context()._parentId.toString(10)
+    }
```
BridgeAR:
```suggestion
    const parentSpanId = span.context()._parentId?.toString(10)
```

packages/datadog-plugin-next/src/index.js
```diff
@@ -89,7 +96,22 @@ class NextPlugin extends ServerPlugin {
...
+
+    // Try current span first, then parent span.
+    // This handles cases where pageLoad runs in a child span context
+    // (e.g., when OpenTelemetry TracerProvider creates "resolve page components" span)
+    let req = this._requestsBySpanId.get(spanId)
+    if (!req && parentSpanId) {
+      req = this._requestsBySpanId.get(parentSpanId)
+    }
```
BridgeAR:
```suggestion
    const req = this._requestsBySpanId.get(spanId) ?? this._requestsBySpanId.get(parentSpanId)
```

PR #7002, opened 2025-11-27
feat(electron): add initial support

rochdev:
### What does this PR do?

Add initial support for Electron.

- Fixes various incompatibilities like the missing `unref()` function.
- Add `net` integration to support `fetch` and `request`.
- Add `ipc` integrations to support `ipcMain` and `ipcRenderer` communication.
    -  Only works for renderer to main in non-sandboxed mode.
- Exporter to a diagnostics channel for the Electron SDK to subscribe to.

### Motivation

APM support for the [Electron SDK](https://github.com/DataDog/electron-sdk).

### Plugin Checklist

- [x] Unit tests.
- [x] Integration tests.
- [ ] Benchmarks.
- [x] TypeScript [tests][2].
- [x] API [documentation][3].
- [x] CI [jobs/workflows][4].

### Additional Notes

Integration tests were written using Claude.


packages/dd-trace/src/ritm.js
```diff
@@ -31,6 +31,7 @@ const builtinModules = new Set(Module.builtinModules.map(stripNodePrefix))
 
 function isBuiltinModuleName (name) {
```
tlhunter, line 32: Does this function name still make sense?

packages/dd-trace/src/dogstatsd.js
```diff
@@ -158,7 +158,7 @@ class DogStatsDClient {
     const socket = dgram.createSocket(type)
 
     socket.on('error', () => {})
-    socket.unref()
+    socket.unref?.()
```
tlhunter, line 164:
```suggestion
    socket.unref?.() // Electron doesn't support dogstatsd
```

packages/dd-trace/src/dogstatsd.js
```diff
@@ -361,7 +361,7 @@ class CustomMetrics {
     const flush = this.flush.bind(this)
 
     // TODO(bengl) this magic number should be configurable
-    setInterval(flush, 10 * 1000).unref()
+    setInterval(flush, 10 * 1000).unref?.()
```
tlhunter, line 364:
```suggestion
    setInterval(flush, 10 * 1000).unref?.() // Electron doesn't support dogstatsd
```

packages/dd-trace/src/telemetry/endpoints.js
```diff
@@ -63,7 +63,7 @@ function scheduleFlush () {
   flushScheduled = true
   // this used to be setImmediate() instead, but it was making the system test flaky
   // don't ask me why
-  setTimeout(flushAndSend).unref()
+  setTimeout(flushAndSend).unref?.()
```
tlhunter, line 66: All of the `?.()` changes are a little scary. Is there a change it would accidentally be made undefined outside of Electron and that we wouldn't catch it? I suppose we could throw guard `if` statements around all of them but that would get messy. Maybe instantiate a noop class in the case of Electron?

```suggestion
  setTimeout(flushAndSend).unref?.() // Electron doesn't support telemetry
```

.github/workflows/apm-integrations.yml
```diff
@@ -454,6 +454,22 @@ jobs:
...
+  electron:
+    runs-on: ubuntu-latest
+    env:
+      PLUGINS: electron
+    steps:
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+      - uses: ./.github/actions/testagent/start
+      - uses: ./.github/actions/node/latest
+      - uses: ./.github/actions/install
+      - run: sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
```
tlhunter, line 455: This deserves a comment; doesn't look like we use it elsewhere.

.github/workflows/apm-integrations.yml
```diff
@@ -454,6 +454,22 @@ jobs:
...
+      - run: |
+          sudo apt-get update
+          sudo apt-get install -y xvfb
```
tlhunter, line 460:
```suggestion
          # run an in memory X Windows Framebuffer to render Electron
          sudo apt-get install -y xvfb
```

integration-tests/electron/app/main.js
```diff
@@ -0,0 +1,69 @@
...
+require('dd-trace').init({
+  service: 'electron-integration-test',
+  flushInterval: 0,
+  plugins: false,
+  experimental: { exporter: 'electron' },
+}).use('electron', true)
+
+// Forward spans exported via the electron exporter to the test process via IPC.
+channel('datadog:apm:electron:export').subscribe(traces => {
+  process.send({ name: 'traces', payload: traces })
```
tlhunter, line 16: Is `name: 'traces'` too generic here? Maybe `datadog-traces` instead?

tlhunter (changes requested):
I mostly have concerns about all of the method nullish checks since they might make real issues outside of an Electron context slip through. Can you instead implement stub classes and drop all of the `?.()` entries?

packages/dd-trace/src/dogstatsd.js
rochdev: Why this comment in this spot specifically?

rochdev: What Electron doesn't support in `unref`.

packages/dd-trace/src/telemetry/endpoints.js
rochdev: It's because Electron doesn't have `unref` on timers, presumably because it's an app so it doesn't exit without something explicit even if idle.

rochdev: Let me see if I can add a lint rule for this.

packages/dd-trace/src/ritm.js
rochdev: It does, but it used to only check for things specific to Node. Electron also has its own built-in modules.

integration-tests/electron/app/main.js
rochdev: Wouldn't that be redundant if we're in dd-trace?

packages/dd-trace/src/dogstatsd.js
rochdev: Added eslint rule with explanation about Electron on violation and auto-fix.

packages/dd-trace/src/dogstatsd.js
rochdev: Added eslint rule with explanation about Electron on violation and auto-fix.

packages/dd-trace/src/telemetry/endpoints.js
rochdev: Added eslint rule with explanation about Electron on violation and auto-fix.

.github/workflows/apm-integrations.yml
rochdev: Added comments for all these commands.

.github/workflows/apm-integrations.yml
rochdev: Added comments for all these commands.

packages/dd-trace/src/dogstatsd.js
tlhunter: I must've misunderstood what the intent of these changes is. Why was this particular change made?

packages/dd-trace/src/telemetry/endpoints.js
tlhunter: Okay, I understand now. The Electron `Timeout` instance doesn't have an `unref()` method. Which makes sense since browsers don't have it either. Making it optional makes total sense where we create the timeout and immediately call unref, like in the case of this line I flagged.

In Node.js it's used to prevent the timeout from keeping the process alive which is very important for third party code in an application like the tracer. Conceptually it makes sense that Electron wouldn't have it since closing the app is a very intentional act.

Overall though I'm worried that we might allow a bug to get created in the tracer with this change made everywhere. Consider the following:

```js
let timer

function doStuff() {
  timer = setTimeout(() => {}, 1000);
}

function cleanup() {
  timer.unref();
  // versus
  timer.unref?.();
}
```

If we forget to call `doStuff()` before `cleanup()` in a Node.js process as of today we'll get an error that we can't call `.unref()` of undefined. However with the `.unref?.()` change we now might let it slip through the cracks.

Is that a valid concern or am I overreacting?

tlhunter: Actually, upon reviewing every example in your PR, we're always (except maybe one debugger situation) calling unref immediately after creating the timer so this concern is invalid.

We do inconsistently do the following though which made it hard to understand at first:

```js
// we mostly do this
const timer = setTimeout().unref();

// we sometimes do this
const timer = setTimeout();
timer.unref();
```

(Of course, `.unref()` is method chained to return the timer instance.)

tlhunter: Actually, looking at my last code example, I believe these code changes need another tweak. I think we need to replace all instances of `const timer = setTimeout().unref?.();` with `const timer = setTimeout(); timer.unref?.();`.

If you call `const timer = setTimeout().unref?.();` in Electron then `timer` will be `undefined`.

packages/datadog-plugin-openai/src/services.js
```diff
@@ -36,7 +36,7 @@ module.exports.init = function (tracerConfig) {
 
   interval = setInterval(() => {
     metrics.flush()
-  }, FLUSH_INTERVAL).unref()
+  }, FLUSH_INTERVAL).unref?.()
```
tlhunter: Here's an example of how we might need to make the code Node.js and Electron compatible:

```diff
-  interval = setInterval(() => {
-    metrics.flush()
-  }, FLUSH_INTERVAL).unref?.()
+  interval = setInterval(() => {
+    metrics.flush()
+  }, FLUSH_INTERVAL)
+  interval.unref?.()
```

packages/dd-trace/src/telemetry/endpoints.js
rochdev: Fixed, I made `unref` always return null in tests and updated all usage to be on a separate line.

tlhunter (approved):

PR #7006, opened 2025-11-27
feat(debugger): implement intelligent snapshot pruning for oversized payloads

watson:
### What does this PR do?

Implements an intelligent snapshot pruning algorithm for Dynamic Instrumentation/Live Debugger that selectively removes the largest and deepest leaf nodes from oversized payloads while preserving the overall schema structure. This replaces the previous approach which simply deleted all captured variables when a snapshot exceeded the 1MB size limit.

### Motivation

Align with how the other tracers prunes large snapshots.

Previously, when a snapshot payload exceeded 1MB, we would delete all captures entirely and show an error message to users. This was a poor user experience because users would lose all captured variable data, even though most of the snapshot was likely still valuable.

With this new pruning algorithm, we can intelligently reduce the size of oversized snapshots by removing only the least valuable data (deepest, largest nodes, and nodes that were already truncated due to depth limits), allowing users to still see most of their captured variables even when dealing with large data structures.

### Additional Notes

The pruning algorithm:
- Parses snapshots into a tree structure tracking JSON object positions
- Uses a priority queue to select nodes for pruning based on:
  1. Presence of `notCapturedReason: 'depth'` (highest priority - already truncated data)
  2. Depth level (deeper nodes pruned first)
  3. Presence of any `notCapturedReason` (any truncated data)
  4. Size (larger nodes pruned first)
- Only prunes nodes at level 6 or deeper (`locals` and below)
- Promotes parent nodes when all children are pruned to reduce overhead
- Iteratively prunes if needed to reach target size
- Falls back to prune all everything if pruning fails or errors out

### DIFF
--- a/packages/dd-trace/src/debugger/devtools_client/snapshot-pruner.js
+++ b/packages/dd-trace/src/debugger/devtools_client/snapshot-pruner.js
@@ -63,29 +63,17 @@ class TreeNode {
   }
 
   /**
-   * Priority key for sorting in queue (higher values = higher priority for pruning). Checks ancestors for
-   * `notCapturedReason` flags to prioritize children of partially captured objects (where the flag is on the parent).
-   * Cached to avoid repeated ancestor chain walks during heap operations.
+   * Priority key for sorting in queue (higher values = higher priority for pruning).
+   * Cached to avoid repeated computation during heap operations.
    *
    * @returns {[number, number, number, number]} Priority key tuple: [not_captured_depth, level, not_captured, size]
    */
   get priorityKey () {
     if (this.#priorityKeyCache === null) {
-      let hasNotCapturedDepth = this.notCapturedDepth
-      let hasNotCaptured = this.notCaptured
-
-      // Check ancestors for notCapturedReason flags
-      let ancestor = this.parent
-      while (ancestor) {
-        if (ancestor.notCapturedDepth) hasNotCaptured = hasNotCapturedDepth = true
-        else if (ancestor.notCaptured) hasNotCaptured = true
-        ancestor = ancestor.parent
-      }
-
       this.#priorityKeyCache = [
-        hasNotCapturedDepth ? 1 : 0,
+        this.notCapturedDepth ? 1 : 0,
         this.level,
-        hasNotCaptured ? 1 : 0,
+        this.notCaptured ? 1 : 0,
         this.size
       ]
     }
--- a/packages/dd-trace/test/debugger/devtools_client/snapshot-pruner.spec.js
+++ b/packages/dd-trace/test/debugger/devtools_client/snapshot-pruner.spec.js
@@ -73,7 +73,7 @@ describe('snapshot-pruner', function () {
 
       const result = pruneSnapshot(json, size, maxSize)
 
-      assert.notStrictEqual(result, undefined, 'Expected pruneSnapshot() to successfully prune')
+      assert.ok(result, 'Expected pruneSnapshot() to successfully prune')
       assert.strictEqual(result, json)
     })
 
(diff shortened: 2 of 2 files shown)


watson:
* **#7006** <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-js/7006?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-js/7006?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/dd-trace/src/debugger/devtools_client/snapshot-pruner.js
```diff
@@ -0,0 +1,473 @@
...
+  return index
+}
+
+/**
+ * Find notCapturedReason value in a JSON object string.
+ *
+ * @param {TreeNode} node - The node to search in
+ * @returns {string|undefined} The reason value or undefined if not found
+ */
+function findNotCapturedReason (node) {
```
shatzi: Might be simpler to put this logic at parse json logic. Everytime you skip a string you check if was noCapturrReason (maybe check length first to really speed this check). And the check if the next string is "depth".

This would avoid checking if we are inside children ranges logic

packages/dd-trace/src/debugger/devtools_client/snapshot-pruner.js
```diff
@@ -0,0 +1,473 @@
...
+   *
+   * @returns {[number, number, number, number]} Priority key tuple: [not_captured_depth, level, not_captured, size]
+   */
+  get priorityKey () {
+    let hasNotCapturedDepth = this.notCapturedDepth
+    let hasNotCaptured = this.notCaptured
+
+    // Check ancestors for notCapturedReason flags
+    let ancestor = this.parent
+    while (ancestor) {
```
shatzi: can we avoid this look up? this seems to be very heavy as we keep comparing the heap again and agin.

watson: Fixed

packages/dd-trace/src/debugger/devtools_client/snapshot-pruner.js
watson: Fixed

packages/dd-trace/src/debugger/devtools_client/snapshot-pruner.js
```diff
@@ -0,0 +1,416 @@
...
+   * @returns {[number, number, number, number]} Priority key tuple: [not_captured_depth, level, not_captured, size]
+   */
+  get priorityKey () {
+    if (this.#priorityKeyCache === null) {
+      let hasNotCapturedDepth = this.notCapturedDepth
+      let hasNotCaptured = this.notCaptured
+
+      // Check ancestors for notCapturedReason flags
+      let ancestor = this.parent
+      while (ancestor) {
```
shatzi: if we calculate this key when the get inserted, none of his ancestor would have notCaptureDepth. I would also say most of them will notCaptured.

So why do we need this logic in the first place?

watson: You're right about this in case a parent is promoted, but there's a different use-case we also need to consider: Children of partially-captured objects from the original JSON.

When the debugger captures a snapshot, objects can have `notCapturedReason` set on the parent (e.g., `"collectionSize"`, `"fieldCount"`). During parsing, these flags are set on the parent node, but the children are regular leaf nodes without flags.

Example:

```js
{
  "truncatedList": {
    "type": "array",
    "notCapturedReason": "collectionSize", // Flag on parent
    "elements": [
      { "type": "string", "value": "xxx..." }, // Child leaf (no flag)
      { "type": "string", "value": "yyy..." }  // Child leaf (no flag)
    ]
  }
}
```

When these children calculate their `priorityKey`:
1. They don't have `notCaptured` flags themselves
2. They walk up and find the flag on their parent
3. This gives them higher priority for pruning

Without the ancestor walk, children of truncated collections have the same priority as children of normal collections, causing the wrong nodes to be pruned.

watson: Discussed on Slack: Turned out the RFC doesn't expect children of parents with a flag to have higher priority. I've removed that logic from the implementation now

PR #7007, opened 2025-11-27
chore: add no-unused-expressions lint rule

simon-id:
### What does this PR do?
Add the eslint rule `no-unused-expressions`, it prevents stuff like:
```javascript
function a () {
  if (x) false // forgot the return statement

  blabla()
}
```
It also allows the use of the `void` operator, to explicitly allow statements with only side effects.

### Motivation
Mistakes have been made in previous PRs that could have been prevented with this rule.

### DIFF
--- a/packages/datadog-plugin-mariadb/test/index.spec.js
+++ b/packages/datadog-plugin-mariadb/test/index.spec.js
@@ -261,7 +261,7 @@ describe('Plugin', () => {
           })
 
           it('should run event listeners in the parent context', done => {
-            if (typeof connection.queryStream !== 'function') return done()
+            if (typeof connection.queryStream !== 'function') return this.skip()
 
             const stream = connection.queryStream('SELECT 1 + 1 AS solution')
 


packages/datadog-instrumentations/src/mariadb.js
```diff
@@ -66,7 +66,7 @@ function createWrapQuery (options) {
           finishCh.publish(ctx)
           return result
         }, error => {
-          ctx.error
+          ctx.error = error
```
simon-id, line 68: @rochdev need your approval on this, the change i made is only a guess

eslint.config.mjs
```diff
@@ -420,7 +420,9 @@ export default [
       'no-await-in-loop': 'error',
       'no-else-return': ['error', { allowElseIf: true }],
       'no-implicit-coercion': ['error', { boolean: true, number: true, string: true, allow: ['!!'] }],
+      'no-unused-expressions': 'error',
       'no-useless-assignment': 'error',
+      'no-void': 'off',
```
BridgeAR: I think it is better to use the eslint exception for accessing the stack, while I am uncertain if accessing it is still needed at all. There were some V8 changes that might make that obsolete (I did not check).

packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -117,7 +117,7 @@ function extractTraceContext (tracer, ctx) {
     case 'Http':
       return tracer.extract('http_headers', Object.fromEntries(ctx.httpRequest.headers))
     default:
-      null
+      return null
```
BridgeAR: AFAIK it's even possible to just remove the default which would allow this to be a simple if statement. While being on this: the `String()` call is not needed.

packages/datadog-instrumentations/src/mariadb.js
BridgeAR: This looks like a bug fix. Could you please add a test for that?

simon-id: yes, but i'll let roch do this, because i don't even know what bug this fixes

eslint.config.mjs
simon-id: any reasoning against void ? just picked that because i never remember the eslint exception syntax, void is easy to remember

packages/datadog-plugin-azure-functions/src/index.js
simon-id: Of course the default here is not needed if the caller accept undefined instead of null, but i'm trying to keep the changes as iso as possible, this is just a linter PR, if i stop at every little thing that can be improved, the PR will be very off topic

simon-id: I thought it was unnecessary when i saw it too, but if i go and change it, i run the risk of creating more problems and failing tests and stuff, that i'm not interested in investing time on now. Especially since I'm making this PR completely online without running it locally, so iteration time is way slower. Small PR for a small task

packages/dd-trace/src/remote_config/scheduler.js
```diff
@@ -17,7 +17,7 @@ class Scheduler {
   runAfterDelay (interval = this._interval) {
     this._timer = setTimeout(this._callback, interval, () => this.runAfterDelay())
 
-    this._timer.unref && this._timer.unref()
+    this._timer.unref?.()
```
watson: This seems like a mistake - The `unref` function was added in Node.js v0.9.1, so why can't we just:

```suggestion
    this._timer.unref()
```

packages/datadog-plugin-azure-functions/src/index.js
watson: Since it didn't return `null` before, but `undefined`, I'd argue that there's a bigger risk making this change vs just removing the entire `default` statement. If you're unsure, look at how the caller handles the return value.

packages/dd-trace/src/remote_config/scheduler.js
simon-id: probably very old copy pasted code, good point!

packages/datadog-plugin-azure-functions/src/index.js
simon-id: Very good point, you convinced me, I'll make it a if and ping whoever wrote this

packages/datadog-instrumentations/src/mariadb.js
BridgeAR: Well, that is to figure out in that case, no?

Because it would either not be needed, or it would fix a bug and we should add a regression test.
Not adding the test is not good in my opinion.

eslint.config.mjs
BridgeAR: I don't think `void` is ergonomic as it is unclear why that is needed reading the code. Most developers would also not know what it does (in general), while `// eslint-disable-next-line foobar` is commonly used and we need that anyway in a few spots.

packages/datadog-plugin-azure-functions/src/index.js
BridgeAR: Generally, it is good to check where it is used and in what context. The call site uses `null` next to it working with `undefined` in that case.
Interestingly, it does have a different behavior using `null` or `undefined` looking at the code that uses it. It does seem `undefined` would be expected in all cases, as far as I can tell by briefly looking at it. The behavior in that case is that the async local storage store is checked if there is no child and that span is then defined as child span in case it exists. The type definition for startSpan childOf option officially only supports `undefined`, so that is another indicator for that.

Could you please add a TODO entry for the `null` part and ask someone who maintains this code to look into the `null` case? I could imagine it is a bug.

packages/datadog-instrumentations/src/mariadb.js
simon-id: Yeah of course I agree, I'd like a test for this in this PR too. I'll try to steal some of Roch's time for this.

eslint.config.mjs
simon-id: Ok sounds reasonable, i'll change that

packages/datadog-plugin-azure-functions/src/index.js
BridgeAR: Discussed with @rochdev that null is the expected part. I added JSDoc to clarify that.

eslint.config.mjs
BridgeAR: I just changed to rule to only allow usages as you did. That way it should be fine.

packages/datadog-plugin-azure-functions/src/index.js
BridgeAR: @jcstorms1 PTAL. We believe returning `null` is actually a fix (using the span as root instead of attaching it to a potential parent)

packages/datadog-plugin-azure-functions/test/integration-test/eventhubs-test/eventhubs.spec.js
simon-id, line 1: what are the changes in these 3 files doing in this PR ?

simon-id: also the changes in apollo.js and tracing.js are unrelated ?

packages/datadog-plugin-azure-functions/src/index.js
jcstorms1: @BridgeAR Yes, this should be returning `null` when the function is not triggered by an `HTTP` trigger. I must have missed this in my review.

packages/datadog-plugin-mariadb/test/index.spec.js
```diff
@@ -201,16 +201,212 @@ describe('Plugin', () => {
...
+          })
+
+          it('should run promise continuations in the parent context', async () => {
+            await connection.query('SELECT 1 + 1 AS solution').then(() => {
+              assert.strictEqual(tracer.scope().active(), null)
+            })
+          })
+
+          it('should run event listeners in the parent context', done => {
+            if (typeof connection.queryStream !== 'function') return done()
```
simon-id: I prefere using skips to avoid any confusion. We do this in other tests, does it work here ?
```suggestion
            if (typeof connection.queryStream !== 'function') return this.skip()
```

PR #7010, opened 2025-11-28
feat(anthropic): add anthropic beta client support

PROFeNoM:
https://datadoghq.atlassian.net/browse/MLOB-4728

### What does this PR do?

Implement beta message handling for Anthropic SDK and add corresponding test cases

- Added a new hook for the beta version of the Anthropic SDK to support message creation.
- Introduced test cases for the beta messages.create and messages.stream methods, ensuring proper span creation and handling of streaming responses.
- Created two new YAML test cassettes to validate interactions with the beta API.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
+++ b/packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
@@ -3,7 +3,7 @@
 const { describe, before, it } = require('mocha')
 const { withVersions } = require('../../../setup/mocha')
 const assert = require('node:assert')
-const semver = require('semver')
+const semifies = require('semifies')
 const { useEnv } = require('../../../../../../integration-tests/helpers')
 
 const {
@@ -44,7 +44,7 @@ describe('Plugin', () => {
 
   const getEvents = useLlmObs({ plugin: 'anthropic' })
 
-  withVersions('anthropic', '@anthropic-ai/sdk', (version, moduleName, resolvedVersion) => {
+  withVersions('anthropic', '@anthropic-ai/sdk', (version, moduleName, realVersion) => {
     let client
 
     before(() => {
@@ -52,7 +52,7 @@ describe('Plugin', () => {
       client = new Anthropic({ baseURL: 'http://127.0.0.1:9126/vcr/anthropic' })
     })
 
-    const isBetaSupported = semver.gte(semver.coerce(resolvedVersion), '0.33.0')
+    const isBetaSupported = semifies(realVersion, '>=0.33.0')
 
     describe('messages.create', () => {
       it('creates a span', async () => {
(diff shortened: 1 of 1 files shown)


packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
```diff
@@ -38,19 +39,21 @@ function assertLLMObsSpan (apmSpans, llmobsSpans) {
...
-  withVersions('anthropic', '@anthropic-ai/sdk', (version) => {
+  withVersions('anthropic', '@anthropic-ai/sdk', (version, moduleName, resolvedVersion) => {
     let client
 
     before(() => {
       const { Anthropic } = require(`../../../../../../versions/@anthropic-ai/sdk@${version}`).get()
       client = new Anthropic({ baseURL: 'http://127.0.0.1:9126/vcr/anthropic' })
     })
 
+    const isBetaSupported = semver.gte(semver.coerce(resolvedVersion), '0.33.0')
```
sabrenner: let's use `semifies` i think the rest of the guild is trying to move away from `semver` if possible since `semifies` is a smaller dev dep and does most of what we want
```suggestion
    const isBetaSupported = semifies(realVersion, '>=0.33.0')
```
with requiring `semifies` at the top
```javascript
const semifies = require('semifies')
```
but lmk if that doesn't work!

packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
```diff
@@ -133,5 +136,64 @@ describe('Plugin', () => {
         })
       })
     })
+
+    describe('beta.messages.create', () => {
+      before(function () {
+        if (!isBetaSupported || !client.beta?.messages) this.skip()
```
sabrenner: just to double check for myself - i would think that just the `isBetaSupported` is enough, if there's also `!client.beta.messages` for some of those versions, should we refine our version patching/support?

packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
```diff
@@ -133,5 +136,64 @@ describe('Plugin', () => {
...
+
+          const { apmSpans, llmobsSpans } = await getEvents()
+          assertLLMObsSpan(apmSpans, llmobsSpans)
+        })
+      })
+    })
+
+    describe('beta.messages.stream', () => {
+      before(function () {
+        if (!isBetaSupported || !client.beta?.messages?.stream) this.skip()
```
sabrenner, line 180: likewise here

sabrenner (approved):
lgtm just a couple thoughts for testing! glad it was easy enough to do with another patching hook 😄

packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
PROFeNoM: The `stream` method was added **after** 0.33.0, so unlike the check above, I'd argue this one is necessary :+1:

To be more precise, from the changelog, it got added in 0.35.0

sabrenner (approved):
🚢

PR #7021, opened 2025-12-02
chore(deps): Bump Next.js test-version

CarlesDD:
### What does this PR do?

- Bumps Next.js test version to **16.0.5**
- Makes `next.config.js` version-aware:
  - Reads the installed Next.js version at build time.
  - Applies `eslint.ignoreDuringBuilds` only for **Next.js < 16** (this property was removed in Next.js 16).
  - Applies `turbopack.root `only for **Next.js ≥ 16**. Since the repository contains multiple lockfiles, setting an explicit root is required for Turbopack to resolve the correct workspace.

### Motivation

Our test suite runs across a matrix of Next.js versions, and a static configuration led to build errors due to breaking changes introduced in newer releases.
This PR ensures that the configuration adapts automatically to each version, keeping all test runs compatible.

### Additional Notes
[APPSEC-60069](https://datadoghq.atlassian.net/browse/APPSEC-60069)

[APPSEC-60069]: https://datadoghq.atlassian.net/browse/APPSEC-60069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/test/appsec/next/pages-dir/next.config.js
+++ b/packages/dd-trace/test/appsec/next/pages-dir/next.config.js
@@ -13,4 +13,10 @@ if (major < 16) {
   }
 }
 
+if (major >= 16) {
+  nextConfig.turbopack = {
+    root: __dirname,
+  }
+}
+
 module.exports = nextConfig


packages/dd-trace/test/appsec/next/app-dir/next.config.js
```diff
@@ -1,13 +1,20 @@
...
   output: 'standalone',
   outputFileTracingRoot: __dirname
 }
 
+if (major < 16) {
+  nextConfig.eslint = {
+    ignoreDuringBuilds: true,
+  }
+}
+
```
IlyasShabi, line 19: Why we do not have nextConfig.turbopack config on the else section?

packages/dd-trace/test/appsec/next/app-dir/next.config.js
```diff
@@ -1,13 +1,20 @@
+const nextPkg = require('next/package.json')
+const [majorStr] = nextPkg.version.split('.')
+const major = Number(majorStr)
```
IlyasShabi, line 3: Why not using semver?

simon-id: we don't use semver in dd-trace-js anymore :p
in a lot of places we use this trick when we only need the major version.

simon-id: it's an effort to reduce dependency bloat and perf

packages/dd-trace/test/appsec/next/app-dir/next.config.js
CarlesDD: Because using separate conditionals makes the intent clearer:
- `eslint.ignoreDuringBuilds` applies only when the version that supports it (Next < 16).
- `turbopack.root` applies only to the version that supports it (Next ≥ 16).

These are capability-based rules, not two branches of the same decision, they are independent compatibility rules that apply depending on the Next.js version.
As a result, making them as independent conditions better reflects the actual semantics of the configuration.

Also, this structure is also more maintainable and extensible.
If in the future we need to introduce additional version-specific settings, adding another independent `if` is straightforward. An `if/else` structure would have to be reworked or extended awkwardly, forcing us into an artificial branching model that doesn't fit the real logic.

packages/dd-trace/test/appsec/next/app-dir/next.config.js
CarlesDD: I avoided using `semver` because our version check is very simple (only depends on the major version). If we ever need more complex version handling, we can revisit it.

IlyasShabi (approved):

PR #7023, opened 2025-12-03
Conti/all tooling changes

wconti27:
### What does this PR do?

Summary of Changes:
- added assertion helper that allows for asserting type matching for assertObjectContains (useful for asserting error tags since stacktrace / error message is difficult to get exact values)
- added abstraction layer for plugin tests. The abstraction handles withVersions setup, mock agent setup, tracer initialization, and mock agent closing. Allows plugin test files to be simpler, less to generate for an AI Agent.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/dd-trace/src/plugins/tracing.js
```diff
@@ -66,31 +66,35 @@ class TracingPlugin extends Plugin {
     this.addError(ctxOrError)
   }
 
-  addTraceSubs () {
+  registerOperation (channelName) {
```
rochdev: Why the alias? The way this is named I would expect it takes only an operation name and do some additional magic. Otherwise it's redundant.

.gitignore
```diff
@@ -135,3 +135,10 @@ packages/datadog-plugin-azure-functions/test/integration-test/fixtures/node_modu
 __azurite_db_queue__.json
 __azurite_db_queue_extent__.json
 __queuestorage__/AzuriteConfig
+
+
+# ignore apm-instrumentation-toolkit analysis results dirs
+.analysis/
+
+# ignore claude settings
+.claude/
```
rochdev:
```suggestion
.claude/

```

packages/dd-trace/src/plugins/consumer.js
```diff
@@ -7,14 +7,20 @@ class ConsumerPlugin extends InboundPlugin {
   static kind = 'consumer'
   static type = 'messaging'
 
-  startSpan (options, enterOrCtx) {
+  startSpan (name, options, enterOrCtx) {
+    if (typeof name === 'object' && name !== null && enterOrCtx === undefined) {
```
rochdev: Why the change? I don't see any usage change?

rochdev (dismissed):

packages/dd-trace/src/plugins/tracing.js
wconti27: removed.

.gitignore
wconti27: fixed

packages/dd-trace/src/plugins/consumer.js
wconti27: change no longer in PR, removed..

integration-tests/helpers/index.js
```diff
@@ -635,14 +640,8 @@ function setShouldKill (value) {
   })
 }
 
-// @ts-expect-error assert.partialDeepStrictEqual does not exist on older Node.js versions
-// eslint-disable-next-line n/no-unsupported-features/node-builtins
-const assertObjectContains = assert.partialDeepStrictEqual || function assertObjectContains (actual, expected, msg) {
```
BridgeAR, line 640: While I understand the wish for asymmetric matchers, this is less than ideal. The output of partialDeepStrictEqual in error cases is just vastly better than the fallback. As such, I think we should not use this by default.

If we want to use these, right away (I do plan on implementing these in Node.js assert soon as well), than we should run the native method first and only in the failure case run the fallback with asymmetric matchers.

packages/dd-trace/src/plugins/composite.js
```diff
@@ -6,8 +6,21 @@ class CompositePlugin extends Plugin {
   constructor (...args) {
     super(...args)
 
+    // Track instantiated plugin names for configure()
+    this._pluginNames = []
```
BridgeAR: Please use a private property

BridgeAR (changes requested):

integration-tests/helpers/index.js
BridgeAR: Another way to improve the output: we could create Node.js AssertionErrors. That way the output should be good for the fallback for the non asymmetric matcher cases.

packages/dd-trace/src/plugins/composite.js
wconti27: done

integration-tests/helpers/index.js
wconti27: done

packages/dd-trace/test/setup/helpers/integration-test-helpers/index.js
```diff
@@ -0,0 +1,62 @@
+'use strict'
+
+const { describe, it, beforeEach, afterEach, before, after } = require('mocha')
+const { expect } = require('chai')
```
BridgeAR: Chai was removed from the code, please use Node.js assert instead

packages/dd-trace/src/plugins/composite.js
```diff
@@ -3,11 +3,23 @@
...
 class CompositePlugin extends Plugin {
+  #pluginNames = []
+
   constructor (...args) {
     super(...args)
 
     for (const [name, PluginClass] of Object.entries(this.constructor.plugins)) {
-      this[name] = new PluginClass(...args)
+      // Handle case where value is an array of plugin classes
+      if (Array.isArray(PluginClass)) {
```
BridgeAR, line 13: Do we already have cases where this would be used?

integration-tests/helpers/index.js
```diff
@@ -24,6 +24,11 @@ const { DEBUG } = process.env
 // This is set by the setShouldKill function
 let shouldKill
 
+// Symbol constants for dynamic value matching in assertObjectContains
+const ANY_STRING = Symbol.for('test.ANY_STRING')
+const ANY_NUMBER = Symbol.for('test.ANY_NUMBER')
+const ANY_VALUE = Symbol.for('test.ANY_VALUE')
```
BridgeAR, line 30: Could we already use these somewhere in our code where it would make the assertion more readable?

I think it is best to not only add a tool but to use it right away to justify its addition.

Should we just export these instead of using Symbol.for? That way we guarantee that there is no typo while using them.

integration-tests/helpers/index.js
```diff
@@ -670,15 +687,47 @@ const assertObjectContains = assert.partialDeepStrictEqual || function assertObj
...
 
+// Use native partialDeepStrictEqual when available and no asymmetric matchers,
+// otherwise fall back to custom implementation for better asymmetric matcher support
+// @ts-expect-error assert.partialDeepStrictEqual does not exist on older Node.js versions
+// eslint-disable-next-line n/no-unsupported-features/node-builtins
+const nativePartialDeepStrictEqual = assert.partialDeepStrictEqual
+
+const assertObjectContains = function assertObjectContains (actual, expected, msg) {
+  // If expected contains asymmetric matchers, use our custom implementation
+  if (containsAsymmetricMatchers(expected)) {
```
BridgeAR: Are we expecting to use the asymmetric matchers in most situations? Depending on that, I would use a different implementation: Instead of having to iterate through the object as a whole each time, we could just use a try / catch. If an error is thrown, we rerun the assertion using the asymmetric matchers. If that throws as well, we throw the original caught error. That way the average case stays fast and there is no need to check anything in addition (because we expect no asymmetric matchers being used).

packages/dd-trace/src/plugins/composite.js
wconti27: No, but the idea would be for us to still be able to use `.assertSomeTrace({span object})`, mainly for tags that we don't know the exact value, such as error stack traces.

packages/dd-trace/test/setup/helpers/integration-test-helpers/index.js
wconti27: Done.

integration-tests/helpers/index.js
wconti27: Really they are mainly for targeting scenarios, especially error tests since we can't know stacktrace reliably but should test for it. Or any other tags where the value may change between tests. Okay, I made that change 👍

packages/dd-trace/src/plugins/composite.js
wconti27: Being used in this PR: https://github.com/DataDog/dd-trace-js/pull/7123/files#:~:text=%7D)-,it(%27should%20generate%20span%20with%20error%20tags%20(error%20path)%27,%7D),-describe(%27Queue

integration-tests/helpers/index.js
wconti27: Being used in this PR: https://github.com/DataDog/dd-trace-js/pull/7123/files#:~:text=%7D)-,it(%27should%20generate%20span%20with%20error%20tags%20(error%20path)%27,%7D),-describe(%27Queue

wconti27: I added it to exports 👍

PR #7028, opened 2025-12-03
fix(otel): support setting DD_METRICS_OTEL_ENABLED to False

mabdinur:
### What does this PR do?

Fixes `DD_METRICS_OTEL_ENABLED=false` being treated as truthy by using `isTrue()` to parse the environment variable.

### Motivation

The string `"false"` was truthy, incorrectly enabling OpenTelemetry metrics when disabled.

### Plugin Checklist

N/A - Configuration fix.

### Additional Notes

- Updated `config.js` line 630 to use `isTrue(DD_METRICS_OTEL_ENABLED)`
- Added test case in `packages/dd-trace/test/opentelemetry/metrics.spec.js`

### DIFF
--- a/packages/dd-trace/src/config.js
+++ b/packages/dd-trace/src/config.js
@@ -626,11 +626,11 @@ class Config {
     target.otelMaxExportBatchSize = nonNegInt(OTEL_BSP_MAX_EXPORT_BATCH_SIZE, 'OTEL_BSP_MAX_EXPORT_BATCH_SIZE', false)
     target.otelMaxQueueSize = nonNegInt(OTEL_BSP_MAX_QUEUE_SIZE, 'OTEL_BSP_MAX_QUEUE_SIZE', false)
 
-    const otelMetricsExporter = OTEL_METRICS_EXPORTER?.toLowerCase() !== 'none'
+    const otelMetricsExporterEnabled = OTEL_METRICS_EXPORTER?.toLowerCase() !== 'none'
     this.#setBoolean(
       target,
       'otelMetricsEnabled',
-      DD_METRICS_OTEL_ENABLED && isTrue(DD_METRICS_OTEL_ENABLED) && otelMetricsExporter
+      DD_METRICS_OTEL_ENABLED && isTrue(DD_METRICS_OTEL_ENABLED) && otelMetricsExporterEnabled
     )
     // Set OpenTelemetry metrics configuration with specific _METRICS_ vars
     // taking precedence over generic _EXPORTERS_ vars


packages/dd-trace/src/config.js
```diff
@@ -627,7 +627,11 @@ class Config {
     target.otelMaxQueueSize = nonNegInt(OTEL_BSP_MAX_QUEUE_SIZE, 'OTEL_BSP_MAX_QUEUE_SIZE', false)
 
     const otelMetricsExporter = !OTEL_METRICS_EXPORTER || OTEL_METRICS_EXPORTER.toLowerCase() !== 'none'
-    this.#setBoolean(target, 'otelMetricsEnabled', DD_METRICS_OTEL_ENABLED && otelMetricsExporter)
+    this.#setBoolean(
+      target,
+      'otelMetricsEnabled',
+      DD_METRICS_OTEL_ENABLED && isTrue(DD_METRICS_OTEL_ENABLED) && otelMetricsExporter
```
mabdinur: We use this pattern (`DD_METRICS_OTEL_ENABLED && isTrue(DD_METRICS_OTEL_ENABLED)`) in many places in our configs. It would be nice if we had a MaybeBool() helper that returned `undefined` if a config was not set. 

This change is beyond the scope of this PR

BridgeAR: Nit

```suggestion
      otelMetricsExporter && isTrue(DD_METRICS_OTEL_ENABLED)
```

`isTrue()` handles undefined and there is no need to call the method in case the other part is already falsy :)

packages/dd-trace/src/config.js
```diff
@@ -627,7 +627,11 @@ class Config {
     target.otelMaxQueueSize = nonNegInt(OTEL_BSP_MAX_QUEUE_SIZE, 'OTEL_BSP_MAX_QUEUE_SIZE', false)
 
     const otelMetricsExporter = !OTEL_METRICS_EXPORTER || OTEL_METRICS_EXPORTER.toLowerCase() !== 'none'
-    this.#setBoolean(target, 'otelMetricsEnabled', DD_METRICS_OTEL_ENABLED && otelMetricsExporter)
+    this.#setBoolean(
+      target,
+      'otelMetricsEnabled',
+      otelMetricsExporter && isTrue(DD_METRICS_OTEL_ENABLED)
```
mabdinur: revert to previous pattern to fix failing tests

```suggestion
      DD_METRICS_OTEL_ENABLED && isTrue(DD_METRICS_OTEL_ENABLED) && otelMetricsExporter
```

packages/dd-trace/src/config.js
mabdinur: When `otelMetricsExporter` is False and `DD_METRICS_OTEL_ENABLED` is undefined, this sets `otelMetricsEnabled` to False. I will clean up how we set `otelMetricsExporter`

mabdinur: Ahhh I found it. The issue here is that `isTrue` always returns a boolean: https://github.com/DataDog/dd-trace-js/blob/0354f5789808d2964e5d43fd0ab36b4b3761ba4f/packages/dd-trace/src/util.js#L7-L8. 

If `DD_METRICS_OTEL_ENABLED ` is unset we want the env config value to be undefined so that we use the default value (or a value from another source).

PR #7042, opened 2025-12-06
fix(debugger): initialize lastCaptureNs to ensure first probe hit is captured

watson:
### What does this PR do?

Initialize a probes `lastCaptureNs` to `BigInt(Number.MIN_SAFE_INTEGER)`, ensuring the first probe hit always passes the sampling check regardless of when it occurs.
    
### Motivation

When a probe is added with any sampling rate, the first probe hit could be incorrectly skipped if it occurred within the sampling interval after process start.
    
This happened because `lastCaptureNs` was initialized to `0n`, causing the sampling check `start - probe.lastCaptureNs < probe.nsBetweenSampling` to evaluate to true when the process had been running for less time than the sampling interval.
    
For example:
- At 10 samples/sec: first hit skipped if within first 100ms
- At 5 samples/sec: first hit skipped if within first 200ms
- At 1 sample/sec: first hit skipped if within first 1 second
- At 0.5 samples/sec: first hit skipped if within first 2 seconds

### DIFF
--- a/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
+++ b/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
@@ -64,7 +64,7 @@ async function addBreakpoint (probe) {
   const snapshotsPerSecond = probe.sampling?.snapshotsPerSecond ?? (probe.captureSnapshot
     ? MAX_SNAPSHOTS_PER_SECOND_PER_PROBE
     : MAX_NON_SNAPSHOTS_PER_SECOND_PER_PROBE)
-  probe.nsBetweenSampling = BigInt(1 / snapshotsPerSecond * 1e9)
+  probe.nsBetweenSampling = BigInt(Math.trunc(1 / snapshotsPerSecond * 1e9))
   // Initialize to a large negative value to ensure first probe hit is always captured
   probe.lastCaptureNs = BigInt(Number.MIN_SAFE_INTEGER)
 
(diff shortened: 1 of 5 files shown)


packages/dd-trace/src/debugger/devtools_client/breakpoints.js
```diff
@@ -65,7 +65,8 @@ async function addBreakpoint (probe) {
     ? MAX_SNAPSHOTS_PER_SECOND_PER_PROBE
     : MAX_NON_SNAPSHOTS_PER_SECOND_PER_PROBE)
   probe.nsBetweenSampling = BigInt(Math.trunc(1 / snapshotsPerSecond * 1e9))
-  probe.lastCaptureNs = 0n
+  // Initialize to a large negative value to ensure first probe hit is always captured
+  probe.lastCaptureNs = BigInt(Number.MIN_SAFE_INTEGER)
```
simon-id, line 69: nit: since you're using BigInts, is it really useful to use MIN_SAFE_INTEGER instead of just `-99999999`

watson: Sorry?

watson: Did you mean *not useful*? I don't care as long as it's a large negative number. I know that `BigInt` and `Number.MIN_SAFE_INTEGER` is not technically meant to be used together like this, but it's perfectly safe and produces the expected result

simon-id: yeah yeah, it was a question "is it really useful ...?", but as you said, it works so i approved already :+1:

PR #7049, opened 2025-12-08
ci: safer dependabot updates

BridgeAR:
See commit messages for details


.github/dependabot.yml
```diff
@@ -95,12 +95,51 @@ updates:
...
+  - package-ecosystem: "npm"
+    directories:
+      - "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 1
+    cooldown:
+      default-days: 3
+      exclude:
+        - "@datadog/*"
```
simon-id: probably don't need this exclude since this is just for otel no ?
```suggestion
```

.github/dependabot.yml
```diff
@@ -95,12 +95,51 @@ updates:
...
+
+  # Instrumented library support range
   - package-ecosystem: "npm"
     directories:
       - "/packages/dd-trace/test/plugins/versions"
     schedule:
       interval: "daily"
     open-pull-requests-limit: 1
+    cooldown:
+      default-days: 1
```
simon-id: the security team would probably like this to be higher :/

BridgeAR: I bumped it further and will discuss it with the team again. I think we should merge it as is as this is discussed with the guild so far and already makes it more secure.

simon-id: I agree let's merge like this

PR #7051, opened 2025-12-08
chore: prevent most double wrapping in shimmer

BridgeAR:
This prevents double wrapping in case it would happen.


BridgeAR:
Not planned anymore until this ever becomes an issue again

Issue #7025, opened 2025-12-03
[BUG]: ws instrumentation breaks `removeListener`


twooster:
### Tracer Version(s)

5.66.0 and later

### Node.js Version(s)

22.20, 24.11.1

### Bug Report

The trace instrumentation for the `ws` module wraps the `on` method on websockets. The implementation for on-"message" wraps the passed-in callback before passing it to the original `on` method.

However, the implementation does not add related bookkeeping to `removeListener`. This means that a subsequent call to `removeListener` will fail to remove the original listener.

For certain patterns (handshake with one handler, attach another handler for the desired protocol version), this entirely breaks applications.

### Reproduction Code

Completely untested, but an example:

```
import { WebSocketServer } from "ws";

const wss = new WebSocketServer({
  port: 8080,
});

wss.on("connection", ws => {
  ws.on("error", console.error);

  const v1ProtocolHandler = message => {
    console.log("received", message.toString());
    ws.send(message);
  };

  const handshakeHandler = message => {
    ws.removeListener(handshakeHandler);
    if (message.toString() === "v1") {
      ws.on("message", v1ProtocolHandler);
      ws.send("v1");
    } else {
      console.log(`Unknown protocol version: ${message}, terminating socket`);
      ws.terminate();
    }
  }
  ws.on("message", handshakeHandler);
});
```

Repro is going to be to connect to the server, send "v1", then send anything that isn't "v1", and watch the connection terminate because the handshake handler was never removed.

### Error Logs

No errors are logged, this is purely a "breaks userspace" problem.

### Tracer Config

This is for any/all ws instrumentation. Bug is plain here:

[Wrapping the argument to 'on'](https://github.com/DataDog/dd-trace-js/blob/0bb1f17270289fecdd042e2c5a769b54995a5677/packages/datadog-instrumentations/src/ws.js#L79-L86)
[Only patching 'on' but not patching 'removeListener'](https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-instrumentations/src/ws.js#L117-L119)

### Operating System

24.6.0 Darwin Kernel Version

### Bundling

Unsure

PR #7052, opened 2025-12-08
fix(ws): wrap event remove handler for message event removal

BridgeAR:
The remove handler has to understand what method was originally wrapped and remove the wrapped method. Otherwise it would still exist and be triggered.

Fixes: https://github.com/DataDog/dd-trace-js/issues/7025


packages/datadog-instrumentations/src/ws.js
```diff
@@ -76,13 +76,35 @@ function createWrappedHandler (handler) {
   }
 }
 
-function wrapListener (originalOn) {
-  return function (eventName, handler) {
-    if (eventName === 'message') {
-      return originalOn.call(this, eventName, createWrappedHandler(handler))
+function wrapListeners (wsPrototype) {
+  const callbackMap = new WeakMap()
```
simon-id: nit: I am unsure why this function and thus wrapListener() and removeListener() have to be a closure. Does a new callbackMap have to be created everytime it's called ? Can't it be global-ish ?

simon-id (dismissed):

packages/datadog-instrumentations/src/ws.js
BridgeAR: I updated the code. I agree about your perspective. After looking at it further, I actually believe we should not even wrap all these handlers, but that is for another PR.

simon-id:
I assume the CI failures are because of what happened yesterday so I'll update the branch and enable automerge

PR #7070, opened 2025-12-10
ci: fix dependabot.yml to not have overlapping directories

BridgeAR:
A package.json can only be handled by a single config. Individual dependencies can not be handled separately in a different config. Thus, this simplifies our overall config by defining increase for all our dependencies besides the OTEL api one, since that must be widened. We have to manually update it for now, until a better solution is found.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -83,9 +83,10 @@ updates:
       - dependency-name: "@opentelemetry/resources"
         # 2.0.0 onwards only supports Node.js 18.19.0 and above
         update-types: ["version-update:semver-major"]
-        # @opentelemetry/api has to be widened, not increased.
-        # Thus, we ignore it and update it manually.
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -77,9 +77,15 @@ updates:
       - dependency-name: "glob"
         # 11.0.0 onwards only supports Node.js 20 and above
         update-types: ["version-update:semver-major"]
-        # OTEL is handled separately due to the version increase here.
-        # The package.json range should only manually be adjusted for OTEL.
-      - dependency-name: "@opentelemetry/*"
+      - dependency-name: "@opentelemetry/core"
```
pabloerhard, line 80: Considering opentelemetry/* was removed and individual ignores were added, should an ignore also be added for @opentelemetry/api-logs which we also depend on?

simon-id (approved):

PR #7072, opened 2025-12-10
feat(prisma): Prisma Output and V7 Support

pabloerhard:
### What does this PR do?
This PR intends to add support prisma client generated through the output field. The output field creates the prisma client in a custom path, which dd-trace has no context of. In order to properly instrument the client generated in this custom output, support was added to hook into the internals of the client. 

To disregard the client’s actual location, we hooked into the relative path that the client consistently uses regardless of where it is located. This relative path can be found in the hooks.js file.

In addition to supporting the output field for Prisma v6, support for Prisma v7 has been added. Prisma v7 introduces several breaking changes, such as disallowing the use of url inside the schema, enforcing the use of adapters, and providing official support for ESM.

### Motivation
Have proper support for Prisma Client instrumentation

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/loader-hook.mjs
+++ b/loader-hook.mjs
@@ -29,9 +29,9 @@ function addInstrumentations (data) {
   const instrumentations = Object.keys(hooks)
 
   for (const moduleName of instrumentations) {
-    if (isRelativeRequire(moduleName)) {
-      continue
-    } else {
+    // Skip instrumentation hooks with relative module names
+    // since there is no current business need of instrumenting imports outside of the node_modules folder
+    if (!isRelativeRequire(moduleName)) {
       data.include.push(new RegExp(`node_modules/${moduleName}/(?!node_modules).+`), moduleName)
     }
   }


packages/dd-trace/src/supported-configurations.json
```diff
@@ -134,6 +134,7 @@
     "DD_OPENAI_SPAN_CHAR_LIMIT": ["A"],
     "DD_PIPELINE_EXECUTION_ID": ["A"],
     "DD_PLAYWRIGHT_WORKER": ["A"],
+    "DD_PRISMA_OUTPUT": ["A"],
```
BridgeAR:
```suggestion
```

scripts/install_plugin_modules.js
```diff
@@ -167,7 +169,8 @@ async function assertPeerDependencies (rootFolder, parent = '') {
     const versionPkgJsonPath = join(folder, 'package.json')
     const versionPkgJson = require(versionPkgJsonPath)
 
-    for (const { dep, name } of externalDeps.get(externalName)) {
+    for (const { dep, name, node } of externalDeps.get(externalName)) {
+      if (node && !semver.satisfies(process.versions.node, node)) continue
```
BridgeAR, line 173: I like that change!

.github/workflows/apm-integrations.yml
```diff
@@ -1047,11 +1056,26 @@ jobs:
...
+      - uses: ./.github/actions/install
+      - name: Run tests
+        run: yarn test:plugins:ci
+      - if: always()
+        uses: ./.github/actions/testagent/logs
+        with:
+          suffix: plugins-${{ github.job }}-${{ matrix.node-version }}-${{ matrix.range_clean }}
+      - uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
+      - uses: DataDog/junit-upload-github-action@762867566348d59ac9bcf479ebb4ec040db8940a  # v2.0.0
+        if: always() && github.actor != 'dependabot[bot]'
```
BridgeAR, line 1075: Is this change something we should generalize? I think we should always use codecov, or why does this need special handling here?

packages/dd-trace/src/ritm.js
```diff
@@ -167,6 +188,30 @@ Hook.reset = function () {
...
+  if (moduleName.startsWith('./') || moduleName.startsWith('../') || moduleName.startsWith('/')) {
+    return true
+  }
+
+  if (moduleName.includes('/') && !moduleName.includes('node_modules/') && !moduleName.startsWith('@')) {
+    return true
+  }
+
+  return false
+}
```
BridgeAR: This exists abstracted in a separate file as well and seems duplicated :)

packages/datadog-instrumentations/src/helpers/shared-utils.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+
+function isFilePath (moduleName) {
```
BridgeAR: `isFilePath()` is AFAIK only used for relativePaths, right? Should we also name it as such?
I think we can also simplify the check to just checking for `.` or `/` to be at the start and return that directly.

packages/datadog-instrumentations/src/helpers/shared-utils.js
```diff
@@ -0,0 +1,17 @@
...
+  if (moduleName.startsWith('./') || moduleName.startsWith('../') || moduleName.startsWith('/')) {
+    return true
+  }
+
+  if (moduleName.includes('/') && !moduleName.includes('node_modules/') && !moduleName.startsWith('@')) {
```
BridgeAR: The first condition can not be false anymore, due to matching for that in the first if statement :)

packages/datadog-plugin-prisma/test/integration-test/client.spec.js
```diff
@@ -14,75 +14,152 @@ const {
...
+          })
+
+          // node v18 needs the package.json to have type module to treat .js files as esm
+          if (config.ts && config.name.includes('v6')) {
+            const fs = require('fs')
+            const path = require('path')
+            const distPath = path.join(cwd, 'dist')
+            if (!fs.existsSync(distPath)) {
+              fs.mkdirSync(distPath, { recursive: true })
             }
```
BridgeAR:
```suggestion
            try {
              fs.mkdirSync(distPath, { recursive: true })
            } catch {}
```

That way there is no race condition and it should be faster

packages/datadog-plugin-prisma/test/index.spec.js
```diff
@@ -7,299 +7,433 @@ const path = require('node:path')
 
 const { after, before, beforeEach, describe, it } = require('mocha')
 const semifies = require('semifies')
-
+const semver = require('semver')
```
BridgeAR: How come we need semver and semifies? 😢

packages/datadog-plugin-prisma/test/index.spec.js
```diff
@@ -7,299 +7,433 @@ const path = require('node:path')
...
+              assert.strictEqual(traces[0][0].service, expectedSchema.client.serviceName)
+
+              // grabbing actual db query span
+              if (config.v7) {
+                const pgSpan = traces[0].find(span => span.name === 'pg.query')
+                assert.strictEqual(pgSpan.resource, 'SELECT 1')
+                assert.strictEqual(pgSpan.type, 'sql')
+                assert.strictEqual(pgSpan.meta['span.kind'], 'client')
+                assert.strictEqual(pgSpan.name, 'pg.query')
+                assert.strictEqual(pgSpan.service, 'test-postgres')
```
BridgeAR:
```suggestion
                assertObjectContains(
                  traces[0],
                  {
                    resource: 'SELECT 1',
                    type: 'sql',
                    meta: {
                      'span.kind': 'client'
                    },
                    name: 'pg.query',
                    service: 'test-postgres'
                  }
                )
```

Using this helper is favorable due to getting much better errors. A rule that I use: the less assertions for a single input, the better.

packages/dd-trace/src/ritm.js
```diff
@@ -104,7 +103,7 @@ function Hook (modules, options, onrequire) {
     // The module has already been loaded,
     // so the patching mark can be cleaned up.
     delete patching[filename]
-
+    const isRelativeHook = isFilePath(request) && moduleHooks[request]
```
BridgeAR: I believe this is only needed in the non-core case and we should move it as such.

packages/dd-trace/src/ritm.js
```diff
@@ -5,7 +5,7 @@ const Module = require('module')
 const parse = require('../../../vendor/dist/module-details-from-path')
 const dc = require('dc-polyfill')
 const { getEnvironmentVariable } = require('../../dd-trace/src/config-helper')
-
+const fs = require('fs')
```
BridgeAR: Nit: please move this to the top and sort the require calls by 1. Node.js modules, 2. node_modules, 3. relative require calls.

packages/dd-trace/src/ritm.js
```diff
@@ -117,32 +116,54 @@ function Hook (modules, options, onrequire) {
...
       }
 
-      if (!res || res !== filename) {
-        // this is a module-internal file
-        // use the module-relative path to the file, prefixed by original module name
-        name = name + path.sep + path.relative(basedir, filename)
+      // figure out if this is the main module file, or a file inside the module
+      // Skip this for paths since it's intended for node_modules paths
+
+      if (!isRelativeHook) {
```
BridgeAR: I believe this is always false in case `stat` was true, right? If that is the case, we could simplify the code accordingly.

packages/dd-trace/src/ritm.js
```diff
@@ -117,32 +116,54 @@ function Hook (modules, options, onrequire) {
...
-      }
+        hooks = moduleHooks[name]
+        if (!hooks) return exports // abort if module name isn't on whitelist
+      } else {
+        // check if the full filename if the filename was registered, if not then prefix matching will be done
+        hooks = moduleHooks[filename]
+
+        if (!hooks && isRelativeHook) {
+          hooks = moduleHooks[request]
+          name = request
```
BridgeAR: Could this name ever not be true in here?

.github/workflows/apm-integrations.yml
pabloerhard: What do you mean by it should always use codecov? This is essentially doing the same as ./.github/actions/plugins/test but with some extra nits such as the version handling

packages/datadog-plugin-prisma/test/integration-test/client.spec.js
pabloerhard: Nice change

packages/dd-trace/src/ritm.js
pabloerhard: Nope, that was leftover of the past implementation.

packages/dd-trace/src/ritm.js
pabloerhard: true

packages/datadog-plugin-prisma/test/index.spec.js
pabloerhard: semifies was used before this changes, but the condition was never working because the version had to be cleaned before being send to semifies (hence why semver was added to the scope). Considering we now have semver as well we could just use semver.satisfies().

loader-hook.mjs
```diff
@@ -28,7 +29,11 @@ function addInstrumentations (data) {
   const instrumentations = Object.keys(hooks)
 
   for (const moduleName of instrumentations) {
-    data.include.push(new RegExp(`node_modules/${moduleName}/(?!node_modules).+`), moduleName)
+    if (isRelativeRequire(moduleName)) {
+      continue
+    } else {
```
BridgeAR: Could you add a comment why this is ignored here? :)

It is not really clear for me reading this code.

```suggestion
    // Relative requires are skipped due to ...
    if (!isRelativeRequire(moduleName)) {
```

BridgeAR (commented):
LGT, just one last comment about the loader-hook that I can not follow by reading the code including a nit about the if statement at the same spot.

loader-hook.mjs
pabloerhard: Yep, the reason we skip them is that there is no current business need of supporting a relative import

PR #7090, opened 2025-12-11
mongodb-core: Enhance isHeartbeat to check for ismaster command

thena-seer:
### What does this PR do?
Updated isHeartbeat function to include ismaster check. This will cause such commands to be skipped when heartbeatEnabled=false is configured.

### Motivation

Flame graphs that look like this, even when heartbeatEnabled=false
Screenshot 2025-12-11 at 10 58 29 AM

ismaster is another operation name used for both handshake and heartbeat purposes in older mongo versions. If a user of datadog chooses to configure `heartbeatEnabled=false` they should reasonably expect these operations to be suppressed.

### Plugin Checklist

- [x] Unit tests. (logic covered sufficiently by existing test case, additional test case would not add meaningful value)
- [x] Integration tests. (integration tests exist for this plugin, but do not go as fine grained as this logic)
- [x] Benchmarks. (trivial impact on performance)
- [x] TypeScript [definitions][1]. (no impact)
- [x] TypeScript [tests][2]. (no impact)
- [x] API [documentation][3]. (no impact)
- [x] CI [jobs/workflows][4]. (no impact)

### DIFF
--- a/packages/datadog-plugin-mongodb-core/src/index.js
+++ b/packages/datadog-plugin-mongodb-core/src/index.js
@@ -204,11 +204,11 @@ function isBinary (val) {
 }
 
 function isHeartbeat(ops, config) {
-  // Check if it's a heartbeat command hello: 1, helloOk: 1, or ismaster: 1
+  // Check if it's a heartbeat command https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.md
   return (
     ops &&
     typeof ops === "object" &&
-    (ops.hello === 1 || ops.helloOk === true || ops.ismaster === 1)
+    (ops.hello === 1 || ops.helloOk === true || ops.ismaster === 1 || ops.isMaster === 1)
   );
 }
 


packages/datadog-plugin-mongodb-core/src/index.js
```diff
@@ -203,9 +203,13 @@ function isBinary (val) {
...
 
-function isHeartbeat (ops, config) {
-  // Check if it's a heartbeat command hello: 1 or helloOk: 1
-  return ops && typeof ops === 'object' && (ops.hello === 1 || ops.helloOk === true)
+function isHeartbeat(ops, config) {
+  // Check if it's a heartbeat command hello: 1, helloOk: 1, or ismaster: 1
+  return (
+    ops &&
+    typeof ops === "object" &&
+    (ops.hello === 1 || ops.helloOk === true || ops.ismaster === 1)
```
simon-id: I'm not familiar with this part of the code, but I see `helloOk` is in camelCase but `ismaster` is not. Is this intended and the correct fields, or a little typo ?

tlhunter: It looks like both are supported:

https://specifications.readthedocs.io/en/latest/mongodb-handshake/handshake/#terms

Or in other words, the code should probably look like this:

```
    (ops.hello === 1 || ops.helloOk === true || ops.ismaster === 1 || ops.isMaster === 1)
```

tlhunter: It's hard to find docs for this old feature. Do you know what versions of mongo it applies to? And are the values always exactly the integer `1`?

thena-seer: You're right that we should include both ismaster and isMaster according to [this spec](https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.md). I will make that update.

[The 4.4 docs](https://github.com/mongodb/docs/blob/v4.4/source/reference/command/isMaster.txt) confirm that the integer 1 is the correct value. The isMaster.txt file is removed from the 5.0 docs.

thena-seer: Changes have been made, ready for another look @tlhunter

PR #7099, opened 2025-12-12
move vendored dependencies to a folder than is not deleted by yarn

rochdev:
### What does this PR do?

Move vendored dependencies to a folder than is not deleted by yarn.

### Motivation

Yarn >=2 deletes files in the packages it installs including any `node_modules` folder.


BridgeAR:
Superseded

PR #7100, opened 2025-12-12
[SVLS-8161] Update Azure Function metadata detection logic

kathiehuang:
### What does this PR do?
If we're in an Azure Function on the flex consumption plan, set `aas.resource.group` to the `DD_AZURE_RESOURCE_GROUP` env var.

### Motivation
- We are currently adding support for profiling in Azure Functions, but the `aas.resource.group` tag in profiles for Node.js Azure Functions on the Flex Consumption Plan is currently being set to `flex`
  - This is because Flex Consumption functions don't have a `WEBSITE_RESOURCE_GROUP` env var, and the structure of the `WEBSITE_OWNER_NAME` env var is different from all other Azure Function hosting plans. As a result, we ask customers to [manually set DD_AZURE_RESOURCE_GROUP as an app setting](https://docs.datadoghq.com/serverless/azure_functions/?tab=nodejs#:~:text=DD_AZURE_RESOURCE_GROUP)
- Consistent with libdatadog changes made [here](https://github.com/DataDog/libdatadog/pull/1241)
  - See related ticket [here](https://datadoghq.atlassian.net/browse/SVLS-6337)

### Testing
- Deployed a Flex Consumption function and saw the correct `aas.resource.group`
image

Deployed a Node.js Windows Web App with the site extension, a Linux Web App with sidecar, and a Azure Container App to make sure this doesn't affect their profile tags

### How to test
1. From this branch, run `npm pack`
2. Move the newly created `.tgz` file to the root of your app directory
3. Update `package.json` to have the tracer point to this file: `"dd-trace": file:<name_of_tgz_file>`
4. If deploying a Windows Web App with the site extension, also add an env var `NODE_PATH=C:\home\site\wwwroot\node_modules`. If deploying apps using in-process, update the Dockerfile to also copy the `.tgz` file before it runs `npm ci`

### DIFF
--- a/packages/dd-trace/src/azure_metadata.js
+++ b/packages/dd-trace/src/azure_metadata.js
@@ -46,7 +46,7 @@ function buildMetadata () {
     WEBSITE_OWNER_NAME,
     WEBSITE_OS,
     WEBSITE_RESOURCE_GROUP,
-    WEBSITE_SITE_NAME,
+    WEBSITE_SITE_NAME
   } = getEnvironmentVariables()
 
   const subscriptionID = extractSubscriptionID(WEBSITE_OWNER_NAME)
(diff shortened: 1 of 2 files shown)


kathiehuang:
@codex review

PR #7104, opened 2025-12-15
fix(prisma): use real trace context in getTraceParent for DBM correlation.

### LINKED ISSUE #6063: [BUG]: Prisma DBM propagation does not work correctly.
### Tracer Version(s)

5.58.0

### Node.js Version(s)

22

### Bug Report

DBM propagation should work with Prisma but it doesn't because the `traceparent` is invalid.

### Reproduction Code

1. Setup Prisma
2. Setup dd-trace with dbmPropagationMode=full.
3. Register the dd-trace OTel TracerProvider
4. Run some queries.

Prisma supports trace propagation via OTel and adds the traceparent comment based on the OTel context. However using the new dd-trace Prisma support this comment always comes out as `/* traceparent='00-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-bbbbbbbbbbbbbbbb-01' */` because it is hard coded to that value [here](https://github.com/DataDog/dd-trace-js/blob/a57677c113b7b1e135604b17be99d83d7d257f3f/packages/datadog-instrumentations/src/prisma.js#L26).

### Error Logs

_No response_

### Tracer Config

_No response_

### Operating System

_No response_

### Bundling

Unsure

tristal:
### What does this PR do?

Fixes #6063 

This fixes the Prisma instrumentation to return actual trace IDs instead of hardcoded placeholder values, enabling proper Database Monitoring (DBM) correlation. The implementation:
- Retrieves active span context from async local storage
- Calls toTraceparent() to get the W3C traceparent with real trace IDs
- Forces sampled flag to ensure Prisma generates engine spans
- Falls back to placeholder when no active span exists

### Motivation
Database Monitoring correlation requires real trace IDs to link database queries with application traces. The hardcoded placeholder prevents this correlation from working, making it impossible to identify which application requests are causing database load.

### Plugin Checklist

- [ ] Unit tests.
- [X] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes

This is a bug fix (should be labeled `semver-patch`). The change maintains backward compatibility while fixing the DBM correlation issue.

### Reviewer's Guide

The forced 01 sampling flag ensures Prisma generates engine spans, while dd-trace still makes the actual sampling decision when recording/sending those spans.


rochdev:
> The overall idea is sound while having a brief look at it. The main issue is just that an instrumentation should never have access to the state.

I wonder if the TracingHelper could be defined on the subscriber side instead. This would make more sense as its whole purpose is to do tracing which is more inline with our plugins than the instrumentation. Maybe a channel could publish an event that would be handled by the plugin to get the helper back to the instrumentation?

cc @bojbrook

jeff-traba:
huge!

tristal:
> @tristal thank you very much for the PR, we did by now look into this to fit it into our plugin / instrumentation architecture and landed the fix.
> 
> This one was a bit tricky to decide what to do, so having it as community PR was difficult. Otherwise I would have gone ahead and merged this since the code itself seems good!

Thanks for the context, @BridgeAR. I have a similar question to Jeff's above. I understand the desire to align this with your long-term plugin/instrumentation architecture, but this is a breaking issue affecting many consumers of this package. Could you share a timeline for when your fix will be released? If it's not soon, it might be worth reconsidering this PR as an interim solution to unblock people in the meantime.

PR #7105, opened 2025-12-15
chore: remove plugin checklist from PR template

simon-id:
### What does this PR do?
This removes the "plugin" checklist from the PR description template, as no one really uses it much, and it's a relic of when dd-trace-js was mostly just plugins.


.github/pull_request_template.md
```diff
@@ -1,26 +1,8 @@
+<ins>**Please make sure your changes are properly tested**!</ins>
```
watson, line 1: Why `<ins>`?

.github/pull_request_template.md
```diff
@@ -1,26 +1,8 @@
...
-- [ ] API [documentation][3].
-- [ ] CI [jobs/workflows][4].
-
-[1]: https://github.com/DataDog/dd-trace-js/blob/master/index.d.ts
-[2]: https://github.com/DataDog/dd-trace-js/blob/master/docs/test.ts
-[3]: https://github.com/DataDog/documentation/blob/master/content/en/tracing/trace_collection/library_config/nodejs.md
-[4]: https://github.com/DataDog/dd-trace-js/blob/master/.github/workflows/plugins.yml
-
-### Additional Notes
-<!-- Anything else we should know when reviewing? -->
```
watson, line 24: Why remove the notes section?

CONTRIBUTING.md
```diff
@@ -214,4 +214,19 @@ $ yarn bench
 ```
 
 
+### Checklist
```
watson: Nit:

```suggestion
### Checklist

```

CONTRIBUTING.md
```diff
@@ -214,4 +214,19 @@ $ yarn bench
...
+Here is a simple checklist to help write a more complete PR:
+- [ ] Unit tests.
+- [ ] Integration tests.
+- [ ] Benchmarks.
+- [ ] TypeScript [definitions][2].
+- [ ] TypeScript [tests][3].
+- [ ] API [documentation][4].
+- [ ] CI [jobs/workflows][5].
```
watson: Why the `[ ]`? This is no longer an issue/PR

CONTRIBUTING.md
watson: Also the headline is a bit misleading in the context of `CONTRIBUTING.md`. It should maybe be `PR checklist` or something similar to indicate this is related to PRs?

.github/pull_request_template.md
simon-id: afaik that's how github supports underlining: 
image

.github/pull_request_template.md
simon-id: Because I find it superfluous. There are two possible cases:
1. you don't have any additional notes, and thus you need to remove this section manually.
2. you have additional notes, and do I REALLY need to remind you that you're allowed to add additional notes to your PR description ? like if you wanna write more stuff, just write more stuff.

wdyt ?

CONTRIBUTING.md
simon-id: so people that want to use the checklist can copy paste it in their description or somewhere else ;)

CONTRIBUTING.md
simon-id:
```suggestion

### PR Checklist

```

simon-id: Oh I was following the double empty line that the benchmark section above uses, but it looks like all the other ones are only one single empty lines

simon-id:
```suggestion
### PR Checklist

```

.github/pull_request_template.md
watson: WTF, I see you're right, but that's not what the `<ins>` HTML tag is normally used for 🤯 

image

simon-id: Yeah ins means "inserted", but don't ask me why github considers it underline :shrug:

.github/pull_request_template.md
watson: I agree people will find a place to put it regardless, but I think it's nice to have a place that's dedicated so the its information isn't accidentally mixed in with to the two other sections. E.g. I usually use the section for information relevant to the reviewer only. Recent examples:

- https://github.com/DataDog/dd-trace-js/pull/6912
- https://github.com/DataDog/dd-trace-js/pull/7028
- https://github.com/DataDog/dd-trace-js/pull/6958
- https://github.com/DataDog/dd-trace-js/pull/6415
- https://github.com/DataDog/dd-trace-js/pull/6782
- https://github.com/DataDog/dd-trace-js/pull/6877

.github/pull_request_template.md
simon-id: Google AI said "\<u>" also works, but i tried it actually doesn't -_- i love AI

CONTRIBUTING.md
watson: If it's meant to be copied it should be a code block, otherwise it's not possible. In its current state, it just shows disabled checkboxes that can't be checked

CONTRIBUTING.md
watson: Yeah I actually have PR in flight that removes those extra line breaks around the benchmark section. But feel free to fix it here instead

.github/pull_request_template.md
simon-id: Yes that makes sense, and as my previous message, that falls into the second case, if you have it, you can add `## additional info` yourself. I just feel like the template PR should be more about what sections are required, than what sections are possible.
We require people to add tests, to explain what a PR does and why it does. But having a forced section for additional info looks meh to me. Does that make sense ?

watson: No, my point was that people don't necessarily add it themselves. If there's a placeholder for it, it's much more likely to be used. You could use your argument for the two remaining headlines in your PR in which case we should just delete the template all together.

CONTRIBUTING.md
simon-id: It's a compromise between making it look good in markdown, and making the original copyable. Either you're reading the markdown in an editor and the original copyable is there, or you're on github interface and there is literally a button to switch to the original: 
image

.github/pull_request_template.md
watson: The official way to underline with GitHub markdown is `<ins>` so I say we stick with that - it was just a surprise to me.

.github/pull_request_template.md
simon-id: >You could use your argument for the two remaining headlines in your PR in which case we should just delete the template all together.

No because those we actually require for every PR.

>If there's a placeholder for it, it's much more likely to be used.

Granted. The only remaining question is what's the expected return of keeping it to exhort authors to use it, vs the annoyance of people having to remove it for all the small PRs that really don't need it. I think the expected return isn't that much, considering we don't have that many PRs from outsiders, and by definition small PRs where it's not needed should be more common than big PRs where it is. Buuuut I won't die on that hill so i'll let you decide. You may add a commit to add the section back if you think it's best. I don't mind :+1:

CONTRIBUTING.md
watson: I question the usefulness of this and would probably vote to just remove this section altogether

simon-id: I'm fine with removing it. It was a request from @bengl on my original slack thread.

.github/pull_request_template.md
watson: I question how big of an annoyance this actually is. And I think the benefits of having it outweighs the stated downsides. Especially in a world where AI is starting to make the first draft of a lot of these PR descriptions

simon-id: > Especially in a world where AI is starting to make the first draft of a lot of these PR descriptions

I am an agent of the resistance, and you can be too! One must try to live, instead of passively accepting his fate.

More seriously, go ahead and add it back if you're convinced. I'm not convinced, but I can live with it :smile:
I'm using the lazy-check technique here: if you're too lazy to add the commit back, then it's probably not that important to add it back hehe

watson: This section is actively being used by people in many PRs every week. I don't understand why that's not useful then - a lot of people seem to think it is. The `AGENTS.md` file [being added here](https://github.com/DataDog/dd-trace-js/pull/7109) is being instructed to look at the PR template when authoring PR bodies.

CONTRIBUTING.md
watson: I don't think he intended for people to click "edit source" to actually use it 😉

simon-id: i'll have to let him say what he intended himself, don't wanna put words into his mouth. Although, once again it's not clicking "edit source", it's clicking the file then the tab "code", which is available for all markdown files to see the original code instead of the markdown render.

.github/pull_request_template.md
simon-id: I'm still not convinced. I think it's unnecessary hand-holding. If a PR should have more informations, just add more informations, nothing stopping you from doing it. Unlike the other two sections, additional informations are optional. And this is why I think it's superfluous, even if I understand the argument about inviting authors to use it. Also, about your AGENTS.md argument, I'm not one to enjoy sacrificing human experience just because a fancy autocorrect is too stupid to do a good job.
Now I'll repeat what we can do:
1. you can add it back by making a commit on this PR, I won't object (if you don't wanna take the time to do it, it's probably not a very important section then).
2. Find other people that support your point, and I will fold to consensus, potentially in a guild meeting or something

BridgeAR (commented):
I have no strong opinion either way.

.github/pull_request_template.md
simon-id: Good news @watson we talked about this in the guild meeting and decided it wasn't worth our time, so i added it back hahhaa. Let's merge this

simon-id: Good news @watson we talked about this in the guild meeting and decided it wasn't worth our time, so i added it back hahhaa. Let's merge this

CONTRIBUTING.md
watson: The intent of this is not to show it nicely, but to make it easy to copy into your PR. So if we want to keep it (which I still doubt the usefulness of), it should be turned into a code block. If you show it "nicely", many users will just be very confused about why they can't click the check boxes.

simon-id: I removed it completely :+1: not sure it's useful either

PR #7115, opened 2025-12-15
Move vendored dependencies to non-standard folder

rochdev:
### What does this PR do?

Move vendored dependencies to non-standard folder.

### Motivation

Yarn v2+ deletes all `node_modules` folder in the entire tree of dependencies, making it impossible to write standard Node code.

### DIFF
--- a/.github/workflows/platform.yml
+++ b/.github/workflows/platform.yml
@@ -36,7 +36,7 @@ jobs:
           - name: yarn
             install: yarn add
           - name: yarn-berry
-            install: npm install -g corepack && yarn set version stable && yarn config set nodeLinker node-modules && yarn install
+            install: npm install -g corepack && yarn set version stable && yarn config set nodeLinker node-modules && yarn add
           - name: bun
             install: bun add
     runs-on: ubuntu-latest


BridgeAR (commented):
LGTM % linting issues

integration-tests/profiler/profiler.spec.js
```diff
@@ -14,7 +14,7 @@ const fs = require('fs/promises')
 const fsync = require('fs')
 const net = require('net')
 const zlib = require('zlib')
-const { Profile } = require('pprof-format')
+const { Profile } = require('../../vendor/dist/pprof-format')
```
watson, line 17: This is really annoying. I'm almost considering a build step just to not have to do this

BridgeAR: Agreed. I think we can look into this again at a later point

rochdev: > I'm almost considering a build step just to not have to do this

That should be how we do this anyway, but unfortunately it is not currently possible as the code is not modularized, so we can't bundle individual features properly, and bundling a single output is unacceptable because of Serverless.

PR #7119, opened 2025-12-16
feat(profiler): AsyncContextFrame used by profiler with Node.js 22+, on by default

szegedi:
### What does this PR do?
* Updates `pprof-nodejs` to 5.13.1, which allows enabling ACF sampling context already with Node.js 22
* Lowers the requirement to enable ACF sampling context to Node.js 22.9.0.
* When ACF is available on Linux and macOS, use it by default for profiling sampling context. In practical terms, this means either:
  * Node.js 24+ without explicit `--no-async-context-frame` command line switch, or
  * Node.js 22.9.0-23.99 with explicit `--experimental-async-context-frame` command line switch.

### Motivation
ACF works well starting with Node.js 22.9.0. Also, it is time we turn it on by default where possible, as it has lower overhead and has been tested for long time internally.

### DIFF
--- a/packages/dd-trace/src/profiling/config.js
+++ b/packages/dd-trace/src/profiling/config.js
@@ -229,10 +229,7 @@ class Config {
         canUseAsyncContextFrame = hasExecArg('--experimental-async-context-frame')
       }
     }
-    this.asyncContextFrameEnabled = isTrue(
-      options.useAsyncContextFrame ??
-      DD_PROFILING_ASYNC_CONTEXT_FRAME_ENABLED ?? canUseAsyncContextFrame
-    )
+    this.asyncContextFrameEnabled = isTrue(DD_PROFILING_ASYNC_CONTEXT_FRAME_ENABLED ?? canUseAsyncContextFrame)
     if (this.asyncContextFrameEnabled && !canUseAsyncContextFrame) {
       if (!samplingContextsAvailable) {
         turnOffAsyncContextFrame(`on ${process.platform}`)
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/profiling/config.js
```diff
@@ -221,20 +221,24 @@ class Config {
...
-      } else if (satisfies(process.versions.node, '>=23.0.0')) {
-        if (!hasExecArg('--experimental-async-context-frame')) {
-          turnOffAsyncContextFrame('without --experimental-async-context-frame')
-        }
+        canUseAsyncContextFrame = !hasExecArg('--no-async-context-frame')
+      } else if (satisfies(process.versions.node, '>=22.9.0')) {
+        canUseAsyncContextFrame = hasExecArg('--experimental-async-context-frame')
+      }
+    }
+    this.asyncContextFrameEnabled = isTrue(options.useAsyncContextFrame ?? DD_PROFILING_ASYNC_CONTEXT_FRAME_ENABLED ?? canUseAsyncContextFrame)
```
BridgeAR: `options.useAsyncContextFrame` is AFAIK coming from the user, while we have no public documentation of that config. So this is effectively never used, if I am not mistaken and we should likely either document it or remove it.

Would a manual override be something we really want anyway? I personally believe it would be best to just use it if we can and that is it?

szegedi: I agree; made a change to remove it.

nsavoire (approved):

PR #7120, opened 2025-12-16
// AI GENERATED: DO NOT MERGE, PURELY TO REVIEW CODE QUALITY- new nats integration

wconti27:
### What does this PR do?
// AI GENERATED PR

this pr is purely to discuss the quality of ai written integrations. please give any reviews, comments, etc. also feel free to reach out to @wconti27 with review comments.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/dd-trace/src/supported-configurations.json
tlhunter, line 53: is this a rebase issue?

packages/datadog-plugin-nats/test/index.spec.js
```diff
@@ -0,0 +1,279 @@
+'use strict'
+
+// ⚠️ MUST be set BEFORE any requires that initialize the tracer!
```
tlhunter, line 3: Might want to add "no emoji" to the prompt lol

packages/datadog-instrumentations/src/nats.js
```diff
@@ -0,0 +1,14 @@
...
+
+const {
+  addHook,
+  getHooks
+} = require('./helpers/instrument')
+
+// Register hooks for orchestrion (publish and processMsg)
+// NOTE: request() instrumentation is NOT supported due to NATS internal architecture.
+// Any wrapping of request() breaks the internal inbox/reply correlation mechanism.
+// This is a known limitation documented in the integration.
```
tlhunter, line 11: Is this referencing the three lines above it?

packages/datadog-plugin-nats/src/consumer.js
```diff
@@ -0,0 +1,100 @@
+'use strict'
+
+const ConsumerPlugin = require('../../dd-trace/src/plugins/consumer')
+const Extractors = require('./extractors')
+const { getSizeOrZero } = require('../../dd-trace/src/datastreams')
+
+class NatsConsumerPlugin extends ConsumerPlugin {
+  static get id () { return 'nats' }
+  static operation = 'processMsg'
```
tlhunter, line 9: Should use the same syntax for both these lines

tlhunter, line 9: Should `processMsg` be consume? I think we have common a convention for queues?

wconti27: we dont have any common conventions for operation naming.

wconti27: Good catch! will update the reviewer agent to catch these!

packages/datadog-instrumentations/src/nats.js
wconti27: This is really referencing this file: packages/datadog-instrumentations/src/helpers/rewriter/instrumentations/nats.json. and the fact that we tried to instrument `request` but failed for whatever reason. Will update the reviewer agent prompt again to ensure that we cleanup these comments.

packages/datadog-plugin-nats/src/consumer.js
```diff
@@ -0,0 +1,100 @@
...
+        this.tracer.decodeDataStreamsContext(plainHeaders)
+      }
+      const payloadSize = msg?.hdr > 0 ? getSizeOrZero(data?.subarray?.(msg.hdr)) : getSizeOrZero(data)
+      this.tracer.setCheckpoint(['direction:in', `topic:${subject}`, 'type:nats'], span, payloadSize)
+    }
+
+    return ctx.currentStore
+  }
+
+  end (ctx) {
```
wconti27, line 46: remove event functions if they match parent event functions.

packages/datadog-plugin-nats/src/extractors.js
```diff
@@ -0,0 +1,38 @@
...
+      'messaging.system': 'nats',
+      'messaging.destination.name': subject,
+      'messaging.operation': 'process'
+    }
+  }
+}
+
+module.exports = {
+  'tracing:orchestrion:nats:NatsConnectionImpl_publish': natsconnectionimplPublishProducer,
+  'tracing:orchestrion:nats:ProtocolHandler_processMsg': protocolhandlerProcessmsgConsumer
```
bengl, line 37: This whole approach is a bit bizarre to look at. The function names are very long, and the property names are channels.

It would make sense to include these directly in the producer and consumer files themselves, but I can't easily tell if both are used in both or not. I guess `registerOperation` does something here to select one of them? If that's the case, these should be directly inside the file that uses them. (I'm not sure, that's a new method to me.)

packages/datadog-plugin-nats/src/consumer.js
bengl, line 9: Here and throughout this PR, both a static getter and a static member are used. The static getter is a holdover from when static members weren't supported by the runtime. Today they are, so everything should be a static member.

packages/datadog-plugin-nats/src/consumer.js
```diff
@@ -0,0 +1,100 @@
...
+  constructor (...args) {
+    super(...args)
+    this.registerOperation('tracing:orchestrion:nats:ProtocolHandler_processMsg')
+  }
+
+  bindStart (ctx, channel) {
+    const options = Extractors[channel]?.(ctx)
+    if (!options) return ctx.currentStore
+
+    const msg = ctx.arguments?.[0]
```
bengl, line 20: In the start event, there's _always_ an arguments property on the context. There's no need to check for its presence. That's true here and throughout the PR.

packages/datadog-plugin-nats/src/consumer.js
```diff
@@ -0,0 +1,100 @@
...
+      if (line) {
+        const colonIdx = line.indexOf(':')
+        if (colonIdx > 0) {
+          headers[line.slice(0, colonIdx)] = line.slice(colonIdx + 1).trim()
+        }
+      }
+    }
+    return Object.keys(headers).length > 0 ? headers : null
+  } catch {
+    return null
```
bengl, line 88: This is a good candidate for logging.

packages/datadog-plugin-nats/src/consumer.js
```diff
@@ -0,0 +1,100 @@
+'use strict'
+
+const ConsumerPlugin = require('../../dd-trace/src/plugins/consumer')
+const Extractors = require('./extractors')
```
bengl, line 4: Things that aren't classes shouldn't have a leading capital letter. This is not the only place in this PR where this happens.

packages/datadog-instrumentations/src/nats.js
bengl, line 11: Documented where?

packages/datadog-plugin-nats/test/test-setup.js
```diff
@@ -0,0 +1,224 @@
...
+    // Set up responder BEFORE any tests run to ensure it's outside instrumentation context
+    this.requestResponder = this.nc.subscribe('time.request')
+    ;(async () => {
+      for await (const msg of this.requestResponder) {
+        msg.respond(this.sc.encode(JSON.stringify({ time: new Date().toISOString() })))
+      }
+    })()
+
+    // Send a test message to verify responder is working
+    await new Promise(resolve => setTimeout(resolve, 200))
```
bengl, line 24: No need for the `Promise` constructor here. Just use `node:timers/promises`.

Broadly though, since it doesn't look like subprocesses are being used, there's no need to wait or time anything. We should be able to do the appropriate things when they're ready, not after timeouts. Timeouts like this nearly always add flakiness.

packages/datadog-plugin-nats/test/test-setup.js
```diff
@@ -0,0 +1,224 @@
...
+        )
+        await Promise.race([
+          (async () => {
+            await this.nc.drain()
+            await this.nc.close()
+          })(),
+          timeout
+        ]).catch(() => {
+          // Ignore timeout errors
+        })
```
bengl, line 50: This catch call is unnecessary, since you're already catching errors on line 52.

packages/datadog-plugin-nats/test/test-setup.js
```diff
@@ -0,0 +1,224 @@
...
+      }
+    } catch (error) {
+      // Ignore errors during teardown
+    }
+  }
+
+  // --- Operations ---
+  async natsconnectionimpl_publish () {
```
bengl, line 58: None of these function names follow our coding style.

packages/dd-trace/src/supported-configurations.json
```diff
@@ -365,6 +365,7 @@
     "DD_TRACE_MYSQL_ENABLED": ["A"],
     "DD_TRACE_MYSQL2_ENABLED": ["A"],
     "DD_TRACE_NATIVE_SPAN_EVENTS": ["A"],
+    "DD_TRACE_NATS_ENABLED": ["A"],
```
bengl, line 368: This appears to be the only change in this file that has anything to do with this PR.

wconti27: explained over slack

PR #7121, opened 2025-12-16
refactor(remote-config): replace kPreUpdate with batch handler API (DEBUG-4402)

watson:
### What does this PR do?

- Removes the leaky `kPreUpdate` hook from `RemoteConfigManager` and replaces it with a first-class batch handler API (`setBatchHandler`/`removeBatchHandler`) that receives an explicit transaction (`ack`/`error`/`markHandled`).
- Introduces explicit RC product subscription via `subscribeProducts(...products)` / `unsubscribeProducts(...products)` and makes `setProductHandler` / `removeProductHandler` subscribe/unsubscribe for clarity.
- Wires AppSec/WAF RC updates to the new batch transaction API and centralizes ASM/WAF RC product names in `appsec/rc-products.js`.

### Motivation

`kPreUpdate` exposed internal mutable state and blurred responsibility between RC internals and consumers. WAF needs to reconcile multiple products in one logical update, so the new batch API provides a safer and clearer surface.

### DIFF
--- a/packages/dd-trace/src/appsec/rule_manager.js
+++ b/packages/dd-trace/src/appsec/rule_manager.js
@@ -16,7 +16,7 @@ let appliedActions = new Map()
 
 /**
  * @typedef {object} AsmConfigFile
- * @property {Array<Record<string, any>>} [actions]
+ * @property {Array<Record<string, unknown>>} [actions]
  */
 
 /**
@@ -93,7 +93,7 @@ function updateWafFromRC (transaction) {
 
   waf.checkAsmDdFallback()
 
-  if (wafUpdated && waf.wafManager) {
+  if (wafUpdated) {
     Reporter.reportWafUpdate(waf.wafManager.ddwafVersion, waf.wafManager.rulesVersion, !wafUpdatedFailed)
   }
 
--- a/packages/dd-trace/src/appsec/waf/index.js
+++ b/packages/dd-trace/src/appsec/waf/index.js
@@ -50,7 +50,7 @@ function init (rules, config) {
 
   limiter = new Limiter(config.rateLimit)
 
-  // dirty require to make startup faster for serverless
+  // Lazy loading improves the startup time
   WAFManager = require('./waf_manager')
 
   waf.wafManager = new WAFManager(rules, config)
@@ -111,7 +111,6 @@ function removeConfig (configPath) {
 }
 
 function run (data, req, raspRule) {
-  if (!waf.wafManager) return
   if (!req) {
     const store = storage('legacy').getStore()
     if (!store || !store.req) {
@@ -138,7 +137,6 @@ function run (data, req, raspRule) {
 }
 
 function disposeContext (req) {
-  if (!waf.wafManager) return
   const wafContext = waf.wafManager.getWAFContext(req)
 
   if (wafContext && !wafContext.ddwafContext.disposed) {
--- a/packages/dd-trace/src/remote_config/manager.js
+++ b/packages/dd-trace/src/remote_config/manager.js
@@ -264,14 +264,14 @@ class RemoteConfigManager {
     const toUnapply = /** @type {RcConfigState[]} */ ([])
     const toApply = /** @type {RcConfigState[]} */ ([])
     const toModify = /** @type {RcConfigState[]} */ ([])
-    const txByPath = new Map()
-    const txHandledPaths = new Set()
-    const txOutcomes = new Map()
+    const transactionByPath = new Map()
+    const transactionHandledPaths = new Set()
+    const transactionOutcomes = new Map()
 
     for (const appliedConfig of this.appliedConfigs.values()) {
       if (!clientConfigs.includes(appliedConfig.path)) {
         toUnapply.push(appliedConfig)
-        txByPath.set(appliedConfig.path, appliedConfig)
+        transactionByPath.set(appliedConfig.path, appliedConfig)
       }
     }
 
@@ -316,7 +316,7 @@ class RemoteConfigManager {
           hashes: meta.hashes,
           file: fromBase64JSON(file.raw)
         })
-        txByPath.set(path, newConf)
+        transactionByPath.set(path, newConf)
       }
 
       this.state.client.state.targets_version = targets.signed.version
@@ -324,22 +324,26 @@ class RemoteConfigManager {
     }
 
     if (toUnapply.length || toApply.length || toModify.length) {
-      const transaction = createUpdateTransaction({ toUnapply, toApply, toModify }, txHandledPaths, txOutcomes)
+      const transaction = createUpdateTransaction(
+        { toUnapply, toApply, toModify },
+        transactionHandledPaths,
+        transactionOutcomes
+      )
 
       if (this.#batchHandlers.size) {
         for (const [handler, products] of this.#batchHandlers) {
-          const txView = filterTransactionByProducts(transaction, products)
-          if (txView.toUnapply.length || txView.toApply.length || txView.toModify.length) {
-            handler(txView)
+          const transactionView = filterTransactionByProducts(transaction, products)
+          if (transactionView.toUnapply.length || transactionView.toApply.length || transactionView.toModify.length) {
+            handler(transactionView)
           }
         }
       }
 
-      applyOutcomes(txByPath, txOutcomes)
+      applyOutcomes(transactionByPath, transactionOutcomes)
 
-      this.dispatch(toUnapply, 'unapply', txHandledPaths)
-      this.dispatch(toApply, 'apply', txHandledPaths)
-      this.dispatch(toModify, 'modify', txHandledPaths)
+      this.dispatch(toUnapply, 'unapply', transactionHandledPaths)
+      this.dispatch(toApply, 'apply', transactionHandledPaths)
+      this.dispatch(toModify, 'modify', transactionHandledPaths)
 
       this.state.cached_target_files = /** @type {RcCachedTargetFile[]} */ ([])
 
@@ -368,7 +372,7 @@ class RemoteConfigManager {
   dispatch (list, action, handledPaths) {
     for (const item of list) {
       if (!handledPaths.has(item.path)) {
-        this._callHandlerFor(action, item)
+        this.#callHandlerFor(action, item)
       }
 
       if (action === 'unapply') {
@@ -383,7 +387,7 @@ class RemoteConfigManager {
    * @param {'apply' | 'modify' | 'unapply'} action
    * @param {RcConfigState} item
    */
-  _callHandlerFor (action, item) {
+  #callHandlerFor (action, item) {
     // in case the item was already handled by a batch hook
     if (item.apply_state !== UNACKNOWLEDGED && action !== 'unapply') return
 
@@ -453,22 +457,13 @@ class RemoteConfigManager {
  */
 
 /**
- * @typedef {Object} RcConfigDescriptor
- * @property {string} path
- * @property {string} product
- * @property {string} id
- * @property {number} version
- * @property {unknown} file
- */
-
-/**
  * Remote Config batch update transaction passed to batch handlers registered via
  * `RemoteConfigManager.setBatchHandler()`.
  *
  * @typedef {Object} RcBatchUpdateTransaction
- * @property {RcConfigDescriptor[]} toUnapply
- * @property {RcConfigDescriptor[]} toApply
- * @property {RcConfigDescriptor[]} toModify
+ * @property {RcConfigState[]} toUnapply
+ * @property {RcConfigState[]} toApply
+ * @property {RcConfigState[]} toModify
  * @property {(path: string) => void} ack
  * @property {(path: string, err: unknown) => void} error
  */
@@ -483,9 +478,9 @@ class RemoteConfigManager {
  */
 function createUpdateTransaction ({ toUnapply, toApply, toModify }, handledPaths, outcomes) {
   return {
-    toUnapply: toUnapply.map(toDescriptor),
-    toApply: toApply.map(toDescriptor),
-    toModify: toModify.map(toDescriptor),
+    toUnapply,
+    toApply,
+    toModify,
     ack (path) {
       if (typeof path !== 'string') return
       outcomes.set(path, { state: ACKNOWLEDGED, error: '' })
@@ -533,26 +528,13 @@ function filterTransactionByProducts (transaction, products) {
   }
 }
 
-/**
- * @param {RcConfigState} conf
- * @returns {RcConfigDescriptor}
- */
-function toDescriptor (conf) {
-  return {
-    path: conf.path,
-    product: conf.product,
-    id: conf.id,
-    version: conf.version,
-    file: conf.file
-  }
-}
-
 function applyOutcomes (byPath, outcomes) {
   for (const [path, outcome] of outcomes) {
     const item = byPath.get(path)
-    if (!item) continue
-    item.apply_state = outcome.state
-    item.apply_error = outcome.error
+    if (item) {
+      item.apply_state = outcome.state
+      item.apply_error = outcome.error
+    }
   }
 }
 
--- a/packages/dd-trace/test/remote_config/manager.spec.js
+++ b/packages/dd-trace/test/remote_config/manager.spec.js
@@ -360,8 +360,7 @@ describe('RemoteConfigManager', () => {
 
     it('should do nothing if passed an empty payload', () => {
       payload = {}
-
-      assert.doesNotThrow(parsePayload)
+      parsePayload()
       sinon.assert.notCalled(rc.dispatch)
       assert.deepStrictEqual(rc.state, previousState)
     })


watson:
* **#7121** <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-js/7121?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-js/7121?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/dd-trace/src/appsec/rule_manager.js
```diff
@@ -1,82 +1,90 @@
...
+const { ASM_WAF_PRODUCTS_SET } = require('./rc-products')
 
 /*
   ASM Actions must be tracked in order to update the defaultBlockingActions in blocking. These actions are used
   by blockRequest method exposed in the user blocking SDK (see packages/dd-trace/src/appsec/sdk/user_blocking.js)
  */
 let appliedActions = new Map()
 
+/**
+ * @param {{ rules?: string, [key: string]: any }} config
```
BridgeAR: Nit: I see the improvement, while I would like to prohibit adding any `any` to the code base

packages/dd-trace/src/appsec/rule_manager.js
```diff
@@ -1,82 +1,90 @@
...
   for (const item of toUnapply) {
-    if (!ASM_PRODUCTS.has(item.product)) continue
+    if (!ASM_WAF_PRODUCTS_SET.has(item.product)) continue
 
     try {
       waf.removeConfig(item.path)
 
-      item.apply_state = ACKNOWLEDGED
+      tx.ack(item.path)
+      tx.markHandled(item.path)
```
BridgeAR: I think we could just implicitly mark it as handled when there is an ack or an error. That way the code is simpler.

packages/dd-trace/src/remote_config/manager.js
```diff
@@ -326,6 +424,150 @@ class RemoteConfigManager extends EventEmitter {
...
+ */
+function toDescriptor (conf) {
+  const desc = {
+    path: conf.path,
+    product: conf.product,
+    id: conf.id,
+    version: conf.version,
+    file: conf.file
+  }
+  return Object.freeze(desc)
```
BridgeAR: Do we need to freeze the objects?

packages/dd-trace/src/remote_config/manager.js
```diff
@@ -238,20 +313,32 @@ class RemoteConfigManager extends EventEmitter {
...
+        txByPath.set(path, newConf)
       }
 
       this.state.client.state.targets_version = targets.signed.version
       this.state.client.state.backend_client_state = targets.signed.custom.opaque_backend_state
     }
 
     if (toUnapply.length || toApply.length || toModify.length) {
-      this.emit(RemoteConfigManager.kPreUpdate, { toUnapply, toApply, toModify })
+      const tx = createUpdateTransaction({ toUnapply, toApply, toModify }, txHandledPaths, txOutcomes)
```
BridgeAR: I would strongly favor writing transaction instead of tx.

packages/dd-trace/src/remote_config/manager.js
```diff
@@ -49,6 +43,8 @@ class RemoteConfigManager extends EventEmitter {
       : config.tags
 
     this._handlers = new Map()
+    this._products = new Set()
+    this._batchHandlers = new Map()
```
BridgeAR: Please use private properties instead

packages/dd-trace/src/appsec/rule_manager.js
watson: Fixed 👍

packages/dd-trace/src/appsec/rule_manager.js
watson: Fixed 👍

packages/dd-trace/src/remote_config/manager.js
watson: No, I guess it's not required. I removed it 👍

packages/dd-trace/src/remote_config/manager.js
watson: Ok. Renamed across entire PR 👍

packages/dd-trace/src/remote_config/manager.js
watson: Fixed 👍

packages/dd-trace/src/appsec/rule_manager.js
```diff
@@ -1,82 +1,99 @@
...
 
 /*
   ASM Actions must be tracked in order to update the defaultBlockingActions in blocking. These actions are used
   by blockRequest method exposed in the user blocking SDK (see packages/dd-trace/src/appsec/sdk/user_blocking.js)
  */
 let appliedActions = new Map()
 
+/**
+ * @typedef {object} AsmConfigFile
+ * @property {Array<Record<string, any>>} [actions]
```
BridgeAR:
```suggestion
 * @property {Array<Record<string, unknown>>} [actions]
```

Do you mind opening a separate PR (I am also fine to include it here as separate commit) that updates Claude.md to never use `any` and instead tries to infer the type as detailed as possible? We should probably also update the Contributing.md accordingly.

packages/dd-trace/src/appsec/rule_manager.js
```diff
@@ -1,82 +1,99 @@
...
+      const error = e instanceof waf.WafUpdateError ? JSON.stringify(extractErrors(e.diagnosticErrors)) : e
+      transaction.error(item.path, error)
       wafUpdatedFailed = true
     }
   }
 
   waf.checkAsmDdFallback()
 
-  if (wafUpdated) {
+  if (wafUpdated && waf.wafManager) {
```
BridgeAR: `wafUpdated` would only ever be truthy in case waf.wafManager is truthy as well. I believe this is a type fix? I think it would be nice to just refactor the code in a way that code changes like these would not be necessary anymore. What do you think? Should we postpone this as such?

packages/dd-trace/src/appsec/waf/index.js
```diff
@@ -31,13 +41,17 @@ const waf = {
...
+/**
+ * @param {object} rules
+ * @param {WAFConfig} config
+ */
 function init (rules, config) {
   destroy()
 
   limiter = new Limiter(config.rateLimit)
 
   // dirty require to make startup faster for serverless
```
BridgeAR: Independent wish:
```suggestion
  // Lazy loading improves the startup time
```

packages/dd-trace/src/appsec/waf/index.js
```diff
@@ -97,6 +111,7 @@ function removeConfig (configPath) {
 }
 
 function run (data, req, raspRule) {
+  if (!waf.wafManager) return
```
BridgeAR: Similar to the comment above about code changes for type fixes. I think instead of adding the safeguard here for something that is impossible to be falsy is not ideal and refactoring the code differently later would be better instead.

packages/dd-trace/src/appsec/waf/index.js
```diff
@@ -123,6 +138,7 @@ function run (data, req, raspRule) {
 }
 
 function disposeContext (req) {
+  if (!waf.wafManager) return
```
BridgeAR: Similar to the comment above about code changes for type fixes. I think instead of adding the safeguard here for something that is impossible to be falsy is not ideal and refactoring the code differently later would be better instead.

packages/dd-trace/test/remote_config/manager.spec.js
```diff
@@ -360,7 +361,7 @@ describe('RemoteConfigManager', () => {
     it('should do nothing if passed an empty payload', () => {
       payload = {}
 
-      assert.doesNotThrow(func)
+      assert.doesNotThrow(parsePayload)
```
BridgeAR: Nit while being on it (I guess a separate PR might actually be better for that, where we also update Claude.md and the linter):

```suggestion
      parsePayload()
```

packages/dd-trace/src/remote_config/manager.js
```diff
@@ -189,13 +261,17 @@ class RemoteConfigManager extends EventEmitter {
...
   }) {
-    const toUnapply = []
-    const toApply = []
-    const toModify = []
+    const toUnapply = /** @type {RcConfigState[]} */ ([])
+    const toApply = /** @type {RcConfigState[]} */ ([])
+    const toModify = /** @type {RcConfigState[]} */ ([])
+    const txByPath = new Map()
+    const txHandledPaths = new Set()
+    const txOutcomes = new Map()
```
BridgeAR: Do you mind renaming these to transaction as well? :)

packages/dd-trace/src/remote_config/manager.js
```diff
@@ -115,24 +116,95 @@ class RemoteConfigManager extends EventEmitter {
...
+    if (hadProducts && this.#products.size === 0) {
       this.scheduler.stop()
     }
   }
 
   updateProducts () {
-    this.state.client.products = [...this._handlers.keys()]
+    this.state.client.products = [...this.#products]
+  }
+
```
BridgeAR, line 186: I believe this is dead code and only used in tests without actually doing anything.

packages/dd-trace/src/remote_config/manager.js
```diff
@@ -240,20 +316,32 @@ class RemoteConfigManager extends EventEmitter {
...
+      }
+
+      applyOutcomes(txByPath, txOutcomes)
 
-      this.dispatch(toUnapply, 'unapply')
-      this.dispatch(toApply, 'apply')
-      this.dispatch(toModify, 'modify')
+      this.dispatch(toUnapply, 'unapply', txHandledPaths)
+      this.dispatch(toApply, 'apply', txHandledPaths)
+      this.dispatch(toModify, 'modify', txHandledPaths)
```
BridgeAR: Do we always have to dispatch these, even if there are no changes?

packages/dd-trace/src/remote_config/manager.js
```diff
@@ -269,12 +357,19 @@ class RemoteConfigManager extends EventEmitter {
...
+   * @param {Set<string>} handledPaths
+   */
+  dispatch (list, action, handledPaths) {
     for (const item of list) {
-      // TODO: we need a way to tell if unapply configs were handled by kPreUpdate or not, because they're always
-      // emitted unlike the apply and modify configs
-
-      this._callHandlerFor(action, item)
+      if (!handledPaths.has(item.path)) {
+        this._callHandlerFor(action, item)
```
BridgeAR: Is it possible to make this a private property as well? :)

packages/dd-trace/src/remote_config/manager.js
```diff
@@ -328,6 +427,135 @@ class RemoteConfigManager extends EventEmitter {
...
+    file: conf.file
+  }
+}
+
+function applyOutcomes (byPath, outcomes) {
+  for (const [path, outcome] of outcomes) {
+    const item = byPath.get(path)
+    if (!item) continue
+    item.apply_state = outcome.state
+    item.apply_error = outcome.error
```
BridgeAR: Nit (I know this is more indentation, while I feel it is easier to follow)

```suggestion
    if (item) {
      item.apply_state = outcome.state
      item.apply_error = outcome.error
    }
```

packages/dd-trace/src/remote_config/manager.js
```diff
@@ -240,20 +316,32 @@ class RemoteConfigManager extends EventEmitter {
...
       this.state.client.state.backend_client_state = targets.signed.custom.opaque_backend_state
     }
 
     if (toUnapply.length || toApply.length || toModify.length) {
-      this.emit(RemoteConfigManager.kPreUpdate, { toUnapply, toApply, toModify })
+      const transaction = createUpdateTransaction({ toUnapply, toApply, toModify }, txHandledPaths, txOutcomes)
+
+      if (this.#batchHandlers.size) {
+        for (const [handler, products] of this.#batchHandlers) {
+          const txView = filterTransactionByProducts(transaction, products)
```
BridgeAR: Do you mind renaming this to transaction as well? :)

packages/dd-trace/src/remote_config/manager.js
```diff
@@ -328,6 +427,135 @@ class RemoteConfigManager extends EventEmitter {
...
+function toDescriptor (conf) {
+  return {
+    path: conf.path,
+    product: conf.product,
+    id: conf.id,
+    version: conf.version,
+    file: conf.file
+  }
+}
+
```
BridgeAR: Is the filtering of the properties really needed? I understand that this limits the surface for us where we do not need other properties, while I feel it is mostly additional CPU work.

packages/dd-trace/src/appsec/rule_manager.js
watson: 👍 I'll open another PR to update `AGENTS.md` and `CONTRIBUTING.md`

packages/dd-trace/src/appsec/rule_manager.js
watson: Yes this is a type fix. For the sake of scope, I'll remove it now, leaving TS to complain and we can fix properly later

packages/dd-trace/src/appsec/waf/index.js
watson: Yes this is a type fix. For the sake of scope, I'll remove it now, leaving TS to complain and we can fix properly later

packages/dd-trace/src/appsec/waf/index.js
watson: Yes this is a type fix. For the sake of scope, I'll remove it now, leaving TS to complain and we can fix properly later

packages/dd-trace/src/remote_config/manager.js
watson: I felt it was important to not expose the internal structure, but make a clear API. This isn't a hot code path, so I felt it was better to make a cleaner API like this. Do you think the additional work isn't worth it?

packages/dd-trace/src/remote_config/manager.js
watson: Are you referring to the `updateProducts` function, or the content of the `state.client.products` array?

watson: The `updateProducts` function is called by both `subscribeProducts` and `unsubscribeProducts` and the `products` array is sent to the `/v0.7/config` endpoint. So unless I'm missing something, it's all used

packages/dd-trace/src/remote_config/manager.js
watson: Strictly speaking, no. But inside of `dispatch` we iterate over the provided array, so the `dispatch` function doesn't do anything if the array is empty. I could guard each of these 3 calls with an `if (<arr>.length)`, but I felt it was just making the code more complex, so I wrote it as so. WDYT?

packages/dd-trace/src/appsec/rule_manager.js
watson: Fixed 👍

packages/dd-trace/src/appsec/rule_manager.js
watson: Fixed 👍

packages/dd-trace/src/appsec/waf/index.js
watson: Updated 👍

packages/dd-trace/src/appsec/waf/index.js
watson: Fixed 👍

packages/dd-trace/src/appsec/waf/index.js
watson: Fixed 👍

packages/dd-trace/test/remote_config/manager.spec.js
watson: Fixed 👍

packages/dd-trace/src/remote_config/manager.js
watson: Fixed 👍

packages/dd-trace/src/remote_config/manager.js
watson: Fixed 👍

packages/dd-trace/src/remote_config/manager.js
watson: Fixed 👍

packages/dd-trace/src/remote_config/manager.js
watson: Fixed 👍

packages/dd-trace/src/remote_config/manager.js
BridgeAR: I guess it is fine, since it is no hot code path

packages/dd-trace/src/appsec/rule_manager.js
watson: Here's the separate PR to update both `AGENTS.md` and `CONTRIBUTING.md`: https://github.com/DataDog/dd-trace-js/pull/7182

packages/dd-trace/src/remote_config/manager.js
BridgeAR: `updateProducts` is called, correct. But the state defined is not used anymore, if I am not mistaken. Thus, I believe it not needed anymore including the call sites.

packages/dd-trace/src/remote_config/manager.js
BridgeAR: I already expected the wish for the clear API, I just think if this is ever used, it would also only ever be used with properties that are needed to be accessed. Thus, I feel the extra step here is not needed and I would remove it as it is also an extra step to follow when understanding the code.

packages/dd-trace/src/remote_config/manager.js
watson: "The state defined"? Are you referring to the `state.client.products` array being set inside this function? That is stringified inside the `getPayload` function which is used to generate the payload for the request to `/v0.7/config` in the `poll` function, which is called periodically (by default every 5 seconds).

packages/dd-trace/src/remote_config/manager.js
watson: Removed 👍

packages/dd-trace/src/remote_config/manager.js
```diff
@@ -328,6 +431,113 @@ class RemoteConfigManager extends EventEmitter {
...
+    toUnapply,
+    toApply,
+    toModify,
+    ack (path) {
+      if (typeof path !== 'string') return
+      outcomes.set(path, { state: ACKNOWLEDGED, error: '' })
+      handledPaths.add(path)
+    },
+    error (path, err) {
+      if (typeof path !== 'string') return
```
BridgeAR, line 490: I didn't check, is that actually possible to have an invalid path?

BridgeAR (approved):
LGTM, just left a question

packages/dd-trace/src/remote_config/manager.js
watson: You're right, unless the API is misused, these should both always be strings. I'll remove these if-statements in the next PR in this stack

watson: Fixed in https://github.com/DataDog/dd-trace-js/pull/7122

PR #7122, opened 2025-12-16
refactor(remote_config): move appsec-specific logic to appsec folder

watson:
## What Does This Do?

Resolves the TODO in `remote_config/index.js` by moving all appsec-specific remote configuration logic into the appsec folder, improving separation of concerns and module independence.

## Motivation

The `remote_config` module contained appsec-specific code including:
- WAF update capabilities and subscriptions
- ASM_FEATURES product handlers
- User tracking configuration
- Appsec enable/disable logic via remote config

This created tight coupling between core remote config and appsec, making both modules harder to understand and maintain. The `remote_config` module should focus on core APM tracing capabilities and be agnostic to appsec.

## Benefits

- **Clear ownership**: All appsec functionality lives in the appsec folder
- **Reduced coupling**: `remote_config` has zero knowledge of appsec
- **Better testability**: Each module tested independently with proper mocks
- **Improved maintainability**: Easier to understand each module's scope
- **Preserved optimizations**: Lazy loading still works for serverless cold starts

### DIFF
--- a/packages/dd-trace/src/appsec/remote_config.js
+++ b/packages/dd-trace/src/appsec/remote_config.js
@@ -11,13 +11,12 @@ let rc
 
 /**
  * Configures remote config handlers for appsec features
- *
  * @param {Object} rcInstance - RemoteConfigManager instance
+ *
  * @param {Object} config - Tracer config
- * @param {(config: Object, rc: Object) => void} enable - Function to enable appsec
- * @param {() => void} disable - Function to disable appsec
+ * @param {Object} appsec - Appsec module
  */
-function enable (rcInstance, config, enable, disable) {
+function enable (rcInstance, config, appsec) {
   rc = rcInstance
   const activation = Activation.fromConfig(config)
 
@@ -50,7 +49,7 @@ function enable (rcInstance, config, enable, disable) {
       }
 
       if (activation === Activation.ONECLICK) {
-        enableOrDisableAppsec(action, rcConfig, config, enable, disable)
+        enableOrDisableAppsec(action, rcConfig, config, appsec)
       }
     })
   }
@@ -62,10 +61,9 @@ function enable (rcInstance, config, enable, disable) {
  * @param {string} action - 'apply', 'modify', or 'unapply'
  * @param {Object} rcConfig - Remote config
  * @param {Object} config - Tracer config
- * @param {(config: Object, rc: Object) => void} enable - Function to enable appsec
- * @param {() => void} disable - Function to disable appsec
+ * @param {Object} appsec - Appsec module
  */
-function enableOrDisableAppsec (action, rcConfig, config, enable, disable) {
+function enableOrDisableAppsec (action, rcConfig, config, appsec) {
   if (typeof rcConfig.asm?.enabled === 'boolean') {
     const isRemoteConfigControlling = action === 'apply' || action === 'modify'
     const shouldEnable = isRemoteConfigControlling
@@ -73,9 +71,9 @@ function enableOrDisableAppsec (action, rcConfig, config, enable, disable) {
       : config.appsec.enabled // give back control to local config
 
     if (shouldEnable) {
-      enable(config, rc)
+      appsec.enable(config)
     } else {
-      disable()
+      appsec.disable()
     }
 
     updateConfig([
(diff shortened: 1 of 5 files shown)


packages/dd-trace/test/appsec/remote_config.spec.js
```diff
@@ -0,0 +1,371 @@
...
+
+      describe('update config origin activation', () => {
+        const rcConfigAsmEnabling = { asm: { enabled: true } }
+        const rcConfigAsmDisabling = { asm: { enabled: true } }
+
+        it('should update appsec.enabled when applying asm enabling by RC', () => {
+          listener('apply', rcConfigAsmEnabling)
+
+          sinon.assert.calledOnce(telemetry.updateConfig)
+          assert.strictEqual(telemetry.updateConfig.firstCall.args[0][0].name, 'appsec.enabled')
```
BridgeAR: Nit: we could use `assertObjectContains(telemetry.updateConfig.firstCall.args, [[{ ... }]])` instead. The same in the next test

watson: Fixed 👍

packages/dd-trace/src/appsec/remote_config.js
```diff
@@ -0,0 +1,177 @@
...
+      // this is put before other handlers because it can reject the config
+      if (typeof rcConfig.auto_user_instrum?.mode === 'string') {
+        if (action === 'apply' || action === 'modify') {
+          // check if there is already a config applied with this field
+          if (autoUserInstrumModeId && configId !== autoUserInstrumModeId) {
+            log.error('[RC] Multiple auto_user_instrum received in ASM_FEATURES. Discarding config')
+            // eslint-disable-next-line no-throw-literal
+            throw 'Multiple auto_user_instrum.mode received in ASM_FEATURES'
+          }
+
```
BridgeAR, line 42: In what circumstances could something like that be true? Should we really guard against AFAIK programming errors like these?

packages/dd-trace/src/proxy.js
```diff
@@ -156,6 +156,11 @@ class Tracer extends NoopProxy {
           this._flare.module.send(conf.args)
         })
 
+        if (this._modules.appsec) {
+          const appsecRemoteConfig = require('./appsec/remote_config')
+          appsecRemoteConfig.enable(rc, config, this._modules.appsec)
+        }
```
BridgeAR: I feel this would better be placed in the appsec enable part.

We would just pass through rc to the `.enable()` call. That way the receiver is able to activate RC, if needed instead of having all that logic placed inside of proxy.

packages/dd-trace/src/appsec/remote_config.js
watson: Good question. I just moved the existing code, so I feel that's out of scope for this PR?

BridgeAR: It is indeed nothing new, I just feel we could clean this up at the same time or open a separate PR afterwards (while I favor just removing it here), if it is indeed not needed. It throws a string that would prevent appsec from being loaded.

@simon-id PTAL

packages/dd-trace/src/proxy.js
watson: The place [where we call the appsec `enable` function](https://github.com/DataDog/dd-trace-js/blob/7a2b4276f4b9293fb2b93909b1c86ed5592ad992/packages/dd-trace/src/proxy.js#L254), we don't have a reference to `rc`. So it would be a fairly complex change (if I understand your suggestion correctly).

packages/dd-trace/src/appsec/remote_config.js
watson: I prefer to do something like this in a separate PR. Just in case we're wrong and might want to revert it later, it's easier to have this land a separate commit. Also keeps the discussion focused.

packages/dd-trace/src/proxy.js
watson: Fixed 👍

BridgeAR (dismissed):
LGTM

packages/dd-trace/src/proxy.js
watson: Had to revert these changes because they introduced a bug: With the changes we would only enable appsec RC if appsec itself is enabled. However, we still need to advertise the appsec RC capabilities even if appsec isn't enabled, as it can also be enabled remotely via RC

BridgeAR (approved):
LGTM

PR #7127, opened 2025-12-17
refactor(remote_config): move capability registration to feature modules

watson:
### What does this PR do?

Refactors remote config capability and handler registration to improve separation of concerns. Moves feature-specific remote config setup from a central location into their respective feature modules, following the same pattern already used by AppSec.

**Key changes:**
- Creates `config/remote_config.js` for client library config capabilities (`APM_TRACING_*`)
- Creates `openfeature/remote_config.js` for OpenFeature capabilities (`FFE_FLAG_CONFIGURATION_RULES`)
- Renames `RemoteConfigManager` → `RemoteConfig` and `remote_config/manager.js` → `remote_config/index.js`
- Simplifies `proxy.js` to instantiate `RemoteConfig` directly instead of through wrapper function

### Motivation

This refactoring establishes a clear pattern where each feature manages its own remote config capabilities and handlers together:
- **Core features** (like APM tracing) → always registered
- **Optional features** (like AppSec) → registered only when feature is enabled

OpenFeature is an exception to this pattern: it always registers the `FFE_FLAG_CONFIGURATION_RULES` capability (required by parametric tests since v5.72.0), but only registers its handler when `config.experimental.flaggingProvider.enabled` is true. This hybrid approach satisfies test requirements while keeping the experimental feature properly gated.

### Additional Notes

**Reviewing this PR:**
The diff is large but mostly consists of code being moved between files. To review more easily:

1. **File renames** (no logic changes, just find/replace):
   - `remote_config/manager.js` → `remote_config/index.js`
   - `test/remote_config/manager.spec.js` → `test/remote_config/index.spec.js`
   - Class `RemoteConfigManager` → `RemoteConfig` (throughout)

2. **New files with moved code**:
   - `config/remote_config.js` - client library config capabilities moved from old `remote_config/index.js` + `APM_TRACING` handler moved from `proxy.js`
   - `openfeature/remote_config.js` - FFE capability + handler moved from `proxy.js`

3. **Modified files**:
   - `proxy.js` - Removed old wrapper, now instantiates `RemoteConfig` directly; handlers moved to feature modules
   - `test/proxy.spec.js` - Updated imports and assertions to match new structure

4. **Assertion lib**:
   - Replaced `chai` with `node:assert/strict` in remote_config tests for consistency

No functional changes to remote config behavior, purely structural refactoring.

### DIFF
--- a/packages/dd-trace/src/remote_config/index.js
+++ b/packages/dd-trace/src/remote_config/index.js
@@ -207,6 +207,7 @@ class RemoteConfig {
     this.#batchHandlers.delete(handler)
   }
 
+  // TODO: Cache the return value of this method, to avoid recomputing it on every poll
   getPayload () {
     this.state.client.client_tracer.extra_services = getExtraServices()
 
--- a/packages/dd-trace/test/config/remote_config.spec.js
+++ b/packages/dd-trace/test/config/remote_config.spec.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { describe, it, beforeEach } = require('tap').mocha
+const { describe, it, beforeEach } = require('mocha')
 const sinon = require('sinon')
 
 const RemoteConfigCapabilities = require('../../src/remote_config/capabilities')
@@ -59,7 +59,7 @@ describe('Tracing Remote Config', () => {
         handler('apply', { lib_config: libConfig })
 
         sinon.assert.calledOnceWithExactly(config.configure, libConfig, true)
-        sinon.assert.calledOnceWithExactly(enableOrDisableTracing, config)
+        sinon.assert.calledOnceWithExactly(enableOrDisableTracing, config, rc)
       })
 
       it('should reset config on unapply action', () => {
@@ -70,7 +70,7 @@ describe('Tracing Remote Config', () => {
         handler('unapply', {})
 
         sinon.assert.calledOnceWithExactly(config.configure, {}, true)
-        sinon.assert.calledOnceWithExactly(enableOrDisableTracing, config)
+        sinon.assert.calledOnceWithExactly(enableOrDisableTracing, config, rc)
       })
     })
   })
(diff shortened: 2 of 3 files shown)


packages/dd-trace/src/remote_config/index.js
```diff
@@ -1,28 +1,610 @@
...
+      ? {
+          ...config.tags,
+          [GIT_REPOSITORY_URL]: config.repositoryUrl,
+          [GIT_COMMIT_SHA]: config.commitSHA
+        }
+      : config.tags
+
+    this._handlers = new Map()
+    this._products = new Set()
+    this._batchHandlers = new Map()
```
BridgeAR: Similar as in the main PR: using private properties would be nice :)

packages/dd-trace/src/remote_config/index.js
```diff
@@ -1,28 +1,610 @@
...
+    const options = {
+      url: this.url,
+      method: 'POST',
+      path: '/v0.7/config',
+      headers: {
+        'Content-Type': 'application/json; charset=utf-8'
+      }
+    }
+
+    request(this.getPayload(), options, (err, data, statusCode) => {
```
BridgeAR, line 227: Should we cache this already? Potentially add a TODO entry for that?

packages/dd-trace/test/config/remote_config.spec.js
```diff
@@ -0,0 +1,77 @@
+'use strict'
+
+const { describe, it, beforeEach } = require('tap').mocha
```
BridgeAR:
```suggestion
const { describe, it, beforeEach } = require('mocha')
```

BridgeAR: Is this file picked up? I can not see the related test failure

watson: Good catch. Seems like `packages/dd-trace/test/config/disabled_instrumentations.spec.js` is actually not being run either

watson: I'll add them to the `test:trace:core` script

packages/dd-trace/src/remote_config/index.js
watson: Do you mean cache the result of `this.getPayload()`, or? Then we need to deal with cache invalidation as it can change from request to request, but it's doable. But I would say outside the scope of this PR as this code is just moved from the old `packages/dd-trace/src/remote_config/manager.js` file. I've added a TODO for now

packages/dd-trace/test/config/remote_config.spec.js
watson: Fixed 👍

PR #7132, opened 2025-12-17
[DI] add process tags to dynamic instrumentation

tlhunter:
### What does this PR do?
- adds process tags to dynamic instrumentation
  - this is additional context about the currently running process
- RFC https://docs.google.com/document/d/1AFdLUuVk70i0bJd5335-RxqsvwAV9ovAqcO2z5mEMbA

### Motivation
- help with process renaming

### DIFF
--- a/integration-tests/debugger/basic.spec.js
+++ b/integration-tests/debugger/basic.spec.js
@@ -800,6 +800,58 @@ describe('Dynamic Instrumentation', function () {
       )
     })
   })
+
+  describe('DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED=true', function () {
+    const t = setup({
+      env: { DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED: 'true' },
+      dependencies: ['fastify']
+    })
+
+    describe('input messages', function () {
+      it('should include process_tags in snapshot when enabled', function (done) {
+        t.agent.on('debugger-input', ({ payload }) => {
+          const snapshot = payload[0].debugger.snapshot
+
+          // Assert that process_tags are present
+          assert.ok(snapshot.process_tags)
+          assert.strictEqual(typeof snapshot.process_tags, 'object')
+
+          // Check for expected process tags keys
+          assert.ok(snapshot.process_tags['entrypoint.name'])
+          assert.ok(snapshot.process_tags['entrypoint.type'])
+          assert.strictEqual(snapshot.process_tags['entrypoint.type'], 'script')
+
+          done()
+        })
+
+        t.triggerBreakpoint()
+        t.agent.addRemoteConfig(t.rcConfig)
+      })
+    })
+  })
+
+  describe('DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED=false', function () {
+    const t = setup({
+      env: { DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED: 'false' },
+      dependencies: ['fastify']
+    })
+
+    describe('input messages', function () {
+      it('should not include process_tags in snapshot when disabled', function (done) {
+        t.agent.on('debugger-input', ({ payload }) => {
+          const snapshot = payload[0].debugger.snapshot
+
+          // Assert that process_tags are not present
+          assert.strictEqual(snapshot.process_tags, undefined)
+
+          done()
+        })
+
+        t.triggerBreakpoint()
+        t.agent.addRemoteConfig(t.rcConfig)
+      })
+    })
+  })
 })
 
 function testBasicInput (t, done) {
--- a/packages/dd-trace/src/debugger/devtools_client/index.js
+++ b/packages/dd-trace/src/debugger/devtools_client/index.js
@@ -12,6 +12,7 @@ const { MAX_SNAPSHOTS_PER_SECOND_GLOBALLY } = require('./defaults')
 const log = require('./log')
 const { version } = require('../../../../../package.json')
 const { NODE_MAJOR } = require('../../../../../version')
+const processTags = require('../../process-tags')
 
 require('./remote_config')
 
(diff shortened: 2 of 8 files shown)


watson:
@tlhunter this PR could do with a better description. I'm not sure what process tags is and why it's added to the snapshot

integration-tests/debugger/basic.spec.js
```diff
@@ -800,6 +800,58 @@ describe('Dynamic Instrumentation', function () {
...
+      dependencies: ['fastify']
+    })
+
+    describe('input messages', function () {
+      it('should include process_tags in snapshot when enabled', function (done) {
+        t.agent.on('debugger-input', ({ payload }) => {
+          const snapshot = payload[0].debugger.snapshot
+
+          // Assert that process_tags are present
+          assert.ok(snapshot.process_tags)
```
watson, line 816: Nit: I usually don't add extra asserts to validate that a given property is present, if I can assert its value instead. Especially if you as here do that on the very next line.

Same in the section below

packages/dd-trace/src/debugger/devtools_client/index.js
```diff
@@ -215,6 +216,10 @@ session.on('Debugger.paused', async ({ params }) => {
       language: 'javascript'
     }
 
+    if (config.propagateProcessTags?.enabled) {
```
watson, line 219: As far as I can see in the client lib config, `propagateProcessTags` will always be an object, right?

```suggestion
    if (config.propagateProcessTags.enabled) {
```

watson (approved):
I would prefer a few minor things to be cleaned up, but since the holidays are coming up, I will not want to be a blocker, so I'll approve this PR (as it's just minor optimizations). It would be best to just fix them, and feel free to do so and to get someone else to approve the PR afterwards, but it could also be fixed in a follow-up PR if you prefer.

packages/dd-trace/test/process-tags.spec.js
```diff
@@ -37,41 +49,56 @@ describe('process-tags', () => {
...
     })
 
-    it('should serialize tags correctly', () => {
-      const result = getProcessTags()
+    it('should sort tags alphabetically', () => {
+      assert.strictEqual(processTags.tags[0][0], 'entrypoint.basedir')
+      assert.strictEqual(processTags.tags[1][0], 'entrypoint.name')
+      assert.strictEqual(processTags.tags[2][0], 'entrypoint.type')
+      assert.strictEqual(processTags.tags[3][0], 'entrypoint.workdir')
+      assert.strictEqual(processTags.tags[4][0], 'package.json.name')
```
BridgeAR, line 95: Please always use assertObjectContains (partial deep strict equal), if possible. That way the output would be much better and more informative in case of an error. The rule I use is: using as few assertions as needed and possible for a single variable.

packages/dd-trace/test/process-tags.spec.js
```diff
@@ -280,11 +307,16 @@ describe('process-tags', () => {
...
       delete require.cache[require.resolve('../src/span_processor')]
+      delete require.cache[require.resolve('../src/process-tags')]
     })
 
     it('should enable process tags propagation when set to true', () => {
       process.env.DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED = 'true'
 
+      // Need to reload config first, then process-tags (which reads from config)
+      delete require.cache[require.resolve('../src/config')]
+      delete require.cache[require.resolve('../src/process-tags')]
```
BridgeAR, line 318: You could also use the `getConfigFresh()` helper

packages/dd-trace/test/process-tags.spec.js
```diff
@@ -37,41 +49,56 @@ describe('process-tags', () => {
...
 
-    // note that these tests may fail if the tracer folder structure changes
-    it('should set sensible values based on tracer project structure and be sorted alphabetically', () => {
-      const result = getProcessTags()
+    it('should set sensible values', () => {
+      const basedirTag = processTags.tags.find(([name]) => name === 'entrypoint.basedir')
+      const nameTag = processTags.tags.find(([name]) => name === 'entrypoint.name')
+      const typeTag = processTags.tags.find(([name]) => name === 'entrypoint.type')
+      const workdirTag = processTags.tags.find(([name]) => name === 'entrypoint.workdir')
+      const packageNameTag = processTags.tags.find(([name]) => name === 'package.json.name')
```
BridgeAR, line 71: The next test asserts that the order is sorted, thus, the find is actually not needed :)

integration-tests/debugger/basic.spec.js
tlhunter: This value can be `null` which complicates things

tlhunter: I removed this line and the next one in https://github.com/DataDog/dd-trace-js/pull/7176 as they're entirely redundant with the following assertions which check object properties.

packages/dd-trace/test/process-tags.spec.js
tlhunter: This is addressed in https://github.com/DataDog/dd-trace-js/pull/7176

packages/dd-trace/test/process-tags.spec.js
tlhunter: This is addressed in https://github.com/DataDog/dd-trace-js/pull/7176

packages/dd-trace/test/process-tags.spec.js
tlhunter: This is addressed in https://github.com/DataDog/dd-trace-js/pull/7176

tlhunter:
@watson had this to say:

> Hi Thomas
> I just realized that we might have implemented debugger support for process tags incorrectly in the Node.js client lib.
> In your PR (https://github.com/DataDog/dd-trace-js/pull/7132) the process_tags property was possibly added in the wrong location in the JSON payload sent to the backend. It was added under debugger.snapshot.process_tags, but if I'm not mistaken it should probably have been in the root?
> Does this sound right?

PR #7133, opened 2025-12-17
test: use partialDeepStrictEqual instead of individual assertions

BridgeAR:
This makes the test output in case of errors easier to understand as well as making the code easier to read.

The diff is big while it should be easy to review.

The code change was done with an AI, concentrating solely on this migration in files with most offenders.

### DIFF
--- a/packages/datadog-plugin-pg/test/index.spec.js
+++ b/packages/datadog-plugin-pg/test/index.spec.js
@@ -155,16 +155,18 @@ describe('Plugin', () => {
           it('should handle callback errors', done => {
             let error
 
-            agent.assertFirstTraceSpan({
-              meta: {
-                [ERROR_TYPE]: error.name,
-                [ERROR_MESSAGE]: error.message,
-                [ERROR_STACK]: error.stack,
-                component: 'pg'
-              },
-              metrics: {
-                'network.destination.port': 5432
-              }
+            agent.assertSomeTraces(traces => {
+              assertObjectContains(traces[0][0], {
+                meta: {
+                  [ERROR_TYPE]: error.name,
+                  [ERROR_MESSAGE]: error.message,
+                  [ERROR_STACK]: error.stack,
+                  component: 'pg'
+                },
+                metrics: {
+                  'network.destination.port': 5432
+                }
+              })
             })
               .then(done)
               .catch(done)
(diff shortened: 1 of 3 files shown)


simon-id:
@BridgeAR did you review yourself carefully every line of the diff to make sure the output of the AI is correct and functionally equivalent to the original ?

packages/datadog-plugin-azure-event-hubs/test/integration-test/client.spec.js
```diff
@@ -92,8 +116,8 @@ describe.skip('esm', () => {
       const res = agent.assertMessageReceived(({ headers, payload }) => {
         assert.ok(!('_dd.span_links' in payload[2][0]))
       })
-      const envVar = { DD_TRACE_AZURE_EVENTHUBS_BATCH_LINKS_ENABLED: false, ...spawnEnv }
-      proc = await spawnPluginIntegrationTestProc(sandboxCwd(), 'server.mjs', agent.port, undefined, envVar)
+      const envVar = { DD_TRACE_AZURE_EVENTHUBS_BATCH_LINKS_ENABLED: 'false', ...spawnEnv }
+      proc = await spawnPluginIntegrationTestProc(sandboxCwd(), 'server.mjs', agent.port, envVar)
```
watson, line 120: Why did this work before?

packages/datadog-plugin-pg/test/index.spec.js
```diff
@@ -609,7 +649,9 @@ describe('Plugin', () => {
 
         it('query should inject _dd.dbm_trace_injected into span', done => {
           agent.assertSomeTraces(traces => {
-            assert.strictEqual(traces[0][0].meta['_dd.dbm_trace_injected'], 'true')
+            assertObjectContains(traces[0][0].meta, {
+              '_dd.dbm_trace_injected': 'true'
+            })
```
watson, line 654: Nit: This one could be reverted. Maybe there's similar situations in the files above that I missed. Consider validating

packages/datadog-plugin-azure-event-hubs/test/integration-test/client.spec.js
BridgeAR: Envs are always converted to a string when being assigned. It is a magic setter.

`undefined` was fine due to the overloading happening in the method.

PR #7134, opened 2025-12-18
test: rewrite chai to assert

BridgeAR:
These changes are done with a script. The script is visible in another commit that is reverted.

Manual fixes will be pushed after checking test failures.


packages/dd-trace/test/appsec/iast/context/context-plugin.spec.js
```diff
@@ -68,8 +67,8 @@ describe('IastContextPlugin', () => {
     it('should add a subscription to the channel', () => {
       plugin.startCtxOn(channelName, tag)
 
-      expect(addSub).to.be.calledOnceWith(channelName)
-      expect(getAndRegisterSubscription).to.be.calledOnceWith({ channelName, tag, tagKey: TagKey.SOURCE_TYPE })
+      sinon.assert.calledOnceWithMatch(addSub, channelName)
+      sinon.assert.calledOnceWithExactly(getAndRegisterSubscription, { channelName, tag, tagKey: TagKey.SOURCE_TYPE })
```
simon-id, line 71: nit: why is one converted to calledOnceWithMatch and the other to calledOnceWithExactly ? just because otherwise the test fails ?

packages/dd-trace/test/appsec/sdk/set_user.spec.js
```diff
@@ -80,11 +80,11 @@ describe('set_user', () => {
...
-        expect(rootSpan.setTag.getCall(0)).to.have.been.calledWithExactly('usr.id', '123')
-        expect(rootSpan.setTag.getCall(1)).to.have.been.calledWithExactly('usr.email', 'a@b.c')
-        expect(rootSpan.setTag.getCall(2)).to.have.been.calledWithExactly('usr.custom', 'hello')
-        expect(rootSpan.setTag.getCall(3)).to.have.been.calledWithExactly('usr.session_id', '133769')
-        expect(rootSpan.setTag.getCall(4)).to.have.been.calledWithExactly('_dd.appsec.user.collection_mode', 'sdk')
+        assert.strictEqual(rootSpan.setTag.getCall(0).calledWithExactly('usr.id', '123'), true)
+        assert.strictEqual(rootSpan.setTag.getCall(1).calledWithExactly('usr.email', 'a@b.c'), true)
+        assert.strictEqual(rootSpan.setTag.getCall(2).calledWithExactly('usr.custom', 'hello'), true)
+        assert.strictEqual(rootSpan.setTag.getCall(3).calledWithExactly('usr.session_id', '133769'), true)
+        assert.strictEqual(rootSpan.setTag.getCall(4).calledWithExactly('_dd.appsec.user.collection_mode', 'sdk'), true)
```
simon-id, line 87: Doesn't that work better ?
```javascript
sinon.assert.calledWithExactly(rootSpan.setTag.getCall(0), 'usr.id', '123')
```

integration-tests/playwright/playwright.spec.js
```diff
@@ -159,14 +158,14 @@ versions.forEach((version) => {
...
-            assert.equal(testModuleEvent.content.meta[TEST_STATUS], 'fail')
-            assert.equal(testSessionEvent.content.meta[TEST_TYPE], 'browser')
-            assert.equal(testModuleEvent.content.meta[TEST_TYPE], 'browser')
+            assert.strictEqual(testModuleEvent.content.meta[TEST_STATUS], 'fail')
+            assert.strictEqual(testSessionEvent.content.meta[TEST_TYPE], 'browser')
+            assert.strictEqual(testModuleEvent.content.meta[TEST_TYPE], 'browser')
 
-            assert.exists(testSessionEvent.content.meta[ERROR_MESSAGE])
-            assert.exists(testModuleEvent.content.meta[ERROR_MESSAGE])
+            assert.ok(testSessionEvent.content.meta[ERROR_MESSAGE] != null)
```
watson: I think `exists` is better replaced by:

```suggestion
            assert.notEqual(testSessionEvent.content.meta[ERROR_MESSAGE], null)
```

That will fail only if the tested value is either `undefined` or `null` and will therefore give a better, more meaningful error message

packages/datadog-instrumentations/test/helpers/check-require-cache.spec.js
```diff
@@ -17,7 +16,7 @@ describe('check-require-cache', () => {
   it('should be no warnings when tracer is loaded first', (done) => {
     exec(`${process.execPath} ./check-require-cache/good-order.js`, opts, (error, stdout, stderr) => {
       assert.strictEqual(error, null)
-      expect(stderr).to.not.include("Package 'express' was loaded")
+      assert.ok(!(stderr).includes("Package 'express' was loaded"))
```
watson:
```suggestion
      assert.doesNotMatch(stderr, /Package 'express' was loaded/)
```

packages/dd-trace/test/appsec/iast/overhead-controller.spec.js
```diff
@@ -670,3 +670,17 @@ describe('Overhead controller', () => {
     })
   })
 })
+
+function hasNestedProperty (obj, path) {
```
watson, line 674: Can we move this function to a helper and use it elsewhere? I'm sure we have similar needs in other tests

packages/dd-trace/test/openfeature/writers/exposures.spec.js
```diff
@@ -71,14 +71,14 @@ describe('OpenFeature Exposures Writer', () => {
...
-      expect(writer._buffer).to.be.an('array').that.is.empty
+      assert.strictEqual(writer._interval, 1000)
+      assert.strictEqual(writer._timeout, 5000)
+      assert.strictEqual(writer._bufferLimit, 1000)
+      assert.deepStrictEqual(writer._buffer, [])
     })
 
     it('should set up periodic flushing', () => {
-      expect(writer._periodic).to.exist
+      assert.ok(writer._periodic != null)
```
watson: Same here:

```suggestion
      assert.notEqual(writer._periodic, null)
```

packages/dd-trace/test/appsec/iast/context/context-plugin.spec.js
BridgeAR: Yes

packages/dd-trace/test/appsec/sdk/set_user.spec.js
BridgeAR: Both works. I did not look into all transformations if they are perfect as long as they keep the overall behavior as expected.

packages/dd-trace/test/appsec/iast/overhead-controller.spec.js
BridgeAR: I gave it some more thought and I think it is actually good to not use it as generic helper. I believe we should normally just set up the whole object and check for that instead of using some magic string that will end up as the string we expect.

This particular case is special, since it needs to understand the user input.

packages/datadog-instrumentations/test/helpers/check-require-cache.spec.js
BridgeAR: I checked through the whole code for this pattern and updated them where possible

integration-tests/playwright/playwright.spec.js
BridgeAR: I went for just removing the not null check. That way assert is able to tell what was passed through and it is simpler to write and produces an even nicer message overall.

packages/dd-trace/test/openfeature/writers/exposures.spec.js
BridgeAR: See above

watson (approved):

PR #7135, opened 2025-12-18
test: migrate tap to mocha

BridgeAR:
This tries to align with the former behavior as close as possible. This is for example the case by using more parallel threads, even if the CPU has less cores.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,6 +24,14 @@ dd-trace is the Datadog client library for Node.js.
 - `integration-tests/` - E2E integration tests
 - `benchmark/` - Performance benchmarks
 
+**Package Structure:**
+
+Each package follows a consistent structure:
+- `src/` - Source code for the package
+- `test/` - Unit tests for the package
+- Unit test files always follow the `*.spec.js` naming convention
+- Test directories may also contain helper files (e.g., `naming.js`, `helpers.js`, `suite.js`) and integration test subdirectories
+
 ## Testing Instructions
 
 ### Testing Workflow
@@ -106,7 +114,10 @@ assertObjectContains(response, { status: 200, body: { user: { name: 'Alice' } } 
 ### Linting & Naming
 - Lint: `yarn lint` / `yarn lint:fix`
 - Files: kebab-case
-- JSDoc: TypeScript-compatible syntax (`@param {string}`, `@returns {Promise<void>}`, `@typedef`)
+
+### JSDoc
+- Use TypeScript-compatible syntax (`@param {string}`, `@returns {Promise<void>}`, `@typedef`)
+- Never use `any` (be specific or use `unknown` if type is truly unknown)
 
 ### Import Ordering
 
@@ -136,7 +147,6 @@ const log = require('../log')
   const { NODE_MAJOR } = require('./version')
   if (NODE_MAJOR >= 20) { /* Use Node.js 20+ API */ }
   ```
-- **Prefix Node.js core modules with `node:`** (e.g., `require('node:assert')`)
 
 ### Performance and Memory
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -240,9 +240,11 @@ DD_TRACE_DEBUG=true node your-app.js
 DD_TRACE_DEBUG=true yarn test:debugger
 ```
 
-### Documentation
+### JSDoc
 
 Document all APIs with TypeScript-compatible JSDoc to ensure proper types without using TypeScript. This enables type checking and IDE autocompletion while maintaining the JavaScript codebase. Use TypeScript type syntax in JSDoc annotations (e.g., `@param {string}`, `@returns {Promise<void>}`).
+
+Never use the type `any` - be specific.
 
 ## Adding New Configuration Options
 
--- a/integration-tests/remote_config.spec.js
+++ b/integration-tests/remote_config.spec.js
@@ -56,22 +56,17 @@ describe('Remote config client id', () => {
 
           const processTags = client.client_tracer.process_tags
 
-          // Verify process_tags is an object
-          assert.strictEqual(typeof processTags, 'object')
+          // Verify process_tags is an array of strings
+          assert.ok(Array.isArray(processTags), 'process_tags should be an array')
 
           // Verify required process tags are present
-          assert.ok('entrypoint.basedir' in processTags)
-          assert.ok('entrypoint.name' in processTags)
-          assert.ok('entrypoint.type' in processTags)
-          assert.ok('entrypoint.workdir' in processTags)
+          assert.ok(processTags.some(tag => tag.startsWith('entrypoint.basedir:')))
+          assert.ok(processTags.some(tag => tag.startsWith('entrypoint.name:')))
+          assert.ok(processTags.some(tag => tag.startsWith('entrypoint.type:')))
+          assert.ok(processTags.some(tag => tag.startsWith('entrypoint.workdir:')))
 
           // Verify entrypoint.type has the expected value
-          assert.strictEqual(processTags['entrypoint.type'], 'script')
-
-          // Verify values are strings (not undefined)
-          assert.strictEqual(typeof processTags['entrypoint.name'], 'string')
-          assert.strictEqual(typeof processTags['entrypoint.workdir'], 'string')
-          assert.strictEqual(typeof processTags['entrypoint.basedir'], 'string')
+          assert.ok(processTags.some(tag => tag === 'entrypoint.type:script'))
           done()
         } catch (err) {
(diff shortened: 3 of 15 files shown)


scripts/mocha-parallel-files.js
```diff
@@ -0,0 +1,506 @@
...
+
+function stableUnique (items) {
+  const seen = new Set()
+  const out = []
+  for (const item of items) {
+    if (seen.has(item)) continue
+    seen.add(item)
+    out.push(item)
+  }
+  return out
```
tlhunter:
```suggestion
  return [...new Set(items)]
```

tlhunter (dismissed):
As mentioned elsewhere my preference for test suites is:
all node:test > all tap > all mocha > tap + mocha

So I would consider this PR an improvement to what we have today even if it's not in the direction I would ultimately prefer.

packages/dd-trace/test/ci-visibility/exporters/ci-visibility-exporter.spec.js
```diff
@@ -1,18 +1,17 @@
...
-const { describe, it, beforeEach, afterEach, context } = require('tap').mocha
-const sinon = require('sinon')
-const nock = require('nock')
 const cp = require('node:child_process')
 const fs = require('node:fs')
 const zlib = require('node:zlib')
 
-require('../../../../dd-trace/test/setup/core')
+const { describe, it, beforeEach, afterEach } = require('mocha')
+const context = describe
```
watson, line 9: FYI, this would technically also be possible, but I'm ok with splitting it over two lines and keeping it as is

```suggestion
const { describe, describe: context, it, beforeEach, afterEach } = require('mocha')
```

packages/dd-trace/test/encode/coverage-ci-visibility.spec.js
```diff
@@ -1,24 +1,22 @@
 'use strict'
 
 const assert = require('node:assert/strict')
-const { assertObjectContains } = require('../../../../integration-tests/helpers')
+/**
```
watson: Nit: Not sure why this typedef was moved up in the middle of the require block, but if it needs to be here, I prefer a blank line above it

```suggestion

/**
```

packages/dd-trace/test/plugins/util/inferred_proxy.spec.js
```diff
@@ -1,16 +1,16 @@
 'use strict'
 
 const assert = require('node:assert/strict')
-const { describe, it, afterEach } = require('tap').mocha
-const axios = require('axios')
 const { Agent } = require('node:http')
+// Create axios instance with no connection pooling
```
watson: This comment seemed to have been moved by a mistake

packages/dd-trace/test/profiling/exporters/agent.spec.js
```diff
@@ -439,7 +438,7 @@ describe('exporters/agent', function () {
     })
   })
 
-  describe('using UDS', () => {
+  ;(os.platform() === 'win32' ? describe.skip : describe)('using UDS', () => {
```
watson: It's better to use `this.skip()`, something along the lines of:


```suggestion
  describe('using UDS', () => {
    if (os.platform() === 'win32') this.skip()
```

packages/dd-trace/test/azure_metadata.spec.js
```diff
@@ -1,14 +1,42 @@
...
+    for (const key of AZURE_ENV_KEYS) {
+      const value = initialAzureEnv[key]
+      if (value === undefined) {
+        delete process.env[key]
+      } else {
+        process.env[key] = value
+      }
+    }
+  })
+
```
watson, line 39: Did you mean to include this in this PR?

packages/dd-trace/test/custom-metrics.spec.js
```diff
@@ -1,14 +1,14 @@
...
 const assert = require('node:assert/strict')
-/* eslint-disable no-console */
-const { describe, it, beforeEach, afterEach } = require('tap').mocha
-
 const http = require('node:http')
 const path = require('node:path')
 const os = require('node:os')
 const { exec } = require('node:child_process')
 
+/* eslint-disable no-console */
```
watson: Nit: This comment was originally at the top of the file where it belongs, but after several automatic updates to the file, it as now been moved down in the middle of the require block. Consider a manual edit to just move it back up.

packages/dd-trace/test/log.spec.js
```diff
@@ -1,16 +1,15 @@
 'use strict'
 
 const assert = require('node:assert/strict')
-const { describe, it, beforeEach, afterEach } = require('tap').mocha
+/* eslint-disable no-console */
```
watson: This comment shouldn't be in the middle of the require block

packages/dd-trace/test/runtime_metrics.spec.js
```diff
@@ -1,11 +1,12 @@
...
 const assert = require('node:assert')
 const os = require('node:os')
 
+const { describe, it, beforeEach, afterEach } = require('mocha')
 const proxyquire = require('proxyquire')
 const sinon = require('sinon')
+
 const performance = require('node:perf_hooks').performance
 const { setImmediate, setTimeout } = require('node:timers/promises')
 const util = require('node:util')
```
watson: Nit: you seem to have moved all other node-core require calls to the top of the files in this PR, except for this file. Feel free to fix or leave as is

packages/dd-trace/test/runtime_metrics.spec.js
```diff
@@ -14,8 +15,6 @@ require('./setup/core')
 
 const { DogStatsDClient } = require('../src/dogstatsd')
 
-const isWindows = os.platform() === 'win32'
```
watson, line 17: I assume you managed to make these tests work on Windows, since you removed this logic?

scripts/mocha-parallel-files.js
watson, line 1: If I'm not mistaken, this isn't related to migrating from tap to mocha? If so, I would prefer this being a separate PR, otherwise I fear it will not get the review it deserves.

packages/dd-trace/test/setup/mocha.js
```diff
@@ -148,11 +149,11 @@ function withPeerService (tracer, pluginName, spanGenerationFn, service, service
       const useCallback = spanGenerationFn.length === 1
       const spanGenerationPromise = useCallback
         ? new Promise((resolve, reject) => {
```
watson: I would prefer this instead of having to change calls to `resolve()` to `resolve(undefined)`

```suggestion
        ? new Promise((/** @type {() => void} */ resolve, reject) => {
```

packages/dd-trace/test/setup/mocha.js
watson, line 1: The changes in this file is not entirely clear to me

packages/dd-trace/test/runtime_metrics.spec.js
BridgeAR: Yes, the tests are passing fine. I am unsure when this particular changed. I did fix the non-native part recently, while the native part seemed to be supported before as well. We worked quite a bit on runtime metrics in the last year, so it was probably just overlooked that it was fixed for Windows at that time.

scripts/mocha-parallel-files.js
BridgeAR: It could indeed be separated out into an individual PR. The downside is, that our CI will be much slower when we do not land this at the same time.

packages/dd-trace/test/ci-visibility/exporters/ci-visibility-exporter.spec.js
BridgeAR: I feel like it is easier to follow when having it as two separate lines.

We could also rename context to describe as alternative, while this requires less code changes.

packages/dd-trace/test/custom-metrics.spec.js
BridgeAR: I will inline it to only disable it for the current two lines.

packages/dd-trace/test/azure_metadata.spec.js
BridgeAR: Yes, I forgot why, but the changed parallel mode conflicted with the envs. This is actually just always going to reset the keys to their original state and as such make sure tests may run fully independent from each other in the future, which is a good improvement one way or the other.

packages/dd-trace/test/setup/mocha.js
BridgeAR: The setup is used in many of our test files and we did not fully clean up different states. The changes were needed to run them in the new parallel mode with mocha.

PR #7137, opened 2025-12-18
feat(debugger): add Remote Enablement support

watson:
### What does this PR do?

Implements Remote Enablement (RE) for Dynamic Instrumentation/Live Debugger and Code Origin via `APM_TRACING_MULTICONFIG`.

**Key Remote Config changes:**
- Add `RCClientLibConfigManager` class for priority-based merging of multiple `APM_TRACING` configs (Service+Env > Service > Env > Cluster > Org)
- Add `APM_TRACING_MULTICONFIG`, `APM_TRACING_ENABLE_DYNAMIC_INSTRUMENTATION`, and `APM_TRACING_ENABLE_CODE_ORIGIN` capability flags
- Centralize client library configuration remote config logic in `config/remote_config.js`

**Key Debugger changes:**
- Add lifecycle methods (`isStarted()`, `stop()`, `configure()`) to `debugger/index.js` for dynamic control

**Key Code Origin changes:**
- Add runtime enable/disable support for Code Origin via Remote Config
- Update Express and Fastify code origin plugins to pre-compute entry tags at route registration time when RC is enabled (allowing runtime enabling to work immediately)
- Add integration tests for Code Origin Remote Config toggling

### Motivation

The Debugger (Dynamic Instrumentation/Live Debugging) needs to support Remote Enablement, allowing operators to dynamically enable, disable, or reconfigure the debugger without requiring service restarts.

This implementation follows the `APM_TRACING_MULTICONFIG` RFC.

### DIFF
--- a/packages/datadog-plugin-express/src/code_origin.js
+++ b/packages/datadog-plugin-express/src/code_origin.js
@@ -16,7 +16,6 @@ class ExpressCodeOriginForSpansPlugin extends Plugin {
 
     // Middleware/request handling: apply pre-computed tags to spans
     const handleMiddlewareEnter = ({ req, layer }) => {
-      if (!this._enabled) return
       const tags = layerTags.get(layer)
       if (!tags) return
       web.getContext(req)?.span?.addTags(tags)
--- a/packages/datadog-plugin-fastify/src/code_origin.js
+++ b/packages/datadog-plugin-fastify/src/code_origin.js
@@ -15,7 +15,6 @@ class FastifyCodeOriginForSpansPlugin extends Plugin {
     super(...args)
 
     this.addSub('apm:fastify:request:handle', ({ req, routeConfig }) => {
-      if (!this._enabled) return
       const tags = routeConfig?.[kCodeOriginForSpansTagsSym]
       if (!tags) return
       web.getContext(req)?.span?.addTags(tags)
--- a/packages/dd-trace/src/config.js
+++ b/packages/dd-trace/src/config.js
@@ -375,9 +375,11 @@ class Config {
     this.#merge()
   }
 
+  // TODO: Remove the `updateOptions` method. We don't want to support updating the config this way
   /**
    * Updates the configuration with new programmatic options.
    *
+   * @deprecated This method should not be used and will be removed in a future version.
    * @param {object} options - Configuration options to apply (same format as tracer init options)
    */
   updateOptions (options) {
--- a/packages/dd-trace/src/config/remote_config.js
+++ b/packages/dd-trace/src/config/remote_config.js
@@ -152,7 +152,7 @@ class RCClientLibConfigManager {
       libConfigCount++
 
       for (const [key, value] of Object.entries(libConfig)) {
-        if (key in merged) continue
+        if (Object.hasOwn(merged, key)) continue
 
         // Set the value even if it's null (to reset) but not if it's undefined (missing)
         if (value === null) {
@@ -201,6 +201,9 @@ function enable (rc, config, onConfigUpdated) {
   rc.updateCapabilities(RemoteConfigCapabilities.APM_TRACING_ENABLE_CODE_ORIGIN, true)
 
   const rcClientLibConfigManager = new RCClientLibConfigManager(config.service, config.env)
+
+  // Subscribe to APM_TRACING product (setBatchHandler used below doesn't automatically subscribe)
+  rc.subscribeProducts('APM_TRACING')
 
   // Use a batch handler to process all changes before updating the config. This is important in case there's
   // conflicting configs between, for example, the org and service level.
--- a/packages/dd-trace/src/proxy.js
+++ b/packages/dd-trace/src/proxy.js
@@ -131,8 +131,8 @@ class Tracer extends NoopProxy {
 
         const tracingRemoteConfig = require('./config/remote_config')
         tracingRemoteConfig.enable(rc, config, () => {
-          this._updateTracing(config, rc)
-          this._updateDebugger(config, rc)
+          this.#updateTracing(config)
+          this.#updateDebugger(config, rc)
         })
 
         rc.setProductHandler('AGENT_CONFIG', (action, conf) => {
@@ -186,7 +186,7 @@ class Tracer extends NoopProxy {
         runtimeMetrics.start(config)
       }
 
-      this._updateTracing(config)
+      this.#updateTracing(config)
 
       this._modules.rewriter.enable(config)
 
@@ -245,7 +245,7 @@ class Tracer extends NoopProxy {
     }
   }
 
-  _updateTracing (config) {
+  #updateTracing (config) {
     if (config.tracing !== false) {
       if (config.appsec.enabled) {
         this._modules.appsec.enable(config)
@@ -297,19 +297,21 @@ class Tracer extends NoopProxy {
    * @param {object} config - The tracer configuration object
    * @param {object} rc - The RemoteConfig instance
    */
-  _updateDebugger (config, rc) {
+  #updateDebugger (config, rc) {
     const shouldBeEnabled = config.dynamicInstrumentation.enabled
     const isCurrentlyStarted = DynamicInstrumentation.isStarted()
 
-    if (shouldBeEnabled && !isCurrentlyStarted) {
-      log.debug('[proxy] Starting Dynamic Instrumentation via remote config')
-      DynamicInstrumentation.start(config, rc)
-    } else if (!shouldBeEnabled && isCurrentlyStarted) {
+    if (shouldBeEnabled) {
+      if (isCurrentlyStarted) {
+        log.debug('[proxy] Reconfiguring Dynamic Instrumentation via remote config')
+        DynamicInstrumentation.configure(config)
+      } else {
+        log.debug('[proxy] Starting Dynamic Instrumentation via remote config')
+        DynamicInstrumentation.start(config, rc)
+      }
+    } else if (isCurrentlyStarted) {
       log.debug('[proxy] Stopping Dynamic Instrumentation via remote config')
       DynamicInstrumentation.stop()
-    } else if (shouldBeEnabled && isCurrentlyStarted) {
-      log.debug('[proxy] Reconfiguring Dynamic Instrumentation via remote config')
-      DynamicInstrumentation.configure(config)
     }
   }
 
--- a/packages/dd-trace/test/proxy.spec.js
+++ b/packages/dd-trace/test/proxy.spec.js
@@ -953,6 +953,7 @@ function createApmTracingTransaction (configId, libConfig, action = 'apply') {
     toUnapply: action === 'unapply' ? [item] : [],
     toApply: action === 'apply' ? [item] : [],
     toModify: action === 'modify' ? [item] : [],
-    ack: sinon.spy()
+    ack: sinon.spy(),
+    error: sinon.spy()
   }
 }
--- a/integration-tests/code-origin-remote-config.spec.js
+++ b/integration-tests/code-origin-remote-config.spec.js
@@ -5,6 +5,8 @@ const path = require('node:path')
 const Axios = require('axios')
 const { FakeAgent, sandboxCwd, useSandbox, spawnProc } = require('./helpers')
 
+const ACKNOWLEDGED = 2
+
 describe('Code Origin Remote Config', function () {
   this.timeout(20000)
 
@@ -37,6 +39,7 @@ describe('Code Origin Remote Config', function () {
       cwd,
       env: {
         DD_TRACE_AGENT_PORT: agent.port,
+        DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS: 0.1,
         ...envVars
       }
     })
@@ -44,6 +47,28 @@ describe('Code Origin Remote Config', function () {
       baseURL: proc.url,
       headers: { 'Content-Type': 'application/json' }
     })
+  }
+
+  const addRemoteConfigAndWaitForAck = (libConfig) => {
+    return /** @type {Promise<void>} */ (new Promise((resolve) => {
+      // Random config id - Just needs to be unique between calls to this function
+      const configId = Math.random().toString(36).slice(2)
+      const handler = (id, version, state) => {
+        if (id === configId && state === ACKNOWLEDGED) {
+          agent.removeListener('remote-config-ack-update', handler)
+          resolve()
+        }
+      }
+      agent.on('remote-config-ack-update', handler)
+      agent.addRemoteConfig({
+        product: 'APM_TRACING',
+        id: configId,
+        config: {
+          service_target: { service: 'node', env: '*' },
+          lib_config: libConfig
+        }
+      })
+    }))
   }
 
   const assertCodeOriginPresent = (framework, url = '/hello') => {
@@ -81,7 +106,7 @@ describe('Code Origin Remote Config', function () {
           DD_REMOTE_CONFIG_ENABLED: 'true'
         }))
 
-        it('should disable code origin tags at runtime via setRemoteConfig', async () => {
+        it('should disable code origin tags at runtime via remote config', async () => {
           // Step 1: Make a request with code origin enabled (default)
           await assertCodeOriginPresent(framework)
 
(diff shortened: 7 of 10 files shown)


packages/dd-trace/src/config/remote_config.js
```diff
@@ -1,34 +1,218 @@
...
+const log = require('../log')
+
+module.exports = {
+  enable
+}
+
+/**
+ * @typedef {object} RemoteConfigOptions
+ * @property {boolean} [dynamic_instrumentation_enabled] - Enable Dynamic Instrumentation
+ * @property {boolean} [live_debugging_enabled] - Enable Live Debugging (deprecated alias)
```
shatzi: You didn't add exception_replay field. Because I assume it is not supported yet. Live debugging is also not used as it should enable trigger probes and capture code origin spans variables - see the spec for distributed debugger.

So, for now we only need to look at code origin and dynamic instrumentation flags.

packages/dd-trace/src/config/remote_config.js
```diff
@@ -1,34 +1,218 @@
...
+        // Set the value even if it's null (to reset) but not if it's undefined (missing)
+        if (value === null) {
+          merged[key] = undefined // TODO: Should this be null?
+        } else if (value !== undefined) {
+          merged[key] = value
+        }
+      }
+    }
+
+    log.debug('[config/remote_config] Merged %d configs into lib_config', libConfigCount)
```
shatzi, line 171: Nit: if all libConfigs were null, we didn't merged anything, should we return {} or null. 

Not sure if there is a difference,

packages/dd-trace/src/config/remote_config.js
```diff
@@ -1,34 +1,218 @@
...
   rc.updateCapabilities(RemoteConfigCapabilities.APM_TRACING_LOGS_INJECTION, true)
-  rc.updateCapabilities(RemoteConfigCapabilities.APM_TRACING_SAMPLE_RATE, true)
-  rc.updateCapabilities(RemoteConfigCapabilities.APM_TRACING_ENABLED, true)
-  rc.updateCapabilities(RemoteConfigCapabilities.APM_TRACING_SAMPLE_RULES, true)
 
-  // APM_TRACING product handler - manages tracer configuration
-  rc.setProductHandler('APM_TRACING', (action, conf) => {
+  // Debugger
+  rc.updateCapabilities(RemoteConfigCapabilities.APM_TRACING_ENABLE_DYNAMIC_INSTRUMENTATION, true)
+  rc.updateCapabilities(RemoteConfigCapabilities.APM_TRACING_ENABLE_LIVE_DEBUGGING, true)
```
shatzi: Not sure about this one

packages/dd-trace/src/config/remote_config.js
```diff
@@ -1,34 +1,218 @@
...
+
+  // Code Origin
+  rc.updateCapabilities(RemoteConfigCapabilities.APM_TRACING_ENABLE_CODE_ORIGIN, true)
+
+  const rcClientLibConfigManager = new RCClientLibConfigManager(config.service, config.env)
+
+  rc.setProductHandler('APM_TRACING', (action, conf, configId) => {
```
shatzi: does this run on every config we get? this mean are calling config.updateRemoteConfig multiple times. Maybe there is a better way to only call updateRemoteConfig only once we receive all the configs. I assume we would get all of the relevant configs on the first request.

packages/dd-trace/src/proxy.js
```diff
@@ -130,7 +130,7 @@ class Tracer extends NoopProxy {
         const rc = new RemoteConfig(config)
 
         const tracingRemoteConfig = require('./config/remote_config')
-        tracingRemoteConfig.enable(rc, config, this._enableOrDisableTracing.bind(this))
+        tracingRemoteConfig.enable(rc, config, this._updateTracing.bind(this), this._updateDebugger.bind(this))
```
shatzi: Why did we choose this design where we pass multiple functions to handle a new config?

I’m hoping we can simplify this so we don’t have to update multiple files every time we add a new product that needs to react to config changes.

Maybe something like a subscriber pattern, or `proxy.on('config-change', updateDebugger)`.

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -69,7 +69,9 @@ class LLMObs extends NoopLLMObs {
     }
 
     this._config.llmobs.enabled = true
-    this._config.configure({ ...this._config, llmobs: llmobsConfig })
+    // TODO: It's not nice that it's copying the entire config object to just update the llmobs config.
+    // We should find a better way to do this.
+    this._config.updateOptions({ ...this._config, llmobs: llmobsConfig })
```
shatzi: Does the config also include merged values from the remote config?

I’m wondering if there’s a bug where, when we copy the config, merged remote properties get converted into local options properties.

packages/dd-trace/src/config/remote_config.js
watson: > You didn't add exception_replay field. Because I assume it is not supported yet. 

Correct

> Live debugging is also not used as it should enable trigger probes and capture code origin spans variables - see the spec for distributed debugger.
> 
> So, for now we only need to look at code origin and dynamic instrumentation flags.

Got'ya. Thanks for explaining. I'll update the PR

packages/dd-trace/src/config/remote_config.js
watson: I'll remove as per this thread: https://github.com/DataDog/dd-trace-js/pull/7137#discussion_r2637181830

packages/dd-trace/src/proxy.js
watson: Yeah, it's not pretty. I tried to just extend the exiting pattern, but I'll try and refactor it to be more generic

packages/dd-trace/src/llmobs/sdk.js
watson: Hmm I think you're on to something 🤔 This isn't a bug introduced by this PR, but it's something we should deal with ASAP. I wonder if it's a blocker for this PR though or if it can be fixed in parallel?

packages/dd-trace/src/config/remote_config.js
watson: But does the protocol allow us to know if there's any remaining configs - i.e. when to finally call `updateRemoteConfig`? I always prefer a state-less algorithm if possible. And calling `updateRemoteConfig` multiple times isn't an issue as far as I see it, or am I missing something?

packages/dd-trace/src/config/remote_config.js
watson: I guess that's a protocol question that we need to align on across all client libs... If we return `null`, it would trigger a reset of all RC fields to their default value, instead of just not making any changes. I assumed we wanted the latter, but maybe I've misunderstood?

packages/dd-trace/src/config/remote_config.js
watson: Fixed 👍

packages/dd-trace/src/config/remote_config.js
watson: Fixed 👍

packages/dd-trace/src/proxy.js
watson: I've updated the PR now with a simplified implementation. Let me know what you think 😃

packages/dd-trace/src/llmobs/sdk.js
shatzi: I think this might be a bug on our side: if we disable DI remotely and it gets converted to a local disable, then we can’t re-enable it remotely.... other flags I think are ok.

packages/dd-trace/src/config/remote_config.js
shatzi: Assume we have org-level settings that enable ER, and service-level settings that disable it. Depending on the order, we could enable it first and then disable it, which would generate workers and tracer actions that could be avoided if we merged the configs and saw upfront that it should be disabled.

watson: Fixed by using the batch handler instead (that's what it is for)

packages/dd-trace/src/llmobs/sdk.js
watson: I've fixed this in a separate PR: https://github.com/DataDog/dd-trace-js/pull/7183
Once that is merged, I'll rebase this PR on top.

packages/dd-trace/src/config/remote_config.js
watson: I updated the code to account for if all `lib_config` are set to `null`, in which case this function now also returns `null` to properly unset all configs previously set via RC

packages/dd-trace/src/llmobs/sdk.js
watson: Fixed in `master` and rebased 👍

packages/dd-trace/src/config/remote_config.js
```diff
@@ -1,34 +1,225 @@
...
+   */
+  constructor (currentService, currentEnv) {
+    this.configs = new Map() // config_id -> { conf, priority }
+    this.currentService = currentService
+    this.currentEnv = currentEnv
+  }
+
+  /**
+   * Calculate priority based on target specificity
+   * Priority order: Service+Env (5) > Service (4) > Env (3) > Cluster (2) > Org (1)
```
tylfin: Maybe just mention explicitly that higher values take precedence:

```suggestion
   * Calculate priority based on target specificity.
   * Higher values take precedence.

   * Priority order (highest → lowest):
   *   Service+Env (5) > Service (4) > Env (3) > Cluster (2) > Org (1)
```

tylfin:
This generally looks good to me, you might want to run it against this system-test https://github.com/DataDog/system-tests/blob/da17bdc4b54f09001bfd39cf50fd63d8f377cb8d/tests/debugger/test_debugger_inproduct_enablement.py#L55, and try a few manual failure scenarios flipping and deleting the config

packages/dd-trace/test/proxy.spec.js
```diff
@@ -936,3 +940,19 @@ describe('TracerProxy', () => {
...
+    id: configId,
+    file: { lib_config: libConfig },
+    path: `datadog/1/APM_TRACING/${configId}`
+  }
+
+  return {
+    toUnapply: action === 'unapply' ? [item] : [],
+    toApply: action === 'apply' ? [item] : [],
+    toModify: action === 'modify' ? [item] : [],
+    ack: sinon.spy()
```
BridgeAR: Should this potentially also already contain an error method? If we ever want to test that otherwise, it will be confusing that it is missing. This has just happened to me in another PR where a mock did not contain the `once` method next to `on`.

tylfin (approved):

packages/datadog-plugin-express/src/code_origin.js
```diff
@@ -12,29 +14,34 @@ class ExpressCodeOriginForSpansPlugin extends Plugin {
 
     const layerTags = new WeakMap()
 
-    this.addSub('apm:express:middleware:enter', ({ req, layer }) => {
+    // Middleware/request handling: apply pre-computed tags to spans
+    const handleMiddlewareEnter = ({ req, layer }) => {
+      if (!this._enabled) return
```
BridgeAR: If the plugin is not enabled, should this ever be reached?

packages/datadog-plugin-fastify/src/code_origin.js
```diff
@@ -13,16 +15,28 @@ class FastifyCodeOriginForSpansPlugin extends Plugin {
     super(...args)
 
     this.addSub('apm:fastify:request:handle', ({ req, routeConfig }) => {
+      if (!this._enabled) return
```
BridgeAR: If the plugin is disabled, should the method ever be called?

packages/dd-trace/src/config.js
```diff
@@ -353,15 +352,36 @@ class Config {
...
+    this.#applyCalculated()
+    this.#merge()
+  }
+
+  /**
+   * Updates the configuration with new programmatic options.
+   *
+   * @param {object} options - Configuration options to apply (same format as tracer init options)
+   */
+  updateOptions (options) {
```
BridgeAR, line 385: Can you please add a deprecation notice + a TODO entry since we should never have an internal `updateOptions`

packages/dd-trace/src/config/remote_config.js
```diff
@@ -1,34 +1,226 @@
...
+
+    // Merge configs: take first non-null/undefined value for each field
+    // If a field is explicitly set to null, that means "reset to default"
+    for (const { conf } of sortedConfigs) {
+      const libConfig = conf.lib_config
+      if (libConfig == null) continue
+      libConfigCount++
+
+      for (const [key, value] of Object.entries(libConfig)) {
+        if (key in merged) continue
```
BridgeAR:
```suggestion
        if (Object.hasOwn(merged, key)) continue
```

packages/dd-trace/src/proxy.js
```diff
@@ -287,6 +290,29 @@ class Tracer extends NoopProxy {
...
   }
 
+  /**
+   * Updates the debugger (Dynamic Instrumentation) state based on remote config changes.
+   * Handles starting, stopping, and reconfiguring the debugger dynamically.
+   *
+   * @param {object} config - The tracer configuration object
+   * @param {object} rc - The RemoteConfig instance
+   */
+  _updateDebugger (config, rc) {
```
BridgeAR: Please use private properties

packages/dd-trace/src/proxy.js
```diff
@@ -287,6 +290,29 @@ class Tracer extends NoopProxy {
...
+    if (shouldBeEnabled && !isCurrentlyStarted) {
+      log.debug('[proxy] Starting Dynamic Instrumentation via remote config')
+      DynamicInstrumentation.start(config, rc)
+    } else if (!shouldBeEnabled && isCurrentlyStarted) {
+      log.debug('[proxy] Stopping Dynamic Instrumentation via remote config')
+      DynamicInstrumentation.stop()
+    } else if (shouldBeEnabled && isCurrentlyStarted) {
+      log.debug('[proxy] Reconfiguring Dynamic Instrumentation via remote config')
+      DynamicInstrumentation.configure(config)
+    }
```
BridgeAR:
```suggestion
    if (shouldBeEnabled) {
      if (isCurrentlyStarted) {
        log.debug('[proxy] Reconfiguring Dynamic Instrumentation via remote config')
        DynamicInstrumentation.configure(config)
      } else {
        log.debug('[proxy] Starting Dynamic Instrumentation via remote config')
        DynamicInstrumentation.start(config, rc)
      }
    } else if (isCurrentlyStarted) {
      log.debug('[proxy] Stopping Dynamic Instrumentation via remote config')
      DynamicInstrumentation.stop()
    }
```

packages/dd-trace/test/proxy.spec.js
watson: Added 👍

packages/datadog-plugin-express/src/code_origin.js
watson: You were right - this wasn't needed. I removed it again

packages/datadog-plugin-fastify/src/code_origin.js
watson: You were right - this wasn't needed. I removed it again

packages/dd-trace/src/config.js
watson: Done 👍

packages/dd-trace/src/config/remote_config.js
watson: Done 👍

packages/dd-trace/src/proxy.js
watson: I changed `_updateDebugger` to `#updateDebugger` and `_updateTracing` to `#updateTracing`, but left the other method and properties in this class to use the underscore, as they were not otherwise touched by this PR. Those should be changed in another PR

packages/dd-trace/src/proxy.js
watson: Updated 👍

BridgeAR (approved):

PR #7138, opened 2025-12-18
[appsec] Stripe business logic events

simon-id:
### What does this PR do?
This PR adds instrumentation for the Stripe SDK, and sends payloads from a couple of methods to the AppSec WAF, that will in turn convert them into span tags for the backend to receive.

### Motivation
These tags will be used by the WAF to detect endpoints that handle payments in the Endpoint Catalog, and will be useful for fraud detection and general security observability.

ST: https://github.com/DataDog/system-tests/pull/6219


packages/dd-trace/src/appsec/channels.js
```diff
@@ -37,10 +37,13 @@ module.exports = {
...
   responseSetHeader: dc.channel('datadog:http:server:response:set-header:start'),
   responseWriteHead: dc.channel('apm:http:server:response:writeHead:start'),
-  routerParam: dc.channel('datadog:router:param:start'),
   routerMiddlewareError: dc.channel('apm:router:middleware:error'),
+  routerParam: dc.channel('datadog:router:param:start'),
   setCookieChannel: dc.channel('datadog:iast:set-cookie'),
   setUncaughtExceptionCaptureCallbackStart: dc.channel('datadog:process:setUncaughtExceptionCaptureCallback:start'),
   startGraphqlResolve: dc.channel('datadog:graphql:resolver:start'),
+  stripeCheckoutSessionCreate: dc.channel('datadog:stripe:checkoutSession:create:finish'),
+  stripeConstructEvent: dc.channel('datadog:stripe:constructEvent:finish')
```
simon-id:
```suggestion
  stripeConstructEvent: dc.channel('datadog:stripe:constructEvent:finish'),
```

packages/datadog-instrumentations/src/stripe.js
```diff
@@ -0,0 +1,92 @@
...
+      shimmer.wrap(stripe.webhooks, 'constructEventAsync', wrapConstructEventAsync)
+    }
+
+    return stripe
+  }
+}
+
+addHook({
+  name: 'stripe',
+  versions: ['9', '10', '11', '12', '13', '14', '15', '16', '17', '18', '19', '>=20.0.0'],
```
uurien, line 89: Is this long list (instead of `>=9`) used here only for testing purpose? To force the the execution in the versions 9, 10...? 

I think that we shold add a comment explainig why, or else we can set this as `>=9` and in externals.json add these version in stripe section. I think that it is the main purpose of having an array for versions in externals.json:

```json
  "stripe": [
    {
      "name": "express",
      "versions": ["^4"]
    },
    {
      "name": "body-parser",
      "versions": ["1.20.1"]
    },
    {
      "name": "stripe",
      "versions": ['9', '10', '11', '12', '13', '14', '15', '16', '17', '18', '19']
    }
  ],
```

simon-id: yes, it's to make sure we test all the majors, as it seems like stripe sdk loves majors. I don't think a comment is super useful here as this syntax is used in other instrumentations too. If you want a comment, please make a github suggestion, it's easy and handy!
And about externals.json I am not as sure as oyu this is why we have it. What's wrong with having it in the instrmentation anyway ?

uurien: > What's wrong with having it in the instrumentation anyway ?

I am afraid that someone could see that and change it to `<=9`.

packages/datadog-instrumentations/src/stripe.js
```diff
@@ -0,0 +1,92 @@
...
+const shimmer = require('../../datadog-shimmer')
+const { channel, addHook } = require('./helpers/instrument')
+
+const checkoutSessionCreateFinishCh = channel('datadog:stripe:checkoutSession:create:finish')
+const paymentIntentCreateFinishCh = channel('datadog:stripe:paymentIntent:create:finish')
+const constructEventFinishCh = channel('datadog:stripe:constructEvent:finish')
+
+function wrapSessionCreate (create) {
+  return function wrappedSessionCreate () {
+    const promise = create.apply(this, arguments)
```
uurien, line 12: These methods are implemented in stripe using the function `stripeMethod`: https://github.com/stripe/stripe-node/blob/47cdb5eb972d00bd682aa9078e0cfdc1049f7226/src/StripeMethod.ts#L23

This method, tecnically supports callbacks, and when the callback is used, the promise always succeed with empty result.
https://github.com/stripe/stripe-node/blob/47cdb5eb972d00bd682aa9078e0cfdc1049f7226/src/utils.ts#L378-L379

In the `.d.ts` files of stripe, there is no mention to callback usage. 

Should we support also this use of the SDK?

simon-id: Oof good point, any mention of callbacks was removed in version 8 (which we support), but they never actually removed support for it. Considering it was already a legacy API 8+ years ago according to their readme, and that it would require significant changes to the instrumentations and the test suite, I'd say we leave it like this until we get a customer complaining

uurien: > in version 8 (which we support)

(which we **don't** support) we support >=9.

If there is no mention to callbacks since v8, lgtm

simon-id: Ah yes, i changed the support range at the last minute to >=9 because i was too lazy to code something for version 8

packages/datadog-instrumentations/src/stripe.js
BridgeAR: I believe we should just improve our install script to always run tests for all majors in a range

uurien (approved):

PR #7147, opened 2025-12-19
refactor(config): use getEnv instead of getEnvironmentVariables

watson:
## What Does This Do?

This PR refactors the `Config` class to standardize environment variable access by replacing `getEnvironmentVariables()` with `getEnv()` for consistency throughout the codebase.

## Motivation

Using a consistent method for environment variable access improves code maintainability and readability.

### DIFF
--- a/packages/dd-trace/src/config.js
+++ b/packages/dd-trace/src/config.js
@@ -326,23 +326,20 @@ class Config {
     this.#applyRemote({})
     this.#merge()
 
-    // Use typed version of `this` to appease the type checker
-    const self = /** @type {Record<string, any>} */ (this)
-
-    tagger.add(self.tags, {
-      service: self.service,
-      env: self.env,
-      version: self.version,
+    tagger.add(this.tags, {
+      service: this.service,
+      env: this.env,
+      version: this.version,
       'runtime-id': runtimeId
     })
 
-    if (self.isCiVisibility) {
-      tagger.add(self.tags, {
+    if (this.isCiVisibility) {
+      tagger.add(this.tags, {
         [ORIGIN_KEY]: 'ciapp-test'
       })
     }
 
-    if (self.gitMetadataEnabled) {
+    if (this.gitMetadataEnabled) {
       this.#loadGitMetadata()
     }
   }
@@ -1461,11 +1458,7 @@ class Config {
         }
       }
     }
-
-    // Use typed version of `this` to appease the type checker
-    const self = /** @type {Record<string, any>} */ (this)
-    self.sampler.sampleRate = self.sampleRate
-
+    this.sampler.sampleRate = this.sampleRate
     updateConfig(changes, this)
   }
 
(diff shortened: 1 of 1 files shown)


packages/dd-trace/src/config.js
```diff
@@ -321,28 +319,31 @@ class Config {
...
     this.#applyCalculated()
     this.#applyRemote({})
     this.#merge()
 
-    tagger.add(this.tags, {
-      service: this.service,
-      env: this.env,
-      version: this.version,
+    // Use typed version of `this` to appease the type checker
+    const self = /** @type {Record<string, any>} */ (this)
```
BridgeAR: I don't think we should do that. In fact, we should prohibit `any` usage in general.

This is just hiding the underlying issue of the types not being detected properly. So let us tackle that instead.

packages/dd-trace/src/config.js
```diff
@@ -1460,7 +1461,11 @@ class Config {
         }
       }
     }
-    this.sampler.sampleRate = this.sampleRate
+
+    // Use typed version of `this` to appease the type checker
+    const self = /** @type {Record<string, any>} */ (this)
```
BridgeAR: There is a definition for that in the sampler, if I am not mistaken. So we should just import that type :)

packages/dd-trace/src/config.js
```diff
@@ -1473,58 +1478,59 @@ class Config {
     }
   }
 
-  #loadGitMetadata (envs) {
+  #loadGitMetadata () {
+    // Use typed version of `this` to appease the type checker
+    const self = /** @type {Record<string, any>} */ (this)
```
BridgeAR: As above

packages/dd-trace/src/config.js
```diff
@@ -1473,58 +1478,59 @@ class Config {
...
       }
     }
     // otherwise, try to read Git metadata from the .git/ folder
-    if (!this.repositoryUrl || !this.commitSHA) {
-      const DD_GIT_FOLDER_PATH = envs.DD_GIT_FOLDER_PATH ??
-        path.join(process.cwd(), '.git')
-      if (!this.repositoryUrl) {
+    if (!self.repositoryUrl || !self.commitSHA) {
+      const DD_GIT_FOLDER_PATH = getEnv('DD_GIT_FOLDER_PATH')
+      const gitFolderPath = DD_GIT_FOLDER_PATH ?? path.join(process.cwd(), '.git')
```
BridgeAR, line 1503: ❤️

BridgeAR (changes requested):
Mostly looking good, while the self part should be reverted in my opinion.

packages/dd-trace/src/config.js
watson: Reverted

packages/dd-trace/src/config.js
watson: Reverted

packages/dd-trace/src/config.js
watson: Reverted

BridgeAR (approved):

PR #7149, opened 2025-12-19
FIx: Ws -add guardrail to bindAsyncStart in Publisher.

crysmags:
### What does this PR do?

Adding a check to verify the span is present on the context before calling finish. 

### Motivation

There is a current [escalation](https://datadoghq.atlassian.net/browse/APMS-18103) where the customer is using a long standing websocket connection and migrating traffic kubernetes pods while maintaining the same websocket. In doing so it appears the span on the context, which is the  is lost and when attempting to call finish there is no longer a present span.


BridgeAR:
I believe this got superseded

PR #7158, opened 2025-12-23
feat(llmobs): implement multi-tenant routing context support

PROFeNoM:
## What does this PR do?

Adds multi-tenant API key support for LLM Observability, enabling platforms to route LLMObs telemetry to different Datadog organizations from a single process.

**New API:**
```javascript
await llmobs.routingContext({ ddApiKey, ddSite }, async () => {
  // All LLMObs spans here route to the specified Datadog org
})
```

### Motivation

Multi-tenant AI platforms need to send LLM Observability data to their customers' Datadog organizations. This enables:
- Customer-specific trace visibility
- Data isolation between tenants
- Compliance with data residency requirements (different DD sites per customer)

###  Implementation
```
  withRoutingContext({ ddApiKey, ddSite }, fn)
      ↓ AsyncLocalStorage
  registerLLMObsSpan()
      ↓ captures routing to span tags
  span.finish()
      ↓
  SpanProcessor.process()
      ↓ extracts routing from tags
  writer.append(event, routing)
      ↓ routes to correct buffer
  setInterval flush()
      ↓ flushes each buffer to its endpoint
```

Key Changes:
- `withRoutingContext(options, fn)` method added to LLMObs SDK (uses `AsyncLocalStorage`)
- Routing context captured at span creation time
- `BaseLLMObsWriter` refactored to multi-buffer architecture (one buffer per API key)
- Each buffer flushes independently to its corresponding endpoint
- Agentless mode only

##  Testing

- `writers/multi-tenant.spec.js` - Tests for buffer routing, endpoint construction, API key isolation, context nesting, and concurrency

### DIFF
--- a/packages/dd-trace/test/llmobs/index.spec.js
+++ b/packages/dd-trace/test/llmobs/index.spec.js
@@ -256,9 +256,9 @@ describe('module', () => {
 
     const payload = {}
 
-    evalMetricAppendCh.publish(payload)
+    evalMetricAppendCh.publish({ payload })
 
-    sinon.assert.calledWith(LLMObsEvalMetricsWriterSpy().append, payload)
+    sinon.assert.calledWith(LLMObsEvalMetricsWriterSpy().append, payload, undefined)
   })
 
   it('removes all subscribers when disabling', () => {


PROFeNoM:
@codex review

packages/dd-trace/src/llmobs/writers/base.js
```diff
@@ -22,11 +22,11 @@ class BaseLLMObsWriter {
     this._timeout = timeout ?? getEnvironmentVariable('_DD_LLMOBS_TIMEOUT') ?? 5000 // 5s
     this._eventType = eventType
 
-    this._buffer = []
+    this._buffers = new Map()
```
BridgeAR: Please use private properties instead of underscores for all changed properties and new ones. Changing old ones is nice to have as separate commit :)

Please also add a comment to our Claude.md file that highlights this. Otherwise new code changes will likely come up by the AI that use the underscore instead.

PROFeNoM: Done. Converted to # private syntax where possible (#buffers, #getMaskedRoutingKey, #cleanupEmptyBuffers, #getUrlForRouting)

Kept _getRoutingKey and _getOrCreateBuffer with underscore because LLMObsSpanWriter (subclass) needs to access them.

Also added the convention to AGENTS.md (symlink to claude.md ?) under "Class Properties and Methods".

PROFeNoM: FYI, I did keep things like `get _buffer, set _buffer` with the underscore because they are backward compatibility shims. You were not thinking about changing them as well, right?

packages/dd-trace/src/llmobs/routing-context.js
sabrenner, line 1: i don't think this extra file is needed, we can inline this to the `withRoutingContext` function in the sdk directly

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -447,6 +448,22 @@ class LLMObs extends NoopLLMObs {
...
+  /**
+   * Execute a function within a routing context, directing all LLMObs spans
+   * created within to a specific Datadog organization.
+   *
+   * @param {Object} options - Routing options
+   * @param {string} options.ddApiKey - The Datadog API key for the target org
+   * @param {string} [options.ddSite] - The Datadog site (e.g., 'datadoghq.eu')
+   * @param {Function} fn - The function to execute within the routing context
+   * @returns {*} The return value of fn
+   */
```
sabrenner: i would make this part of the type definitions in `index.d.ts` (ex: [annotationContext](https://github.com/DataDog/dd-trace-js/blob/5c8cd4f8bf955603a011ba7c04df7158d5cf0129/index.d.ts#L2981-L2987), i think the typing could be exactly the same, and we can just swap out the text description/variable names for what you have here)

packages/dd-trace/src/llmobs/routing-context.js
```diff
@@ -0,0 +1,28 @@
+'use strict'
+
+const { storage } = require('./storage')
+
+function withRoutingContext (options, fn) {
+  if (!options || !options.ddApiKey) {
```
sabrenner:
```suggestion
  if (!options?.ddApiKey) {
```

packages/dd-trace/src/llmobs/tagger.js
```diff
@@ -110,6 +113,16 @@ class LLMObsTagger {
     // apply annotation context name
     const annotationContextName = annotationContext?.name
     if (annotationContextName) this._setTag(span, NAME, annotationContextName)
+
+    const routing = getCurrentRouting()
```
sabrenner: we can read directly from `storage.getStore()`. maybe we can make a private function `#getStorageProperties` that does this for `annotationContext`, and then `routing` as well.

packages/dd-trace/src/llmobs/tagger.js
```diff
@@ -110,6 +113,16 @@ class LLMObsTagger {
...
+    if (routing) {
+      if (routing.apiKey) {
+        this._setTag(span, ROUTING_API_KEY, routing.apiKey)
```
sabrenner: i would say since `routing.apiKey` _should_ exist, we can remove the conditional block here, and if for some reason it's not set, we probably want it to fail over

packages/dd-trace/src/llmobs/routing-context.js
```diff
@@ -0,0 +1,28 @@
...
+    routingContext: {
+      apiKey: options.ddApiKey,
+      site: options.ddSite
+    }
+  }
+
+  return storage.run(store, fn)
+}
+
+function getCurrentRouting () {
```
sabrenner: i don't think this function is needed - the only place in source it is used can be replaced by this `storage.getStore()?.routingContext` call directly, and I'll leave some suggestions for the tests (edit: the suggestion is to just remove the test file that calls this function as it is inherently tested by the other test file i think 😅)

packages/dd-trace/src/llmobs/writers/base.js
```diff
@@ -42,48 +43,118 @@ class BaseLLMObsWriter {
...
+  }
+
   get url () {
     if (this._agentless == null) return null
+    return this.#getUrlForRouting()
+  }
 
-    const baseUrl = this._baseUrl.href
-    const endpoint = this._endpoint
+  // Backward-compatible getters - only access default routing buffer.
```
sabrenner: i think ideally we don't want to access `_buffer` in tests - let's make a ticket to clean it up so we can make this private (something i can take on, unless you want to! 😅)

packages/dd-trace/src/llmobs/writers/base.js
```diff
@@ -42,48 +43,118 @@ class BaseLLMObsWriter {
...
+
+      const events = buffer.events
+      buffer.events = []
+      buffer.size = 0
+
+      const payload = this._encode(this.makePayload(events))
+      const options = this._getOptions(buffer.routing)
+      const url = this.#getUrlForRouting(buffer.routing)
+
+      log.debug('Encoded LLMObs payload for %s: %s', this.#getMaskedRoutingKey(buffer.routing), payload)
```
sabrenner: if we are sending data to the agent proxy, and we log this, there is no site, and instead we'll log a masked routing key of something like `****:datadoghq.com`, since there will be no api key and a default site. i don't think this conveys much info so i wonder how valuable the extra info might be

i have an idea for this, so i'll put it as a comment on this file

packages/dd-trace/src/llmobs/writers/base.js
sabrenner, line 1: a lot of this logic seems a bit confusing, although i do understand it. i wonder if we can do something like this

1. keep the `buffer` and `bufferSize` as the default buffer information for the writer.
2. have the buffers map be specific for multi-tenant buffers, and maybe re-named appropriately, like `#multiTenantBuffers`
3. When we enqueue an event, we can do pretty much what you already have. If there is no routing information passed in, we do all operations on the default/global writer buffer. Otherwise, get the multi-tenant routing key, appropriate buffer, and add to it instead.
4. When we flush, do the following:
    1. Flush the main writer buffer as we normally do before this change
    2. Iterate over all of the multi-tenant buffers if there are any. For each one we do, maybe we have a log like `Encoding and flushing multi-tenant buffer for {site} with {truncated api key}`, and then similar `Encoded LLMObs payload ...` logs subsequently.

I think that would get rid of the need for `get`/`set` on buffers and buffer size, and `url`. The writer will have one default buffer/size and url (as we did before, depending on agentless or not), and then when we append items, they're done conditionally on if there was any routing information for the given event, and when we flush, we do the global buffer first, then multi-tenant buffers if there are any, with logs denoting which is which.

Lmk if that makes sense! I think it'll just let us clean everything up a bit

packages/dd-trace/test/llmobs/routing-context.spec.js
sabrenner, line 1: i don't think this test file is needed, its behavior is inherently tested in the `multi-tenant.spec.js` file

packages/dd-trace/test/llmobs/writers/multi-tenant.spec.js
sabrenner, line 1: I think we can stick this in the [writers test folder](https://github.com/DataDog/dd-trace-js/tree/master/packages/dd-trace/test/llmobs/writers)

sabrenner (commented):
mostly small comments but one larger implementation comment i'm curious on your thoughts on. but overall the logic makes sense to me and the sdk api change looks good pending some type definition updates!

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -447,6 +447,32 @@ class LLMObs extends NoopLLMObs {
...
+  withRoutingContext (options, fn) {
+    if (!this.enabled) return fn()
+    if (!options?.ddApiKey) {
+      throw new Error('ddApiKey is required for routing context')
+    }
+    const currentStore = storage.getStore()
```
sabrenner, line 450: what if someone nests routing contexts
```javascript
llmobs.withRoutingContext({ ...tenant1Options }, () => {
  // create span1
  llmobs.withRoutingContext({ ...tenant2Options }, () => {
    // create span2
  })
})
```
i think as it is now, `span1` would be routed to `tenant1`, and `span2` would be routed to `tenant2`, although I'm not sure if a user would expect that, or for `span2` to be sent to _both_ `tenant1` and `tenant2`.

maybe we can/should throw if there is already an active routing context? lmk what you think, but i think it could discourage the above if it's not a situation we want to promote. we can always relax that restraint in the future, but maybe we wanna make sure users get a good understanding & signals from our API. 

but lmk if you disagree, i'm also fine to keep it as is, maybe with a warning log or smth but still going through with doing the nested routing context

packages/dd-trace/src/llmobs/writers/base.js
sabrenner, line 1: i think this file looks a ton better now imho, thanks for cleaning it up!

we can make all of the new private methods truly private i think (ie `#buildUrl` instead of `_buildUrl`, for example, and then call it with `this.#buildUrl`)

PROFeNoM: Oh ok I was actually wondering about that. I didn't what to do so I used the underscore, but i'll go ahead and change that

packages/dd-trace/src/llmobs/sdk.js
PROFeNoM: Your understanding is correct. As shown in the `nested contexts override outer and restore after` test, the inner routing context overrides the outer one, and restores after exit. It can indeed be considered a "silent override".

I personally believe that sending to both is problematic. It's complex, potentially dangerous (more easily to do an accidental data leakage), and harder (from a DEV-X) to reason about imo.

I believe that temporarily switching routing for a subset of operation can be a valid use case, hence why I believe that throwing may not be appropriate.
However, it may indeed be useful to provide visibility and to alert users of a potentially unintended behavior.

Right now, based on your suggestion, I've added a warning message:
```
Nested routing context detected. Inner context (****-key2) will override outer context (****-key1).  
Spans created in the inner context will only be sent to the inner context.
```

WDYT?

sabrenner: > I personally believe that sending to both is problematic. It's complex, potentially dangerous (more easily to do an accidental data leakage), and harder (from a DEV-X) to reason about imo.

+1

> I believe that temporarily switching routing for a subset of operation can be a valid use case, hence why I believe that throwing may not be appropriate.

ok sounds good! i think a warning log can be fine, and if it become more popular of a use case we can demote it to a debug log or something since it might be nice to have some kind of logged signal still in that case. but im good with a warning log 😄 

but i think we can remove the redacted api key bits, idt it's needed for this warning log, but lmk what you think. i think just

```
[LLM Observability] Nested routing context detected. Inner context will override outer context.  
Spans created in the inner context will only be sent to the inner context.
```

is good

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -447,6 +447,43 @@ class LLMObs extends NoopLLMObs {
...
+    if (currentStore?.routingContext) {
+      const outerKey = currentStore.routingContext.apiKey
+      const innerKey = options.ddApiKey
+      const maskedOuter = outerKey ? `****${outerKey.slice(-4)}` : ''
+      const maskedInner = innerKey ? `****${innerKey.slice(-4)}` : ''
```
sabrenner: per my response, if you're ok with it, i think we can forgo logging redacted api keys, and just do the log

packages/dd-trace/test/llmobs/writers/multi-tenant.spec.js
```diff
@@ -0,0 +1,174 @@
...
+    const payload = request.getCall(0).args[0]
+    assert.ok(!payload.includes('secret-tenant-key'))
+    assert.ok(!payload.includes('default-key'))
+  })
+
+  describe('routing context behavior', () => {
+    const { storage } = require('../../../src/llmobs/storage')
+    let sdkLogger
+
+    function withRoutingContext (options, fn) {
```
sabrenner: can these tests be done against the actual sdk? i don't think it's right to re-define the function here and use it in the tests below

AGENTS.md
```diff
@@ -115,6 +115,11 @@ assertObjectContains(response, { status: 200, body: { user: { name: 'Alice' } }
 - Files: kebab-case
 - JSDoc: TypeScript-compatible syntax (`@param {string}`, `@returns {Promise<void>}`, `@typedef`)
 
+### Class Properties and Methods
+- Use `#privateField` syntax for private properties and methods in new code
+- Do not use underscore prefix (`_property`) for new private members
+- Existing underscore properties may remain for backward compatibility
```
BridgeAR:
```suggestion
- Existing underscore properties should be refactored, when a file is touched for other
  code changes, if no other files rely on these properties (test files do not count).
```

packages/dd-trace/src/llmobs/writers/base.js
```diff
@@ -42,48 +44,110 @@ class BaseLLMObsWriter {
...
-    const baseUrl = this._baseUrl.href
-    const endpoint = this._endpoint
-
-    // Split on protocol separator to preserve it
-    // path.join will remove some slashes unnecessarily
-    const [protocol, rest] = baseUrl.split('://')
-    return protocol + '://' + path.join(rest, endpoint)
+  _getBuffer (routing) {
+    if (!routing?.apiKey) {
+      return { events: this._buffer, size: this._bufferSize, isDefault: true }
```
sabrenner: nit: to help with typing, we could turn this into a small class encapsulating this data

```javascript
class Buffer {
  constructor ({ events, size, routing, isDefault }) {
    this.events = events
    this.size = size
    this.routing = routing ?? {}
    this.isDefault = isDefault ?? false
  }
}
```

although it doesn't matter too much, so feel free to ignore

packages/dd-trace/src/llmobs/writers/base.js
```diff
@@ -42,48 +44,110 @@ class BaseLLMObsWriter {
...
+    }
+    const apiKey = routing.apiKey
+    if (!this.#multiTenantBuffers.has(apiKey)) {
+      this.#multiTenantBuffers.set(apiKey, {
+        events: [],
+        size: 0,
+        routing
+      })
+    }
+    return this.#multiTenantBuffers.get(apiKey)
```
sabrenner:
```suggestion
    const apiKey = routing.apiKey
    let buffer = this.#multiTenantBuffers.get(apiKey)
    if (!buffer) {
       buffer = { events: [], size: 0, routing }
       // buffer = new Buffer({  events: [], size: 0, routing }) // if doing the above class-based approach
      
      this.#multiTenantBuffers.set(apiKey, buffer)
    }
    
    return buffer
```

small nit so we only do one access from the map

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -447,6 +447,38 @@ class LLMObs extends NoopLLMObs {
...
+  /**
+   * Execute a function within a routing context, directing all LLMObs spans
+   * created within to a specific Datadog organization.
+   *
+   * @param {Object} options - Routing options
+   * @param {string} options.ddApiKey - The Datadog API key for the target org
+   * @param {string} [options.ddSite] - The Datadog site (e.g., 'datadoghq.eu')
+   * @param {Function} fn - The function to execute within the routing context
+   * @returns {*} The return value of fn
+   */
```
sabrenner: this docstring isn't needed now as it's documented in `index.d.ts`

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -447,6 +447,38 @@ class LLMObs extends NoopLLMObs {
...
+  withRoutingContext (options, fn) {
+    if (!this.enabled) return fn()
+    if (!options?.ddApiKey) {
+      throw new Error('ddApiKey is required for routing context')
+    }
+    const currentStore = storage.getStore()
+    if (currentStore?.routingContext) {
+      logger.warn(
+        'Nested routing context detected. Inner context will override outer context. ' +
```
sabrenner:
```suggestion
        '[LLM Observability] Nested routing context detected. Inner context will override outer context. ' +
```

our logs aren't namespaced at all by default so i like to add these demarcations here to signal that it's an LLMObs log

packages/dd-trace/test/llmobs/writers/multi-tenant.spec.js
```diff
@@ -0,0 +1,202 @@
...
+    })
+
+    after(() => {
+      delete process.env.DD_API_KEY
+      delete process.env.DD_SITE
+      llmobs.disable()
+      agent.wipe()
+    })
+
+    it('nested contexts route spans correctly and log warning', () => {
```
sabrenner, line 166: looks like these tests are [broken](https://github.com/DataDog/dd-trace-js/actions/runs/20820100574/job/60078402094?pr=7158) in CI - i can help take a look if needed!

sabrenner (commented):
looks good to me now! just nits, plus some of the nested routing context tests are failing, i'm happy to help with looking into those, so lmk

packages/dd-trace/src/llmobs/writers/base.js
PROFeNoM: Note: had to change tests that were directly accessing `_buffer`

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -447,6 +447,28 @@ class LLMObs extends NoopLLMObs {
     return storage.run(store, fn)
   }
 
+  withRoutingContext (options, fn) {
```
sabrenner: just before we lock in this api - while i think "context" maps well to a node.js environment, i don't think "with" does (as it's more of a python thing)

if u agree, we can also change in `index.d.ts` and tests, but def lmk if you disagree!
```suggestion
  routingContext (options, fn) {
```

packages/dd-trace/src/llmobs/writers/base.js
```diff
@@ -42,48 +60,109 @@ class BaseLLMObsWriter {
...
+        url: new URL(format({
+          protocol: 'https:',
+          hostname: `${this._intake}.${site}`
+        })),
+        path: this._baseEndpoint
+      }
+      const url = this.#buildUrl(options.url.href, options.path)
+      const maskedApiKey = apiKey ? `****${apiKey.slice(-4)}` : ''
+
+      log.debug('Encoding and flushing multi-tenant buffer for %s', maskedApiKey)
```
sabrenner:
```suggestion
      log.debug('Encoding and flushing multi-tenant %s buffer for %s', this.constructor.name, maskedApiKey)
```

so it's something like `Encoding and flushing multi-tenant LLMObsSpanWriter buffer for ****abcd`

PR #7168, opened 2026-01-01
AppSec optimizations

simon-id:
### What does this PR do?
Optimize an old TODO that no one else bothered fixing. It prevents a few addresses to be sent to the WAF if they're empty objects, and it prevents body and cookies to be analyzed in the end-of-request handler if the same object pointer has been analyzed by a just-in-time WAF call earlier.

### Performance impact: it reduces the amount of time spent in WAF/bindings when running the default system test scenario **by 21.7%**

### DIFF
--- a/.github/workflows/all-green.yml
+++ b/.github/workflows/all-green.yml
@@ -26,3 +26,33 @@ jobs:
           checks_exclude: devflow.*
           fail_fast: false
           verbose: true # What checks are still waited for?
+      - name: Require vendor validation when vendor/ changes
+        if: github.event_name == 'pull_request'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          set -euo pipefail
+
+          vendor_changed="$(gh api "repos/${REPO}/compare/${BASE_SHA}...${HEAD_SHA}" --jq 'any(.files[].filename; startswith("vendor/"))')"
+          if [ "$vendor_changed" != "true" ]; then
+            exit 0
+          fi
+
+          # If vendor/ was touched, require *some* vendor validation check to be present:
+          # - For most PRs: `Validate vendored bundle / validate-vendored-bundle`
+          # - For Dependabot vendor PRs: `Dependabot Automation / vendor-validate`
+          check_names="$(gh api "repos/${REPO}/commits/${{ github.event.pull_request.head.sha }}/check-runs" \
+            -H "Accept: application/vnd.github+json" \
+            --paginate \
+            --jq '.check_runs[].name')"
+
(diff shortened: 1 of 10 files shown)


packages/dd-trace/src/appsec/index.js
```diff
@@ -120,27 +122,37 @@ function onRequestBodyParsed ({ req, res, body, abortController }) {
...
 
+  if (!Object.keys(body).length) return
+
   const results = waf.run({
     persistent: {
       [addresses.HTTP_INCOMING_BODY]: body
     }
   }, req)
 
+  analyzedBodies.add(body)
```
CarlesDD: If body is a primitive (e.g. string), this will thow a TypeError, since WeakSet only accepts objects.

packages/dd-trace/src/appsec/index.js
```diff
@@ -178,19 +190,32 @@ function incomingHttpEndTranslator ({ req, res }) {
...
 
   // we need to keep this to support other body parsers
-  // TODO: no need to analyze it if it was already done by the body-parser hook
-  if (req.body !== undefined && req.body !== null) {
+  if (
+    req.body !== undefined &&
+    req.body !== null &&
+    !analyzedBodies.has(req.body) &&
+    Object.keys(req.body).length
+  ) {
```
CarlesDD: Nit: this condition is getting quite complex (null checks, type checks, WeakSet, emptiness). Extracting it into a helper would make the intent clearer

packages/dd-trace/src/appsec/index.js
```diff
@@ -178,19 +190,32 @@ function incomingHttpEndTranslator ({ req, res }) {
...
   }
 
   // we need to keep this to support other cookie parsers
-  if (req.cookies !== null && typeof req.cookies === 'object') {
+  if (
+    req.cookies !== null &&
+    typeof req.cookies === 'object' &&
+    !analyzedCookies.has(req.cookies) &&
+    Object.keys(req.cookies).length
+  ) {
```
CarlesDD, line 211: Same nit here.

packages/dd-trace/src/appsec/index.js
CarlesDD, line 1: These changes add checks to avoid analyzing empty objects, which is great. As a small optimization, `Object.keys(obj).length` allocates an array just to check emptiness. Given this runs in a hot path, a `for…in` early-return check could be more efficient, especially for large objects.
```
for (const _ in obj) return true
return false
```

simon-id: After discussing with the current worldwide expert in object key counting, turns out `Object.keys().length` is the best way currently available.

simon-id: Turns out I'm now the worldwide expert, replacing the previous incorrect worldwide ""expert"". You were right, and I changed it

packages/dd-trace/src/appsec/index.js
simon-id: Pushed a fix!

CarlesDD (approved):
LGTM

PR #7171, opened 2026-01-05
feat(aap): update tags for inferred proxy spans

CarlesDD:
### What does this PR do?

- API Gateway v2 (HTTP API) support: Added `aws-httpapi` to the supported proxies map, generating spans with name `aws.httpapi` (in addition to existing `aws.apigateway` for REST API v1).

- Span attributes improvements:
    - Added **span.kind**: `server` to inferred proxy spans
    - Prefixed **http.url** with `https://` scheme
    - Added support for new `x-dd-proxy-resource-path` header to populate **http.route** tag
    - **resource.name** now uses the route pattern (from `x-dd-proxy-resource-path`) when available, falling back to the path

- AppSec integration:
    - Added `_dd.appsec.enabled: 1` metric to inferred proxy spans when AppSec is enabled
    - Propagated appsec events to inferred proxy spans

- Additional span tags (when corresponding headers are present):
    - `account_id` from `x-dd-proxy-account-id`
    - `apiid` from `x-dd-proxy-api-id`
    - `region` from `x-dd-proxy-region`
    - `aws_user` from `x-dd-proxy-user`
    - `dd_resource_key` computed as API Gateway ARN:
        - REST API (v1): `arn:aws:apigateway:{region}::/restapis/{api-id}`
        - HTTP API (v2): `arn:aws:apigateway:{region}::/apis/{api-id}`

### Motivation
Implementation of RFC-1081 on Endpoint Discovery & Correlation from API Gateway Inferred Spans, summary of changes:

image

### Additional Notes
[APPSEC-60015](https://datadoghq.atlassian.net/browse/APPSEC-60015)

System test: https://github.com/DataDog/system-tests/pull/6245
System test: https://github.com/DataDog/system-tests/pull/5935

[APPSEC-60015]: https://datadoghq.atlassian.net/browse/APPSEC-60015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dd-trace/src/appsec/index.js
```diff
@@ -168,6 +168,13 @@ function incomingHttpStartTranslator ({ req, res, abortController }) {
     [HTTP_CLIENT_IP]: clientIp,
   })
 
+  const context = web.getContext(req)
+  if (context?.inferredProxySpan) {
+    context.inferredProxySpan.addTags({
+      '_dd.appsec.enabled': 1,
+    })
+  }
```
uurien: I think that this would be better in performance if we not get the context if inferredProxySpans are not enabled, the config is already loaded, is false by default and `web.getContext(req)` finds in a Map.

```suggestion
  if (config.inferredProxyServicesEnabled) {
    const context = web.getContext(req)
    if (context?.inferredProxySpan) {
      context.inferredProxySpan.addTags({
        '_dd.appsec.enabled': 1,
      })
    }
  }
```

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -362,6 +362,14 @@ function reportAttack ({ events: attackData, actions }) {
 
   rootSpan.addTags(newTags)
 
+  // Add _dd.appsec.json tag to inferred proxy span
+  const context = web.getContext(req)
```
uurien: same here, but we would need more changes because to reporter receives only the appsec config.

packages/dd-trace/test/plugins/util/inferred_proxy.spec.js
```diff
@@ -159,6 +195,112 @@ describe('Inferred Proxy Spans', function () {
       })
     })
 
+    it('should create a parent span with aws.httpapi for API Gateway v2', async () => {
```
uurien: I miss some tests like these when appsec is enabled `_dd.appsec.json` and `_dd.appsec.enabled` is filled. Maybe in `test/appsec/index.spec.js`?

zarirhamza (dismissed):
The changes are all sufficient aside from the performance improvements mentioned before.

However, to get CI working, it seems that system-tests need to be updated to account for the new `https` string prefixed to the URLs. Until this change is rolled out to the other languages as well, we should avoiding changing the test entirely and just create a separate use case specific for JS to include the `https` prefix.

packages/dd-trace/test/plugins/util/inferred_proxy.spec.js
CarlesDD: Added a test

packages/dd-trace/src/appsec/index.js
```diff
@@ -168,6 +168,15 @@ function incomingHttpStartTranslator ({ req, res, abortController }) {
     [HTTP_CLIENT_IP]: clientIp,
   })
 
+  if (config.inferredProxyServicesEnabled) {
+    const context = web.getContext(req)
+    if (context?.inferredProxySpan) {
+      context.inferredProxySpan.addTags({
+        '_dd.appsec.enabled': 1,
+      })
```
simon-id: if you're setting only one tag you can use `setTag()` instead so you dont need to use an object for no reason right ?

packages/dd-trace/src/plugins/util/inferred_proxy.js
```diff
@@ -50,9 +61,15 @@ function createInferredProxySpan (headers, childOf, tracer, reqCtx, traceCtx, co
...
       component: proxySpanInfo.component,
       [SPAN_TYPE]: 'web',
+      [SPAN_KIND]: 'server',
       [HTTP_METHOD]: proxyContext.method,
-      [HTTP_URL]: proxyContext.domainName + proxyContext.path,
+      [HTTP_URL]: 'https://' + proxyContext.domainName + proxyContext.path,
       stage: proxyContext.stage,
+      ...(proxyContext.resourcePath && { [HTTP_ROUTE]: proxyContext.resourcePath }),
+      ...(proxyContext.accountId && { account_id: proxyContext.accountId }),
+      ...(proxyContext.apiId && { apiid: proxyContext.apiId }),
```
simon-id, line 75: there is account_id just above with an underscore, are you sure this isn't named api_id ?

integration-tests/appsec/inferred-proxy-spans.spec.js
```diff
@@ -0,0 +1,91 @@
...
+      // Server span
+      assert.strictEqual(spans[1].name, 'web.request')
+      assert.strictEqual(spans[1].metrics['_dd.appsec.enabled'], 1)
+    })
+  })
+
+  it('should add _dd.appsec.json to inferred proxy span when attack is detected', async () => {
+    await axios.get('/testattack', {
+      headers: inferredHeaders,
+      validateStatus: () => true,
```
simon-id: hold my beer, watch this
```suggestion
      validateStatus: false,
```

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -362,6 +364,16 @@ function reportAttack ({ events: attackData, actions }) {
 
   rootSpan.addTags(newTags)
 
+  // Add _dd.appsec.json tag to inferred proxy span
+  if (config.inferredProxyServicesEnabled) {
+    const context = web.getContext(req)
+    if (context?.inferredProxySpan) {
+      context.inferredProxySpan.addTags({
+        '_dd.appsec.json': newTags['_dd.appsec.json'],
+      })
```
simon-id: setTag() aqui tambien

packages/dd-trace/src/plugins/util/inferred_proxy.js
```diff
@@ -50,9 +61,15 @@ function createInferredProxySpan (headers, childOf, tracer, reqCtx, traceCtx, co
       service: proxyContext.domainName || tracer._config.service,
       component: proxySpanInfo.component,
       [SPAN_TYPE]: 'web',
+      [SPAN_KIND]: 'server',
       [HTTP_METHOD]: proxyContext.method,
-      [HTTP_URL]: proxyContext.domainName + proxyContext.path,
+      [HTTP_URL]: 'https://' + proxyContext.domainName + proxyContext.path,
```
simon-id, line 70: little bit of a breaking change no ?

packages/dd-trace/src/plugins/util/inferred_proxy.js
CarlesDD: Counterintuitively, it is the correct value. You can check the RFC and the STs.

packages/dd-trace/src/plugins/util/inferred_proxy.js
CarlesDD: Technically is a format change aligned with the RFC, and fixes an invalid `http.url` value (missing scheme per RFC-3986). The previous format could lead to incorrect backend parsing and route inference.
Adding the schema makes the value a valid absolute URI and enables the expected enrichment behavior.
Impact should be limited and justified as a bugfix rather than a breaking change.

packages/dd-trace/src/appsec/index.js
```diff
@@ -174,6 +174,13 @@ function incomingHttpStartTranslator ({ req, res, abortController }) {
     [HTTP_CLIENT_IP]: clientIp,
   })
 
+  if (config.inferredProxyServicesEnabled) {
+    const context = web.getContext(req)
+    if (context?.inferredProxySpan) {
+      context.inferredProxySpan.setTag('_dd.appsec.enabled', 1)
+    }
```
simon-id, line 181:
```suggestion
    context?.inferredProxySpan?.setTag('_dd.appsec.enabled', 1)
```

simon-id: actually nvm, i think it makes better semantic sense in your original change

simon-id (approved):
LGTM, but as i wrote before, maybe worth a little eye from whoever owns the inferred proxy code ? up to you

PR #7177, opened 2026-01-05
feat: process tags for client trace statistics

tlhunter:
### What does this PR do?
- adds process tags to client trace statistics

### Motivation
- [RFC](https://docs.google.com/document/d/1AFdLUuVk70i0bJd5335-RxqsvwAV9ovAqcO2z5mEMbA)

### DIFF
--- a/packages/dd-trace/src/encode/span-stats.js
+++ b/packages/dd-trace/src/encode/span-stats.js
@@ -95,7 +95,7 @@ class SpanStatsEncoder extends AgentEncoder {
   }
 
   _encode (bytes, stats) {
-    this._encodeMapPrefix(bytes, 8)
+    this._encodeMapPrefix(bytes, stats.ProcessTags ? 9 : 8)
 
     this._encodeString(bytes, 'Hostname')
     this._encodeString(bytes, stats.Hostname)
@@ -123,6 +123,11 @@ class SpanStatsEncoder extends AgentEncoder {
 
     this._encodeString(bytes, 'Sequence')
     this._encodeLong(bytes, stats.Sequence)
+
+    if (stats.ProcessTags) {
+      this._encodeString(bytes, 'ProcessTags')
+      this._encodeString(bytes, stats.ProcessTags)
+    }
   }
 }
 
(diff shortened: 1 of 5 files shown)


BridgeAR (approved):

packages/dd-trace/src/encode/span-stats.js
```diff
@@ -123,6 +123,9 @@ class SpanStatsEncoder extends AgentEncoder {
 
     this._encodeString(bytes, 'Sequence')
     this._encodeLong(bytes, stats.Sequence)
+
+    this._encodeString(bytes, 'ProcessTags')
+    this._encodeObject(bytes, stats.ProcessTags)
```
vandonr:
```suggestion
    this._encodeString(bytes, stats.ProcessTags)
```

the proto is defined here https://github.com/DataDog/datadog-agent/blob/main/pkg/proto/datadog/trace/stats.proto#L52-L55, the type is a string

packages/dd-trace/src/encode/span-stats.js
```diff
@@ -95,7 +95,7 @@ class SpanStatsEncoder extends AgentEncoder {
   }
 
   _encode (bytes, stats) {
-    this._encodeMapPrefix(bytes, 8)
+    this._encodeMapPrefix(bytes, 9)
```
vandonr: nit : this could still be 8 when process tags are disabled, if we don't write them to the payload (which would be cleaner in the sense that the behavior wouldn't change as long as the flag is disabled).

ajwerner:
While you're here, let me highlight https://github.com/DataDog/system-tests/pull/5949 -- for remote config it's a list of strings...

packages/dd-trace/src/encode/span-stats.js
```diff
@@ -123,6 +123,9 @@ class SpanStatsEncoder extends AgentEncoder {
 
     this._encodeString(bytes, 'Sequence')
     this._encodeLong(bytes, stats.Sequence)
+
+    this._encodeString(bytes, 'ProcessTags')
+    this._encodeString(bytes, stats.ProcessTags)
```
BridgeAR: Should these not be skipped if `stats.ProcessTags` is not defined?

BridgeAR (commented):
LGTM, it just needs a small rebase

BridgeAR (approved):

PR #7178, opened 2026-01-06
Changes for AI Integration Toolkit Project

wconti27:
### What does this PR do?
A few changes to make the AI integrations toolkit project a bit easier. Adds an abstraction layer function for setting up plugin tests. Also allows composite plugin constructor to take in a list of plugins. finally adds the ability to do type based assertions using assertObjectContains (helps with testing error tag `error.stacktrace` presence since the value may change between runs).

Opened in favor of #7023 due to requested changes from someone OOO until next week.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/integration-tests/helpers/index.js
+++ b/integration-tests/helpers/index.js
@@ -25,9 +25,9 @@ const { DEBUG } = process.env
 let shouldKill
 
 // Symbol constants for dynamic value matching in assertObjectContains
-const ANY_STRING = Symbol.for('test.ANY_STRING')
-const ANY_NUMBER = Symbol.for('test.ANY_NUMBER')
-const ANY_VALUE = Symbol.for('test.ANY_VALUE')
+const ANY_STRING = Symbol('test.ANY_STRING')
+const ANY_NUMBER = Symbol('test.ANY_NUMBER')
+const ANY_VALUE = Symbol('test.ANY_VALUE')
 
 /**
  * @param {string} filename


packages/dd-trace/test/setup/helpers/integration-test-helpers/index.js
```diff
@@ -0,0 +1,62 @@
...
+      await testSetup.setup(mod)
+    })
+
+    after(async () => {
+      if (testSetup?.teardown) {
+        await testSetup.teardown()
+      }
+    })
+
+    if (testCallback) {
```
bengl: In what situations would testCallback not be provided? I think you can omit this check, no?

packages/dd-trace/test/setup/helpers/integration-test-helpers/index.js
```diff
@@ -0,0 +1,62 @@
...
+      testCallback({
+        testSetup,
+        agent,
+        tracer,
+        assert,
+        describe,
```
bengl: What's the value in all these re-exports? If we're not modifying them, it just adds to confusion when reading a test file.

packages/dd-trace/test/setup/helpers/integration-test-helpers/index.js
```diff
@@ -0,0 +1,62 @@
...
+    })
+
+    before(async () => {
+      tracer = require('../../../../../dd-trace').init()
+      mod = require(`../../../../../../versions/${packageName}@${version}`)
+      mod = options.subModule ? mod.get(options.subModule) : mod.get()
+      await testSetup.setup(mod)
+    })
+
+    after(async () => {
```
bengl: Why bother with this, rather than just letting the caller call `after` directly, for clarity?

packages/dd-trace/test/setup/helpers/integration-test-helpers/index.js
```diff
@@ -0,0 +1,62 @@
...
+      await agent.close({ ritmReset: false })
+    })
+
+    before(async () => {
+      tracer = require('../../../../../dd-trace').init()
+      mod = require(`../../../../../../versions/${packageName}@${version}`)
+      mod = options.subModule ? mod.get(options.subModule) : mod.get()
+      await testSetup.setup(mod)
```
bengl: Why bother with this, rather than just letting the caller call `before` directly, for clarity?

packages/dd-trace/test/setup/helpers/integration-test-helpers/index.js
```diff
@@ -0,0 +1,62 @@
...
+  describe('Plugin', () => {
+    describe(pluginName, () => {
+      withVersions(pluginName, packageName, version => {
+        createVersionedTests(pluginName, packageName, testSetup, options, version, testCallback)
+      })
+    })
+  })
+}
+
+function createVersionedTests (pluginName, packageName, testSetup, options = {}, version, testCallback) {
```
bengl: This function is only ever called on line 12 of this file, so you might as well in-line it for clarity.

packages/dd-trace/test/setup/helpers/integration-test-helpers/index.js
```diff
@@ -0,0 +1,62 @@
+'use strict'
```
bengl: The fact that both this file and `integration-tests/helpers/index.js` both exist is confusing. Can you consolidate them?

packages/dd-trace/src/plugins/composite.js
```diff
@@ -3,11 +3,23 @@
...
 
 class CompositePlugin extends Plugin {
+  #pluginNames = []
+
   constructor (...args) {
     super(...args)
 
     for (const [name, PluginClass] of Object.entries(this.constructor.plugins)) {
-      this[name] = new PluginClass(...args)
+      // Handle case where value is an array of plugin classes
```
bengl: Why would we do this? `this.constructor.plugins` is already an array of plugin classes.

integration-tests/helpers/index.js
```diff
@@ -635,14 +640,56 @@ function setShouldKill (value) {
   })
 }
 
-// @ts-expect-error assert.partialDeepStrictEqual does not exist on older Node.js versions
-// eslint-disable-next-line n/no-unsupported-features/node-builtins
-const assertObjectContains = assert.partialDeepStrictEqual || function assertObjectContains (actual, expected, msg) {
-  if (expected === null || typeof expected !== 'object') {
-    assert.strictEqual(actual, expected, msg)
+// Internal implementation that handles asymmetric matchers (ANY_STRING, ANY_NUMBER, ANY_VALUE)
+function assertObjectContainsWithMatchers (actual, expected, msg) {
```
bengl: Can you show example usage of this? It seems more confusing and more potentially  error-prone than just looking at the properties we're interested in directly.

packages/dd-trace/test/setup/helpers/integration-test-helpers/index.js
wconti27: removed the check.

packages/dd-trace/test/setup/helpers/integration-test-helpers/index.js
wconti27: removed the unnecessary re exports, only keeping the test app, initialized tracer, initialized mock agent, and the loaded module.

packages/dd-trace/test/setup/helpers/integration-test-helpers/index.js
wconti27: Because all of this setup is boilerplate that doesn't change between different plugins. I don't see the need for the plugin tests to have to do any of this setup, it should just be handled for them.

packages/dd-trace/test/setup/helpers/integration-test-helpers/index.js
wconti27: Same comment as above.

packages/dd-trace/test/setup/helpers/integration-test-helpers/index.js
wconti27: done.

packages/dd-trace/test/setup/helpers/integration-test-helpers/index.js
wconti27: changed to plugin-test-helpers

packages/dd-trace/src/plugins/composite.js
wconti27: in situations where we have several plugins exported from one file, useful for when we are tracing similar operations, such as two clients that differ slightly and require separate plugins. Avoids us have to iterate the list of plugins to correctly make the Composite plugin map, example of code that we won't need anymore: https://github.com/DataDog/dd-trace-js/blob/2681a5e0a5432716173460bcc3192391be8acb61/packages/datadog-plugin-langchain/src/index.js#L13

integration-tests/helpers/index.js
wconti27: it allows us to use the assertObjectContains helper and assert the presence of a tag while still being able to use the object version of the assertion. Makes the test code much cleaner. 

Its used in the Bullmq pr I have up, but i couldn't copy a link to the highlight 😢 . The test is in this file: https://github.com/DataDog/dd-trace-js/pull/7123/files#:~:text=packages/datadog%2Dplugin%2Dbullmq/test/index.spec.js

```
    it('should generate span with error tags (error path)', async () => {
      const traceAssertion = agent.assertFirstTraceSpan({
        name: 'bullmq.addBulk',
        error: 1,
        meta: {
          'span.kind': 'producer',
          'messaging.system': 'bullmq',
          'error.type': ANY_STRING,
          'error.message': ANY_STRING
        }
      })

      try {
        await testSetup.queueAddBulkError()
      } catch (err) {
        // Expected error
      }

      return traceAssertion
    })
```

packages/dd-trace/src/plugins/composite.js
BridgeAR: I would rather keep the current behavior as it is in the example and not add this functionality.

packages/dd-trace/test/plugins/composite.spec.js
```diff
@@ -0,0 +1,100 @@
+'use strict'
+
+const assert = require('node:assert')
+const { describe, it, beforeEach, afterEach } = require('tap').mocha
```
BridgeAR: Is it possible to use mocha instead (I believe it depends on how we start the test)? We are in the process of removing tap.

integration-tests/helpers/index.js
BridgeAR: Is the error type and message here not a fixed value that we could check for explicitly?

integration-tests/helpers/index.js
```diff
@@ -671,14 +718,29 @@ const assertObjectContains = assert.partialDeepStrictEqual || function assertObj
...
       assert.ok(actual, msg)
       assert.strictEqual(actual[key], expected[key], msg)
     }
   }
 }
 
+// Main assertObjectContains: tries fast path first, falls back to matcher-aware path if needed
+const assertObjectContains = function assertObjectContains (actual, expected, msg) {
+  try {
+    assertObjectContainsStrict(actual, expected, msg)
```
BridgeAR: While I am not yet sure about adding the helpers is a good idea, if we do want to add them, we can just run assert.partialDeepStrictEqual here if existent and fallback, to the asymmetric matcher if not.
That prevents code duplication and achieves the good error messages.

That is what I meant in the other PR comment :)

packages/dd-trace/test/plugins/composite.spec.js
wconti27: done.

packages/dd-trace/src/plugins/composite.js
bengl: Ah, I see, `.plugins` is actually an object. My mistake.

Still, can't the llmobs plugin files just export the appropriate collections (i.e. objects) to be added to `.plugins`?

integration-tests/helpers/index.js
wconti27: Sorry, my assertion was wrong, they are indeed something we can check! I updated the bullmq pr to have exact assertions for `error message` and `error type`. But for `error.stackTrace`, we cannot do exact assertions since the stack trace will change according to where the test is run (CI vs my local setup vs your local setup, will all have different stack traces!)

```
    it('should generate span with error tags (error path)', async () => {
      const traceAssertion = agent.assertFirstTraceSpan({
        name: 'bullmq.add',
        error: 1,
        meta: {
          'span.kind': 'producer',
          'messaging.system': 'bullmq',
          'error.type': 'Error',
          'error.message': 'Validation error, cannot resolve alias "inv"',
          'error.stack': ANY_STRING
        }
      })
```
link: https://github.com/DataDog/dd-trace-js/pull/7123/files#diff-18d85ce9ed2f37175e92429ff80c48b3bcd6828542884bd82732df35464b856d:~:text=packages/datadog%2Dplugin%2Dbullmq/test/index.spec.js

integration-tests/helpers/index.js
wconti27: I think i made the change that you wanted now.

packages/dd-trace/test/plugins/composite.spec.js
wconti27: removed this file.

packages/dd-trace/src/plugins/composite.js
wconti27: removed this change

packages/dd-trace/test/setup/helpers/integration-test-helpers/index.js
bengl: This is for `after`, and there's nothing in here except calling the teardown if it exists, meaning this could just be an `after()` in the calling code whenever needed. For the setup case, `before()`s are called sequentially, so you _could_ just remove `await testSetup.setup(mod)` and then have the calling code add their own `before()`, just the same.

integration-tests/helpers/index.js
bengl: This test API is confusing. Can we not just use `node:assert` functions? Yes, doing complex tests on an object is slightly more code, but that extra code makes everything clearer.

packages/dd-trace/test/setup/helpers/plugin-test-helpers/index.js
```diff
@@ -0,0 +1,42 @@
...
+          before(async () => {
+            tracer = require('../../../../../dd-trace').init()
+            mod = require(`../../../../../../versions/${packageName}@${version}`)
+            mod = options.subModule ? mod.get(options.subModule) : mod.get()
+          })
+
+          testCallback({
+            agent,
+            tracer,
+            mod
```
bengl: `mod` gets reassigned in the `before` block above. AFAICT the value will always be `null` in this object that's passed to `testCallback()`.

wconti27: Should be fixed now.

integration-tests/helpers/index.js
```diff
@@ -24,6 +24,11 @@ const { DEBUG } = process.env
 // This is set by the setShouldKill function
 let shouldKill
 
+// Symbol constants for dynamic value matching in assertObjectContains
+const ANY_STRING = Symbol.for('test.ANY_STRING')
+const ANY_NUMBER = Symbol.for('test.ANY_NUMBER')
+const ANY_VALUE = Symbol.for('test.ANY_VALUE')
```
bengl: Non-blocker, but these don't have to be `Symbol.for`, since you're exporting them anyway. You can just use `Symbol('test.ANY_STRING')`, etc.

In fact, that's preferable here, considering the symbol text is so generic.

bengl (dismissed):

integration-tests/helpers/index.js
wconti27: Fixed this.

PR #7180, opened 2026-01-06
Fix: updating remote config manager to send process_tags as an array

crysmags:
### What does this PR do?

Fixes a bug in remote config where the tracer was sending `process_tags` as an object, but the Agent expects an array of strings.
```
[Error: [RC] Error in request]
NoTransmitError: Error from http://127.0.0.1:8136/v0.7/config: 400 Bad Request. Response from the endpoint: "json: cannot unmarshal object into Go struct field ClientTracer.client.client_tracer.process_tags of type []string
"
```

### DIFF
--- a/packages/dd-trace/src/process-tags/index.js
+++ b/packages/dd-trace/src/process-tags/index.js
@@ -40,8 +40,9 @@ function getProcessTags () {
 
   for (const [key, value] of tags) {
     if (value !== undefined) {
-      tagsArray.push(`${key}:${sanitize(value)}`)
-      tagsObject[key] = sanitize(value)
+      const sanitizedValue = sanitize(value)
+      tagsArray.push(`${key}:${sanitizedValue}`)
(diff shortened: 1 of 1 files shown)


packages/dd-trace/src/process-tags/index.js
```diff
@@ -35,19 +35,23 @@ function getProcessTags () {
...
-  const serialized = serialize(tags)
+  const tagsArray = []
+  const tagsObject = {}
 
-  const tagsObject = tags.reduce((acc, [key, value]) => {
+  for (const [key, value] of tags) {
     if (value !== undefined) {
-      acc[key] = value
+      tagsArray.push(`${key}:${sanitize(value)}`)
+      tagsObject[key] = value
```
BridgeAR: I think this also has to be sanitized :)
(Using a variable to only do it once)

BridgeAR (approved):

PR #7181, opened 2026-01-06
ci: update openldap image with maintained one

BridgeAR:
Alternative to https://github.com/DataDog/dd-trace-js/pull/6569
Closes: #6569

### DIFF
--- a/.github/workflows/appsec.yml
+++ b/.github/workflows/appsec.yml
@@ -78,14 +78,14 @@ jobs:
       PLUGINS: ldapjs
     services:
       openldap:
-        image: cleanstart/openldap:latest
+        image: registry.opensuse.org/opensuse/openldap
         ports:
           - '1389:389'
           - '1636:636'
         env:
           LDAP_DOMAIN: 'example.org'
           LDAP_ADMIN_PASSWORD: 'adminpassword'
-          LDAP_ORGANISATION: 'Example Inc'
+          LDAP_CONFIG_PASSWORD: 'adminpassword'
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/appsec.yml
```diff
@@ -78,15 +78,14 @@ jobs:
       PLUGINS: ldapjs
     services:
       openldap:
-        image: bitnamilegacy/openldap:latest
+        image: registry.opensuse.org/opensuse/openldap
         ports:
-          - '1389:1389'
-          - '1636:1636'
+          - '1389:389'
+          - '1636:636'
```
simon-id, line 84: nit: doesn't that mean the opensuse image is running on root ? is that gonna be a sec problem ? prob not?

.github/workflows/appsec.yml
```diff
@@ -78,15 +78,14 @@ jobs:
       PLUGINS: ldapjs
     services:
       openldap:
-        image: bitnamilegacy/openldap:latest
+        image: registry.opensuse.org/opensuse/openldap
```
simon-id, line 81: should we pin it to a specific hash if that's even possible ? the internal sec team always come asking us to use pinned deps right

.github/workflows/appsec.yml
BridgeAR: It is the default port for openldap and the other port is not supported by the container by default. It should not be a security concern either.

.github/workflows/appsec.yml
BridgeAR: I agree with that and think we should handle it for all of our containers in a separate step

PR #7188, opened 2026-01-07
feat: add runtime detection for bun

zmzlois:
### What does this PR do?
Detect if user is using Bun and log additional information. The plan is to add logging for now, and as soon as each one of the pieces are done we turn of the flag 

### Motivation
This is the first initial PR to add Bun support. The roadmap identified for dd-trace including: 
- `pprof-nodejs` compatible Bun bindings 
- `dd-native-metrics-js` 
- `heap_snapshot` + `runtime_metrics` from `v8` (this one might be a bit bumpy to add) 
potentially also: 
- for firewall `native_appspec` 
- for security analysis`native-iast-taint-tracking`
must add but esbuild/bun has limited support, will have to see what comes out of the runtime API: 
- module loading, it'd be even harder if the output it's a binary itself 
- extend `AsyncLocalStorage` for context propagation 
- `worker_threads` related 
- `perf_hooks` 
- WASM rewriter 

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### Additional Notes

I think @szegedi or @rochdev might want to have a look at this? (pardon my linter issue, will fix)

### DIFF
--- a/version.js
+++ b/version.js
@@ -22,7 +22,7 @@ module.exports = {
   NODE_PATCH: parseInt(nodeMatches[3]),
   NODE_VERSION: nodeMatches[0],
   RUNTIME: runtime.runtimeName,
-  RUNTIME_MAJOR: parseInt(runtimeVersion[1]), // in case it breaks downstream consumer 
+  RUNTIME_MAJOR: parseInt(runtimeVersion[1]), // in case it breaks downstream consumer
   RUNTIME_MINOR: parseInt(runtimeVersion[2]),
   RUNTIME_PATCH: parseInt(runtimeVersion[3]),
   RUNTIME_VERSION: runtime.runtimeVersion


zmzlois:
i suppose i could add a feature flag for bun so user can 
```
tracer.init({ experimental: { bun: { enabled: true } } })
```
if it's needed i could add it too

packages/dd-trace/src/config.js
```diff
@@ -287,6 +288,39 @@ class Config {
...
+        this.runtimeMetrics = { ...this.runtimeMetrics, enabled: false }
+        log.warn(
+          'Runtime metrics are disabled on Bun runtime (requires V8). We are adding support for Bun soon.'
+        )
+      }
+
+      // Auto-instrumentation may be limited
+      if (this.tracing !== false) {
+        log.warn(
+          'Auto-instrumentation may be limited on Bun runtime. We are adding support for Bun soon.'
```
szegedi: I'd rather we don't make promises like this.
```suggestion
          'Profiling is disabled on Bun runtime (requires V8).'
        )
      }

      // Runtime metrics won't work (V8-specific)
      if (this.runtimeMetrics?.enabled !== false) {
        this.runtimeMetrics = { ...this.runtimeMetrics, enabled: false }
        log.warn(
          'Runtime metrics are disabled on Bun runtime (requires V8).'
        )
      }

      // Auto-instrumentation may be limited
      if (this.tracing !== false) {
        log.warn(
          'Auto-instrumentation may be limited on Bun runtime.'
```

packages/dd-trace/src/config.js
```diff
@@ -287,6 +288,39 @@ class Config {
...
+      isNode: !isBun,
+    }
+
+    // Disable incompatible features on Bun
+    if (isBun) {
+      // Profiling won't work (V8-specific)
+      if (this.profiling?.enabled !== false) {
+        this.profiling = { ...this.profiling, enabled: false }
+        log.warn(
+          'Profiling is disabled on Bun runtime (requires V8). We are adding support for Bun soon.'
```
BridgeAR, line 304: There is no guarantee for Bun support and everything that is done here is a best effort.
Stating that we officially plan on doing that is as such not possible.

The same about the other comments.

packages/dd-trace/src/guardrails/index.js
```diff
@@ -19,6 +20,29 @@ function guard (fn) {
...
+        { name: 'abort', tags: ['reason:incompatible_runtime'] },
+        { name: 'abort.runtime', tags: ['runtime:' + runtime.runtimeName] }
+      ], undefined, {
+        result: 'abort',
+        result_class: 'incompatible_runtime',
+        result_reason: 'Bun runtime detected. Some features may not work. Use DD_INJECT_FORCE=true to proceed.'
+      }
+    )
+    log.warn(
+      'Bun runtime detected. Some dd-trace features are not yet fully supported.'
```
BridgeAR, line 35:
```suggestion
      'Bun runtime detected. Some dd-trace features may not work as expected.'
```

packages/dd-trace/src/utils/runtime.js
```diff
@@ -0,0 +1,45 @@
+'use strict'
+
+/**
+ * Detects the JavaScript runtime environment
+ * @returns {{ name: string, version: string, isBun: boolean, isNode: boolean }}
+ */
+function detectRuntime () {
+  // Bun sets process.isBun = true
+  if (typeof process !== 'undefined' && process.isBun === true) {
```
BridgeAR, line 9: We can just require `process`. That way it should always be available.

packages/dd-trace/src/utils/runtime.js
```diff
@@ -0,0 +1,45 @@
...
+    return {
+      name: 'bun',
+      version: process.versions?.bun || 'unknown',
+      isBun: true,
+      isNode: false,
```
BridgeAR, line 14: Instead of using booleans for each runtime, we could just use a `runtime` field that is set to the runtime name. That way it is easier to add support for additional ones with less changes required.

packages/dd-trace/src/utils/runtime.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+  }
+
+  // Node.js detection
+  if (typeof process !== 'undefined' && process.versions?.node) {
```
BridgeAR, line 19: This should be the first check as it is the most likely one

packages/dd-trace/src/utils/runtime.js
```diff
@@ -0,0 +1,45 @@
...
+  // Fallback
+  return {
+    name: 'unknown',
+    version: 'unknown',
+    isBun: false,
+    isNode: false,
+  }
+}
+
+const runtime = detectRuntime()
```
BridgeAR, line 37: Instead of using a method, it could just be code that is executed directly.

packages/dd-trace/src/utils/runtime.js
```diff
@@ -0,0 +1,45 @@
...
+
+module.exports = {
+  runtime,
+  isBun: runtime.isBun,
+  isNode: runtime.isNode,
+  runtimeName: runtime.name,
+  runtimeVersion: runtime.version,
```
BridgeAR, line 44: Please only export one object instead of having the object and a copy of that as well

packages/dd-trace/test/utils/runtime.spec.js
```diff
@@ -0,0 +1,227 @@
+'use strict'
+
+const assert = require('node:assert/strict')
+const { describe, it, beforeEach, afterEach } = require('tap').mocha
```
BridgeAR, line 4:
```suggestion
const { describe, it, beforeEach, afterEach } = require('mocha')
```

Tap was recently removed

packages/dd-trace/test/utils/runtime.spec.js
```diff
@@ -0,0 +1,227 @@
...
+        } catch (e2) {
+          // ignore if we can't restore
+        }
+      }
+    }
+    // clear module cache after test
+    delete require.cache[runtimeModulePath]
+  })
+
+  function setVersions (versions) {
```
BridgeAR, line 52: Instead of mocking these, having just integration tests should in my opinion be fine.

That way we are safe that we always test the right things and can fully rely on it, while keeping the test setup simple.

version.js
```diff
@@ -14,5 +20,10 @@ module.exports = {
   NODE_MAJOR: parseInt(nodeMatches[1]),
   NODE_MINOR: parseInt(nodeMatches[2]),
   NODE_PATCH: parseInt(nodeMatches[3]),
-  NODE_VERSION: nodeMatches[0]
+  NODE_VERSION: nodeMatches[0],
+  RUNTIME: runtime.runtimeName,
+  RUNTIME_MAJOR: parseInt(runtimeVersion[1]), // in case it breaks downstream consumer
+  RUNTIME_MINOR: parseInt(runtimeVersion[2]),
+  RUNTIME_PATCH: parseInt(runtimeVersion[3]),
+  RUNTIME_VERSION: runtime.runtimeVersion
```
BridgeAR, line 28: I don't think we should add any of these as long as there is no consumer of these entries

packages/dd-trace/src/config.js
```diff
@@ -287,6 +288,39 @@ class Config {
...
+      name: runtimeName,
+      version: runtimeVersion,
+      isBun,
+      isNode: !isBun,
+    }
+
+    // Disable incompatible features on Bun
+    if (isBun) {
+      // Profiling won't work (V8-specific)
+      if (this.profiling?.enabled !== false) {
```
BridgeAR, line 301: This is not the only way profiling can be enabled. The programmatic way is in fact less common than using for example environment variables (while there is also stable configuration and more).

Checking for this should as such not be part of config.

Instead, it could be part of proxy (the main tracer class).

BridgeAR (changes requested):
I do not see us supporting Bun officially anytime soon, while I could of course be mistaken.

I did review the code nevertheless to give some hints of the current suggestion

Issue #7110, opened 2025-12-15
[BUG]: MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 connect listeners added to [TLSSocket].


johnkrzywanek:
### Tracer Version(s)

5.80.0, also happened in earlier versions

### Node.js Version(s)

22

### Bug Report

Frequent warnings occurring on a Node.js Express-based application, running on Kubernetes (GKE).

```
(node:1) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 connect listeners added to [TLSSocket]. MaxListeners is 10. Use emitter.setMaxListeners() to increase limit
at genericNodeError (node:internal/errors:983:15)
at wrappedFn (node:internal/errors:537:14)
at _addListener (node:events:593:17)
at TLSSocket.addListener (node:events:611:10)
at Readable.on (node:internal/streams/readable:1128:35)
at TLSSocket.once (node:events:655:8)
at once (/app/node_modules/got/dist/source/core/utils/unhandle.js:11:20)
at TLSSocket.<anonymous> (/app/node_modules/got/dist/source/core/utils/timed-out.js:89:29)
at Object.onceWrapper (node:events:634:26)
at TLSSocket.emit (node:events:531:35)
at TLSSocket.emit (/app/node_modules/dd-trace/packages/datadog-instrumentations/src/net.js:65:25)
at GetAddrInfoReqWrap.emitLookup (node:net:1471:14)
at GetAddrInfoReqWrap.wrappedLookupCallback (/app/node_modules/@opentelemetry/instrumentation-dns/build/src/instrumentation.js:131:29)
at run (node:diagnostics_channel:168:14)
at DatadogStorage.run (/app/node_modules/dd-trace/packages/datadog-core/src/storage.js:81:22)
at node:diagnostics_channel:100:18
```

### Tracer Config

```
        const {TracerProvider} = tracer.init({
            enabled: process.env.DD_TRACING_ENABLED ?? false,
            service: 'service-name',
            version: process.env.DD_VERSION ?? version, 
            environment: process.env.TRACING_ENVIRONMENT,
            logInjection: false,
            appsec: {
                enabled: process.env.DD_APPSEC_ENABLED ?? false,
            },
            analytics: process.env.DD_ANALYTICS_ENABLED ?? false,
            profiling: process.env.DD_PROFILING_ENABLED ?? false,
            logLevel: process.env.LOG_LEVEL ?? 'debug',
[...]

PR #7201, opened 2026-01-08
fix: remove event emitter leak warning

BridgeAR:
This is done by limiting the connect listeners added to one instead of up to three.

It also changes many listeners to become `once` listeners instead of `on` to limit the addition to a single execution, if sufficient.

The code is also changed that only a single `beforeExit` listener is added. That is done by using a global
symbol that may now contain different dd-trace related state.

In addition, this adds a test helper that verifies we are never above 6 listeners.
Exceptions are special handled (pino-pretty adds additional listeners).

Fixes: #7110

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -179,10 +179,10 @@ const log = require('../log')
 
 ### Event handlers
 
+- Avoid adding new listeners, if possible
 - Use monitor symbols like `events.errorMonitor` when available
 - Use `.once()` methods instead of `.on()`, if the event is only needed once
-- Limit event additions to the absolute minimum required
-- If new 'beforeExit' events on `process` are needed, add them to globalThis[Symbol.for('dd-trace')].beforeExitHandlers
+- If new `beforeExit` events on `process` are needed, add them to `globalThis[Symbol.for('dd-trace')].beforeExitHandlers`
 
 ### Performance and Memory
 
--- a/packages/dd-trace/test/llmobs/writers/base.spec.js
+++ b/packages/dd-trace/test/llmobs/writers/base.spec.js
@@ -8,6 +8,7 @@ const sinon = require('sinon')
 
 require('../../setup/core')
 const { useEnv } = require('../../../../../integration-tests/helpers')
+const { removeDestroyHandler } = require('../util')
 
 describe('BaseLLMObsWriter', () => {
   let BaseLLMObsWriter
@@ -50,12 +51,7 @@ describe('BaseLLMObsWriter', () => {
 
   afterEach(() => {
     clock.restore()
-    for (const handler of globalThis[Symbol.for('dd-trace')].beforeExitHandlers) {
-      if (handler.name.endsWith('destroy')) {
-        globalThis[Symbol.for('dd-trace')].beforeExitHandlers.delete(handler)
-        break
-      }
-    }
+    removeDestroyHandler()
   })
 
   it('constructs an agentless writer', () => {
--- a/packages/dd-trace/test/llmobs/index.spec.js
+++ b/packages/dd-trace/test/llmobs/index.spec.js
@@ -8,6 +8,8 @@ const proxyquire = require('proxyquire')
 const sinon = require('sinon')
 
 const AgentInfoExporter = require('../../src/exporters/common/agent-info-exporter')
+const { removeDestroyHandler } = require('./util')
+
 const spanFinishCh = channel('dd-trace:span:finish')
 const evalMetricAppendCh = channel('llmobs:eval-metric:append')
 const flushCh = channel('llmobs:writers:flush')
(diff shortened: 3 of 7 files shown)


packages/dd-trace/index.js
```diff
@@ -10,6 +10,23 @@ if (!global._ddtrace) {
...
+    value: {
+      beforeExitHandlers: new Set(),
+    },
+    enumerable: false,
+    configurable: true, // Allow this to be overridden by loading the tracer
+    writable: false
+  })
+
+  process.once('beforeExit', function mainBeforeExit () {
+    for (const handler of globalThis[ddTraceSymbol].beforeExitHandlers) {
```
simon-id: little check to make sure this set still exists so it doesn't crash with an ugly error if it doesn't exists anymroe ?

BridgeAR: I thought about that, while our tracer would be fully broken, if that does not exist.

I am not sure if that is a case that we have to protect against. Opinions?

simon-id: Cost of adding the check: almost none.
Benefits: avoid a misleading stacktrace out of nowhere before exit, in a rare case.

You never know what weird stuff the customers will do, for sure it would mean the library won't work in that state, but who knows maybe they remove it in a beforeExit callback just before ours :shrug:, I don't see a reason not to have the check.

packages/dd-trace/index.js
```diff
@@ -10,6 +10,25 @@ if (!global._ddtrace) {
...
+    value: {
+      beforeExitHandlers: new Set(),
+    },
+    enumerable: false,
+    configurable: true, // Allow this to be overridden by loading the tracer
+    writable: false
+  })
+
+  process.once('beforeExit', function mainBeforeExit () {
+    if (globalThis[ddTraceSymbol] && globalThis[ddTraceSymbol].beforeExitHandlers) {
```
simon-id:
```suggestion
    if (globalThis[ddTraceSymbol]?.beforeExitHandlers) {
```

AGENTS.md
```diff
@@ -134,23 +146,34 @@ const log = require('../log')
...
   const { NODE_MAJOR } = require('./version')
   if (NODE_MAJOR >= 20) { /* Use Node.js 20+ API */ }
   ```
 
+### Event handlers
+
+- Use monitor symbols like `events.errorMonitor` when available
+- Use `.once()` methods instead of `.on()`, if the event is only needed once
+- Limit event additions to the absolute minimum required
+- If new 'beforeExit' events on `process` are needed, add them to globalThis[Symbol.for('dd-trace')].beforeExitHandlers
```
watson:
```suggestion
- If new `beforeExit` events on `process` are needed, add them to `globalThis[Symbol.for('dd-trace')].beforeExitHandlers`
```

AGENTS.md
```diff
@@ -134,23 +146,34 @@ const log = require('../log')
...
   ```js
   const { NODE_MAJOR } = require('./version')
   if (NODE_MAJOR >= 20) { /* Use Node.js 20+ API */ }
   ```
 
+### Event handlers
+
+- Use monitor symbols like `events.errorMonitor` when available
+- Use `.once()` methods instead of `.on()`, if the event is only needed once
+- Limit event additions to the absolute minimum required
```
watson: Not sure I understand this line?

packages/dd-trace/test/plugins/agent.js
```diff
@@ -696,6 +699,9 @@ module.exports = {
     delete require.cache[require.resolve('../..')]
     delete global._ddtrace
 
+    process.removeAllListeners('exit')
+    process.removeAllListeners('beforeExit')
```
watson, line 726: Do we really want to remove all listeners here? I.e. can we be sure that there's only the listeners we expect there to be?

packages/dd-trace/test/llmobs/writers/base.spec.js
```diff
@@ -49,7 +50,12 @@ describe('BaseLLMObsWriter', () => {
 
   afterEach(() => {
     clock.restore()
-    process.removeAllListeners('beforeExit')
+    for (const handler of globalThis[Symbol.for('dd-trace')].beforeExitHandlers) {
+      if (handler.name.endsWith('destroy')) {
+        globalThis[Symbol.for('dd-trace')].beforeExitHandlers.delete(handler)
+        break
+      }
+    }
```
watson: This snippet is used in quite a few places. Would it make sense to move it to a helper, or is that overkill?

packages/dd-trace/test/plugins/agent.js
BridgeAR: It is going to be tricky to determine if an exit listener is meant to stay. When we start our agent, it should just be as if it is a fresh service start. That is why removing all seems best to me. Any former state should not exist anymore.

AGENTS.md
BridgeAR: In `net` we added up to three distinct `connect` listener handlers instead of one.

watson: Ah, I see. It should probably be rephrased then. Maybe something like:

```suggestion
- Limit unique event listeners for a given event to the absolute minimum required
```

BridgeAR: I changed to just not adding listeners, when possible

PR #7205, opened 2026-01-09
Add pubsub.delivery_duration_ms to push subscriptions

nina9753:
### What does this PR do?
Small clean up PR to :
- remove redundant `pubsub.batch.request_span_id` span tag which is the same as the `pubsub.batch.request_trace_id` tag. 
- added `pubsub.delivery_duration_ms` to push subscriptions for tag parity with pull subscriptions
- Update synthetic span for push subscription name from `pubsub.delivery` to `pubsub.push.receive`
- Updates tests to test the new changes with garbage collector

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/integration-tests/helpers/index.js
+++ b/integration-tests/helpers/index.js
@@ -172,11 +172,20 @@ function assertTelemetryPoints (pid, msgs, expectedTelemetryPoints) {
 }
 
 /**
- * @typedef {childProcess.ChildProcess & { url: string }} SpawnedProcess
+ * @typedef {childProcess.ChildProcess & {
+ *   url: string,
+ *   stdout: NodeJS.ReadableStream,
+ *   stderr: NodeJS.ReadableStream
+ * }} SpawnedProcess
  */
 
 /**
  * Spawns a Node.js script in a child process and returns a promise that resolves when the process is ready.
+ *
+ * This function expects the spawned process to stay alive (e.g., a server). If the process exits
+ * (even with code 0), the promise will reject with an error.
+ *
+ * For processes that are expected to run and exit cleanly, use `spawnProcAndExpectExit` instead.
  *
  * @param {string|URL} filename - The filename of the Node.js script to spawn in a child process.
  * @param {childProcess.ForkOptions} [options] - The options to pass to the child process.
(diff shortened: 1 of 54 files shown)


packages/datadog-plugin-google-cloud-pubsub/src/consumer.js
```diff
@@ -134,7 +134,6 @@ class GoogleCloudPubsubConsumerPlugin extends ConsumerPlugin {
     }
 
     if (batchRequestTraceId && batchRequestSpanId) {
-      meta['pubsub.batch.request_trace_id'] = batchRequestTraceId
       meta['pubsub.batch.request_span_id'] = batchRequestSpanId
       meta['_dd.pubsub_request.trace_id'] = batchRequestTraceId
       meta['_dd.pubsub_request.span_id'] = batchRequestSpanId
```
BridgeAR: If this is redundant, would `pubsub.batch.request_span_id` not also be redundant?

packages/datadog-plugin-google-cloud-pubsub/src/pubsub-push-subscription.js
```diff
@@ -173,6 +173,12 @@ class GoogleCloudPubsubPushSubscriptionPlugin extends TracingPlugin {
       return null
     }
 
+    // Calculate delivery latency (queue time from publish to delivery)
+    if (publishStartTime) {
+      const deliveryDuration = Date.now() - Number(publishStartTime)
+      span.setTag('pubsub.delivery_duration_ms', deliveryDuration)
```
BridgeAR, line 180: Please update the tests to verify this change.

BridgeAR (changes requested):
Our tests seem to not reflect this change properly.

When I ran the tests locally, the span assertion always returned early and never found a match. It should actually not contain the return. Our test helper would pass, if any trace contains the correct span, which does not seem to be the case.

Please have a look at that and update the test so that it would now need an update due to the changed code.

I just removed the early returns in the tests and now the following all fail:

```
      1) should create BOTH pubsub.delivery span AND HTTP span in same trace
      2) should propagate distributed trace context from producer to push delivery
      3) should add batch metadata to delivery span
      4) should set service name with -pubsub suffix
```

packages/datadog-plugin-google-cloud-pubsub/test/pubsub-push-subscription.spec.js
```diff
@@ -361,78 +234,51 @@ describe('Push Subscription Plugin', () => {
...
-            // Wait for request to complete
+            await sendPushRequest(port)
             await wait(100)
 
-            // Force garbage collection
-            // @ts-expect-error We expect the test to be started with --trace-gc
+            // Force garbage collection multiple times
             global.gc()
             await wait(100)
-            // @ts-expect-error We expect the test to be started with --trace-gc
```
BridgeAR, line 425: Please keep these or alternatively just assign e.g., `const gc = global.gc ?? (() => {})` and use that to prevent type issues.

BridgeAR (commented):
Nice! The tests are now seemingly running correct, thank you for fixing them! I believe this should be good to land as soon as the type changes are undone

packages/datadog-plugin-google-cloud-pubsub/test/index.spec.js
```diff
@@ -15,6 +15,9 @@ const { computePathwayHash } = require('../../dd-trace/src/datastreams/pathway')
 const { DataStreamsProcessor, ENTRY_PARENT_HASH } = require('../../dd-trace/src/datastreams/processor')
 const { expectedSchema, rawExpectedSchema } = require('./naming')
 
+// @ts-expect-error We expect the test to be started with --expose-gc
+const gc = global.gc ?? (() => {})
```
BridgeAR, line 17: I believe the error is now not happening here due to the fallback?

The same in the other file

BridgeAR (approved):

PR #7207, opened 2026-01-09
chore: improve parallel mocha output

BridgeAR:
This adds a global counter next to the local error numbers as well as adding the file name next to the failure counter of each file.

### DIFF
--- a/scripts/mocha-parallel-files.js
+++ b/scripts/mocha-parallel-files.js
@@ -217,9 +217,9 @@ async function main () {
     code: /** @type {number|null} */ (null),
     signal: /** @type {NodeJS.Signals|null} */ (null),
 
-    // Output buffers (in-memory)
-    stdoutBuf: /** @type {string[]} */ ([]),
-    stderrWarnBuf: /** @type {string[]} */ ([]),
+    // Output buffers (in-memory). For non-active entries, preserve stdout/stderr warning ordering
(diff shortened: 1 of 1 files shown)


scripts/mocha-parallel-files.js
```diff
@@ -488,6 +514,9 @@ async function main () {
     for (const { file, code, signal } of failed) {
       process.stdout.write(`- ${file} (exit=${code ?? 'null'} signal=${signal ?? 'null'})\n`)
     }
+    process.stdout.write(
+      'Legend: [n] = global failure index across all files; n) = local failure index within a file.'
```
watson: Shouldn't this end with a line break?

scripts/mocha-parallel-files.js
```diff
@@ -478,16 +511,20 @@ async function main () {
 
   process.stdout.write('\n=== Summary ===\n')
   process.stdout.write(`Passed: ${totalPasses}\n`)
-  process.stdout.write(`Failed: ${totalFailures}\n`)
+  process.stdout.write(`Failed: ${totalFailures + crashedFiles}\n`)
```
watson, line 514: Do we need to add `crashedFiles` here if we also output it below?

BridgeAR: I think it makes sense to count those as test failures, no?

PR #7209, opened 2026-01-09
chore: update AGENTS.md for improved types and assertions

BridgeAR:


AGENTS.md
```diff
@@ -36,34 +56,53 @@ Each package under `packages/` follows a consistent structure:
 
 ### Running Individual Tests
 
-**IMPORTANT**: Never run `yarn test` directly. Use `mocha` directly on test files.
+**IMPORTANT**: Avoid running the root `npm test` unless you explicitly want the full repo test run. Prefer targeted `npm run test:<area>` scripts or running mocha on specific test files.
```
pabloerhard: What is the reason behind the agents running tests with npm and not yarn?

BridgeAR: It is just a step to slowly remove yarn. I actually updated more for that now

AGENTS.md
pabloerhard, line 1: We should also consider adding the instruction of using varySandbox when doing ESM tests.

AGENTS.md
```diff
@@ -36,38 +56,56 @@ Each package under `packages/` follows a consistent structure:
...
```
watson: A few points:
- It's not easy for an agent to know which `<area>` to use, which is why I asked it to run the mocha files directly before.
- I'm not sure there's ever a good case for running the full test suite?

AGENTS.md
```diff
@@ -98,23 +137,36 @@ assert.equal(actual, expected)
...
 - Files: kebab-case
 
 ### JSDoc
+
 - Use TypeScript-compatible syntax (`@param {string}`, `@returns {Promise<void>}`, `@typedef`)
 - Never use `any` (be specific or use `unknown` if type is truly unknown)
+- Write the most specific types possible by reading the overall context
+- Always define types for method arguments as method params
+- Never define argument types inside of a method
+- Only define types inside of a method, if it can not be inferred otherwise
```
watson, line 144: Consider merging these two bullets as the second one defies the first, e.g:

```suggestion
- Never define argument types inside of a method unless it can not be inferred otherwise
```

AGENTS.md
BridgeAR: I think you actually know these best. Could you just add them in the ESM tests PR that you are currently working on?

AGENTS.md
```diff
@@ -134,8 +186,11 @@ const log = require('../log')
 ### ECMAScript and Node.js API Standards
 
 **Target Node.js 18.0.0 compatibility:**
+
 - Use modern JS features supported by Node.js (e.g., optional chaining `?.`, nullish coalescing `??`)
+- Use `undefined` over `null`, if not required otherwise
```
watson, line 172: Not sure I understand this bullet?

AGENTS.md
```diff
@@ -145,12 +200,17 @@ const log = require('../log')
 
 **CRITICAL: Tracer runs in application hot paths - every operation counts.**
 
+- Use fast paths and most performant APIs
```
watson: Should this bullet be split into two? The two things seems unrelated

AGENTS.md
```diff
@@ -145,12 +200,17 @@ const log = require('../log')
...
+- Understanding the context to write CPU performant code and skip unnecessary steps
```
watson: This seems a bit vague for an agent. Can it be more specific?

AGENTS.md
```diff
@@ -2,21 +2,40 @@
...
 
-- Node.js >= 18
+- Node.js >= 18 (recommended: use `.nvmrc` via `nvm use`; supported range is in `package.json#engines`)
 - yarn 1.x
 - Docker + docker-compose (for running service dependencies in tests)
 
 ## Setup
 
+- If you use nvm: `nvm use`
 - Install dependencies: `yarn install`
```
watson, line 11: I'm a bit unsure about these two sections in general. They don't add anything for members of the Node.js guild who already have the environment set up. But I added it anyway because they might help 3rd party contributions. But I'd still like to keep it as brief as possible so that we in the Node.js guild don't have to waste a lot of context reading this section every time.

With that in mind, I'd probably drop the change to the `- Node.js >= 18` line as the added "Setup" line should give enough context for the agent to be able to work with the repo.

AGENTS.md
```diff
@@ -2,21 +2,40 @@
...
+# Install deps
+yarn install
+
+# Lint
+npm run lint
+npm run lint:fix
+
+# Typecheck
+npm run type:check
+```
```
watson: Not sure about the value of repeating these here? Isn't it just bloat?

AGENTS.md
```diff
@@ -36,38 +56,56 @@ Each package under `packages/` follows a consistent structure:
 
 ### Running Individual Tests
 
-**IMPORTANT**: Never run `yarn test` directly. Use `mocha` directly on test files.
+**IMPORTANT**: Avoid running the root `npm test` unless you explicitly want the full repo test run. Prefer targeted `npm run test:<area>` scripts or running mocha on specific test files.
 
 **Mocha unit tests:**
```
watson: While you're here, this was left over from a previous refactor when we removed tap:

```suggestion
**Unit tests:**
```

AGENTS.md
```diff
@@ -36,38 +56,56 @@ Each package under `packages/` follows a consistent structure:
...
+
 ```bash
 ./node_modules/.bin/mocha --timeout 60000 -r "packages/dd-trace/test/setup/core.js" path/to/test.spec.js
 ```
 
+**If a test expects “spec file is entrypoint” semantics (tap-like):**
+
+```bash
+node scripts/mocha-run-file.js path/to/test.spec.js
+```
```
watson, line 59: Not sure I understand this, even outside the scope of being an agent. Please explain

AGENTS.md
```diff
@@ -98,23 +137,36 @@ assert.equal(actual, expected)
 assertObjectContains(response, { status: 200, body: { user: { name: 'Alice' } } })
 ```
 
+When writing assertions, use the minimal amount of individual assertion calls possible.
+So instead of writing multiple assert.strictEqual calls, use a single assert.deepStrictEqual call or assertObjectContains.
+
+Never use a `doesNotThrow()` assertion. Instead, just write execute the method directly.
```
watson:
```suggestion
Never use a `doesNotThrow()` assertion. Instead, just execute the method directly.
```

AGENTS.md
```diff
@@ -98,23 +137,36 @@ assert.equal(actual, expected)
 assertObjectContains(response, { status: 200, body: { user: { name: 'Alice' } } })
 ```
 
+When writing assertions, use the minimal amount of individual assertion calls possible.
+So instead of writing multiple assert.strictEqual calls, use a single assert.deepStrictEqual call or assertObjectContains.
```
watson: Just trying to make it less verbose for the agent:

```suggestion
Favor fewer `assert.deepStrictEqual`/`assertObjectContains` calls over many `assert.strictEqual` calls.
```

AGENTS.md
BridgeAR: Most of the time I get code as: `return null` instead of `return`, even if both are possible. Or a variable is assigned null instead of just `let foo`

AGENTS.md
BridgeAR: After I edited a few things, I asked the LLM what could be improved and it suggested the common commands to be added. I just accepted that

AGENTS.md
BridgeAR: When our tests used tap, the entry file would be the .spec.js file itself. When using mocha, the entry is actually mocha that reads the .spec.js file, but it is not the main file according to e.g., require.main.filename or process.argv[1]. That script uses a hack to make that work.

AGENTS.md
BridgeAR: That changes the meaning though. I really do not want any type definitions for arguments and type definitions inside of a method should only happen, if they can not be inferred from the call sites.

AGENTS.md
BridgeAR: This helped my llm to understand when to use nvm use ..., so I would prefer to keep it

AGENTS.md
watson: Ah sorry, I thought the 2nd one was about method args also. But now I understand that it is about the method body

AGENTS.md
watson: Yeah, my agent did similar things when I asked it - but it all depends on how you frame the question. I ended up removing it because I think it's just bloat. But I don't have that much experience with `AGENTS.md`, so maybe I'm wrong. I just am very conscious about not using too many tokens in the context window. This is also why I try to keep the sentences short and to the point. As long as we haven't seen an issue with the agent not understanding to use these commands, I'd be careful adding them here.

AGENTS.md
watson: The presence of nvm is environment specific, so it might not be present everywhere. In what scenario did you need the agent to change Node.js version?

AGENTS.md
watson: So are you saying we shouldn't run `./node_modules/.bin/mocha path/to/test.spec.js` but instead `node scripts/mocha-run-file.js path/to/test.spec.js`?

BridgeAR: At least in some cases, yes. <s>Running `node path/to/test.spec.js` should actually also work in that case</s>

AGENTS.md
BridgeAR: I can remove it. I normally do not have to switch Node.js versions with the agent. It is just a config that I have to change locally

AGENTS.md
watson: Sorry, which cases is that? I'm not sure I follow

BridgeAR: E.g., `packages/dd-trace/test/pkg.spec.js` and `packages/dd-trace/test/span_processor.spec.js`

watson: Ok, after looking closer at `packages/dd-trace/test/pkg.spec.js` I now understand the problem. I'm not sure it was very clear from the `AGENTS.md` description. The key is if the test requires `require.main.filename` to be something specific as far as I understand.

However, determining this requires knowledge of the files being tested, so to avoid this extra lookup, I recommend that our request to the agent is simply to always use `scripts/mocha-run-file.js` and never run spec files directly via `./node_modules/.bin/mocha`. That way the agent doesn't need to spend time resolving this.

watson: Discussed on Slack, and TL;DR: `scripts/mocha-run-file.js` doesn't support running all spec files yet which would require additional work. So let's leave this as is for now and get this merged, though I still think we should reword the sentence `**If a test expects “spec file is entrypoint” semantics (tap-like):**` as, at least to my ears, it's very confusing.

Issue #7193, opened 2026-01-08
OTel Span links - TypeError: context.toTraceId is not a function


naseemkullah:
When using the Datadog SDK [with the OpenTelemetry API Bridge](https://docs.datadoghq.com/opentelemetry/instrument/api_support/nodejs/traces), providing a standard OTel SpanContext in the links option of a span causes a crash. The internal Datadog formatter assumes the context object has proprietary methods like `.toTraceId()`.

### Reproduce

Initialize the Datadog tracer with the OTel API bridge enabled.

Extract a SpanContext from a carrier using standard OTel propagation (e.g., from Pub/Sub attributes).

Start a new span using the OTel API and include the extracted context in the links array.

```js
const otel = require('@opentelemetry/api');

// 1. Extract context from incoming message attributes
const extractedContext = otel.propagation.extract(otel.context.active(), carrier);
const spanContext = otel.trace.getSpanContext(extractedContext);

// 2. Start a span with a link to that context
const tracer = otel.trace.getTracer('my-service');
tracer.startActiveSpan('handle-message', {
  links: [{ context: spanContext }] 
}, (span) => {
  // Logic here
  span.end();
});
```
### Expected behaviour

The tracer should handle standard OpenTelemetry SpanContext objects. The bridge should be responsible for converting these into Datadog-compatable internal objects or accessing the traceId property directly.

### Actual behaviour 

The application crashes with the following error:

```
TypeError: context.toTraceId is not a function
    at /usr/src/app/node_modules/dd-trace/packages/dd-trace/src/span_format.js:81:25
    at Array.map (<anonymous>)
    at formatLinks (/usr/src/app/node_modules/dd-trace/packages/dd-trace/src/span_format.js:79:16)
```

### Additional context 

The crash occurs at [this line](https://github.com/DataDog/dd-trace-js/blob/06156ff7fb9f0e1e3cda35d2370d45a7564250ec/packages/dd-trace/src/span_format.js#L81). A standard OTel SpanContext contains traceId and spanId as strings, but does not implement the .toTraceId() method. This makes it impossible to use Span Links in Pub/Sub consumers when the parent context is extracted via the standard propagation API.

PR #7210, opened 2026-01-09
fix: handle OTel span links properly

BridgeAR:
The formatter did not expect the span links to be of the otel style.
This is now fixed and the formatter can handle these just fine.

Fixes: https://github.com/DataDog/dd-trace-js/issues/7193

### DIFF
--- a/packages/dd-trace/src/opentelemetry/tracer.js
+++ b/packages/dd-trace/src/opentelemetry/tracer.js
@@ -5,11 +5,47 @@ const { sanitizeAttributes } = require('../../../../vendor/dist/@opentelemetry/c
 
 const id = require('../id')
 const log = require('../log')
+const DatadogSpanContext = require('../opentracing/span_context')
 const TextMapPropagator = require('../opentracing/propagation/text_map')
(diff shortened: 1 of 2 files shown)


khanayan123 (approved):

packages/dd-trace/src/opentelemetry/tracer.js
```diff
@@ -5,12 +5,48 @@ const { sanitizeAttributes } = require('../../../../vendor/dist/@opentelemetry/c
...
+  if (context._ddContext) return context._ddContext
+
+  // Datadog span context
+  if (typeof context.toTraceId === 'function' && typeof context.toSpanId === 'function') {
+    return context
+  }
+
+  // Standard OTel SpanContext (traceId/spanId)
+  if (typeof context.traceId !== 'string' || typeof context.spanId !== 'string') {
+    // Invalid
```
watson, line 28: Should we add a test for this?

BridgeAR: I will add one when I work on it again (I hope I remember)

PR #7212, opened 2026-01-09
feat(dsm,dbm): add process tags support for enhanced trace correlation

tlhunter:
## Summary

This PR adds propagation hash support to Data Streams Monitoring (DSM) and Database Monitoring (DBM) to enable correlation between traces, database operations, and data stream pathways using process and container metadata.

The propagation hash is an FNV-1a 64-bit hash combining:
- **Process tags**: entrypoint info (workdir, name, basedir, type), package.json name
- **Container tags hash**: received from the Datadog agent via response headers

This hash is used to:
- **DSM**: Include in pathway hash computation and add `ProcessTags` field to payloads
- **DBM**: Add `ddsh='<hex_hash>'` parameter to SQL comments for database query correlation

## Related PRs

- dd-trace-py: https://github.com/DataDog/dd-trace-py/pull/15356
- dd-trace-java: https://github.com/DataDog/dd-trace-java/pull/9282
- dd-trace-rb: https://github.com/DataDog/dd-trace-rb/pull/5208 https://github.com/DataDog/dd-trace-rb/pull/5296
- dd-trace-rb: https://github.com/DataDog/dd-trace-rb/pull/5209


packages/dd-trace/src/datastreams/writer.js
```diff
@@ -24,7 +25,15 @@ function makeRequest (data, url, cb) {
 
   log.debug('Request to the intake: %j', options)
 
-  request(data, options, (err, res) => {
+  request(data, options, (err, res, status, headers) => {
+    // Capture container tags hash from agent response headers
+    if (headers) {
+      const containerTagsHash = headers['datadog-container-tags']
```
wantsui: The agent's container tags hash header is `Datadog-Container-Tags-Hash`

See https://github.com/DataDog/datadog-agent/blob/4fbbc4a16aebbaea34b1248e36c285a46361c590/pkg/trace/api/info.go#L23

packages/dd-trace/src/exporters/agent/writer.js
```diff
@@ -53,6 +49,15 @@ class AgentWriter extends BaseWriter {
 
       log.debug('Response from the agent: %s', res)
 
+      // Capture container tags hash from agent response headers
+      if (headers) {
+        const containerTagsHash = headers['datadog-container-tags']
```
wantsui: See the comment from above about `Datadog-Container-Tags-Hash`: https://github.com/DataDog/datadog-agent/blob/4fbbc4a16aebbaea34b1248e36c285a46361c590/pkg/trace/api/info.go#L23

packages/dd-trace/src/datastreams/pathway.js
```diff
@@ -26,17 +26,28 @@ function shaHash (checkpointString) {
...
   if (value) {
     return value
   }
-  const currentHash = shaHash(`${service}${env}` + hashableEdgeTags.join(''))
+
+  // Include propagation hash in hash input if present
+  const baseString = `${service}${env}` + hashableEdgeTags.join('')
+  const hashInput = propagationHashBigInt
+    ? `${baseString}:${propagationHashBigInt.toString(16)}`
+    : baseString
```
vandonr, line 56: I feel like this is doing roughly the same job as lines 36-37 above, but in a different way, and I cannot really tell if it's same same or different...

It'd probably deserve a comment explaining how `hashinput` is different from `key`, if so, and why.

packages/dd-trace/src/datastreams/writer.js
```diff
@@ -24,7 +25,15 @@ function makeRequest (data, url, cb) {
...
   log.debug('Request to the intake: %j', options)
 
-  request(data, options, (err, res) => {
+  request(data, options, (err, res, status, headers) => {
+    // Capture container tags hash from agent response headers
+    if (headers) {
+      const containerTagsHash = headers['Datadog-Container-Tags-Hash']
+      if (containerTagsHash) {
+        propagationHash.updateContainerTagsHash(containerTagsHash)
+        log.debug('Updated container tags hash from DSM response: %s', containerTagsHash)
```
vandonr: maybe only log this if the hash we got is different from the one we had stored to avoid being too verbose ? Or not at all if this adds too much logic.
Otherwise, we're going to get this log for every single request

packages/dd-trace/src/exporters/agent/writer.js
```diff
@@ -53,6 +49,15 @@ class AgentWriter extends BaseWriter {
 
       log.debug('Response from the agent: %s', res)
 
+      // Capture container tags hash from agent response headers
+      if (headers) {
+        const containerTagsHash = headers['Datadog-Container-Tags-Hash']
+        if (containerTagsHash) {
+          propagationHash.updateContainerTagsHash(containerTagsHash)
+          log.debug('Updated container tags hash from agent response: %s', containerTagsHash)
```
vandonr: same remark as above. Also, not sure we need to request the container tags hash in DSM writer _and_ here. We mostly need it once, so I think it's ok to do it here only (which is the most common path I assume), to avoid duplicating code and adding complexity.

packages/dd-trace/src/exporters/agent/writer.js
vandonr, line 86: I'm not sure where this is coming from, but if we want to read the container tags hash from the response, we need to provide the container ID as part of the request headers (under the key `Datadog-Container-ID`).
Maybe it was done already, but I don't see it in your changes here.

packages/dd-trace/src/plugins/database.js
```diff
@@ -73,13 +74,25 @@ class DatabasePlugin extends StoragePlugin {
 
     const servicePropagation = this.createDBMPropagationCommentService(dbmService, span)
 
+    let dbmComment = servicePropagation
+
+    // Add propagation hash if enabled
+    if (propagationHash.isEnabled()) {
```
vandonr: I know this sounds fucked up, but this injection is actually controlled by a _separate_ setting, `DB_DBM_INJECT_SQL_BASEHASH`
see [java](https://github.com/DataDog/dd-trace-java/blob/6bbcd6103723b58d041ae811a05f3e88ed5db240/dd-trace-api/src/main/java/datadog/trace/api/config/TraceInstrumentationConfig.java#L71), [python](https://github.com/DataDog/dd-trace-py/blob/984a634671a2fa93164c71a5a77bfa34c22c6cbc/ddtrace/internal/settings/_database_monitoring.py#L17-L22)

packages/dd-trace/src/datastreams/processor.js
```diff
@@ -171,6 +174,12 @@ class DataStreamsProcessor {
       Lang: 'javascript',
       Tags: Object.entries(this.tags).map(([key, value]) => `${key}:${value}`)
     }
+
+    // Add ProcessTags only if feature is enabled and process tags exist
+    if (propagationHash.isEnabled() && processTags.serialized) {
+      payload.ProcessTags = processTags.serialized
```
wantsui: I got this error `400 Bad Request. Response from the endpoint: "{"acknowledged":false,"error":"Error decoding request msgp: attempted to decode type \"str\" with method for \"array\" at ProcessTags"}`, so I believe this means we need to actually send it as an array of strings instead.
```suggestion
      payload.ProcessTags = processTags.serialized.split(',')
```

packages/dd-trace/src/exporters/agent/writer.js
tlhunter: `Datadog-Container-ID` is automatically injected by `docker.inject()` in `exporters/common/request.js`.

packages/dd-trace/src/datastreams/processor.js
```diff
@@ -171,6 +174,12 @@ class DataStreamsProcessor {
       Lang: 'javascript',
       Tags: Object.entries(this.tags).map(([key, value]) => `${key}:${value}`),
     }
+
+    // Add ProcessTags only if feature is enabled and process tags exist
+    if (propagationHash.isEnabled() && processTags.serialized) {
+      payload.ProcessTags = processTags.serialized.split(',')
```
wantsui, line 180: nit: do we need to be worried about paths having commas if we're just splitting by the comma?
(Not saying it's a big concern so this is just a nit)

tlhunter: I believe the process tag sanitize function would have already replaced comma characters in filenames with underscores.

packages/dd-trace/src/datastreams/pathway.js
tlhunter: @vandonr do these comments I added since your review make sense?

vandonr: yes, sorry I didn't circle back after I left those comments :)
I'd say it's pretty verbose now 😂 but looks good, I mostly missed the fact that one had the parent hash and not the other.

vandonr (approved):

packages/dd-trace/src/plugins/database.js
```diff
@@ -59,13 +60,25 @@ class DatabasePlugin extends StoragePlugin {
...
 
+    let dbmComment = servicePropagation
+
+    // Add propagation hash if both process tags and SQL base hash injection are enabled
+    if (propagationHash.isEnabled() && this.config['dbm.injectSqlBaseHash']) {
+      const hashBase64 = propagationHash.getHashBase64()
+      if (hashBase64) {
+        dbmComment += `,ddsh='${hashBase64}'`
+        // Add hash to span meta as a tag
+        span.setTag('_dd.dbm.propagation_hash', hashBase64)
```
tlhunter, line 71: This should be `_dd.propagated_hash`.

tlhunter: https://github.com/DataDog/dd-trace-js/pull/8643

PR #7213, opened 2026-01-11
chore: activate more eslint imports and jsdoc rules

BridgeAR:
This sorts the imports (a lot of former code was changed before in that direction) and adds stricter type rules.

### DIFF
--- a/scripts/mocha-run-file.js
+++ b/scripts/mocha-run-file.js
@@ -4,8 +4,6 @@ const path = require('path')
 
 const Mocha = require('mocha')
 
-const mocharc = require('../.mocharc.js')
-
 function parseJson (value, fallback) {
   if (!value) return fallback
   try {
@@ -49,7 +47,7 @@ async function main () {
    *   require?: string[]
    * }}
    */
-  const config = parseJson(process.env.MOCHA_RUN_FILE_CONFIG, mocharc)
+  const config = parseJson(process.env.MOCHA_RUN_FILE_CONFIG, {})
 
   const mocha = new Mocha({
     ui: 'bdd',
@@ -57,7 +55,6 @@ async function main () {
     color: config.color ?? true,
     bail: config.bail ?? false,
     retries: config.retries,
-    require: config.require ?? [],
     fullTrace: config.fullTrace ?? false,
     reporter: config.reporter ?? 'spec',
     reporterOptions: config.reporterOptions


packages/datadog-plugin-openai/src/stream-helpers.js
```diff
@@ -1,10 +1,30 @@
...
+ *   text?: string,
+ *   delta?: ChatDelta
+ * }} StreamChoice
+ *
+ * @typedef {JsonObject & { choices: StreamChoice[], usage?: unknown }} StreamChunk
+ * @typedef {JsonObject & { choices: Array<StreamChoice | undefined>, usage?: unknown }} StreamResponseBody
+ *
+ * @typedef {JsonObject & { status?: string }} ResponsesApiResponse
+ * @typedef {JsonObject & { response?: ResponsesApiResponse }} ResponsesApiChunk
+ */
```
sabrenner, line 21: just a question out of curiosity - was this the eslint auto-formatter adding these types as inferred from the method? or did either yourself or an llm write it inferred from the actual method? i think this looks correct though!

packages/dd-trace/src/llmobs/plugins/ai/util.js
```diff
@@ -10,16 +10,35 @@ const MODEL_METADATA_KEYS = new Set([
...
- * @param {Record<string, any>} ctx
- * @returns {Record<string, any>}
+ * @param {AiPluginContext} ctx
+ * @returns {SpanTags}
  */
 function getSpanTags (ctx) {
   const span = ctx.currentStore?.span
   const carrier = ctx.attributes ?? span?.context()._tags ?? {}
-  return carrier
+  return /** @type {SpanTags} */ (carrier)
```
sabrenner, line 41: also just out of curiosity (possibly just a knowledge gap), do we need both the `@returns` in the jsdoc and a "typecast" of `/** @type {SpanTags} */`, or is just the `@returns` ok?

sabrenner: ah, is it maybe because the `span?.context()._tags` might not be typed correctly in its return value, so we want to cast in case?

packages/datadog-plugin-openai/src/stream-helpers.js
BridgeAR: It is inferred by an llm from reading the code

packages/dd-trace/src/llmobs/plugins/ai/util.js
BridgeAR: Yes, while that actually looks like a bug to cast it. I'll merge it now, but I think we have to continue working on our types anyway a lot, so this could be looked at again a tad later.

PR #7215, opened 2026-01-11
chore: simplify running mocha tests by auto requiring setup file

BridgeAR:
This adds the setup file to the .mocharc file. That way it is not necessary anymore to include that to run tests.


packages/dd-trace/test/setup/mocha.js
```diff
@@ -1,10 +1,10 @@
 'use strict'
 
-const assert = require('node:assert')
-const fs = require('node:fs')
-const { platform } = require('node:os')
-const path = require('node:path')
-const util = require('node:util')
```
simon-id, line 7: why remove the node prefix ? to run on older versions of node or something ?

BridgeAR: Yes, that is required for guardrails, since it is now included in all tests.

simon-id (approved):

package.json
```diff
@@ -18,60 +18,60 @@
...
+    "test:integration:openfeature": "mocha --timeout 60000 -r packages/dd-trace/test/setup/core.js integration-tests/openfeature/*.spec.js",
+    "test:integration:jest": "mocha --timeout 60000 -r packages/dd-trace/test/setup/core.js integration-tests/jest/*.spec.js",
+    "test:integration:mocha": "mocha --timeout 60000 -r packages/dd-trace/test/setup/core.js integration-tests/mocha/*.spec.js",
+    "test:integration:playwright": "mocha --timeout 60000 -r packages/dd-trace/test/setup/core.js integration-tests/playwright/*.spec.js",
+    "test:integration:selenium": "mocha --timeout 60000 -r packages/dd-trace/test/setup/core.js integration-tests/selenium/*.spec.js",
+    "test:integration:vitest": "mocha --timeout 60000 -r packages/dd-trace/test/setup/core.js integration-tests/vitest/*.spec.js",
+    "test:integration:testopt": "mocha --timeout 120000 -r packages/dd-trace/test/setup/core.js integration-tests/ci-visibility/*.spec.js",
+    "test:integration:profiler": "mocha --timeout 180000 -r packages/dd-trace/test/setup/core.js integration-tests/profiler/*.spec.js",
```
watson: The `packages/dd-trace/test/setup/mocha.js` file is now also required in all these cases, which it wasn't before. Is that going to be an issue, or was it actually an oversight before?

package.json
```diff
@@ -18,60 +18,60 @@
...
-    "test:aiguard": "mocha -r \"packages/dd-trace/test/setup/mocha.js\" \"packages/dd-trace/test/aiguard/**/*.spec.js\"",
-    "test:aiguard:ci": "nyc --no-clean --include \"packages/dd-trace/src/aiguard/**/*.js\" -- npm run test:aiguard",
-    "test:appsec": "mocha -r \"packages/dd-trace/test/setup/mocha.js\" --exclude \"packages/dd-trace/test/appsec/**/*.plugin.spec.js\" \"packages/dd-trace/test/appsec/**/*.spec.js\"",
-    "test:appsec:ci": "nyc --no-clean --include \"packages/dd-trace/src/appsec/**/*.js\" --exclude \"packages/dd-trace/test/appsec/**/*.plugin.spec.js\" -- npm run test:appsec",
-    "test:appsec:plugins": "mocha -r \"packages/dd-trace/test/setup/mocha.js\" \"packages/dd-trace/test/appsec/**/*.@($(echo $PLUGINS)).plugin.spec.js\"",
-    "test:appsec:plugins:ci": "yarn services && nyc --no-clean --include \"packages/dd-trace/src/appsec/**/*.js\" -- npm run test:appsec:plugins",
-    "test:debugger": "mocha -r 'packages/dd-trace/test/setup/mocha.js' 'packages/dd-trace/test/debugger/**/*.spec.js'",
+    "test": "SERVICES=* yarn services && node --expose-gc ./node_modules/mocha/bin/mocha.js packages/dd-trace/test/setup/node.js packages/*/test/**/*.spec.js",
```
watson: Did you mean to change this call to `node` instead of `mocha`, and if so why?

scripts/mocha-parallel-files.js
```diff
@@ -349,15 +351,21 @@ async function main () {
...
+        if (opts.require.length) {
+          options.require = opts.require
+        }
+
         const childEnv = {
           ...process.env,
-          MOCHA_RUN_FILE_CONFIG: JSON.stringify({
-            timeout,
-            color: true,
-            reporter: 'mocha-multi-reporters',
```
watson, line 356: These two options are no longer included in `options`. Was that on purpose?

package.json
BridgeAR: That was a mistake, I will remove that again. In fact, we should not even need the core require either anymore. Let me add that as well.

scripts/mocha-parallel-files.js
BridgeAR: Yes, they are always included in the script that is triggered from this one. So it is redundant and could in fact conflict at some point.

package.json
BridgeAR: Yes, this is a necessary change in case we want to use Node.js flags (--expose-gc). It was working before due to ... I forgot and have to look into that 😅 

I am actually considering to just always pass through --expose-gc. That should be fine for our tests and I see no downside. What do you think?

watson: Why do we have it here in the first place? The only downside is that if you actually try to use `global.gc` in the production code, our tests will not flag it. But I guess we could ESLint our way out of that one

BridgeAR: Some tests use it for runtime metrics for example. Also some other tests try to make sure there is no leak or that a GC run keeps a reference / cleans a reference.

watson: I guess accidental use of `global.gc` in production code would be caught by system and integration tests, so I think it's ok to have here 👍

BridgeAR: I decided to keep it explicit for now. I felt it is more intuitive to know where it is coming from than injecting it by the parallel script. Is it ok as is for you?

package.json
watson: In the newest iteration of this PR, the `-r packages/dd-trace/test/setup/core.js` is still removed and it's implicitly requiring `packages/dd-trace/test/setup/mocha.js`. Before approving this PR, I just want to make sure that's on purpose. Your comment above makes it seem like it might not be.

BridgeAR: The `core` part is required by setup/mocha.js. Thus, it is still loaded.

`setup/mocha.js` has the side effect of adding a afterEach hook that cleans up some state. That is not necessary in many tests but it does not hurt either. It should as such be fine to always load it. I think it is nicer to have the setup simpler and consistent.

PR #7219, opened 2026-01-12
refactor(test): improve type safety for spawned process URLs

watson:
### What does this PR do?

Refactors the `spawnProc` test helper to improve TypeScript type safety by splitting it into two distinct functions with clear lifecycle expectations and return types.

### Motivation

When using `spawnProc`, TypeScript would complain that `proc.url` might be undefined, even in cases where we know the process is long-running and will definitely send a port message. This was because the function could return either `SpawnedProcess` (with `url`) or `void` depending on whether the process exited or stayed alive.

This refactoring:
- Makes `spawnProc` always return `SpawnedProcess` with a guaranteed `url` property (rejects if process exits)
- Adds `spawnProcAndExpectExit` for processes that should run and exit cleanly (returns `void` on success)
- Implements `url` as a lazy getter that throws if accessed before the port message is received (prevents timing bugs)
- Provides clear JSDoc documentation explaining when to use each function

This ensures TypeScript correctly understands that `proc.url` is always defined when using `spawnProc`, eliminating false-positive type errors.

### Additional Notes

This is purely a refactoring of test infrastructure with no impact on production code.

**Changes to test helpers:**
- Added `spawnPluginIntegrationTestProcAndExpectExit` wrapper for short-lived plugin integration tests

**Updated tests:**
Updated various plugin integration tests and other test files to use the appropriate spawn function based on whether they spawn long-running servers (use `spawnProc`/`spawnPluginIntegrationTestProc`) or short-lived scripts that run and exit (use `spawnProcAndExpectExit`/`spawnPluginIntegrationTestProcAndExpectExit`).

### DIFF
--- a/integration-tests/helpers/index.js
+++ b/integration-tests/helpers/index.js
@@ -630,7 +630,7 @@ function checkSpansForServiceName (spans, name) {
 }
 
 /**
- * @typedef {Record<string, string|undefined> & { execArgv?: string[] }} AdditionalSpawnOptions
+ * @typedef {Record<string, string|undefined>} AdditionalEnvArgs
  */
 
 /**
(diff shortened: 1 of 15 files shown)


integration-tests/helpers/index.js
```diff
@@ -599,11 +670,66 @@ async function spawnPluginIntegrationTestProc (
...
+
+/**
+ * Spawns a plugin integration test process that runs a long-lived server.
+ *
+ * The spawned process should call `process.send({ port })` to signal it's ready.
+ * Use this for tests that spawn HTTP servers or other long-running processes.
+ *
+ * For short-lived scripts that run and exit, use `spawnPluginIntegrationTestProcAndExpectExit` instead.
+ *
+ * @overload
```
BridgeAR: I do not know if that is true or not but I found that using `overload` only in one will not remove the type errors in case the other one is matched (at least I noticed it so far in my editor).

I can see that also happening here in packages/datadog-plugin-anthropic/test/integration-test/client.spec.js

By adding overload to spawnPluginIntegrationTestProcAndExpectExit, it understands that 5 arguments are fine. Otherwise not.

watson: Fixed by refactoring the functions so we don't need the overloads at all

PR #7224, opened 2026-01-12
fix: only send process tags in first span of first chunk of a network request

tlhunter:
### What does this PR do?
- modifies behavior to only send process tags in first span of first chunk of a network request

### Motivation
- previous behavior is to send process tags in first span of every chunk of a network request
- the agent only needs to examine the tags in the first span so this is more efficient

### DIFF
--- a/packages/dd-trace/src/span_processor.js
+++ b/packages/dd-trace/src/span_processor.js
@@ -55,12 +55,18 @@ class SpanProcessor {
       this._gitMetadataTagger.tagGitMetadata(spanContext)
 
       let isFirstSpanInChunk = true
+      // Only add process tags to the first span of the first chunk
+      const shouldAddProcessTags = this._processTags && !trace._processTagsSent
+      if (shouldAddProcessTags) {
+        trace._processTagsSent = true
+      }
 
       for (const span of started) {
         if (span._duration === undefined) {
           active.push(span)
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/span_processor.js
```diff
@@ -55,12 +55,18 @@ class SpanProcessor {
       this._gitMetadataTagger.tagGitMetadata(spanContext)
 
       let isFirstSpanInChunk = true
+      // Only add process tags to the first span of the first chunk
+      const shouldAddProcessTags = this._processTags && !trace._processTagsSent
+      if (shouldAddProcessTags) {
+        trace._processTagsSent = true
+      }
```
simon-id, line 62: if the outgoing request fails to send the first spans, this will stay true and the process tags will never be sent no ?

BridgeAR (changes requested):
Was there anything wrong before? Would it not already have worked due to our `isFirstSpanInChunk` argument?

The test is actually not very helpful in my opinion due to testing just the direct functionality but not the outcome. So by changing the test to test the outcome instead of internal method calls, we can verify the actual behavior and determine the action better in my opinion.
Please update that no matter if the code needs actual adjustment or not as the current test makes assumptions that we should in my opinion not do.

PR #7231, opened 2026-01-13
[test optimization] Add automatic NX service detection

juan-fernandez:
### What does this PR do?

* If `process.env. DD_ENABLE_NX_SERVICE_NAME ` and `process.env.NX_TASK_TARGET_PROJECT` are both set and the user hasn't configured test service in any other way, `service` is set to `process.env.NX_TASK_TARGET_PROJECT`

### Motivation
Make it simpler for `nx` users to set an automatic test service when `nx` is testing multiple different projects. 

### Plugin Checklist

- [x] Unit tests.


packages/dd-trace/src/config.js
```diff
@@ -862,7 +864,12 @@ class Config {
     this.#setSamplingRule(target, 'sampler.rules', safeJsonParse(DD_TRACE_SAMPLING_RULES))
     unprocessedTarget['sampler.rules'] = DD_TRACE_SAMPLING_RULES
     this.#setString(target, 'scope', DD_TRACE_SCOPE)
-    this.#setString(target, 'service', DD_SERVICE || tags.service || OTEL_SERVICE_NAME)
+    // Priority: DD_SERVICE > tags.service > OTEL_SERVICE_NAME > NX_TASK_TARGET_PROJECT (if DD_IS_NX) > default
+    let serviceName = DD_SERVICE || tags.service || OTEL_SERVICE_NAME
+    if (!serviceName && isTrue(DD_IS_NX) && NX_TASK_TARGET_PROJECT) {
```
BridgeAR: Since this is just a fallback in case we have no serviceName defined at all and `NX_TASK_TARGET_PROJECT` would (hopefully) only be defined in the corresponding environment, I think we could just directly read it instead of having to add `DD_IS_NX`. A user would still be able to replace that by setting `DD_SERVICE`, in case they do not want that behavior.

juan-fernandez: without `DD_IS_NX` this is effectively a breaking change: customers which haven't set `DD_SERVICE` and are using `nx` see the default service name today and would see a different thing after this change

BridgeAR: Good point. What about changing it in our next major to be the new default? We could already implement it as such and name the env accordingly.

BridgeAR: Due to preventing the breaking change, I guess we do not get around adding the env. We could just potentially name it a bit more about what it stands for. So e.g., `DD_ENABLE_NX_SERVICE_NAME`

juan-fernandez: applied your suggestions in https://github.com/DataDog/dd-trace-js/pull/7231/changes/5afa390ff66456fa5c0f1fc357775ea46bd1dd0a

juan-fernandez:
> Hello 😄
> 
> I think this could be a bit confusing. Do we really need a flag to detect Nx? The existence of `NX_TASK_TARGET_PROJECT` already indicates Nx is being used.
> 
> Is the goal to let the consumer decide whether to use it or not? If so, I’d name it something like `DD_NX_AWARE`.
> 
> If we want the service name to default to the Nx project name, we can always read `NX_TASK_TARGET_PROJECT` than.

hey thanks for the feedback. We will use `DD_ENABLE_NX_SERVICE_NAME` which should be more explanatory. As for its necessity: if we didn't include it, this would be a breaking change, causing `nx` users to see a changing test service name. 

We're going to ship a breaking change in v6 though, removing the need for passing `DD_ENABLE_NX_SERVICE_NAME` altogether. 

> Also, `DD_SERVICE="env://NX_TASK_TARGET_PROJECT"` could be a good alternative if it’s supported and relevant.

We've decided not to support `env://$ENV_VAR` kind of syntax because it increases the likelihood of sensitive data being leaked. Since `NX_TASK_TARGET_PROJECT` is the only env var for which this is needed at the moment, we've preferred to add custom logic for it

PR #7241, opened 2026-01-14
fix(ci): make sure dependabot is secure and able to update everthing

BridgeAR:
- Use synchronize to fix running the dependabot on updates to the branch
- Separate the vendoring into two steps. One for installing and vendoring with read permissions only. The other for pushing with write permissions.
   This is safer, since it does not allow the install step to push to the branch.

### DIFF
--- a/.github/workflows/dependabot-automation.yml
+++ b/.github/workflows/dependabot-automation.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     types:
       - opened
+      - synchronize
 
 env:
   GROUPS: '["dev-minor-and-patch-dependencies", "gh-actions-packages", "test-versions"]'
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot-automation.yml
```diff
@@ -1,9 +1,10 @@
 name: 'Dependabot Automation'
 
 on:
-  pull_request:
+  pull_request_target:
```
KSerrania: The changes I made to the `dependabot-automation` policy should fix the `dd-octo-sts` errors we were getting. Therefore, we should be able to revert to:

```suggestion
  pull_request:
```

.github/workflows/dependabot-automation.yml
```diff
@@ -15,6 +16,26 @@ jobs:
...
+            exit 1
+          fi
+
+          case "${{ github.workflow_ref }}" in
+            DataDog/dd-trace-js/.github/workflows/dependabot-automation.yml@refs/heads/master) ;;
+            *)
+              echo "Unexpected workflow_ref; expected master workflow ref for STS trust policy."
+              exit 1
+              ;;
+          esac
```
KSerrania: The `dd-octo-sts` policy file takes care of verifying the workflow ref and event type, no need to verify again here

BridgeAR: Yes, this was just as simpler error message :)

I am fine to remove it again

BridgeAR: I removed it to simplify the PR

rochdev:
> Separate the vendoring into two steps. One for installing and vendoring with read permissions only. The other for pushing with write permissions. This is safer, since it does not allow the install step to push to the branch.

In what way is it safer? With the current setup, any bad actor would be able to just change the workflow so the contents of the individual jobs doesn't really matter, what matters is the Octo STS claims.

I'm a bit worried about adding so much untestable shell scripting if there isn't a strong argument for, so let's make that argument clear 😄

KSerrania (approved):
> In what way is it safer? With the current setup, any bad actor would be able to just change the workflow so the contents of the individual jobs doesn't really matter, what matters is the Octo STS claims.

The main idea behind separating steps is not to protect against someone already able to write to the repository.- in this case, they do not need to go through such lengths to get `contents: write` and `pull-requests: write` access, since they already have it.

The goal is to prevent malicious code that could run as part of the vendoring scripts (e.g. if it ends up being vulnerable to arbitrary code execution) from being able to get access to valuable credentials (write access to the repository).

Giving as few steps as possible sensitive access within a workflow limits this risk, which contributes to improving the security of the repository. However, security is about trade-offs - here, between keeping these scripts maintainable and reducing the access they may provide. If you deem this approach harmful to the maintainability of your repository, we can try to think about alternative ways to secure the repository.

PR #7248, opened 2026-01-14
feat(esbuild): warn when using --minify w/o --keep-names

tlhunter:
### What does this PR do?
- prints a warning when the user opts to minify ESBuild output but doesn't preserve class names

### Motivation
- this can have subtle breaking effects on the tracer

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -932,6 +932,46 @@ declare namespace tracer {
      * Configuration enabling LLM Observability. Enablement is superseded by the DD_LLMOBS_ENABLED environment variable.
      */
     llmobs?: llmobs.LLMObsEnableOptions
+
+    /**
+     * Configuration for Dynamic Instrumentation (Live Debugging).
+     */
+    dynamicInstrumentation?: {
+      /**
+       * Whether to enable Dynamic Instrumentation.
+       * @default false
+       */
+      enabled?: boolean
+
+      /**
+       * Path to a custom probes configuration file.
+       */
+      probeFile?: string
+
+      /**
+       * Timeout in milliseconds for capturing variable values.
+       * @default 100
+       */
+      captureTimeoutMs?: number
+
+      /**
+       * Interval in seconds between uploads of probe data.
+       * @default 1
+       */
+      uploadIntervalSeconds?: number
+
+      /**
+       * List of identifier names to redact in captured data.
+       * @default []
+       */
+      redactedIdentifiers?: string[]
+
+      /**
+       * List of identifier names to exclude from redaction.
+       * @default []
+       */
+      redactionExcludedIdentifiers?: string[]
+    }
   }
 
   /**


packages/datadog-esbuild/index.js
```diff
@@ -97,6 +97,10 @@ function getGitMetadata () {
 }
 
 module.exports.setup = function (build) {
+  if (build.initialOptions.minify && !build.initialOptions.keepNames && DEBUG) {
+    console.warn('WARNING: using --minify without --keep-names will break some dd-trace behavior.')
```
```suggestion
    process.emitWarning(
      'WARNING: using --minify without --keep-names will break some dd-trace behavior and may cause the application to fail.',
      { code: 'DATADOG_0001' }
    )
```

packages/datadog-esbuild/index.js
```diff
@@ -97,6 +97,13 @@ function getGitMetadata () {
 }
 
 module.exports.setup = function (build) {
+  if (build.initialOptions.minify && !build.initialOptions.keepNames && DEBUG) {
```
```suggestion
  if (build.initialOptions.minify && !build.initialOptions.keepNames) {
```

tlhunter:
@BridgeAR I can modify this to throw an error and fail the build process entirely (if that's what you mean by bail). I'd rather do that than display a warning and continue the build but without the tracer doing any instrumentation.

The main reason being that if a user is intentionally choosing to include the plugin then they obviously want datadog present.

It would also be weird to succeed the build but without instrumentation in place. Since the user's code has `require('dd-trace')` in it the bundle _will_ contain the tracer even if all of the require call interposing that the plugin does is disabled.

integration-tests/appsec/iast-esbuild-cjs/esbuild.common-config.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
   entryPoints: ['app.js'],
   bundle: true,
   minify: true,
+  keepNames: true,
```
tlhunter, line 9: FYI @CarlesDD. Not sure if this test depended on the broken behaviors of combining minify without keepNames.

packages/datadog-esbuild/index.js
```diff
@@ -98,9 +98,8 @@ function getGitMetadata () {
 
 module.exports.setup = function (build) {
   if (build.initialOptions.minify && !build.initialOptions.keepNames) {
-    process.emitWarning(
-      'WARNING: using --minify without --keep-names will break some dd-trace behavior.',
-      { code: 'DATADOG_0001' }
+    throw new Error(
+      'ERROR: using --minify without --keep-names will break some dd-trace behavior. Refusing to bundle the tracer.'
```
```suggestion
      'Using --minify without --keep-names will break some dd-trace behavior. Refusing to bundle the tracer.'
```

BridgeAR (dismissed):

integration-tests/esbuild/build-and-test-minify.js
```diff
@@ -0,0 +1,35 @@
...
+  target: ['node18'],
+  external: [
+    'knex'
+  ]
+}).then(() => {
+  console.error('Expected build to throw an error, but it succeeded')
+  process.exit(1)
+}).catch((err) => {
+  assert(
+    err.message.includes('using --minify without --keep-names'),
```
BridgeAR:
```suggestion
    err.message.includes('--minify without --keep-names'),
```

PR #7251, opened 2026-01-15
fix(test-optimization): jest instrumentation applies auto test retries on top of other retry mechanisms

calvinbayer:
### What does this PR do?
Introduce an order of priority for applying retry mechanisms:
- Attempt to Fix
- Early Flake Detection
- Auto Test Retries

This change aligns the behavior of the Jest instrumentation with how other tracers do it. It also decreases the number of retries and thus the performance overhead introduced by instrumenting tests with Test Optimization. E.g., before this change, if both EFD and ATR were enabled, a new failing test would be executed 66 times: `(1 execution + 10 EFD retries) * (1 execution + 5 ATR retries)`. After this change, ATR retries will not be applied anymore and this test is only executed 11 times as expected.

### Motivation
Working on adding the `@test.final_status` tag had unexpected side effects with multiple retry mechanisms being applied.

### DIFF
--- a/integration-tests/jest/jest.spec.js
+++ b/integration-tests/jest/jest.spec.js
@@ -4167,7 +4167,7 @@ describe(`jest@${JEST_VERSION} commonJS`, () => {
         ])
       })
 
-      it('attempt to fix takes precedence over EFD', async () => {
+      it('attempt to fix takes precedence over EFD for new tests', async () => {
         const NUM_RETRIES_EFD = 2
         receiver.setKnownTests({ jest: {} })
         receiver.setSettings({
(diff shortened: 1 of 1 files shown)


anmarchenko (approved):

integration-tests/jest/jest.spec.js
```diff
@@ -2055,6 +2055,46 @@ describe(`jest@${JEST_VERSION} commonJS`, () => {
   })
 
   context('early flake detection', () => {
+    it('takes precedence over flaky test retries for new tests', (done) => {
```
juan-fernandez: could we try to make the test `async` instead of using `done`? I've been trying to get rid of `done`

integration-tests/jest/jest.spec.js
```diff
@@ -4075,6 +4115,116 @@ describe(`jest@${JEST_VERSION} commonJS`, () => {
...
+          early_flake_detection: {
+            enabled: true,
+            slow_test_retries: {
+              '5s': NUM_RETRIES_EFD
+            },
+            faulty_session_threshold: 100
+          },
+          known_tests_enabled: true
+        }
+        )
```
juan-fernandez: nitpick
```suggestion
        })
```

integration-tests/jest/jest.spec.js
```diff
@@ -4075,6 +4115,116 @@ describe(`jest@${JEST_VERSION} commonJS`, () => {
...
+            stdio: 'inherit'
+          }
+        )
+
+        await Promise.all([
+          once(childProcess, 'exit'),
+          eventsPromise
+        ])
+      })
+
```
juan-fernandez, line 4228: should we also check that attempt to fix takes precedence over retries triggered by `impacted_tests_enabled: true`?

packages/datadog-instrumentations/src/jest.js
```diff
@@ -434,19 +439,23 @@ function getWrappedEnvironment (BaseEnvironment, jestVersion) {
...
-            this.retryTest({
-              jestEvent: event,
-              retryCount: testManagementAttemptToFixRetries,
-              retryType: 'Test Management (Attempt to Fix)'
-            })
-          }
+        const isAttemptToFix = this.isTestManagementTestsEnabled &&
+          this.testManagementTestsForThisSuite?.attemptToFix?.includes(testFullName)
+        if (
+          this.isTestManagementTestsEnabled &&
```
juan-fernandez: nitpick: this is redundant as it's already a requirement for `isAttemptToFix`
```suggestion
```

integration-tests/jest/jest.spec.js
calvinbayer: Done 🫡

integration-tests/jest/jest.spec.js
calvinbayer: I was surprised to see we don't have code formatting as part of linting. Fixed it manually

integration-tests/jest/jest.spec.js
calvinbayer: Good point, added a test for that too

packages/datadog-instrumentations/src/jest.js
calvinbayer: Good catch, removed that part of the condition

PR #7259, opened 2026-01-16
Apply per-endpoint limits to active requests

bojbrook:
### What does this PR do?
This PR introduces per-endpoint active request limits instead of a single global limit.

This ensures that slow or high-latency endpoints cannot monopolize the active requests, preventing unrelated endpoints from having their payloads dropped under load.

### Motivation
Working with @BridgeAR on issue APMS-16671 where the customer was hitting the max requests limit of the tracer

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].

### DIFF
--- a/packages/dd-trace/src/datastreams/writer.js
+++ b/packages/dd-trace/src/datastreams/writer.js
@@ -41,10 +41,6 @@ class DataStreamsWriter {
   }
 
   flush (payload) {
-    if (!request.writable) {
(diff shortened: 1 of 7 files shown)


BridgeAR:
Since we now measure bytes instead of requests, it is probably best to close this and the root cause for the issue was fixed as well.

Issue #6389, opened 2025-09-06
[BUG]: Memory leak when http integration is enabled in Nestjs


elton-okawa:
### Tracer Version(s)

5.65.0

### Node.js Version(s)

22.18.0

## Problem
When http instrumentation is loaded, memory is not released

Might be related to: https://github.com/DataDog/dd-trace-js/issues/5780

## Setup

I'd found this issue during my work and I was able to create this minimal setup that replicated the exact scenario I'm facing, it has the following features:

- `dd-tracer` version `5.65.0`
- `node` version `22.18.0`
- Nestjs
- `AppService`, `OtherService` and `DataRepository` with `scope.REQUEST`
- `AppService` injects both `OtherService` and `DataRepository`
- `AppService` reads data from `DataRepository` and provides it to `OtherService`

## Observations

- When `http` integration is **disabled** there's **no leak**
- When `http` integration is **enabled** there's **a leak**
- When `http` integration is **enabled** and `OtherService` releases provided data reference, there's **no leak**
- When `http` integration is **enabled** and using Node.js `24.7.0` there's **no leak**
  - **Obs**: there's no leak for the provided project but for the production one the leak still happens

Here's the heapsnapshot for each case (I couldn't upload directly to GitHub)
- [Enabling/disabling http](https://drive.google.com/file/d/1ai3jqhYOOL0YvWnOZXtSHKc2TZjtdiFR/view?usp=sharing)
- [Using experimental async context](https://drive.google.com/file/d/12ee831J2gucH04RDn4oVCKxInz19xCJY/view?usp=sharing)

## Workaround

Our current solution is releasing provided data reference from `OtherService`.

### Reproduction Code

https://github.com/elton-okawa/dd-tracer-memory-leak

### Tracer Config

```
[...]

PR #7276, opened 2026-01-19
fix: limit memory leak to AppSec being enabled

BridgeAR:
<s>This fixes a memory leak by making sure the exposed req and res of
http are not hold onto strongly.
It does that by skipping adding these to the store, if not needed
as well as creating a WeakRef when it is needed. That way AppSec
still has access to these as long as the request is alive.</s>

This limits a AppSec specific memory leak to only surface in case AppSec
is enabled. That is just a stop gap to limit the impact on our users until we
have a proper fix that needs further rework.

Refs: https://github.com/DataDog/dd-trace-js/issues/6389

Router now correlates the storeStacks to the correct middleware exit.
Before, these were expected to always be in order while multiple parallel
calls could potentially interfere with that and correlate wrong entries.

### DIFF
--- a/packages/datadog-instrumentations/test/passport-http.spec.js
+++ b/packages/datadog-instrumentations/test/passport-http.spec.js
@@ -17,7 +17,11 @@ withVersions('passport-http', 'passport-http', version => {
     let port, server, subscriberStub
 
     before(() => {
-      return agent.load(['http', 'express', 'passport', 'passport-http'], { client: false })
+      return agent.load(
+        ['http', 'express', 'passport', 'passport-http'],
+        { client: false },
+        { appsec: { enabled: true } }
+      )
     })
 
     before((done) => {
--- a/packages/datadog-instrumentations/test/passport-local.spec.js
+++ b/packages/datadog-instrumentations/test/passport-local.spec.js
@@ -17,7 +17,11 @@ withVersions('passport-local', 'passport-local', version => {
     let port, server, subscriberStub
 
     before(() => {
-      return agent.load(['http', 'express', 'passport', 'passport-local'], { client: false })
+      return agent.load(
+        ['http', 'express', 'passport', 'passport-local'],
+        { client: false },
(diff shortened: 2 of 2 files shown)


packages/datadog-plugin-http/src/server.js
```diff
@@ -32,11 +31,31 @@ class HttpServerPlugin extends ServerPlugin {
...
-    this.enter(span, { ...store, req, res })
-
     const context = web.getContext(req)
 
+    if (context) {
+      context.parentStore = store
+    }
+
+    if (incomingHttpRequestStart.hasSubscribers) {
+      const weakReq = new WeakRef(req)
```
rochdev: This might actually introduce a memory leak, along with a performance degradation. We've used `WeakRef` in the past for a few things and it ended up holding things in memory for longer. Instead we should update the HTTP integration to properly use `runStores` and `bindStore` so that the storage itself handles cleaning up, which wouldn't require any manual handling.

If you prefer to still go with this approach for now to fix an immediate issue I would say to add a TODO to remove this later once we have transitioned to diagnostics channel handling this for us. I'm also curious how this was tested, because again, in the past we've tried the same strategy without much success.

packages/datadog-plugin-router/src/index.js
```diff
@@ -9,50 +9,59 @@ const { COMPONENT } = require('../../dd-trace/src/constants')
 class RouterPlugin extends WebPlugin {
   static id = 'router'
 
+  #storeStacks = new WeakMap()
```
rochdev, line 12: These changes are a nice to have, but unrelated to this PR and in the future it would be best to make those unrelated changes in another PR to avoid unnecessary noise.

packages/datadog-plugin-http/src/server.js
BridgeAR: This is a draft WIP with failing tests ;)

PR #7278, opened 2026-01-19
refactor(config): simplify remote config merging logic

watson:
### What does this PR do?

Refactors the `getMergedLibConfig()` method in the remote config manager to use simpler iteration and overwrite semantics for better performance and readability.

### Motivation

The original implementation used a "highest-to-lowest priority with first-wins" approach that required checking if keys already existed before setting them. By reversing the sort order to lowest-to-highest priority, we can leverage JavaScript's natural "last write wins" behavior, eliminating unnecessary checks and making the code simpler to understand.

### DIFF
--- a/packages/dd-trace/src/config/remote_config.js
+++ b/packages/dd-trace/src/config/remote_config.js
@@ -133,28 +133,16 @@ class RCClientLibConfigManager {
   getMergedLibConfig () {
     if (this.configs.size === 0) return null
 
-    const sortedConfigs = [...this.configs.values()]
+    let hasLibConfig = false
+
+    const merged = [...this.configs.values()]
       .sort((a, b) => a.priority - b.priority)
+      .reduce((merged, { conf }) => {
+        if (conf.lib_config != null) hasLibConfig = true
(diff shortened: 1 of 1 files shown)


khanayan123 (approved):

packages/dd-trace/src/config/remote_config.js
```diff
@@ -126,47 +126,32 @@ class RCClientLibConfigManager {
...
-
-        // Set the value even if it's null (to reset) but not if it's undefined (missing)
-        if (value === null) {
-          merged[key] = undefined // TODO: Should this be null?
-        } else if (value !== undefined) {
-          merged[key] = value
+      // Filter out undefined (but keep null) since undefined means "not set" and null means "reset to default"
+      for (const key in conf.lib_config) {
+        if (conf.lib_config[key] !== undefined) {
+          merged[key] = conf.lib_config[key]
```
BridgeAR:
```suggestion
      for (const [key, value] of Object.entries(conf.lib_config)) {
        if (value !== undefined) {
          merged[key] = value
```

PR #7279, opened 2026-01-19
fix: improve baggage implementation

BridgeAR:
This addresses multiple baggage issues and adds further documentation to the new implementation.
The regular API is now spec compliant by ignoring non string keys as well as returning the new context consistently when needed.

Baggage headers are also improved by not accepting arbitrary strings as keys anymore. They now adhere to the spec as outlined in the code comment.
In addition, the parsing now handles optional properties that may be attached.

TODO: Add system tests later.


packages/dd-trace/src/baggage.js
```diff
@@ -1,41 +1,66 @@
...
+const baggageStorage =
+  /** @type {{ enterWith: (store?: BaggageStore) => void, getStore: () => (BaggageStore | undefined) }} */ (
+    /** @type {unknown} */ (storage('baggage'))
+  )
+
+// TODO: Implement metadata https://opentelemetry.io/docs/specs/otel/baggage/api/#set-value
 /**
  * @param {string} key
  * @param {string} value
+ * @param {object} [metadata] Not used yet
```
watson, line 25: Does this require a TODO?

packages/dd-trace/src/baggage.js
```diff
@@ -1,41 +1,66 @@
...
+  const store = baggageStorage.getStore() ?? {}
+  const { [keyToRemove]: _, ...newBaggage } = store
+  baggageStorage.enterWith(newBaggage)
   return newBaggage
 }
 
 function removeAllBaggageItems () {
-  storage('baggage').enterWith()
-  return storage('baggage').getStore()
+  const newContext = /** @type {BaggageStore} */ ({})
```
pabloerhard, line 61: nit: wouldn't  something like emptyContext be more suitable for this method?
```suggestion
  const emptyContext = /** @type {BaggageStore} */ ({})
```

packages/dd-trace/src/opentracing/propagation/text_map.js
```diff
@@ -668,19 +665,38 @@ class TextMapPropagator {
...
         tracerMetrics.count('context_header_style.malformed', ['header_style:baggage']).inc()
         removeAllBaggageItems()
         return
       }
-      let [key, value] = keyValue.split('=')
-      key = this._decodeOtelBaggageKey(key.trim())
-      value = decodeURIComponent(value.trim())
-      if (!key || !value) {
+
+      const key = member.slice(0, eqIdx).trim()
```
pabloerhard, line 692: What was the reason behind moving from .split() to two separate .slice() calls?
It seems the original approach avoided three operations (indexOf and two slices).

bengl: I'd guess this is because of a bug that might occur if there are two `=` in the string.

index.d.ts
```diff
@@ -161,18 +161,54 @@ interface Tracer extends opentracing.Tracer {
...
    * Provide same functionality as OpenTelemetry Baggage:
    * https://opentelemetry.io/docs/concepts/signals/baggage/
    *
    * Since the equivalent of OTel Context is implicit in dd-trace-js,
    * these APIs act on the currently active baggage
    *
-   * Work with storage('baggage'), therefore do not follow the same continuity as other APIs
+   * Work with storage('baggage'), therefore do not follow the same continuity as other APIs.
+   */
+  setBaggageItem (key: string, value: string, metadata?: object): Record<string, string>;
```
bengl, line 179: This is a semver-minor change. Will update the label accordingly.

BridgeAR: Well... metadata has no functionality so far though. It will just be ignored. The spec is very vague and it is still not clear for me what to do there. I did look at the Otel SDK and it wraps the value into an object and returns it as toString. Not sure this is really what we want to do.

packages/dd-trace/src/baggage.js
BridgeAR: Is a new context not by default empty? I feel new is better :)

packages/dd-trace/src/opentracing/propagation/text_map.js
BridgeAR: Correct @bengl 

That aside: split is way slower than using indexOf and slice. A high level method can hide the internal CPU operations needed. Think about what a method is for and how it likely has to be implemented on a lower level. In that case, the overhead is showing up more.

pabloerhard: Nice, good to know!

index.d.ts
bengl: 🤷 an addtion to the documented args is still semver minor. It's fine. We do a lot of minor releases anyway.

packages/dd-trace/src/baggage.js
```diff
@@ -1,41 +1,66 @@
...
+
+/**
+ * @type {{ enterWith: (store?: BaggageStore) => void, getStore: () => (BaggageStore | undefined) }}
+ */
+const baggageStorage =
+  /** @type {{ enterWith: (store?: BaggageStore) => void, getStore: () => (BaggageStore | undefined) }} */ (
+    /** @type {unknown} */ (storage('baggage'))
+  )
+
+// TODO: Implement metadata https://opentelemetry.io/docs/specs/otel/baggage/api/#set-value
```
rachelyangdog, line 21: IIRC most of the other tracers do not support adding metadata. Was the metadata parameter added for API compatibility?

packages/dd-trace/src/baggage.js
BridgeAR: I have some in other spots. https://github.com/DataDog/dd-trace-js/pull/7279/files#diff-f43ed8423ffd07b7ecf5128f40262b5c1df985cec1e567e6b542a7f0f954a405R21

packages/dd-trace/src/baggage.js
BridgeAR: We do not yet support that either, it is more about laying the foundation for that. We will ignore it as before.

rachelyangdog (approved):

PR #7281, opened 2026-01-20
fix: load files correctly for Cucumber and Playwright

BridgeAR:
The require calls were not correctly aligned. This was detected due to type errors being visible for the wrong require calls.

As drive-by add a few additional types.

TODO: For landing this, we should add tests. @juan-fernandez could you please take a look at that?

### DIFF
--- a/integration-tests/jest/jest.spec.js
+++ b/integration-tests/jest/jest.spec.js
@@ -525,13 +525,11 @@ describe(`jest@${JEST_VERSION} commonJS`, () => {
         })
       })
 
-      it('sends telemetry with test_session metric when telemetry is enabled', async () => {
-        const envVars = reportingOption === 'agentless'
-          ? getCiVisAgentlessConfig(receiver.port)
-          : getCiVisEvpProxyConfig(receiver.port)
-        if (reportingOption === 'evp proxy') {
-          receiver.setInfoResponse({ endpoints: ['/evp_proxy/v4'] })
-        }
+      // TODO: This should also run in agentless mode
+      const maybeSkippped = reportingOption === 'evp proxy' ? it : it.skip
+      maybeSkippped('sends telemetry with test_session metric when telemetry is enabled', async () => {
+        const envVars = getCiVisEvpProxyConfig(receiver.port)
+        receiver.setInfoResponse({ endpoints: ['/evp_proxy/v4'] })
 
         const telemetryPromise = receiver
           .gatherPayloadsMaxTimeout(({ url }) => url.endsWith('/api/v2/apmtelemetry'), (payloads) => {
--- a/integration-tests/mocha/mocha.spec.js
+++ b/integration-tests/mocha/mocha.spec.js
@@ -171,6 +171,7 @@ describe(`mocha@${MOCHA_VERSION}`, function () {
     })
   })
 
+  // TODO: This should also run in agentless mode
   it('sends telemetry with test_session metric when telemetry is enabled', async () => {
     receiver.setInfoResponse({ endpoints: ['/evp_proxy/v4'] })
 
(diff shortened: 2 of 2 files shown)


integration-tests/jest/jest.spec.js
```diff
@@ -525,6 +525,42 @@ describe(`jest@${JEST_VERSION} commonJS`, () => {
         })
       })
 
+      if (reportingOption === 'evp proxy') {
```
juan-fernandez: telemetry is also sent when working in agentless mode, so we should add a test for that

integration-tests/jest/jest.spec.js
```diff
@@ -525,6 +525,42 @@ describe(`jest@${JEST_VERSION} commonJS`, () => {
...
+
+          childProcess = exec(
+            runTestsCommand,
+            {
+              cwd,
+              env: {
+                ...getCiVisEvpProxyConfig(receiver.port),
+                DD_TRACE_AGENT_PORT: String(receiver.port),
+                DD_INSTRUMENTATION_TELEMETRY_ENABLED: 'true',
+                SHOULD_CHECK_RESULTS: '1',
```
juan-fernandez: is this needed?

integration-tests/mocha/mocha.spec.js
```diff
@@ -171,6 +171,39 @@ describe(`mocha@${MOCHA_VERSION}`, function () {
...
+
+        assert.ok(testSessionMetric, 'test_session telemetry metric should be sent')
+      })
+
+    childProcess = exec(
+      runTestsCommand,
+      {
+        cwd,
+        env: {
+          ...getCiVisEvpProxyConfig(receiver.port),
```
juan-fernandez, line 194: same as above, telemetry is also sent via agentless

integration-tests/playwright/playwright.spec.js
```diff
@@ -277,6 +277,38 @@ versions.forEach((version) => {
       })
     })
 
+    it('sends telemetry with test_session metric when telemetry is enabled', async () => {
```
juan-fernandez: couldn't this test be merged with `sends telemetry for RUM browser tests when telemetry is enabled` ?

integration-tests/selenium/selenium.spec.js
```diff
@@ -42,7 +43,11 @@ versionRange.forEach(version => {
       cwd = sandboxCwd()
 
       webAppServer.listen(0, () => {
-        webAppPort = webAppServer.address().port
+        const address = webAppServer.address()
+        if (!address || typeof address === 'string') {
```
juan-fernandez, line 47: why is this necessary?

integration-tests/selenium/selenium.spec.js
```diff
@@ -97,24 +102,43 @@ versionRange.forEach(version => {
...
+              })
+            })
+
           childProcess = exec(
             command,
             {
               cwd,
               env: {
-                ...getCiVisAgentlessConfig(receiver.port),
+                ...getCiVisEvpProxyConfig(receiver.port),
```
juan-fernandez, line 128: this is only testing EVP proxy, so I'd say we need to add agentless too

integration-tests/selenium/selenium.spec.js
```diff
@@ -127,10 +151,10 @@ versionRange.forEach(version => {
           cwd,
           env: {
             ...getCiVisAgentlessConfig(receiver.port),
+            DD_CIVISIBILITY_AGENTLESS_ENABLED: '1',
```
juan-fernandez: why is this needed?

integration-tests/selenium/selenium.spec.js
```diff
@@ -140,12 +164,16 @@ versionRange.forEach(version => {
...
+      if (childProcess.stdout) {
+        childProcess.stdout.on('data', (chunk) => {
+          testOutput += chunk.toString()
+        })
+      }
+      if (childProcess.stderr) {
+        childProcess.stderr.on('data', (chunk) => {
+          testOutput += chunk.toString()
+        })
+      }
```
juan-fernandez: is this needed? I don't think I've ever seen errors with accessing `stderr` or `stdout`. If it is needed, I wonder if we need it in other places

integration-tests/cucumber/cucumber.spec.js
```diff
@@ -110,6 +110,38 @@ describe(`cucumber@${version} commonJS`, () => {
...
   })
 
+  it('sends telemetry with test_session metric when telemetry is enabled', async () => {
+    receiver.setInfoResponse({ endpoints: ['/evp_proxy/v4'] })
+
+    const telemetryPromise = receiver
+      .gatherPayloadsMaxTimeout(({ url }) => url.endsWith('/api/v2/apmtelemetry'), (payloads) => {
+        const telemetryMetrics = payloads.flatMap(({ payload }) => payload.payload.series)
+
+        const testSessionMetric = telemetryMetrics.find(
```
juan-fernandez, line 120: nitpick: why not use the chance to check other telemetry tags?

BridgeAR: Can you tell me which ones I shall check for? I do not know this code that well :D

integration-tests/selenium/selenium.spec.js
BridgeAR: It fixes a type issue and makes it easier to debug if this would ever fail

integration-tests/selenium/selenium.spec.js
BridgeAR: That is a type fix. Let me remove this for now. I think we should use a different pattern to fix the reported type issues

integration-tests/cucumber/cucumber.spec.js
juan-fernandez: at least https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/ci-visibility/telemetry.js#L53-L55 for different `event_type` (corresponding to test levels). I can do it since I have more context though

PR #7285, opened 2026-01-20
doc: update AGENTS.md for clarity and less tokens

BridgeAR:
This adds a few improvements around handling types as well as how to write the code (using optional chaining and no for-in) while minimizing the document size. That way we use less tokens just for the reading this and overall gain better code output.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,9 +8,14 @@
 
 ## Setup
 
-- Install dependencies: `yarn install`
+**Package manager policy:**
 
-**This project uses yarn, not npm. Always use `yarn` commands instead of `npm` commands.**
+- Use **yarn only for installing dependencies and services**:
+  - `yarn install`
+  - `yarn services`
+- Use **npm for running scripts and other commands**: `npm run <script>`
+- In this repo, **everything else** (tests, lint, build, etc.) should use **npm**, not yarn.
+- `yarn services` is the only non-install yarn command: it sets up test service/plugin dependencies.
 
 ## Project Overview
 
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -129,6 +120,16 @@ Never use the `doesNotThrow()` assertion. Instead, execute the method directly.
 
 ## Code Style & Linting
 
+### Style
```
sabrenner, line 123: should we add a line for `use strict`? i admittedly don't use agents much to code/write new files, so maybe it just does it automatically, but i believe we have a linter check for this, which falls in line with the other points here (ish), so maybe it could be a callout, but let me know!

AGENTS.md
```diff
@@ -311,31 +253,8 @@ In case an issue is actually happening outside of dd-trace, suggest to fix it up
 **New instrumentations go in `packages/datadog-instrumentations/`.** The instrumentation system uses diagnostic channels for communication.
 
 Many integrations have corresponding plugins in `packages/datadog-plugin-*/` that work with the instrumentation layer.
-
-### What Are Plugins?
```
sabrenner, line 315: just a question - does _not_ having this make a difference? i would think that having this extra context on how auto-instrumentation works could be good for an LLM, but, I could be wrong. I think this is maybe the only place we would want to be a bit verbose as we've become very opinionated on how we do our Plugin system/what we do and/or don't put in instrumentation/publishers vs subscribers, etc., but curious on your thoughts.

AGENTS.md
```diff
@@ -158,30 +159,13 @@ Separate groups with empty line, sort alphabetically within each:
...
 
 - Use modern JS features supported by Node.js (e.g., optional chaining `?.`, nullish coalescing `??`)
 - Use `undefined` over `null`, if not required otherwise
-- Guard newer APIs with version checks using [`version.js`](./version.js):
-
-  ```js
-  const { NODE_MAJOR } = require('./version')
-  if (NODE_MAJOR >= 20) { /* Use Node.js 20+ API */ }
-  ```
+- Guard newer APIs with version checks using [`version.js`](./version.js)
```
sabrenner, line 168:
```suggestion
- Guard newer Node.js APIs with version checks using [`version.js`](./version.js)
```

when i first read this i thought it was referring to APIs we write 😅 but feel free to ignore if you think this is sorta a moot point, I don't mind either way

sabrenner (approved):
looks good to me, i think this cuts down on context fed to agents succinctly! just added a few comments/questions, but not blocking, i think the cut down here does its job

AGENTS.md
BridgeAR: I added things that were happening to me. I didn't see use strict as an issue :)

We also do not have eslint rules for all of these entries (e.g., for-in and optional chaining)

AGENTS.md
BridgeAR: I think we do not often add a plugin and having a dedicated documentation for that should be sufficient. The overall structure is also already likely enough for most cases.

AGENTS.md
BridgeAR: I just looked at it again. It is visible from the context above and also explained in a part below. I would keep it due to that.

PR #7293, opened 2026-01-21
fix log publish format

rochdev:
### What does this PR do?

Fix log publish format.

### Motivation

The format was changed a while ago which broke all subscribers except the user logger writer. This is because a lot of the logic to reformat the log entry was added to the subscriber, meaning it never applied to the other subscribers. Instead of duplicating that code everywhere, or having to expose helpers to be reused in every subscriber, I thought it made more sense to just reformat in the original format right at the source in the publisher.

### DIFF
--- a/packages/dd-trace/test/flare.spec.js
+++ b/packages/dd-trace/test/flare.spec.js
@@ -3,7 +3,6 @@
 const assert = require('node:assert/strict')
 const http = require('node:http')
 
-const { channel } = require('dc-polyfill')
 const express = require('express')
 const upload = require('multer')()
 const proxyquire = require('proxyquire').noCallThru()
@@ -11,9 +10,8 @@ const { describe, it, beforeEach, afterEach } = require('mocha')
 
 const { assertObjectContains } = require('../../../integration-tests/helpers')
 require('./setup/core')
+const log = require('../src/log')
 const { getConfigFresh } = require('./helpers/config')
-
-const debugChannel = channel('datadog:log:debug')
 
 describe('Flare', () => {
   let flare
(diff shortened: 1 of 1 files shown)


packages/dd-trace/test/log.spec.js
```diff
@@ -457,8 +457,8 @@ describe('log', () => {
 
         sinon.assert.calledOnce(console.error)
         const consoleErrorArg = console.error.getCall(0).args[0]
-        assert.strictEqual(typeof consoleErrorArg, 'object')
-        assert.strictEqual(consoleErrorArg.message, 'message')
+        assert.strictEqual(typeof consoleErrorArg, 'string')
```
rochdev, line 460: Not sure why I had to change this one. It makes sense to me that it would be a string, but then I don't know why it would have been an object before. cc @iunanua

packages/dd-trace/src/log/index.js
```diff
@@ -133,7 +130,25 @@ const log = {
...
 }
 
-logWriter.setStackTraceLimitFunction(log.error)
+function publishFormatted (ch, formatter, ...args) {
+  if (ch.hasSubscribers) {
+    const log = Log.parse(...args)
+    const { formatted, cause } = getErrorLog(log)
+
+    // calling twice ch.publish() because Error cause is only available in Node.js v16.9.0
+    // TODO: replace it with Error(message, { cause }) when cause has broad support
```
BridgeAR, line 139: So we effectively currently send two errors in case we send these to us?

packages/dd-trace/test/log.spec.js
BridgeAR: It should be the string. Otherwise the receiver has to handle that and I do not think that makes sense, since we want to control that fully.

packages/dd-trace/src/log/index.js
rochdev: This was already the case, this part was copied.

rochdev:
> It should be the string. Otherwise the receiver has to handle that and I do not think that makes sense, since we want to control that fully.

That's also my thought, I just don't understand why it was an object before here.

PR #7295, opened 2026-01-21
fix activeRequests leak when requests timeout on Bun

Maru32768:
### What does this PR do?

Ensures `activeRequests` is decremented when a request terminates via `'close'`, even if `'error'` is not emitted, by finalizing cleanup exactly once at the end of the request lifecycle.

### Motivation

When running dd-trace-js on Bun, requests could stop being sent after `activeRequests` was exhausted.
This happens because `request.abort` on timeout may terminate a request without emitting an `'error'` event, leaving `activeRequests` undecremented.

### DIFF
--- a/packages/dd-trace/src/exporters/common/request.js
+++ b/packages/dd-trace/src/exporters/common/request.js
@@ -128,16 +128,14 @@ function request (data, options, callback) {
     }
 
     activeRequests++
-    const finalize = (() => {
-      // Ensures decrementing activeRequests only once across multiple event handlers
-      let called = false
-      return () => {
-        if (!called) {
-          activeRequests--
-          called = true
-        }
+    // Ensures decrementing activeRequests only once across multiple event handlers
+    let called = false
+    const finalize = () => {
+      if (!called) {
+        activeRequests--
+        called = true
       }
-    })()
+    }
 
     const store = storage('legacy').getStore()
 


packages/dd-trace/src/exporters/common/request.js
```diff
@@ -126,17 +128,30 @@ function request (data, options, callback) {
...
+    const finalize = (() => {
+      // Ensures decrementing activeRequests only once across multiple event handlers
+      let called = false
+      return () => {
+        if (!called) {
+          activeRequests--
+          called = true
+        }
+      }
+    })()
```
BridgeAR: I am not sure why an IIFE is being used? Should it not be fine to just define it without?

```suggestion
    // Ensures decrementing activeRequests only once across multiple event handlers
    let called = false
    const finalize = () => {
      if (!called) {
        activeRequests--
        called = true
      }
    }
```

BridgeAR (commented):
Thank you for the PR. It has a strong overlap with https://github.com/DataDog/dd-trace-js/pull/7284. I hope it is fine to just land the PR that addresses the comments first :)

packages/dd-trace/src/exporters/common/request.js
Maru32768: It's just my coding style.
I'll commit it.

PR #7296, opened 2026-01-21
use POST for DogStatsD HTTP proxy requests

Maru32768:
### What does this PR do?

Switches the DogStatsD HTTP proxy to use `POST` when sending custom metrics payloads, ensuring the request body is always transmitted.

### Motivation

When running dd-trace-js on Bun, metrics payloads sent via `GET` requests were dropped because the request body was omitted.
This caused the agent to wait for the body, eventually timing out.


BridgeAR (approved):

bengl (approved):
This is fine by me, but I'd like to get someone from the trace agent team to verify.

PR #7297, opened 2026-01-21
test: fix type issues in test optimization tests

BridgeAR:
These are mostly passing through stdio as argument to exec, which does not support that argument.

In addition, fix a couple of process.env types and add some more types to test helpers.

There is one spot where `parallel` was defined was not used in the test. I just added a TODO entry for that.

I also removed unused variables.


integration-tests/cypress/cypress.spec.js
```diff
@@ -222,46 +220,51 @@ moduleTypes.forEach(({
...
+            name: 'cypress.test',
+            resource: 'cypress/e2e/basic-pass.js.basic pass suite can pass',
+            type: 'test',
+            meta: {
+              TEST_STATUS: 'pass',
+              TEST_NAME: 'basic pass suite can pass',
+              TEST_SUITE: 'cypress/e2e/basic-pass.js',
+              TEST_FRAMEWORK: 'cypress',
+              TEST_TYPE: 'browser',
+              TEST_CODE_OWNERS: JSON.stringify(['@datadog-dd-trace-js']),
```
juan-fernandez: is this change correct?
```suggestion
              [TEST_STATUS]: 'pass',
              [TEST_NAME]: 'basic pass suite can pass',
              [TEST_SUITE]: 'cypress/e2e/basic-pass.js',
              [TEST_FRAMEWORK]: 'cypress',
              [TEST_TYPE]: 'browser',
              [TEST_CODE_OWNERS]: JSON.stringify(['@datadog-dd-trace-js']),
```
if it isn't, you need to import https://github.com/DataDog/dd-trace-js/pull/7297/changes#diff-68748ac051251a008cc2ba15e7824343438d0be3f665d7463f031c07c7fb6baeL25-L27 back

integration-tests/cypress/cypress.spec.js
```diff
@@ -222,46 +220,51 @@ moduleTypes.forEach(({
...
-          assert.strictEqual(failedTestSpan.meta[TEST_FRAMEWORK], 'cypress')
-          assert.strictEqual(failedTestSpan.meta[TEST_TYPE], 'browser')
-          assert.ok(failedTestSpan.meta[TEST_SOURCE_FILE])
+
+          assertObjectContains(failedTestSpan, {
+            name: 'cypress.test',
+            resource: 'cypress/e2e/basic-fail.js.basic fail suite can fail',
+            type: 'test',
+            meta: {
+              TEST_STATUS: 'fail',
```
juan-fernandez: same as above

integration-tests/cypress/cypress.spec.js
juan-fernandez: this would explain tests failing in https://github.com/DataDog/dd-trace-js/actions/runs/21214065817/job/61030133890?pr=7297

BridgeAR: Ah, yes

BridgeAR: Fixed

integration-tests/cypress/cypress.spec.js
```diff
@@ -222,46 +222,51 @@ moduleTypes.forEach(({
...
+              customTag: 'customValue',
+              addTagsBeforeEach: 'customBeforeEach',
+              addTagsAfterEach: 'customAfterEach',
+              addTagsAfterFailure: 'customAfterFailure'
+            },
+          })
           assert.match(failedTestSpan.meta[TEST_SOURCE_FILE], /cypress\/e2e\/basic-fail\.js/)
           assert.ok(failedTestSpan.meta[TEST_FRAMEWORK_VERSION])
           assert.ok(failedTestSpan.meta[COMPONENT])
           assert.ok(failedTestSpan.meta[ERROR_MESSAGE])
```
juan-fernandez, line 265: any reason not to additionally include these `meta` tags in the `assertObjectContains` above?

integration-tests/cypress/cypress.spec.js
```diff
@@ -222,46 +222,51 @@ moduleTypes.forEach(({
...
+              [TEST_STATUS]: 'fail',
+              [TEST_NAME]: 'basic fail suite can fail',
+              [TEST_SUITE]: 'cypress/e2e/basic-fail.js',
+              [TEST_FRAMEWORK]: 'cypress',
+              [TEST_TYPE]: 'browser',
+              [TEST_CODE_OWNERS]: JSON.stringify(['@datadog-dd-trace-js']),
+              customTag: 'customValue',
+              addTagsBeforeEach: 'customBeforeEach',
+              addTagsAfterEach: 'customAfterEach',
+              addTagsAfterFailure: 'customAfterFailure'
```
juan-fernandez: is this correct? Don't we check `addTagsAfterFailure` is not present in `meta` ? https://github.com/DataDog/dd-trace-js/pull/7297/changes#diff-68748ac051251a008cc2ba15e7824343438d0be3f665d7463f031c07c7fb6baeR271

integration-tests/vitest/vitest.spec.js
```diff
@@ -2118,6 +2113,10 @@ versions.forEach((version) => {
               assert.strictEqual(retriedTestNew, isNew ? NUM_RETRIES_EFD : 0)
               assert.strictEqual(retriedTestsWithReason, NUM_RETRIES_EFD)
             }
+
+            if (isParallel) {
```
juan-fernandez: `vitest` is always parallel, so this is not needed. I realise there was a `isParallel` in the test before, but it wasn't needed

integration-tests/cypress/cypress.spec.js
BridgeAR: I do not know the exact error message. We could actually use the new ANY_STRING helper for that but I forgot when working on it

juan-fernandez: we can exclude `ERROR_MESSAGE` if need be

integration-tests/cypress/cypress.spec.js
juan-fernandez: that might explain this test error: https://github.com/DataDog/dd-trace-js/actions/runs/21215049696/job/61033692769?pr=7297

integration-tests/cypress/cypress.spec.js
BridgeAR: Shall we keep that for another PR?

integration-tests/cypress/cypress.spec.js
```diff
@@ -153,14 +153,14 @@ moduleTypes.forEach(({
 
       // Create a fresh web server for each test to avoid state issues
       webAppServer = createWebAppServer()
-      await new Promise((resolve, reject) => {
+      await /** @type {Promise<void>} */ (new Promise((resolve, reject) => {
         webAppServer.once('error', reject)
         webAppServer.listen(0, 'localhost', () => {
           webAppPort = webAppServer.address().port
           webAppServer.removeListener('error', reject)
           resolve()
```
juan-fernandez, line 161: shouldn't we do the same as in https://github.com/DataDog/dd-trace-js/pull/7297/changes#diff-95017dabcd69d1ae16bc190da206260b5da7c6c81d033504d6c89413e8373d80R37-R41?

integration-tests/cypress/cypress.spec.js
juan-fernandez: are those changes out of scope of this PR?

integration-tests/mocha/mocha.spec.js
```diff
@@ -3613,6 +3550,18 @@ describe(`mocha@${MOCHA_VERSION}`, function () {
...
+      /**
+       * @param {() => void} done
+       * @param {{
+       *   isAttemptToFix?: boolean,
+       *   isQuarantined?: boolean,
+       *   isDisabled?: boolean,
+       *   shouldAlwaysPass?: boolean,
+       *   shouldFailSometimes?: boolean,
+       *   extraEnvVars?: Record<string, string>,
+       *   isParallel?: boolean
```
juan-fernandez:
```suggestion
```

integration-tests/vitest/vitest.spec.js
```diff
@@ -2125,7 +2119,7 @@ versions.forEach((version) => {
         { isModified, isEfd = false, isParallel = false, isNew = false },
```
juan-fernandez:
```suggestion
        { isModified, isEfd = false, isNew = false },
```

integration-tests/cypress/cypress.spec.js
BridgeAR: Running these tests takes very long. Even, if I skip many. So yes, that was just a drive-by example which I added.

Using the `ANY_STRING` matchers is also not ideal, since that should likely also check for being non empty.

I am happy to open a follow-up PR in some time to change more places to use our helper.

So yes, this is out of scope for me.

integration-tests/cypress/cypress.spec.js
BridgeAR: I addressed issues that were currently flagged as type issues. The webAppServer is currently still `any` and therefore follow-up types are not detected as being wrong. That will be flagged when we prohibit implicit any.

But for now, I would concentrate on the reported, as it would otherwise we too much at once. This is an incremental work anyway.

integration-tests/vitest/vitest.spec.js
BridgeAR: Addressed and I activated the eslint rule to detect these as mistake

integration-tests/cypress/cypress.spec.js
juan-fernandez: that makes sense. Should we revert the `assertObjectContains` changes then and do them at the follow-up PR?

integration-tests/cypress/cypress.spec.js
juan-fernandez: I see. Just for me to understand: why are these lines not flagged as type issues?

juan-fernandez (commented):
CI seems to be failing now: https://github.com/DataDog/dd-trace-js/actions/runs/21246776002/job/61137915349?pr=7297. Is it related to the latest changes?

integration-tests/cypress/cypress.spec.js
BridgeAR: Because `any` means anything is allowed

PR #7298, opened 2026-01-21
fix: endpoint telemetry use setTimeout

simon-id:
### What does this PR do?
Change scheduling type for endpoint detection telemetry payloads.

### Motivation
System test was flaky, and this fixes it :shrug:


simon-id:
@BridgeAR it is confusing to me too, but by observation, when reproducing locally it fails almost all the time with setImmediate(), and it never fails with setTimeout(). I suspect a race condition in the system test, not with us. Either way this PR reduces flakiness stress on the system test CI (and on ours too), while not being a great piece of programming for sure.

PR #7301, opened 2026-01-21
split azure functions integration tests in ci

rochdev:
### What does this PR do?

Split Azure Functions integration tests in CI.

### Motivation

The tests are extremely slow and combined can reach 15+ minutes. With this change the time should be cut in half.

### DIFF
--- a/packages/datadog-plugin-azure-functions/test/integration-test/http-test/client.spec.js
+++ b/packages/datadog-plugin-azure-functions/test/integration-test/http-test/client.spec.js
@@ -74,8 +74,7 @@ describe('esm', () => {
 async function spawnPluginIntegrationTestProc (cwd, command, args, agentPort, stdioHandler, additionalEnvArgs = {}) {
   let env = {
     NODE_OPTIONS: `--loader=${hookFile} func start`,
-    DD_TRACE_AGENT_PORT: agentPort,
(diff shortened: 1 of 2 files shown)


packages/datadog-plugin-azure-functions/test/integration-test/http-test/client.spec.js
```diff
@@ -74,7 +74,7 @@ describe('esm', () => {
 async function spawnPluginIntegrationTestProc (cwd, command, args, agentPort, stdioHandler, additionalEnvArgs = {}) {
   let env = {
     NODE_OPTIONS: `--loader=${hookFile} func start`,
-    DD_TRACE_AGENT_PORT: agentPort,
+    DD_TRACE_AGENT_PORT: agentPort
```
BridgeAR, line 77: Nit: I think it is actually good to keep trailing ports for reducing churn when adding or removing entries.

BridgeAR: I guess this is something to be changed in the linter at some point.

rochdev: I definitely agree, but I think right now it's the opposite and there are cases where the linter actively _removes_ the comma. I would be onboard with reverting that to ensuring that it's always there instead 👍

PR #7306, opened 2026-01-22
ci: fix vendored dependabot PRs

BridgeAR:
The main issue is AFAIK the Node.js version being used. This is now aligned with the installment of the bundle job.

That is done in a secure way where we use least privileges possible.

### DIFF
--- a/.github/workflows/dependabot-automation.yml
+++ b/.github/workflows/dependabot-automation.yml
@@ -48,7 +48,6 @@ jobs:
     # so it is intentionally restricted to read-only permissions and produces a
     # patch artifact instead of pushing directly.
     permissions:
-      actions: write # upload-artifact uses the Actions API
       contents: read
       pull-requests: read
     outputs:
@@ -143,7 +142,7 @@ jobs:
     # Security: this job has write permissions but never runs installs/builds.
     # It only applies the vetted patch artifact and pushes a single commit.
     permissions:
-      actions: read # download-artifact uses the Actions API
+      contents: write
       id-token: write
     steps:
       - uses: DataDog/dd-octo-sts-action@acaa02eee7e3bb0839e4272dacb37b8f3b58ba80 # v1.0.3


rochdev (dismissed):
LGTM, but would like another review from @KSerrania as well on the security side.

.github/workflows/dependabot-automation.yml
```diff
@@ -46,6 +48,7 @@ jobs:
     # so it is intentionally restricted to read-only permissions and produces a
     # patch artifact instead of pushing directly.
     permissions:
+      actions: write # upload-artifact uses the Actions API
```
KSerrania: I don't think that should be needed. You should only need `actions: write` if you want to manipulate actions, their artifacts and caches across the entire repository. See this example usage of `actions/upload-artifact`: https://github.com/DataDog/datadog-agent/blob/main/.github/workflows/docs-dev.yml#L14-L50.

Looking at errors in recent runs like https://github.com/DataDog/dd-trace-js/actions/runs/21231777060/job/61091562886:

```
Error: An action could not be found at the URI 'https://api.github.com/repos/actions/upload-artifact/tarball/ea165f8a6f3f9f5b76b7b70ee0fdd3b5d7d3b6a2' (DC21:30ED6D:81B30:A86C0:697175EE)
```

I believe the problem comes from the definition of the action. You define:
```
uses: actions/upload-artifact@ea165f8a6f3f9f5b76b7b70ee0fdd3b5d7d3b6a2 # v4.6.2
```

But commit `ea165f8a6f3f9f5b76b7b70ee0fdd3b5d7d3b6a2` does not exist in the `actions/upload-artifact` repository: https://github.com/actions/upload-artifact/commit/ea165f8a6f3f9f5b76b7b70ee0fdd3b5d7d3b6a2.

The `v4.6.2` tag of `actions/upload-artifact` is on commit `ea165f8d65b6e75b540449e92b4886f43607fa02`: https://github.com/actions/upload-artifact/commit/ea165f8d65b6e75b540449e92b4886f43607fa02.

.github/workflows/dependabot-automation.yml
```diff
@@ -133,10 +143,17 @@ jobs:
     # Security: this job has write permissions but never runs installs/builds.
     # It only applies the vetted patch artifact and pushes a single commit.
     permissions:
-      contents: write
+      actions: read # download-artifact uses the Actions API
```
KSerrania: Likewise, `actions/download-artifact` should not need permissions to read artifacts from its own workflow. Example: https://github.com/DataDog/integrations-core/blob/4dc075d2d6529e120442d7c555fe8c75a2149c67/.github/workflows/pr-all.yml#L39-L66

.github/workflows/dependabot-automation.yml
BridgeAR: The version is definitely broken. I just fixed that and updated it to the current latest.

I do believe the permission is required though: in the example, write permission is granted. Since the permissions block is defined here, we have to grant permissions that are needed, as far as I know.

.github/workflows/dependabot-automation.yml
```diff
@@ -134,13 +143,20 @@ jobs:
     # It only applies the vetted patch artifact and pushes a single commit.
     permissions:
       contents: write
```
KSerrania, line 145: Minor: I see that you switched to using a `dd-octo-sts` token to push the commit now. Therefore, I don't think you need this `contents: write` permission anymore.

BridgeAR: I believe it would just be a fallback

PR #7307, opened 2026-01-22
fix: do not log pg deprecation message

BridgeAR:
This somehow slipped through, while I do not know why.

This is essential for master to be green.

### DIFF
--- a/packages/datadog-instrumentations/src/pg.js
+++ b/packages/datadog-instrumentations/src/pg.js
@@ -112,8 +112,10 @@ function wrapQuery (query) {
       arguments[0] = pgQuery
 
       const retval = query.apply(this, arguments)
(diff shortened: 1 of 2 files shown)


packages/datadog-instrumentations/src/pg.js
```diff
@@ -112,8 +112,10 @@ function wrapQuery (query) {
       arguments[0] = pgQuery
 
       const retval = query.apply(this, arguments)
-      const queryQueue = this._queryQueue || this.queryQueue
-      const activeQuery = Object.hasOwn(this, '_activeQuery') ? this._activeQuery : this.activeQuery
+
+      const deperecated = Object.hasOwn(this, '_activeQuery')
```
rochdev, line 116: Not a blocker since CI is red, but this should be `deprecated`.

PR #7308, opened 2026-01-22
feat(debugger): add support for v2 input endpoint detection

watson:
### What does this PR do?

Implements support for the new `/debugger/v2/input` endpoint with automatic agent capability detection and graceful fallback to `/debugger/v1/diagnostics`.

This change enables the Live Debugger/Dynamic Instrumentation client to:
- Automatically detect which debugger input endpoint the agent supports by querying the `/info` endpoint at startup
- Use `/debugger/v2/input` when available for improved functionality
- Gracefully fallback to `/debugger/v1/diagnostics` if the v2 endpoint is not available or returns 404
- Continue using the diagnostics endpoint after fallback for all subsequent requests

The diagnostics endpoint is used as the fallback (rather than v1/input) because it supports the same payload format for snapshots while also being universally available on all agent versions.

### Motivation

This change is part of the "Live Debugger Default-On RFC".

### Additional notes

Tip: Ignore whitespace when reviewing the changes to `packages/dd-trace/src/debugger/index.js`

### DIFF
--- a/integration-tests/debugger/endpoint-fallback.spec.js
+++ b/integration-tests/debugger/endpoint-fallback.spec.js
@@ -29,6 +29,7 @@ describe('Dynamic Instrumentation - Endpoint Fallback', function () {
       assert.deepStrictEqual(response.data, { hello: 'bar' })
 
       const [{ payload }] = await diagnosticsInput
+
       assertObjectContains(payload[0], {
         ddsource: 'dd_debugger',
         service: 'node',
@@ -90,6 +91,7 @@ describe('Dynamic Instrumentation - Endpoint Fallback', function () {
       assert.deepStrictEqual(response.data, { hello: 'bar' })
 
       const [{ payload }] = await v2Input
+
       assertObjectContains(payload[0], {
         ddsource: 'dd_debugger',
         service: 'node',
@@ -120,6 +122,7 @@ describe('Dynamic Instrumentation - Endpoint Fallback', function () {
       assert.deepStrictEqual(response.data, { hello: 'bar' })
 
       const [, [{ payload }]] = await Promise.all([v2404Event, diagnosticsInput])
+
       assertObjectContains(payload[0], {
         ddsource: 'dd_debugger',
         service: 'node',


watson:
* **#7308** <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-js/7308?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-js/7308?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/dd-trace/src/debugger/index.js
```diff
@@ -181,6 +201,44 @@ function cleanup (error) {
...
+
+  agentInfoExporter.getAgentInfo((err, agentInfo) => {
+    if (err) {
+      log.warn('[debugger] Failed to query agent /info endpoint, falling back to %s', DEBUGGER_INPUT_V1, err)
+      return cb(DEBUGGER_INPUT_V1)
+    }
+
+    const endpoints = agentInfo.endpoints || []
+
+    if (endpoints.includes(DEBUGGER_INPUT_V2)) {
```
shatzi, line 224: maybe fallback to diagnostics if we have it.

shatzi (dismissed):

packages/dd-trace/src/debugger/index.js
watson: I've added this topic to the client lib weekly sync meeting

watson: Fixed 👍

packages/dd-trace/src/debugger/index.js
```diff
@@ -38,100 +50,106 @@ function isStarted () {
...
   if (worker !== null) return
 
   log.debug('[debugger] Starting Dynamic Instrumentation client...')
 
-  rc = rcInstance
-  rcAckCallbacks = new Map()
-  const probeChannel = new MessageChannel()
-  const logChannel = new MessageChannel()
-  configChannel = new MessageChannel()
+  detectDebuggerEndpoint(config, (_inputPath) => {
```
shatzi, line 99: There could be a race condition here. Can a probe be invoked before detectDebuggerEndpoint finishes?

packages/dd-trace/src/debugger/index.js
```diff
@@ -181,6 +200,43 @@ function cleanup (error) {
...
+function detectDebuggerEndpoint (config, cb) {
+  log.debug('[debugger] Detecting available debugger endpoints...')
+
+  fetchAgentInfo(getAgentUrl(config), (err, agentInfo) => {
+    if (err) {
+      log.warn('[debugger] Failed to query agent %s endpoint, falling back to %s',
+        DEBUGGER_INPUT_V2,
+        DEBUGGER_DIAGNOSTICS_V1,
+        err)
+      return cb(DEBUGGER_DIAGNOSTICS_V1)
```
shatzi, line 219: I am not sure I like this approach. If we fail to get agent-info, do we default to diagnostics, even if agent-info does not exist? Also, would the tracer retry at some point?

packages/dd-trace/src/debugger/index.js
```diff
@@ -181,6 +200,43 @@ function cleanup (error) {
...
+    }
+
+    const endpoints = agentInfo.endpoints || []
+
+    if (endpoints.includes(DEBUGGER_INPUT_V2)) {
+      log.debug('[debugger] Agent supports %s', DEBUGGER_INPUT_V2)
+      return cb(DEBUGGER_INPUT_V2)
+    }
+    log.debug('[debugger] Agent does not support %s, using %s', DEBUGGER_INPUT_V2, DEBUGGER_DIAGNOSTICS_V1)
```
shatzi, line 228: where did we check DEBUGGER_DIAGNOSTICS_V1 is available?

packages/dd-trace/src/debugger/devtools_client/send.js
```diff
@@ -78,17 +80,63 @@ function send (message, logger, dd, snapshot) {
...
+function handleV2FallbackIfNeeded (statusCode, payload) {
+  if (statusCode !== 404 || config.inputPath !== DEBUGGER_INPUT_V2) {
+    return false
+  }
+
+  log.warn('[debugger:devtools_client] Received 404 from %s, falling back to %s',
+    DEBUGGER_INPUT_V2,
+    DEBUGGER_DIAGNOSTICS_V1)
+
+  setInputPath(DEBUGGER_DIAGNOSTICS_V1)
```
shatzi, line 110: no fallback to v1?

packages/dd-trace/src/debugger/index.js
watson: No it can't. No probes will be added until the worker thread is started which only happens later inside this function

packages/dd-trace/src/debugger/index.js
watson: Correct. I did it this way because it is essentially the same behavior as previously, where we never checked `/info` and just assumed the endpoints would exist. I can change the behavior to simply fail to start the debugger in that case, but I felt that if this was an intermitten failure, it was simpler to fall back to the diagnostics endpoint, than trying to be clever about re-trying etc.

> Also, would the tracer retry at some point?

Yes and no: The `/info` endpoint is only queried when the debugger is starting. This can happen at boot, or via remote enablement. So if you disable and then enable remotely, it will happen agian.

packages/dd-trace/src/debugger/index.js
watson: We don't. Similar to the current code in `master` - we assume that an agent supporting this endpoint is running. This has not been an issue since the beginning.

packages/dd-trace/src/debugger/devtools_client/send.js
watson: Correct. This was the consensus that was reached as far as I understand it: We do not want to accidentally start sending data to `/debugger/v1/input` that will then not be redacted.

integration-tests/helpers/fake-agent.js
```diff
@@ -313,9 +322,11 @@ function buildExpressServer (agent) {
   app.use(bodyParser.json({ limit: Infinity, type: 'application/json' }))
 
   app.get('/info', (req, res) => {
-    res.json({
-      endpoints: ['/evp_proxy/v2'],
-    })
+    const endpoints = ['/evp_proxy/v2', '/debugger/v1/input']
```
BridgeAR, line 325: @juan-fernandez I remember you also using the evp_proxy endpoint and maybe using this would be an alternative for faster tests? :)

integration-tests/helpers/fake-agent.js
```diff
@@ -408,22 +419,57 @@ function buildExpressServer (agent) {
...
+    })
+    agent.emit('debugger-input-v2', {
+      headers: req.headers,
+      query: req.query,
+      payload: req.body,
     })
   })
 
+  app.post('/debugger/v1/diagnostics', upload.any(), (req, res) => {
+    res.status(200).send() // TODO: Should we send a 202 here instead?
```
BridgeAR, line 454: Is that not something you would be best to answer? 😅

packages/dd-trace/src/debugger/devtools_client/send.js
```diff
@@ -78,17 +80,63 @@ function send (message, logger, dd, snapshot) {
...
 function onFlush (payload) {
   log.debug('[debugger:devtools_client] Flushing probe payload buffer')
 
-  const opts = {
+  request(payload, buildRequestOpts(), (err, res, statusCode) => {
+    if (handleV2FallbackIfNeeded(statusCode, payload)) {
+      return
+    }
+
+    if (err) {
```
BridgeAR: Nit

```suggestion
    if (!handleV2FallbackIfNeeded(statusCode, payload) && err) {
```

packages/dd-trace/src/debugger/devtools_client/send.js
```diff
@@ -78,17 +80,63 @@ function send (message, logger, dd, snapshot) {
...
+    return false
+  }
+
+  log.warn('[debugger:devtools_client] Received 404 from %s, falling back to %s',
+    DEBUGGER_INPUT_V2,
+    DEBUGGER_DIAGNOSTICS_V1)
+
+  setInputPath(DEBUGGER_DIAGNOSTICS_V1)
+
+  request(payload, buildRequestOpts(), (err) => {
```
BridgeAR, line 112: The options still seem identical and should likely be passed through instead of having to be defined again.

integration-tests/debugger/endpoint-fallback.spec.js
```diff
@@ -0,0 +1,180 @@
...
+    it('should use diagnostics endpoint when agent does not advertise v2 support', async function () {
+      const diagnosticsPromise = new Promise(/** @type {() => void} */ (resolve) => {
+        t.agent.once('debugger-diagnostics-input', ({ payload }) => {
+          assertObjectContains(payload[0], {
+            ddsource: 'dd_debugger',
+            service: 'node',
+            debugger: { snapshot: {} },
+          })
+          resolve()
+        })
```
BridgeAR: Nit:


```suggestion
      const diagnosticsPromise = once(t.agent, 'debugger-diagnostics-input')
 
// ... later ...

      assertObjectContains(payload[0], {
        ddsource: 'dd_debugger',
        service: 'node',
        debugger: { snapshot: {} },
      })
```

Similar in the other tests here.

integration-tests/debugger/endpoint-fallback.spec.js
```diff
@@ -0,0 +1,180 @@
...
+      })
+
+      t.agent.on('debugger-input-v2', () => {
```
BridgeAR:
```suggestion
      t.agent.once('debugger-input-v2', () => {
```

It could otherwise fail multiple times. I believe it should also theoretically be removed. Otherwise it could theoretically fail after the test passed.

integration-tests/helpers/fake-agent.js
watson: Yes, but I left it there for me to refactor later as it wasn't important right now

packages/dd-trace/src/debugger/devtools_client/send.js
watson: Do you mean the requests options? I specifically chose to rebuild them as `request` is mutating the passed in object. It might work regardless, but the behavior would be undefined and brittle.

integration-tests/debugger/endpoint-fallback.spec.js
watson: Fixed 👍

integration-tests/debugger/endpoint-fallback.spec.js
watson: Fixed 👍

packages/dd-trace/src/debugger/devtools_client/send.js
watson: Fixed 👍

integration-tests/helpers/fake-agent.js
watson: I verified it for the `/debugger/v1/input` event, but didn't get to check the actual status code for `/debugger/v1/diagnostics`, so wanted to do that later

packages/dd-trace/src/debugger/devtools_client/send.js
BridgeAR: Ok, should we not instead just prevent that mutation?

integration-tests/debugger/endpoint-fallback.spec.js
```diff
@@ -0,0 +1,179 @@
...
+        advertiseDebuggerV2IntakeSupport: false,
+      },
+    })
+
+    it('should use diagnostics endpoint when agent does not advertise v2 support', async function () {
+      const diagnosticsPromise = new Promise((resolve) => {
+        t.agent.once('debugger-diagnostics-input', ({ payload }) => {
+          resolve(payload[0])
+        })
+      })
```
BridgeAR: Any reason not to use the Node.js native helper that I mentioned?

integration-tests/debugger/endpoint-fallback.spec.js
```diff
@@ -0,0 +1,179 @@
...
+        service: 'node',
+        debugger: { snapshot: {} },
+      })
+    })
+
+    it('should continue using diagnostics endpoint for multiple requests', async function () {
+      const expectedSnapshots = 2
+      let snapshotsReceived = 0
+
+      const diagnosticsPromise = new Promise(/** @type {() => void} */ (resolve) => {
```
BridgeAR: Same as above. I guess the comments might have been resolved with an LLM and it did not understand things well enough?

integration-tests/debugger/endpoint-fallback.spec.js
```diff
@@ -0,0 +1,179 @@
...
+              debugger: { snapshot: {} },
+            })
+          })
+          if (snapshotsReceived >= expectedSnapshots) {
+            resolve()
+          }
+        })
+      })
+
+      t.agent.on('debugger-input-v2', () => {
```
BridgeAR: As above


```suggestion
      t.agent.once('debugger-input-v2', () => {
```

packages/dd-trace/src/debugger/devtools_client/send.js
watson: Maybe, but there's drawbacks that needs to be considered, so I think it's out of scope. This fallback event hopefully never happens, but if it does, it happens once in the life-time of the process. So the extra object created is not a big deal here - vs the overhead of always cloning the object on all calls to `request` if we "fixed" it at the source.

integration-tests/debugger/endpoint-fallback.spec.js
```diff
@@ -0,0 +1,179 @@
...
+
+      await diagnosticsPromise
+    })
+  })
+
+  describe('v2 endpoint works when agent supports it', function () {
+    const t = setup({ dependencies: ['fastify'], testApp: 'target-app/basic.js' })
+
+    it('should successfully use v2 endpoint when agent supports it', async function () {
+      const v2InputPromise = new Promise(/** @type {() => void} */ (resolve) => {
```
BridgeAR: As above

integration-tests/debugger/endpoint-fallback.spec.js
```diff
@@ -0,0 +1,179 @@
...
+          assertObjectContains(payload[0], {
+            ddsource: 'dd_debugger',
+            service: 'node',
+            debugger: { snapshot: {} },
+          })
+          resolve()
+        })
+      })
+
+      t.agent.on('debugger-diagnostics-input', () => {
```
BridgeAR:
```suggestion
      t.agent.once('debugger-diagnostics-input', () => {
```

integration-tests/debugger/endpoint-fallback.spec.js
```diff
@@ -0,0 +1,179 @@
...
+        debuggerV2IntakeStatusCode: 404,
+      },
+    })
+
+    it('should fallback to diagnostics endpoint when v2 returns 404 at runtime', async function () {
+      const v2404Promise = new Promise(/** @type {() => void} */ (resolve) => {
+        t.agent.once('debugger-input-v2-404', () => resolve())
+      })
+
+      const diagnosticsPromise = new Promise(/** @type {() => void} */ (resolve) => {
```
BridgeAR: As above

integration-tests/debugger/endpoint-fallback.spec.js
```diff
@@ -0,0 +1,179 @@
...
+      agentOptions: {
+        // Agent advertises v2 support in /info, but returns 404 when actually called
+        // This simulates an edge case where agent changes between /info and actual request
+        advertiseDebuggerV2IntakeSupport: true,
+        debuggerV2IntakeStatusCode: 404,
+      },
+    })
+
+    it('should fallback to diagnostics endpoint when v2 returns 404 at runtime', async function () {
+      const v2404Promise = new Promise(/** @type {() => void} */ (resolve) => {
```
BridgeAR: As above

integration-tests/debugger/endpoint-fallback.spec.js
```diff
@@ -0,0 +1,179 @@
...
+      assert.deepStrictEqual(response.data, { hello: 'bar' })
+
+      await Promise.all([v2404Promise, diagnosticsPromise])
+    })
+
+    it('should continue using diagnostics endpoint after runtime fallback', async function () {
+      const expectedSnapshots = 2
+      let snapshotsReceived = 0
+
+      const v2404Promise = new Promise(/** @type {() => void} */ (resolve) => {
```
BridgeAR: As above

integration-tests/debugger/endpoint-fallback.spec.js
```diff
@@ -0,0 +1,179 @@
...
+        t.agent.once('debugger-input-v2-404', () => resolve())
+      })
+
+      const diagnosticsPromise = new Promise(/** @type {() => void} */ (resolve) => {
```
BridgeAR: As above

packages/dd-trace/src/debugger/devtools_client/send.js
BridgeAR: I am ok with that since it is not new with the PR, I am just always concerned about mutating objects like that.

integration-tests/debugger/endpoint-fallback.spec.js
watson: Fixed 👍

integration-tests/debugger/endpoint-fallback.spec.js
watson: Fixed 👍

integration-tests/debugger/endpoint-fallback.spec.js
watson: Fixed 👍

integration-tests/debugger/endpoint-fallback.spec.js
watson: Fixed 👍

integration-tests/debugger/endpoint-fallback.spec.js
watson: Fixed 👍

integration-tests/debugger/endpoint-fallback.spec.js
watson: Fixed 👍

integration-tests/debugger/endpoint-fallback.spec.js
watson: Fixed 👍

integration-tests/debugger/endpoint-fallback.spec.js
watson: Fixed 👍

integration-tests/debugger/endpoint-fallback.spec.js
watson: Fixed 👍

PR #7309, opened 2026-01-22
chore(ts): fix remaining TS errors in fake-agent.js

watson:
### What does this PR do?

As the title says.

### Motivation

Reduce the risk of coding errors.

### DIFF
--- a/integration-tests/helpers/fake-agent.js
+++ b/integration-tests/helpers/fake-agent.js
@@ -10,13 +10,37 @@ const upload = require('multer')()
 
 const noop = () => {}
 
+/**
+ * @typedef {object} RemoteConfigFile
+ * @property {number} orgId
+ * @property {string} product
+ * @property {string} id
+ * @property {string} name
+ * @property {string} config
+ * @property {string} path
+ * @property {string} fileHash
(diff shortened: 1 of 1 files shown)


integration-tests/helpers/fake-agent.js
```diff
@@ -10,13 +10,37 @@ const upload = require('multer')()
...
+
 module.exports = class FakeAgent extends EventEmitter {
+  port = 0
+  /** @type {Set<import('net').Socket>} */
+  #sockets = new Set()
+  /** @type {Record<string, RemoteConfigFile>} */
+  _rcFiles = {}
+  _rcTargetsVersion = 0
+  /** @type {Set<string>} */
+  _rcSeenStates = new Set()
```
BridgeAR, line 38: These here could use a small helper method or a getter to prevent the need for the underscored properties.
I believe they should just not be `public` to reflect the actual agent closer?

This could be seen as out of scope, so I approve anyway.

watson: I considered it, but felt it was out of scope as it would make the diff more complex. Hence why I only changed `_sockets`

PR #7310, opened 2026-01-22
refactor(exporters): split AgentInfoExporter into focused modules

watson:
## What does this PR do?

This PR refactors the `AgentInfoExporter` class, which had mixed responsibilities and confusing naming, by splitting it into two focused modules with clearer purposes.

## Motivation

The `AgentInfoExporter` class was problematic for several reasons:

1. **Misleading name**: Despite being called an "exporter", it was primarily used as a base class for buffering traces
2. **Mixed responsibilities**: Combined agent info fetching with trace buffering/exporting logic
3. **Unnecessary instantiations**: Some modules only needed to fetch agent info but had to instantiate the entire exporter class


watson:
* **#7310** <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-js/7310?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-js/7310?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/dd-trace/src/agent/url.js
```diff
@@ -0,0 +1,25 @@
...
+// additional logic for unix socket auto-detection on Linux. The config version is only used
+// during config initialization, while this one is used throughout the codebase. Consider if
+// the unix socket detection should be part of this general helper or remain config-specific.
+
+/**
+ * Gets the agent URL from config, constructing it from hostname/port if needed
+ * @param {ReturnType<import('../config')>} config - Tracer configuration object
+ * @returns {URL} The agent URL
+ */
+function getAgentUrl (config) {
```
BridgeAR, line 18: I strongly feel that this should be part of config. I also believe we should consolidate the overall logic, while I do not know of all implications (there are some nuanced deviations with the settings, while I believe unifying them would actually resolve customer confusion and simplify customer setups without breaking them). The hostname and the port have a different way of being defined and those are used here and dogstatd is having a separate endpoint as it seems (which I am uncertain about, if that is intentional or not). So it is something to determine first.

I think we could start by moving this into config (including the `fetchAgentInfo()` functionality which I would expand further as the methods that use this are mainly interested in endpoints being available or not on our agent. We can return a promise or a callback for that case and handle everything inside of config.

That way most code is consolidated, it is simpler, and we have less code duplication.

watson: I agree. This is a 1 out of X PRs refactoring this. To make the PR easier to review, I've split it up into multiple

PR #7319, opened 2026-01-22
Fix: Update Opentelemety Active Functionality

bojbrook:
### What does this PR do?
This PR refactors the context manager’s active() method to improve context handling between Datadog and OpenTelemetry and to fix edge cases when converting between Datadog and OpenTelemetry spans.

### Motivation
This issue was reported in `APMS-18328` and can be reproduced with the code below. The core problem is that retrieving and setting the active span works as expected when using the Datadog API, but fails silently when performed through the OpenTelemetry API.
```
function enrichSpan() {
    const ddSpan = ddTrace.scope().active();
    const otelSpan = trace.getActiveSpan();
    // 1. OTel Attempt: SILENT FAILURE
    otelSpan?.setAttribute('my.otel.attr', 'failed');

    // 2. Datadog Attempt: SUCCESS
    ddSpan?.addTags({ 'my.dd.attr': 'success' });
}

function main() {
    otelTracer.startActiveSpan('otel-span', async (span) => {
        enrichSpan();
        await sleep(1000);
        span.end();
    });

    console.log('Ending main function');
}

main();
```
**Expected result:** For this scenario we would expect both tags(attributes) to be present on the same span.

### Plugin Checklist

- [ ] Unit tests.
- [ ] Integration tests.
- [ ] Benchmarks.
- [ ] TypeScript [tests][2].
- [ ] API [documentation][3].
- [ ] CI [jobs/workflows][4].


packages/dd-trace/src/opentelemetry/context_manager.js
```diff
@@ -14,45 +13,64 @@ class ContextManager {
...
         const entries = {}
         for (const [key, value] of Object.entries(baggages)) {
           entries[key] = { value }
         }
         const otelBaggages = propagation.createBaggage(entries)
-        return propagation.setBaggage(context, otelBaggages)
+        return propagation.setBaggage(store, otelBaggages)
       }
-      return context
+      return store
```
rochdev, line 33: Isn't `active` supposed to always return a context? Or is a context also a plain object similar to a store so a store can be used in-place?

PR #7322, opened 2026-01-22
feat(llmobs): support custom telemetry metadata in Vercel AI SDK spans

isaacwasserman:
### What does this PR do?

Adds support for custom telemetry metadata from the Vercel AI SDK's `experimental_telemetry.metadata` to LLM Observability spans.

**Files changed:**
- `packages/dd-trace/src/llmobs/plugins/ai/util.js` - New `getTelemetryMetadata()` function
- `packages/dd-trace/src/llmobs/plugins/ai/index.js` - Integration into span tagging methods

### Motivation

The Vercel AI SDK supports custom metadata via `experimental_telemetry.metadata`, which allows applications to attach contextual information (e.g., user IDs, conversation IDs, session data) to LLM calls:

```typescript
await streamText({
  model,
  prompt: "...",
  experimental_telemetry: {
    isEnabled: true,
    functionId: "chat",
    metadata: {
      conversationId: "abc123",
      userId: "user456",
      organizationId: "org789"
    }
  }
})
```

Currently, this metadata is captured as span attributes (`ai.telemetry.metadata.*`) but not propagated to LLM Observability span metadata, making it unavailable for filtering, grouping, and analysis in Datadog.

### Changes

**New function in `util.js`:**
```javascript
function getTelemetryMetadata (tags) {
  const metadata = {}
  const prefix = 'ai.telemetry.metadata.'

  for (const tag of Object.keys(tags)) {
    if (!tag.startsWith(prefix)) continue

    const metadataKey = tag.slice(prefix.length)
    if (metadataKey) {
      metadata[metadataKey] = tags[tag]
    }
  }

  return Object.keys(metadata).length ? metadata : null
}
```

**Integration in `index.js`:**
Updated `setEmbeddingWorkflowTags()`, `setObjectGenerationTags()`, and `setTextGenerationTags()` to merge telemetry metadata with generation metadata:

```javascript
const generationMetadata = getGenerationMetadata(tags) || {}
const telemetryMetadata = getTelemetryMetadata(tags) || {}
const metadata = { ...generationMetadata, ...telemetryMetadata }
this._tagger.tagMetadata(span, Object.keys(metadata).length ? metadata : null)
```

### Benefits

- **Better observability**: Custom context (users, sessions, experiments) appears in span metadata
- **Enhanced filtering**: Filter traces by custom metadata in Datadog UI
- **Improved debugging**: Correlate LLM calls with business context
- **Analytics**: Group and analyze LLM usage by custom dimensions

### Plugin Checklist

- [x] Unit tests
- [x] Integration tests
- [x] Benchmarks
- [x] TypeScript definitions (if needed)
- [x] TypeScript tests (if needed)
- [x] API documentation
- [x] CI jobs/workflows

### Additional Notes

This follows the existing pattern for `getGenerationMetadata()` and `getModelMetadata()` - extracting structured data from span tags and merging it into the metadata object.

The implementation is backward compatible: if no custom telemetry metadata is present, behavior is unchanged.

### DIFF
--- a/packages/dd-trace/src/llmobs/plugins/ai/util.js
+++ b/packages/dd-trace/src/llmobs/plugins/ai/util.js
@@ -223,7 +223,7 @@ function getTelemetryMetadata (tags) {
     }
   }
 
-  return Object.keys(metadata).length ? metadata : null
+  return metadata
 }
 
 module.exports = {


packages/dd-trace/src/llmobs/plugins/ai/util.js
```diff
@@ -205,6 +205,27 @@ function getLlmObsSpanName (operation, functionId) {
...
+  for (const tag of Object.keys(tags)) {
+    if (!tag.startsWith(prefix)) continue
+
+    const metadataKey = tag.slice(prefix.length)
+    if (metadataKey) {
+      metadata[metadataKey] = tags[tag]
+    }
+  }
+
+  return Object.keys(metadata).length ? metadata : null
```
sabrenner:
```suggestion
  return metadata
```

I checked and indeed the other metadata function does the length check as well, I don't remember why exactly, but I don't think it's needed, so this one doesn't need to (I'll change the other one in a different PR). 

If you agree, we can change the typedoc and default where this function is called.

packages/dd-trace/test/llmobs/plugins/ai/util.spec.js
sabrenner, line 1: While I like the test here, I'd prefer a more end-to-end type test in https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/test/llmobs/plugins/ai/index.spec.js. You should be able to just modify [this test](https://github.com/DataDog/dd-trace-js/blob/b4af38d2d7a6e7cdae4eb91ce79ed97b2d7324e0/packages/dd-trace/test/llmobs/plugins/ai/index.spec.js#L48-L102) to add `metadata` to the `options`, and then assert that the metadata shows up in the below assertions. I'm happy to help out if you have any questions!

If that works out, we can get rid of this file as we test it end-to-end instead.

PR #7327, opened 2026-01-23
fix rewriter support for super with traceCallback

rochdev:

### What does this PR do?

Fix super support for `traceCallback`.

### Motivation

We were previously using `function () {}` instead of an arrow function for the traced wrapper as it was needed for `traceCallback` which meant that `super` couldn't be used in the wrapped function. By switching to the expanded version with explicit `runStores` and `publish` calls, we can switch to an arrow function and solve the issue.

### DIFF
--- a/packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
+++ b/packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { parse } = require('./compiler')
+const { parse, query } = require('./compiler')
 
 const tracingChannelPredicate = (node) => (
   node.specifiers?.[0]?.local?.name === 'tr_ch_apm_tracingChannel' ||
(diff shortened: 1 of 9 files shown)


packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
```diff
@@ -106,25 +106,21 @@ function traceInstanceMethod (state, node, program) {
...
-  const tracedArgs = operator === 'traceCallback'
-    ? `__apm$original_args.splice(${index}, 1, arguments[${index >= 0 ? index : `arguments.length + ${index}`}])`
-    : '__apm$original_args'
-  const traceParams = operator === 'traceCallback'
-    ? `__apm$traced, ${index}`
-    : '__apm$traced'
   const wrapper = parse(`
     function wrapper () {
       const __apm$original_args = arguments;
       const __apm$traced = ${async} function () {
```
crysmags: @rochdev does this need to be changed to an arrow function as well for `tracePromise` and `traceSync` functions?

rochdev: No as this is unique to callbacks because of how the arguments have to be altered, which is not the case for promises or synchronous functions.

rochdev: Oh I see what you're saying, basically the function is no longer needed. Will take a look.

rochdev: Fixed, good catch!

crysmags (approved):

PR #7335, opened 2026-01-26
[test optimization][SDTEST-3004] Upload code coverage reports automatically 

juan-fernandez:
### What does this PR do?

If a code coverage report is present after running `nyc` with `mocha` or `cucumber` or after running with code coverage enabled with `jest` or `vitest` (they don't need `nyc`) and the library configuration includes `coverage_report_upload_enabled:true`, the library automatically uploads the code coverage report.

Since the `nyc` instrumentation runs in a different process (`nyc` uses `foreground-child` to spawn the test process it wraps), it does not have access to the library configuration, which is requested from test framework instrumentation. 

Instead of requesting the library configuration in `nyc` again, we'll set `DD_EXPERIMENTAL_TEST_OPT_SETTINGS_CACHE` for the test process to use. If the test instrumentation detects `DD_EXPERIMENTAL_TEST_OPT_SETTINGS_CACHE`, the libraries configuration request is stored in the file system. The `nyc` instrumentation will check the library configuration in the file system to decide whether to upload the report or not. 

Reference implementation in java: https://github.com/DataDog/dd-trace-java/pull/9425

### Motivation
Make it easier for our customers to use the code coverage product in Datadog.

### Additional Notes

* Cypress is not included because the way they recommend gathering code coverage includes running different processes: https://docs.cypress.io/app/tooling/code-coverage#See-code-coverage-summary 

* Same with playwright

In addition to the integration tests, this has been tested manually.

### DIFF
--- a/.cursor/commands/commit-message.md
+++ b/.cursor/commands/commit-message.md
@@ -1,0 +1,18 @@
+Write a commit message for the staged changes.
+
+Focus on the overall picture, a brief "what has changed", potentially followed by a "why" it was changed. Do not just blindly list every change. Only include specific changes if its relevant to understand the commit.
+
+The subject line should be a maximum of 50 chars.
+Each line of the body should be a maximum of 72 chars. Long URLs are allowed to exceed this.
+
+Format it as a semantic commit message
+- Format: `<type>(<scope>): <subject>` (`<scope>` is optional).
+
+Semantic commit message types:
+- feat: (new feature for the user, not a new feature for build script)
+- fix: (bug fix for the user, not a fix to a build script)
+- docs: (changes to the documentation)
+- style: (formatting, missing semi colons, etc; no production code change)
+- refactor: (refactoring production code, eg. renaming a variable)
+- test: (adding missing tests, refactoring tests; no production code change)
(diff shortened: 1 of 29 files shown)


packages/dd-trace/src/ci-visibility/test-optimization-cache.js
```diff
@@ -0,0 +1,61 @@
...
+function setupSettingsCachePath () {
+  const existing = getSettingsCachePath()
+  if (existing) {
+    return existing
+  }
+
+  const cacheFilePath = path.join(tmpdir(), `dd-test-optimization-${randomUUID()}.json`)
+
+  // eslint-disable-next-line eslint-rules/eslint-process-env
+  process.env.DD_EXPERIMENTAL_TEST_OPT_SETTINGS_CACHE = cacheFilePath
```
juan-fernandez, line 33: just for context, `nyc` already adds environment variables to the test command that it wraps: https://github.com/istanbuljs/nyc/blob/41f4476f9f1010367d62c9e7841e14cfe4a2801a/bin/nyc.js#L45. This is there so the test command knows where to write the cache

PR #7336, opened 2026-01-26
fix(debugger): apply source maps to probe stack traces

watson:
### What does this PR do?

Fixes probe stack traces to correctly show original source file locations when source maps are available. Previously, stack traces in probe snapshots displayed transpiled/generated code locations instead of the original source locations (e.g., TypeScript files), while breakpoint locations were correctly mapped.

### Motivation

When a Dynamic Instrumentation probe is hit in transpiled code, developers need to see stack traces that reference their original source files, not the generated JavaScript. This was already working correctly for breakpoint locations but not for the stack traces in the snapshots, making it difficult to debug transpiled applications.

### DIFF
--- a/packages/dd-trace/src/debugger/devtools_client/state.js
+++ b/packages/dd-trace/src/debugger/devtools_client/state.js
@@ -2,7 +2,7 @@
 
 const { join, dirname } = require('path')
 const { normalize } = require('../../../../../vendor/dist/source-map/lib/util')
-const { loadSourceMapSync } = require('./source-maps')
+const { loadSourceMapSync, getOriginalPosition } = require('./source-maps')
 const session = require('./session')
 const log = require('./log')
 
@@ -36,7 +36,6 @@ module.exports = {
 
   _loadedScripts: loadedScripts, // Only exposed for testing
   _scriptUrls: scriptUrls, // Only exposed for testing
-  _scriptSourceMaps: scriptSourceMaps, // Only exposed for testing
 
   /**
    * Find the script to inspect based on a partial or absolute path. Handles both Windows and POSIX paths.
@@ -127,8 +126,6 @@ module.exports = {
    * @returns {Promise<Array<StackFrame>>} - The stack from call frames.
    */
   getStackFromCallFrames (callFrames) {
-    const { getOriginalPosition } = require('./source-maps')
-
     return Promise.all(callFrames.map(async (frame) => {
       // TODO: Possible race condition: If the breakpoint is in the process of being removed, and this is the last
       // breakpoint, it will also stop the debugging session, which in turn will clear the state, which means clearing


packages/dd-trace/src/debugger/devtools_client/state.js
```diff
@@ -19,12 +36,14 @@ module.exports = {
 
   _loadedScripts: loadedScripts, // Only exposed for testing
   _scriptUrls: scriptUrls, // Only exposed for testing
+  _scriptSourceMaps: scriptSourceMaps, // Only exposed for testing
```
BridgeAR: Could we not use proxyquire instead?

packages/dd-trace/src/debugger/devtools_client/state.js
```diff
@@ -101,21 +120,57 @@ module.exports = {
...
 
+  /**
+   * Get the stack from call frames.
+   *
+   * @param {Array<import('inspector').Debugger.CallFrame>} callFrames - The call frames to get the stack from.
+   * @returns {Promise<Array<StackFrame>>} - The stack from call frames.
+   */
   getStackFromCallFrames (callFrames) {
-    return callFrames.map((frame) => {
+    const { getOriginalPosition } = require('./source-maps')
```
BridgeAR: This should be moved to after knowing if we have a source map or not, since the loading can be prevented if not.

packages/dd-trace/src/debugger/devtools_client/state.js
```diff
@@ -101,21 +120,57 @@ module.exports = {
...
+
+    return Promise.all(callFrames.map(async (frame) => {
```
BridgeAR, line 129: Likely not an issue, while we could optimize this for the case of not having source maps to only create a single promise in that case instead of n, where n is the number of frames.

packages/dd-trace/src/debugger/devtools_client/state.js
watson: Fixed 👍

packages/dd-trace/src/debugger/devtools_client/state.js
watson: Fixed - kinda.. I realized that we're already importing `./source-map` in the root of the module, so there's no need for this runtime require (not sure why I did that)

packages/dd-trace/src/debugger/devtools_client/state.js
watson: To keep the code simpler, I prefer this approach. The rationale is that this code is running in a worker thread at a point in time that's not blocking the main thread, so here simpler code is preferred IMO.

BridgeAR (approved):
The code itself is LGTM. I did not closely look at the tests.

PR #7338, opened 2026-01-26
chore: update supported-configurations.json to v2 and data

BridgeAR:
See individual commits.

### DIFF
--- a/packages/dd-trace/src/config/defaults.js
+++ b/packages/dd-trace/src/config/defaults.js
@@ -36,7 +36,7 @@ function parseDefaultByType (raw, type) {
       return Number(raw)
     }
     case 'array': {
-      if (raw.length === 0) return []
+      if (!raw || raw.length === 0) return []
       // TODO: Make the parsing a helper that is reused.
       return raw.split(',').map(item => {
         const colonIndex = item.indexOf(':')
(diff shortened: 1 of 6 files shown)


packages/dd-trace/src/config/defaults.js
```diff
@@ -18,207 +15,137 @@ const service = getEnv('AWS_LAMBDA_FUNCTION_NAME') ||
...
+      return
+    case 'int':
+    case 'decimal': {
+      return Number(raw)
+    }
+    case 'array': {
+      if (raw.length === 0) return []
+      let parsedAsJson
+      try {
+        parsedAsJson = JSON.parse(raw)
```
rochdev: This is not true for all arrays.

index.d.ts
```diff
@@ -739,24 +824,32 @@ declare namespace tracer {
...
     reportHostname?: boolean
 
     /**
      * A string representing the minimum tracer log level to use when debug logging is enabled
      * @default 'debug'
+     * @env DD_TRACE_LOG_LEVEL, OTEL_LOG_LEVEL
+     * Programmatic configuration takes precedence over the environment variables listed above.
      */
-    logLevel?: 'error' | 'debug'
+    logLevel?: 'debug' | 'info' | 'warn' | 'error'
```
ab-pm, line 846: Ah this is great, I was just about to open a new issue with a "*TypeScript declaration of `TracerOptions.logLevel` is incomplete*" bug report when I saw this commit!
Notice that technically there are some more supported values: https://github.com/DataDog/dd-trace-js/blob/620a8fb2cc06db623c24d42722cb3e48e90d0f8d/packages/dd-trace/src/log/channels.js#L5-L13

PR #7351, opened 2026-01-27
perf(tests): cache dd-trace tarball for sandbox creation

watson:
## What Does This Do

Optimizes sandbox creation for parallel integration tests by caching the `dd-trace.tgz` tarball at the `sandboxRoot` level instead of having each parallel worker create its own copy.

## Motivation

When running integration tests in parallel using `scripts/mocha-parallel-files.js`, multiple processes were simultaneously attempting to pack `dd-trace.tgz` during sandbox creation. This caused severe I/O contention, making parallel execution actually slower than sequential execution.

## Performance Impact Example

Results for debugger integration tests (22 test files):

| Configuration | Time | System CPU | vs Sequential |
|---------------|------|------------|---------------|
| Sequential | 5:06 | 162s | baseline |
| Parallel (`--jobs 2`) - before | 6:47 | 424s | 33% slower ⚠️ |
| Parallel (`--jobs 2`) - after | 3:02 | 164s | **40% faster** ✅ |
| Parallel (`--jobs 4`) - after | 2:51 | 189s | **44% faster** ✅ |

The tarball is now created once per test run and reused by all workers, reducing system CPU overhead by 61% and enabling effective parallelization.

For the debugger integration tests, the optimal parallelism level is currently `--jobs 4` as further increases provide minimal benefit due to a long-running test file (`snapshot-time-budget.spec.js`) that acts as a bottleneck.

### DIFF
--- a/integration-tests/helpers/index.js
+++ b/integration-tests/helpers/index.js
@@ -351,9 +351,6 @@ async function packTarballWithLock (tarballPath, env) {
   }
 
   const lockFile = `${tarballPath}.lock`
-  const maxWaitTime = 60_000 // 60 seconds max wait
-  const pollInterval = 100 // Check every 100ms
-  const lockTimeout = 60_000 // Consider locks older than 60s as stale
   let lockFd
 
   try {
@@ -361,6 +358,7 @@ async function packTarballWithLock (tarballPath, env) {
     lockFd = await fs.open(lockFile, 'wx')
     log('Lock acquired, packing tarball:', tarballPath)
 
+    // Double-check if tarball was created while we were acquiring the lock
     if (existsSync(tarballPath)) {
       log('Tarball already exists (created while waiting for lock):', tarballPath)
       return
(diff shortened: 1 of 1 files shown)


scripts/mocha-parallel-files.js
```diff
@@ -362,8 +362,13 @@ async function main () {
           options.require = opts.require
         }
 
+        // Enable shared tarball caching for parallel runs to avoid redundant packing
+        const DD_TRACE_INTEGRATION_TARBALL_PATH = process.env.DD_TRACE_INTEGRATION_TARBALL_PATH ||
```
BridgeAR: Nit: shall we use something like `DD_TEST_ONLY_INTEGRATION_TARBALL_PATH` instead to make it clearer?

watson: I was actually considering something like `DD_TEST_SANDBOX_TARBALL_PATH` as it technically is unrelated to integrations tests, though I agree integrations tests are probably the only ones using sandboxes 🤷

watson: I've updated the code to use `DD_TEST_SANDBOX_TARBALL_PATH` and simplified the locking parts like you suggested

integration-tests/helpers/index.js
```diff
@@ -325,6 +325,68 @@ function execHelper (command, options) {
...
+
+    // We have the lock, pack the tarball
+    packTarball(tarballPath, env)
+  } catch (err) {
+    if (err.code === 'EEXIST') {
+      // Lock exists, another process is packing - wait for the tarball to appear
+      log('Lock file exists, waiting for tarball:', tarballPath)
+
+      while (!existsSync(tarballPath)) {
+        await new Promise(resolve => setTimeout(resolve, 100))
```
BridgeAR, line 375: Nit: this could use `timers/promise` setTimeout instead.

Maybe we should add a linter?

watson: Unfortunately not as this is used in guardrails

watson: See the "Support guardrails" commit

PR #7356, opened 2026-01-27
feat(test-optimization): create final_status tag on test event for jest

calvinbayer:
### What does this PR do?
Create a `@test.final_status` tag that is added to every test event that represents the final execution of that test (there can be multiple executions of a single test due to retry mechanisms being active)

### Motivation
Tracers send each test execution as a span to the backend. When retry mechanisms are enabled a single test can run several times with different outcomes. These mechanisms differ in behavior. Some stop once the test reaches the desired state such as ATR ending when a test passes. Others always run a fixed number of attempts such as EFD or Attempt to fix.

To ensure the correct final result for the customer some intermediate outcomes are suppressed and the testing framework receives an adjusted final status. For example if an early EFD attempt passes but later attempts fail the test is still reported as passed and the failing attempts are hidden to avoid failing CI pipelines.

Both external and internal users have requested a way to query tests in Datadog by this final status. Their main goal is to build monitors and alerts for hard failures on default branches.

To support this a new tag will be added to the last retry of a test named: `@test.final_status:pass|fail|skip`

### DIFF
--- a/integration-tests/jest/jest.spec.js
+++ b/integration-tests/jest/jest.spec.js
@@ -2155,72 +2155,74 @@ describe(`jest@${JEST_VERSION} commonJS`, () => {
       ])
     })
 
-    it('sets final_status tag only on last ATR retry when EFD is enabled but not active and ATR is active', async () => {
-      receiver.setInfoResponse({ endpoints: ['/evp_proxy/v4'] })
+    it(
+      'sets final_status tag only on last ATR retry when EFD is enabled but not active and ATR is active',
+      async () => {
+        receiver.setInfoResponse({ endpoints: ['/evp_proxy/v4'] })
 
-      // All tests are known, so EFD will not be active
-      receiver.setKnownTests({
-        jest: {
-          'ci-visibility/jest-flaky/flaky-passes.js': [
-            'test-flaky-test-retries can retry flaky tests',
-            'test-flaky-test-retries will not retry passed tests'
-          ]
-        }
-      })
-
-      receiver.setSettings({
-        early_flake_detection: {
-          enabled: true,
-          slow_test_retries: {
-            '5s': 2
+        // All tests are known, so EFD will not be active
+        receiver.setKnownTests({
+          jest: {
+            'ci-visibility/jest-flaky/flaky-passes.js': [
+              'test-flaky-test-retries can retry flaky tests',
+              'test-flaky-test-retries will not retry passed tests',
+            ],
           },
-          faulty_session_threshold: 100
-        },
-        known_tests_enabled: true,
-        flaky_test_retries_enabled: true
-      })
-
-      const eventsPromise = receiver
-        .gatherPayloadsMaxTimeout(({ url }) => url.endsWith('/api/v2/citestcycle'), (payloads) => {
-          const events = payloads.flatMap(({ payload }) => payload.events)
-          const tests = events
-            .filter(event => event.type === 'test')
-            .map(event => event.content)
-            .filter(test => test.meta[TEST_NAME] === 'test-flaky-test-retries can retry flaky tests')
-
-          // We expect 2 executions: the failed (retry) and the passed (last one)
-          assert.strictEqual(tests.length, 3)
-
-          // Only the last execution (the one with status 'pass') should have TEST_FINAL_STATUS tag
-          tests.sort((a, b) => a.meta.start - b.meta.start).forEach((test, idx) => {
-            if (idx < tests.length - 1) {
-              assert.ok(!(TEST_FINAL_STATUS in test.meta),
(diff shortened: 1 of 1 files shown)


integration-tests/jest/jest.spec.js
```diff
@@ -4072,13 +4260,18 @@ describe(`jest@${JEST_VERSION} commonJS`, () => {
               if (isLastAttempt) {
                 if (shouldAlwaysPass) {
                   assert.strictEqual(test.meta[TEST_MANAGEMENT_ATTEMPT_TO_FIX_PASSED], 'true')
+                  assert.strictEqual(test.meta[TEST_FINAL_STATUS], 'pass')
                 } else if (shouldFailSometimes) {
                   assert.ok(!('TEST_HAS_FAILED_ALL_RETRIES' in test.meta))
                   assert.strictEqual(test.meta[TEST_MANAGEMENT_ATTEMPT_TO_FIX_PASSED], 'false')
+                  assert.strictEqual(test.meta[TEST_FINAL_STATUS], isQuarantined ? 'skip' : 'fail')
```
juan-fernandez, line 4338: ❓ Does the spec say that if the status is to be ignored because the test is quarantined, its `final_status` should be `skip`?

integration-tests/jest/jest.spec.js
```diff
@@ -5000,6 +5228,47 @@ describe(`jest@${JEST_VERSION} commonJS`, () => {
         // Exit code should be 1 because a non-quarantined test failed
         assert.strictEqual(exitCode, 1)
       })
+
+      it('sets final_status for quarantined tests', async () => {
```
juan-fernandez:
```suggestion
      it('sets final_status to skip for quarantined tests', async () => {
```

integration-tests/jest/jest.spec.js
juan-fernandez: ok I see that this is explicitly tested later on, so nvm 😄

packages/datadog-instrumentations/src/jest.js
```diff
@@ -635,13 +645,75 @@ function getWrappedEnvironment (BaseEnvironment, jestVersion) {
...
+      if (isAttemptToFixEnabled && isFinalAttemptToFixExecution) {
+        const testStatuses = attemptToFixRetriedTestsStatuses.get(testName)
+        if (testStatuses && testStatuses.every(status => status === 'pass')) {
+          return 'pass'
+        }
+        return 'fail'
+      }
+
+      // For ATR: The last execution's status is what the framework reports
+      return status
```
juan-fernandez: this function is such a beauty 😄

packages/datadog-instrumentations/src/jest.js
```diff
@@ -635,13 +645,75 @@ function getWrappedEnvironment (BaseEnvironment, jestVersion) {
...
+
+      const isAttemptToFixEnabled =
+        this.isTestManagementTestsEnabled &&
+        isAttemptToFix &&
+        Number.isFinite(testManagementAttemptToFixRetries)
+      const isFinalAttemptToFixExecution = isAttemptToFixEnabled &&
+        numberOfExecutedRetries >= (testManagementAttemptToFixRetries + 1)
+
+      // When no retry features are active, every test execution is final
+      const noRetryFeaturesActive = !isEfdEnabled && !isAttemptToFixEnabled
```
juan-fernandez: could we return earlier in this case? (just thinking of ways to simplify this)

packages/datadog-instrumentations/src/jest.js
```diff
@@ -635,13 +645,75 @@ function getWrappedEnvironment (BaseEnvironment, jestVersion) {
...
+        const testStatuses = attemptToFixRetriedTestsStatuses.get(testName)
+        if (testStatuses && testStatuses.every(status => status === 'pass')) {
+          return 'pass'
+        }
+        return 'fail'
+      }
+
+      // For ATR: The last execution's status is what the framework reports
+      return status
+    }
```
juan-fernandez, line 743: I'm thinking maybe we split the function by feature? So we can understand what each feature does more easily

juan-fernandez (approved):
👏 Nice! LGTM other than some minor comments

integration-tests/jest/jest.spec.js
calvinbayer: Yes, updated the test name to be more explicit about the expected state

integration-tests/jest/jest.spec.js
calvinbayer: Done

packages/datadog-instrumentations/src/jest.js
calvinbayer: Done

PR #7359, opened 2026-01-27
refactor(config): consolidate agent URL calculation

watson:
### What does this PR do?

Consolidates two duplicate `getAgentUrl()` implementations into a single source of truth within the `Config` class. The agent URL is now always calculated during initialization and stored in `config.url`, eliminating the need for consumers to call helper functions.

### Motivation

There were two separate implementations of agent URL calculation:
1. In `packages/dd-trace/src/config/index.js` (as a standalone function)
2. In `packages/dd-trace/src/agent/url.js` (as a helper module)

This duplication led to inconsistencies (the helper lacked Unix socket support) and made the codebase harder to maintain.

### To-do

- [ ] Decide if `config` should expose `hostname` and `port` now that `url` will always be there
- [ ] Should the agent url property on config be renamed from `url` to `agentUrl`?


BridgeAR:
This got superseded and fixed by now. Closing

PR #7363, opened 2026-01-28
testing: adds a debug module for logging span lifecycle / channel events

wconti27:
### What does this PR do?

AI GENERATED: ONLY IMPACTS TEST CODE

I AM OPEN TO SUGGESTIONS ON BETTER WAYS TO ACHIEVE THIS LOGGING. THE GOAL WAS TO LOG SPAN LIFECYCLE EVENTS INCLUDING CHANNEL ACTIVITY AND SHIMMER/REWRITER TRANSFORMATIONS TO IMPROVE AI AGENT ABILITY TO DEBUG TESTS. GOAL WAS ALSO TO DO THIS WITHOUT CHANGING SRC CODE.

Adds some environment variables that allow enabling logging for mocha tests. This logging will help AI agents and eng alike easily see breakages in the lifecycle of a tracing event to better improve test time debugging.

Logs the following:
- Channel subscriptions, Channel publishes
- Tracing Channel subscriptions, publishes
- Orchestrion Rewriter transforms
- Span Start & End (Doesn't use DD_TRACE_DEBUG since that environment variable produces a LOT of output, and we are trying to limit output size to prevent context window explosion for AI use cases)

Adds the following env variable options:
- `DD_CHANNEL_DEBUG`=`true` - Enables this debug logging
- `DD_CHANNEL_FILTER="*http*"` - Filter output by pattern (supports `*` wildcards)
- `DD_CHANNEL_VERBOSE=true` - Show span tags on indented lines (3 per line), default is not logging span tags
- `DD_CHANNEL_SHOW_DATA=true` - Include message data in publish logs


packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,433 @@
...
+    const handler = colors.gray(`← ${fn.name || 'anon'}`)
+    log(`${formatTimestamp()} ${colors.blue('[SUB]')} ${colors.cyan(this.name)} ${handler}`)
+  }
+  return subscribe.call(this, fn)
+}
+
+dc.Channel.prototype.publish = function (msg) {
+  if (match(this.name)) {
+    let line = `${formatTimestamp()} ${colors.yellow('[PUB]')} ${colors.cyan(this.name)}`
+    if (!this.hasSubscribers) line += colors.red(' (no subscribers)')
```
tlhunter: I wonder if it makes sense to add some sort of "Dead Letter Queue" functionality directly to diagnostics channel to support this kind of thing? Basically a way to receive messages in a channel but only if nothing else is listening. WDYT @Qard?

simon-id: That defeats a bit the purpose of noop perf without listener

tlhunter: You're right

wconti27:
> Overall I like the idea. But if we're handing two spans in parallel it would be hard to differentiate them. We should log the span ID (or last X characters of the ID) with each line to assist correlating the information.

Done! Added the last 8 characters from the span id into the logs!

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,449 @@
+'use strict'
+
+if (globalThis._ddChannelDebugPatched) return
+globalThis._ddChannelDebugPatched = true
```
bengl: Why do you need this whole ifdef-style thing? Unless you clear the require cache, this file will only ever run once anyway -- OH RIGHT, TESTS.

Maybe put a comment explaining this?

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,449 @@
...
+      red: s => `\x1b[31m${s}\x1b[0m`,
+      magenta: s => `\x1b[35m${s}\x1b[0m`,
+      white: s => `\x1b[37m${s}\x1b[0m`
+    }
+
+const indent = '                    ' // 20 spaces - well past mocha's indentation
+
+/**
+ * Writes a line to stderr synchronously to prevent interleaving with stdout.
+ * @param {string} line - The line to write (can contain newlines)
```
bengl: So by definition, not necessarily a line. Let's just call this `content`.

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,449 @@
...
+ */
+function match (name) {
+  if (!filter) return true
+  const startsWild = filter.startsWith('*')
+  const endsWild = filter.endsWith('*')
+  const pattern = filter.slice(startsWild ? 1 : 0, endsWild ? -1 : undefined)
+  if (startsWild && endsWild) return name.includes(pattern)
+  if (startsWild) return name.endsWith(pattern)
+  if (endsWild) return name.startsWith(pattern)
+  return name.includes(filter)
```
bengl: This whole function seems like it should probably just be a regex?

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,449 @@
...
+ */
+function formatDuration (start) {
+  return colors.gray(`${(performance.now() - start).toFixed(2)}ms`)
+}
+
+// ============================================================
+// Channel Prototype Patching
+// ============================================================
+
+const { subscribe, publish } = dc.Channel.prototype
```
bengl: Do `dc.subscribe` and `dc.unsubscribe` (i.e. the module-level functions) call these prototype methods internally? If not, you'll need to wrap them too.

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,449 @@
...
+// TracingChannel caches runStores at creation time, and wrapping dc.channel()
+// to patch instances breaks async context propagation.
+// Use [TRACEPROMISE]/[TRACESYNC]/[TRACECALLBACK] logs instead.
+
+// ============================================================
+// TracingChannel Patching (dc-polyfill)
+// ============================================================
+
+Hook(['dc-polyfill'], exports => {
+  if (exports._channelDebugPatched) return exports
```
bengl: Why are we expecting double-patching to happen?

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,449 @@
...
+  const spanInfo = `${colors.white(name)} ${spanIdPart} ${colors.cyan(service)}${resourcePart}${kindPart}`
+  log(`${separator}\n${formatTimestamp()} ${tag} ${spanInfo}${meta}`)
+}
+
+/**
+ * Patches the startSpan method on a target object to enable span logging.
+ * @param {object} target - The object containing startSpan method
+ * @param {object} [bindTarget] - Optional object to bind startSpan to
+ */
+function patchStartSpan (target, bindTarget) {
```
bengl: bindTarget is never actually used.

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,449 @@
...
+  if (!target.startSpan || target._startSpanPatched) return
+  target._startSpanPatched = true
+  const origStartSpan = target.startSpan.bind(bindTarget || target)
+  target.startSpan = function (name, options = {}) {
+    const span = origStartSpan(name, options)
+    if (match(name)) logSpanStart(name, span, options)
+    patchSpanFinish(span, name)
```
bengl: This operates on instances. Can we not operate on the prototype?

packages/dd-trace/test/debug/channel-patch.mjs
```diff
@@ -0,0 +1,14 @@
+import { createRequire } from 'node:module'
+
+const require = createRequire(import.meta.url)
+require('./channel-patch.js')
```
bengl: What is this whole loader hook for? It looks like it's just doing pass-throughs. You can omit `initialize`, for sure, probably the other ones too.

bengl (changes requested):

packages/dd-trace/test/debug/channel-patch.js
wconti27: done.

packages/dd-trace/test/debug/channel-patch.js
wconti27: done.

packages/dd-trace/test/debug/channel-patch.js
wconti27: makes sense, done.

packages/dd-trace/test/debug/channel-patch.js
wconti27: good catch, added wrapping for them too.

packages/dd-trace/test/debug/channel-patch.mjs
wconti27: removed.

packages/dd-trace/test/debug/channel-patch.js
wconti27: its used below!

packages/dd-trace/test/debug/channel-patch.js
wconti27: AI Response: 


> The tracer has a layered architecture where the public tracer object wraps an internal _tracer. Users can call tracer.startSpan() or plugins may call tracer._tracer.startSpan() directly.                        
>                                                            
> There's no shared prototype between these - tracer is a Proxy/wrapper and _tracer is the actual DatadogTracer instance. Patching the prototype would only catch one path. Instance patching on both ensures we capture all spans regardless of which interface is used.

wconti27: As you can probably tell, this PR was mostly AI, so I'm not the most aware of why some decisions were made, but this explanation makes sense to me.

packages/dd-trace/test/debug/channel-patch.js
wconti27: AI response: 

  Why double-patching guards are needed:                                                                                                                                                                            
                                                         
  ritm (require-in-the-middle) hooks can fire multiple times for the same module in several scenarios:                                                                                                              
                                                                                                                                                                                                                  
  1. Module required from different paths (e.g., require('dc-polyfill') vs require('../node_modules/dc-polyfill'))
  2. Module cache cleared and re-required (common in test environments)
  3. Multiple instrumentation hooks registered for the same module

  Without the _channelDebugPatched guard, we'd wrap already-wrapped functions, causing duplicate log output and potential stack overflows from infinite recursion.

packages/dd-trace/test/debug/channel-patch.js
bengl: Not really. If bindTarget is omitted, then target is used instead, and this is only ever called in patchTracer, where bindTarget is either omitted or is exactly target.

packages/dd-trace/test/debug/channel-patch.mjs
bengl: Okay, but why do you need the channel patching on the loader thread?

wconti27: Good point, this was added while trying to get the patch working on ESM sandboxed tests. Works now and I've removed it.

packages/dd-trace/test/debug/channel-patch.js
wconti27: double patching guard has been removed.

packages/dd-trace/test/debug/channel-patch.js
wconti27: bindTarget is no longer used, removed.

packages/dd-trace/test/debug/channel-patch.js
wconti27: We are listening via the Diagnostics Channel for start / end span events so this thread is irrelevant.

integration-tests/helpers/index.js
```diff
@@ -720,6 +720,37 @@ function checkSpansForServiceName (spans, name) {
...
+ * @param {AdditionalEnvArgs} envArgs - Environment args that may contain NODE_OPTIONS to merge
+ * @returns {{ nodeOptions: string, envArgs: AdditionalEnvArgs }}
+ */
+function buildDebugNodeOptions (envArgs) {
+  const result = { ...envArgs }
+  let nodeOptions = `--loader=${hookFile}`
+
+  if (process.env.TEST_CHANNEL_DEBUG) {
+    const channelPatchCjs = path.join(__dirname, '../../packages/dd-trace/test/debug/channel-patch.js')
+    // Use --require for CJS context (--import would run in separate loader thread)
```
bengl: Not quite. There's just no need to bother with ESM code because everything you're patching is in CJS.

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,240 @@
...
+})
+
+// Rewriter patching - log when code gets rewritten by orchestrion
+let instrumentations = []
+try { instrumentations = require('../../../datadog-instrumentations/src/helpers/rewriter/instrumentations') } catch {}
+
+try {
+  const rewriter = require('../../../datadog-instrumentations/src/helpers/rewriter')
+  const orig = rewriter.rewrite
+  rewriter.rewrite = function (content, filename, format) {
```
bengl: I didn't realize you were rewriting `rewrite`. Because of that, you _will_ need to do all this in the loader thread too when that's used.

bengl (changes requested):

packages/dd-trace/test/plugins/agent.js
```diff
@@ -14,6 +14,9 @@ const semifies = require('semifies')
 const { assertObjectContains } = require('../../../../integration-tests/helpers')
 const { storage } = require('../../../datadog-core')
 const ritm = require('../../src/ritm')
+
+// Channel debug patching (loaded via require side-effect when TEST_CHANNEL_DEBUG is set)
+if (process.env.TEST_CHANNEL_DEBUG) require('../debug/channel-patch')
```
BridgeAR: Please use a `_DD_` prefix or something the like. It is otherwise difficult to differentiate them from non DD envs.

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,239 @@
...
+  ? { cyan: id, yellow: id, green: id, gray: id, blue: id, red: id, magenta: id, white: id }
+  : {
+      cyan: s => `\x1b[36m${s}\x1b[0m`,
+      yellow: s => `\x1b[33m${s}\x1b[0m`,
+      green: s => `\x1b[32m${s}\x1b[0m`,
+      gray: s => `\x1b[90m${s}\x1b[0m`,
+      blue: s => `\x1b[34m${s}\x1b[0m`,
+      red: s => `\x1b[31m${s}\x1b[0m`,
+      magenta: s => `\x1b[35m${s}\x1b[0m`,
+      white: s => `\x1b[37m${s}\x1b[0m`,
```
BridgeAR: These exist in `util.inspect.colors` and we could access these entries instead of declaring them again. https://nodejs.org/api/util.html#customizing-utilinspect-colors

In newer versions there is also styleText, but that is not available in Node.js v18

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,239 @@
...
+// Prefix mocha output to distinguish from debug logs
+const marker = noColor ? '>>> ' : '\x1b[33;1m>>> \x1b[0m'
+const origWrite = process.stdout.write.bind(process.stdout)
+process.stdout.write = function (chunk, enc, cb) {
+  const str = typeof chunk === 'string' ? chunk : chunk.toString()
+  if (!str.trim()) return origWrite(chunk, enc, cb)
+  return origWrite(str.split('\n').map(l => l.trim() ? marker + l : l).join('\n'), enc, cb)
+}
+
+module.exports = {}
```
BridgeAR, line 222: Could you add a comment that this file is loaded for side effects and does not export anything? It otherwise easily looks like a bug to me

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,239 @@
...
+globalThis._ddChannelDebugPatched = true
+
+const dc = require('node:diagnostics_channel')
+const { performance } = require('node:perf_hooks')
+
+// Config
+const filter = process.env.TEST_CHANNEL_FILTER || ''
+const showData = process.env.TEST_CHANNEL_SHOW_DATA === 'true'
+const verbose = process.env.TEST_CHANNEL_VERBOSE === 'true'
+const noColor = (!process.stderr.isTTY && process.env.FORCE_COLOR !== '1') || process.env.NO_COLOR
```
BridgeAR: Shall we use `hasColors()` instead? https://nodejs.org/api/tty.html#writestreamhascolorscount-env

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,239 @@
+'use strict'
+
+// Prevent double-patching when loaded multiple times
+if (globalThis._ddChannelDebugPatched) return
+globalThis._ddChannelDebugPatched = true
```
BridgeAR: We have the non-enumerable `globalThis[Symbol.for('dd-trace')]` symbol to attach DD specific logic without adding more properties to the global scope.

I am uncertain though in what circumstances we load `core.js` more than once. That should actually not be the case. If it is, we should ideally fix that instead.

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,239 @@
...
+  const origWrap = shimmer.wrap
+  shimmer.wrap = function (obj, method, wrapper) {
+    const name = obj?.constructor?.name || typeof obj
+    if (match(method) || match(name)) log(`${ts()} ${c.magenta('[WRAP]')} ${c.yellow(name)}.${c.cyan(method)}`)
+    return origWrap.call(this, obj, method, wrapper)
+  }
+  if (shimmer.massWrap) {
+    const origMass = shimmer.massWrap
+    shimmer.massWrap = function (obj, methods, wrapper) {
+      const name = obj?.constructor?.name || typeof obj
```
BridgeAR: util.inspect(obj, { depth: -1 }) should achieve a similar output that contains more information than typeof.

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,239 @@
...
+        }
+      }
+    }
+    return tc
+  }
+} catch {}
+
+// Shimmer patching - direct require, no ritm needed
+try {
+  const shimmer = require('datadog-shimmer')
```
BridgeAR: I am unsure what this loads. It is named as module, while no such module exists?

I believe it is meant to be a relative require? That would also explain the need for the `try / catch` since the loading would currently likely always just fail?

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,239 @@
...
+    if (match(name)) {
+      log(`${ts()} ${c.gray('[UNSUB]')} ${c.cyan(name)} ${c.gray('← ' + (fn.name || 'anon'))}`)
+    }
+    return orig.call(this, name, wrappers.get(fn) || fn)
+  }
+}
+/* eslint-enable n/no-unsupported-features/node-builtins */
+
+// TracingChannel patching (dc-polyfill) - direct require, no ritm needed
+try {
```
BridgeAR: I am unsure why all these rewrites are placed in `try / catch`. These should AFAIC never throw. If they do, that would in my opinion be something to look into and fix or find a specific reason to add why something might not be patched.

BridgeAR (commented):
I just had a glimpse and did not do a full review.

I understand the overall wish for an improved logging experience what is happening. I just have to think about patching everything here instead of placing regular debug calls in the code are best.

packages/dd-trace/test/plugins/agent.js
wconti27: @bengl said the exact opposite of this. To _not_ use a DD_ scoped env var.

packages/dd-trace/test/debug/channel-patch.js
wconti27: Done.

packages/dd-trace/test/debug/channel-patch.js
wconti27: Cool, switched to this `utils.inspect.colors` 👍

packages/dd-trace/test/debug/channel-patch.js
wconti27: done.

packages/dd-trace/test/debug/channel-patch.js
wconti27: I removed the usage of globals entirely.

packages/dd-trace/test/plugins/agent.js
BridgeAR: It is in a way correct that these are not DD_ envs. I wrote an underscore at the beginning, so it was in italics. I corrected that. I also remember Attila and me discussing a good prefix for testing envs, I just do not remember what it was anymore.

BridgeAR: It is in a way correct that these are not DD_ envs. I wrote an underscore at the beginning, so it was in italics. I corrected that. I also remember Attila and me discussing a good prefix for testing envs, I just do not remember what it was anymore.

BridgeAR: It is in a way correct that these are not DD_ envs. I wrote an underscore at the beginning, so it was in italics. I corrected that. I also remember Attila and me discussing a good prefix for testing envs, I just do not remember what it was anymore.

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,226 @@
...
+const { performance } = require('node:perf_hooks')
+const { inspect } = require('node:util')
+
+// Config
+const filter = process.env.TEST_CHANNEL_FILTER || ''
+const showData = process.env.TEST_CHANNEL_SHOW_DATA === 'true'
+const verbose = process.env.TEST_CHANNEL_VERBOSE === 'true'
+const useColor = process.env.NO_COLOR
+  ? false
+  : (process.stderr.hasColors?.() || process.env.FORCE_COLOR === '1')
```
BridgeAR: `hasColors` already checks for the envs `NO_COLOR` and `FORCE_COLOR`.

FORCE_COLOR could also be set to other values and should accept colors in that case https://nodejs.org/api/tty.html#writestreamgetcolordepthenv

```suggestion
const useColor = process.stderr.hasColors
  ? process.stderr.hasColors()
  : !(process.env.NO_COLOR || process.env.FORCE_COLOR === '0')
```

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,226 @@
+'use strict'
+
+// Loaded for side effects only (via --require or require()).
+// Patches diagnostic channels, shimmer, rewriter, and span lifecycle for debug logging.
+
+const dc = require('node:diagnostics_channel')
```
BridgeAR: We normally use `dc-polyfill` due to differences in Node.js versions. That way there is also no feature detection needed below.


```suggestion
const dc = require('dc-polyfill')
```

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,226 @@
...
+    let out = `${ts()} ${c.yellow('[PUB]')} ${c.cyan(this.name)}${c.red(' (no subscribers)')}`
+    if (showData && msg) out += ` ${c.gray(JSON.stringify(msg).slice(0, 80))}`
+    log(out)
+  }
+  return publish.call(this, msg)
+}
+
+// Module-level dc.subscribe/unsubscribe (Node 18.7+)
+/* eslint-disable n/no-unsupported-features/node-builtins */
+if (dc.subscribe) {
```
BridgeAR: Not needed with dc-polyfill. The same for unsubscribe.

packages/dd-trace/test/debug/channel-patch.js
```diff
@@ -0,0 +1,226 @@
...
+
+// Shimmer patching
+const shimmer = require('../../../datadog-shimmer')
+const origWrap = shimmer.wrap
+shimmer.wrap = function (obj, method, wrapper) {
+  const name = inspect(obj, { depth: -1 })
+  if (match(method) || match(name)) log(`${ts()} ${c.magenta('[WRAP]')} ${c.yellow(name)}.${c.cyan(method)}`)
+  return origWrap.call(this, obj, method, wrapper)
+}
+if (shimmer.massWrap) {
```
BridgeAR: When would this not be defined?

BridgeAR (commented):
These patches are likely going to break from time to time if we do not run these with our tests. The reason is that we could refactor code without noticing that we our patching also needs to be refactored at the same time.

I put the PR on the guild agenda to discuss the general approach instead of relying on util.debuglog.

packages/dd-trace/test/debug/channel-patch.js
BridgeAR: Please place the comment right above the exports line 😅

bengl: Why is the exports line there at all?

PR #7368, opened 2026-01-28
fix(ts): added filter into ts definition file

pabloerhard:

### What does this PR do?

Added filter definition into index.d.ts, previously it was only defined for the web interface but it is used in multiple places, with out this TypeScript users get an error when trying to add the filter option.

### DIFF
--- a/packages/datadog-plugin-http/test/client.spec.js
+++ b/packages/datadog-plugin-http/test/client.spec.js
@@ -1419,44 +1419,90 @@ describe('Plugin', () => {
       })
 
       describe('with filter configuration', () => {
-        let config
+        describe('when filter is only applied to client', () => {
+          beforeEach(() => {
+            const config = {
+              server: false,
+              client: {
+                filter: (url) => !url.includes('/health'),
+              },
+            }
 
-        beforeEach(() => {
-          config = {
-            server: false,
-            client: {
-              filter: (url) => !url.includes('/health'),
-            },
-          }
-
-          return agent.load('http', config)
-            .then(() => {
-              http = require(pluginToBeLoaded)
-              express = require('express')
-            })
-        })
-
-        it('should skip recording if the filter function returns false', done => {
-          const app = express()
-
-          app.get('/health', (req, res) => {
-            res.status(200).send()
+            return agent.load('http', config)
+              .then(() => {
+                http = require(pluginToBeLoaded)
+                express = require('express')
+              })
           })
 
-          const timer = setTimeout(done, 100)
+          it('should skip recording if the filter function returns false', done => {
+            const app = express()
 
-          agent
-            .assertSomeTraces(() => {
+            app.get('/health', (req, res) => {
+              res.status(200).send()
+            })
+
+            const timer = setTimeout(done, 100)
+
+            agent.assertSomeTraces(() => {
               clearTimeout(timer)
               done(new Error('Filtered requests should not be recorded.'))
             })
-            .catch(done)
 
-          appListener = server(app, port => {
-            const req = http.request(`${protocol}://localhost:${port}/health`, res => {
-              res.on('data', () => {})
+            appListener = server(app, port => {
+              const req = http.request(`${protocol}://localhost:${port}/health`, res => {
+                res.on('data', () => {})
+              })
+              req.end()
             })
-            req.end()
+          })
+        })
+
+        // This tests that multiple filters do not conflict
+        describe('when both server and client apply a filter', () => {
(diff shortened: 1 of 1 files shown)


BridgeAR (commented):
This makes sense! Could you also do the same for queryStringObfuscation and resourceRenamingEnabled where it applies and remove the web interface? We have to adjust the test in that case to ignore web.

index.d.ts
```diff
@@ -2564,28 +2619,7 @@ declare namespace tracer {
...
-       * - `RegExp`: regex used to obfuscate matching values
-       */
-      queryStringObfuscation?: boolean | string | RegExp;
-
-      /**
-       * Whether to enable resource renaming when the framework route is unavailable.
-       */
-      resourceRenamingEnabled?: boolean;
-    }
+    interface web extends HttpServer {}
```
BridgeAR, line 2643: Should we mention this plugin at all? It is more of an internal implementation detail. I did add it because I did want to show the options and it can be used, while it does not seem to be very helpful in this form.

index.d.ts
```diff
@@ -1572,6 +1572,16 @@ declare namespace tracer {
...
 
+      /**
+       * Custom filter function used to decide whether a URL/path is allowed.
+       * When provided, this takes precedence over allowlist/blocklist configuration.
+       * If not provided, allowlist/blocklist logic will be used instead.
+       *
+       * @param urlOrPath - The URL or path to filter
+       * @returns true to instrument the request, false to skip it
+       */
+      filter?: (urlOrPath: string) => boolean;
```
BridgeAR, line 1585: I think this belongs to HttpClient

index.d.ts
```diff
@@ -1624,6 +1634,21 @@ declare namespace tracer {
...
+
+      /**
+       * Whether (or how) to obfuscate querystring values in `http.url`.
+       *
+       * - `true`: obfuscate all values
+       * - `false`: disable obfuscation
+       * - `string`: regex string used to obfuscate matching values (empty string disables)
+       * - `RegExp`: regex used to obfuscate matching values
+       */
+      queryStringObfuscation?: boolean | string | RegExp;
```
BridgeAR, line 1648: This also applies to Http2Server

index.d.ts
```diff
@@ -1624,6 +1634,21 @@ declare namespace tracer {
...
+
+      /**
+       * Whether to enable resource renaming when the framework route is unavailable.
+       */
+      resourceRenamingEnabled?: boolean;
```
BridgeAR, line 1653: This is also used in azure_functions

index.d.ts
BridgeAR: We seem to miss tests that show how this would be used as end user. Please add these

index.d.ts
BridgeAR: I am unsure if this is the correct place for an end user. Could you please write a kind of end user test?

index.d.ts
pabloerhard: web extends HttpServer and web also uses filter

pabloerhard: unless we move forward with removing web completely then it does make sense to just leave it in HttpClient

packages/datadog-plugin-redis/test/client.spec.js
```diff
@@ -244,6 +244,35 @@ describe('Plugin', () => {
...
+          redis = require(`../../../versions/${moduleName}@${version}`).get()
+          client = redis.createClient()
+
+          await client.connect()
+        })
+
+        it('should be able to filter commands', async () => {
+          const promise = agent.assertSomeTraces(traces => {
+            assert.strictEqual(traces[0][0].resource, 'GET')
+          })
```
BridgeAR: This is not an ideal test since we only verify that `any` trace matches and that is definitely the case

packages/datadog-plugin-http/test/client.spec.js
```diff
@@ -1417,6 +1417,49 @@ describe('Plugin', () => {
           })
         })
       })
+
+      describe('with filter configuration', () => {
+        let config
+
+        beforeEach(() => {
+          config = {
+            server: false,
```
BridgeAR: Instead, could we add another filter and check if they conflict? I am uncertain if it is possible to define different filters.

And if they do, we should document that / add a comment here as well.

packages/datadog-plugin-redis/test/client.spec.js
pabloerhard: true

PR #7379, opened 2026-01-29
chore(deps): Update @fastify/multipart used in tests to the latest version

uurien:

### What does this PR do?

Update @fastify/multipart used in tests to the latest version and skips the tests for latest version of multipart with the newest version of multipart.
### Motivation

Test was failing in the automatic PR: https://github.com/DataDog/dd-trace-js/pull/7372

### DIFF
--- a/packages/dd-trace/test/plugins/externals.json
+++ b/packages/dd-trace/test/plugins/externals.json
@@ -188,7 +188,7 @@
     },
     {
       "name": "@fastify/multipart",
-      "versions": [">=6"]
+      "versions": [">=6", "9.3.0"]
     },
     {
       "name": "pg",
--- a/packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
+++ b/packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
@@ -561,6 +561,10 @@ withVersions('fastify', 'fastify', '>=2', (fastifyVersion, _, fastifyLoadedVersi
           this.skip()
         }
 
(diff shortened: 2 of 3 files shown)


packages/dd-trace/test/plugins/externals.json
```diff
@@ -188,7 +188,7 @@
     },
     {
       "name": "@fastify/multipart",
-      "versions": [">=6"]
+      "versions": [">=6", "9.3.0"]
```
uurien, line 191: This forces the execution of the tests with 9.3.0 version of `@fastify/multipart` and prevents to be skipped because it has a newer version.

BridgeAR: Are we still testing newer versions than 9.3 as well?

uurien: Yes, they are. For example here: https://github.com/DataDog/dd-trace-js/actions/runs/21481537300/job/61880423233
```
Suspicious request blocking - multipart
      with @fastify/multipart >=6 (6.0.0)
        - should not block the request without an attack
        - should block the request when attack is detected
      with @fastify/multipart >=6 (9.4.0)
        ✔ should not block the request without an attack
        ✔ should block the request when attack is detected
      with @fastify/multipart 9.3.0 (9.3.0)
        ✔ should not block the request without an attack
        ✔ should block the request when attack is detected
```

But without that, the 9.3.0 would be skipped, because it is not the oldest nor the latest, as it is the latest that works with fastify 3.0.0, I want to run it also

PR #7387, opened 2026-01-29
chore: update iitm and fix initialize

BridgeAR:
The initialize hook was expecting that it could import a cjs module. That expectation was wrong due to only working due to iitm formerly instrumenting the top level module. Since that is not the case anymore it broke our instrumentation, since that would not be possible by Node.js itself.

As drive by, this removes code from not supported versions as well as making some parts faster and cleaning up code.

### DIFF
--- a/integration-tests/startup.spec.js
+++ b/integration-tests/startup.spec.js
@@ -25,7 +25,7 @@ const execArgvs = [
   },
   {
     execArgv: ['--import', 'dd-trace/loader-hook.mjs'],
-    skip: semver.satisfies(process.versions.node, '<20.6')
+    skip: semver.satisfies(process.versions.node, '<20.6'),
   },
   {
     execArgv: ['--loader', 'dd-trace/loader-hook.mjs'],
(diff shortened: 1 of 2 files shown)


initialize.mjs
```diff
@@ -13,48 +12,65 @@
...
 
 let hasInsertedInit = false
-function insertInit (result) {
-  if (!hasInsertedInit) {
-    hasInsertedInit = true
-    result.source = `
-import '${fileURLToPath(new URL('init.js', import.meta.url))}';
-${result.source}`
+const initJsUrl = new URL('init.js', import.meta.url).href
+// Note: `--loader` only reliably influences ESM entrypoints; for CJS apps use `--import`/`--require`.
```
bengl: We'll need to document this elsewhere, for sure.

bengl (approved):

initialize.mjs
BridgeAR: Discussed that it is currently not used and therefore ok

PR #7389, opened 2026-01-29
ci: fix github rate limit

BridgeAR:


.github/actions/node/action.yml
```diff
@@ -19,6 +19,44 @@ runs:
           inputs.version }}
         check-latest: true
         registry-url: ${{ inputs.registry-url || 'https://registry.npmjs.org' }}
+    # Avoid GitHub REST API tag lookups (api.github.com/.../git/refs/tags) which can hit installation rate limits
+    # in large orgs / high-parallelism workflows. Instead, use a direct release asset URL.
+    #
+    # Note: setup-bun's `bun-download-url` skips its own AVX2 detection, so we do a minimal check ourselves for
+    # Linux x64 (fallback to baseline build when AVX2 isn't present).
```
watson, line 26: Would it make sense to contribute this upstream?

.github/workflows/all-green.yml
```diff
@@ -21,8 +21,9 @@ jobs:
     steps:
       - uses: wechuli/allcheckspassed@b0eaef5be76003932c9520e2b3dd03c85cc43d87 # v2.2.0
         with:
-          delay: ${{ github.run_attempt == 1 && '6' || '0' }} # 6m on first attempt, no delay on reruns
-          retries: 30 # once per minute, some checks take up to 15 min, retries are possible
+          delay: ${{ github.run_attempt == 1 && '7' || '0' }} # 7 minutes on first attempt, no delay on reruns
+          retries: 15
+          polling_interval: 2 # Once every 2 minutes
```
watson, line 26: I would love if we could avoid this one. I often sit and wait for a PR to be merged... but I guess one extra minute isn't going to kill us - just a personal annoyance 🙄

.github/workflows/platform.yml
```diff
@@ -36,7 +36,7 @@ jobs:
           - name: yarn
             install: yarn add
           - name: yarn-berry
-            install: npm install -g corepack && yarn set version stable && yarn config set nodeLinker node-modules && yarn add
+            install: (command -v corepack >/dev/null 2>&1 || npm install -g corepack) && yarn set version stable && yarn config set nodeLinker node-modules && yarn add
```
watson, line 39: Do `npm install corepack` use the GitHub API?!?

BridgeAR: No, it would just fail if corepack is already installed. I am not sure why. This will in fact reduce the overall calls one way or the other.

.github/workflows/all-green.yml
BridgeAR: I would also prefer it otherwise. I think for now we should just see how well this fixes it and try to reduce it later again if it does not work well enough.

.github/actions/node/action.yml
BridgeAR: Likely yes, I will look at it at a later point

.github/workflows/platform.yml
watson: What? `npm install` should not fail if it's already installed?

> This will in fact reduce the overall calls one way or the other.

Again, is this GitHub API calls you're talking about? I'm just trying to understand if this is a fly-by fix or actually related to this PR

BridgeAR: It is corepack related, not npm related. Corepack will fail in case it is already present. The issue was pretty weird though.

I believe this will just improve the situation either way: we should not install corepack, if it is already installed.

watson: From the error, npm is failing with an 503 http server error trying to fetch the URL https://registry.npmjs.org/yarn/latest - I bet you it's just a temporary npm server error we hit. But let's keep your improvement regardless

PR #7391, opened 2026-01-30
ci: fix coverage reports and prevent detecting fixture files

BridgeAR:
Our test optimization product is using fixture files that are automatically picked up by CodeCov. Instead of having to declare our coverage files in all CI scripts, this just renames these fixture files, since they are just imported to read their content.

To prevent future issues, a small lint job is added that prevents names that might be detected as coverage.

The scripts are now way simpler due to the fix and a dynamic config is used for nyc.

While this officially 'decreases' our coverage, the opposite is actually the case.

We now track 9000 more lines and 200 files of which we cover 6000 lines. That is why our overall coverage drops while we actually properly handle the coverage now.

```diff
@@            Coverage Diff             @@
##           master    https://github.com/DataDog/dd-trace-js/issues/7391      +/-   ##
==========================================
- Coverage   86.16%   80.58%   -5.58%     
==========================================
  Files         515      731     +216     
  Lines       22249    31234    +8985     
==========================================
+ Hits        19171    25171    +6000     
- Misses       3078     6063    +2985
```

All our uploaded files are now correct and none create errors:

image

This PR makes our coverage way more solid and is a first step towards also tracking the coverage of integration tests.

Coverage is now tracked in individual folders per Node.js version and plugin name. That way CodeCov is merging the coverage instead of nyc and we never replace existing coverage.

### DIFF
--- a/scripts/check-no-coverage-artifacts.js
+++ b/scripts/check-no-coverage-artifacts.js
@@ -14,7 +14,7 @@ const FORBIDDEN_BASENAMES = new Set([
   'cobertura-coverage.xml', // istanbul/nyc: cobertura reporter output
   'clover.xml', // istanbul/nyc: clover reporter output
   'coverage.xml', // Generic coverage XML basename used by various ecosystems/tools (avoid committing it)
-  'cobertura.xml' // Alternative Cobertura XML basename used by some tools/setups (avoid committing it)
+  'cobertura.xml', // Alternative Cobertura XML basename used by some tools/setups (avoid committing it)
 ])
 
 /**
--- a/scripts/verify-coverage.js
+++ b/scripts/verify-coverage.js
@@ -8,9 +8,9 @@ const { parseArgs, format } = require('node:util')
 const { values } = parseArgs({
   args: process.argv.slice(2),
   options: {
-    flags: { type: 'string', default: '' }
+    flags: { type: 'string', default: '' },
   },
-  allowPositionals: true
+  allowPositionals: true,
 })
 
 const cwd = process.cwd()
--- a/nyc.config.js
+++ b/nyc.config.js
@@ -15,7 +15,7 @@ const label = `-${event.replaceAll(/[^a-zA-Z0-9._-]+/g, '-')}`
 module.exports = {
   reporter: [
     'text',
-    'lcov'
+    'lcov',
   ],
   include: [
     '**/ext/**/*.{js,mjs}',
@@ -47,5 +47,5 @@ module.exports = {
   // Baseline coverage is disabled because some of our CI suites only run
   // coverage on a small subset of the codebase. The only value we may trust is
   // the combined coverage of all suites.
-  'check-coverage': false
+  'check-coverage': false,
 }


.nycrc
```diff
@@ -2,5 +2,11 @@
...
     "text",
     "lcov"
-  ]
+  ],
+  "exclude": [
+    "**/test/**/fixtures/**",
+    "**/test/**/resources/**",
+    "**/integration-tests/**/fixtures/**"
+  ],
+  "excludeAfterRemap": true
```
BridgeAR:
```suggestion
```

.github/workflows/apm-integrations.yml
BridgeAR:
```suggestion
```

scripts/check-no-coverage-artifacts.js
```diff
@@ -0,0 +1,52 @@
...
+for (const file of listFilesFromGit()) {
+  const basename = path.basename(file)
+
+  // Be extra conservative: JSON fixtures that look like coverage reports are commonly misidentified by tooling.
+  if (basename.endsWith('.json') && /coverage/i.test(basename)) {
+    matches.push(file.replaceAll('\\', '/'))
+    continue
+  }
+
+  if (FORBIDDEN_BASENAMES.has(basename)) {
```
watson: Nit:

```suggestion
  } else if (FORBIDDEN_BASENAMES.has(basename)) {
```

package.json
```diff
@@ -71,10 +71,10 @@
     "test:integration:vitest": "mocha --timeout 60000 \"integration-tests/vitest/*.spec.js\"",
     "test:integration:testopt": "mocha --timeout 120000 \"integration-tests/ci-visibility/*.spec.js\"",
     "test:integration:profiler": "mocha --timeout 180000 \"integration-tests/profiler/*.spec.js\"",
-    "test:integration:plugins": "mocha \"packages/datadog-plugin-@($(echo $PLUGINS))/test/integration-test/**/*.spec.js\"",
-    "test:unit:plugins": "mocha \"packages/datadog-instrumentations/test/@($(echo $PLUGINS)).spec.js\" \"packages/datadog-plugin-@($(echo $PLUGINS))/test/**/*.spec.js\" --exclude \"packages/datadog-plugin-@($(echo $PLUGINS))/test/integration-test/**/*.spec.js\"",
+    "test:integration:plugins": "mocha \"packages/datadog-plugin-@(${PLUGINS})/test/integration-test/**/*.spec.js\"",
+    "test:unit:plugins": "mocha \"packages/datadog-instrumentations/test/@(${PLUGINS}).spec.js\" \"packages/datadog-plugin-@(${PLUGINS})/test/**/*.spec.js\"",
```
watson: This change will also include the `integration-test` folder in the unit tests. That seems like a mistake

scripts/verify-coverage.js
```diff
@@ -0,0 +1,82 @@
...
+  let lcovContent
+  try {
+    lcovContent = fs.readFileSync(lcovPath, 'utf8')
+  } catch {
+    // ignore
+  }
+
+  // Consider it empty unless we see at least one `SF:` record.
+  const isMissingOrEmpty = lcovContent === undefined || !/(^|\n)SF:/.test(lcovContent)
+
```
watson: Nit:

```suggestion
```

watson (commented):
LGTM in general, just one small thing which I think is a mistake, and two non-blocking nits.

package.json
BridgeAR: That was indeed not meant to be removed

watson (approved):

watson (approved):

PR #7395, opened 2026-01-30
fix(dsm): move dsm plugin init to start from bindStart

robcarlan-datadog:
## Summary
- DSM context is correctly bound when set in `start`, which runs inside the async context that `bindStart` establishes. Setting it in `bindStart` itself caused race conditions where the DSM context could leak across concurrent operations. This can cause DSM to read context from incorrect operations, leading to incorrect pathways and latencies.
- Move DSM checkpoint logic from `bindStart` to `start` in messaging plugins (kafkajs, amqplib, bullmq), which also fixes Confluent Kafka because that extends kafkajs.
- I moved Rhea and Google Cloud Pubsub out of this because the tests are failing. I will fix those implementations separately.

## Test plan
- [x] Regression tests (see failing PR before changes [here](https://github.com/DataDog/dd-trace-js/pull/7580))
- [ x] Tested locally


tlhunter:
Maybe we would be better off having some sort of bindStart static helper in the producer / consumer base classes? Something to signal that it should always happen at the end of bindStart? That would at least reduce the number of requires and hopefully ensure that the author of the next messaging plugin remembers to call the method.

BridgeAR:
I marked it as draft while we do not have a conclusion. @rochdev @robcarlan-datadog what about having a call to discuss the implementation together? :)

packages/datadog-plugin-kafkajs/src/producer.js
```diff
@@ -71,6 +71,27 @@ class KafkajsProducerPlugin extends ProducerPlugin {
     }
   }
 
+  start (ctx) {
+    if (!this.config.dsmEnabled) return
+    const { topic, messages, clusterId, disableHeaderInjection, currentStore: { span } } = ctx
+
+    for (const message of messages) {
```
tlhunter, line 78: It's a shame we need to iterate the message list twice now.

How many messages are usually in the list? 10? 100? more?

tlhunter: It seems like the default is 16kB and the message count is how many messages fits into that limit?

rochdev: The expensive part of a loop is its content, not the loop itself, so I would say having better separation of concerns is better. In the future I'd even want DSM to have separate plugins entirely that are not loaded at all when DSM is disabled.

robcarlan-datadog: Yeah, it really depends on the client configuration how many messages would be present. It at least exits early in the vast majority of cases where DSM isn't enabled

PR #7400, opened 2026-01-31
test: improve install_plugin_modules

BridgeAR:
The script now runs better in parallel as well as installing the right versions. Former, we installed the same versions multiple times and skipped a couple of versions.

Ideally, we always install an upper and lower bound as minimum (or even all majors).

### DIFF
--- a/packages/datadog-instrumentations/test/express-mongo-sanitize.spec.js
+++ b/packages/datadog-instrumentations/test/express-mongo-sanitize.spec.js
@@ -22,9 +22,10 @@ describe('express-mongo-sanitize', () => {
       })
 
       before((done) => {
-        // Highest version of express that supports express-mongo-sanitize is <5
-        // This is defined in externals.js
-        const express = require('../../../versions/express').get()
+        // Highest version of express that supports express-mongo-sanitize is <5 (express 5 makes req.query a
+        // getter-only property). The unversioned `versions/express` folder tracks the latest express (now 5.x), so
+        // pin the <5 range that externals.js installs for this plugin.
+        const express = require('../../../versions/express@>=4 <5').get()
         const expressMongoSanitize = require(`../../../versions/express-mongo-sanitize@${version}`).get()
         const app = express()
 
--- a/packages/datadog-plugin-google-cloud-pubsub/test/index.spec.js
+++ b/packages/datadog-plugin-google-cloud-pubsub/test/index.spec.js
@@ -2,9 +2,10 @@
 
 const assert = require('node:assert/strict')
 const { inspect } = require('node:util')
-const sinon = require('sinon')
 
 const { after, afterEach, before, beforeEach, describe, it } = require('mocha')
+const semver = require('semver')
+const sinon = require('sinon')
 
 const { ERROR_MESSAGE, ERROR_TYPE, ERROR_STACK } = require('../../dd-trace/src/constants')
 const id = require('../../dd-trace/src/id')
@@ -56,6 +57,12 @@ describe('Plugin', () => {
       let v1
       let gax
 
+      // pubsub 2.x bundles its own @grpc/grpc-js, so the internal v1 client rejects credentials minted by the
+      // test-pinned google-gax@3.5.7 ("Channel credentials must be a ChannelCredentials object"). 1.x and 3.x+
+      // stay compatible, so gate only the low-level v1 tests on 2.x; the public-API tests still cover that major.
+      const pkgVersion = require(`../../../versions/@google-cloud/pubsub@${version}`).version()
+      const itInternalApi = semver.satisfies(pkgVersion, '2.x') ? it.skip : it
+
       describe('without configuration', () => {
         beforeEach(() => {
           return agent.load('google-cloud-pubsub', { dsmEnabled: false }, { flushMinSpans: 1 })
@@ -92,7 +99,7 @@ describe('Plugin', () => {
             return expectedSpanPromise
           })
 
-          it('should be instrumented when using the internal API', async () => {
+          itInternalApi('should be instrumented when using the internal API', async () => {
             const publisher = new v1.PublisherClient({
               grpc: gax.grpc,
               projectId: project,
@@ -117,7 +124,7 @@ describe('Plugin', () => {
             return expectedSpanPromise
           })
 
-          it('should be instrumented w/ error', async () => {
+          itInternalApi('should be instrumented w/ error', async () => {
             const expectedSpanPromise = expectSpanWithDefaults({
               name: expectedSchema.controlPlane.opName,
               service: expectedSchema.controlPlane.serviceName,
--- a/packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
+++ b/packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
@@ -27,7 +27,7 @@ function isFastifyPluginVersionMismatch (error) {
 
 withVersions('fastify', 'fastify', '>=2', (fastifyVersion, _, fastifyLoadedVersion) => {
   describe('Suspicious request blocking - query', () => {
-    let server, requestBody, axios
+    let app, server, requestBody, axios
 
     before(() => {
       return agent.load(['fastify', 'http'], { client: false })
@@ -36,7 +36,7 @@ withVersions('fastify', 'fastify', '>=2', (fastifyVersion, _, fastifyLoadedVersi
     before((done) => {
       const fastify = require(`../../../../versions/fastify@${fastifyVersion}`).get()
 
-      const app = fastify()
+      app = fastify()
 
       app.get('/', (request, reply) => {
         requestBody()
@@ -51,9 +51,9 @@ withVersions('fastify', 'fastify', '>=2', (fastifyVersion, _, fastifyLoadedVersi
       server = app.server
     })
 
-    after(() => {
-      server.close()
-      return agent.close()
+    after(async () => {
+      await app.close()
+      await agent.close()
     })
 
     beforeEach(async () => {
@@ -93,7 +93,7 @@ withVersions('fastify', 'fastify', '>=2', (fastifyVersion, _, fastifyLoadedVersi
   })
 
   describe('Suspicious request blocking - body', () => {
-    let server, requestBody, axios
+    let app, server, requestBody, axios
 
     before(() => {
       return agent.load(['fastify', 'http'], { client: false })
@@ -102,7 +102,7 @@ withVersions('fastify', 'fastify', '>=2', (fastifyVersion, _, fastifyLoadedVersi
     before((done) => {
       const fastify = require(`../../../../versions/fastify@${fastifyVersion}`).get()
 
-      const app = fastify()
+      app = fastify()
 
       app.post('/', (request, reply) => {
         requestBody()
@@ -117,9 +117,9 @@ withVersions('fastify', 'fastify', '>=2', (fastifyVersion, _, fastifyLoadedVersi
       server = app.server
     })
 
-    after(() => {
-      server.close()
-      return agent.close()
+    after(async () => {
+      await app.close()
+      await agent.close()
     })
 
     beforeEach(async () => {
@@ -194,7 +194,7 @@ withVersions('fastify', 'fastify', '>=2', (fastifyVersion, _, fastifyLoadedVersi
   })
 
   describe('Appsec blocking with schema validation', () => {
-    let server, axios
+    let app, server, axios
 
     before(() => {
       return agent.load(['fastify', 'http'], { client: false })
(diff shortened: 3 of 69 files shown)


packages/datadog-instrumentations/test/express-mongo-sanitize.spec.js
```diff
@@ -22,9 +22,10 @@ describe('express-mongo-sanitize', () => {
       })
 
       before((done) => {
-        // Highest version of express that supports express-mongo-sanitize is <5
-        // This is defined in externals.js
-        const express = require('../../../versions/express').get()
+        // Highest version of express that supports express-mongo-sanitize is <5 (express 5 makes req.query a
+        // getter-only property). The unversioned `versions/express` folder tracks the latest express (now 5.x), so
+        // pin the <5 range that externals.js installs for this plugin.
+        const express = require('../../../versions/express@>=4 <5').get()
```
BridgeAR, line 28: Potentially change the output folders to match:

```suggestion
        const express = require('../../../versions/express@4').get()
```

packages/datadog-plugin-google-cloud-pubsub/test/index.spec.js
```diff
@@ -56,6 +57,12 @@ describe('Plugin', () => {
       let v1
       let gax
 
+      // pubsub 2.x bundles its own @grpc/grpc-js, so the internal v1 client rejects credentials minted by the
+      // test-pinned google-gax@3.5.7 ("Channel credentials must be a ChannelCredentials object"). 1.x and 3.x+
+      // stay compatible, so gate only the low-level v1 tests on 2.x; the public-API tests still cover that major.
```
BridgeAR, line 62: Improve comment. This is hard to read and follow

packages/dd-trace/test/appsec/index.fastify.plugin.spec.js
```diff
@@ -484,12 +492,18 @@ withVersions('fastify', 'fastify', '>=2', (fastifyVersion, _, fastifyLoadedVersi
...
 
-            app.listen({ port: 0 }, () => {
-              const port = (/** @type {import('net').AddressInfo} */ (server.address())).port
-              axios = Axios.create({ baseURL: `http://localhost:${port}` })
-              done()
-            })
+            try {
+              await app.listen({ host: '127.0.0.1', port: 0 })
+            } catch (error) {
+              if (isFastifyPluginVersionMismatch(error)) {
```
BridgeAR, line 498: How come this is only a problem here and not elsewhere?

packages/dd-trace/test/config/fresh-config-helper.spec.js
```diff
@@ -0,0 +1,23 @@
...
+
+    getConfigFresh({})
+    const childCount = helperModule.children.length
+
+    getConfigFresh({})
+    getConfigFresh({})
+
+    assert.strictEqual(helperModule.children.length, childCount)
+  })
+})
```
BridgeAR, line 23: Unnecessary test file

packages/dd-trace/test/plugins/versions/index.js
```diff
@@ -40,6 +50,117 @@ function capSubrange (name, subrange) {
...
+function getVersionList (name, versions, nonConsecutiveMajors = nonConsecutiveMajorPackages) {
+  /** @type {Map<string, { versionKey: string, range: string }>} */
+  const entries = new Map()
+
+  const add = (versionKey, range) => {
+    if (!entries.has(versionKey)) entries.set(versionKey, { versionKey, range })
+  }
+
+  for (const range of versions) {
+    if (!range) continue
```
BridgeAR, line 81: This should be impossible to reach (it would be a bug in the setup)

packages/dd-trace/test/plugins/versions/index.js
```diff
@@ -40,6 +50,117 @@ function capSubrange (name, subrange) {
...
+
+    if (range === '*') {
+      add('*', range)
+      continue
+    }
+
+    // Exact-version notations collapse to one key so the same version is never installed twice.
+    const exact = clean(range)
```
BridgeAR, line 89: Should we use `loose` option? That might be good to be safe for any possibilities

packages/dd-trace/test/plugins/versions/index.js
```diff
@@ -40,6 +50,117 @@ function capSubrange (name, subrange) {
...
+    add(floor.version, range)
+
+    if (!nonConsecutiveMajors.has(name)) {
+      const topMajor = highestMajor(name, range, floor.major)
+      for (let major = floor.major + 1; major < topMajor; major++) {
+        add(`>=${major}.0.0 <${major + 1}.0.0`, range)
+      }
+    }
+
+    add(range, range)
```
BridgeAR, line 107: Check about using the upper bound instead of the full range.

packages/dd-trace/test/plugins/versions/index.js
```diff
@@ -40,6 +50,117 @@ function capSubrange (name, subrange) {
...
+
+  const add = (versionKey, range) => {
+    if (!entries.has(versionKey)) entries.set(versionKey, { versionKey, range })
+  }
+
+  for (const range of versions) {
+    if (!range) continue
+
+    if (range === '*') {
+      add('*', range)
```
BridgeAR, line 84: When having a star, we always want everything. Do we want any other return values as such?

packages/dd-trace/test/plugins/externals.js
```diff
@@ -653,10 +660,4 @@ module.exports = {
       versions: ['1.20.1'],
     },
   ],
-  ws: [
-    {
-      name: 'ws',
```
BridgeAR, line 658: Check if self references are now impossible (throwing an error would be helpful)

packages/dd-trace/test/plugins/versions.spec.js
```diff
@@ -0,0 +1,136 @@
...
+    assert.deepEqual(keys('mongodb', ['*']), ['*'])
+  })
+
+  it('collapses equivalent exact-version notations to a single key', () => {
+    assert.deepEqual(keys('mongodb', ['1.2.3', '=1.2.3', 'v1.2.3']), ['1.2.3'])
+  })
+
+  it('pins the floor and keeps the range for a single-major range', () => {
+    // `<3` caps the range below any plausible pinned latest, so there is no in-between major to fill.
+    assert.deepEqual(keys('mongodb', ['>=2 <3']), ['2.0.0', '>=2 <3'])
```
BridgeAR, line 23:
```suggestion
    assert.deepEqual(keys('mongodb', ['>=2 <3']), ['2.0.0', '2'])
```

I think using this notation is easier to read and follow

packages/dd-trace/test/plugins/versions.spec.js
```diff
@@ -0,0 +1,136 @@
...
+  })
+
+  it('fills every in-between major between the floor and the range top', () => {
+    // `<5` caps the top at major 4, so only major 3 sits strictly in between major 2 and major 4.
+    assert.deepEqual(keys('mongodb', ['>=2 <5']), ['2.0.0', '>=3.0.0 <4.0.0', '>=2 <5'])
```
BridgeAR, line 28:
```suggestion
    assert.deepEqual(keys('mongodb', ['>=2 <5']), ['2.0.0', '2', '3', '4'])
```

Would be my expected install range. Otherwise we skip testing latest 2.

packages/dd-trace/test/plugins/versions.spec.js
```diff
@@ -0,0 +1,136 @@
...
+  })
+
+  it('orders fills ascending and keeps the declared range last (the top major is covered by the range itself)', () => {
+    // `<6` caps the top at major 5, which the `>=1 <6` range resolves to; only majors 2-4 are filled in between.
+    assert.deepEqual(keys('mongodb', ['>=1 <6']), [
+      '1.0.0',
+      '>=2.0.0 <3.0.0',
+      '>=3.0.0 <4.0.0',
+      '>=4.0.0 <5.0.0',
+      '>=1 <6',
```
BridgeAR, line 38:
```suggestion
      '2',
      '3',
      '4',
      '5',
```

In theory, we could even test the lower bound of each major, while that likely explodes our testing targets too drastically.

packages/dd-trace/test/plugins/versions.spec.js
```diff
@@ -0,0 +1,136 @@
...
+    ])
+  })
+
+  it('de-duplicates a shared floor across multiple ranges', () => {
+    assert.deepEqual(keys('mongodb', ['>=2 <3', '^2.0.0']), ['2.0.0', '>=2 <3', '^2.0.0'])
```
BridgeAR, line 43:
```suggestion
    assert.deepEqual(keys('mongodb', ['>=2 <3', '^2.0.0']), ['2.0.0', '2'])
```

packages/dd-trace/test/plugins/versions.spec.js
```diff
@@ -0,0 +1,136 @@
...
+  })
+
+  it('does not fill in-between majors for non-consecutive packages', () => {
+    assert.deepEqual(keys('mongodb', ['>=1 <6'], new Set(['mongodb'])), ['1.0.0', '>=1 <6'])
```
BridgeAR, line 47: Same as above and I have to check again where the non-consecutive packages are defined. They are seemingly not in a single spot!

packages/dd-trace/test/plugins/versions.spec.js
```diff
@@ -0,0 +1,136 @@
...
+  })
+
+  it('treats the built-in non-consecutive packages as floor + range only', () => {
+    // graphql jumps from 0.x to 14.x; auto-filling 1.x–13.x would fail the install.
+    assert.deepEqual(keys('graphql', ['>=0.10']), ['0.10.0', '>=0.10'])
```
BridgeAR, line 52: How do we handle non semver ranges in general? I guess we should test each minor similarly to how we handle a major version?


```suggestion
    assert.deepEqual(keys('graphql', ['>=0.10']), ['0.10.0', '0.10'])
```

We could also always fill the missing versions with `.x` to make it more explicit

packages/dd-trace/test/plugins/versions.spec.js
```diff
@@ -0,0 +1,136 @@
...
+  })
+
+  it('throws on an unparseable range', () => {
+    assert.throws(() => getVersionList('mongodb', ['not-a-version']), /Invalid version range/)
+  })
+
+  it('ignores empty entries', () => {
+    assert.deepEqual(keys('mongodb', ['', undefined, '>=2 <3']), ['2.0.0', '>=2 <3'])
```
BridgeAR, line 60: I guess we should be stricter and actually throw on these. That way we detect our own wrong values earlier

scripts/install_plugin_modules.js
```diff
@@ -313,9 +351,21 @@ async function assertWorkspaces () {
...
+      return
+    }
+    // A non-transient failure is most often an unresolvable version: a declared range spans a major version that was
+    // never published. Point at the fix instead of leaving a bare yarn error.
+    throw new Error(
+      'yarn failed to install the generated versions/ workspaces. If a plugin declares a version range that spans a ' +
+      'major version that was never published (non-consecutive majors), add that package to ' +
+      '`nonConsecutiveMajorPackages` in packages/dd-trace/test/plugins/versions/index.js (or split the range) so its ' +
+      'in-between majors are not installed.\n' +
+      `Original error: ${error.message}`,
```
BridgeAR, line 366:
```suggestion
      'in-between majors are not installed.',
```

The cause will contain the original error message

scripts/check-version-coverage.js
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  console.log(
+    `\nChecked ${declared.length} packages ` +
+    `(${skipped.length} not on the public registry, ${errors.length} errored).`
+  )
+
+  if (gaps.length) {
+    console.log(
+      `\n${gaps.length} package(s) publish a supported major the test matrix never resolves to. ` +
```
BridgeAR, line 161: Log concrete packages

scripts/check-version-coverage.js
```diff
@@ -0,0 +1,174 @@
...
+  if (errors.length) {
+    console.log(`\nCould not reach the registry for ${errors.length} package(s):`)
+    for (const { name, error } of errors) {
+      console.log(`  ${name}: ${error}`)
+    }
+  }
+
+  console.log(
+    `\nChecked ${declared.length} packages ` +
+    `(${skipped.length} not on the public registry, ${errors.length} errored).`
```
BridgeAR, line 156: name concrete ones skipped

scripts/check-version-coverage.js
```diff
@@ -0,0 +1,174 @@
...
+      if (!published?.length) return { name, skipped: true }
+      return { name, ...analyzeCoverage(name, ranges, published) }
+    } catch (error) {
+      return { name, error: error.message }
+    }
+  })
+
+  const gaps = results.filter(result => result.untestedMajors?.length)
+  const errors = results.filter(result => result.error)
+  const skipped = results.filter(result => result.skipped)
```
BridgeAR, line 138: We can build the three arrays on the fly instead of filtering them afterwards

BridgeAR:
Superseded in smaller PRs

PR #7403, opened 2026-02-01
chore: activate unicorn/no-array-for-each eslint rule

BridgeAR:
This uses autofixes with manual improvements afterwards.

### DIFF
--- a/packages/dd-trace/src/config/index.js
+++ b/packages/dd-trace/src/config/index.js
@@ -1540,7 +1540,7 @@ function setIntegerRangeSet (obj, name, value) {
   value = value.split(',')
   const result = []
 
-  value.forEach(val => {
+  for (const val of value) {
     if (val.includes('-')) {
       const [start, end] = val.split('-').map(Number)
       for (let i = start; i <= end; i++) {
@@ -1549,7 +1549,7 @@ function setIntegerRangeSet (obj, name, value) {
     } else {
       result.push(Number(val))
     }
-  })
+  }
   obj[name] = result
 }
 
--- a/scripts/install_plugin_modules.js
+++ b/scripts/install_plugin_modules.js
@@ -21,12 +21,14 @@ const excludeList = arch() === 'arm64' ? ['aerospike', 'couchbase', 'grpc', 'ora
 const workspaces = new Set()
 const externalDeps = new Map()
 
-Object.keys(externals).forEach(external => externals[external].forEach(thing => {
-  if (thing.dep) {
-    const depsArr = externalDeps.get(external)
-    depsArr ? depsArr.push(thing) : externalDeps.set(external, [thing])
+for (const external of Object.keys(externals)) {
+  for (const thing of externals[external]) {
+    if (thing.dep) {
+      const depsArr = externalDeps.get(external)
+      depsArr ? depsArr.push(thing) : externalDeps.set(external, [thing])
+    }
   }
-}))
+}
 
 run()
 
--- a/scripts/release/helpers/terminal.js
+++ b/scripts/release/helpers/terminal.js
@@ -15,9 +15,21 @@ const PREVIOUS = '\x1B[1A'
 const RED = '\x1B[31m'
 const RESET = '\x1B[0m'
 
-const print = (...msgs) => msgs.forEach(msg => process.stdout.write(msg))
-const log = (...msgs) => msgs.forEach(msg => print(`${msg}\n`))
-const fatal = (...msgs) => fail() || log(...msgs) || process.exit(1)
+const print = (...msgs) => {
+  for (const msg of msgs) {
+    process.stdout.write(msg)
+  }
+}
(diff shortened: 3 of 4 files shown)


packages/datadog-instrumentations/src/mocha/main.js
```diff
@@ -496,6 +497,7 @@ addHook({
         status = 'skip'
       } else {
         // has to check every test in the test file
+        // eslint-disable-next-line unicorn/no-array-for-each
```
tlhunter, line 498: Why skip this one?

tlhunter (approved):

packages/datadog-instrumentations/src/mocha/main.js
BridgeAR: Mocha was weird. I believe they have own methods that are not behaving as the native one.

BridgeAR: @juan-fernandez you might know more

packages/dd-trace/src/appsec/iast/security-controls/parser.js
```diff
@@ -12,23 +12,48 @@ const SANITIZER_TYPE = 'SANITIZER'
...
+ *   file: string,
+ *   method: string,
+ *   parameters: number[] | undefined
+ * }>>}
+ */
 function parse (securityControlsConfiguration) {
   const controls = new Map()
 
-  securityControlsConfiguration?.replaceAll(/[\r\n\t\v\f]*/g, '')
+  const potentialControls = securityControlsConfiguration
```
sabrenner, line 28: looks like before the code assumed `securityControlsConfiguration` could be undefined, we should probably maintain that same behavior still wih `securityControlsConfiguration?.replaceAll(...)`

packages/dd-trace/src/appsec/iast/security-controls/parser.js
```diff
@@ -48,32 +73,39 @@ function parseControl (control) {
...
 
-  file = file?.trim()
+  file = file.trim()
 
   method = method?.trim()
 
   try {
-    parameters = getParameters(parameters)
+    const parsedParameters = getParameters(parameters)
+    return { type, secureMarks, file, method, parameters: parsedParameters }
```
sabrenner, line 90: this seems like different behavior i think - before we were always returning an object, possibly with `parameters` as `undefined`, but now we're only returning when we successfully parse `parameters`. i think maybe we should revert to the original logic for the scope of this pr - but also please lmk if i'm misreading this diff 😄

packages/dd-trace/src/appsec/iast/security-controls/parser.js
```diff
@@ -48,32 +73,39 @@ function parseControl (control) {
...
+  for (const mark of getSecureMarks(marks)) {
+    secureMarks |= mark
+  }
   if (secureMarks === CUSTOM_SECURE_MARK) {
     log.warn('[ASM] Invalid security control mark: %s', marks)
     return
   }
 
-  file = file?.trim()
+  file = file.trim()
```
sabrenner, line 84:
```suggestion
  file = file?.trim()
```

i could be missing it, but any reason this has been changed?

BridgeAR: It is guaranteed to be a string due to the way the code is written above. We check for at least three entries coming back after splitting a string. It is the third field and will therefore always be correct.

packages/dd-trace/src/appsec/iast/security-controls/parser.js
BridgeAR: That behavior did not change. Before, if `getParameters` would fail, we would end up in the catch clause and return `undefined`.

Now, we end up in the catch clause and the method ends without returning anything explicitly = `undefined`.

packages/dd-trace/src/appsec/iast/security-controls/parser.js
BridgeAR: Parse is called in a single spot in our code. That is in `index.js` from security-controls. We pass through a truthy `iastConfig.securityControlsConfiguration`. Thus, the behavior here would not have an impact.

packages/dd-trace/src/appsec/iast/security-controls/parser.js
sabrenner: ah ok i missed that in the collapsed section of the diff 👍

packages/dd-trace/src/appsec/iast/security-controls/parser.js
sabrenner: 🤦 i missed that we returned in the catch before

packages/dd-trace/src/appsec/iast/security-controls/parser.js
sabrenner: ok, thanks for the context!

PR #7404, opened 2026-02-01
perf: cache serverless environment detection at startup

watson:
### What does this PR do?

This PR optimizes serverless environment detection by changing it from a function that's called repeatedly to a constant that's evaluated once at module load time. The serverless environment status is static (it doesn't change during application runtime), so checking it repeatedly throughout execution was unnecessary overhead.

### Motivation

The tracer was performing repeated checks to determine if it's running in a serverless environment (AWS Lambda, GCP Functions, or Azure Functions) throughout execution. Since this status never changes after the application starts, we can safely check once at module load time and cache the result.

- **Improves performance** by eliminating redundant environment variable checks in hot path code across multiple modules
- **Simplifies the codebase** by exporting a pre-computed `IS_SERVERLESS` constant from the `serverless.js` module instead of an `isInServerlessEnvironment()` function
- **Reduces coupling** by removing the `config` parameter from `lazyProxy()` that was only needed for this check and removing the `Config._isInServerlessEnvironment()` method
- **Ensures consistency** by updating all serverless detection callsites across the codebase to use the constant

### DIFF
--- a/packages/dd-trace/test/appsec/index.spec.js
+++ b/packages/dd-trace/test/appsec/index.spec.js
@@ -1412,65 +1412,18 @@ describe('AppSec Index', function () {
 describe('AppSec Index (Serverless)', () => {
   let AppSec
   let log
-  let web
-  let blocking
-  let UserTracking
-  let appsecTelemetry
-  let graphql
-  let apiSecuritySampler
-  let rasp
-  let config
   let RuleManager
-  let Reporter
   let incomingHttpRequestStart
   let incomingHttpRequestEnd
+  let config
 
   beforeEach(() => {
     log = {
-      debug: sinon.stub(),
-      warn: sinon.stub(),
       error: sinon.stub(),
-    }
-
-    web = {
-      root: sinon.stub(),
-    }
-
-    blocking = {
-      setTemplates: sinon.stub(),
-    }
-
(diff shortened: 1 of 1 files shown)


watson:
* **#7404** <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-js/7404?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-js/7404?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/dd-trace/test/appsec/index.spec.js
```diff
@@ -1424,6 +1409,121 @@ describe('AppSec Index', function () {
...
+  let UserTracking
+  let appsecTelemetry
+  let graphql
+  let apiSecuritySampler
+  let rasp
+  let config
+  let RuleManager
+  let Reporter
+  let incomingHttpRequestStart
+  let incomingHttpRequestEnd
```
BridgeAR, line 1417: Do we need to stub all of these?

watson: Fixed 👍

PR #7412, opened 2026-02-02
feat(@langchain/langgraph): initial integration support

crysmags:
This will add support for LangGraph and was made using the APM Instrumentation Toolkit.

### DIFF
--- a/integration-tests/debugger/utils.js
+++ b/integration-tests/debugger/utils.js
@@ -9,7 +9,7 @@ const { randomUUID } = require('crypto')
 const Axios = require('axios')
 
 const { assertObjectContains, assertUUID } = require('../helpers')
-const { sandboxCwd, useSandbox, FakeAgent, spawnProc } = require('../helpers')
+const { sandboxCwd, useSandbox, FakeAgent, spawnProc, stopProc } = require('../helpers')
 const { generateProbeConfig } = require('../../packages/dd-trace/test/debugger/devtools_client/utils')
 const { version } = require('../../package.json')
 
@@ -222,7 +222,7 @@ function setup ({ env, testApp, testAppSource, dependencies, silent, stdioHandle
   })
 
   afterEach(async function () {
-    t.proc?.kill()
+    await stopProc(t.proc)
     await t.agent?.stop()
   })
 
--- a/integration-tests/helpers/index.js
+++ b/integration-tests/helpers/index.js
@@ -28,6 +28,7 @@ let shouldKill
 const ANY_STRING = Symbol('test.ANY_STRING')
 const ANY_NUMBER = Symbol('test.ANY_NUMBER')
 const ANY_VALUE = Symbol('test.ANY_VALUE')
+const defaultStopProcTimeoutMs = 2_000
 
 /**
  * @param {string} filename
@@ -65,7 +66,7 @@ async function runAndCheckOutput (filename, cwd, expectedOut, expectedSource) {
 
   if (expectedSource) {
     assert.match(out, new RegExp(`instrumentation source: ${expectedSource}`),
-    `Expected the process to output "${expectedSource}", but logs only contain: "${out}"`)
+      `Expected the process to output "${expectedSource}", but logs only contain: "${out}"`)
   }
   return pid
 }
@@ -258,6 +259,60 @@ function spawnProcAndExpectExit (filename, options = {}, stdioHandler, stderrHan
         }
         resolve()
       })
+  })
+}
+
+/**
+ * Stop a process and wait for it to fully exit.
+ *
+ * Sends `SIGTERM` first, waits up to `timeoutMs`, and escalates to `SIGKILL` if needed.
+ *
+ * @param {childProcess.ChildProcess|undefined} proc - Process to stop.
+ * @param {object} [options] - Stop options.
+ * @param {number} [options.timeoutMs=defaultStopProcTimeoutMs] - Max wait per signal in milliseconds.
+ * @returns {Promise<void>}
+ */
+async function stopProc (proc, { timeoutMs = defaultStopProcTimeoutMs } = {}) {
+  if (!proc) return
+  if (proc.exitCode !== null || proc.signalCode !== null) return
+
+  proc.kill()
+
+  const exitedAfterSigterm = await waitForProcExit(proc, timeoutMs)
+  if (exitedAfterSigterm) return
+
+  proc.kill('SIGKILL')
+  const exitedAfterSigkill = await waitForProcExit(proc, timeoutMs)
+
+  if (!exitedAfterSigkill) {
+    throw new Error(`Process ${proc.pid} did not exit after SIGKILL`)
+  }
+}
+
+/**
+ * Wait for a process to exit for up to `timeoutMs`.
+ *
+ * @param {childProcess.ChildProcess} proc - Process to wait for.
+ * @param {number} timeoutMs - Max time to wait in milliseconds.
+ * @returns {Promise<boolean>} `true` if the process exited before timeout.
+ */
+function waitForProcExit (proc, timeoutMs) {
+  if (proc.exitCode !== null || proc.signalCode !== null) {
+    return Promise.resolve(true)
+  }
+
+  return new Promise((resolve) => {
+    const timeout = setTimeout(() => {
+      proc.removeListener('exit', onExit)
+      resolve(false)
+    }, timeoutMs)
+
+    proc.once('exit', onExit)
+
+    function onExit () {
+      clearTimeout(timeout)
+      resolve(true)
+    }
   })
 }
 
@@ -954,6 +1009,7 @@ module.exports = {
   hookFile,
   assertObjectContains,
   assertUUID,
+  stopProc,
   spawnProc,
   spawnProcAndExpectExit,
   telemetryForwarder,


packages/datadog-plugin-langgraph/src/internal.js
```diff
@@ -0,0 +1,99 @@
...
+    const span = ctx.currentStore?.span
+    if (!span) {
+      return
+    }
+
+    const asyncIterable = ctx.result
+
+    const originalAsyncIterator = asyncIterable[Symbol.asyncIterator].bind(asyncIterable)
+
+    asyncIterable[Symbol.asyncIterator] = function () {
```
sabrenner: i think we should try and do this in the `langgraph.js` instrumentation file maybe, and add a custom subscriber there and use `shimmer.wrap` for robustness. then we can emit a new event with the context that this `PregelStreamPlugin` can listen for

packages/dd-trace/src/llmobs/plugins/langgraph/index.js
```diff
@@ -0,0 +1,83 @@
...
+const LLMObsPlugin = require('../base')
+const { spanHasError } = require('../../util')
+
+class BaseLangGraphLLMObsPlugin extends LLMObsPlugin {
+  static integration = 'langgraph'
+  static id = 'langgraph'
+
+  getLLMObsSpanRegisterOptions (ctx) {
+    const span = ctx.currentStore?.span
+    const name = span?.context()._tags?.['resource.name'] || 'langgraph.workflow'
```
sabrenner, line 14: i gotta check it out but ideally if we can use `ctx.self.name` if it exists (like `this.name`, `this` referring to the graph instance), that would be better i think

packages/dd-trace/src/llmobs/plugins/langgraph/index.js
```diff
@@ -0,0 +1,83 @@
...
+    const output = hasError
+      ? undefined
+      : (results !== undefined && results !== null ? this.formatIO(results) : undefined)
+
+    this._tagger.tagTextIO(span, input, output)
+  }
+
+  formatIO (data) {
+    // Handle null/undefined explicitly
+    if (data === null || data === undefined) return ''
```
sabrenner, line 39:
```suggestion
    if (data == null) return ''
```

packages/datadog-plugin-langgraph/src/internal.js
crysmags: We are trying to avoid this so that we can continue to use Orchestrion, right now since we only have `tracePromise` we need to handle this on the subscriber.

sabrenner: yeah, what i mean is we can do this out directly in the `langgraph.js` instead

```js
const orchestrionStreamCh = channel('tracing:orchestrion:@langchain/langgraph:Pregel_stream')

orchestrionStreamCh.subscribe({
  asyncStart (ctx) {
    // shimmer.wrap ctx.result
  }
})
```

just so any/all custom wrapping logic happens in instrumentation and not in the plugins directly. lmk if this doesn't make any sense tho as idk if we do something like this currently.

crysmags: I don't think this is a common pattern to mix publisher and subscriber logic in the instrumentation. I would think it would be cleaner to keep it on the subscriber/plugin side, but I can defer to @rochdev on this.

sabrenner: ah ok i don't have too strong a preference, just that we had done something like that previously for langchain before we had some more orchestrion capabilities ([ref](https://github.com/DataDog/dd-trace-js/blob/2244cf0e764263383b673352caf80be9df8b35c8/packages/datadog-instrumentations/src/langchain.js#L96-L109)).

i think in either case we should use `shimmer` to patch instead of modifying properties directly (even if shimmer just does this directly under the hood)

BridgeAR: I believe we have to implement iterator support in orchestrion-js and @rochdev already has an open PR for that, so I guess that is related.

BridgeAR:
I believe this PR is waiting for us to integrate iterator support in orchestrion-js. Thus, I marked the PR as draft until that is done

docs/add-redirects.sh
```diff
@@ -53,6 +53,7 @@ declare -a plugins=(
   "knex"
   "koa"
   "langchain"
+  "langgraph"
```
watson: I don't know about this file, but the note above says:

> Do not add any new entries to this list

tlhunter: This file was used to generate simple redirects from old doc pages to new pages. I added it two years ago when we upgraded the doc generator and the URLs changed. New modules don't need to be added since they never existed with the old URL. Having new entries added isn't beneficial but technically doing so is no harm.

Search engines have undoubtedly updated themselves to link to the new URLs by now so I'll make a PR to remove the file entirely.

tlhunter: #7593

.agents/skills/apm-integrations/references/orchestrion.md
sabrenner, line 1: nit: i think the formatting changes in this file are against our linter, we should probs revert it for the skill

packages/datadog-instrumentations/src/helpers/rewriter/instrumentations/langgraph.js
sabrenner, line 1: i think from our conversation the other week, we can get rid of tracing/rewriting `invoke` since it calls `stream` under the hood

packages/datadog-instrumentations/src/helpers/hooks.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   '@anthropic-ai/sdk': { esmFirst: true, fn: () => require('../anthropic') },
   '@apollo/server': () => require('../apollo-server'),
   '@apollo/gateway': () => require('../apollo'),
+  '@langchain/langgraph': { esmFirst: true, fn: () => require('../langgraph') },
```
sabrenner, line 7: question: i think since we're using orchestrion we don't need this `esmFirst` bit, let's see if removing it still passes with everything. i could definitely be wrong though

packages/datadog-plugin-langgraph/src/index.js
```diff
@@ -0,0 +1,24 @@
...
+
+// CRITICAL: LLMObs plugins MUST come first
+for (const Plugin of langgraphLLMObsPlugins) {
+  plugins[Plugin.id] = Plugin
+}
+
+// Tracing plugins second
+for (const [id, Plugin] of Object.entries(internalPlugin)) {
+  plugins[id] = Plugin
+}
```
sabrenner, line 17: looks like one is exported as an object and the other as an array. could we do one or the other?

packages/datadog-plugin-langgraph/src/internal.js
sabrenner, line 1: we don't use this plugin so i think we can yeet it

packages/datadog-plugin-langgraph/src/stream.js
```diff
@@ -0,0 +1,46 @@
...
+  bindStart (ctx) {
+    this.startSpan('LangGraph', {
+      service: this.config.service,
+      kind: 'internal',
+      component: 'langgraph',
+    }, ctx)
+    return ctx.currentStore
+  }
+}
+class NextStreamPlugin extends PregelStreamPlugin {
```
sabrenner, line 17: does this need to extend `PregelStreamPlugin`? i don't think there's any shared behavior between them

packages/datadog-plugin-langgraph/src/stream.js
```diff
@@ -0,0 +1,46 @@
...
+
+  asyncEnd (ctx) {
+    const span = ctx.currentStore?.span
+    if (!span) return
+    if (ctx.result.done === true) {
+      span.finish()
+    }
+  }
+
+  error (ctx) {
```
sabrenner, line 33: does the built-in error listener from the base `Plugin` class not handle errors correctly? i think in that case we could remove this function and modify the one above to check `ctx.result.done === true || spanHasError`

packages/datadog-plugin-langgraph/test/index.spec.js
```diff
@@ -0,0 +1,116 @@
...
+
+  describe('Pregel.invoke() - invoke', () => {
+    it('should generate span with correct tags (happy path)', async () => {
+      const traceAssertion = agent.assertSomeTraces((traces) => {
+        const allSpans = traces.flat()
+        const invokeSpan = allSpans.find(span => span.name === 'LangGraph')
+
+        if (!invokeSpan) {
+          throw new Error('LangGraph span not found')
+        }
```
sabrenner, line 38:
```suggestion
        assert.ok(invokeSpan)
```

i think this would fit better for a test file. looks like we can replace a few instances of this kind of thing in the file

packages/dd-trace/src/config/supported-configurations.json
```diff
@@ -2978,6 +2978,13 @@
         "default": "true"
       }
     ],
+    "DD_TRACE_LANGGRAPH_ENABLED": [
+      {
+        "implementation": "C",
```
sabrenner, line 2983: i honestly don't know much about supported config expectations, but it looks like everything else in this file is implementation "A" for initial configurations

packages/dd-trace/src/llmobs/plugins/langgraph/index.js
sabrenner: i think we should still be able to try this!

packages/dd-trace/src/llmobs/plugins/langgraph/index.js
```diff
@@ -0,0 +1,170 @@
...
+  start () {
+    // Don't register a new span - the span was already registered by PregelStreamLLMObsPlugin
+    // We just need to tag it when iteration completes
+  }
+
+  end () {
+    // Don't restore context - that will be handled by PregelStreamLLMObsPlugin
+  }
+
+  asyncStart (ctx) {
```
sabrenner, line 97: i think we should have either `asyncStart` or `asyncEnd` but probably not both

packages/dd-trace/test/llmobs/plugins/langgraph/index.spec.js
```diff
@@ -0,0 +1,524 @@
...
+            }
+          }
+
+          // Build and execute graph
+          const workflow = new StateGraph(StateAnnotation)
+            .addNode('chat', chatNode)
+            .addEdge('__start__', 'chat')
+            .addEdge('chat', '__end__')
+
+          const app = workflow.compile()
```
sabrenner, line 57:
```suggestion
          const app = workflow.compile({ name: 'foobarbaz' })
```

something like this should enable us to check if we can set the llmobs span event name from the configured name

packages/dd-trace/src/llmobs/plugins/langgraph/index.js
crysmags: what do you want to set here in case name is not set on the graph?

PR #7414, opened 2026-02-02
vendor @opentelemetry-api

rochdev:

### What does this PR do?

Vendor `@opentelemetry-api`.

### Motivation

Relying on the user to install the dependency doesn't work because the user might use a bundler, in which case our calls to `require('@opentelemetry/api')` would fail in serverless and SSI since `dd-trace` is packaged separately. I checked the code and the way the dependency works is by using globals so that multiple versions can co-exist, so we can just vendor our own and the user can use a different version and both will use the same underlying objects. This is similar to the existing behaviour when bundlers are used, or when there is an actual version mismatch.

### Additional Notes

Ideally we should just move our `TracerProvider` out of the library completely, but that is a larger effort.


rochdev:
Ended up just installing the dependency directly instead for now to solve the immediate issue. Since the strategy in this PR only works for OpenTelemetry because of their usage of globals, it's also not a generic enough solution. We'll revisit a proper solution that works for everything later on.

PR #7419, opened 2026-02-03
fix(ws): handle removing untracked event  listeners

crysmags:
When calling removeListener/off with a handler that was never added to the tracked handlers it will fall back to the original handler instead of passing undefined. This prevents errors when users remove listeners that weren't added.

### DIFF
--- a/packages/datadog-plugin-ws/test/index.spec.js
+++ b/packages/datadog-plugin-ws/test/index.spec.js
@@ -95,6 +95,26 @@ describe('Plugin', () => {
           })
         })
 
+        it('should handle removing a listener that was never added', (done) => {
+          wsServer.on('connection', (ws) => {
+            connectionReceived = true
+            ws.send('test message')
+          })
+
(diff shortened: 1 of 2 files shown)


packages/datadog-plugin-ws/test/index.spec.js
```diff
@@ -95,6 +95,26 @@ describe('Plugin', () => {
...
+          const neverAddedHandler = () => {
+            throw new Error('this should never be called')
+          }
+
+          client.on('message', (msg) => {
+            assert.strictEqual(msg.toString(), 'test message')
+            done()
+          })
+
+          assert.doesNotThrow(() => {
```
BridgeAR: Please never use `doesNotThrow`.

I cite the Node.js documentation:

> Using assert.doesNotThrow() is actually not useful because there is no benefit in catching an error and then rethrowing it. Instead, consider adding a comment next to the specific code path that should not throw and keep error messages as expressive as possible.

PR #7422, opened 2026-02-03
fix(ws): transparent instrumentation + reduced memory retention

BridgeAR:
The instrumentation was noticable by a user so far and hooked into the public event listener addition / removal. The event handlers being added could not be detected as not being the ones by the user.

That is improved here by using an internal only instrumentation that should also reduce overhead and needs less coordination between the additions and removals.

The plugin is refactored to have less memory retantion. So far, the whole span was attached and in case long living ws connections existed with a high load, the spans could add up without ever being freed. Now that tracking overhead is much lower and the span is created when needed.

### DIFF
--- a/packages/datadog-instrumentations/src/ws.js
+++ b/packages/datadog-instrumentations/src/ws.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const tracingChannel = require('dc-polyfill').tracingChannel
+const { tracingChannel } = /** @type {import('node:diagnostics_channel')} */ (require('dc-polyfill'))
 
 const shimmer = require('../../datadog-shimmer')
 const {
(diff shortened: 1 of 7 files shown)


packages/datadog-instrumentations/src/ws.js
```diff
@@ -70,35 +117,42 @@ function createWrapEmit (emit) {
...
+      }
 
-    return receiverCh.traceSync(originalHandler, ctx, this, data, binary)
-  })
-}
+      const websocket = /** @type {WebSocketInstance | undefined} */ (this[kWebSocketSymbol])
+      // Avoid receive spans when no one listens to messages.
+      if (websocket && typeof websocket.listenerCount === 'function' && websocket.listenerCount('message') === 0) {
+        return handler.call(this, data, isBinary)
+      }
```
BridgeAR, line 143: I could imagine it might be worth always tracking these instead of replicating the original implementation.

packages/datadog-instrumentations/src/ws.js
```diff
@@ -70,35 +117,42 @@ function createWrapEmit (emit) {
...
+      const socket = websocket?._sender?._socket
+      if (!socket) {
```
BridgeAR, line 145:
```suggestion
      if (!socket) {
        // This should never happen and is a protection against potential future ws changes.
```

PR #7423, opened 2026-02-04
vendor on npm prepare instead of committing to git

rochdev:

### What does this PR do?

Vendor on npm prepare instead of committing to git.

### Motivation

The choice to commit vendored dependencies was to get slightly better install times locally and in CI, and being able to install from git regardless of package manager. However, these are very small benefits, and the complexity of our CI automation has exploded because of the need to automatically re-vendor after the automation. It also has the downside that every time we touch the bundler config there are dozens of files changed every time. At this point I don't think the trade-offs are worth it to keep the files in git. Vendoring on prepare makes everything much simpler.

### Additional Notes

Only the files outside of `vendor/dist` need review, everything else is just the dist folder being deleted.

### DIFF
--- a/scripts/check_licenses.js
+++ b/scripts/check_licenses.js
@@ -84,7 +84,7 @@ function addNpmProdDeps (deps, cwd) {
 function collectDependencies (deps, obj) {
   if (!obj.dependencies) return
 
-  for (const dep in obj.dependencies) {
+  for (const dep of Object.keys(obj.dependencies)) {
     const resolved = obj.dependencies[dep].resolved
 
     if (!resolved) continue


vendor/rspack.config.js
```diff
@@ -26,9 +26,10 @@ const exclude = new Set([
   'mutexify' // we only ever use `mutexify/promise`
 ])
 
+const difference = new Set([...include].filter(x => !exclude.has(x)))
+
 module.exports = {
-  // @ts-expect-error Array#difference exists in the Node.js version being used here.
-  entry: Object.fromEntries(include.difference(exclude).entries()),
+  entry: Object.fromEntries(difference.entries()),
```
watson, line 32: Unless the expected Node.js version changed, this should work as well, and is a bit simpler:

```suggestion
module.exports = {
  // @ts-expect-error Array#difference exists in the Node.js version being used here.
  entry: [...include.difference(exclude)],
```

... as long as what you want is an array like this:

```js
[
  'dc-polyfill',
  'import-in-the-middle',
  'mutexify/promise',
  'protobufjs/minimal',
  'source-map/lib/util'
]
```

.github/workflows/dependabot-automation.yml
```diff
@@ -4,36 +4,16 @@ on:
   pull_request:
     types:
       - opened
-      - reopened
-      - synchronize
```
BridgeAR, line 8: This should be kept. We should trigger workflows again, if a PR was reopened or updated

rochdev: I guess my worry is what if dependabot opens a PR, then a malicious user adds a commit, and then dependabot updates the PR? In that scenario, the PR would be auto-approved by automation without ever validating the changes from the user.

vendor/rspack.config.js
rochdev: The expected Node version did change, because the `prepare` script needs to run in all our CI environments which include Node 16 in some cases. It wasn't a problem before because the script was run locally or in very specific CI jobs.

.github/workflows/dependabot-automation.yml
rochdev: This might be an existing problem and out of scope of this PR though, so I can add it back for now.

rochdev:
> The license script is now not picking up the vendored dependencies anymore. We should include those though.

cc @watson

watson:
> Since we do effectively the same work in JS for the validation, would it make sense to just generate it locally instead? Why do we need a separate project in Python to handle this?

Hmm good question. The pro for using the Python project, is that all Datadog repos use it, so it's guaranteed to follow our official requirements, any improvement or change in behavior in it benefit us all. You could reverse the question and ask if we should keep maintaining our own js-variant. I've already implemented support for `package-lock.json`, but just need to finish updating the tests and then I'll make a PR. Hopefully it's not a common occurrence that we change lock file formats 😅 

Additionally, our own JS-variant will need significant updates to do the same work that the Python tool currently does. As far as I know, our JS-variant only validates that a dependency exists - it doesn't fetch license info from GitHub and npm.

.github/workflows/update-3rdparty-licenses.yml
```diff
@@ -34,7 +36,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           repository: DataDog/dd-license-attribution
```
watson:
```suggestion
          repository: watson/dd-license-attribution
```

rochdev: Not sure how it was able to pass if it wasn't even the right repo :|

watson: Dunno, maybe GitHub has some magic to find the ref if a PR already exists that contains it 🤷

watson: I took a look in the GitHub Actions logs and found that it ran this command ([source](https://github.com/DataDog/dd-trace-js/actions/runs/22163628537/job/64085966052?pr=7423#step:5:74)):

```bash
git remote add origin https://github.com/DataDog/dd-license-attribution
git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=1 origin 8ea483b9f735bf8da632c89796789cc2a050a9a6
  From https://github.com/DataDog/dd-license-attribution
   * branch            8ea483b9f735bf8da632c89796789cc2a050a9a6 -> FETCH_HEAD
```

So yeah, It Just Worked™

rochdev: Is it fine to just merge using your branch for now? I mean if it works I don't see it as a problem?

.github/workflows/platform.yml
```diff
@@ -43,7 +43,8 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - uses: ./.github/actions/node/active-lts
-      - run: FILENAME=$(npm pack --pack-destination /tmp) && mv /tmp/$FILENAME /tmp/dd-trace.tgz
+      - uses: ./.github/actions/install
+      - run: FILENAME=$(npm pack --silent --pack-destination /tmp) && mv /tmp/$FILENAME /tmp/dd-trace.tgz
```
BridgeAR, line 47: Should we not keep the output to know what happened? The same about other jobs.

integration-tests/helpers/index.js
```diff
@@ -332,7 +332,7 @@ function execHelper (command, options) {
  * @param {NodeJS.ProcessEnv} env - The environment to use for the pack command
  */
 function packTarball (tarballPath, env) {
-  execHelper(`${BUN} pm pack --quiet --gzip-level 0 --filename ${tarballPath}`, { env })
+  execHelper(`${BUN} pm pack --ignore-scripts --quiet --gzip-level 0 --filename ${tarballPath}`, { env })
```
BridgeAR, line 335: If I am not mistaken, this results in our dependencies not being minified for the tests. That is not ideal, since we do not test the same circumstances in that case.

scripts/check_licenses.js
```diff
@@ -68,6 +68,36 @@ function addProdDeps (deps, cwd) {
...
+
+  const parsed = JSON.parse(stdout)
+
+  collectDependencies(deps, parsed)
+}
+
+function collectDependencies (deps, obj) {
+  if (!obj.dependencies) return
+
+  for (const dep in obj.dependencies) {
```
BridgeAR: Let us please not use `for ... in`


```suggestion
  for (const dep of Object.keys(obj.dependencies)) {
```

.github/workflows/platform.yml
rochdev: The output is used as input to the next task. Also, this was already not the case before.

integration-tests/helpers/index.js
rochdev: I don't think this is correct, otherwise the tests would not work at all. It just prevents our custom engines check from running.

scripts/check_licenses.js
rochdev: I don't see why not, it's in the language and it works for the purpose. Still changing it to avoid getting stuck on this.

rochdev:
@watson Any further concerns now that 3rd party dependencies are fixed?

integration-tests/helpers/index.js
BridgeAR: Good point, it just ignores prepare and pack scripts.

PR #7427, opened 2026-02-04
remove vscode settings

rochdev:

### What does this PR do?

Remove VSCode settings.

### Motivation

I use different settings and the ones from the repository are conflicting with them. People use different editors anyway, so committing settings for each editor doesn't make sense.

### Additional Notes

Should we do the same for Cursor etc?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,10 +93,6 @@ typings/
 
 ### VSCode ###
 .vscode/*
-!.vscode/settings.json
-!.vscode/tasks.json
-!.vscode/launch.json
-!.vscode/extensions.json
 .history
 .cursor/skills
 


.vscode/settings.json
```diff
@@ -1,33 +0,0 @@
...
-  "eslint.format.enable": true,
-  "[javascript]": {
-    "editor.defaultFormatter": "dbaeumer.vscode-eslint",
-    "editor.codeActionsOnSave": {
-      "source.fixAll.eslint": "explicit"
-    }
-  },
-  "[typescript]": {
-    "editor.defaultFormatter": "dbaeumer.vscode-eslint",
-    "editor.codeActionsOnSave": {
```
simon-id, line 29: I think this was added recently to have better devX, people's IDE would be picking the wrong linter software or something. Check the blame

rochdev:
> I think this needs a broader discussion

Agree, and the broader discussion should be to _add it_. In the meantime, it should be removed as it's causing issues.

.vscode/settings.json
rochdev: Whatever the reason, it's breaking my editor, so we should remove it until we figure out a better solution. Also, this completely ignores that at least half the team doesn't even use VSCode to begin with.

rochdev:
> I put the topic on the guild meeting

That's great, but can this also be approved so that contributors are unblocked in the meantime?

watson:
I didn't even know about the "Format document" option in VSCode - that's nice 🙂 So your point is that you want to control when the formatting happens instead of the editor doing it on save? I'm fine with that 👍

Personally I'd love it if we can get to a place where we can use prettier or something similar - it avoids all bikeshedding and reduces review turnaround times. Unfortunately a pre-commit hook is opt-in and is therefore going to reach a lot less people than if we commit editor configs 🤷

rochdev:
> I tried to re-create the bug: Is it this error?

Yes, I don't want to have to do that for all file types in the universe, nor do I want anything controlling what my editor uses to being with.

watson:
> I don't want to have to do that for all file types in the universe

I think that is a little bit exaggerated. If you encounter a new file type, you just click the button and it does it for you - it's literally one extra mouse click. And lets be honest, it's very rare that you encounter a new file type. To help you, I've now added json as a file type, so it should just work as you expect for json now.

> nor do I want anything controlling what my editor uses to being with.

I understand - and agree in general. The intention behind committing the editor config isn’t to dictate how anyone works locally, but to make sure the repository has a consistent formatting baseline. Without that, we end up spending review time on whitespace, line wrapping, etc., instead of actual logic (I can't tell you how many times I've had to add a review comment in the past about trailing line space or missing or extra trailing line breaks 🙄).

The config only affects files inside this repo, and only for the aspects we’ve explicitly agreed on (formatting rules, indentation, etc.). It doesn’t prevent you from customizing other parts of your editor.

In practice, this helps us:
- Avoid formatting debates in PRs
- Reduce diff noise and merge conflicts
- Keep the codebase consistent over time
- Make automated formatting predictable

These rules are not set in stone and we should always be open to re-evaluate them. The goal isn’t control - it’s reducing friction in collaboration.

If you don't want to partisipate in this effort, you're free to ignore the `settings.json` file locally. There's several ways you can instruct git to treat it as if it was in `.gitignore`, for example `git update-index --skip-worktree .vscode/settings.json`.

watson:
I think we're misunderstanding each other. Let's have a zoom call to discuss. A lot of context and intent is probably lost in translation here.

PR #7431, opened 2026-02-04
feat(debugger): implement capture expressions

watson:
### What does this PR do?

This PR implements capture expressions for Live Debugger/Dynamic Instrumentation log probes, providing an alternative to full snapshot capture. Instead of serializing entire object graphs, users can now define specific expressions to capture, giving them precise control over what data gets collected and addressing the 1MB Event Platform payload limit.

### Motivation

Log probes currently face a significant limitation: when `captureSnapshot: true` is enabled, the debugger serializes entire object graphs. Even with depth restrictions in place, this approach might hit the 1MB Event Platform limit. The problem becomes frustrating when users need a specific piece of deeply nested data, increasing the capture depth to reach that value inadvertently captures large amounts of uninteresting data from other parts of the object graph.

Capture expressions solve this by letting users define exactly which data they need. When `captureExpressions` is defined on a probe, the debugger switches to expression-only mode, evaluating and serializing only the specified expressions. Each expression can optionally override the default capture limits for depth, collection size, field count, and string length, providing fine-grained control over what gets captured.

The implementation includes robust error handling that distinguishes between two types of failures. Evaluation errors, like `ReferenceError` for undefined variables, are expected JavaScript errors that get reported per-expression while allowing other expressions to continue. Fatal errors, such as protocol failures, indicate something more serious has gone wrong, so the entire capture expressions feature is disabled for that probe until the probe is re-applied, preventing repeated failures on every breakpoint hit.

Time budget enforcement remains a critical aspect of the debugger's performance profile: When capture expressions are being evaluated, the timeout applies across all expressions. If the time budget is exceeded, successfully evaluated expressions are included in the snapshot, while remaining unevaluated expressions appear with a `notCapturedReason` field indicating the timeout was reached. This ensures the snapshot structure remains predictable even when operating under time pressure.

### Additional Notes

- Implements RFC: Capture Expressions (previously: Watches)
- After this PR is merged, we need to potentially update https://github.com/DataDog/system-tests/pull/6210 to the correct version (if this PR lands after 5.85.0 has been released), rerun the tests, and mark it as ready for review.

### DIFF
--- a/integration-tests/debugger/capture-expressions-time-budget.spec.js
+++ b/integration-tests/debugger/capture-expressions-time-budget.spec.js
@@ -43,8 +43,8 @@ describe('Dynamic Instrumentation', function () {
         const { captureExpressions } = snapshot.captures.lines[t.breakpoint.line]
 
         assert.deepStrictEqual(
-          Object.keys(captureExpressions),
-          ['capturedStart', 'capturedObj', 'timedOutObj1', 'timedOutStart', 'timedOutObj2']
+          Object.keys(captureExpressions).sort(),
+          ['capturedObj', 'capturedStart', 'timedOutObj1', 'timedOutObj2', 'timedOutStart']
         )
 
         // First expression should always be captured (simple primitive evaluated before deadline)
--- a/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
+++ b/packages/dd-trace/src/debugger/devtools_client/breakpoints.js
@@ -125,29 +125,30 @@ async function addBreakpoint (probe) {
   if (probe.captureExpressions?.length > 0) {
     probe.compiledCaptureExpressions = []
     for (const captureExpr of probe.captureExpressions) {
-      const limits = {
-        maxReferenceDepth: captureExpr.capture?.maxReferenceDepth ??
-          probe.capture?.maxReferenceDepth ?? DEFAULT_MAX_REFERENCE_DEPTH,
-        maxCollectionSize: captureExpr.capture?.maxCollectionSize ??
-          probe.capture?.maxCollectionSize ?? DEFAULT_MAX_COLLECTION_SIZE,
-        maxFieldCount: captureExpr.capture?.maxFieldCount ??
-          probe.capture?.maxFieldCount ?? DEFAULT_MAX_FIELD_COUNT,
-        maxLength: captureExpr.capture?.maxLength ??
-          probe.capture?.maxLength ?? DEFAULT_MAX_LENGTH,
-      }
-
+      let expression
       try {
-        probe.compiledCaptureExpressions.push({
-          name: captureExpr.name,
-          expression: compile(captureExpr.expr.json),
-          limits,
-        })
+        expression = compile(captureExpr.expr.json)
       } catch (err) {
         throw new Error(
           `Cannot compile capture expression: ${captureExpr.name} (probe: ${probe.id}, version: ${probe.version})`,
           { cause: err }
         )
       }
+
+      probe.compiledCaptureExpressions.push({
+        name: captureExpr.name,
+        expression,
+        limits: {
+          maxReferenceDepth: captureExpr.capture?.maxReferenceDepth ??
+            probe.capture?.maxReferenceDepth ?? DEFAULT_MAX_REFERENCE_DEPTH,
+          maxCollectionSize: captureExpr.capture?.maxCollectionSize ??
+            probe.capture?.maxCollectionSize ?? DEFAULT_MAX_COLLECTION_SIZE,
+          maxFieldCount: captureExpr.capture?.maxFieldCount ??
+            probe.capture?.maxFieldCount ?? DEFAULT_MAX_FIELD_COUNT,
+          maxLength: captureExpr.capture?.maxLength ??
+            probe.capture?.maxLength ?? DEFAULT_MAX_LENGTH,
+        },
+      })
     }
   }
 
--- a/packages/dd-trace/src/debugger/devtools_client/snapshot/index.js
+++ b/packages/dd-trace/src/debugger/devtools_client/snapshot/index.js
@@ -209,7 +209,7 @@ async function evaluateCaptureExpressions (callFrame, expressions, deadlineNs = 
 function extractErrorMessage (exceptionDetails) {
   const description = exceptionDetails.exception?.description
   if (!description) return 'Unknown evaluation error'
-  const newlineIndex = description.indexOf('\n')
-  if (newlineIndex === -1) return description
-  return description.slice(0, newlineIndex)
+  const startOfStackTraceIndex = description.indexOf('\n    at ')
+  if (startOfStackTraceIndex === -1) return description
+  return description.slice(0, startOfStackTraceIndex)
 }


packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
```diff
@@ -29,6 +35,13 @@ let sessionStarted = false
 const breakpointIdToProbe = new Map()
 const probeIdToBreakpointId = new Map()
 
+const limits = {
+  maxReferenceDepth: DEFAULT_MAX_REFERENCE_DEPTH,
+  maxCollectionSize: DEFAULT_MAX_COLLECTION_SIZE,
+  maxFieldCount: DEFAULT_MAX_FIELD_COUNT,
+  maxLength: DEFAULT_MAX_LENGTH,
+}
```
BridgeAR, line 43: I am not sure I understand this indirection instead of keeping it where it is used.

packages/dd-trace/src/debugger/devtools_client/snapshot/index.js
```diff
@@ -63,21 +49,167 @@ async function getLocalStateForCallFrame (
...
+  let processedResult = null
+
+  for (let i = 0; i < expressions.length; i++) {
+    const { name, expression, limits } = expressions[i]
+    const { maxReferenceDepth, maxCollectionSize, maxFieldCount, maxLength } = limits
+
+    try {
+      const { result, exceptionDetails } = /** @type {EvaluateOnCallFrameResult} */ (
+        // eslint-disable-next-line no-await-in-loop
+        await session.post('Debugger.evaluateOnCallFrame', {
```
BridgeAR, line 121: Is it not possible to do that in parallel?

watson: The CDP protocol is seriel anyway, so nothing is gained.

BridgeAR: Shall we add a comment?

packages/dd-trace/src/ci-visibility/dynamic-instrumentation/worker/index.js
watson: This is part of a performance optimization. When a probe is added, we now calculate the actual limits to be used once at add-time, instead of re-calculating it for every invocation of that probe. This means that the limits are now expected to always be present with no default fallback. This unfortunately means that in CI vis, we now can't rely on `getLocalStateForCallFrame` to know its own default limits, but have to pass it in at call time.

packages/dd-trace/src/debugger/devtools_client/snapshot/index.js
watson: We do this a 6 places in the debugger code for the same reason. ESLint doesn't support adding extra text in the same comment that is used to disable the next line, so the comment will have to be standalone on the line above. I can add such a comment at the 6 locations if you want, but personally I'm leaning towards keeping what we have now. WDYT?

jpbempel (dismissed):

integration-tests/debugger/capture-expressions-time-budget.spec.js
```diff
@@ -0,0 +1,84 @@
...
+          { name: 'timedOutObj1', expr: { dsl: 'obj', json: { ref: 'obj' } } },
+          { name: 'timedOutStart', expr: { dsl: 'start', json: { ref: 'start' } } },
+          { name: 'timedOutObj2', expr: { dsl: 'obj', json: { ref: 'obj' } } },
+        ])
+
+        const { captureExpressions } = snapshot.captures.lines[t.breakpoint.line]
+
+        assert.deepStrictEqual(
+          Object.keys(captureExpressions),
+          ['capturedStart', 'capturedObj', 'timedOutObj1', 'timedOutStart', 'timedOutObj2']
```
BridgeAR: Are these guaranteed to be in the same order in all cases? We could otherwise sort the keys.

packages/dd-trace/src/debugger/devtools_client/snapshot/index.js
```diff
@@ -63,21 +49,167 @@ async function getLocalStateForCallFrame (
...
+ *
+ * @param {import('inspector').Runtime.ExceptionDetails} exceptionDetails - The exception details
+ * @returns {string} The error message
+ */
+function extractErrorMessage (exceptionDetails) {
+  const description = exceptionDetails.exception?.description
+  if (!description) return 'Unknown evaluation error'
+  const newlineIndex = description.indexOf('\n')
+  if (newlineIndex === -1) return description
+  return description.slice(0, newlineIndex)
```
BridgeAR: Any reason to not include the full message as long as it is not part of the stack? We could also escape newlines.

packages/dd-trace/src/debugger/devtools_client/breakpoints.js
```diff
@@ -99,14 +105,52 @@ async function addBreakpoint (probe) {
...
+      maxLength: probe.capture?.maxLength ?? DEFAULT_MAX_LENGTH,
+    }
+  }
+
+  if (probe.captureExpressions?.length > 0) {
+    probe.compiledCaptureExpressions = []
+    for (const captureExpr of probe.captureExpressions) {
+      const limits = {
+        maxReferenceDepth: captureExpr.capture?.maxReferenceDepth ??
+          probe.capture?.maxReferenceDepth ?? DEFAULT_MAX_REFERENCE_DEPTH,
```
BridgeAR: Nit: these fallbacks are accessed constantly on each iteration in a worst case while we already have the values if `probe.captureSnapshot` is defined and we could just unconditionally define the object for easier access and assign `probe.capture` to that object, if needed. That would be a tad nicer to read and we do not need to access the variables multiple times in a worst case.

packages/dd-trace/src/debugger/devtools_client/breakpoints.js
```diff
@@ -99,14 +105,52 @@ async function addBreakpoint (probe) {
...
+        maxLength: captureExpr.capture?.maxLength ??
+          probe.capture?.maxLength ?? DEFAULT_MAX_LENGTH,
+      }
+
+      try {
+        probe.compiledCaptureExpressions.push({
+          name: captureExpr.name,
+          expression: compile(captureExpr.expr.json),
+          limits,
+        })
```
BridgeAR: Nit: what about limiting the try part to the compile? I know it is not so nice having the variable being defined earlier, it is just easier to see what this means to catch.

packages/dd-trace/src/debugger/devtools_client/breakpoints.js
```diff
@@ -99,14 +105,52 @@ async function addBreakpoint (probe) {
...
+      } catch (err) {
+        throw new Error(
```
BridgeAR, line 132: Are we interested in the stack trace? If not, what about using Error.stackTraceLimit = 0 for these cases? I am unsure how likely these are though.

packages/dd-trace/src/debugger/devtools_client/index.js
```diff
@@ -228,16 +252,48 @@ session.on('Debugger.paused', async ({ params }) => {
     }
 
     if (probe.captureSnapshot) {
-      if (captureErrors?.length > 0) {
+      if (fatalSnapshotErrors && fatalSnapshotErrors.length > 0) {
```
BridgeAR, line 255: Nit


```suggestion
      if (fatalSnapshotErrors?.length > 0) {
```

integration-tests/debugger/capture-expressions-time-budget.spec.js
watson: Yes, unless the CDP protocol changes its internals.

packages/dd-trace/src/debugger/devtools_client/snapshot/index.js
watson: The full message includes the stack trace, which is not relevant to the user. So I'm operating under the assumption that the actual error message is only a single line since it's coming from V8.

packages/dd-trace/src/debugger/devtools_client/index.js
watson: 😂 Yeah I know. I did this mainly to appease TS, as it complained about doing `undefined > 0`

packages/dd-trace/src/debugger/devtools_client/breakpoints.js
watson: I did it this way to avoid unnecessary object creation. If `probe.captureSnapshot` is not `true`, the `probe.capture` object is never going to be used, so there's no need to create it. I figured this one time penalty when adding the breakpoint was better than the extra object being created.

The best alternative that I can find is this, though it still creates the extra object:

```diff
+  const capture = {
+    maxReferenceDepth: probe.capture?.maxReferenceDepth ?? DEFAULT_MAX_REFERENCE_DEPTH,
+    maxCollectionSize: probe.capture?.maxCollectionSize ?? DEFAULT_MAX_COLLECTION_SIZE,
+    maxFieldCount: probe.capture?.maxFieldCount ?? DEFAULT_MAX_FIELD_COUNT,
+    maxLength: probe.capture?.maxLength ?? DEFAULT_MAX_LENGTH,
+  }
+
   if (probe.captureSnapshot) {
-    probe.capture = {
-      maxReferenceDepth: probe.capture?.maxReferenceDepth ?? DEFAULT_MAX_REFERENCE_DEPTH,
-      maxCollectionSize: probe.capture?.maxCollectionSize ?? DEFAULT_MAX_COLLECTION_SIZE,
-      maxFieldCount: probe.capture?.maxFieldCount ?? DEFAULT_MAX_FIELD_COUNT,
-      maxLength: probe.capture?.maxLength ?? DEFAULT_MAX_LENGTH,
-    }
+    probe.capture = capture
   }
 
   if (probe.captureExpressions?.length > 0) {
     probe.compiledCaptureExpressions = []
     for (const captureExpr of probe.captureExpressions) {
       const limits = {
-        maxReferenceDepth: captureExpr.capture?.maxReferenceDepth ??
-          probe.capture?.maxReferenceDepth ?? DEFAULT_MAX_REFERENCE_DEPTH,
-        maxCollectionSize: captureExpr.capture?.maxCollectionSize ??
-          probe.capture?.maxCollectionSize ?? DEFAULT_MAX_COLLECTION_SIZE,
-        maxFieldCount: captureExpr.capture?.maxFieldCount ??
-          probe.capture?.maxFieldCount ?? DEFAULT_MAX_FIELD_COUNT,
-        maxLength: captureExpr.capture?.maxLength ??
-          probe.capture?.maxLength ?? DEFAULT_MAX_LENGTH,
+        maxReferenceDepth: captureExpr.capture?.maxReferenceDepth ?? capture.maxReferenceDepth,
+        maxCollectionSize: captureExpr.capture?.maxCollectionSize ?? capture.maxCollectionSize,
+        maxFieldCount: captureExpr.capture?.maxFieldCount ?? capture.maxFieldCount,
+        maxLength: captureExpr.capture?.maxLength ?? capture.maxLength,
       }
```

Do you think the tradeoff of this implementation is better than the current one?

packages/dd-trace/src/debugger/devtools_client/breakpoints.js
watson: I wrote it like that first, but ended up with this implementation because I thought it looked nicer (I've never been a fan of having to define the variable outside the try-catch block as a `let` for it to be accessible after the try-catch block). But if you think it's better, I'll happily change it back 🙂

packages/dd-trace/src/debugger/devtools_client/breakpoints.js
watson: We use this pattern throughout the `devtools_client` worker, and it can be good to have the stack trace so you don't have to rely on the error message being unique enough so that you can find it without the stack trace. Also, it can sometimes help identify which version a given error comes from if the lines have moved around between versions and you somehow don't have the exact version number.

packages/dd-trace/src/debugger/devtools_client/breakpoints.js
watson: I love how this is handled in Ruby where you can return a value from a try-catch block and use it outside it:

```rb
result =
  begin
    do_something
  rescue SomeError
    fallback_value
  end
```

BridgeAR: It is more about having a try block being explicit about what it should catch. I agree that the ruby style is much nicer. We should likely add something similar to the language.

packages/dd-trace/src/debugger/devtools_client/breakpoints.js
BridgeAR: I can not tell how likely either case is, so I think it is best for you to decide upon that

packages/dd-trace/src/debugger/devtools_client/breakpoints.js
watson: I understand your reasoning - I just wasn't sure the pros outweighed the cons here. But this is all very nitty gritty, so I'll happily change it if you think it's best

packages/dd-trace/src/debugger/devtools_client/snapshot/index.js
BridgeAR: That could change at any time though, while it is probably not horrible when it does. I would just normally check for the `    at` to detect the end of the message.

integration-tests/debugger/capture-expressions-time-budget.spec.js
BridgeAR: While it is unlikely, should we just do it to be future proof?

packages/dd-trace/src/debugger/devtools_client/breakpoints.js
watson: A probe is not supposed to have both `probe.captureSnapshot: true` and `probe.captureExpressions.length > 0` at the same time. Those should be mutually exclusive. Some customers will lean heavily towards using Capture Expressions all the time, others will only use snapshots, others will not use any of those 🤷 I'll leave it as is for now.

packages/dd-trace/src/debugger/devtools_client/snapshot/index.js
watson: Fixed 👍

integration-tests/debugger/capture-expressions-time-budget.spec.js
watson: Fixed 👍

packages/dd-trace/src/debugger/devtools_client/breakpoints.js
watson: Fixed 👍

PR #7440, opened 2026-02-05
feat(esbuild): add DI worker bundling support

watson:
### What does this PR do?

Add support for bundling the Live Debugging (LD) / Dynamic Instrumentation (DI) worker when building applications with esbuild. This enables LD/DI features to work in bundled applications.
    
The plugin now:
- Emits a separate worker bundle (`dd-trace-debugger-worker.cjs`) alongside the main application bundle
- Patches the debugger module to reference the emitted worker
- Patches the worker's dd-trace lookup to use `global._ddtrace` for bundler compatibility

### Motivation

Without this, LD/DI would not work in a bundled application, as the worker thread would emit an `error` event when started, saying it couldn't find the worker file on disk.

### DIFF
--- a/integration-tests/esbuild/build-and-test-debugger-worker.js
+++ b/integration-tests/esbuild/build-and-test-debugger-worker.js
@@ -1,0 +1,118 @@
+#!/usr/bin/env node
+'use strict'
+
+/* eslint-disable no-console */
+
+const assert = require('node:assert/strict')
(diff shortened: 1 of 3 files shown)


BridgeAR:
Is this still worked upon? :)

BridgeAR:
I am going to close this for now due to the age. Please reopen, if it should still be worked upon.

PR #7441, opened 2026-02-05
[test optimization] Allow multiple code coverage reports per request

juan-fernandez:
### What does this PR do?
Add support for multiple code coverage reports for a single request, which the backend supports.

### Motivation
Upload faster. Example implementation: https://github.com/DataDog/datadog-ci/blob/master/packages/plugin-coverage/src/api.ts#L20


juan-fernandez:
abandoned since the intake only supports multiple files if they have the same format, which I can't make happen within a single test session (all the information gets merged on a single report unless you have multiple formats).

PR #7445, opened 2026-02-05
fix(lambda): handle missing context for some lambda functions

joeyzhao2018:
It is possible to create a lambda function without including the context in the method signature - they only receive an event parameter. Previously, the `extractContext()` function would throw "Could not extract context" when it couldn't find a context object with `getRemainingTimeInMillis`.

This change makes `extractContext()` return `null` instead of throwing, and updates the `datadog()` wrapper to skip timeout checking when context is not available. This allows these types of functions to be instrumented without crashing.

Fixes: https://github.com/DataDog/datadog-lambda-js/issues/721

### DIFF
--- a/packages/dd-trace/test/lambda/context.spec.js
+++ b/packages/dd-trace/test/lambda/context.spec.js
@@ -1,10 +1,11 @@
+'use strict'
+
 const assert = require('node:assert/strict')
-const { afterEach, beforeEach, describe, it } = require('mocha')
+const { describe, it } = require('mocha')
 const { extractContext } = require('../../src/lambda/context')
 
 describe('context', () => {
   describe('extractContext', () => {
-
     const assertExtractContext = (args, doesExtract) => {
       it(`properly extracts context object from args length ${args.length}`, () => {
         const ctx = extractContext(args)
@@ -18,22 +19,22 @@ describe('context', () => {
     }
 
     const contexts = [
-      [ null, false ],
-      [ [], false ],
-      [ {}, false ],
-      [ { getRemainingTimeInMillis: null }, false ],
-      [ { getRemainingTimeInMillis: undefined }, false ],
-      [ { getRemainingTimeInMillis: 'not a function' }, false ],
-      [ { getRemainingTimeInMillis: () => 100 }, true ],
+      [null, false],
+      [[], false],
+      [{}, false],
+      [{ getRemainingTimeInMillis: null }, false],
+      [{ getRemainingTimeInMillis: undefined }, false],
+      [{ getRemainingTimeInMillis: 'not a function' }, false],
+      [{ getRemainingTimeInMillis: () => 100 }, true],
     ]
 
     assertExtractContext([], false)
-    assertExtractContext([ {} ], false)
-    contexts.forEach(([ context, doesExtract ], index) => {
+    assertExtractContext([{}], false)
+    contexts.forEach(([context, doesExtract], index) => {
       describe(`using context case ${index + 1}`, () => {
-        assertExtractContext([ {}, context ], doesExtract)
-        assertExtractContext([ {}, {}, context ], doesExtract)
-        assertExtractContext([ {}, {}, {}, context ], false)
+        assertExtractContext([{}, context], doesExtract)
+        assertExtractContext([{}, {}, context], doesExtract)
+        assertExtractContext([{}, {}, {}, context], false)
       })
     })
   })
--- a/packages/dd-trace/src/lambda/context.js
+++ b/packages/dd-trace/src/lambda/context.js
@@ -20,5 +20,8 @@ exports.extractContext = function extractContext (args) {
       break
     }
   }
+  if (!context) {
+    log.debug('Unable to extract context object from Lambda handler arguments')
+  }
   return context
 }


packages/dd-trace/src/lambda/handler.js
```diff
@@ -63,16 +63,17 @@ function crashFlush () {
...
 function extractContext (args) {
-  let context = args.length > 1 ? args[1] : undefined
-  if (context === undefined || context.getRemainingTimeInMillis === undefined) {
-    context = args.length > 2 ? args[2] : undefined
-    if (context === undefined || context.getRemainingTimeInMillis === undefined) {
-      throw new Error('Could not extract context')
-    }
+  let context = args.length > 1 ? args[1] : null
+  if (context === null || context.getRemainingTimeInMillis === undefined) {
+    context = args.length > 2 ? args[2] : null
```
duncanista: I think we missed the second check where we ensure that the second context has the things we care about? Does it matter?

duncanista (approved):

packages/dd-trace/src/lambda/handler.js
purple4reina: I believe you are correct. Means there's a hole in our testing too. I'll fix.

duncanista (dismissed):

packages/dd-trace/src/lambda/handler.js
```diff
@@ -63,15 +63,20 @@ function crashFlush () {
 /**
  * Extracts the context from the given Lambda handler arguments.
  *
+ * It is possible for users to define a lambda function without specifying a
+ * context arg. In these cases, this function returns null instead of throwing
+ * an error.
+ *
  * @param {unknown[]} args any amount of arguments
- * @returns the context, if extraction was succesful.
+ * @returns the context if extraction was successful, null otherwise.
```
BridgeAR:
```suggestion
 * @returns {object | null}
```

packages/dd-trace/src/lambda/handler.js
```diff
@@ -63,15 +63,20 @@ function crashFlush () {
...
 function extractContext (args) {
-  let context = args.length > 1 ? args[1] : undefined
-  if (context === undefined || context.getRemainingTimeInMillis === undefined) {
-    context = args.length > 2 ? args[2] : undefined
-    if (context === undefined || context.getRemainingTimeInMillis === undefined) {
-      throw new Error('Could not extract context')
+  let context = args.length > 1 ? args[1] : null
+  if (context === null || context.getRemainingTimeInMillis === undefined) {
+    context = args.length > 2 ? args[2] : null
+    if (context === null || context.getRemainingTimeInMillis === undefined) {
```
BridgeAR: If `args[2]` would ever be undefined, this would fail accessing `context.getRemaining...`. Is that something that would be good to guard against?

BridgeAR: I believe this could be simplified to (this is a tad slower due to potentially accessing out of bounds. That could be prevented by still using the length check):

```js
let context = args?.[1]
if (!context?.getRemainingTimeInMillis && args?.[2]?.getRemainingTimeInMillis) {
  context = args[2]
}
if (!context) {
  log.debug('Could not extract the context from the Lambda handler arguments. No timeout will be tracked.')
  return
}
```

packages/dd-trace/test/lambda/index.spec.js
```diff
@@ -218,6 +218,125 @@ describe('lambda', () => {
...
+      datadog = require('./fixtures/datadog-lambda')
+
+      // Run the function - should throw but not because of missing context
+      try {
+        await datadog(app.authorizerErrorHandler)(_event)
+        assert.fail('Expected error to be thrown')
+      } catch (e) {
+        assert.strictEqual(e.name, 'AuthorizationError')
+        assert.strictEqual(e.message, 'Unauthorized')
+      }
```
BridgeAR, line 326:
```suggestion
      await assert.rejects(
        datadog(app.authorizerErrorHandler)(_event),
        { name: 'AuthorizationError', message: 'Unauthorized' }
      )
```

packages/dd-trace/test/lambda/index.spec.js
```diff
@@ -218,6 +218,125 @@ describe('lambda', () => {
...
+
+      // Expect traces to be correct with error.
+      const checkTraces = agent.assertSomeTraces((_traces) => {
+        const traces = _traces[0]
+        assert.strictEqual(traces.length, 1)
+        traces.forEach((trace) => {
+          assert.strictEqual(trace.error, 1)
+        })
+      })
+      await checkTraces
```
BridgeAR, line 336: Nit
```suggestion
      await agent.assertSomeTraces(([traces]) => {
        assert.strictEqual(traces.length, 1)
        assert.strictEqual(traces[0].error, 1)
      })
```

packages/dd-trace/test/lambda/context.spec.js
```diff
@@ -0,0 +1,40 @@
+const assert = require('node:assert/strict')
+const { afterEach, beforeEach, describe, it } = require('mocha')
+const { extractContext } = require('../../src/lambda/context')
```
BridgeAR, line 5: I understand that this creates a higher coverage for the particular internal method.

If I could choose, I would just rather always solely test public APIs or entry points as we do with the other tests. That way it is easier to refactor internals without having to change tests as well as having a stronger safe guard how the user would ever reach such state. Testing internal methods could at some point not be used anymore while still showing full coverage by being tested directly.

packages/dd-trace/test/lambda/index.spec.js
```diff
@@ -218,6 +218,125 @@ describe('lambda', () => {
...
+
+      const _event = {
+        type: 'REQUEST',
+        methodArn: 'arn:aws:execute-api:us-east-1:123456789012:api-id/stage/GET/resource',
+      }
+
+      // Mock `datadog-lambda` handler resolve and import.
+      const _handlerPath = path.resolve(__dirname, './fixtures/handler.js')
+      const app = require(_handlerPath)
+      datadog = require('./fixtures/datadog-lambda')
```
BridgeAR, line 317: These lines are identical in all tests new created tests here besides for the environment variable. What about abstracting the tests here to something like:

```js
const tests = [{ name: 'should do foo', env: 'FOO', error: 1, fn() { ... } }, ...]

for (const test of tests) {
  it(test.name, async () => {
    process.env.DD_LAMBDA_HANDLER = test.env

    await loadAgent()
    ...
    const assertionPromise = agent.assertSomeTraces(([traces]) => {
      assert.strictEqual(traces.length, 1)
      assert.strictEqual(traces[0].error, test.error)
    })
    await Promise.all([fn(), assertionPromise])
  })
}
```

That way creating new test cases is not a lot of code and we could likely add enough test cases here instead of using the new context.spec.js tests.

PR #7447, opened 2026-02-05
refactor(prisma): Prisma instrumentation rewrite

bojbrook:
### What does this PR do?
Introduces an explicit separation of the instrumentation logic and plugin logic for the Prisma plugin.
- The instrumentation layer  now wraps `Prisma’s getPrismaClient()` and publishes a single diagnostic-channel event to let the plugin provide the tracing helper.
- The plugin layer subscribes to `apm:prisma:helper:init` and injects a DatadogTracingHelper instance, keeping span creation and Datadog-specific logic out of the instrumentation layer.

### Motivation
We want a clear boundary between the instrumentation layer `packages/datadog-instrumentations/src/prisma.js` and the plugin logic `packages/datadog-plugin-prisma)`. That separation was previously not being respected and blocked supporting specific Prisma tracing features (`getTraceParent`, `getActiveContext`) without leaking responsibilities across layers.

### DIFF
--- a/packages/datadog-plugin-prisma/test/index.spec.js
+++ b/packages/datadog-plugin-prisma/test/index.spec.js
@@ -205,10 +205,12 @@ describe('Plugin', () => {
 
       assert.strictEqual(result, 'ok')
       assert.strictEqual(calls.length, 1)
-      assert.strictEqual(calls[0].type, 'promise')
-      assert.strictEqual(calls[0].ctx.resourceName, 'operation')
-      assertObjectContains(calls[0].ctx, {
-        attributes: { method: 'findMany', model: 'users' },
+      assertObjectContains(calls[0], {
+        ctx: {
+          resourceName: 'operation',
+          attributes: { method: 'findMany', model: 'users' },
+        },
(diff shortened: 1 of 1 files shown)


packages/datadog-plugin-prisma/test/index.spec.js
```diff
@@ -87,6 +92,165 @@ describe('Plugin', () => {
...
+        () => 'ok'
+      )
+
+      assert.strictEqual(result, 'ok')
+      assert.strictEqual(calls.length, 1)
+      assert.strictEqual(calls[0].type, 'promise')
+      assert.strictEqual(calls[0].ctx.resourceName, 'operation')
+      assertObjectContains(calls[0].ctx, {
+        attributes: { method: 'findMany', model: 'users' },
+      })
```
BridgeAR:
```suggestion
      assertObjectContains(calls[0], {
        ctx: {
          resourceName: 'operation',
          attributes: { method: 'findMany', model: 'users' },
        },
        type: 'promise',
      })
```

PR #7450, opened 2026-02-05
Improve graphql instrumentation

### LINKED ISSUE #2398: Add graphql parent type to "graphql.resolve" spans
Often, when I'm trying to determine if we can deprecate a field, query or mutation in our GraphQL schema, I want to see if the particular field is in use.  I can sort of get at this information by looking at all spans and filtering by the Operation Name `graphql.resolve` the `@graphql.field.path` tag on the span. However, this is imperfect, as it relies on my knowledge of all the paths within the graph to a particular field or making assumptions about what the parent type's field name will be.

What I'd *really* like to be able to do is to filter down the spans in DataDog APM by the parent type.  So, for example, if I had a schema like this:
```graphql
type MyType {
  id: ID!
  name: String!
}
```
I'd like to be able to find all spans that have, say, the `name` field with a parent type of `MyType`.  This would work for queries or mutations as well, since those are just special types (e.g. find a resource_name with a parent type of 'Query' or 'Mutation'.

ab-pm:
### What does this PR do?

* Change `resource_name` of `graphql.resolve` spans to consist of parent type and field name, making it a [schema coordinate](https://spec.graphql.org/draft/#sec-Schema-Coordinates). This uniquely identifies the field in the schema - where previously `name:String` was rather useless, it now properly distinguishes a `Human.name` from a `Pet.name`
  This imo resolves #2398, even though it does not add an extra `@graphql.field.parentType` tag as originally suggested there.
* Make `@graphql.field.type` work with non-nullable and list types

### Motivation

I was about to [patch](https://www.npmjs.com/package/patch-package) the integration for our deployment to make the traces more useful when I noticed that other people share the same issue.

### Additional Notes

I'm not sure if this is considered a breaking change ("*changing the contents or value of any existing data being reported*") and requires \<semver-major\>?
Please let me know when I should rebase or split up the PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -407,6 +407,8 @@ You can also run the tests for multiple plugins at once by separating them with 
 PLUGINS="amqplib|bluebird" yarn test:plugins
 ```
 
+The necessary shell commands for the setup can also be executed at once by the `yarn env` script.
+
 ### Other Unit Tests
 
 There are several types of unit tests, for various types of components. The
--- a/packages/datadog-instrumentations/src/graphql.js
+++ b/packages/datadog-instrumentations/src/graphql.js
@@ -177,8 +177,8 @@ function wrapExecute (execute) {
 
       return startExecuteCh.runStores(ctx, () => {
         if (schema) {
-          wrapFields(schema._queryType)
-          wrapFields(schema._mutationType)
+          wrapFields(schema.getQueryType())
+          wrapFields(schema.getMutationType())
         }
 
         contexts.set(contextValue, ctx)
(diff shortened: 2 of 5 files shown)


BridgeAR (commented):
This looks great, thank you very much!

@rochdev what is your take on the semverness? I fear this might be major as pointed out?

ab-pm:
Yes, this is still relevant.

BridgeAR:
We worked quite a lot on GraphQL recently and the requested information is now surfaced on another tag to not break existing dashboards. This got superseded as such, while it was very helpful for thinking this through! So thank you again!

PR #7453, opened 2026-02-06
sync span based baggage APIs with global baggage APIs

ida613:
### What does this PR do?
Make sure baggage added using the [global APIs](https://github.com/DataDog/dd-trace-js/pull/5365) are in sync with the [span APIs](https://github.com/DataDog/dd-trace-js/pull/4563). Since span APIs are legacy, they take precedence. It is still possible to add baggage using the global APIs and to not have them also added to the legacy baggages.

### DIFF
--- a/packages/dd-trace/src/opentracing/span.js
+++ b/packages/dd-trace/src/opentracing/span.js
@@ -13,7 +13,6 @@ const log = require('../log')
 const { storage } = require('../../../datadog-core')
 const telemetryMetrics = require('../telemetry/metrics')
 const { getValueFromEnvSources } = require('../config/helper')
-const { setBaggageItem, removeBaggageItem, removeAllBaggageItems } = require('../baggage')
 const SpanContext = require('./span_context')
 
 const tracerMetrics = telemetryMetrics.manager.namespace('tracers')
@@ -164,7 +163,6 @@ class DatadogSpan {
 
   setBaggageItem (key, value) {
     this._spanContext._baggageItems[key] = value
-    setBaggageItem(key, value)
     return this
   }
 
@@ -178,12 +176,10 @@ class DatadogSpan {
 
   removeBaggageItem (key) {
     delete this._spanContext._baggageItems[key]
-    removeBaggageItem(key)
   }
 
   removeAllBaggageItems () {
     this._spanContext._baggageItems = {}
-    removeAllBaggageItems()
   }
 
   setTag (key, value) {


packages/dd-trace/src/opentelemetry/context_manager.js
```diff
@@ -34,7 +44,7 @@ class ContextManager {
     }
 
     if (!activeSpan) {
-      const storedBaggageItems = storedSpan?._spanContext?._ddContext?._baggageItems
+      const storedBaggageItems = this._mergeGlobalBaggageWith(storedSpan?._spanContext?._ddContext?._baggageItems)
```
BridgeAR: This will manipulate `storedSpan?._spanContext?._ddContext?._baggageItems` and write all entries into that object. Is that intended?

packages/dd-trace/src/opentelemetry/context_manager.js
```diff
@@ -11,6 +12,15 @@ class ContextManager {
     this._store = storage('opentelemetry')
   }
 
+  _mergeGlobalBaggageWith (baggages) {
```
BridgeAR: Please use private properties or even better (the method does not need anything from the class) a pure top level function instead.


```suggestion
  #mergeGlobalBaggageWith (baggages) {
```

packages/dd-trace/src/opentelemetry/context_manager.js
```diff
@@ -86,6 +96,10 @@ class ContextManager {
     if (baggages) {
       baggageItems = baggages.getAllEntries()
     }
+    removeAllBaggageItems()
+    for (const baggage of baggageItems) {
+      setBaggageItem(baggage[0], baggage[1].value)
+    }
```
BridgeAR, line 76: Could you elaborate about this? I am unsure what this does.

ida613: This is calling the [global baggage APIs](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/baggage.js) to make sure the currently active baggage is in sync with baggage on the [context](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/opentelemetry/context_manager.js#L77). It essentially does the same thing as [these operations](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/opentelemetry/context_manager.js#L91-L94) but for global baggage

packages/dd-trace/src/opentelemetry/context_manager.js
ida613: thanks Ruben! made it a top level function

packages/dd-trace/src/opentelemetry/context_manager.js
ida613: Good call :D fixed

packages/dd-trace/src/opentelemetry/context_manager.js
BridgeAR: Should Otel baggage overrule DD baggage? Should it be merged? Do we have any documentation or RFCs around that or was there any discussion around that?

ida613: Good question. The code lets whichever comes later override, but I'm not sure if this is what's intended. @rachelyangdog @zacharycmontoya do you think we need to discuss this before proceeding?

packages/dd-trace/src/opentelemetry/context_manager.js
BridgeAR: I guess you might have forgotten to include that change? :)

ida613: apologies, here you go ^ ^

packages/dd-trace/src/opentelemetry/context_manager.js
```diff
@@ -2,10 +2,20 @@
 
 const { trace, ROOT_CONTEXT, propagation } = require('@opentelemetry/api')
 const { storage } = require('../../../datadog-core')
+const { getAllBaggageItems, setBaggageItem, removeAllBaggageItems } = require('../baggage')
 
 const tracer = require('../../')
 const SpanContext = require('./span_context')
 
+function mergeGlobalBaggageWith (baggages) {
+  const combinedBaggages = structuredClone(baggages) || {}
```
BridgeAR:
```suggestion
  const combinedBaggages = baggages ? rfdc(baggages) : {}
```

Please use rfdc. `structuredClone` is relatively slow. This must be required at the top and I believe we used a vendored rfdc. Check other usages across the repo.

In addition: we do not need to clone anything if we called `JSON.parse()` before. We could just do the cloning before we pass it to the method.

ida613: updated, thank you Ruben! this is really insightful :)

BridgeAR: Could you please move the cloning to only happen where it is needed?

ida613: is this good practice when we could also [pass an object](https://github.com/DataDog/dd-trace-js/blob/0767df5214eaa11391921450a40f36ff43113106/packages/dd-trace/src/opentelemetry/context_manager.js#L47) into it?

packages/dd-trace/src/opentelemetry/context_manager.js
ida613: so turns out dd baggage should always override otel baggage, updated the code

BridgeAR: Do you mind elaborating on how that is determined?

zacharycmontoya: It seems like we haven't properly documented this discussion internally, but most of our implementations (including Node.js) can track both the OpenTelemetry and Datadog baggage. In these cases, if there are baggage items with conflicting keys, the most recently updated item should be in the final baggage.

zacharycmontoya: I believe the only outlier tracer where we can't properly keep track of this dd-trace-go, so in that case we're opting to always prefer DD when there are conflicts

ida613: thanks Zach! that makes it easy for node then, I think the first version of my code follows this rule. I'll change it back

BridgeAR: @zacharycmontoya that sounds reasonable.

@ida613 I believe the implementation does not yet follow that. We would have to track that immediately on each individual change. I believe that is not how this works?

packages/dd-trace/src/opentelemetry/context_manager.js
BridgeAR: I am not sure I understand. I meant we should just copy that object before passing it through. That way we do not have to copy in all cases but only in one.

packages/dd-trace/src/opentelemetry/context_manager.js
ida613: @BridgeAR you are correct, I'm working on updating the code accordingly
@zacharycmontoya there are two sources of truth in node right now, one is from [legacy span apis](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/opentracing/span.js#L164-L179), the other is the [global baggage api](https://github.com/DataDog/dd-trace-js/pull/5365). I can keep the latter in sync with otel api, and use the legacy api exclusively for legacy baggage. When injecting/extracting, I can merge baggages from both legacy and global, letting legacy take precedence when there is a conflict. Does that sound like a viable solution?

zacharycmontoya: > I can keep the latter in sync with otel api, and use the legacy api exclusively for legacy baggage.

Yes I support this approach. Legacy baggage won't be included in this support. We will only keep the _Current_ OTel API Baggage in sync with the global Datadog baggage that's tracked _independently_ of spans.

> When injecting/extracting, I can merge baggages from both legacy and global, letting legacy take precedence when there is a conflict. Does that sound like a viable solution?

Did we already establish a precedence for this scenario or are we only now adding that? My preference would be to prefer global baggage, aka from the `baggage` HTTP header. How feasible would that be and do you see any issues with that preference?

ida613: > Yes I support this approach. Legacy baggage won't be included in this support. We will only keep the Current OTel API Baggage in sync with the global Datadog baggage that's tracked independently of spans.

Thank you Zach! I can do this no problem, but currently in node, dd and otel baggages only get synced when we call [active](https://github.com/DataDog/dd-trace-js/blob/015654a1a0408618aeb666fb9fb2ed47f2be41b7/packages/dd-trace/src/opentelemetry/context_manager.js#L15) or [with](https://github.com/DataDog/dd-trace-js/blob/015654a1a0408618aeb666fb9fb2ed47f2be41b7/packages/dd-trace/src/opentelemetry/context_manager.js#L77). In order to do what dotnet does (syncing on each individual operation), we would have to replace the otel baggage operations within dd-trace, which will involve a ton of work. Is it acceptable to only sync when calling `with` or `active`?

> Did we already establish a precedence for this scenario or are we only now adding that? My preference would be to prefer global baggage, aka from the baggage HTTP header. How feasible would that be and do you see any issues with that preference?

Sounds good with me, it's not hard to do. I'll work on it :)

zacharycmontoya: Yes I think it should be fine to only update when calling `active` or `with`. Let me describe the interactions I expect, but if my understanding is wrong please point it out so we're operating off the same understanding.

### DD API
Each time the user wants to read/write to the active global baggage, they call one of the Baggage API's [here](https://github.com/DataDog/dd-trace-js/blob/aa68f2976daf99f3c1a4c32fb621678272eccd74/packages/dd-trace/src/baggage.js). Notably, when clearing or setting baggage, these API's will _automatically_ update the current baggage before returning to the caller. As long as we instrument each time `active` or `with` is called in the OpenTelemetry context manager, we don't need to update any code here.

### OTel API
Each time the user wants to read/write to the active global baggage, they will need to get the current baggage by calling [getActiveBagage](https://github.com/open-telemetry/opentelemetry-js/blob/ad92be4c2c1094745a85b0b7eeff1444a11b1b4a/api/src/baggage/context-helpers.ts#L43) which gets the current context by calling [active](https://github.com/DataDog/dd-trace-js/blob/015654a1a0408618aeb666fb9fb2ed47f2be41b7/packages/dd-trace/src/opentelemetry/context_manager.js#L15). Then, once the user has updated the baggage and wants to set it as current, the user must attach the baggage to a Context using the OTel API [setBaggage](https://github.com/open-telemetry/opentelemetry-js/blob/ad92be4c2c1094745a85b0b7eeff1444a11b1b4a/api/src/baggage/context-helpers.ts#L52) and call the [api.context.with](https://github.com/open-telemetry/opentelemetry-js/blob/ad92be4c2c1094745a85b0b7eeff1444a11b1b4a/api/src/context/types.ts#L62) API with the Context and a callback, so the baggage will become active.

So with this flow, we just need our `active` operation to copy over the Datadog global baggage to overwrite all the OpenTelemetry baggage in the context and our `with` operation to copy over the OpenTelemetry baggage to override all the Datadog baggage.

And actually, this is exactly what my .NET implementation does! 😄 It just turns out that the .NET OTel Baggage API's directly set the "current" OTel baggage API within each API call (Set/Remove) so we just have to do the `active`/`with` implementation in more places

ida613: thanks Zach! that's my understanding too. I've updated the code accordingly

packages/dd-trace/src/opentelemetry/context_manager.js
ida613: I've removed this helper function in the latest commits

packages/dd-trace/src/opentelemetry/context_manager.js
ida613: @zacharycmontoya turns out legacy baggage and the new baggage are extracted/injected and propagated independently, so there will be never be a key conflict. But the legacy baggage never interact with the otel drop in code, is this okay?

zacharycmontoya: Yes that's okay 👍🏼

rochdev: The whole reason for the new API was because the global baggage is disconnected from the concept of traces and can't be tied to it, so I'm not sure why we're touching the legacy API. They operate on 2 completely different contextual hierarchies, tying them together doesn't make sense. If there is a specific problem we're trying to solve, please describe it clearly so that we can look into alternatives.

rochdev (changes requested):

rochdev (approved):

Issue #7370, opened 2026-01-28
[BUG]:  LLM Observability should not throw an unhandled exception if the agent is unavailable.


ianjkaplan:
### Tracer Version(s)

5.82

### Node.js Version(s)

24.9

### Bug Report

When instrumenting with dd-trace 5.82 with LLM Observability enabled, the library can throw an unhandled exception if connectivity to the Datadog agent is lost (or intermittently unavailable). If the exception is not caught by a top-level uncaughtException handler, it crashes the entire Node.js process.

This behavior does not feel appropriate for non-critical telemetry functionality.

Expected behavior
If the agent is unavailable, dd-trace should gracefully degrade by temporarily disabling LLM Observability until the agent becomes available again.

Actual behavior
The process can crash due to an unhandled exception originating from the LLM Observability path when the agent is unreachable.

Additional notes
If this is a desired improvement, I’m happy to submit a PR implementing this change.

### Reproduction Code

n/a The agent was temporarily unavailable. See error stack trace below.

### Error Logs

```
/app/node_modules/.pnpm/dd-trace@5.82.0_@openfeature+core@1.9.1_@openfeature+server-sdk@1.18.0_@openfeature+core@1.9.1_/node_modules/dd-trace/packages/dd-trace/src/llmobs/index.js:71
      throw new Error(
      ^

Error: Cannot send LLM Observability data without a running agent or without both a Datadog API key and site.
Ensure these configurations are set before running your application.
    at /app/node_modules/.pnpm/dd-trace@5.82.0_@openfeature+core@1.9.1_@openfeature+server-sdk@1.18.0_@openfeature+core@1.9.1_/node_modules/dd-trace/packages/dd-trace/src/llmobs/index.js:71:13
    at /app/node_modules/.pnpm/dd-trace@5.82.0_@openfeature+core@1.9.1_@openfeature+server-sdk@1.18.0_@openfeature+core@1.9.1_/node_modules/dd-trace/packages/dd-trace/src/llmobs/writers/util.js:25:7
    at /app/node_modules/.pnpm/dd-trace@5.82.0_@openfeature+core@1.9.1_@openfeature+server-sdk@1.18.0_@openfeature+core@1.9.1_/node_modules/dd-trace/packages/dd-trace/src/exporters/common/agent-info-exporter.js:15:14
[...]

PR #7457, opened 2026-02-07
fix(llmobs): gracefully handle missing API key or site configuration

ianjkaplan:
### What does this PR do?

 Instead of throwing an error when agentlessEnabled is true but API key or site is missing, log an error and disable LLM Observability. This prevents crashing user applications if the agent is misconfigured or temporarily unavailable once deployed.
### Motivation

See issue [this issue](https://github.com/DataDog/dd-trace-js/issues/7370)

### Additional Notes

This change brings the LLM Observability feature in line with the CONTRIBUTING.md rule that the tracer "should never crash a user application"

### DIFF
--- a/packages/dd-trace/src/llmobs/index.js
+++ b/packages/dd-trace/src/llmobs/index.js
@@ -67,9 +67,8 @@ function enable (config) {
   setAgentStrategy(config, useAgentless => {
     if (useAgentless && !(config.apiKey && config.site)) {
       log.error(
-        // eslint-disable-next-line eslint-rules/eslint-log-printf-style -- concatenating strings to avoid long lines
-        '[LLMObs] Cannot send LLM Observability data without a running agent or without both a Datadog API key ' +
-        'and site. Ensure these configurations are set before running your application. Disabling LLM Observability.'
+        // eslint-disable-next-line @stylistic/max-len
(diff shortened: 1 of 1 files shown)


ianjkaplan:
thanks for the feedback @BridgeAR. I can add the retry with backoff while your team discusses internally.

packages/dd-trace/src/llmobs/index.js
```diff
@@ -66,10 +66,13 @@ function enable (config) {
 
   setAgentStrategy(config, useAgentless => {
     if (useAgentless && !(config.apiKey && config.site)) {
-      throw new Error(
-        'Cannot send LLM Observability data without a running agent or without both a Datadog API key and site.\n' +
-        'Ensure these configurations are set before running your application.'
+      log.error(
+        // eslint-disable-next-line eslint-rules/eslint-log-printf-style -- concatenating strings to avoid long lines
+        '[LLMObs] Cannot send LLM Observability data without a running agent or without both a Datadog API key ' +
+        'and site. Ensure these configurations are set before running your application. Disabling LLM Observability.'
```
watson:
```suggestion
        // eslint-disable-next-line @stylistic/max-len
        '[LLMObs] Cannot send LLM Observability data without a running agent or without both a Datadog API key and site. Ensure these configurations are set before running your application. Disabling LLM Observability.'
```

sabrenner:
@ianjkaplan to @BridgeAR's point above, I'll be putting up a separate PR that accomplishes this similarly within the next week or so - will keep you updated, and thanks for your continued patience! 🙇

Issue #7074, opened 2025-12-10
[BUG]: mysql2 instrumentation causes stack overflow with large prepared statement    params


myeongseoklee:
### Tracer Version(s)

5.28.0

### Node.js Version(s)

18.19.1

### Bug Report

After calling `connection.execute()` with a large number of parameters (~3,300+), the application crashes with `Maximum call stack size exceeded`.

**Stack Trace:**

RangeError: Maximum call stack size exceeded
    at AsyncResource.runInAsyncScope (node:async_hooks:197:18)
    at Prepare.bound (node:async_hooks:235:16)
    at AsyncResource.runInAsyncScope (node:async_hooks:203:9)
    at Prepare.bound (node:async_hooks:235:16)
    ... (repeating pattern)

**Root Cause:**

In `packages/datadog-instrumentations/src/mysql2.js`, the `bindExecute` function re-wraps `onResult` on every `execute()` call (Line 108-109):

The `onResult` callback gets wrapped with `asyncResource.bind()` on every packet. MySQL's prepared statement protocol sends one packet per parameter definition. For 3,366 params, `Prepare.execute()` is called 3,369 times, nesting the callback 3,369 layers deep.

**Evidence:**
- 3,060 params → 3,063 execute() calls → ✅ OK
- 3,366 params → 3,369 execute() calls → ❌ Stack Overflow

**Suggested Fix:**
Only wrap `onResult` once by checking if already wrapped, or move wrapping to `addCommand`.

**Reproduction Repository:**
https://github.com/myeongseoklee/mysql2-callstack-reproduction

### Reproduction Code

```javascript
// Problematic pattern
const params = []; // 3,366+ params
connection.execute(
  `INSERT INTO table (...) VALUES (?, ?, ...), (?, ?, ...), ...`,
  params,
  callback
);

// Workaround: inline values, pass empty array
const values = rows.map(r => `(${escape(r.col1)}, ...)`).join(',');
connection.execute(`INSERT ... VALUES ${values}`, [], callback);
```

### Error Logs

```
RangeError: Maximum call stack size exceeded
    at AsyncResource.runInAsyncScope (node:async_hooks:197:18)
    at Prepare.bound (node:async_hooks:235:16)
    at AsyncResource.runInAsyncScope (node:async_hooks:203:9)
    at Prepare.bound (node:async_hooks:235:16)
    at AsyncResource.runInAsyncScope (node:async_hooks:203:9)
[...]

PR #7459, opened 2026-02-07
fix: wrap mysql2 addCommand only once

BridgeAR:
[fix: wrap mysql2 addCommand only once](https://github.com/DataDog/dd-trace-js/commit/77dbc12410a4d0846eb16d040e7b690d89bca77b) 

Prepare calls would currently cause the method to be rewrapped
frequently. This is now only done a single time.

In addition this adds types to the instrumentation and adds a few
more safe guards for type safety.

[perf: improve database plugin performance](https://github.com/DataDog/dd-trace-js/commit/a41ce914137ccdddd1589678389c16a68a51eb5a) 

The database plugin has a few intermediate objects that are not
necessary as well as calling methods when they do not have to be
called.
This is not a huge difference but it is not needed.

In addition, add types and make private methods actually privat.

[chore: add types to tracing and outbound plugins](https://github.com/DataDog/dd-trace-js/commit/787c2034931c5472ccedcb60efcf32dbfc299b8b)

Fixes: https://github.com/DataDog/dd-trace-js/issues/7074

### DIFF
--- a/packages/datadog-instrumentations/src/mysql2.js
+++ b/packages/datadog-instrumentations/src/mysql2.js
@@ -49,7 +49,13 @@ function wrapConnection (Connection, version) {
     }
 
     wrapCommandOnResult(command, ctx)
-    command.execute = bindExecute(command.execute, ctx)
+
+    command.execute = shimmer.wrapFunction(
+      command.execute,
+      execute => function executeWithTrace (_packet_, _connection_) {
+        return commandStartCh.runStores(ctx, execute, this, ...arguments)
+      }
+    )
 
     return commandAddCh.runStores(ctx, addCommand, this, ...arguments)
   })
@@ -145,17 +151,6 @@ function wrapConnection (Connection, version) {
 
     wrappedOnResult.set(cmd, wrapped)
     cmd.onResult = wrapped
-  }
-
-  /**
-   * @param {Function} execute
-   * @param {object} ctx
-   * @returns {Function}
-   */
-  function bindExecute (execute, ctx) {
-    return shimmer.wrapFunction(execute, execute => function executeWithTrace (packet, connection) {
-      return commandStartCh.runStores(ctx, execute, this, ...arguments)
-    })
   }
 
   /**


packages/datadog-instrumentations/src/mysql2.js
```diff
@@ -19,28 +35,33 @@ function wrapConnection (Connection, version) {
...
   })
 
   shimmer.wrap(Connection.prototype, 'query', query => function (sql, values, cb) {
     if (!startOuterQueryCh.hasSubscribers) return query.apply(this, arguments)
 
-    if (sql !== null && typeof sql === 'object') sql = sql.sql
-
-    if (!sql) return query.apply(this, arguments)
+    const resolvedSql = resolveSqlString(sql)
+    if (resolvedSql === undefined) return query.apply(this, arguments)
```
uurien, line 67: This is now assuming that sql can't be `null`

packages/datadog-instrumentations/src/mysql2.js
```diff
@@ -104,33 +126,59 @@ function wrapConnection (Connection, version) {
...
 
-  function bindExecute (execute, ctx) {
-    return shimmer.wrapFunction(execute, execute => function executeWithTrace (packet, connection) {
-      const onResult = this.onResult
+  /**
+   * @param {object} cmd
+   * @param {object} ctx
+   * @returns {void}
+   */
+  function wrapCommandOnResult (cmd, ctx) {
```
uurien, line 140: why are we extractinng this from `bindExecute`? 
In the original approach the check for `cmd.onResult` was into `execute` function, I don't know the mysql2 internals, but if it is adding a field for `cmd.onResult` into `addCommand` before the execute function when it is not set, it could be undefined when this is executed and it would have some value when `execute`.

packages/datadog-instrumentations/src/mysql2.js
BridgeAR: Why would it ever be `null`? The only reason to check for it like this is to guard against driver changes, if I am not mistaken. If that is changed, it would normally just be `undefined`. I can also catch `null`, it just seems unlikely that this would be needed.

packages/datadog-instrumentations/src/mysql2.js
BridgeAR: It is just a small helper method that is called `bindExecute`. I guess the name might imply other things, so I am going ahead and inline the method again.

packages/datadog-instrumentations/src/mysql2.js
uurien: It is not probable, and I think we can leave at it is now, but in the same way that it could be `undefined` in a wrong usage of the method, it could be `null`. Is a user who can use it incorrectly, sending null in the query object. `connection.query({ sql: null, ... })`

packages/datadog-instrumentations/src/mysql2.js
uurien: But you have changed also the order of how this is executed.
Previous behavior:
addCommand => execute (beforeExecute, check `cmd.onResult`)

Current behavior:
addCommand (beforeAddCommand, check `cmd.onResult`) => execute

Anyway, I'm checking the mysql2 source code and doesn't seem that it affects, so we can resolve this comment.

packages/datadog-instrumentations/src/mysql2.js
BridgeAR: Good point. In theory, we may only accept strings, which we formerly also did not. We are able to handle almost all types (all types that work with `str + sql` coercion). The question is: should we track these user mistakes? Should we just ignore them? I feel users might want to know about these mistakes?

packages/datadog-instrumentations/src/mysql2.js
```diff
@@ -141,11 +184,14 @@ function wrapConnection (Connection, version) {
             finishCh.runStores(ctx, onResult, this, ...arguments)
           })
         } else {
-          this.once(errorMonitor, error => {
-            ctx.error = error
-            errorCh.publish(ctx)
-          })
-          this.once('end', () => finishCh.publish(ctx))
+          const command = /** @type {{ once?: Function }} */ (this)
```
bengl, line 187: Does TS not provide a nice way to do this, say on line 164, so that you don't need to do this kinda weird re-assignment?

packages/datadog-instrumentations/src/mysql2.js
```diff
@@ -206,21 +256,22 @@ function wrapPool (Pool, version) {
...
       if (typeof values === 'function') {
         cb = values
       }
 
-      process.nextTick(() => {
-        cb(abortController.signal.reason)
-      })
+      if (typeof cb === 'function') {
+        process.nextTick(() => {
+          /** @type {Function} */ (cb)(abortController.signal.reason)
```
bengl, line 272: Does line 270 not make this type comment unnecessary?

packages/datadog-instrumentations/src/mysql2.js
```diff
@@ -305,9 +360,21 @@ function wrapPoolCluster (PoolCluster) {
   return PoolCluster
 }
 
-addHook({ name: 'mysql2', file: 'lib/base/connection.js', versions: ['>=3.11.5'] }, wrapConnection)
-addHook({ name: 'mysql2', file: 'lib/connection.js', versions: ['1 - 3.11.4'] }, wrapConnection)
-addHook({ name: 'mysql2', file: 'lib/pool.js', versions: ['1 - 3.11.4'] }, wrapPool)
+addHook(
+  { name: 'mysql2', file: 'lib/base/connection.js', versions: ['>=3.11.5'] },
+  /** @type {(moduleExports: unknown, version: string) => unknown} */ (wrapConnection)
```
bengl, line 365: This is pretty repetitive. Can we not make this implicit via `addHooks`'s definition?

BridgeAR: This is already the definition in addHook. The reason is that the receiving method defines more concrete types that do not match these 1-to-1. Thus the need for the cast.

packages/datadog-instrumentations/src/mysql2.js
BridgeAR: Somehow the types detect that the cb might have changed its content due to the async operation in between.

packages/datadog-instrumentations/src/mysql2.js
BridgeAR: Yes, the issue here is something else: we actually truly do not know about the type because they might change in any version. Due to that, we have to 'detect' each individual type and even when we detect it, we have to cast that part so that this access is accepted.

packages/datadog-instrumentations/src/mysql2.js
BridgeAR: Briefly discussed in the guild channel: we likely want to track all types. This would accommodate that mostly.

packages/dd-trace/src/plugins/database.js
```diff
@@ -7,71 +7,57 @@ class DatabasePlugin extends StoragePlugin {
...
-    const { encodedDddb, encodedDddbs, encodedDde, encodedDdh, encodedDdps, encodedDdpv } = this.serviceTags
+  /**
+   * @param {string} serviceName
+   * @param {import('../../../..').Span} span
+   * @param {object} peerData
+   * @returns {string}
+   */
+  #createDBMPropagationCommentService (serviceName, span, peerData) {
+    const spanTags = span.context()._tags
+    const encodedDddb = encode(spanTags['db.name'])
```
uurien, line 18: aren't you changing the behavior here? is intentional?
in the original behavior, when `spanTags['db.name']` was `dddb`, the value for `encodedDddb` would be `''`, and here, it would be `dddb`.

BridgeAR: It was an intentional change. I believe the if statement is comparing wrong values and it was potentially meant to prevent encoding something that is already encoded.

PR #7460, opened 2026-02-07
test: fix flaky confluenctinc kafka error

BridgeAR:
The broker needs some time after creating a new topic to settle in. I was able to run it 400 times in a row without error while it could theoretically still fail in some rare circumstances.

I also removed the old commented error test case. It could not work like that due to
expecting to catch a user error (which we should not).

### DIFF
--- a/packages/datadog-plugin-confluentinc-kafka-javascript/test/index.spec.js
+++ b/packages/datadog-plugin-confluentinc-kafka-javascript/test/index.spec.js
@@ -165,17 +165,11 @@ describe('Plugin', () => {
                 })
               }))
 
-              await withTimeout(
+              await Promise.all([
                 sendMessages(kafka, testTopic, messages),
-                20000,
-                `Timeout: Did not produce message on topic "${testTopic}" within 20000ms`
-              )
-              await withTimeout(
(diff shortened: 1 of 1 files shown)


packages/datadog-plugin-confluentinc-kafka-javascript/test/index.spec.js
```diff
@@ -153,15 +157,23 @@ describe('Plugin', () => {
...
                   eachMessage: () => {
                     resolve()
                   },
                 })
-              })
-              await sendMessages(kafka, testTopic, messages).then(
-                async () => await consumerReceiveMessagePromise
+              }))
+
+              await withTimeout(
```
pabloerhard: Doubt: Why is doing this with 'withTimeout' needed if the topic is already being run with 'waitForReady'? Is it just for safety / to prevent hanging?

BridgeAR: I just ran it again locally and you seem to be right. This is likely not needed

pabloerhard (approved):
LGTM!

PR #7465, opened 2026-02-09
fix: reduce ACF overhead in storage and noop contexts

BridgeAR:
Replace manual enterWith/restore patterns with run() across all noop context wrappers (log writer, exporters, git utils). This halves AsyncContextFrame transitions on Node.js 22+: run() restores via AsyncContextFrame.set() (no extra frame), while enterWith/restore created a second frame that triggered the profiler's setContext() again.

In DatadogStorage, feature-detect whether ALS.run() delegates to enterWith() (ACF behavior) and only fall back to the manual approach on older runtimes without AsyncContextFrame.

### DIFF
--- a/packages/dd-trace/test/appsec/iast/analyzers/path-traversal-analyzer.spec.js
+++ b/packages/dd-trace/test/appsec/iast/analyzers/path-traversal-analyzer.spec.js
@@ -15,7 +15,6 @@ const { newTaintedString } = require('../../../../src/appsec/iast/taint-tracking
 
 const { prepareTestServerForIast } = require('../utils')
 const { HTTP_REQUEST_PARAMETER } = require('../../../../src/appsec/iast/taint-tracking/source-types')
-const { isACFActive } = require('../../../../../datadog-core/src/storage')
 
 const iastContext = {
   rootSpan: {
@@ -500,7 +499,7 @@ prepareTestServerForIast('integration test', (testThatRequestHasVulnerability, t
         const callArgs = [fsAsyncWayMethodPath]
         callArgs[0] = newTaintedString(iastCtx, callArgs[0], 'param', 'Request')
         return require(fsAsyncWayMethodPath).doubleCallIgnoringCb('stat', callArgs)
-      }, 'PATH_TRAVERSAL', { occurrences: isACFActive ? 1 : 2 })
+      }, 'PATH_TRAVERSAL', { occurrences: 2 })
     })
   })
 
--- a/packages/dd-trace/src/appsec/iast/path-line.js
+++ b/packages/dd-trace/src/appsec/iast/path-line.js
@@ -22,6 +22,7 @@ const EXCLUDED_PATH_PREFIXES = [
   'child_process',
   'node:async_hooks',
   'async_hooks',
+  'node:internal/async_local_storage',
 ]
 
 function getNonDDCallSiteFrames (callSiteFrames, externallyExcludedPaths) {


packages/datadog-core/src/storage.js
```diff
@@ -59,28 +59,46 @@ class DatadogStorage extends AsyncLocalStorage {
       return stores.get(handle)
     }
   }
+}
+
+// To handle all versions always correct, feature detect AsyncContextFrame and
+// fallback to manual approach if not active.
```
szegedi: I'd add a comment here explaining why this detection works: that with ACF, `run` delegates to `enterWith`, and without ACF it does not. It might be an implementation detail, and it might change in the future (hopefully it doesn't, 'cause the implementation is actually better that way) so leaving a breadcrumb here for anyone trying to understand the logic seems like a good idea.

packages/datadog-core/src/storage.js
```diff
@@ -59,28 +59,46 @@ class DatadogStorage extends AsyncLocalStorage {
...
+const isACFActive = (() => {
+  let active = false
+  const als = new AsyncLocalStorage()
+  const orig = als.enterWith
+  als.enterWith = () => { active = true }
+  als.run(1, () => {})
+  als.enterWith = orig
```
szegedi: I think it's a good idea to call `als.disable()`. Also, we don't need to bother with restoring original `enterWith` value, since this als instance is not reachable/usable from anywhere else after this method returned.

```suggestion
  const als = new AsyncLocalStorage()
  als.enterWith = () => { active = true }
  als.run(1, () => {})
  als.disable()
```

packages/datadog-core/src/storage.js
BridgeAR: I explained it in the `run` part below. Shall I duplicate that or move it (I explained it implicitely)?

BridgeAR: I added a short description here as well.

packages/dd-trace/src/profiling/config.js
```diff
@@ -219,16 +222,8 @@ class Config {
       that.asyncContextFrameEnabled = false
     }
 
-    const hasExecArg = (arg) => process.execArgv.includes(arg) || String(NODE_OPTIONS).includes(arg)
+    const canUseAsyncContextFrame = samplingContextsAvailable && isACFActive
```
rochdev, line 225: Not a fan of exposing `isACFActive` as that should be a hidden implementation detail, but it seems better than the current alternative.

packages/dd-trace/test/appsec/iast/analyzers/path-traversal-analyzer.spec.js
```diff
@@ -499,7 +500,7 @@ prepareTestServerForIast('integration test', (testThatRequestHasVulnerability, t
         const callArgs = [fsAsyncWayMethodPath]
         callArgs[0] = newTaintedString(iastCtx, callArgs[0], 'param', 'Request')
         return require(fsAsyncWayMethodPath).doubleCallIgnoringCb('stat', callArgs)
-      }, 'PATH_TRAVERSAL', { occurrences: 2 })
+      }, 'PATH_TRAVERSAL', { occurrences: isACFActive ? 1 : 2 })
```
simon-id: we think this is actually a regression, not a fix :/

packages/datadog-core/src/storage.js
```diff
@@ -59,28 +59,46 @@ class DatadogStorage extends AsyncLocalStorage {
...
+// fallback to manual approach if not active. With ACF `run` delegates to
+// `enterWith`, without ACF `run` does not.
+const isACFActive = (() => {
+  let active = false
+  const als = new AsyncLocalStorage()
+  als.enterWith = () => { active = true }
+  als.run(1, () => {})
+  als.disable()
+  return active
+})()
```
rochdev, line 74: Doing this right at file load time can have side-effects, but I think that's fine because if this file is loaded then the rest of the library is probably loaded as well at which point we'll end up getting those side-effects anyway.

rochdev (approved):
LGTM! Added a few comments but mostly just as notes.

packages/dd-trace/src/profiling/config.js
BridgeAR: Exactly my thoughts :D

PR #7469, opened 2026-02-09
fix(ws): avoid retaining connection span

BridgeAR:
Run ws socket setup in a store scope without the connection span so internal ws handlers don't capture and keep it alive.

### DIFF
--- a/packages/datadog-plugin-ws/src/server.js
+++ b/packages/datadog-plugin-ws/src/server.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const { storage } = require('../../datadog-core')
 const TracingPlugin = require('../../dd-trace/src/plugins/tracing.js')
 const tags = require('../../../ext/tags.js')
 const { HTTP_HEADERS } = require('../../../ext/formats')
@@ -23,9 +22,7 @@ class WSServerPlugin extends TracingPlugin {
 
     // Bind the setSocket channel so internal ws event handlers (data, close)
     // don't capture their async context.
-    this.addBind('tracing:ws:server:connect:setSocket', () => {
-      return
-    })
+    this.addBind('tracing:ws:server:connect:setSocket', () => {})
   }
 
   bindStart (ctx) {


packages/datadog-plugin-ws/src/server.js
```diff
@@ -17,6 +18,21 @@ class WSServerPlugin extends TracingPlugin {
...
 
+  constructor (...args) {
+    super(...args)
+
+    // Bind the setSocket channel so internal ws event handlers (data, close)
+    // don't capture the connection span in their async context.
+    this.addBind('tracing:ws:server:connect:setSocket', () => {
+      const store = storage('legacy').getStore()
+      if (store?.span) {
+        const { span: _, ...storeWithoutSpan } = store
```
rochdev: Why keep the rest of the store? It seems like this could still leak at that point. If we want these events to be in the root context, that should be the case instead of creating a new child context.

BridgeAR: Fixed. I just have a noop function now to always return undefined as context.

packages/datadog-plugin-ws/src/server.js
```diff
@@ -17,6 +17,14 @@ class WSServerPlugin extends TracingPlugin {
   static get type () { return 'websocket' }
   static get kind () { return 'request' }
 
+  constructor (...args) {
+    super(...args)
+
+    // Bind the setSocket channel so internal ws event handlers (data, close)
+    // don't capture their async context.
+    this.addBind('tracing:ws:server:connect:setSocket', () => {})
```
rochdev, line 25: Nit, but I think this would be a case where explicitly returning `undefined` makes the intent clearer.

packages/datadog-plugin-ws/test/index.spec.js
```diff
@@ -65,6 +67,29 @@ describe('Plugin', () => {
           agent.close({ ritmReset: false, wipe: true })
         })
 
+        it('should not retain the connection span during socket setup', async () => {
```
rochdev, line 70: Nit: I'd probably reword this, as that's actually only a side effect. If a span shouldn't be active, it means that those handlers were previously in the wrong context, so that's the real problem that the test is testing for.

packages/datadog-plugin-ws/src/server.js
BridgeAR: Our linter prohibits that. I would have to add an exception for that which is also meh

PR #7470, opened 2026-02-09
fix(startup-log): emit immediately after init and send to stderr

bm1549:
## Summary
Changes startup logs to emit immediately after tracer initialization and send output to stderr instead of stdout.

## Motivation
1. **Immediate emission**: Previously, startup logs were only emitted on the first payload sent to the agent, which could be delayed or never happen if the agent is unreachable. Emitting immediately after init ensures users always see the logs.

2. **stderr output**: Following Unix conventions and aligning with other Datadog language tracers, diagnostic/log output should go to stderr rather than stdout. This allows proper separation of application output from diagnostic information.

## Changes
- Added `startupLog()` call in `proxy.js` after tracer initialization completes
- Changed output from `info()` to `error()` to use stderr instead of stdout
- Removed duplicate late call in `writer.js` that was firing on first payload
- Updated all tests to expect `console.error` instead of `console.info`

## Test plan
- [x] All startup log tests pass with updated expectations
- [x] Tests verify logs are emitted immediately, not delayed

## Related PR
This PR is part of a larger effort split into two PRs:
- This PR: Behavioral changes (timing and output stream)
- #7474: Enable startup logs by default (semver-major)

### DIFF
--- a/packages/dd-trace/src/startup-log.js
+++ b/packages/dd-trace/src/startup-log.js
@@ -4,7 +4,7 @@ const os = require('os')
 const { inspect } = require('util')
 const tracerVersion = require('../../../package.json').version
 const { getAgentUrl } = require('./agent/url')
-const { error, warn } = require('./log/writer')
+const { warn } = require('./log/writer')
 
 const errors = {}
 let config
@@ -25,7 +25,7 @@ function startupLog () {
   startupLogRan = true
 
   const out = tracerInfo()
-  error('DATADOG TRACER CONFIGURATION - ' + out)
+  warn('DATADOG TRACER CONFIGURATION - ' + out)
 }
 
 /**
--- a/packages/dd-trace/test/startup-log.spec.js
+++ b/packages/dd-trace/test/startup-log.spec.js
@@ -17,7 +17,6 @@ describe('startup logging', () => {
   let tracerInfoMethod
 
   before(() => {
-    sinon.stub(console, 'error')
     sinon.stub(console, 'warn')
     delete require.cache[require.resolve('../src/startup-log')]
     const {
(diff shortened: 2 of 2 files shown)


packages/dd-trace/src/proxy.js
```diff
@@ -288,6 +288,8 @@ class Tracer extends NoopProxy {
       this._pluginManager.configure(config)
       DynamicInstrumentation.configure(config)
       setStartupLogPluginManager(this._pluginManager)
+      // Emit startup log immediately after tracer is fully initialized
+      startupLog()
```
BridgeAR, line 292: This removes the error handling. I am fine with that, while it a) removes information that the customer might want and b) if we remove it, we also have to remove the agentError part. I believe that was actually spected somewhere.

bm1549: Can you elaborate? I had expected that this would just front-load information we were already providing, not remove information

Is there a different place you'd recommend that this go to accomplish both goals?

BridgeAR: `agentError` came from the information being passed through to the method when being called https://github.com/DataDog/dd-trace-js/pull/7470/changes#diff-a3fcd7c24d4ec8042528afc821a9e9541c66809f4f79d2371274bef28f955342L34

I would have considered this to be part of the initialization of the tracer (the first payload to the agent is during startup because it checks for the agent to be available or not and what the agent supports. Without that, the tracer might not know how to work).

Thus, I would have expected the former placement to be correct.

Are you certain the agent support is not checked in other tracers during startup for the startup log?

bm1549: This is something other tracers have, but it's de-coupled from the initialization. I'll re-work slightly to have two separate log lines as other tracers do

BridgeAR: If we decouple it, we should probably log to stdout, not err (due to not having an error anymore)

bm1549: I'll have it as warn instead since that also goes to stderr, but strongly believe it should not go to stdout. Reason being that going to stdout, especially by default, can mess with any tooling that may be processing the logs

For example, imagine someone uses SSI, which instruments everything running Node.js. If they were to run a small Node.js script and depend on the stdout of that script for further processing, this may break their tool

bm1549: Discussed offline with @BridgeAR 

1. stderr is the convention for diagnostic logs coming from outside of the primary application. We will follow it here
2. agent error was re-introduced and will be logged separately, in the same conditions as before

rochdev: Isn't this too early? We weren't waiting just for the errors, we were also waiting to give time to the app to load any dependencies it's going to use, otherwise we lose the information about loaded integrations (which is one of the most important aspects of startup logs). Unless I'm misunderstanding the PR, I think this is broken now.

PR #7472, opened 2026-02-10
add rewriter support for generator functions

rochdev:
### What does this PR do?

Add rewriter support for generator functions.

### Motivation

They can't be patched using the current available functionality.

### DIFF
--- a/packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
+++ b/packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
@@ -298,16 +298,16 @@ function wrapIterator (state, node, program) {
         return iter.then(result => {
           ctx.result = result;
 
-          asyncStart.publish(ctx);
-          asyncEnd.publish(ctx);
+          ${channelVariable}.asyncStart.publish(ctx);
+          ${channelVariable}.asyncEnd.publish(ctx);
 
           return wrap(result);
         }, err => {
           ctx.error = err;
 
-          error.publish(ctx);
-          asyncStart.publish(ctx);
-          asyncEnd.publish(ctx);
+          ${channelVariable}.error.publish(ctx);
+          ${channelVariable}.asyncStart.publish(ctx);
+          ${channelVariable}.asyncEnd.publish(ctx);
 
           return Promise.reject(err);
         });


packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
```diff
@@ -194,3 +230,42 @@ function wrapCallback (state, node) {
...
+
+      if (!${channelVariable}.start.hasSubscribers) return __apm$traced();
+
+      {
+        const ctx = {
+          arguments,
+          self: this,
+          moduleVersion: "1.0.0"
+        };
+        const gen = ${channelVariable}.traceSync(__apm$traced, ctx);
```
bengl: I was going to suggest that this should be `${traceMethod}`, but then I remembered that async generators return synchronously, despite being called `async`. 😆

packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
```diff
@@ -133,6 +145,30 @@ function wrap (state, node) {
...
+  traverse(
+    node.body,
+    '[object.type=Super]',
+    (node, parent) => {
+      const method = node.property.name
+      const { callee } = parent
+
+      if (callee) {
+        const expression = parse(`
+          Reflect.get(Object.getPrototypeOf(this.constructor.prototype), '${method}').call(this)
```
bengl: You may as well be consistent and use `Reflect.getPrototypeOf` here.

Also, can you leave a comment explaining why `Reflect.get` is necessary?

packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
rochdev: Yeah I actually did the same thing and wrote it wrong at first lol.

packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
bengl: Alright, I see you've added the comment, it it makes sense, but it doesn't explain why `Reflect.get` in particular is necessary.

rochdev: Wasn't really needed in this case so I just removed it.

packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
```diff
@@ -194,3 +233,63 @@ function wrapCallback (state, node) {
...
+        return __apm$wrapped.apply(this, arguments);
+      };
+
+      if (!${channelVariable}.start.hasSubscribers) return __apm$traced();
+
+      {
+        const wrap = it => {
+          const { next: itNext, return: itReturn, throw: itThrow } = it;
+
+          it.next = (...args) => ${traceNext}(itNext, ctx, it, ...args);
```
bengl: Nit: I usually prefer `iter` as an instance of an iterator, but that's not a blocker.

bengl (dismissed):

packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
rochdev: I need to fix `wrapSuper` so I'll do this at the same time.

rochdev: Fixed `wrapSuper` and renamed the variable.

packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
```diff
@@ -194,3 +256,67 @@ function wrapCallback (state, node) {
...
+          asyncStart.publish(ctx);
+          asyncEnd.publish(ctx);
+
+          return wrap(result);
+        }, err => {
+          ctx.error = err;
+
+          error.publish(ctx);
+          asyncStart.publish(ctx);
+          asyncEnd.publish(ctx);
```
crysmags:
```suggestion
          ${channelVariable}.asyncStart.publish(ctx);
          ${channelVariable}.asyncEnd.publish(ctx);

          return wrap(result);
        }, err => {
          ctx.error = err;

          ${channelVariable}.error.publish(ctx);
          ${channelVariable}.asyncStart.publish(ctx);
          ${channelVariable}.asyncEnd.publish(ctx);
```
we need to add the channel names

PR #7474, opened 2026-02-10
feat(config): enable startup logs by default

bm1549:
### What does this PR do?
Enable startup logs by default for dd-trace v6+, gated behind a `DD_MAJOR >= 6` version check so the change is safely backportable to v5.

### Motivation
Startup logs provide helpful diagnostic information about tracer configuration. Currently they must be explicitly enabled via `DD_TRACE_STARTUP_LOGS=true`. By enabling them by default, we make it easier for users to diagnose configuration issues and understand their tracer setup.

Main benefits:

1. Installing our libraries is hard for some users and is error-prone. Providing an on-by-default log as a heartbeat gives folks the confidence that the SDK was installed.
2. Changing configs is often the second battle. Having the startup log off-by-default with a config to turn it on effectively couples installing and configuring. Users can end up not knowing if the tracer isn't installed or if the config isn't being set properly.
3. Moving to stderr (#7502) makes this much less risky since other dependencies may also log to stderr.
4. Consistency with other languages. Nearly every other Datadog tracer has logging to console on by default, confirmed by https://github.com/DataDog/system-tests/pull/6241

### Additional Notes
- The default change is gated behind `DD_MAJOR >= 6` in `defaults.js` so it is safely backportable (v5 retains `startupLogs: false`)
- `supported-configurations.json` keeps the canonical default as `false` (implementation level `D`), with the runtime override in `defaultsWithConditionalRuntimeBehavior`
- Users who want to opt out can set `DD_TRACE_STARTUP_LOGS=false`

**Test plan:**
- [x] Config defaults test updated and passing (197/197)
- [x] RASP integration tests set `DD_TRACE_STARTUP_LOGS=false` to avoid interference

### DIFF
--- a/packages/dd-trace/test/config/index.spec.js
+++ b/packages/dd-trace/test/config/index.spec.js
@@ -578,7 +578,7 @@ describe('Config', () => {
       { name: 'spanAttributeSchema', value: 'v0', origin: 'default' },
       { name: 'spanComputePeerService', value: false, origin: 'calculated' },
       { name: 'spanRemoveIntegrationFromService', value: false, origin: 'default' },
-      { name: 'startupLogs', value: true, origin: 'default' },
+      { name: 'startupLogs', value: DD_MAJOR >= 6, origin: 'default' },
       { name: 'stats.enabled', value: false, origin: 'calculated' },
       { name: 'tagsHeaderMaxLength', value: 512, origin: 'default' },
       { name: 'telemetry.debug', value: false, origin: 'default' },


integration-tests/appsec/index.spec.js
```diff
@@ -37,6 +37,7 @@ describe('RASP', () => {
           DD_APPSEC_RASP_ENABLED: 'true',
           DD_APPSEC_RULES: path.join(cwd, 'appsec/rasp/rasp_rules.json'),
           DD_APPSEC_RASP_COLLECT_REQUEST_BODY: String(collectRequestBody),
+          DD_TRACE_STARTUP_LOGS: 'false',
```
bm1549, line 40: This appsec test makes sure that no logs exist when blocking a request. Rather than patching it to ignore the DD startup log, this seemed more correct

packages/dd-trace/test/config/index.spec.js
```diff
@@ -578,7 +578,7 @@ describe('Config', () => {
       { name: 'spanAttributeSchema', value: 'v0', origin: 'default' },
       { name: 'spanComputePeerService', value: false, origin: 'calculated' },
       { name: 'spanRemoveIntegrationFromService', value: false, origin: 'default' },
-      { name: 'startupLogs', value: false, origin: 'default' },
+      { name: 'startupLogs', value: true, origin: 'default' },
```
BridgeAR: This also needs to check for the version being used, otherwise it would fail in the release branch.

bm1549: Ya know I was confused about how this was working. Locally (and in CI) it was running as v6 so I thought tests always ran for v6 and, when I tried leaving it, it failed

Fixed it up!

PR #7489, opened 2026-02-11
fix(test): Node 24 deprecation warnings and url.parse refactors

watson:
### What does this PR do?

- **Test setup:** Add `url.parse()` (DEP0169) and `punycode` (DEP0040) to the global deprecation warning exceptions in `packages/dd-trace/test/setup/core.js` so tests do not fail on Node.js 24.13.1+.
- **express-session spec:** Add a middleware that registers the url.parse deprecation as a temporary warning exception before the express-session middleware runs, so the warning emitted when express-session@1.5.0 calls `url.parse()` during request handling does not fail the test.
- **Refactors:** Replace deprecated `url.parse()` with the WHATWG `URL` API in test/helper code:
  - `packages/datadog-plugin-microgateway-core/test/proxy.js` — use `new URL()` and `node:` core requires.
  - `packages/datadog-plugin-next/test/server.js` and `packages/dd-trace/test/appsec/next/pages-dir/server.js` — parse request URLs with `new URL()` and pass a URL-like object to the Next request handler.

### Motivation

Node.js 24.13.1+ triggers these deprecation warnings in more code paths (e.g. url instrumentation, express-session, dependencies using punycode). The test framework treats unhandled deprecation warnings as failures. Adding the warnings to the global exceptions list keeps tests green; the refactors remove `url.parse()` from test/helper code where it is under our control.

### DIFF
--- a/packages/datadog-instrumentations/test/url.spec.js
+++ b/packages/datadog-instrumentations/test/url.spec.js
@@ -5,7 +5,6 @@ const { describe, it, beforeEach, afterEach, before, after } = require('mocha')
 const sinon = require('sinon')
 
 const agent = require('../../dd-trace/test/plugins/agent')
-const { temporaryWarningExceptions, URL_PARSE_DEPRECATION } = require('../../dd-trace/test/setup/core')
 const { channel } = require('../src/helpers/instrument')
 const names = ['url', 'node:url']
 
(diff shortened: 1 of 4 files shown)


watson:
I have a feeling we'll see a 24.13.2 soon reverting this deprecation change, as it might be considered a breaking change

PR #7491, opened 2026-02-11
[test optimization] Fix quarantine + EFD

juan-fernandez:
### What does this PR do?

Fix some edge cases where quarantined tests run with EFD would result in us not correctly ignoring quarantined test failures.

### Motivation
Make sure quarantine and EFD work correctly.


packages/dd-trace/src/plugins/util/test.js
```diff
@@ -573,7 +573,7 @@ function getTestSuitePath (testSuiteAbsolutePath, sourceRoot) {
     ? testSuiteAbsolutePath
     : path.relative(sourceRoot, testSuiteAbsolutePath)
 
-  return testSuitePath.replace(path.sep, '/')
+  return testSuitePath.replaceAll(path.sep, '/')
```
juan-fernandez, line 576: unrelated improvement

integration-tests/jest/jest.spec.js
```diff
@@ -5460,6 +5460,136 @@ describe(`jest@${JEST_VERSION} commonJS`, () => {
...
+
+            // Session should pass because the only failing test is quarantined
+            assert.strictEqual(testSession.meta[TEST_STATUS], 'pass')
+            assert.strictEqual(testSession.meta[TEST_MANAGEMENT_ENABLED], 'true')
+
+            // All executions of the quarantined test should be tagged as quarantined
+            const quarantinedTests = tests.filter(
+              test => test.meta[TEST_NAME] === 'quarantine tests can quarantine a test'
+            )
+            assert.ok(quarantinedTests.length > 1, 'quarantined test should have been retried by ATR')
```
calvinbayer, line 5484: Unrelated to this PR, but what's the motivation of retrying quarantined tests? Does this bring any value?

juan-fernandez: It brings value, but one could argue that it might not be enough for the cost. The value is getting a stronger signal about how the test behaves

PR #7494, opened 2026-02-11
Only add host tag to profiles when DD_TRACE_REPORT_HOSTNAME is set

szegedi:
### What does this PR do?
Only adds the `host:` tag to profiles when `DD_TRACE_REPORT_HOSTNAME` is set to true.

### Motivation
Node.js profiler unconditionally emits host tag with the value of `os.hostname()`. This causes issues for serverless usage, where host names are ephemeral and unnecessarily increase cardinality of various lists and metrics.

The agent also adds a host tag when necessary, so the profiler needn’t do it. There is also a `DD_TRACE_REPORT_HOSTNAME` env var that regulates whether the tracer includes host name in the traces, it defaults to false. We changed the logic so we only emit the host tag in profiles if the env var is explicitly set to true.

Jira: [PROF-13728](https://datadoghq.atlassian.net/browse/PROF-13728)

### DIFF
--- a/packages/dd-trace/src/profiling/config.js
+++ b/packages/dd-trace/src/profiling/config.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const os = require('os')
 const path = require('path')
 const { pathToFileURL } = require('url')
 const satisfies = require('../../../../vendor/dist/semifies')
@@ -50,7 +49,6 @@ class Config {
       DD_TAGS,
     } = getProfilingEnvValues()
 
-    const host = os.hostname()
     // Must be longer than one minute so pad with five seconds
     const flushInterval = options.interval ?? (Number(DD_PROFILING_UPLOAD_PERIOD) * 1000 || 65 * 1000)
     const uploadTimeout = options.uploadTimeout ?? (Number(DD_PROFILING_UPLOAD_TIMEOUT) || 60 * 1000)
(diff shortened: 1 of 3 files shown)


packages/dd-trace/src/profiling/config.js
```diff
@@ -71,7 +65,7 @@ class Config {
       tagger.parse(options.tags),
       tagger.parse({
         env: options.env,
-        host,
+        host: options.reportHostname ? require('os').hostname() : undefined,
```
BridgeAR, line 68: Nit: 
Ideally we create an object before and add `host` as property only if hostname is available. I guess loading os here and reading it lazy is fine, since we should normally just construct this once.

szegedi: `tagger.parse` discards key-value pairs where the value is non-strictly equal to null.

szegedi: I'm not actually sure what is it you're suggesting as a change here.

BridgeAR: It would have been about not needing the property, no matter that it is removed later. I personally like to prevent overhead and have shapes staying stable.

That is why I marked it as nit. It is nothing that needs addressing. Let's merge it as is. This is not going to be called more than once normally.

PR #7495, opened 2026-02-11
fix(prisma): resolve db config from adapters

pabloerhard:
fix(prisma): resolve db config from adapters

Prisma v7 can surface db info via datasource names or adapters.
Resolve the runtime datasource name and prefer adapter config,
then client overrides, then runtime/env fallback so tags stay
correct across pg/mariadb/mssql.
Parse sqlserver:// strings with semicolon keys to recover
db/user/host/port reliably.

perf(prisma): use parent child relation for resolving nested spans

Engine spans now build a parent->children map and start at
roots so nested spans are emitted without rescanning.
Short-circuit tracing when there are no subscribers and ignore
empty span lists to avoid unnecessary work.

test(prisma): cover v7 adapters and variants

Add pg field config plus mariadb/mssql adapter fixtures so
adapter parsing paths are exercised across providers.
Servers toggle between connection string and field configs via
env flags to hit both code paths with minimal scripts.
Integration harness now selects a single import variant per
config and asserts the adapter-specific db span name.
Update prisma schemas to use a named datasource and the test
env var to validate runtime datasource resolution.
Register adapter deps in plugin externals and versions, and
let varySandbox accept a binding name plus package override.

test(prisma): stabilize mongodb integration

Run mongo as a replica set and wait for primary readiness so
Prisma's connector can attach consistently.
Reset the database with the driver before each run to prevent
state leakage across sandboxes.
Skip migrate reset for mongo and generate unique emails to
avoid duplicate key errors.

test(mongo): make waitForMongo replset-aware

When the configured URL requests a replica set, initialize it
and wait until a primary is elected before proceeding.
Keep mongodb-core for the lightweight connectivity check and
use the driver only for replset setup.

chore: make externals.json a js file to allow commenting

Next to allowing for comments a `forced` entry was added which
means that library will be installed as dependency. In addition, the
install script now does not rewrite the package.json per dependency
anymore. Instead, it gathers the entries up front and writes it once.

chore(prisma): add types to the prisma instrumentation

### DIFF
--- a/packages/datadog-plugin-prisma/test/integration-test/client.spec.js
+++ b/packages/datadog-plugin-prisma/test/integration-test/client.spec.js
@@ -62,8 +62,8 @@ const prismaClientConfigs = [{
   importPath: '@prisma/client',
   variant: 'default',
   env: {
-    PRISMA_TEST_DATABASE_URL: TEST_DATABASE_URL
-  }
+    PRISMA_TEST_DATABASE_URL: TEST_DATABASE_URL,
+  },
 },
 {
   name: 'prisma-generator-js with custom output',
--- a/.github/workflows/apm-integrations.yml
+++ b/.github/workflows/apm-integrations.yml
@@ -1133,7 +1133,7 @@ jobs:
         ports:
           - 1433:1433
       mongodb:
-        image: circleci/mongo
+        image: prismagraphql/mongo-single-replica:4.4.3-bionic
         ports:
           - 27017:27017
       mysql:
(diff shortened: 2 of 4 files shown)


packages/datadog-plugin-prisma/test/integration-test/client.spec.js
```diff
@@ -16,49 +16,206 @@ const {
...
+      'db.type': 'mssql',
+    },
+  },
 }]
 
 describe('esm', () => {
   let agent
   let proc
   prismaClientConfigs.forEach(config => {
+    // if (!config.name.includes('prisma-generator v7 mssql adapter (url)')) return
```
BridgeAR, line 209:
```suggestion
```

packages/dd-trace/test/setup/services/mssql.js
```diff
@@ -1,10 +1,11 @@
 'use strict'
 
 const RetryOperation = require('../operation')
-const tedious = require('../../../../../versions/tedious').get()
 
-function waitForMssql () {
-  return new Promise((resolve, reject) => {
+function waitForMssql (isSandbox) {
+  const tedious = isSandbox ? require('tedious') : require('../../../../../versions/tedious').get()
```
BridgeAR, line 6: ❤️  thank you for finding this!

BridgeAR (approved):
LGTM. Just one outcommented line should be removed

packages/datadog-plugin-prisma/test/integration-test/client.spec.js
BridgeAR: I will fix this as a drive-by in another PR.

PR #7496, opened 2026-02-11
feat(apim): add inferred span support for Azure API Managment

jcstorms1:
### What does this PR do?

Adds support for Azure APIM inferred proxy spans in HTTP triggered azure functions. The inferred span is only added when the appropriate headers are detected from API Management.

image

### Motivation

[APMSVLS-383](https://datadoghq.atlassian.net/browse/APMSVLS-383)
### Additional Notes

[APMSVLS-383]: https://datadoghq.atlassian.net/browse/APMSVLS-383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/datadog-plugin-azure-functions/test/integration-test/http-test/client.spec.js
+++ b/packages/datadog-plugin-azure-functions/test/integration-test/http-test/client.spec.js
@@ -50,7 +50,14 @@ describe('esm', () => {
         assert.strictEqual(payload.length, 1)
         assert.ok(Array.isArray(payload[0]))
         assert.strictEqual(payload[0].length, 1)
-        assert.strictEqual(payload[0][0].name, 'azure.functions.invoke')
+
+        const span = payload[0][0]
+
+        assert.strictEqual(span.name, 'azure.functions.invoke')
+        assert.strictEqual(span.meta['_dd.integration'], 'azure-functions')
+        assert.strictEqual(span.meta.component, 'azure-functions')
+        assert.strictEqual(span.meta['http.route'], '/api/httptest')
+        assert.strictEqual(span.resource, 'GET /api/httptest')
       })
     }).timeout(60_000)
 


packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -26,19 +26,48 @@ class AzureFunctionsPlugin extends TracingPlugin {
...
+        method: httpRequest.method,
+        headers: Object.fromEntries(httpRequest.headers),
+        url: path,
+      }
+      // Patch the request to create web context
+      const webContext = web.patch(req)
+      webContext.config = this.config
+      webContext.tracer = this.tracer
+      // Creates a standard span and an inferred proxy span if headers are present
+      span = serverless.startServiceEntrypointSpan(
```
bengl: This is a new method you've added to `web.js`, but here you're referring to it as belonging to `serverless`. Does `serverless` include all of `web.js`? `web.js` is also required in this file. We could use some consistency here.

packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -26,19 +26,48 @@ class AzureFunctionsPlugin extends TracingPlugin {
...
-      meta,
-    }, ctx)
-
-    if (isMessagingService) {
-      setSpanLinks(triggerType, this.tracer, span, ctx)
-    }
 
+    let span
+
+    if (isHttpTrigger) {
```
bengl, line 33: This if-block seems like it ought to be common across all integrations supported inferred proxy spans. Can we encapsulate it within the definition of `startServiceEntrypointSpan`?

packages/datadog-plugin-azure-functions/src/index.js
jcstorms1: Yeah we can just get rid of serverless I think.

packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -26,19 +26,48 @@ class AzureFunctionsPlugin extends TracingPlugin {
...
+      const { httpRequest } = ctx
+      const path = (new URL(httpRequest.url)).pathname
+      const req = {
+        method: httpRequest.method,
+        headers: Object.fromEntries(httpRequest.headers),
```
tlhunter, line 38: This code seems a little expensive... That said I've not looked at the Azure integration before.

Shouldn't we already have path and header information by this point?

tlhunter: Also, does this only run when the app is in Azure? Or in other words, it's not like all incoming HTTP requests run this code path?

jcstorms1: This only runs when an [HTTP Azure Function](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-trigger?tabs=python-v2%2Cisolated-process%2Cnodejs-v4%2Cfunctionsv2&pivots=programming-language-javascript) is triggered.

packages/dd-trace/src/plugins/util/web.js
```diff
@@ -275,7 +274,7 @@ const web = {
     return context.middleware.at(-1)
   },
 
-  startChildSpan (tracer, config, name, req, traceCtx) {
+  startServiceEntrypointSpan (tracer, config, name, req, traceCtx) {
```
rochdev: All root spans are service entry point spans. Wasn't this supposed to be `startServerlessEntrypointSpan`?

Although when I think about it, even that is not really accurate, because it could apply outside of serverless. Maybe `startSpanWithProxy`? Naming is hard.

packages/datadog-plugin-azure-functions/src/index.js
jcstorms1: We looked into this and the refactor is fairly complex for this PR. Should be handled in a separate project.

packages/datadog-plugin-azure-functions/src/index.js
jcstorms1: Spoke offline. Refactor is too complex for this PR.

PR #7503, opened 2026-02-12
[MLOS-459] Support enriched evalmetric event submission

gsvigruha:
### What does this PR do?

Adds reasoning, metadata and assessment to the submitEvaluation endpoint.
Add support for JSON value type.

### Motivation

Close feature gaps between the Python SDK.
Customer FR: https://datadoghq.atlassian.net/browse/MLOS-459

### Test
 - [score with reasoning, metadata](https://app.datadoghq.com/llm/traces?query=%40ml_app%3Atest%20%40event_type%3Aspan%20%40parent_id%3Aundefined&agg_m=count&agg_m_source=base&agg_t=count&fromUser=false&selectedTab=evaluations&sp=%5B%7B%22mp%22%3A%7B%22className%22%3A%22%22%7D%2C%22p%22%3A%7B%22eventId%22%3A%22AgAAAZxVAB4kr-oBgwAAAAAAAAAYAAAAAEFaeFZBQjRrQUFDZE9iVjA0X3JzQUFBQQAAACQAAAAAMDE5YzU1MDAtYjE2MC00ZmYxLWE1MzItOWJhZTJiYjJhMjNi%22%7D%2C%22i%22%3A%22llm-obs-panel%22%7D%5D&spanId=8035829709590228610&start=1770949005698&end=1770952605698&paused=false)
 - [failed bool eval](https://app.datadoghq.com/llm/traces?query=%40ml_app%3Atest%20%40event_type%3Aspan%20%40parent_id%3Aundefined&agg_m=count&agg_m_source=base&agg_t=count&fromUser=false&selectedTab=evaluations&sp=%5B%7B%22mp%22%3A%7B%22className%22%3A%22%22%7D%2C%22p%22%3A%7B%22eventId%22%3A%22AgAAAZxVAwbMapnVAQAAAAAAAAAYAAAAAEFaeFZBd2JNQUFCaTlzV3Zab2tnQUFBQQAAACQAAAAAMDE5YzU1MDMtNjRlZS00NDA2LWJiYWMtYmQ2NDM0NjY1MTlk%22%7D%2C%22i%22%3A%22llm-obs-panel%22%7D%5D&spanId=5810447116032819132&start=1770949211384&end=1770952811384&paused=false)
 - [json value](https://app.datadoghq.com/llm/traces?query=%40ml_app%3Atest%20%40event_type%3Aspan%20%40parent_id%3Aundefined&agg_m=count&agg_m_source=base&agg_t=count&fromUser=true&selectedTab=evaluations&sp=%5B%7B%22mp%22%3A%7B%22className%22%3A%22%22%7D%2C%22p%22%3A%7B%22eventId%22%3A%22AgAAAZxU_h_n1HS7BQAAAAAAAAAYAAAAAEFaeFVfaF9uQUFCM2ZiOXltaFVUQUFBQQAAACQAAAAAZjE5YzU0ZmUtOWZiMS00ODBkLThjNjMtMmQwN2QxMjZhZDgw%22%7D%2C%22i%22%3A%22llm-obs-panel%22%7D%5D&spanId=8706237192051336874&start=1770948909996&end=1770952509996&paused=false)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3269,12 +3269,12 @@ declare namespace tracer {
       /**
        * Reasoning for the evaluation result.
        */
-      reasoning?: 'pass' | 'fail',
+      reasoning?: string,
 
       /**
        * Whether the evaluation passed or failed. Valid values are pass and fail.
        */
-      assessment?: string,
+      assessment?: 'pass' | 'fail',
 
       /**
        * Arbitrary JSON data associated with the evaluation.


packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -381,6 +381,18 @@ class LLMObs extends NoopLLMObs {
         err = 'invalid_metric_value'
         throw new Error('value must be a boolean for a boolean metric')
       }
+      if (assessment !== undefined && assessment !== 'pass' && assessment !== 'fail') {
```
sabrenner:
```suggestion
      if (!['pass', 'fail'].includes(assessment)) {
```

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -381,6 +381,18 @@ class LLMObs extends NoopLLMObs {
...
+        err = 'invalid_assessment'
+        throw new Error('assessment must be pass or fail')
+      }
+      if (reasoning !== undefined && typeof reasoning !== 'string') {
```
sabrenner:
```suggestion
      if (typeof reasoning !== 'string') {
```

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -381,6 +381,18 @@ class LLMObs extends NoopLLMObs {
...
+        err = 'invalid_reasoning'
+        throw new Error('reasoning must be a string')
+      }
+      if (metadata !== undefined && !(typeof metadata === 'object' && metadata !== null && !Array.isArray(metadata))) {
```
sabrenner:
```suggestion
      if (metadata == null || typeof metadata !== 'object' || Array.isArray(metadata)) {
```

although i saw the `Array.isArray` check, is `metadata` expected to be an object or an array?

packages/dd-trace/src/llmobs/sdk.js
sabrenner, line 1: we can add these three optional fields to our typedocs as well! https://github.com/DataDog/dd-trace-js/blob/875e58513b5cde4e3ddd1233fb2a12c5c7008cf5/index.d.ts#L3237

packages/dd-trace/src/llmobs/sdk.js
gsvigruha: it's meant to be an object not an array. i guess im not sure if `'object'` type only means json objects only or both

sabrenner: ah ok, i see what you're doing here then. let me amend my original suggestion if we do not want to allow arrays through
```
> const obj = {}
undefined
> const arr = []
undefined
> typeof obj
'object'
> typeof arr
'object'
```

gsvigruha: yeah it's really weird how you have a dichotomy of objects vs arrays on one level but also they are both objects

packages/dd-trace/src/llmobs/sdk.js
gsvigruha: unit test seems to fail on non specified assessment w/o the assessment !== undefined  part

index.d.ts
```diff
@@ -3265,6 +3265,21 @@ declare namespace tracer {
...
+
+      /**
+       * Reasoning for the evaluation result.
+       */
+      reasoning?: string,
+
+      /**
+       * Whether the evaluation passed or failed. Valid values are pass and fail.
+       */
+      assessment?: string,
```
sabrenner:
```suggestion
      assessment?: 'pass' | 'fail',
```

should be able to do this to help guide typing in the right direction (although we still wanna keep the validation in the sdk method as you already have too) - [example](https://github.com/DataDog/dd-trace-js/blob/c589ad403bbfba61c314dae4d0732f3a143bf6ae/index.d.ts#L3246)

packages/dd-trace/src/llmobs/sdk.js
BridgeAR:
```suggestion
      if (Array.isArray(metadata) || metadata != null && typeof metadata !== 'object') {
```

That should work the same, if I am not mistaken.

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -381,6 +381,22 @@ class LLMObs extends NoopLLMObs {
...
+        err = 'invalid_metadata'
+        throw new Error('metadata must be a JSON object')
```
BridgeAR, line 398: A JSON object meaning the plain object must be possible to be stringified with JSON.stringify?

gsvigruha: JSON.stringify should work on it, but it also works on types we don't accept like JSON arrays

packages/dd-trace/src/llmobs/sdk.js
gsvigruha: fails for the case when i dont specify metadata in the API

index.d.ts
aniszoubiramar: As we support `json` we should probably add `object` here too.

aniszoubiramar: metric type `json` is missing

gsvigruha: good catch

gsvigruha: good catch

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -425,6 +441,15 @@ class LLMObs extends NoopLLMObs {
...
       }
+      if (reasoning !== undefined) {
+        payload.reasoning = reasoning
+      }
+      if (metadata !== undefined) {
+        payload.metadata = metadata
+      }
+      if (assessment !== undefined) {
+        payload.assessment = assessment
+      }
```
sabrenner: these will allow `null` values through (in js, `null !== undefined`). we can do this instead

```suggestion
      if (reasoning != null) {
        payload.reasoning = reasoning
      }
      if (metadata != null) {
        payload.metadata = metadata
      }
      if (assessment != null) {
        payload.assessment = assessment
      }
```

as `null == undefined`

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -381,6 +381,22 @@ class LLMObs extends NoopLLMObs {
         err = 'invalid_metric_value'
         throw new Error('value must be a boolean for a boolean metric')
       }
+      if (metricType === 'json' && !(typeof value === 'object' && value !== null && !Array.isArray(value))) {
+        err = 'invalid_metric_value'
+        throw new Error('value must be a JSON object for a json metric')
```
sabrenner, line 386:
```suggestion
        throw new Error('value must be a non-Array object for a json metric')
```

packages/dd-trace/src/llmobs/sdk.js
sabrenner:
```suggestion
if (metadata != null && (typeof metadata !== 'object' || Array.isArray(metadata))) {
```

i think this one might work then

index.d.ts
```diff
@@ -3243,13 +3243,13 @@ declare namespace tracer {
...
-      metricType: 'categorical' | 'score' | 'boolean',
+      metricType: 'categorical' | 'score' | 'boolean' | 'json',
 
       /**
        * The value of the evaluation metric.
-       * Must be string for 'categorical' metrics, number for 'score' metrics, and boolean for 'boolean' metrics.
+       * Must be string for 'categorical' metrics, number for 'score' metrics, boolean for 'boolean' metrics and a JSON object for 'json' metrics.
        */
-      value: string | number | boolean,
+      value: string | number | boolean | object,
```
sabrenner:
```suggestion
      value: string | number | boolean | { [key: string]: any },
```

we can do this to help with typing enforcement up front (for typescript users), as it's more restrictive, it'll only pass for things like

```js
const metadata = {
  someKey: ...
}
```

although we still need to keep the validation you have already in the actual sdk

packages/dd-trace/src/llmobs/sdk.js
sabrenner: meh actually take or leave this one, i see it would require test changes too

sabrenner (commented):
just a couple more js-specific comments! thanks for attaching manual tests in the pr description 🙇

index.d.ts
sabrenner: fyi looks like this had gotten changed but got reverted

index.d.ts
gsvigruha: i can do both

packages/dd-trace/src/llmobs/sdk.js
gsvigruha: i think the current one is ever so slightly more clear

packages/dd-trace/src/llmobs/sdk.js
gsvigruha: thanks! i can never wrap my head around TS `null` vs `undefined` and  `==` vs `===`

index.d.ts
gsvigruha: no added it to the wrong place 🤦

PR #7505, opened 2026-02-12
ci(all-green): make workflow event-driven from check_suite

watson:
### What does this PR do?

Makes the All Green workflow event-driven: it now triggers on `check_suite` completed instead of `pull_request`, `push`, and cron. Concurrency is keyed by head SHA with `cancel-in-progress: true` so only the latest run per commit proceeds. The workflow uses `wechuli/allcheckspassed` with `commit_sha` from the event and a 1-minute `delay` for debounce.

### Motivation

Reduce total GitHub API usage and the risk of hitting rate limits. Previously the workflow ran on a schedule and polled every 2 minutes; now it runs only when a check suite completes, and concurrency plus the action’s delay ensure we don’t run redundant polls. Fewer workflow runs and fewer API calls per commit.

### DIFF
--- a/.github/workflows/all-green.yml
+++ b/.github/workflows/all-green.yml
@@ -1,18 +1,13 @@
 name: All Green
 on:
-  pull_request:
-  push:
-    branches:
-      - master
-  schedule:
-    - cron: 0 4 * * *
+  check_suite:
+    types: [completed]
 
 concurrency:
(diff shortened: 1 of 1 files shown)


BridgeAR:
I believe this is currently outdated. In general, it would be a nice to have but I am going to close it for now. Please reopen, when you want to continue on it.

PR #7509, opened 2026-02-12
feat(iast): Improve callsite processing on IAST vuln location computing

CarlesDD:
### What does this PR do?
Simplifies IAST stack trace filtering by replacing the manually maintained `EXCLUDED_PATH_PREFIXES` list with `filename.startsWith('node:')` to automatically exclude all Node.js built-in modules. Also adds fallback to all processed frames when no client frames are found.

### Motivation
Since dd-trace v5 only supports Node.js >= 18, and all built-in modules have the node: prefix in stack traces, we can exclude them automatically. This eliminates maintenance burden when new Node.js modules appear in stack traces.

[APPSEC-61435](https://datadoghq.atlassian.net/browse/APPSEC-61435)

[APPSEC-61435]: https://datadoghq.atlassian.net/browse/APPSEC-61435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dd-trace/src/appsec/iast/path-line.js
```diff
@@ -76,14 +80,12 @@ function isExcluded (callsite, externallyExcludedPaths) {
     excludedPaths = [...excludedPaths, ...externallyExcludedPaths]
   }
 
-  for (const excludedPath of excludedPaths) {
-    if (filename.includes(excludedPath)) {
-      return true
-    }
+  if (filename.startsWith('node:')) {
+    return true
```
BridgeAR: Ideally, this is moved up into the `!filename` if statement with an OR. That way there is no additional work done while checking for that.

packages/dd-trace/src/appsec/iast/path-line.js
```diff
@@ -47,18 +47,22 @@ function getNonDDCallSiteFrames (callSiteFrames, externallyExcludedPaths) {
...
+        !isExcluded(callsite, externallyExcludedPaths) &&
+        (!filepath.includes(pathLine.ddBasePath) || globalThis.__DD_ESBUILD_IAST_WITH_NO_SM)
+      ) {
+        clientFrames.push(callsite)
+      }
     }
   }
 
-  return result
+  return clientFrames.length > 0 ? clientFrames : allFrames
```
BridgeAR, line 76: Do we need to copy the frames for `allFrames`? Would it not be fine to just return `callSiteFrames`?

```suggestion
  return clientFrames.length > 0 ? clientFrames : callSiteFrames
```

packages/dd-trace/src/appsec/iast/path-line.js
```diff
@@ -8,29 +8,29 @@ const { getOriginalPathAndLineFromSourceMap } = require('./taint-tracking/rewrit
...
-]
 
-function getNonDDCallSiteFrames (callSiteFrames, externallyExcludedPaths) {
+/**
+ * Processes and filters call site frames to find the best location for a vulnerability.
+ * Returns client frames if available, otherwise falls back to all processed frames.
+ * Excludes dd-trace frames and all Node.js built-in/internal modules (node:*).
+ * @param {Array} callSiteFrames
+ * @param {Array} externallyExcludedPaths
+ * @returns {Array} Client frames if available, otherwise all processed frames
```
BridgeAR: Nit: define the content of the arrays. That could be defined in the Node.js types already

packages/dd-trace/src/appsec/iast/analyzers/vulnerability-analyzer.js
```diff
@@ -35,9 +35,8 @@ class Analyzer extends SinkIastPlugin {
 
   _reportEvidence (value, context, evidence) {
     const callSiteFrames = getVulnerabilityCallSiteFrames()
-    const nonDDCallSiteFrames = getNonDDCallSiteFrames(callSiteFrames, this._getExcludedPaths())
-
-    const location = this._getLocation(value, nonDDCallSiteFrames)
+    const frames = getCallSiteFramesForLocation(callSiteFrames, this._getExcludedPaths())
+    const location = this._getLocation(value, frames)
```
BridgeAR, line 39: Nit: are these underscored properties really meant as `internally private`? A customer would normally not get hold of them, if I am not mistaken, while we seem to use them across classes.

I am just commenting about it because it would be great to make truly internal methods truly private and keep others accessible (as long as they are not directly accessible for customers).

CarlesDD: No, they're not meant as `internally private`. `_getLocation` is overridden by several subclasses (e.g., `CookieAnalyzer`, `HardcodedBaseAnalyzer`, `MissingHeaderAnalyzer`) and some call `super._getLocation()`

BridgeAR: In that case we should not use the underscore, right?

CarlesDD: The underscore prefix here acts as "protected" in OOP terms: not for external/customer use, but accessible to internal subclasses. This is a consistent pattern across the codebase, not just in appsec:

- Profiling: Profiler defines [_collect()](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/profiling/profiler.js#L302), ServerlessProfiler [overrides it](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/profiling/profiler.js#L418) and calls super._collect()
- Exporters: Writer base class [calls](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/exporters/common/writer.js#L21) `this._sendPayload()`, implemented by [AgentWriter](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/exporters/agent/writer.js#L28), [CoverageWriter](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/ci-visibility/exporters/agentless/coverage-writer.js#L27), [DynamicInstrumentationLogsWriter](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/ci-visibility/exporters/agentless/di-logs-writer.js#L24), and others

Changing this naming convention across the codebase would be a larger refactor outside the scope of this PR.

BridgeAR: I understand why it was used but we are currently moving away from that in most code.
Thus, when we touch something, it would be great to change it as well. In this particular case making them non-underscored would in my opinion be the way to go.

I marked it as nit, so it is not something that I expect to be done as part of the PR, while it would be nice.

packages/dd-trace/src/appsec/iast/path-line.js
CarlesDD: We can't return `callSiteFrames` directly because `allFrames` is a subset of processed frames that have valid file information (only contains frames with a valid `filepath` - line 61 filter). Frames without `filepath` are excluded and wouldn't have the `path`/`isInternal` properties added.

packages/dd-trace/src/appsec/iast/path-line.js
```diff
@@ -67,10 +82,12 @@ function getRelativePath (filepath) {
 
 function isExcluded (callsite, externallyExcludedPaths) {
   if (callsite.isNative) return true
+
   const filename = globalThis.__DD_ESBUILD_IAST_WITH_SM ? callsite.path : callsite.file
-  if (!filename) {
+  if (!filename || filename.startsWith('node:')) {
```
uurien, line 87: this would work with all the node core modules that are included as `node:`, but what when they don't have the `node:`? In the original list we had both, `node:child_process` and `child_process` etc.

CarlesDD: In Node.js 18+, built-in modules in stack traces use the `node:` prefix. The original list had both variants to support older Node.js versions where the behavior was inconsistent.

PR #7526, opened 2026-02-13
[test optimization] Report suppressed errors 

juan-fernandez:
### What does this PR do?

Depends on #7524

Add a report message at the end of the execution to let users know that some test errors have been suppressed:

```
FAIL ci-visibility/test-management/test-quarantine-1.js
  quarantine tests
    ✕ can quarantine a test (10 ms)
    ✓ can pass normally

  ● quarantine tests › can quarantine a test
    ...assertion error details...

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 passed, 2 total
Snapshots:   0 total
Time:        0.22 s
Ran all test suites.

--------------------------------------------------
Datadog Test Optimization
--------------------------------------------------
2 test failure(s) were ignored. Exit code set to 0.

  • ci-visibility/test-early-flake-detection/occasionally-failing-test.js › fail occasionally fails (Early Flake Detection)
```

### Motivation
Make it easier to users to understand our product.


packages/datadog-instrumentations/src/jest.js
```diff
@@ -129,13 +129,41 @@ function getTestEnvironmentOptions (config) {
...
 
+/**
+ * @param {string[]} efdNames
+ * @param {string[]} quarantineNames
+ * @param {number} totalCount
+ */
+function logIgnoredFailuresSummary (efdNames, quarantineNames, totalCount) {
+  const names = []
+  for (const n of efdNames) {
+    names.push({ name: n, reason: 'Early Flake Detection' })
```
calvinbayer, line 149: What about ATR?

juan-fernandez: `jest` completely swallows those errors already and I would say we don't want to be more clever than `jest`

PR #7528, opened 2026-02-13
ci: fix workflow YAML schema and standardize formatting

watson:
### What does this PR do?

Fixes YAML lint errors in GitHub workflow files and standardizes formatting across all workflows.

1. **Schema fix**: Adds `yaml.schemas` in `.vscode/settings.json` so `.github/workflows/*.yml` uses the GitHub Actions schema. Previously, `serverless.yml` was validated against the Serverless Framework schema, which caused false "missing provider/service" errors.
2. **Formatting**: Re-formats all workflow files with the yaml-language-server formatter for consistent style (quotes, spacing, indentation).

### Motivation

`serverless.yml` was incorrectly validated as Serverless Framework config because of its filename, leading to misleading lint errors. Additionally, format-on-save was constantly introducing unrelated formatting changes into diffs when editing workflows. Standardizing formatting once should keep future diffs focused on actual changes.

### DIFF
--- a/.github/workflows/apm-integrations.yml
+++ b/.github/workflows/apm-integrations.yml
@@ -554,7 +554,7 @@ jobs:
   http:
     strategy:
       matrix:
-        node-version: [oldest, maintenance, "latest"]
+        node-version: [oldest, maintenance, latest]
     runs-on: ubuntu-latest
     env:
       PLUGINS: http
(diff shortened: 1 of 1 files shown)


.github/workflows/apm-integrations.yml
```diff
@@ -555,7 +554,7 @@ jobs:
   http:
     strategy:
       matrix:
-        node-version: [oldest, maintenance, 'latest']
+        node-version: [oldest, maintenance, "latest"]
```
pabloerhard: nit: Would be nice to standardize these ones

watson: I agree. I'll update the PR. This was just an auto-lint-fix

watson: Fixed 👍

PR #7531, opened 2026-02-13
feat: support for OTLP trace export (http/json)

ida613:
### What does this PR do?
This is the dd-trace-js counterpart to this [dotnet PR](https://github.com/DataDog/dd-trace-dotnet/pull/8211). But since there is no trace stats computation in dd-trace-js, this PR doesn't export any metrics, so is smaller than the dotnet PR


packages/dd-trace/src/config/defaults.js
```diff
@@ -149,6 +149,11 @@ module.exports = {
   otelMetricsHeaders: '',
   otelMetricsProtocol: 'http/protobuf',
   otelMetricsTimeout: 10_000,
+  otelTracesEnabled: false,
```
ida613: this config var is not exactly in the requirement doc. This was added as the counterpart to `config.otelMetricsEnabled` and `config.otelLogsEnabled`, and is equals to OTEL_TRACES_EXPORTER==otlp

integration-tests/helpers/fake-agent.js
```diff
@@ -495,6 +495,14 @@ function buildExpressServer (agent) {
     })
   })
 
+  app.post('/v1/traces', (req, res) => {
```
zacharycmontoya, line 506: FYI, there's a separate effort to make a "v1" of the Datadog trace payload. I'm not sure if that's going to conflict with this request path or not

packages/dd-trace/src/config/supported-configurations.json
```diff
@@ -474,6 +474,10 @@
     "OTEL_EXPORTER_OTLP_METRICS_HEADERS": ["A"],
     "OTEL_EXPORTER_OTLP_METRICS_PROTOCOL": ["A"],
     "OTEL_EXPORTER_OTLP_METRICS_TIMEOUT": ["A"],
+    "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT": ["A"],
```
zacharycmontoya: I think we should update `OTEL_TRACES_EXPORTER` to version "B" which has the new understanding of `otlp` enabling our OTLP Traces export. I'll do the same on my corresponding .NET PR

packages/dd-trace/src/config/index.js
```diff
@@ -1120,9 +1140,10 @@ class Config {
 
     // Compute OTLP logs and metrics URLs to send payloads to the active Datadog Agent
     const agentHostname = this.#getHostname()
-    calc.otelLogsUrl = `http://${agentHostname}:${DEFAULT_OTLP_PORT}`
-    calc.otelMetricsUrl = `http://${agentHostname}:${DEFAULT_OTLP_PORT}/v1/metrics`
-    calc.otelUrl = `http://${agentHostname}:${DEFAULT_OTLP_PORT}`
+    calc.otelLogsUrl = `http://${agentHostname}:${DEFAULT_OTLP_HTTP_PORT}`
+    calc.otelMetricsUrl = `http://${agentHostname}:${DEFAULT_OTLP_HTTP_PORT}/v1/metrics`
+    calc.otelTracesUrl = `http://${agentHostname}:${DEFAULT_OTLP_HTTP_PORT}/v1/traces`
```
zacharycmontoya: Is this the only place where the final configuration gets calculated? If so, this is okay for now since it seems like `grpc` is not allowed, but in a subsequent PR where we introduce `grpc` support, we'll need the Traces Endpoint calculation logic to look like this:

- Calculate the traces protocol. If `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` is set, prefer this value over `OTEL_EXPORTER_OTLP_PROTOCOL`. If neither are set, use the library default ✅ 
- Use `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` as the absolute URL for traces, if set ✅ 
- Otherwise, calculate the _base path_ for the endpoint then the _subpath_. Use `OTEL_EXPORTER_OTLP_ENDPOINT` for the _base path_ if it is set, or use the default path: `http://localhost:4317` for `otelTracesProtocol == grpc` or `http://localhost:4318` otherwise. Then once you have the _base path_, if  `otelTracesProtocol == grpc` then there is no _subpath_, otherwise the _subpath_ is `/v1/traces`. Combine the two to get the full URL for the traces endpoint ❌

zacharycmontoya: This is where I've gotten this additional logic for HTTP vs gRPC: https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/

packages/dd-trace/src/opentelemetry/trace/otlp_http_trace_exporter.js
```diff
@@ -0,0 +1,56 @@
...
+ * https://opentelemetry.io/docs/specs/otlp/#otlphttp
+ *
+ * It receives DD-formatted spans (from span_format.js), transforms them
+ * to OTLP ExportTraceServiceRequest JSON format, and sends them to the
+ * configured OTLP endpoint via HTTP POST.
+ *
+ * @class OtlpHttpTraceExporter
+ * @augments OtlpHttpExporterBase
+ */
+class OtlpHttpTraceExporter extends OtlpHttpExporterBase {
```
zacharycmontoya, line 21: It looks like this follows the convention set by OpenTelemetry to name the `http/json` exporter [exporter-trace-otlp-http](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/exporter-trace-otlp-http) and the `http/protobuf` exporter [exporter-trace-otlp-proto](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/exporter-trace-otlp-proto), so the naming looks good to me 👍🏼

packages/dd-trace/test/config/index.spec.js
```diff
@@ -297,6 +297,17 @@ describe('Config', () => {
     assert.strictEqual(config.sampleRate, 0.1)
   })
 
+  it('should enable OTLP traces export when OTEL_TRACES_EXPORTER is set to otlp', () => {
+    process.env.OTEL_TRACES_EXPORTER = 'otlp'
+    const config = getConfig()
+    assert.strictEqual(config.otelTracesEnabled, true)
```
zacharycmontoya, line 443: Let's also assert that the sampling strategy is set to `parentbased_always_on`, which means priority sampling with a general sample rate of 1.0. I think that's represented as 

```javascript
assert.strictEqual(config.sampleRate, 1.0)
```

packages/dd-trace/src/span_processor.js
```diff
@@ -52,6 +54,19 @@ class SpanProcessor {
     }
     if (started.length === finished.length || finished.length >= flushMinSpans) {
       this.sample(span)
+
+      // With OTLP export there is no agent to apply sampling, so drop
+      // rejected traces here to avoid sending unsampled data.
+      if (this._dropUnsampled && spanContext._sampling.priority < AUTO_KEEP) {
```
zacharycmontoya: I'll leave this to your JS colleagues to review how we should do this, but this is a similar scenario to client-side stats calculation, where we avoid sending ALL spans to the local agent. Instead, we drop "P0" or unsampled spans and we still obtain accurate trace stats by calculating stats in the SDK (with all sampled and unsampled data). However, in this PR we are not implementing the trace stats calculation.

packages/dd-trace/test/span_processor.spec.js
```diff
@@ -212,4 +212,75 @@ describe('SpanProcessor', () => {
...
+      prioritySampler.sample.callsFake((ctx) => { ctx._sampling.priority = 1 })
+
+      trace.started = [finishedSpan]
+      trace.finished = [finishedSpan]
+      proc.process(finishedSpan)
+
+      sinon.assert.calledOnce(exporter.export)
+    })
+
+    it('should preserve active spans when dropping a rejected trace', () => {
```
zacharycmontoya: This looks like it behaves differently than the test case where `it('should skip traces with unfinished spans'`. Is that what you intended?

packages/dd-trace/test/opentelemetry/traces.spec.js
```diff
@@ -0,0 +1,499 @@
...
+      const { resource, scopeSpans } = decoded.resourceSpans[0]
+
+      const resourceAttrs = extractAttrs(resource.attributes)
+      assert.strictEqual(resourceAttrs['service.name'], 'test-service')
+
+      assert.strictEqual(scopeSpans.length, 1)
+      assert.strictEqual(scopeSpans[0].scope.name, 'dd-trace-js')
+
+      const otlpSpan = scopeSpans[0].spans[0]
+      assert.strictEqual(otlpSpan.name, 'test.operation')
```
zacharycmontoya: I expect this to be `/api/test`, which corresponds to the DD span's `resource` field

packages/dd-trace/test/opentelemetry/traces.spec.js
```diff
@@ -0,0 +1,499 @@
...
+            'exception.type': 'Error',
+          },
+        }],
+      })
+
+      const decoded = decodePayload(transformer.transformSpans([span]))
+      const otlpSpan = decoded.resourceSpans[0].scopeSpans[0].spans[0]
+
+      assert.strictEqual(otlpSpan.events.length, 1)
+      assert.strictEqual(otlpSpan.events[0].name, 'exception')
```
zacharycmontoya, line 372: Let's also assert that the timestamp was mapped correctly

ida613: will do! copying the requirement from your message just to keep track: get the current time stamp, ensure the start time stamp in the trace is greater than or equal to this time stamp, then assert the end time timestamp is greater than or equal to the start time

packages/dd-trace/test/opentelemetry/traces.spec.js
```diff
@@ -0,0 +1,499 @@
...
+
+      const tracer = setupTracer()
+      const processor = tracer._tracer._processor
+      const exporter = processor._exporter
+
+      exporter.export([createMockSpan()])
+    })
+
+    it('includes custom headers from OTEL_EXPORTER_OTLP_TRACES_HEADERS', () => {
+      process.env.OTEL_EXPORTER_OTLP_TRACES_HEADERS = 'x-api-key=secret123'
```
zacharycmontoya, line 480: Would you mind adding another to test that we handle comma-separated values and assert that value too? Such as `x-api-key=secret123,other-config-value=value` and then adding a test to ensure `assert.strictEqual(headers['other-config-value'], 'value')`

packages/dd-trace/test/opentelemetry/traces.spec.js
```diff
@@ -0,0 +1,499 @@
...
+    })
+
+    it('uses port 4318 for default OTLP HTTP endpoint', () => {
+      delete process.env.OTEL_EXPORTER_OTLP_TRACES_PROTOCOL
+
+      const config = getConfigFresh()
+      assert(config.otelTracesUrl.includes(':4318'), `expected port 4318 in URL, got: ${config.otelTracesUrl}`)
+    })
+
+    it('respects explicit endpoint', () => {
```
zacharycmontoya: In a [previous comment I made](https://github.com/DataDog/dd-trace-js/pull/7531#discussion_r2879542334), I outlined the full logic for calculating the traces endpoint when configuring `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`/`OTEL_EXPORTER_OTLP_ENDPOINT`/default with different protocol values. For this PR that only implements `http/json`, can you implement the following test matrix?

1. Default endpoint with `http/json` => `http://localhost:4318/v1/traces`
2. `OTEL_EXPORTER_OTLP_ENDPOINT` endpoint with `http/json` => `${OTEL_EXPORTER_OTLP_ENDPOINT}/v1/traces`
3. `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` endpoint with `http/json` => `${OTEL_EXPORTER_OTLP_TRACES_ENDPOINT}`
  - You could also set `OTEL_EXPORTER_OTLP_ENDPOINT` to ensure that `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` takes precedence

integration-tests/opentelemetry-traces.spec.js
```diff
@@ -0,0 +1,153 @@
...
+    assert.strictEqual(dbSpan.name, 'db.query')
+    assert.strictEqual(errSpan.name, 'error.operation')
+
+    // Validate span kind (server=2, client=3 per OTLP proto SpanKind enum)
+    assert.strictEqual(webSpan.kind, 2, 'web.request should be SERVER kind')
+    assert.strictEqual(dbSpan.kind, 3, 'db.query should be CLIENT kind')
+
+    // Validate timing fields
+    for (const span of spans) {
+      assert.ok(span.startTimeUnixNano > 0, 'span should have a positive startTimeUnixNano')
```
zacharycmontoya: One thing I did in my .NET tests is before launching the program that starts/stops spans, I took the unix time and then when I asserted against all the spans, I tested that it was greater than that time, rather than comparing against 0. If you could do the same here, that would make the test even better

zacharycmontoya (commented):
I've left a couple of comments. From the OTLP and requirements side of things, this PR is nearly complete. I've left a couple of comments to address and once those are completed I think this will meet the requirements

packages/dd-trace/src/config/supported-configurations.json
ida613: There's been a [recent refactoring](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/config/supported-configurations.json#L3919-L3938) of this file, which puts B for OTEL_EXPORTER_OTLP_LOGS_HEADERS and D for OTEL_EXPORTER_OTLP_LOGS_PROTOCOL. Should I follow suit for the corresponding TRACES env vars?

packages/dd-trace/src/config/index.js
ida613: thanks Zach I'll keep this in mind!

packages/dd-trace/src/span_processor.js
ida613: thanks Zach! @rochdev or @BridgeAR do you want to pitch in here?

packages/dd-trace/test/config/index.spec.js
ida613: thanks Zach! just added a test in this [commit ](https://github.com/DataDog/dd-trace-js/pull/7531/changes/6aea9301df29f6e5755c58fd173484a40749d40b#diff-21166de853b46f477b1eb0f82172e751350b701db0f08bdf298662253138cb53)

packages/dd-trace/test/span_processor.spec.js
ida613: I [updated the existing test](https://github.com/DataDog/dd-trace-js/pull/7781) to clarify behaviour :) does it make sense?

packages/dd-trace/test/opentelemetry/traces.spec.js
ida613: [updated](https://github.com/DataDog/dd-trace-js/pull/7531/changes/0704ecc46ec51aedafcc11baa79020e1253735d7)! thanks Zach :'D

packages/dd-trace/test/opentelemetry/traces.spec.js
ida613: thank you Zach for taking the time to flush out all the details! do the unit test changes in [this commit](https://github.com/DataDog/dd-trace-js/pull/7531/changes/c9462300d229fa3d941dbca5b966ac18e9ed791e#diff-952871828b539c34abdf9b9e6bb3ab61d2f2bcf8625065316f5d4209d8e92e85) look good?

packages/dd-trace/test/opentelemetry/traces.spec.js
ida613: [here](https://github.com/DataDog/dd-trace-js/pull/7531/changes/372022958133b5c02760a49cc5d6e32b04752f88) it is!

packages/dd-trace/test/opentelemetry/traces.spec.js
ida613: [here](https://github.com/DataDog/dd-trace-js/pull/7531/commits/8f6428ac3cf6f2e26f909449f3ba0907f25f5058) it is!

integration-tests/opentelemetry-traces.spec.js
ida613: pushed [this commit](https://github.com/DataDog/dd-trace-js/pull/7531/commits/594554276c7b59ea2ec0bc614301622343d7d669) :)

integration-tests/opentelemetry/otlp-traces.js
ida613, line 1: note: double check this integration test

packages/dd-trace/test/opentelemetry/traces.spec.js
zacharycmontoya: Thanks for the update! I have a couple more small unit tests that should be added:
- Setting `OTEL_EXPORTER_OTLP_HEADERS` by itself should result in the same headers being added to the OTLP traces
- When both `OTEL_EXPORTER_OTLP_HEADERS` and `OTEL_EXPORTER_OTLP_TRACES_HEADERS` are set,  only the value of the `OTEL_EXPORTER_OTLP_TRACES_HEADERS` setting is used

packages/dd-trace/test/opentelemetry/traces.spec.js
zacharycmontoya: It's almost there, but not quite.

For the signal-agnostic endpoint, should only add the `/v1/traces` path if `otelTracesProtocol` is `http/protobuf` or `http/json`. Otherwise (`grpc`) you don't add the subpath. So when it comes to the signal-agnostic endpoint tests you've added, each test case should basically have two versions: protocol=grpc or protocol is http. And the ones you've added cover the "protocol is http" case.

This means in the implementation code, you should move the traces protocol calculation first, and then calculate the endpoint second

packages/dd-trace/src/span_processor.js
zacharycmontoya: Sorry, I didn't mean to indicate that there's anything wrong with this approach. I just wanted to highlight this new behavior in the library for other reviewers that in this mode we no longer export _all_ spans, rather we only export _sampled_ spans

ida613: thanks Zach! there's been no misunderstanding! ^ ^ Ruben said he will think about this one and get back to me

packages/dd-trace/test/opentelemetry/traces.spec.js
ida613: Would it make sense to add this in a subsequent grpc PR?

packages/dd-trace/test/opentelemetry/traces.spec.js
ida613: thanks Zach! [here](https://github.com/DataDog/dd-trace-js/pull/7531/commits/d17f8f958af1487720107951ef7ee9ba99792858) it is!

packages/dd-trace/src/config/supported-configurations.json
ida613: this was resolved yesterday by Ruben during our meeting! I know what to do now and will update the code :'D

packages/dd-trace/src/opentelemetry/trace/index.js
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @param {Config} config - Tracer configuration instance
+ * @returns {import('@opentelemetry/api').Attributes} Resource attributes
+ */
+function buildResourceAttributes (config) {
+  const resourceAttributes = {
+    'service.name': config.service || config.tags.service,
+    'telemetry.sdk.name': 'datadog',
+    'telemetry.sdk.version': VERSION,
+    'telemetry.sdk.language': 'javascript',
```
zacharycmontoya: opentelemetry-js uses the value ['nodejs'](https://github.com/open-telemetry/opentelemetry-js/blob/3c797190f3e175f23858a4917d9af175862e2979/semantic-conventions/src/stable_attributes.ts#L1131). Do you think we should do the same or keep our 'javascript' value?

zacharycmontoya: And this is where it's actually referenced and used: https://github.com/open-telemetry/opentelemetry-js/blob/3c797190f3e175f23858a4917d9af175862e2979/packages/opentelemetry-core/src/platform/node/sdk-info.ts

packages/dd-trace/src/opentelemetry/trace/otlp_http_trace_exporter.js
```diff
@@ -0,0 +1,66 @@
...
+   * @returns {void}
+   */
+  export (spans) {
+    if (spans.length === 0) {
+      return
+    }
+
+    // Drop unsampled traces — OTLP endpoints have no agent-side sampling.
+    const priority = spans[0]?.metrics?.[SAMPLING_PRIORITY_KEY]
+    if (priority !== undefined && priority < AUTO_KEEP) {
```
zacharycmontoya, line 50: I'm not super familiar with the inner workings of this library, but let's make sure that a sampling decision is always made before we get to this point. Can you confirm this behavior?

packages/dd-trace/src/config/index.js
```diff
@@ -500,6 +504,29 @@ class Config {
         setString(target, 'otelMetricsTemporalityPreference', temporalityPref)
       }
     }
+    if (OTEL_TRACES_EXPORTER) {
+      setBoolean(target, 'otelTracesEnabled', OTEL_TRACES_EXPORTER.toLowerCase() === 'otlp')
+    }
+    // Set OpenTelemetry traces configuration with specific _TRACES_ vars
+    // taking precedence over generic _EXPORTERS_ vars
+    if (OTEL_EXPORTER_OTLP_ENDPOINT || OTEL_EXPORTER_OTLP_TRACES_ENDPOINT) {
```
zacharycmontoya: If neither `OTEL_EXPORTER_OTLP_ENDPOINT` or `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` are set, then by default the value for http/json and http/protobuf must be set to `http://localhost:4318/v1/traces` (see this for [reference](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_traces_endpoint)). Can you include that in this PR?

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
```diff
@@ -0,0 +1,332 @@
...
+   *
+   * @param {DDFormattedSpan} span - DD-formatted span to transform
+   * @returns {object} OTLP Span object
+   */
+  #transformSpan (span) {
+    const parentId = span.parent_id
+    const links = this.#extractLinks(span.meta?.['_dd.span_links'])
+
+    return {
+      traceId: this.#idToBytes(span.trace_id, 16),
```
zacharycmontoya, line 120: Note for later: When we later implement http/protobuf and grpc protocols, I think this DD->OTLP Span code can be re-used with some small tweaks. If we instead return traceId, spanId, and parentSpanId with them having the `Buffer` type, the http/protobuf and gRPC writers can use the buffer directly whereas the JSON serializer can call `toString('hex')` on the Buffer objects

packages/dd-trace/src/opentelemetry/trace/index.js
ida613: thanks Zach XD [updated](https://github.com/DataDog/dd-trace-js/pull/7531/changes/016cfbbb3e1831d8e3a1a6f8c4235848a0fc3843)

packages/dd-trace/src/opentelemetry/trace/otlp_http_trace_exporter.js
ida613: thanks Zach for calling it out 😄 in span processor, [sample](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/span_processor.js#L54) is always called before [export](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/span_processor.js#L71)

integration-tests/opentelemetry-traces.spec.js
```diff
@@ -0,0 +1,154 @@
...
+
+    // Validate non-error span status
+    assert.strictEqual(webSpan.status.code, 0, 'non-error span should have STATUS_CODE_UNSET')
+
+    // Validate span attributes include service.name and resource.name
+    const webAttrs = Object.fromEntries(
+      webSpan.attributes.map(({ key, value }) => [key, value])
+    )
+    assert.deepStrictEqual(webAttrs['service.name'], { stringValue: 'otlp-test-service' })
+    assert.ok(webAttrs['resource.name'], 'span should have resource.name attribute')
```
zacharycmontoya: Don't forget to assert against the `operation.name` and `resource.name` attributes after the RFC updates

packages/dd-trace/test/config/index.spec.js
```diff
@@ -303,6 +304,93 @@ describe('Config', () => {
...
+  it('should not warn when OTEL_EXPORTER_OTLP_TRACES_PROTOCOL is http/json', () => {
+    process.env.OTEL_EXPORTER_OTLP_TRACES_PROTOCOL = 'http/json'
+    getConfig()
+    const warnCall = log.warn.getCalls().find(
+      (call) => call.args[0]?.includes?.('OTEL_EXPORTER_OTLP_TRACES_PROTOCOL')
+    )
+    assert.strictEqual(warnCall, undefined)
+  })
+
+  it('should default sample rate to 1 when OTLP traces export is enabled', () => {
```
zacharycmontoya: nit: This test currently looks the same as the `should default OTEL_TRACES_SAMPLER to parentbased_always_on when OTEL_TRACES_EXPORTER is otlp` test case. Perhaps you could rewrite this one to assert that a configured `DD_TRACE_SAMPLE_RATE` setting takes precedence over the default 'parent_based_always_on' sampler?

packages/dd-trace/test/opentelemetry/traces.spec.js
```diff
@@ -0,0 +1,707 @@
...
+      assert.match(otlpSpan.spanId, /^[0-9a-f]+$/, 'spanId must be lowercase hex')
+      assert.strictEqual(typeof otlpSpan.parentSpanId, 'string', 'parentSpanId must be a string')
+      assert.strictEqual(otlpSpan.parentSpanId.length, 16, 'parentSpanId must be 16 hex chars (8 bytes)')
+    })
+
+    it('maps span kind correctly', () => {
+      const transformer = new OtlpTraceTransformer({})
+
+      const kinds = ['internal', 'server', 'client', 'producer', 'consumer']
+      const expected = [1, 2, 3, 4, 5]
```
zacharycmontoya: nit: I recommend using the named constants (e.g. the const SPAN_KIND_INTERNAL) rather than just integer values so it's easier to read the test case at first glance

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
```diff
@@ -0,0 +1,332 @@
...
+          attributes.push({ key, value: { doubleValue: value } })
+        }
+      }
+    }
+
+    // Add meta_struct as bytesValue attributes (JSON-serialized, base64-encoded per proto JSON mapping)
+    if (span.meta_struct) {
+      for (const [key, value] of Object.entries(span.meta_struct)) {
+        const bytes = Buffer.from(JSON.stringify(value))
+        attributes.push({ key, value: { bytesValue: bytes.toString('base64') } })
```
zacharycmontoya, line 185: Note for follow-up PR's: When setting properties that originated from the meta_struct dictionary, their resulting protobuf type should be `bytes`. So when we implement protobuf-based protocols, we shouldn't be calling `bytes.toString('base64')` so we'll need to do some refactoring when we get there

zacharycmontoya (dismissed):
From the requirements level, LGTM! I left a number of nits but nothing is blocking from my perspective. And this will also need approval from a Node.js expert

packages/dd-trace/src/config/index.js
ida613: thanks Zach :D [here](https://github.com/DataDog/dd-trace-js/pull/7531/commits/0d139adda1e4b96630cd1bbc5d24f7938799a116)

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
ida613: thank you Zach for this note! :D

integration-tests/opentelemetry-traces.spec.js
ida613: thanks Zach XD [here](https://github.com/DataDog/dd-trace-js/pull/7531/changes/8065a214feab334d716aac9684bb0f85fa427618) it is

packages/dd-trace/test/config/index.spec.js
ida613: thank you so much for spotting this Zach! I fixed it [here](https://github.com/DataDog/dd-trace-js/pull/7531/changes/e715a207f77cf8494e7d81279d37509e487d9014)

packages/dd-trace/test/opentelemetry/traces.spec.js
ida613: very true! [updated](https://github.com/DataDog/dd-trace-js/pull/7531/changes/e225831f53e51cf4e077d958bf6d051af435b2a7)

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
ida613: thanks Zach I'll keep it in mind ^ ^

packages/dd-trace/src/config/index.js
zacharycmontoya: Oh actually, I think the calculation was correct before so we should revert that. After checking these OTLP endpoint env vars, we should check the value of `DD_AGENT_HOST` and use that value if available (e.g. `http://{DD_AGENT_HOST}:4318/v1/traces`. Otherwise, the default value will fall back to `http://localhost:4318/v1/traces`. Sorry for my miscommunication

packages/dd-trace/src/opentelemetry/trace/otlp_http_trace_exporter.js
zacharycmontoya: Ah I see, thanks for the clarification 🙏🏼

packages/dd-trace/test/span_processor.spec.js
zacharycmontoya: I think the updated code makes sense, thanks!

packages/dd-trace/test/opentelemetry/traces.spec.js
zacharycmontoya: Yes feel free to add that logic in the subsequent grpc PR. No need to do that here since you're meeting the functional requirements

integration-tests/opentelemetry-traces.spec.js
```diff
@@ -0,0 +1,155 @@
...
+    // Validate non-error span status
+    assert.strictEqual(webSpan.status.code, 0, 'non-error span should have STATUS_CODE_UNSET')
+
+    // Validate span attributes include service.name and resource.name
+    const webAttrs = Object.fromEntries(
+      webSpan.attributes.map(({ key, value }) => [key, value])
+    )
+    assert.deepStrictEqual(webAttrs['service.name'], { stringValue: 'otlp-test-service' })
+    assert.deepStrictEqual(webAttrs['operation.name'], { stringValue: 'web.request' })
+    assert.deepStrictEqual(webAttrs['resource.name'], { stringValue: 'web.request' })
```
zacharycmontoya: Is the resource name expected to be `web.request`? I would have expected the resource name for this span to be something like `GET /api/test`

packages/dd-trace/src/config/index.js
ida613: okie dokes! thank you Zach for the clarification ^ ^

ida613: [updated](https://github.com/DataDog/dd-trace-js/pull/7531/changes/c3c849fa3a0761d83aa9b643a616e267a86ec343)

integration-tests/opentelemetry-traces.spec.js
ida613: sorry about the confusion Zach! [updated](https://github.com/DataDog/dd-trace-js/pull/7531/changes/2f4ecabb49b48b20f86c6c240db0867391626abf)

packages/dd-trace/src/config/index.js
zacharycmontoya: I'm so sorry, I have a really really small nitpick: In OpenTelemetry, the default endpoint value is [expected](https://opentelemetry.io/docs/specs/otel/protocol/exporter/#configuration-options) to be `localhost`, not the IP loopback address value `127.0.0.1`. Can you make sure it's specifically set to `localhost` so that we maintain compatibility?

ida613: reasonable! will update

ida613: [updated](https://github.com/DataDog/dd-trace-js/pull/7531/changes/58b85ba356b48b354e45fecd2cdc7a3376e63e2f)

zacharycmontoya: Thanks!

zacharycmontoya (dismissed):
LGTM. Thanks for addressing all my feedback 🙏🏼

integration-tests/helpers/fake-agent.js
rochdev: I would expect not if the OTLP endpoint is on the agent to begin with. If it's not on the Datadog agent, we should probably have a specialized OTel fake agent instead.

packages/dd-trace/src/telemetry/telemetry.js
```diff
@@ -470,6 +470,11 @@ const nameMapping = {
   otelMetricsUrl: 'OTEL_EXPORTER_OTLP_METRICS_ENDPOINT',
   otelMetricsExportInterval: 'OTEL_METRIC_EXPORT_INTERVAL',
   otelMetricsTemporalityPreference: 'OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE',
+  otelTracesEnabled: 'OTEL_TRACES_EXPORTER',
```
rochdev: Is this accurate? The exporter is a string, not a boolean.

packages/dd-trace/src/proxy.js
```diff
@@ -268,7 +268,12 @@ class Tracer extends NoopProxy {
         const prioritySampler = config.apmTracingEnabled === false
           ? require('./standalone').configure(config)
           : undefined
-        this._tracer = new DatadogTracer(config, prioritySampler)
+        let otlpExporter
+        if (config.otelTracesEnabled) {
```
rochdev, line 283: There is a switch in the base tracer class to pick the right exporter, why set it externally for only this case?

packages/dd-trace/src/span_processor.js
```diff
@@ -52,6 +52,7 @@ class SpanProcessor {
     }
     if (started.length === finished.length || finished.length >= flushMinSpans) {
       this.sample(span)
+
```
rochdev:
```suggestion
```

packages/dd-trace/src/opentelemetry/trace/index.js
```diff
@@ -0,0 +1,78 @@
...
+  const env = config.env || config.tags.env
+  if (env) resourceAttributes['deployment.environment'] = env
+  const version = config.version || config.tags.version
+  if (version) resourceAttributes['service.version'] = version
+
+  if (config.tags) {
+    const filteredTags = { ...config.tags }
+    delete filteredTags.service
+    delete filteredTags.version
+    delete filteredTags.env
```
rochdev:
```suggestion
    const { service, version, env, ...filteredTags } = config.tags
```

packages/dd-trace/src/opentelemetry/trace/otlp_http_trace_exporter.js
```diff
@@ -0,0 +1,66 @@
...
+      return
+    }
+
+    const additionalTags = [`spans:${spans.length}`]
+    this.recordTelemetry('otel.traces_export_attempts', 1, additionalTags)
+
+    const payload = this.#transformer.transformSpans(spans)
+    this.sendPayload(payload, (result) => {
```
rochdev, line 58: Is it expected that we're sending each traces as their own payload?

rochdev (commented):
A few small things and then I think this is good to go!

packages/dd-trace/src/opentelemetry/trace/otlp_http_trace_exporter.js
zacharycmontoya: For our purposes today this is totally fine, since all of the spans have finished and have chosen to be sampled. If we want to export spans so that payloads look similar to the OpenTelemetry SDK's [Batch Processor](https://opentelemetry.io/docs/specs/otel/trace/sdk/#batching-processor) (sends all spans in the last X seconds to be exported in one payload), then we would need to revisit this

packages/dd-trace/src/opentelemetry/trace/index.js
ida613: thank you Roch, appreciate the call-out :P

packages/dd-trace/src/proxy.js
ida613: I wrote otlp_http_trace_exporter by emulating otlp_http_metric_exporter and otlp_http_log_exporter, which are child classes of OtlpHttpExporterBase. But I can certainly wrap this class and put it under packages/dd-trace/src/exporters if that's a better way :)

packages/dd-trace/src/telemetry/telemetry.js
ida613: make sense! I deleted this line

packages/dd-trace/src/opentelemetry/trace/otlp_http_trace_exporter.js
rochdev: Is this even usable in the real world though? It would result in potentially up to tens of thousands of requests the the collector per second.

packages/dd-trace/src/telemetry/telemetry.js
rochdev: I didn't necessarily mean to delete it entirely if we want to collect some telemetry, but what we report should be accurate and semantically correct.

packages/dd-trace/src/proxy.js
rochdev: Hmm that's a good point, we might want to go the opposite way to what I said and isolate it even more. Let me look into this a bit further.

rochdev: Actually as part of a separate effort we'll be revamping how exporters are registered and it won't be like either the current approaches, so let's keep it like this for now 👍

packages/dd-trace/src/telemetry/telemetry.js
ida613: thanks Roch :D [added it back](https://github.com/DataDog/dd-trace-js/blob/556b436908be3b8460c12c34ff4b88139b1caaa0/packages/dd-trace/src/config/index.js#L377-L384) as a boolean after rebasing on master. [setAndTrack](https://github.com/DataDog/dd-trace-js/blob/37aac5605beb6a08ea84b604846c55fccaed9d5c/packages/dd-trace/src/config/index.js#L122) calls [generateTelemetry](https://github.com/DataDog/dd-trace-js/blob/37aac5605beb6a08ea84b604846c55fccaed9d5c/packages/dd-trace/src/config/index.js#L140)

packages/dd-trace/src/opentelemetry/trace/otlp_http_trace_exporter.js
zacharycmontoya: I agree that performance could be a concern from a high-traffic production environment, but I would like for us to merge in this feature (it's opt-in) and start to get feedback on it

rochdev (approved):

integration-tests/helpers/fake-agent.js
BridgeAR: I don't know where we have the definition for the endpoint.

I would rather use something OTLP specific, if possible, since this seems to be only about that?

integration-tests/opentelemetry-traces.spec.js
```diff
@@ -0,0 +1,155 @@
...
+        if (code !== 0) {
+          reject(new Error(`Process exited with status code ${code}`))
+        } else {
+          resolve()
+        }
+      })
+    })
+
+    const { headers, payload } = await tracesPromise
+    await exitPromise
```
BridgeAR, line 65:
```suggestion
    const [{ headers, payload }] = await Promise.all([
      tracesPromise,
      exitPromise,
    ])
```

Otherwise this could end up as unhandled rejection

integration-tests/opentelemetry-traces.spec.js
```diff
@@ -0,0 +1,155 @@
...
+    const resource = resourceSpan.resource
+    assert.ok(resource, 'resourceSpan should have resource')
+    assert.ok(Array.isArray(resource.attributes), 'resource should have attributes array')
+
+    const resourceAttrs = Object.fromEntries(
+      resource.attributes.map(({ key, value }) => [key, value])
+    )
+    assert.deepStrictEqual(resourceAttrs['service.name'], { stringValue: 'otlp-test-service' })
+    assert.deepStrictEqual(resourceAttrs['deployment.environment'], { stringValue: 'test' })
+    assert.deepStrictEqual(resourceAttrs['service.version'], { stringValue: '1.0.0' })
```
BridgeAR, line 85: With these transformations, it is difficult to know how the actual output looks like.

Would it be possible to just use `assertObjectContains()`? I think that should be sufficient. If we want to guarantee a number of properties, we can also handle that by adding additional length assertions.

If the order is not guaranteed, we can sort the entries before passing it to `assertObjectContains`.

This also applies to the assertions below this code.

packages/dd-trace/src/config/supported-configurations.json
```diff
@@ -3886,6 +3886,47 @@
         "internalPropertyName": "otelHeaders"
       }
     ],
+    "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT": [
+      {
+        "implementation": "A",
+        "type": "string",
+        "default": null,
+        "internalPropertyName": "otelTracesUrl"
```
BridgeAR, line 3894: I would rather not add internalPropertyNames, if not needed. It is simpler to reason about these as env name. So removing it and using `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` would be good. The same about the other configs.

packages/dd-trace/src/config/index.js
```diff
@@ -600,6 +625,13 @@ class Config extends ConfigBase {
     if (!this.otelMetricsUrl) {
       setAndTrack(this, 'otelMetricsUrl', `http://${agentHostname}:${DEFAULT_OTLP_PORT}/v1/metrics`)
     }
+    if (!trackedConfigOrigins.has('otelTracesUrl') && this.OTEL_EXPORTER_OTLP_ENDPOINT) {
+      // Generic OTLP endpoint: per spec, append /v1/traces signal-specific subpath
+      setAndTrack(this, 'otelTracesUrl', this.OTEL_EXPORTER_OTLP_ENDPOINT.replace(/\/$/, '') + '/v1/traces')
+    } else if (!this.otelTracesUrl) {
+      const tracesHostname = agentHostname === '127.0.0.1' ? 'localhost' : agentHostname
```
BridgeAR, line 632: If this is important for traces, it would likely also be important for all other otel related URLs?

Is it defined somewhere that it has to be localhost and not 127.0.0.1?

packages/dd-trace/src/config/index.js
```diff
@@ -600,6 +625,13 @@ class Config extends ConfigBase {
     if (!this.otelMetricsUrl) {
       setAndTrack(this, 'otelMetricsUrl', `http://${agentHostname}:${DEFAULT_OTLP_PORT}/v1/metrics`)
     }
+    if (!trackedConfigOrigins.has('otelTracesUrl') && this.OTEL_EXPORTER_OTLP_ENDPOINT) {
+      // Generic OTLP endpoint: per spec, append /v1/traces signal-specific subpath
+      setAndTrack(this, 'otelTracesUrl', this.OTEL_EXPORTER_OTLP_ENDPOINT.replace(/\/$/, '') + '/v1/traces')
```
BridgeAR, line 630: If that is correct, the metrics URL handling is likely also wrong. Could you please verify that?

packages/dd-trace/src/config/index.js
```diff
@@ -374,6 +374,23 @@ class Config extends ConfigBase {
...
+      setAndTrack(this, 'otelTracesEnabled', otelTracesEnabled)
+    }
+
+    if (this.otelTracesProtocol && this.otelTracesProtocol !== 'http/json') {
+      log.warn(
+        'OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=%s is not yet supported; only http/json is currently implemented',
+        this.otelTracesProtocol
+      )
+      setAndTrack(this, 'otelTracesProtocol', 'http/json')
+    }
```
BridgeAR, line 392: Please replace that with an `allowed` entry in supported-configurations.json.


```suggestion
```

packages/dd-trace/src/config/index.js
```diff
@@ -600,6 +625,13 @@ class Config extends ConfigBase {
     if (!this.otelMetricsUrl) {
       setAndTrack(this, 'otelMetricsUrl', `http://${agentHostname}:${DEFAULT_OTLP_PORT}/v1/metrics`)
     }
+    if (!trackedConfigOrigins.has('otelTracesUrl') && this.OTEL_EXPORTER_OTLP_ENDPOINT) {
```
BridgeAR, line 628: I believe the whole if else could be replaced with a single if in case we add `OTEL_EXPORTER_OTLP_ENDPOINT` as fallback alias to `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` and just add the `/v1/traces` afterwards, if we detect the fallback being used. That would simplify the code quite a bit.

packages/dd-trace/src/config/index.js
```diff
@@ -374,6 +374,23 @@ class Config extends ConfigBase {
...
     }
 
+    const otelTracesEnabled = trackedConfigOrigins.has('OTEL_TRACES_EXPORTER') &&
+      this.OTEL_TRACES_EXPORTER === 'otlp'
+    if (this.protocolVersion && this.protocolVersion !== '0.4' && otelTracesEnabled) {
+      log.warn('DD_TRACE_AGENT_PROTOCOL_VERSION is set, disabling OTLP traces export')
+      setAndTrack(this, 'otelTracesEnabled', false)
+    } else {
+      setAndTrack(this, 'otelTracesEnabled', otelTracesEnabled)
+    }
```
BridgeAR, line 384: `otelTracesEnabled` is not a configuration. The config should be about plain config values. By using `setAndTrack` this would also emit telemetry, which would have to be special handled. That is definitely not ideal.

Instead, please move this logic as a simple if statement to the single usage spot.

packages/dd-trace/src/config/index.js
```diff
@@ -440,8 +457,16 @@ class Config extends ConfigBase {
...
     }
 
-    if (!trackedConfigOrigins.has('sampleRate') && trackedConfigOrigins.has('OTEL_TRACES_SAMPLER')) {
-      setAndTrack(this, 'sampleRate', getFromOtelSamplerMap(this.OTEL_TRACES_SAMPLER, this.OTEL_TRACES_SAMPLER_ARG))
+    if (!trackedConfigOrigins.has('sampleRate')) {
+      const effectiveSampler = (trackedConfigOrigins.has('OTEL_TRACES_EXPORTER') &&
+        this.OTEL_TRACES_EXPORTER === 'otlp' &&
+        !trackedConfigOrigins.has('OTEL_TRACES_SAMPLER'))
+        ? 'parentbased_always_on'
+        : this.OTEL_TRACES_SAMPLER
```
BridgeAR, line 465: I am not sure I understand this logic. If I am not mistaken reverting the whole block to the code as it was before would work identically. If not, I think it should be handled inside `getFromOtelSamplerMap` and we should closely check what state is the wanted one :).

packages/dd-trace/src/opentelemetry/otlp/otlp_http_exporter_base.js
```diff
@@ -116,11 +116,26 @@ class OtlpHttpExporterBase {
...
-   * @param {string} [headersString=''] - Comma-separated key=value pairs
+   * Parses additional HTTP headers from a comma-separated string or pre-parsed map.
+   * @param {string|Record<string, string>} [headersString=''] - Comma-separated key=value pairs or map
    * @returns {Record<string, string>} Parsed headers object
    */
   #parseAdditionalHeaders (headersString = '') {
+    if (headersString !== null && typeof headersString === 'object') {
+      // The config MAP parser uses tagger.add (which splits on ':'), so OTEL-format
+      // headers ('key=value') arrive with the full 'key=value' string as the map key
+      // and an empty string as the value. Re-split on '=' to get the correct pairs.
```
BridgeAR, line 127: We can adjust the parser instead of doing this re-mapping here.
That way our telemetry also looks correct. It would be wrong right now.

I think a good way of handling this would be to check for the name to start with `OTEL_` in the parser after removing the internalPropertyNames. If that is the case, we can use a otel specific parser. I do not know if the otel values could contain colons or not. My implementation would depend on that.

packages/dd-trace/src/opentelemetry/trace/index.js
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Builds resource attributes from the tracer configuration.
+ *
+ * @param {Config} config - Tracer configuration instance
+ * @returns {import('@opentelemetry/api').Attributes} Resource attributes
+ */
+function buildResourceAttributes (config) {
+  const resourceAttributes = {
+    'service.name': config.service || config.tags.service,
```
BridgeAR, line 36:
```suggestion
    'service.name': config.service,
```

This will already have incorporated the tag's service :)

packages/dd-trace/src/opentelemetry/trace/index.js
```diff
@@ -0,0 +1,75 @@
...
+    'telemetry.sdk.name': 'datadog',
+    'telemetry.sdk.version': VERSION,
+    'telemetry.sdk.language': 'nodejs',
+  }
+
+  const env = config.env || config.tags.env
```
BridgeAR, line 42:
```suggestion
  const env = config.env
```

The config already does that for us

packages/dd-trace/src/opentelemetry/trace/index.js
```diff
@@ -0,0 +1,75 @@
...
+  if (env) resourceAttributes['deployment.environment'] = env
+  const version = config.version || config.tags.version
```
BridgeAR, line 44:
```suggestion
  const version = config.version
```

The config already does that for us and it also picks up other ways for detecting the version. It is almost impossible that the version is not defined, while I guess we can keep the safety check below anyway

packages/dd-trace/src/opentelemetry/trace/index.js
```diff
@@ -0,0 +1,75 @@
...
+  if (version) resourceAttributes['service.version'] = version
+
+  if (config.tags) {
```
BridgeAR, line 47: This is guaranteed to be truthy

packages/dd-trace/src/opentelemetry/trace/index.js
```diff
@@ -0,0 +1,75 @@
...
+function buildResourceAttributes (config) {
+  const resourceAttributes = {
+    'service.name': config.service || config.tags.service,
+    'telemetry.sdk.name': 'datadog',
+    'telemetry.sdk.version': VERSION,
+    'telemetry.sdk.language': 'nodejs',
+  }
+
+  const env = config.env || config.tags.env
+  if (env) resourceAttributes['deployment.environment'] = env
```
BridgeAR, line 43: Can you please point me to the documentation for this? If I am not mistaken, it should be `deployment.environment.name`?

packages/dd-trace/src/proxy.js
```diff
@@ -279,7 +279,12 @@ class Tracer extends NoopProxy {
...
+          const { buildResourceAttributes, createOtlpTraceExporter } = require('./opentelemetry/trace')
+          otlpExporter = createOtlpTraceExporter(config, buildResourceAttributes(config))
```
BridgeAR, line 285: The attributes should be an internal handling of the trace file. The method already receives the config, so using the exported method here with config seems not ideal.

```suggestion
          const { createOtlpTraceExporter } = require('./opentelemetry/trace')
          otlpExporter = createOtlpTraceExporter(config)
```

packages/dd-trace/src/opentelemetry/trace/otlp_http_trace_exporter.js
```diff
@@ -0,0 +1,66 @@
...
+
+const OtlpHttpExporterBase = require('../otlp/otlp_http_exporter_base')
+const { SAMPLING_PRIORITY_KEY } = require('../../constants')
+const { AUTO_KEEP } = require('../../../../../ext/priority')
+const OtlpTraceTransformer = require('./otlp_transformer')
+
+/**
+ * OtlpHttpTraceExporter exports DD-formatted spans via OTLP over HTTP/JSON.
+ *
+ * This implementation follows the OTLP HTTP v1.7.0 specification:
```
BridgeAR, line 11: Please use the latest spec

packages/dd-trace/src/opentelemetry/trace/otlp_http_trace_exporter.js
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * Creates a new OtlpHttpTraceExporter instance.
+   *
+   * @param {string} url - OTLP endpoint URL
+   * @param {string} headers - Additional HTTP headers as comma-separated key=value string
+   * @param {number} timeout - Request timeout in milliseconds
+   * @param {import('@opentelemetry/api').Attributes} resourceAttributes - Resource attributes
+   */
+  constructor (url, headers, timeout, resourceAttributes) {
+    super(url, headers, timeout, 'http/json', '/v1/traces', 'traces')
```
BridgeAR, line 33: Passing through `/v1/traces` allows to simplify our config. It does not need any special handling in that case.

Should `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` not be used as protocol source?

packages/dd-trace/src/opentelemetry/trace/otlp_http_trace_exporter.js
BridgeAR: @ida613 can you please add a TODO to add the batch handler?

I would imagine this being something we soon have to get back to for all Otel related things.

@bm1549 FYI for prioritization

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
```diff
@@ -0,0 +1,332 @@
...
+ * @augments OtlpTransformerBase
+ */
+class OtlpTraceTransformer extends OtlpTransformerBase {
+  /**
+   * Creates a new OtlpTraceTransformer instance.
+   *
+   * @param {import('@opentelemetry/api').Attributes} resourceAttributes - Resource attributes
+   */
+  constructor (resourceAttributes) {
+    super(resourceAttributes, 'http/json', 'traces')
```
BridgeAR, line 70: I think the protocol should be defined by our new config, no?

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
```diff
@@ -0,0 +1,332 @@
...
+// DD meta keys that are mapped to dedicated OTLP span fields and should not appear as attributes
+const EXCLUDED_META_KEYS = new Set([
+  '_dd.span_links',
+  'span.kind',
+])
+
+/**
+ * OtlpTraceTransformer transforms DD-formatted spans to OTLP trace JSON format.
+ *
+ * This implementation follows the OTLP Trace v1.7.0 Data Model specification:
```
BridgeAR, line 54: Same as above: please use 1.10.0 (or is there a reason to use the older protocol?)

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
```diff
@@ -0,0 +1,332 @@
...
+    super(resourceAttributes, 'http/json', 'traces')
+  }
+
+  /**
+   * Transforms DD-formatted spans to OTLP JSON format.
+   *
+   * @param {DDFormattedSpan[]} spans - Array of DD-formatted spans to transform
+   * @returns {Buffer} JSON-encoded trace data
+   */
+  transformSpans (spans) {
```
BridgeAR, line 79: Is this used outside as well? If not, could we use a private method instead? :)

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
```diff
@@ -0,0 +1,332 @@
...
+      }
+    }
+
+    // Add metrics as numeric attributes
+    if (span.metrics) {
+      for (const [key, value] of Object.entries(span.metrics)) {
+        if (Number.isInteger(value)) {
+          attributes.push({ key, value: { intValue: value } })
+        } else {
+          attributes.push({ key, value: { doubleValue: value } })
```
BridgeAR, line 176: I guess this is fine, while I do not know if we could get invalid values here somehow. Do you know if that is possible?

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
BridgeAR: Can you please add a TODO entry to change this in the backend as well as here?

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  /**
+   * Maps a DD span.kind string to an OTLP SpanKind enum value.
+   *
+   * @param {string | undefined} kind - DD span kind string
+   * @returns {number} OTLP SpanKind enum value
+   */
+  #mapSpanKind (kind) {
+    if (!kind) return SPAN_KIND_UNSPECIFIED
```
BridgeAR, line 199:
```suggestion
```

This will work the same by just falling back below :)

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
```diff
@@ -0,0 +1,332 @@
...
+   * Maps DD span error state to an OTLP Status object.
+   *
+   * @param {DDFormattedSpan} span - DD-formatted span
+   * @returns {object} OTLP Status object with code and message
+   */
+  #mapStatus (span) {
+    if (span.error === 1) {
+      return {
+        code: STATUS_CODE_ERROR,
+        message: span.meta?.['error.message'] || '',
```
BridgeAR, line 213: I believe we also have other properties by default for regular errors. Can these somehow also be transported? Just relying on the message would loose information. One part is the `type`, if I remember correct.

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
```diff
@@ -0,0 +1,332 @@
...
+   * @returns {object} OTLP Link object
+   */
+  #transformLink (link) {
+    return {
+      traceId: this.#hexToBytes(link.trace_id, 16),
+      spanId: this.#hexToBytes(link.span_id, 8),
+      traceState: link.tracestate || '',
+      attributes: link.attributes && Object.keys(link.attributes).length > 0
+        ? this.transformAttributes(link.attributes)
+        : [],
```
BridgeAR, line 270:
```suggestion
      attributes: this.transformAttributes(link.attributes ?? {}),
```

That way the transformAttributes already handles it appropriately and we do not have to pay the extra price when attributes are defined :)

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
```diff
@@ -0,0 +1,332 @@
...
+    }
+    if (buffer.length > targetLength) {
+      return Buffer.from(buffer.slice(buffer.length - targetLength)).toString('hex')
+    }
+    // Pad with leading zeros to reach target length
+    const result = Buffer.alloc(targetLength)
+    const offset = targetLength - buffer.length
+    for (let i = 0; i < buffer.length; i++) {
+      result[offset + i] = buffer[i]
+    }
```
BridgeAR, line 298:
```suggestion
    result.copy(buffer, offset);
```

https://nodejs.org/api/buffer.html#bufcopytarget-targetstart-sourcestart-sourceend

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
```diff
@@ -0,0 +1,332 @@
...
+   * @param {object} identifier - DD Identifier object with toBuffer() method
+   * @returns {boolean} True if the identifier is all zeros
+   */
+  #isZeroId (identifier) {
+    const buffer = identifier.toBuffer()
+    for (let i = 0; i < buffer.length; i++) {
+      if (buffer[i] !== 0) return false
+    }
+    return true
+  }
```
BridgeAR, line 314: Please remove this and instead use `identifier.equals(zeroId)`, where zeroId would be a id created at the top that is buffered.

```suggestion
```

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
```diff
@@ -0,0 +1,332 @@
...
+      flags: link.flags,
+    }
+  }
+
+  /**
+   * Converts a DD Identifier object to a hex-encoded string of the specified byte length.
+   * Pads with leading zeros if the identifier buffer is shorter than the target.
+   * Per the OTLP http/json spec, trace-ids and span-ids must be hex-encoded strings.
+   *
+   * @param {object} identifier - DD Identifier object with toBuffer() method
```
BridgeAR, line 281: Please use the actual class as type :)

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
```diff
@@ -0,0 +1,332 @@
...
+    if (!Array.isArray(parsedLinks)) return []
+
+    return parsedLinks.map(link => this.#transformLink(link))
+  }
+
+  /**
+   * Transforms a single DD span link to an OTLP Link object.
+   *
+   * @param {object} link - DD span link with trace_id, span_id, attributes, flags, tracestate
+   * @returns {object} OTLP Link object
```
BridgeAR, line 261: Please update the types to be the precise ones. It is possible to import other classes as type :)

This also applies to other parts of the code.

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
```diff
@@ -0,0 +1,332 @@
...
+   * Per the OTLP http/json spec, trace-ids and span-ids must be hex-encoded strings.
+   *
+   * @param {string | undefined} hexString - Hex string to normalize
+   * @param {number} targetLength - Target byte length
+   * @returns {string} Hex-encoded string of the specified length
+   */
+  #hexToBytes (hexString, targetLength) {
+    if (!hexString) return '0'.repeat(targetLength * 2)
+    const cleanHex = hexString.startsWith('0x') ? hexString.slice(2) : hexString
+    return cleanHex.padStart(targetLength * 2, '0')
```
BridgeAR, line 328: Is the hexString a public defined one or an internal one? I am a bit uncertain about that.

Depending on that state, we could optimize this further.

BridgeAR (commented):
I just saw this got merged while I reviewed it 😓 

@ida613 could you please go through my comments and open a follow-up?

I am almost done, I just wanted to post this briefly.

packages/dd-trace/test/opentelemetry/traces.spec.js
```diff
@@ -0,0 +1,717 @@
+'use strict'
+
+// Increase max listeners to avoid warnings in tests
+process.setMaxListeners(50)
```
BridgeAR, line 4: Please rework the tests so that this is not triggered. This actively works around a safeguard.

packages/dd-trace/test/opentelemetry/traces.spec.js
```diff
@@ -0,0 +1,717 @@
...
+  })
+
+  afterEach(() => {
+    process.env = originalEnv
+    sinon.restore()
+  })
+
+  describe('Transformer', () => {
+    const OtlpTraceTransformer = require('../../src/opentelemetry/trace/otlp_transformer')
+    const { getProtobufTypes } = require('../../src/opentelemetry/otlp/protobuf_loader')
```
BridgeAR, line 136: Could a user access these methods in a similar way? Or are these internals for us?

Could we instead create a tracer and call Otel methods instead? That would be nearer to an integration-test, while the runtime would still be similar and the safety net would be higher, if I am not mistaken.

We would just mock some parts of that tracer / use a fake agent.

packages/dd-trace/src/proxy.js
```diff
@@ -279,7 +279,12 @@ class Tracer extends NoopProxy {
...
+        }
+        this._tracer = new DatadogTracer(config, prioritySampler, otlpExporter)
```
BridgeAR, line 287: We already pass down the config to `DatadogTracer`, so we could move this inside of the lower layer. Passing through the argument here seems not ideal to me.

packages/dd-trace/src/config/index.js
zacharycmontoya: Yes this endpoint logic should apply to all 3 OTel signals (traces, logs, and metric). The requirement for the host name `localhost` is defined here in the OTel specification: https://opentelemetry.io/docs/specs/otel/protocol/exporter/#configuration-options

packages/dd-trace/src/opentelemetry/trace/index.js
zacharycmontoya: Historically, the name of this attribute was previously `deployment.environment` and it has since been renamed `deployment.environment.name`. I authored the new system-test test cases to allow both [here](https://github.com/DataDog/system-tests/blob/a37b31bb9495036e438ebcb97d5818b51cb46d63/tests/otel/test_tracing_otlp.py#L52) however it has just occurred that this name is now **[stable](https://github.com/open-telemetry/semantic-conventions/pull/3584)**. We should migrate this to `deployment.environment.name` accordingly, and the internal documentation we have [here](https://docs.google.com/document/d/1HHcIAa9GzoMsQaOU_G9G19cwwvfXVaVf9UOi6yFKaT0/edit?tab=t.0#heading=h.rjs8v32j0uvx) documents this choice of name already.

packages/dd-trace/src/opentelemetry/trace/otlp_transformer.js
zacharycmontoya: This is a public defined one. The [traceID](https://github.com/open-telemetry/opentelemetry-proto/blob/85e63b1ad6d0667e48707e8c0a88f366e79a68ce/opentelemetry/proto/trace/v1/trace.proto#L96) must be a 16-byte array and the [spanID](https://github.com/open-telemetry/opentelemetry-proto/blob/85e63b1ad6d0667e48707e8c0a88f366e79a68ce/opentelemetry/proto/trace/v1/trace.proto#L104) must be a 8-byte array. And then for the [JSON export specifically](https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding) these fields are represented as lowercase hexadecimal strings rather than base64-encoded strings like other "bytes" fields would be encoded

PR #7533, opened 2026-02-13
ci: add rate limit monitoring to GitHub Actions

watson:
### What does this PR do?

Adds a reusable composite action (`.github/actions/rate-limit-status/`) that queries the GitHub API rate limit endpoint and logs the current usage (limit, used, remaining, reset time) to the workflow log. The action is then invoked before and after steps that consume the GitHub API across CI workflows and composite actions.

### Motivation

We've been seeing intermittent CI failures that may be caused by GitHub API rate limit exhaustion. By logging rate limit status around API-consuming steps (Codecov uploads, artifact uploads, `allcheckspassed`, CodeQL, `required-labels`, `heaviest-objects-in-the-universe`, `dd-octo-sts-action`, etc.), we can correlate failures with rate limit pressure and identify which steps are the biggest consumers.

### Additional Notes

- The action uses `if: always()` on the trailing invocations so we still capture rate limit status even when a preceding step fails.
- For jobs that don't already check out the repo (e.g. `all-green`, `pr-labels`, `yarn-dedupe`), a sparse checkout of just `.github/actions/rate-limit-status` is added to make the composite action available.
- This is a diagnostic/debugging addition with no impact on existing CI behavior -- the rate limit check is a single lightweight `curl` call to `/rate_limit`.

### DIFF
--- a/.github/actions/rate-limit-status/action.yml
+++ b/.github/actions/rate-limit-status/action.yml
@@ -1,0 +1,27 @@
+name: GitHub API Rate Limit Status
+description: Output current GitHub API rate limit usage to the workflow log.
+runs:
+  using: composite
+  steps:
+    - name: Check GitHub API rate limit
+      shell: bash
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+      run: |
+        response=$(curl -sS -H "Authorization: Bearer $GITHUB_TOKEN" \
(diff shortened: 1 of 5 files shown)


BridgeAR:
I believe this was mainly used for testing. Should we just close this?

PR #7535, opened 2026-02-13
feat: add durable-functions integration

ojproductions:
### What does this PR do?

Adds auto instrumentation for azure durable-functions package
* Instruments Activity and Entity operations
* Purposely does not instrument orchestrator functions
   * [It was found](https://datadoghq.atlassian.net/browse/SVLS-7644) orchestrators dont play well with instrumentation due to non-deterministic functionality
* Tests

### Motivation

[APMSVLS-332](https://datadoghq.atlassian.net/browse/APMSVLS-332)
[SVLS-7644](https://datadoghq.atlassian.net/browse/SVLS-7644)

### Additional Notes

[APMSVLS-332]: https://datadoghq.atlassian.net/browse/APMSVLS-332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SVLS-7644]: https://datadoghq.atlassian.net/browse/SVLS-7644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/datadog-plugin-azure-durable-functions/src/index.js
```diff
@@ -0,0 +1,49 @@
...
+  static get type () { return 'serverless' }
+  static get kind () { return 'server' }
+
+  bindStart (ctx) {
+    const span = this.startSpan(this.operationName(), {
+      kind: 'internal',
+      type: 'serverless',
+
+      meta: {
+        component: 'azure-functions',
```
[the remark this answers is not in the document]
jcstorms1: Yeah, are we considering these to be the same component as standard azure functions or should we make this have a more specific component tag for durable functions?

packages/datadog-instrumentations/src/azure-functions.js
```diff
@@ -6,11 +6,16 @@ const {
...
 } = require('./helpers/instrument')
 
+let alreadyPatched = false
+
 const azureFunctionsChannel = dc.tracingChannel('datadog:azure:functions:invoke')
 
 addHook({ name: '@azure/functions', versions: ['>=4'], patchDefault: false }, (azureFunction) => {
   const { app } = azureFunction
 
+  if (alreadyPatched) return azureFunction
```
ojproductions: for some reason, when using esm imports, azure functions gets patched twice, resulting in extra spans. hence I added this. Why does this happen?

pabloerhard: For ESM we always patch both the default export and the outer export.

In some cases, the default export and the outer export end up triggering a patch on the same object or function. For example, we observed this behavior when a patch was applied to the prototype of an object that was shared by both the default export and the outer export.

packages/datadog-plugin-azure-durable-functions/src/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+  end (ctx) {
+    // We only want to run finish here if this is a synchronous operation
+    // Only synchronous operations would have `result` or `error` on `end`
+    // So we skip operations that dont
+    if (!ctx.hasOwnProperty('result') && !ctx.hasOwnProperty('error')) return
+    super.finish(ctx)
+  }
+
+  asyncEnd (ctx) {
```
jcstorms1: I know it seems strange, but it is better to make this `asyncStart` for a more accurate trace. My initial intuition was to use these to capture the end of `traceAsync`, but I was corrected as well. 

`asyncStart` and `asyncEnd` are essentially occurring right after each other without anything happening in-between. In fact, `asyncEnd` isn't necessary in the majority of cases. The edge case would be if we need information from a call back function, which we do not patch in this integration. 

`start`
code patched
`end`
`asyncStart` 
if there was a callback it would execute here.
`asyncEnd`

packages/datadog-instrumentations/src/azure-functions.js
ojproductions: does the `patchDefault: false` option in addhook not affect this behavior at all?

packages/datadog-plugin-azure-durable-functions/src/index.js
```diff
@@ -0,0 +1,49 @@
...
+  static get prefix () { return 'tracing:datadog:azure:durable-functions:invoke' }
+  static get type () { return 'serverless' }
+  static get kind () { return 'server' }
+
+  bindStart (ctx) {
+    const span = this.startSpan(this.operationName(), {
+      kind: 'internal',
+      type: 'serverless',
+
+      meta: {
```
jcstorms1, line 17: I'm not sure if you have this in a doc or not, but I'd double check we have all the metadata we need and compare it with a normal function. I imagine a lot of the Azure tags won't be necessary for durable functions, but we should make sure there is as much in common as possible.

packages/datadog-instrumentations/src/azure-functions.js
```diff
@@ -6,11 +6,16 @@ const {
   addHook,
 } = require('./helpers/instrument')
 
+let alreadyPatched = false
```
jcstorms1: Does adding this functionality to `azure-functions.js` affect the durable integration? Curious why this was added. Is it the result of patching standard  and `esm` modules? It might be worth bringing this to the node guild for more innput.

packages/datadog-plugin-azure-durable-functions/test/integration-test/client.spec.js
```diff
@@ -0,0 +1,124 @@
...
+    afterEach(async () => {
+      // after each test, kill process and wait for exit before continuing
+      if (proc) {
+        proc.kill('SIGINT')
+        await new Promise(resolve => proc.on('exit', resolve))
+      }
+      await agent.stop()
+    })
+
+    it('is instrumented', async () => {
```
jcstorms1, line 45: Do we need a separate test for entity and non-entity functions or is this sufficient? [I believe there is separate logic in the case of an entity function and we should probably capture that.](https://github.com/DataDog/dd-trace-js/pull/7535/changes#diff-86a7d1865ef0df369f50e17ffc8fb2b58e23fd24a5004f28ed873ac4d3f63d96R25)

packages/datadog-plugin-azure-functions/test/integration-test/http-test/client.spec.js
```diff
@@ -77,7 +77,7 @@ describe('esm', () => {
 
 async function spawnPluginIntegrationTestProc (cwd, command, args, agentPort, stdioHandler, additionalEnvArgs = {}) {
   let env = {
-    NODE_OPTIONS: `--loader=${hookFile} func start`,
+    NODE_OPTIONS: `--loader=${hookFile}`,
```
jcstorms1, line 81: How does `func start` initialize now? I may have missed it.

jcstorms1 (commented):
Looks good overall. Left a few questions and a couple suggestions. Nice job!

packages/datadog-instrumentations/src/azure-functions.js
jcstorms1: I solved this in service bus by using a weakmap like [this](https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-instrumentations/src/azure-service-bus.js#L10). I'm not sure if that's the approach we want to take here though.

packages/datadog-plugin-azure-functions/test/integration-test/http-test/client.spec.js
ojproductions: I figured this was a typo, as the func start command is getting passed in on line [45](https://github.com/DataDog/dd-trace-js/blob/e547255b9886d5c41e0c41a250309d06c57e6d73/packages/datadog-plugin-azure-functions/test/integration-test/http-test/client.spec.js#L45) & [68](https://github.com/DataDog/dd-trace-js/blob/e547255b9886d5c41e0c41a250309d06c57e6d73/packages/datadog-plugin-azure-functions/test/integration-test/http-test/client.spec.js#L68), then initialized on [line 91](https://github.com/DataDog/dd-trace-js/blob/e547255b9886d5c41e0c41a250309d06c57e6d73/packages/datadog-plugin-azure-functions/test/integration-test/http-test/client.spec.js#L91). Does passing it to NODE_OPTIONS do anything?

packages/datadog-instrumentations/src/azure-functions.js
pabloerhard: I'm looking into it. Ideally patchDefault should prevent the patching of the defaultExport.

pabloerhard: I was not able to replicate it by running the client.spec.js in the http-test folder. I might need some more context on how the double patching was identified. 

Does removing that conditional make the tests fail?

packages/datadog-instrumentations/src/azure-functions.js
ojproductions: Me and @pabloerhard talked and it seems this may be due to how durable-functions is importing @azure/functions under the hood using require. He's working on a fix in #7601

packages/datadog-instrumentations/src/azure-functions.js
ojproductions: Pablo and I talked offline but I'm answering here for record. This is happening in the durable-functions tests not azure functions. Using esm imports produces an extra span. This is being addressed in #7601

packages/datadog-plugin-azure-durable-functions/test/integration-test/client.spec.js
ojproductions: Hi, I added a [few more asserts](https://github.com/DataDog/dd-trace-js/blob/69d1e320027d23d14653b8c40f338d8c8bfb2f5f/packages/datadog-plugin-azure-durable-functions/test/integration-test/client.spec.js#L63-L93) that should address the separate logic (which adds operationName to for entity functions). I think this should cover it but please lmk what you think

packages/datadog-plugin-azure-durable-functions/src/index.js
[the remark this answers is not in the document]
ojproductions: I believe we decided to keep them to the same component and name since durable-functions is basically just an extension of azure-functions. cc @pablomartinezbernardo please correct me if im wrong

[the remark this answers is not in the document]
pablomartinezbernardo: I don't think it makes sense from the point of view of the customer to have Functions and Durable Functions separately. 

> durable-functions is basically just an extension of azure-functions

But this was the whole process behind it, happy to hear other opinions though

packages/dd-trace/src/config/supported-configurations.json
```diff
@@ -2172,6 +2172,13 @@
         "default": "true"
       }
     ],
+    "DD_TRACE_AZURE_DURABLE_FUNCTIONS_ENABLED": [
+      {
+        "implementation": "B",
```
jcstorms1, line 2187: Is this "B" because it is an extension of azure functions?

ojproductions: no, In the feature parity [dashboard](https://feature-parity.us1.prod.dog/configurations?viewType=configurations&configSearch=DD_TRACE_AZURE&configKey=2299) I had to create a new version "B" of the config to add a default value, and then update this to use that version

packages/datadog-instrumentations/src/azure-durable-functions.js
pabloerhard, line 1: Any reason on why there is no .hasSubscribers guard? The other azure instrumentation seem to also not have one so this might be more of a general design question.

ojproductions: I hadn't heard of it before. Just looked into it and If I understand correctly this would prevent unnecessary logic when the tracing channels are not being used?

pabloerhard: Yep, it would involve adding something like this inside your wrapper functions

```
    if (!azureDurableFunctionsChannel.start.hasSubscribers) {
      return handler.apply(this, args)
    }
```

ojproductions: sg. I've added the guards in

packages/datadog-instrumentations/src/azure-durable-functions.js
```diff
@@ -0,0 +1,75 @@
...
+      shimmer.wrap(arg, 'handler', handler => entityHandler(handler, entityName))
+    }
+
+    return method.apply(this, arguments)
+  }
+}
+
+function entityHandler (handler, entityName) {
+  return function () {
+    const entityContext = arguments[0]
```
pabloerhard, line 44: I think it's better to add the .hasSubscribers guard at this level, this way the check happens at invocation time.

```
function entityHandler (handler, entityName) {
  return function () {
    if (!azureDurableFunctionsChannel.hasSubscribers) return handler.apply(this, arguments)
    const entityContext = arguments[0]
    return azureDurableFunctionsChannel.traceSync(
      handler,
      { trigger: 'Entity', functionName: entityName, operationName: entityContext?.df?.operationName },
      this, ...arguments)
  }
}
```

Same would apply for the activityHandler

packages/dd-trace/src/config/supported-configurations.json
tlhunter: It looks like the value has never existed before? If so I would think A would be the anticipated value.

Usually we do an A to B switch for something like tracers switching from being disabled by default to enabled by default.

packages/datadog-instrumentations/src/azure-durable-functions.js
ojproductions: makes sense. I moved the guards

packages/dd-trace/src/config/supported-configurations.json
ojproductions: @tlhunter In the feature parity dashboard, [this config](https://feature-parity.us1.prod.dog/configurations?viewType=configurations&configSearch=DD_TRACE_AZURE&configKey=2299) has Version A with default value:null and version B with default value:true. Im not entirely sure how the implementation works but I assume we want to use the second one no? looking at other configs it looks like we can create new versions to switch whether or not its enabled by default

tlhunter: @bm1549 can you help us out here? It looks like a bug in the dashboard that two versions are listed despite there being no associated tracer languages.

mhlidd: Makes sense that there are no tracer languages associated as @ojproductions added this env var to the FPD. The tracer version will be added automatically once the next NodeJS release is made.

@ojproductions seems like you added both Configuration versions 🤔. Was one of them by accident? Regardless, list the one that matches the behavior expected in dd-trace-js.

ojproductions: Yes the first was by accident, and then it didnt let me edit it so I made a new version B, which is the one Im targetting

mhlidd: Sounds good, use `B` in the local file. 

Not being allowed to edit existing versions is intended, as in the past devs have modified existing versions, unknowingly breaking other SDK CIs w/ the change.

PR #7536, opened 2026-02-13
ci: cache resolved Node.js version in CI

watson:
### What does this PR do?

Caches the resolved Node.js version in the shared `.github/actions/node/action.yml` composite action so that `actions/setup-node` can skip its GitHub API manifest lookups on subsequent runs.

When `check-latest: true` is set, `setup-node` makes two GitHub API requests to resolve the latest patch for a given major version. With hundreds of parallel CI jobs all doing this, the requests count towards our GitHub API rate limit. This PR introduces a lightweight `actions/cache` step that persists the resolved version to a file. On cache hit, the exact version is passed directly to `setup-node` with `check-latest: false`, eliminating both API calls entirely.

The cache key rotates every 20 minutes (based on epoch time) so new Node.js patch releases are still picked up promptly.

### Motivation

Every invocation of `setup-node` with `check-latest: true` triggers two GitHub REST API requests to resolve the latest version. Across our highly parallel CI matrix these add up and count towards our GitHub API rate limit. Caching the resolved version removes this overhead for the vast majority of runs.

### Additional Notes

- Only `.github/actions/node/action.yml` is changed; no workflow files need updating since they all delegate to this composite action.
- The cache is keyed on OS, architecture, and the requested Node.js version, so different matrix entries don't interfere with each other.
- On the first run (or after the 20-minute window rotates), `setup-node` behaves exactly as before (`check-latest: true`), then the resolved version is saved for subsequent runs.

### DIFF
--- a/.github/actions/node/action.yml
+++ b/.github/actions/node/action.yml
@@ -8,17 +8,54 @@ inputs:
 runs:
   using: composite
   steps:
-    - uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
-      with:
-        node-version: ${{
+    # Resolve the version from the input alias so we can use it in cache keys.
+    - name: Resolve Node.js version
+      id: node-version
+      env:
+        NODE_VERSION: ${{
           inputs.version == 'eol' && '16' ||
           inputs.version == 'oldest' && '18' ||
           inputs.version == 'maintenance' && '20' ||
           inputs.version == 'active' && '22' ||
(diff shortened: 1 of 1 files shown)


watson:
* **#7536** <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-js/7536?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-js/7536?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #7539, opened 2026-02-16
switch all-green to custom script with less api calls

rochdev:
### What does this PR do?

Switch All-Green to custom script with less API calls.

### Motivation

The action we are currently using for All-Green makes a lot of API calls as it has to paginate through every workflows and jobs and status checks and validates them individually. This is only needed once all jobs have completed, so the new script only checks for the number of in-progress check runs, and only paginates once when all runs are completed instead of every time.

### Additional Notes

This should result in ~22 API calls at most for a PR commit compared to ~400 right now.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -685,9 +685,9 @@
   integrity sha512-JaaYhsX4CA/QTIXdQfDpD9zcxew5EHuWxJjL+Ctp3L4FXC2gEAZ3s1DaCJU6aFOF285dJH3zAoeU64EJvxCaAw==
 
 "@opentelemetry/api-logs@<1.0.0":
-  version "0.213.0"
-  resolved "https://registry.yarnpkg.com/@opentelemetry/api-logs/-/api-logs-0.213.0.tgz#c7abc7d3c4586cfbfd737c0a2fcfb2323a9def75"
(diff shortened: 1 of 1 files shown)


scripts/all-green.mjs
```diff
@@ -0,0 +1,86 @@
...
+  const completedRuns = await octokit.paginate(
+    'GET /repos/:owner/:repo/commits/:ref/check-runs',
+    {
+      ...params,
+      per_page: 100,
+      status: 'completed',
+    }
+  )
+
+  return !completedRuns.some(run => (
```
BridgeAR: What about logging these at the end in case of a failure? I think that would be very helpful. It is otherwise unclear what is happening.

rochdev: This only reports the status of the jobs that are already in the PR, so they are already available and obvious. I figured this was not worth the effort.

rochdev: Is this a blocker? I'd prefer to do this in another PR if you really want it so we can fix the immediate problem and figure out the UI of this in a second step.

BridgeAR: Since we already lowered the number of requests and we have not run into the limit right now, I think it's best to include that before merging. It otherwise removes information that was formerly there.

rochdev: I restored the output, please have another look.

PR #7548, opened 2026-02-17
chore(eslint): enable rules globally and fix violations

watson:
### What does this PR do?

Enables several ESLint rules in the **global** config so they apply to all files (not only `src` dirs), and fixes existing violations across the repo. The shared `GLOBAL_RESTRICTED_REQUIRES` list is extracted and reused so restricted-require config stays in one place.

Rules now applied globally:
- `eslint-rules/eslint-safe-typeof-object`
- `no-implicit-coercion`
- `no-useless-assignment`
- `no-void`
- `operator-assignment`
- `prefer-exponentiation-operator`
- `prefer-object-has-own`
- `n/no-restricted-require` (using `GLOBAL_RESTRICTED_REQUIRES`)

Those rules are removed from the `dd-trace/src/all` block to avoid duplication (except `n/no-restricted-require` which adds `semver` as a restricted require to `src` only).

Fix tests and app code for the new global rules:
- Safe typeof checks
- Remove useless assignments
- Use `**` and template literals
- Use `filter.includes()` instead of `~indexOf`
- Fix typo in vertexai test (`} if` → `} else if`)

### Motivation

Unify lint rules across the codebase so test and non-src files follow the same style and safety rules.

### DIFF
--- a/packages/datadog-plugin-next/test/app/api/appDir/error/route.js
+++ b/packages/datadog-plugin-next/test/app/api/appDir/error/route.js
@@ -1,11 +1,7 @@
 import { NextResponse } from 'next/server'
 
 export async function GET (req) {
-  try {
-    throw new Error('error in app dir api route')
-  } catch (error) {
-    req.error = error
-  }
+  req.error = new Error('error in app dir api route')
 
   return NextResponse.json({}, { status: 500 })
 }
--- a/packages/dd-trace/test/appsec/iast/taint-tracking/plugin.spec.js
+++ b/packages/dd-trace/test/appsec/iast/taint-tracking/plugin.spec.js
@@ -54,25 +54,29 @@ describe('IAST Taint tracking plugin', () => {
   })
 
   it('Should subscribe to body parser, qs, cookie and process_params channel', () => {
-    assert.strictEqual(taintTrackingPlugin._subscriptions.length, 17)
-    let i = 0
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:body-parser:read:finish')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:multer:read:finish')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:fastify:body-parser:finish')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'apm:express:middleware:next')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:query:read:finish')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:fastify:query-params:finish')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:express:query:finish')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:cookie:parse:finish')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:fastify-cookie:read:finish')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:sequelize:query:finish')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'apm:pg:query:finish')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:express:process_params:start')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:router:param:start')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:fastify:path-params:finish')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'apm:graphql:resolve:start')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:url:parse:finish')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i]._channel.name, 'datadog:url:getter:finish')
+    const expectedChannels = [
+      'datadog:body-parser:read:finish',
+      'datadog:multer:read:finish',
+      'datadog:fastify:body-parser:finish',
+      'apm:express:middleware:next',
+      'datadog:query:read:finish',
+      'datadog:fastify:query-params:finish',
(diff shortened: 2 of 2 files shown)


packages/dd-trace/test/appsec/iast/taint-tracking/plugin.spec.js
```diff
@@ -72,7 +72,7 @@ describe('IAST Taint tracking plugin', () => {
     assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:fastify:path-params:finish')
     assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'apm:graphql:resolve:start')
     assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:url:parse:finish')
-    assert.strictEqual(taintTrackingPlugin._subscriptions[i++]._channel.name, 'datadog:url:getter:finish')
+    assert.strictEqual(taintTrackingPlugin._subscriptions[i]._channel.name, 'datadog:url:getter:finish')
```
BridgeAR: Nit: when working on this code: I would just use an array of expected names and iterate through them. That is nicer to read than having a big bunch of assertion calls.

packages/datadog-plugin-google-cloud-pubsub/test/index.spec.js
```diff
@@ -602,7 +602,7 @@ describe('Plugin', () => {
           prefixedResource = resource
         }
 
-        let defaultOpName = 'pubsub.receive'
+        let defaultOpName
```
simon-id, line 605: sure it's functionally useless, but it's useful for human reading. It shows that this variable should always have a value and never be empty. Either add a typing or a comment or something, otherwise you're reducing readability imo ?

packages/datadog-plugin-next/test/app/api/appDir/error/route.js
```diff
@@ -1,13 +1,11 @@
 import { NextResponse } from 'next/server'
 
 export async function GET (req) {
-  let status = 200
   try {
     throw new Error('error in app dir api route')
   } catch (error) {
     req.error = error
-    status = 500
   }
```
BridgeAR:
```suggestion
  req.error = new Error('error in app dir api route')
```

packages/dd-trace/test/appsec/iast/utils.js
```diff
@@ -201,8 +201,8 @@ function checkVulnerabilityInRequest (
       assert.notStrictEqual(vulnerabilitiesTrace, null)
       const vulnerabilitiesCount = new Map()
       vulnerabilitiesTrace.vulnerabilities.forEach(v => {
-        let count = vulnerabilitiesCount.get(v.type) || 0
-        vulnerabilitiesCount.set(v.type, ++count)
+        const count = (vulnerabilitiesCount.get(v.type) || 0) + 1
```
simon-id, line 204:
```suggestion
        const count = vulnerabilitiesCount.get(v.type) + 1 || 1
```

packages/datadog-plugin-google-cloud-pubsub/test/index.spec.js
simon-id: and this apply to all the places you removed the useless "default" assignements. I think they're worth either a typing or a comment

watson: I don't agree. Reading the prior version of this, you'd think there was a scenario where `defaultOpName` would not be overwritten below and where the value would therefore be `pubsub.receive`. That's not the case. We probably have hundreds of places in our code where we create a variable using `let` with no value, only to set it below in an if-elseif-else block. It's a very common pattern and that shouldn't confuse the reader.

packages/dd-trace/test/appsec/iast/utils.js
watson: That's not identical. If `vulnerabilitiesCount.get(v.type)` is `undefined` we need the fallback to `0`.

BridgeAR (dismissed):
LGTM, just left two suggestions

packages/dd-trace/test/appsec/iast/taint-tracking/plugin.spec.js
watson: Fixed 👍

packages/datadog-plugin-next/test/app/api/appDir/error/route.js
watson: Fixed 👍

packages/dd-trace/test/appsec/iast/utils.js
simon-id: Before suggesting I of course double checked it:
```javascript
> (undefined || 0) + 1
1
> undefined + 1 || 1
1
```

packages/dd-trace/test/profiling/profilers/poisson.spec.js
```diff
@@ -181,7 +181,7 @@ describe('PoissonProcessSamplingFilter', () => {
     const initialCount = filter.samplingInstantCount
     for (let i = 0; i < 5; i++) {
       nowValue += 20
-      const event = { startTime: 0, duration: filter.nextSamplingInstant - 0 }
+      const event = { startTime: 0, duration: filter.nextSamplingInstant }
```
simon-id, line 184: would - 0 be a way to coalesce the value to a number ? sounds improbable that someone actually wrote - 0 without a reason no ?

packages/dd-trace/test/appsec/iast/utils.js
watson: That's only because `NaN` is falsy. I don't think relying on that is such a good idea. Our TS checker would definitely not allow it.

packages/dd-trace/test/profiling/profilers/poisson.spec.js
watson: Yes, I had the same thought and looked over the code to see if `nextSamplingInstant` would ever not be a number. As far as I can see, `nextSamplingInstant` is ALWAYS a number.

watson: ... and the tests pass, so 👍

packages/dd-trace/test/appsec/iast/utils.js
BridgeAR: It is also more difficult to process in my opinion. We have to know that the OR is of higher precedence and that this works due to being coerced to NaN which is falsy.
The brackets are straight forward to understand.

packages/dd-trace/test/profiling/profilers/poisson.spec.js
BridgeAR: This is in a test case, so if there is an issue with the test data, it would surface here. Neither before, nor after would it detect actual faulty user input that could theoretically cause issues.

Since the test passes, the simplification seems good.

PR #7556, opened 2026-02-17
feat(llmobs): support custom telemetry metadata in Vercel AI SDK spans

sabrenner:
### What does this PR do?
Closes #7322 (by porting its commits and cleaning it up)

### Motivation
Adds support for recording telemetry metadata for Vercel AI SDK telemetry on Datadog LLM Observability span events' metadata field.


packages/dd-trace/src/llmobs/plugins/ai/index.js
```diff
@@ -197,8 +198,10 @@ class VercelAILLMObsPlugin extends BaseLLMObsPlugin {
 
     this._tagger.tagTextIO(span, parsedInputs, output)
 
-    const metadata = getGenerationMetadata(tags)
-    this._tagger.tagMetadata(span, metadata)
+    const generationMetadata = getGenerationMetadata(tags) || {}
+    const telemetryMetadata = getTelemetryMetadata(tags) || {}
+    const metadata = { ...generationMetadata, ...telemetryMetadata }
+    this._tagger.tagMetadata(span, Object.keys(metadata).length ? metadata : null)
```
BridgeAR: Before, this was not checking for tag existence. Is that intentional?

packages/dd-trace/src/llmobs/plugins/ai/index.js
```diff
@@ -197,8 +198,10 @@ class VercelAILLMObsPlugin extends BaseLLMObsPlugin {
 
     this._tagger.tagTextIO(span, parsedInputs, output)
 
-    const metadata = getGenerationMetadata(tags)
-    this._tagger.tagMetadata(span, metadata)
+    const generationMetadata = getGenerationMetadata(tags) || {}
+    const telemetryMetadata = getTelemetryMetadata(tags) || {}
+    const metadata = { ...generationMetadata, ...telemetryMetadata }
```
BridgeAR: `getGenerationMetadata` seems to always create a new object and it seems to be always called by the three changed spots.
We could therefore just call `getTelemetryMetadata` inside that method and add it directly to the newly created object. That way there is less overhead and less code edits needed.

packages/dd-trace/src/llmobs/plugins/ai/index.js
```diff
@@ -270,7 +283,9 @@ class VercelAILLMObsPlugin extends BaseLLMObsPlugin {
 
     this._tagger.tagLLMIO(span, parsedInputMessages, outputMessage)
 
-    const metadata = getModelMetadata(tags)
+    const modelMetadata = getModelMetadata(tags)
+    const telemetryMetadata = getTelemetryMetadata(tags)
```
BridgeAR: This seems to be a similar case to `getGenerationMetadata`. I would just move the `getTelemetryMetadata` call inside that method and add the entries directly.

sabrenner: agreed. addressed in https://github.com/DataDog/dd-trace-js/pull/7556/changes/4c6e9fb483570e390ca0132116dabf30dfddd096 i believe!

packages/dd-trace/src/llmobs/plugins/ai/index.js
sabrenner: this has been reverted as a result of applying other review comments

packages/dd-trace/src/llmobs/plugins/ai/index.js
sabrenner: yep, agreed, i have changed it to also do a pass on the telemetry metadata tags in https://github.com/DataDog/dd-trace-js/pull/7556/changes/4c6e9fb483570e390ca0132116dabf30dfddd096. lmk what you think!

packages/dd-trace/src/llmobs/plugins/ai/util.js
```diff
@@ -107,17 +111,28 @@ function getJsonStringValue (str, defaultValue) {
...
-    const metadataValue = tags[metadataTagKey]
-    if (metadataValue) {
-      modelMetadata[metadata] = metadataValue
+  for (const tag of Object.keys(tags)) {
+    const isModelMetadata = tag.startsWith(VERCEL_AI_MODEL_METADATA_PREFIX)
+    const isTelemetryMetadata = tag.startsWith(VERCEL_AI_TELEMETRY_METADATA_PREFIX)
+    if (!isModelMetadata && !isTelemetryMetadata) continue
+
+    if (isModelMetadata) {
+      const metadataKey = tag.split('.').pop()
```
BridgeAR:
```suggestion
      const lastCommaPosition = tag.lastIndexOf('.')
      const metadataKey = lastCommaPosition === -1 ? tag :  tag.slice(lastCommaPosition + 1)
```

packages/dd-trace/src/llmobs/plugins/ai/util.js
```diff
@@ -107,17 +111,28 @@ function getJsonStringValue (str, defaultValue) {
...
   const modelMetadata = {}
-  for (const metadata of MODEL_METADATA_KEYS) {
-    const metadataTagKey = `gen_ai.request.${metadata}`
-    const metadataValue = tags[metadataTagKey]
-    if (metadataValue) {
-      modelMetadata[metadata] = metadataValue
+  for (const tag of Object.keys(tags)) {
+    const isModelMetadata = tag.startsWith(VERCEL_AI_MODEL_METADATA_PREFIX)
+    const isTelemetryMetadata = tag.startsWith(VERCEL_AI_TELEMETRY_METADATA_PREFIX)
+    if (!isModelMetadata && !isTelemetryMetadata) continue
```
BridgeAR:
```suggestion
    const isModelMetadata = tag.startsWith(VERCEL_AI_MODEL_METADATA_PREFIX)
    if (!isModelMetadata) {
      const isTelemetryMetadata = tag.startsWith(VERCEL_AI_TELEMETRY_METADATA_PREFIX)
      if (!isTelemetryMetadata) {
        continue
      }
    }
```

packages/dd-trace/src/llmobs/plugins/ai/util.js
```diff
@@ -134,14 +150,23 @@ function getGenerationMetadata (tags) {
   const metadata = {}
 
   for (const tag of Object.keys(tags)) {
-    if (!tag.startsWith('ai.settings')) continue
+    const isGenerationMetadata = tag.startsWith(VERCEL_AI_GENERATION_METADATA_PREFIX)
+    const isTelemetryMetadata = tag.startsWith(VERCEL_AI_TELEMETRY_METADATA_PREFIX)
+    if (!isGenerationMetadata && !isTelemetryMetadata) continue
```
BridgeAR:
```suggestion
    const isGenerationMetadata = tag.startsWith(VERCEL_AI_GENERATION_METADATA_PREFIX)
    if (!isGenerationMetadata) {
      const isTelemetryMetadata = tag.startsWith(VERCEL_AI_TELEMETRY_METADATA_PREFIX)
      if (!isTelemetryMetadata) {
        continue
      }
    }
```

packages/dd-trace/src/llmobs/plugins/ai/util.js
```diff
@@ -134,14 +150,23 @@ function getGenerationMetadata (tags) {
...
 
-    const settingKey = tag.split('.').pop()
-    const transformedKey = settingKey.replaceAll(/[A-Z]/g, letter => '_' + letter.toLowerCase())
-    if (MODEL_METADATA_KEYS.has(transformedKey)) continue
+    if (isGenerationMetadata) {
+      const settingKey = tag.split('.').pop()
```
BridgeAR:
```suggestion
      const lastCommaPosition = tag.lastIndexOf('.')
      const settingKey = lastCommaPosition === -1 ? tag :  tag.slice(lastCommaPosition + 1)
```

BridgeAR (dismissed):
LGTM, I just left a few suggestions for a faster implementation. There is still a bit of code duplication, while abstracting these would likely not improve much, so I think that is good as is.

packages/dd-trace/src/llmobs/plugins/ai/util.js
sabrenner: question: does this constitute a significant performance improvement (other than just not creating one bool/doing the corresponding check)? i had it the other way intentionally because i don't love all the nesting haha

BridgeAR: It won't be a significant change here, no (because other things are just much slower). When writing generic performant code, it gives others the chance to also write code that is fast in all cases. That way we reduce overhead in general. I personally can also process things easier when I know what is actually relevant for the code execution and what not.

sabrenner: i refactored it to be something like this 😄

PR #7557, opened 2026-02-17
chore(graphql): improve path logic in graphql

pabloerhard:
Improve path handling logic in graphql to avoid performing duplicatge operations. This PR reduces unnecessary processiong of the path property.


pabloerhard:
> If this is addressing a performance issue, can you please add a benchmark that highlights the improvement?

Did this more a microbenchmark kind of script (with claude help), comparing old function vs new function, since on an outer scope it will probably be harder to visualize the improvement.

pathToArray changes will always be faster considering it went from processing a path to an array to just processing it once and passing it through. 

These results include multiple depths

| Function | Old (ms) | New (ms) | Improvement |
|---|---:|---:|---:|
| getPath (collapse) | 62.3 | 23.8 | **+62%** |
| getPath (no-collapse) | 33.6 | 4.2 | **+87%** |
| getParentField (worst-case) | 857.8 | 340.3 | **+60%** |
| getParentField (best-case) | 133.4 | 39.1 | **+71%** |
| getParentField (miss) | 937.1 | 429.9 | **+54%** |

PR #7565, opened 2026-02-18
chore: add packaging reminder to eslint conf

simon-id:
### What does this PR do?
Adds a little comment to remind ourselves to package our eslint conf in a few quarters, as decided in the guild meeting.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,3 +1,7 @@
+// TODO: once we've well iterated on this file (6 months+),
+// we can plan to move it to its own package to reuse elsewhere.
+// written on 2026-02-12
+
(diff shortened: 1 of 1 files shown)


eslint.config.mjs
```diff
@@ -1,3 +1,7 @@
+// TODO: once we've well iterated on this file (6 months+),
+// we can plan to move it to its own package to reuse elsewhere.
+// written on 2026-02-12
```
BridgeAR, line 3: Ideally this would be written as https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/expiring-todo-comments.md#expiry-date

PR #7567, opened 2026-02-18
feat(llmobs): Add support for LangGraph

crysmags:
### What does this PR do?
 Adds APM tracing and LLM Observability instrumentation for @langchain/langgraph (>=1.1.2).

### Motivation
LangGraph is a widely used framework for building stateful, multi-step LLM workflows. Instrumenting it gives users           out-of-the-box visibility into graph execution.

### Additional Notes
`Pregel.stream()` is the single instrumentation point since `invoke()` calls it internally, covering both execution modes. Stream inputs are captured at call time and the last emitted chunk is tagged as output. Each graph node produces a child workflow span named after that node.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2580,14 +2580,14 @@ declare namespace tracer {
 
     /**
      * This plugin automatically instruments the
-     * [ldapjs](https://github.com/ldapjs/node-ldapjs/) module.
-     */
-    /**
-     * This plugin automatically instruments the
      * [langgraph](https://github.com/npmjs/package/langgraph) library.
      */
     interface langgraph extends Instrumentation {}
 
+      /**
+     * This plugin automatically instruments the
+     * [ldapjs](https://github.com/ldapjs/node-ldapjs/) module.
+     */
     interface ldapjs extends Instrumentation {}
 
     /**


packages/datadog-plugin-langgraph/src/stream.js
```diff
@@ -0,0 +1,46 @@
+'use strict'
+
+const TracingPlugin = require('../../dd-trace/src/plugins/tracing')
+class PregelStreamPlugin extends TracingPlugin {
+  static id = 'langgraph'
+  static prefix = 'tracing:orchestrion:@langchain/langgraph:Pregel_stream'
+
+  bindStart (ctx) {
+    this.startSpan('LangGraph', {
```
wconti27, line 13: @sabrenner any tags that should be getting set on these spans?

packages/datadog-plugin-langgraph/src/stream.js
```diff
@@ -0,0 +1,46 @@
+'use strict'
```
wconti27, line 1: also i dnot see the tracing for langgraph.pregel.invoke, did a file not get included?

packages/datadog-plugin-langgraph/src/stream.js
sabrenner: nope, we only tag model name and provider for apm spans and they're not relevant for langgraph

packages/datadog-plugin-langgraph/test/index.spec.js
```diff
@@ -0,0 +1,116 @@
...
+
+  afterEach(async () => {
+    await agent.close({ ritmReset: false })
+  })
+
+  describe('Pregel.invoke() - invoke', () => {
+    it('should generate span with correct tags (happy path)', async () => {
+      const traceAssertion = agent.assertSomeTraces((traces) => {
+        const allSpans = traces.flat()
+        const invokeSpan = allSpans.find(span => span.name === 'LangGraph')
```
wconti27: span.name = LangGraph?

packages/datadog-plugin-langgraph/test/index.spec.js
```diff
@@ -0,0 +1,116 @@
+'use strict'
```
wconti27, line 1: I dont see pregel.invoke tracing, this test file is using spans based on name Langgraph, which is not only missing the actual operation (langgraph.stream, langgraph.invoke, or whatever), but it is really only testing the stream plugin. also i think if span names match between langgraph invoke and stream, we should set a stream tag = True on streams, since we need some way to differentiate them

packages/dd-trace/test/plugins/externals.json
```diff
@@ -0,0 +1,782 @@
+{
```
wconti27: woah, whole file got reformatted 😢

docs/add-redirects.sh
sabrenner, line 1: looks like the addition of this file might be unrelated?

packages/dd-trace/src/llmobs/plugins/langgraph/index.js
```diff
@@ -0,0 +1,151 @@
...
+
+    const input = inputs !== undefined && inputs !== null ? this.formatIO(inputs) : undefined
+    const output = hasError
+      ? undefined
+      : (results !== undefined && results !== null ? this.formatIO(results) : undefined)
+
+    this._tagger.tagTextIO(span, input, output)
+  }
+
+  formatIO (data) {
```
sabrenner: nit for a follow-up: we can probably move this to a util file/share it with the general langchain integration

packages/dd-trace/src/llmobs/plugins/langgraph/index.js
```diff
@@ -0,0 +1,151 @@
...
+      return String(data)
+    }
+  }
+}
+
+class PregelStreamLLMObsPlugin extends BaseLangGraphLLMObsPlugin {
+  static id = 'llmobs_langgraph_pregel_stream'
+  static prefix = 'tracing:orchestrion:@langchain/langgraph:Pregel_stream'
+
+  asyncEnd (ctx) {
```
sabrenner: i gotta try this out locally, and i know tests are passing but it seems like this could be overwriting the `asyncEnd` from the `LlmObsPlugin` of the `BaseLangGraphLLMObsPlugin` above? although since tests pass this is a bit confusing - just leaving this as a note for myself.

sabrenner: i gotta try this out locally, and i know tests are passing but it seems like this could be overwriting the `asyncEnd` from the `LlmObsPlugin` of the `BaseLangGraphLLMObsPlugin` above? although since tests pass this is a bit confusing - just leaving this as a note for myself.

packages/dd-trace/src/llmobs/plugins/langgraph/index.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    // Tag when done
+    if (ctx.result?.done) {
+      this.#tagOnComplete(ctx)
+    }
+  }
+
+  error (ctx) {
+    this.#tagOnComplete(ctx)
```
sabrenner: is this needed? will we not hit the `asyncEnd` of the parent `BaseLangGraphLLMObsPlugin` plugin and tag what we have on the end there?

packages/dd-trace/test/llmobs/plugins/langgraph/index.spec.js
```diff
@@ -0,0 +1,212 @@
...
+          const { apmSpans, llmobsSpans } = await getEvents()
+
+          assertLlmObsSpanEvent(llmobsSpans[0], {
+            span: apmSpans[0],
+            spanKind: 'workflow',
+            name: 'foobarzoo',
+            inputValue: JSON.stringify({
+              messages: [{ role: 'user', content: 'Stream test' }],
+            }),
+            outputValue: MOCK_STRING, // Final streamed output
```
sabrenner: since this should be deterministic, let's assert the actual aggregated output

packages/dd-trace/test/llmobs/plugins/langgraph/index.spec.js
```diff
@@ -0,0 +1,212 @@
...
+          assert.ok(chunks.length > 0)
+
+          const { apmSpans, llmobsSpans } = await getEvents()
+
+          assertLlmObsSpanEvent(llmobsSpans[0], {
+            span: apmSpans[0],
+            spanKind: 'workflow',
+            name: 'foobarzoo',
+            inputValue: JSON.stringify({ count: 0 }),
+            outputValue: MOCK_STRING,
```
sabrenner, line 109: ditto on asserting real output

packages/dd-trace/test/plugins/externals.json
sabrenner, line 1: this file can be removed

packages/dd-trace/test/plugins/versions/package.json
sabrenner, line 1: looks like there are some unrelated version updates in this file (prisma stuff, langchain core) - if we could only add langgraph, and any zod packages, i think that's best!

docs/add-redirects.sh
tlhunter: Yeah, I deleted this file recently, it must've been added accidentally during a rebase. The file needs to be deleted in this PR.

crysmags: This file, keeps on reappearing, yes will delete

.github/workflows/apm-integrations.yml
```diff
@@ -671,6 +671,17 @@ jobs:
...
 
+  langgraph:
+    runs-on: ubuntu-latest
+    env:
+      PLUGINS: langgraph
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: ./.github/actions/plugins/test
+        with:
+          dd_api_key: ${{ secrets.DD_API_KEY }}
```
wconti27: We should move this to the llmobs integrations job

index.d.ts
```diff
@@ -2541,6 +2542,12 @@ declare namespace tracer {
      * This plugin automatically instruments the
      * [ldapjs](https://github.com/ldapjs/node-ldapjs/) module.
      */
+    /**
+     * This plugin automatically instruments the
+     * [langgraph](https://github.com/npmjs/package/langgraph) library.
+     */
+    interface langgraph extends Instrumentation {}
```
wconti27, line 2585: need to fix the comments here to align with the interfaces.

packages/dd-trace/test/llmobs/plugins/langgraph/index.spec.js
```diff
@@ -0,0 +1,283 @@
...
+          assert.ok(finalOutput.length > 0)
+
+          const { apmSpans, llmobsSpans } = await getEvents()
+
+          assertLlmObsSpanEvent(llmobsSpans[0], {
+            span: apmSpans[0],
+            spanKind: 'workflow',
+            name: 'foobarzoo',
+            inputValue: JSON.stringify({ text: '' }),
+            outputValue: MOCK_STRING,
```
wconti27, line 150: lets assert real output everywhere

packages/datadog-plugin-langgraph/test/index.spec.js
```diff
@@ -0,0 +1,69 @@
...
+  })
+
+  beforeEach(async () => {
+    await agent.load('langgraph')
+  })
+
+  afterEach(async () => {
+    await agent.close({ ritmReset: false })
+  })
+
```
wconti27, line 29: we should also be testing against normal pregel.invoke as well as stream IMO

crysmags: but we arent instrumenting invoke....

wconti27: fair enough, was just thinking so because its instrumented under the hood with stream, and it was an original instrumentation target

sabrenner: yeah i kinda agree it might be nice to have tests for `invoke` just in case they ever change the underlying implementation to not call stream, we wanna make sure we catch it and can update our orchestrion hooks/instrumentation

crysmags: can we do this in the focused update you mentioned?

sabrenner: yep that's good w me!

packages/dd-trace/src/llmobs/plugins/langgraph/index.js
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+
+  #tagAndCleanup (span, hasError) {
+    const streamData = streamDataMap.get(span)
+    if (!streamData) return
+
+    const { streamInputs: inputs, chunks } = streamData
+    const input = inputs == null ? undefined : formatIO(inputs)
+    const lastChunk = chunks.length > 0 ? chunks[chunks.length - 1] : undefined
```
sabrenner, line 102: just leaving as a comment to confirm, but we said in a follow-up we'd try to aggregate this, right? instead of just using the last chunk?

packages/dd-trace/test/plugins/versions/package.json
crysmags: done

packages/dd-trace/src/llmobs/plugins/langgraph/index.js
```diff
@@ -0,0 +1,114 @@
...
+
+  getLLMObsSpanRegisterOptions (ctx) {
+    const name = ctx.self.name || 'LangGraph'
+
+    const enabled = this._tracerConfig.llmobs.enabled
+    if (!enabled) return
+
+    const span = ctx.currentStore?.span
+    if (!span) return
+    streamDataMap.set(span, {
```
sabrenner, line 47: nit: for the follow-up pr or here if easy enough to add, we can just stick the chunks on the context i think instead of using a weakmap

something like 

```js
ctx.chunks = []
```

and then below

```js
ctx.chunks?.push(ctx.result.value)
```

and since `streamInputs` is already on the context, I don't think it's needed as well and we can just grab it below.

or, is the `ctx` here not the same `ctx` object that gets passed into the `NextStreamLLMObsPlugin` plugin?

packages/dd-trace/src/llmobs/plugins/langgraph/index.js
```diff
@@ -0,0 +1,114 @@
...
+  start () {} // no-op: span was already registered by PregelStreamLLMObsPlugin
+
+  end () {} // no-op: context restore is handled by PregelStreamLLMObsPlugin
+
+  error (ctx) {
+    const span = ctx.currentStore?.span
+    if (!span) return
+
+    this.#tagAndCleanup(span, true)
+  }
```
sabrenner, line 74: saw your follow up on slack but i think this function is still included in the base base `Plugin` class, so we might be ok to remove it? but we can also try that in a follow up, not needed here for now.

PR #7564, opened 2026-02-18
chore: add add-new-instrumentation agent skill

watson:
### What does this PR do?

Adds an `add-new-instrumentation` agent skill that provides a step-by-step guide for creating new dd-trace instrumentation files and plugins. The skill is stored under `.agents/skills/add-new-instrumentation/SKILL.md` and symlinked into both `.claude/skills/` and `.cursor/skills/`, making it available to Claude and Cursor AI agents alike.

### Motivation

Adding new instrumentation requires coordinating changes across several files (`datadog-instrumentations/`, `datadog-plugin-*/`, `plugins/index.js`, `index.d.ts`, `docs/API.md`, CI config, etc.). Encoding this workflow as a shared skill gives AI agents consistent, up-to-date guidance so they can scaffold new integrations correctly without missing steps.

### Additional Notes

`.gitignore` is updated to stop ignoring `.claude/skills/` and `.cursor/skills/` so the symlinks are tracked in version control, while other Claude/Cursor local settings (e.g. `settings.local.json`) remain ignored.

PR #7568, opened 2026-02-18
add integration skill

wconti27:
Adds a Claude skill for integrations. Uses skill best practices including limiting Skill.MD to under 300 lines, and using reference links in the MD to reference files that hold more specific information. Pulls in information from the recently added PR for adding a new integration, and adds this information as a reference for new integration implementation guide. Prioritizes orchestrion for writing new integrations.

https://github.com/DataDog/dd-trace-js/pull/7564

### DIFF
--- a/.agents/skills/apm-integrations/SKILL.md
+++ b/.agents/skills/apm-integrations/SKILL.md
@@ -2,8 +2,8 @@
 name: apm-integrations
 description: |
   This skill should be used when the user asks to "add a new integration",
-  "create a new plugin", "instrument a library", "add instrumentation for",
-  "write a plugin", "create instrumentation", "new dd-trace plugin",
+  "instrument a library", "add instrumentation for",
+  "create instrumentation", "new dd-trace integration",
   "add tracing for", "TracingPlugin", "DatabasePlugin", "CachePlugin",
   "ClientPlugin", "ServerPlugin", "CompositePlugin", "ConsumerPlugin",
   "ProducerPlugin", "addHook", "shimmer.wrap", "orchestrion",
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -139,7 +139,6 @@ __queuestorage__/AzuriteConfig
 
 # ignore apm-instrumentation-toolkit analysis results dirs
 .analysis/
-
```
watson, line 142: Nit: Re-add empty line

wconti27: fixed!

.agents/skills/apm-integrations/SKILL.md
```diff
@@ -0,0 +1,133 @@
...
+  This skill should be used when the user asks to "add a new integration",
+  "create a new plugin", "instrument a library", "add instrumentation for",
+  "write a plugin", "create instrumentation", "new dd-trace plugin",
+  "add tracing for", "TracingPlugin", "DatabasePlugin", "CachePlugin",
+  "ClientPlugin", "ServerPlugin", "CompositePlugin", "ConsumerPlugin",
+  "ProducerPlugin", "addHook", "shimmer.wrap", "orchestrion",
+  "bindStart", "bindFinish", "startSpan", "diagnostic channel",
+  "runStores", "reference plugin", "example plugin", "similar integration",
+  or needs to build, modify, or debug the instrumentation and plugin layers
+  for a third-party library in dd-trace-js.
```
watson, line 13: Nit: I think there's a difference between adding new instrumentation, and adding a new plugin. You can add a new plugin without having it be related to instrumentation. In that case, do we really want to the agent to try this skill?

.agents/skills/apm-integrations/SKILL.md
```diff
@@ -0,0 +1,133 @@
...
+- **Argument modification** — instrumentations that need to mutate arguments before the original call
+
+If none of these apply, use orchestrion. For shimmer patterns, refer to existing shimmer-based instrumentations in the codebase (e.g., `packages/datadog-instrumentations/src/pg.js`). Always try to use Orchestrion when beginning a new integration!
+
+## Plugin Base Classes
+
+Plugins extend a base class matching the library type. The base class provides automatic channel subscriptions, span lifecycle, and type-specific tags.
+
+```
+Plugin
```
watson, line 61: Is there a case for using just `Plugin`? If so, you should probably add a description to this line like you do for the other classes - otherwise just ignore me

.agents/skills/apm-integrations/SKILL.md
watson, line 1: There's nothing in this file about testing. Only a reference to the `testing.md` file at the very bottom that feels kinda optional. Consider including a flow for how the work should be implemented, including a way for the agent to validate its work, eg:

1. Investigate X
2. Implement Y
3. Add tests
4. This is how you verify that things work

.agents/skills/apm-integrations/SKILL.md
wconti27: Yeah there's a few cases where we directly extend Plugin (see aws-sdk)

.agents/skills/apm-integrations/SKILL.md
wconti27: Added!

.agents/skills/apm-integrations/SKILL.md
wconti27: Taken into account, though for placing the new plugin creation into another skill, I do not think that's the right move. Considering that plugins rely on the instrumentation side, and the LLM should have some context on instrumentations when writing any plugin.

PR #7569, opened 2026-02-18
fix(kafkajs): include kafka_cluster_id in DSM backlog offset tracking

robcarlan-datadog:
## Summary

- DSM checkpoints correctly included `kafka_cluster_id` in edge tags, but the backlog/offset tracking (which feeds `data_streams.kafka.lag_messages` and `data_streams.kafka.lag_seconds`) did not include it
- When the same topic exists on multiple Kafka clusters, producer offsets from different clusters were mixed into a single metric series, causing incorrect lag calculations
- This fix threads `clusterId` through to `setOffset()` calls for both producer and consumer commit paths so that backlog entries are properly scoped per cluster

### Changes

- **`packages/datadog-instrumentations/src/kafkajs.js`**: Capture resolved `clusterId` in closure and include it in consumer COMMIT_OFFSETS event data
- **`packages/datadog-plugin-kafkajs/src/producer.js`**: Extract `clusterId` from context and pass to `transformProduceResponse`, include `kafka_cluster_id` in backlog
- **`packages/datadog-plugin-kafkajs/src/consumer.js`**: Extract `clusterId` from commit items and include `kafka_cluster_id` in backlog
- **`packages/datadog-plugin-kafkajs/test/dsm.spec.js`**: Assert `kafka_cluster_id` is present in backlog entries when cluster ID is available

### Customer impact

Customers with the same topic on multiple Kafka clusters saw `data_streams.kafka.lag_messages` oscillate between cluster offsets (e.g., ~60k and ~84k), producing ~23k phantom lag messages. All cluster-scoped metrics (CloudWatch, DD Agent) showed 0-1 messages lag during the same window.

## Testing

Ran DSM without the fix and verified no kafka_cluster_id for the lag_messages and lag_seconds metric:
Screenshot 2026-02-19 at 11 09 17 am

Ran DSM with the fix and verified the kafka_cluster_id appeared:
Screenshot 2026-02-19 at 11 09 59 am


johannbotha (approved):
LGTM

packages/datadog-instrumentations/src/kafkajs.js
```diff
@@ -146,13 +131,29 @@ addHook({ name: 'kafkajs', file: 'src/index.js', versions: ['>=1.4'] }, (BaseKaf
...
+      }
+      consumerCommitCh.publish(commitList)
+    })
 
     const run = consumer.run
     const groupId = arguments[0].groupId
 
     consumer.run = function ({ eachMessage, eachBatch, ...runArgs }) {
       const wrapConsume = (clusterId) => {
+        resolvedClusterId = clusterId
```
rochdev, line 158: This relies on an assumption of an implementation detail to work, which is that `COMMIT_OFFSETS` will _always_ happen in the context of a run, synchronously, and that no 2 runs can run concurrently. Have you validated that this assumption is correct? If yes, then I would add a comment clarifying that as it's critical for this to work properly and for future readers.

robcarlan-datadog: That's a good point. I checked that's the case and added a comment
There should be only one run ([here](https://github.com/tulios/kafkajs/blob/55b0b416308b9e597a5a6b97b0a6fd6b846255dc/src/consumer/index.js#L198) and the `COMMIT_OFFSETS` only happens in the [offsetManager](https://github.com/tulios/kafkajs/blob/55b0b416308b9e597a5a6b97b0a6fd6b846255dc/src/consumer/offsetManager/index.js#L265) which is only used within the context of run.

PR #7570, opened 2026-02-18
fix(confluent): add kafka_cluster_id to DSM backlog and checkpoints

robcarlan-datadog:
## Summary

Builds on #7569 (kafkajs backlog cluster_id fix).

The `@confluentinc/kafka-javascript` instrumentation never called `getKafkaClusterId`, so `kafka_cluster_id` was missing from **both** DSM checkpoints (edge tags) **and** backlog offset tracking — a worse version of the kafkajs bug.

- **Extract** `getKafkaClusterId` and `isPromise` to shared `helpers/kafka.js` (used by both kafkajs and confluent instrumentations)
- **Update kafkajs** instrumentation to import from the shared helper (no behavior change)
- **Add cluster ID retrieval** to confluent KafkaJS producer and consumer paths
- **Thread `clusterId`** through producer `ctx`, consumer `extractedArgs`, and offset tracking (`updateLatestOffset` key now includes `clusterId` to prevent cross-cluster mixing)

### Not addressed

The **native module path** (`Producer`/`KafkaProducer`/`Consumer`/`KafkaConsumer` classes) does not yet support cluster ID retrieval since it lacks a KafkaJS-style admin API. This could be added in a follow-up using librdkafka's `getMetadata()`.

## Test plan

- [ ] Existing confluent DSM tests pass
- [ ] Verify with multi-cluster setup that lag metrics and checkpoints include `kafka_cluster_id`

**Depends on:** #7569

### DIFF
--- a/packages/datadog-instrumentations/src/confluentinc-kafka-javascript.js
+++ b/packages/datadog-instrumentations/src/confluentinc-kafka-javascript.js
@@ -7,6 +7,7 @@ const {
   addHook,
   channel,
 } = require('./helpers/instrument')
+const { getKafkaClusterId, isPromise } = require('./helpers/kafka')
 
 // Create channels for Confluent Kafka JavaScript
 const channels = {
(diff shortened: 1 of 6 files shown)


BridgeAR:
Superseded

PR #7571, opened 2026-02-18
feat(ai, llmobs): properly support ToolLoopAgent via existing patching

### LINKED ISSUE #7146: [BUG]: ToolLoopAgent in AI SDK 6 is not captured in LLM Observability
### Tracer Version(s)

5.81.0

### Node.js Version(s)

25.2.15.8

### Bug Report

This is the sample code for upcoming V6 SDK (which is to be GA next week): https://v6.ai-sdk.dev/docs/agents/overview

However it is not being captured in LLM Observability. I tried regular stream text, object are captured. This is only a wrapper though and I think it will be a very popular implementation moving forward.

If there is any interim steps I can do here to make them available, I'd be glad to do it as well.

      const weatherAgent = new ToolLoopAgent({
        model: google("gemini-2.5-flash-lite"),
        tools: {
          weather: tool({
            description: 'Get the weather in a location (in Fahrenheit)',
            inputSchema: z.object({
              location: z.string().describe('The location to get the weather for'),
            }),
            execute: async ({ location }) => ({
              location,
              temperature: 72 + Math.floor(Math.random() * 21) - 10,
            }),
          }),
          convertFahrenheitToCelsius: tool({
            description: 'Convert temperature from Fahrenheit to Celsius',
            inputSchema: z.object({
              temperature: z.number().describe('Temperature in Fahrenheit'),
            }),
            execute: async ({ temperature }) => {
              const celsius = Math.round((temperature - 32) * (5 / 9));
              return { celsius };
            },
          }),
        },
        // Agent's default behavior is to stop after a maximum of 20 steps
        // stopWhen: stepCountIs(20),
      });
      
      const result = await weatherAgent.generate({
        prompt: 'What is the weather in San Francisco in celsius?',
      });
      

### Reproduction Code

_No response_

### Error Logs

_No response_

### Tracer Config

_No response_

### Operating System

_No response_

### Bundling

Unsure

sabrenner:
### What does this PR do?
Properly supports `ToolLoopAgent` invocations by properly patching the traced methods.

Before, what we were doing was individually patching `generateText`, `generateStream`, etc. and patching the tracer, or defaulting to a no-op one, or not doing anything if explicitly disabled. The problem here is the `ToolLoopAgent` is defined in the same file, and since we're patching the exports, it was never hitting the patched functions.

Now, we're using our orchestrion-js implementation to rewrite the source to patch `getTracer` and `selectTelemetryAttributes`, such that anytime a function calls those functions to perform telemetry operations, we patch them at that point (with the same patching de-duping logic we already had). This means that we should properly support `ToolLoopAgent`, since it seems it's a relatively thin abstraction layer over just calling `generateText` and `streamText`.

### Motivation
Closes #7146 

MLOB-5644

### Additional Notes
- Most LOC are from cassette fixture diffs for testing, and adding the tests for `ToolLoopAgent`
- Need to add support for `rerank` in a follow-up, should be an easy win, just don't want to pollute this PR with out-of-scope stuff (should now not require an instrumentation/patching change)


packages/datadog-instrumentations/src/ai.js
```diff
@@ -63,29 +29,35 @@ function wrapTracer (tracer) {
 
       arguments[arguments.length - 1] = shimmer.wrapFunction(cb, function (originalCb) {
         return function (span) {
-          shimmer.wrap(span, 'end', function (spanEnd) {
+          // the below is necessary in the case that the span is vercel ai's noopSpan.
+          // while we don't want to patch the noopSpan more than once, we do want to treat each as a
+          // fresh instance. However, this is really not necessary for non-noop spans, but not sure
+          // how to differentiate.
+          const freshSpan = Object.create(span) // TODO: does this cause memory leaks?
```
sabrenner, line 36: see note here - not sure if we'd really leak here since we should be 1:1 with already-created spans that will be GC'd anyways, so this should be as well.

i also thought about trying to detect the no-op span via a WeakSet, potentially with just one entry, as the only time we should see the same span reference here is that case, and then do this cloning if so. Otherwise, if from a custom or real OpenTelemetry tracer, all spans should be unique.

packages/datadog-plugin-ai/test/index.spec.js
sabrenner, line 1: assertions removed from this file are due to not actually adding a tracer/modifying `isEnabled` on the options specific to each `ai` operation, but patching instead from `getTracer`

packages/datadog-instrumentations/src/helpers/rewriter/index.js
sabrenner, line 1: before, it seems we didn't support different ranges in the instrumentation definition file. i think i could have also just put the instrumentation definitions in reverse order (`>=6.0.0`, then `>=4.0.0 <6.0.0`, whereas now it is flipped, although that's an implementation detail that could be hard to discover, especially if supporting newer versions later and just appending them after existing entries.

happy to do this however, though

packages/datadog-instrumentations/src/helpers/rewriter/index.js
```diff
@@ -101,22 +102,24 @@ function disable (instrumentation) {
...
 }
 
-function satisfies (filename, filePath, versions) {
+function satisfies (filename, filePath, versions, functionQuery) {
   const [basename] = filename.split(filePath)
 
-  if (supported[basename] === undefined) {
+  supported[basename] ??= new Map()
+
+  if (supported[basename].get(functionQuery) === undefined) {
```
rochdev: The `functionQuery` variable will be new every time, so this will always be true. But more importantly, the check shouldn't rely on a function query to begin with, as it cannot be checked statically. Instead this should just store a list of versions. I'd probably use another plain object to keep is easy to write.

Probably out of scope, but it would be nice to also be able to list multiple versions in the same object directly in the rewriter config.

sabrenner: ah ok yeah i think that makes sense. so the `supported` object would end up looking something like

```javascript
{
  ai: ['>=4.0.0', '>=4.0.0 <6.0.0'],
  ... // other integrations
}
```

ie, in the case of `ai@5.0.0`, we would not add the function queries specified for `>=6.0.0`. am i understanding this correctly?

sabrenner: i _thinkkkk_ i've addressed this in https://github.com/DataDog/dd-trace-js/pull/7571/changes/2871d693d13234fc8e0e36e45128ea0d92447986 but please lmk if i misunderstood!

rochdev: Ah, yeah although I was thinking an object as an extension of how it was before, but actually a Set might be better. An array is definitely as simple as a Set but more costly, so let's use a Set instead.

sabrenner: agreed, fixed!

packages/datadog-instrumentations/src/ai.js
rochdev: Couldn't we just check if it was patched instead of just making it so that it's always patched through a wrapper?

sabrenner: yeah i had tried that, but the problem is that the first time it's patched some of the variables the span instance patching depends on (`name`, `attributes`) are already bound to the wrapper function, which we want to be _different_ on subsequent spans. the problem is, for no-op cases where a user does not provide their own tracer, the ai sdk uses the [same global span instance](https://github.com/vercel/ai/blob/97c7b9f80838f600d03832275a25039244882a51/packages/ai/src/telemetry/noop-tracer.ts#L29), instead of a [new one per span start of the no-op tracer](https://github.com/vercel/ai/blob/97c7b9f80838f600d03832275a25039244882a51/packages/ai/src/telemetry/noop-tracer.ts#L11). so i found that just creating a fresh object was ok for this.

wheeler-pushpress:
Thank you @sabrenner !

PR #7575, opened 2026-02-19
test: enable --allow-uncaught in mocha runs

watson:
### What does this PR do?

Updates Mocha-based test execution to use `--allow-uncaught` so uncaught exceptions are no longer absorbed by Mocha and instead fail fast by propagating to the process. This change applies to the repo's Mocha test scripts in `package.json` and to a direct Mocha invocation in `.github/workflows/platform.yml` that does not go through those scripts.

### Motivation

Ensure errors do not go unnoticed.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,3 +1,7 @@
+// TODO: once we've well iterated on this file (6 months+),
+// we can plan to move it to its own package to reuse elsewhere.
+// written on 2026-02-12
+
 import { readFileSync } from 'fs'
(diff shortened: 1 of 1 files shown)


.mocharc.js
```diff
@@ -1,6 +1,7 @@
 'use strict'
 
 module.exports = {
+  allowUncaught: true,
   color: true,
   exit: true,
```
BridgeAR, line 6: Independent: we should likely remove this option. It might actually hide issues https://mochajs.org/running/cli/#--exit

PR #7581, opened 2026-02-19
fix(rhea): DSM context leak between concurrent consumers

robcarlan-datadog:
## Summary

- Adds a regression test for the DSM context leak bug in rhea
- Fixes the bug by moving the consumer DSM checkpoint from `bindStart` to `start`, matching the pattern from #7395

## Root Cause

Same underlying issue as #7395 — `decodeDataStreamsContext` in `bindStart` calls `legacyStorage.enterWith(...)` on the **outer** async context (outside the store.run), contaminating all subsequent code running in that outer context. If two consumers fire sequentially (e.g. two messages arrive before either handler completes), consumer B's `decodeDataStreamsContext` overwrites the outer context's `dataStreamsContext`, so any producer triggered by consumer A incorrectly chains from consumer B's pathway instead of consumer A's.

## Fix

Move `decodeDataStreamsContext` + `setCheckpoint` to the `start` hook (which runs **inside** the consumer's store.run scope). `enterWith` there only affects the consumer's isolated async context — it cannot leak to siblings.

**No separate producer fix is needed.** The `encode` hook in the producer fires inside `startSendCh.runStores(...)` in the rhea instrumentation, so it's already within the producer's store.run scope. With the consumer fix in place, the producer's store inherits the correct `dataStreamsContext` from the consumer's async context, and the pathway chains correctly.

## Test plan

- [x] Regression test fails deterministically without the fix (10/10 runs, rhea 1.0.0 and 3.0.4)
- [x] Regression test covers the full consume→produce chain (asserts `produceA`'s parent context hash equals `consumeA`'s checkpoint hash)


robcarlan-datadog:
@codex review

ericfirth (approved):

PR #7584, opened 2026-02-19
feat(lambda): adopt mini agent for Lambda Lite environments

litianningdatadog:
https://datadoghq.atlassian.net/browse/SVLS-8571

### What does this PR do?                                                                                                                                                                         
                                                                                                                                                                                              
Introduces DATADOG_MINI_AGENT_PATH (/tmp/datadog/mini_agent_ready) as a sentinel constant to detect Lambda Lite + mini agent environments, and adjusts two behaviours when that path is present:                                                                                                                                                                                      
                                                                                                                                                                                              
1. Exporter selection (exporter.js): treats the mini agent the same as the Datadog Lambda Extension — selects AgentExporter instead of LogExporter.
2. flushInterval override (config/index.js): standard Lambda forces flushInterval = 0 so every span is sent before the invocation window closes. With the mini agent the process is a long-running web server, so batching is safe and flushInterval = 0 would saturate the 8-request exporter concurrency limit. The override is skipped, leaving the default (2000 ms).

This change depends on [this serverless-components PR](https://github.com/DataDog/serverless-components/pull/81)

### Motivation

Lambda Lite runs the tracer in a persistent server process alongside a mini agent, rather than the ephemeral invocation model of standard Lambda. The existing Lambda fast-path assumptions (flushInterval = 0, log exporter fallback) are counter-productive in this model and need to be gated on the absence of the mini agent ready file.

### Additional Notes

- DATADOG_MINI_AGENT_PATH mirrors the existing DATADOG_LAMBDA_EXTENSION_PATH pattern — both are file-existence checks, no new env vars.
- Unit tests added for both the exporter selection and the flushInterval branching covering all four combinations (standard Lambda, Lambda + extension, Lambda + mini agent, non-Lambda).

### DIFF
--- a/packages/dd-trace/src/config/index.js
+++ b/packages/dd-trace/src/config/index.js
@@ -21,7 +21,7 @@ const {
   getIsAzureFunction,
   enableGCPPubSubPushSubscription,
 } = require('../serverless')
-const { ORIGIN_KEY } = require('../constants')
+const { ORIGIN_KEY, DATADOG_MINI_AGENT_PATH } = require('../constants')
 const { appendRules } = require('../payload-tagging/config')
 const { getGitMetadataFromGitProperties, removeUserSensitiveInfo, getRemoteOriginURL, resolveGitHeadSHA } =
   require('./git_properties')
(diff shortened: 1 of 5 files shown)


packages/dd-trace/src/config/index.js
```diff
@@ -618,7 +618,11 @@ class Config {
     unprocessedTarget['experimental.aiguard.timeout'] = DD_AI_GUARD_TIMEOUT
     setBoolean(target, 'experimental.enableGetRumData', DD_TRACE_EXPERIMENTAL_GET_RUM_DATA_ENABLED)
     setString(target, 'experimental.exporter', DD_TRACE_EXPERIMENTAL_EXPORTER)
-    if (AWS_LAMBDA_FUNCTION_NAME) {
+    if (AWS_LAMBDA_FUNCTION_NAME && !fs.existsSync(DATADOG_MINI_AGENT_PATH)) {
+      // Standard Lambda (with or without the Datadog Extension): flush every span
+      // immediately so traces reach the extension before the invocation window closes.
+      // Skipped for Lambda Lite + mini agent, which is a long-running web server where
+      // batching is safe and flushInterval=0 would saturate the 8-request exporter limit.
```
rochdev: The 8 request limit has been removed so this could be confusing to future readers.

rochdev (approved):

PR #7591, opened 2026-02-20
Update eventbridge.js

Dogbu-cyber:


BridgeAR:
@Dogbu-cyber thank you for the fix!

I ported it and added test cases so that the limit applies in a way that we never exceed the limit by adding the meta information across multiple events.

It will be released in the next version.

PR #7595, opened 2026-02-22
add codeowners validation and ensure teams for all tests

rochdev:
### What does this PR do?

Add CODEOWNERS validation and ensure teams for all tests.

### Motivation

Right now a lot of files are not covered by CODEOWNERS and fallback to dd-trace-js. Instead, every file should have a team.

### Additional Notes

For now the check is limited to test files as the main goal is to ensure Test Optimization can properly identify the owner of a test, but ideally every file would have an owner even outside of tests eventually.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,3 +1,5 @@
+# TODO: Restructure the project by product and clean up this file.
+
 # AppSec
 /integration-tests/appsec/ @DataDog/asm-js
 /packages/dd-trace/src/appsec/ @DataDog/asm-js


CODEOWNERS
```diff
@@ -1,26 +1,62 @@
...
 /integration-tests/appsec/ @DataDog/asm-js
 /packages/dd-trace/src/appsec/ @DataDog/asm-js
 /packages/dd-trace/test/appsec/ @DataDog/asm-js
 
 /integration-tests/aiguard/ @DataDog/asm-js
 /packages/dd-trace/src/aiguard/ @DataDog/asm-js
 /packages/dd-trace/test/aiguard/ @DataDog/asm-js
+/packages/dd-trace/test/plugins/util/ip_extractor.spec.js @DataDog/asm-js
 
+/packages/dd-trace/*/standalone @DataDog/asm-js
```
CarlesDD, line 13: Nit: The `*/standalone` glob also covers `src/standalone`. Since this PR is scoped to test files, should this be `/packages/dd-trace/test/standalone` instead? Happy either way, just flagging for consistency.

juan-fernandez (approved):
looks good from test optimization's perspective

CODEOWNERS
rochdev: The current CI check is only for tests, but in theory we would want source files to be covered as well, so that's actually fine. Eventually we'd want every file to have an owner anyway.

CODEOWNERS
```diff
@@ -76,14 +111,22 @@
 /packages/datadog-plugin-openai/ @DataDog/ml-observability
 /packages/dd-trace/src/llmobs/ @DataDog/ml-observability
 /packages/dd-trace/test/llmobs/ @DataDog/ml-observability
+/packages/dd-trace/test/plugins/util/llm.spec.js @DataDog/ml-observability
```
sabrenner, line 114: this made me realize that the source file isn't covered in codeowners for this test file, if you wanna add it in, otherwise i'll do it as a driveby in whatever my next pr is 😄

rochdev: Yeah there are still a lot of files that are missing coverage. We should eventually expand the check to all files. In the meantime it's more of a best effort.

tlhunter (approved):

PR #7596, opened 2026-02-22
fix(plugins): fix registry and use() validation

watson:
### What does this PR do?

This PR fixes inconsistencies in the `tracer.use()` API and the runtime plugin registry for configurable plugins.

It does three things:

- Adds `fs` (including the `node:fs` alias) and `web` to the internal runtime plugin registry so `tracer.use()` can resolve them instead of silently ignoring them.
- Marks `fs` as `experimental` so it remains opt-in and does not get auto-enabled just because the instrumentation loads.
- Removes `knex` and `generic-pool` from `index.d.ts` and the generated docs because they are internal context-propagation patches, not configurable tracing plugins.

It also updates the `fs` plugin tests so the instrumentation-only setup no longer accidentally enables `fs` via the registry-backed `agent.load('fs')` path.

### Motivation

The current state is inconsistent in both directions:

- valid plugin targets like `fs` were documented/configurable in spirit but missing from the runtime registry, so `tracer.use('fs')` did not actually work as expected
- non-configurable internals like `knex` and `generic-pool` were exposed in the public type/docs surface even though they do not behave like normal configurable plugins

This PR makes the public/configurable surface and the runtime behavior line up more closely while keeping `fs` disabled by default.

### Additional Notes

This PR is stacked on top of `watson/fix-appsec-fs-store`, which isolates AppSec `fs` state from tracing context. That lower PR removes the AppSec/store coupling so this layer can focus on the `tracer.use()` and registry behavior.


BridgeAR:
Superseded

PR #7597, opened 2026-02-22
fix(fs): avoid leaking noop context from fs plugin

watson:
### What does this PR do?

Fixes a bug where an untraced `fs` operation could accidentally affect surrounding tracing context on the same execution path. When the `fs` plugin decides not to create an `fs` span because there is no active parent span, that noop state should stay local to the current `fs` call and preserve any unrelated context that is already active.

This PR makes that skip behavior scoped to the current `fs` operation by returning a scoped noop store instead of mutating ambient context. A regression test in `packages/datadog-plugin-fs/test/index.spec.js` verifies that noop state does not leak into the parent context.

Additionally, since the `fs` plugin was the last caller of the `Plugin#skip()` method, this PR removes it from the base `Plugin` class and fixes pre-existing TypeScript errors in `plugin.js` (incorrect span type narrowing and a wrong config parameter type).

### Motivation

From a user perspective, this bug can show up as missing spans or lost context after ordinary `fs` calls, even though those later operations are unrelated to `fs`. The `fs` plugin should avoid creating standalone `fs` spans when nothing is currently traced, but it should do that without mutating or discarding unrelated tracing state that happens afterward.

### Additional Notes

- `fs` plugin now returns `{ noop: true }` directly instead of calling `this.skip()`, which used `enterWith` to mutate ambient context.
- `Plugin#skip()` has been removed since no other plugin uses it.
- Fixed TypeScript errors in `plugin.js`: cast `store.span` to `DatadogSpan` for `_spanContext`/`setTag` access, and corrected the `configure()` param type from `ConfigBase` (tracer config) to the actual plugin config shape.

### DIFF
--- a/packages/dd-trace/src/plugins/plugin.js
+++ b/packages/dd-trace/src/plugins/plugin.js
@@ -162,12 +162,12 @@ module.exports = class Plugin {
    *
    * TODO: Remove the overloading with `enabled` and use the config object directly.
    *
-   * @param {boolean | {enabled: boolean} & Record<string, unknown>} config Either a boolean to
+   * @param {boolean | import('../config/config-base') & {enabled: boolean}} config Either a boolean to
    * enable/disable or a configuration object containing at least `{ enabled: boolean }`.
    */
   configure (config) {
     if (typeof config === 'boolean') {
-      config = { enabled: config }
+      config = /** @type {import('../config/config-base') & {enabled: boolean}} */ ({ enabled: config })
     }
     this.config = config
     if (config.enabled) {


BridgeAR (commented):
Should skip not just be changed in that case to generally function like that?

watson:
@BridgeAR Good question. I don't have a good overview of the consequences of making such a change. We probably would need some input from APM.

packages/dd-trace/src/plugins/plugin.js
```diff
@@ -167,8 +162,8 @@ module.exports = class Plugin {
    *
    * TODO: Remove the overloading with `enabled` and use the config object directly.
    *
-   * @param {boolean|import('../config/config-base')} config Either a boolean to enable/disable
-   * or a configuration object containing at least `{ enabled: boolean }`.
+   * @param {boolean | {enabled: boolean} & Record<string, unknown>} config Either a boolean to
+   * enable/disable or a configuration object containing at least `{ enabled: boolean }`.
```
BridgeAR: This looks like a partial regression (it should keep the config-base). I see the enabled part, but I believe it is never used as such, only as boolean.

watson: Fixed 👍

PR #7599, opened 2026-02-23
chore(test): Fix cmd injection telemetry flaky tests

uurien:
### What does this PR do?

Instead of checking only the first 2 messages coming in the telemetry, this will wait until the first telemetry message that asserts everithing.
And also, the check will fail if the telemetry is not in appsec namespace.
### Additional Notes

[APPSEC-60075]

[APPSEC-60075]: https://datadoghq.atlassian.net/browse/APPSEC-60075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/appsec.yml
+++ b/.github/workflows/appsec.yml
@@ -23,10 +23,7 @@ env:
 
 jobs:
   macos:
-    strategy:
-      matrix:
-        run: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
-    name: ${{ github.workflow }} / macos (run ${{ matrix.run }})
+    name: ${{ github.workflow }} / macos
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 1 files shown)


packages/dd-trace/test/appsec/rasp/command_injection.integration.spec.js
```diff
@@ -58,7 +58,7 @@ describe('RASP - command_injection - integration', () => {
       const checkMessages = agent.assertMessageReceived(({ headers, payload }) => {
         assert.ok(Object.hasOwn(payload[0][0].meta, '_dd.appsec.json'))
         assert.match(payload[0][0].meta['_dd.appsec.json'], new RegExp(`"rasp-command_injection-rule-id-${ruleId}"`))
-      })
+      }, 4_000)
```
uurien, line 61: Tests have a default timeout of 5000, and this method default timeout of 30_000. I'm changing this to have a readable message if test timeout happens. (same in line 85)

packages/dd-trace/test/appsec/rasp/command_injection.integration.spec.js
```diff
@@ -79,8 +79,10 @@ describe('RASP - command_injection - integration', () => {
           assert.ok(matchSerie.tags.includes('rule_type:command_injection'))
           assert.ok(matchSerie.tags.includes(`rule_variant:${variant}`))
           assert.strictEqual(matchSerie.type, 'count')
+        } else {
+          assert.fail('namespace should be appsec')
         }
-      }, 'generate-metrics', 30_000, 2)
+      }, 'generate-metrics', 4_000, 1, true)
```
uurien, line 85: the number `1` doesn't matter, because the `true` is a flag that ignores the first property. This test is going to success when a telemetry that success the asserts happen.

PR #7607, opened 2026-02-24
chore(test): Fixing iast overhead controller flaky tests

uurien:
### What does this PR do?

Tries to fix flaky test in overhead controller. 
Changes the test check to do subscription to the agent before the execution of the request, seems that there is a race condition when the traces are received before the subscription. Now should be fixed.

### Additional Notes

CI with +200 success executions: https://github.com/DataDog/dd-trace-js/actions/runs/22357293612?pr=7607 (two iteration of +100)

### DIFF
--- a/packages/dd-trace/test/appsec/iast/overhead-controller.integration.spec.js
+++ b/packages/dd-trace/test/appsec/iast/overhead-controller.integration.spec.js
@@ -66,9 +66,10 @@ describe('IAST - overhead-controller - integration', () => {
         })
       }, 1000, 1, true)
 
-      await axios.request(path, { method })
-
-      await assertPromise
+      await Promise.all([
+        axios.request(path, { method }),
+        assertPromise,
+      ])
     }
 
     async function checkNoVulnerabilitiesInEndpoint (path, method = 'GET') {
@@ -78,8 +79,10 @@ describe('IAST - overhead-controller - integration', () => {
         assert.ok(!('_dd.iast.json' in payload[0][0].meta))
       }, 1000, 1, true)
 
-      await axios.request(path, { method })
-      await assertPromise
+      await Promise.all([
+        axios.request(path, { method }),
+        assertPromise,
+      ])
     }
 
     it('should report vulnerability only in the first request', async () => {


packages/dd-trace/test/appsec/iast/overhead-controller.integration.spec.js
```diff
@@ -67,16 +65,21 @@ describe('IAST - overhead-controller - integration', () => {
           assert.strictEqual(vulnerabilities[vType], vulnerabilitiesAndCount[vType], `route: ${path} - type: ${vType}`)
         })
       }, 1000, 1, true)
-    }
 
-    async function checkNoVulnerabilitiesInEndpoint (path, method = 'GET') {
       await axios.request(path, { method })
 
-      await agent.assertMessageReceived(({ payload }) => {
+      await assertPromise
```
BridgeAR: If assertMessageReceived would reject before axios.request resolves (due to potential any side effects), this would cause an unhandled rejection.
I would therefore use `Promise.all()` here.

packages/dd-trace/test/appsec/iast/overhead-controller.integration.spec.js
```diff
@@ -67,16 +65,21 @@ describe('IAST - overhead-controller - integration', () => {
...
+
+    async function checkNoVulnerabilitiesInEndpoint (path, method = 'GET') {
+      const assertPromise = agent.assertMessageReceived(({ payload }) => {
         assert.strictEqual(payload[0][0].type, 'web')
         assert.strictEqual(payload[0][0].metrics['_dd.iast.enabled'], 1)
         assert.ok(!('_dd.iast.json' in payload[0][0].meta))
       }, 1000, 1, true)
+
+      await axios.request(path, { method })
+      await assertPromise
```
BridgeAR:
```suggestion
      await Promise.all([
        axios.request(path, { method }),
        assertPromise,
      ])
```

packages/dd-trace/test/appsec/iast/overhead-controller.integration.spec.js
uurien: makes sense, changed.

CarlesDD (approved):

PR #7608, opened 2026-02-24
feat(dsm): add trackTransaction manual API

ericfirth:
### What does this PR do?

Adds a `trackTransaction(transactionId, checkpointName)` method to the DSM manual API (`dataStreamsCheckpointer`). Users can call it to record that a transaction ID was seen at a named processing stage, without any pathway propagation or carrier injection/extraction.

### Motivation

Ports the transaction tracking feature from the Go tracer to Node.js using the same binary wire format, so the backend can decode payloads from either tracer consistently.

Wire format per entry: `[checkpointId uint8][timestamp int64BE 8 bytes][idLen uint8][id bytes]`

Checkpoint names are mapped to stable 1-byte IDs (1–254) via a per-processor `CheckpointRegistry`. The full name→ID mapping is serialized alongside transaction bytes in each stats bucket as `TransactionCheckpointIds`.

### Additional Notes

- `trackTransaction` is gated behind `DD_DATA_STREAMS_ENABLED`. Without it the periodic flush timer is not set up, so data would only be sent at process exit — effectively useless for long-running services. A `log.warn` is emitted when called with DSM disabled so users know why nothing is happening.
- `Transactions` and `TransactionCheckpointIds` fields are omitted from buckets that have no transactions, keeping the msgpack payload backward-compatible.
- 18 new unit tests added across `processor.spec.js` and `data_streams_checkpointer.spec.js`.

### Semver

`semver-minor` (new public API)

### DIFF
--- a/packages/dd-trace/src/datastreams/processor.js
+++ b/packages/dd-trace/src/datastreams/processor.js
@@ -75,15 +75,13 @@ class Backlog {
  * Scope is per-processor so IDs are stable across bucket boundaries within a process lifetime.
  */
 class CheckpointRegistry {
-  constructor () {
-    /** @type {Map<string, number>} */
-    this._nameToId = new Map()
-    this._nextId = 1
-    /** @type {Buffer[]} Pre-built [id uint8][nameLen uint8][name bytes] entries, one per registered name. */
-    this._entryBuffers = []
-    /** @type {Buffer | null} Cached concat of _entryBuffers; reset when a new name is added. */
-    this._encodedKeysCache = null
-  }
+  /** @type {Map<string, number>} */
+  #nameToId = new Map()
+  #nextId = 1
+  /** @type {Buffer[]} Pre-built [id uint8][nameLen uint8][name bytes] entries, one per registered name. */
+  #entryBuffers = []
+  /** @type {Buffer | null} Cached concat of #entryBuffers; reset when a new name is added. */
+  #encodedKeysCache = null
 
   /**
    * Returns the byte ID for the given checkpoint name, assigning one if not seen before.
@@ -92,11 +90,11 @@ class CheckpointRegistry {
    * @returns {number | undefined}
    */
   getId (name) {
-    const existing = this._nameToId.get(name)
+    const existing = this.#nameToId.get(name)
     if (existing !== undefined) return existing
-    if (this._nextId > 254) return
-    const id = this._nextId++
-    this._nameToId.set(name, id)
+    if (this.#nextId > 254) return
+    const id = this.#nextId++
+    this.#nameToId.set(name, id)
     // Build the wire entry now with a bounded write so long names never materialise
     // their full UTF-8 encoding — buf.write() stops at the supplied byte limit.
     const nameBuf = Buffer.allocUnsafe(255)
@@ -105,8 +103,8 @@ class CheckpointRegistry {
     entry.writeUInt8(id, 0)
     entry.writeUInt8(nameByteLen, 1)
     nameBuf.copy(entry, 2, 0, nameByteLen)
-    this._entryBuffers.push(entry)
-    this._encodedKeysCache = null
+    this.#entryBuffers.push(entry)
+    this.#encodedKeysCache = null
     return id
   }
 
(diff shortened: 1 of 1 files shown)


packages/dd-trace/src/datastreams/processor.js
```diff
@@ -93,6 +141,16 @@ class StatsBucket {
...
   }
 
+  /**
+   * Appends pre-encoded transaction bytes to this bucket.
+   * @param {Buffer} bytes
+   */
+  addTransaction (bytes) {
+    this._transactions = this._transactions === null
+      ? bytes
+      : Buffer.concat([this._transactions, bytes])
```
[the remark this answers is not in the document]
tlhunter: @codex how would you recommend solving this?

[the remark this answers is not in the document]
ericfirth: fixed

packages/dd-trace/src/datastreams/processor.js
```diff
@@ -294,9 +371,47 @@ class DataStreamsProcessor {
...
+    const timestampNs = BigInt(Date.now()) * 1_000_000n
+
+    const entry = Buffer.alloc(1 + 8 + 1 + idBytes.length)
+    entry.writeUInt8(checkpointId, 0)
+    entry.writeBigInt64BE(timestampNs, 1)
+    entry.writeUInt8(idBytes.length, 9)
+    idBytes.copy(entry, 10)
+
+    // Number() cast is safe here: 10s bucket granularity tolerates ~0.5ns precision loss
+    this.bucketFromTimestamp(Number(timestampNs)).addTransaction(entry)
```
tlhunter, line 401: What are the implications of being in the wrong bucket? Would it mean something like a metric close to a time boundary being reported in a previous flush instead of a subsequent flush?

packages/dd-trace/test/datastreams/processor.spec.js
```diff
@@ -358,6 +359,230 @@ describe('DataStreamsProcessor', () => {
...
+    assert.strictEqual(registry.getId('beta'), 2)
+    assert.strictEqual(registry.getId('gamma'), 3)
+  })
+
+  it('returns the same ID for repeated names', () => {
+    const first = registry.getId('alpha')
+    const second = registry.getId('alpha')
+    assert.strictEqual(first, second)
+    assert.strictEqual(first, 1)
+  })
```
tlhunter, line 379: We try to make tests so that they're self contained. Like one should be able to isolate and only run one `it` block and have it pass (e.g. using `it.only`). But it looks like the second `it` depends on side effects from the first `it`. Can you throw the shared work into the `beforeEach` block? And to that end clear the records out in a new `afterEach`?

packages/dd-trace/src/datastreams/processor.js
ericfirth: Yes — the worst case is a transaction falling into an adjacent 10-second flush window rather than the current one, which doesn't matter really.

In practice this is very unlikely: Date.now() only has millisecond precision (the nanosecond value always has 6 trailing zeros), and the float64 ULP at this timestamp magnitude is ~200–400 ns — far smaller than 1 ms

packages/dd-trace/test/datastreams/processor.spec.js
ericfirth: The beforeEach at the top of the describe block already creates a fresh CheckpointRegistry() for each test, so they're fully isolated — running any single one with .only will pass on its own. I've dropped the `assert.strictEqual(first, 1)` from the second test since asserting the specific value belongs in the first test.

packages/dd-trace/src/datastreams/processor.js
```diff
@@ -69,10 +69,68 @@ class Backlog {
...
 
+/**
+ * Maps checkpoint name strings to single-byte IDs (1–254).
+ * ID 0 is reserved; 254 is the maximum number of unique names.
+ * Scope is per-processor so IDs are stable across bucket boundaries within a process lifetime.
+ */
+class CheckpointRegistry {
+  constructor () {
+    /** @type {Map<string, number>} */
+    this._nameToId = new Map()
```
BridgeAR: Please use actual private properties instead of underscores :)


```suggestion
    this.#nameToId = new Map()
```

BridgeAR (commented):
I can not say much about the business requirements here.

I guess there is no way to do this without programmatic API?

packages/dd-trace/src/datastreams/processor.js
ericfirth: done. there were more than just that one, so i had claude look rather than accept the suggested commit

PR #7617, opened 2026-02-25
chore(test): Fixing flaky ssrf

uurien:
### What does this PR do?

Tries to fix the flaky test happening in the rasp test file `ssrf.express.plugin.spec.js`.
There is no easy way to reproduce the failure, but I figured out that the reason why the web root span is not found is because the trace is not complete and it is not sent to the agent. 
This could happen when timeout happens before the outgoing http request finishes. That is why I made some changes in the tested controller to be sure that the outgoing request span is closed as soon as possible, calling to `res.resume()` and finishing the main request when the outgoing request has finished.

### Motivation

Reduce flaky tests

### Additional Notes

50 success executions in the CI: https://github.com/DataDog/dd-trace-js/actions/runs/22406751414/job/64868757093?pr=7617

### DIFF
--- a/packages/dd-trace/test/appsec/rasp/ssrf.express.plugin.spec.js
+++ b/packages/dd-trace/test/appsec/rasp/ssrf.express.plugin.spec.js
@@ -88,7 +88,7 @@ describe('RASP - ssrf', () => {
             }
 
             await Promise.all([
-              checkRaspExecutedAndNotThreat(agent, true, 9_000),
+              checkRaspExecutedAndNotThreat(agent),
               axios.get('/?host=www.datadoghq.com'),
             ])
           })
@@ -148,7 +148,7 @@ describe('RASP - ssrf', () => {
 
             await Promise.all([
               axios.get('/?host=www.datadoghq.com'),
-              checkRaspExecutedAndNotThreat(agent, true, 9_000),
+              checkRaspExecutedAndNotThreat(agent),
             ])
           })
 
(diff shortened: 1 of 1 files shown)


packages/dd-trace/test/appsec/rasp/ssrf.express.plugin.spec.js
```diff
@@ -194,8 +201,10 @@ describe('RASP - ssrf', () => {
             }
 
             axios.get('/?host=www.datadoghq.com')
-
-            return checkRaspExecutedAndNotThreat(agent)
+            await Promise.all([
+              axios.get('/?host=www.datadoghq.com'),
```
tlhunter: Is this intentionally making the same request twice?

uurien: Not, it is not. Good catch 😅 
Changed.

packages/dd-trace/test/appsec/rasp/ssrf.express.plugin.spec.js
```diff
@@ -193,9 +200,10 @@ describe('RASP - ssrf', () => {
               })
             }
 
-            axios.get('/?host=www.datadoghq.com')
-
-            return checkRaspExecutedAndNotThreat(agent)
+            await Promise.all([
+              axios.get('/?host=www.datadoghq.com'),
+              checkRaspExecutedAndNotThreat(agent, true, 9_000),
```
BridgeAR: AFAIK this method only has two arguments (the second one implicitly being true, so there is no change to only using one as originally)?

The same about the other changes.

uurien: That's true, there are remains of other tests I did before fixing the flaky.

PR #7632, opened 2026-02-27
agentless via json intake

bengl:
Enables agentless via the RUM span intake. Also adds a manual stress-test script.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -199,10 +199,13 @@
 /packages/dd-trace/test/dogstatsd.spec.js @DataDog/lang-platform-js
 /packages/dd-trace/test/encode/0.4.spec.js @DataDog/lang-platform-js
 /packages/dd-trace/test/encode/0.5.spec.js @DataDog/lang-platform-js
+/packages/dd-trace/test/encode/agentless-json.spec.js @DataDog/lang-platform-js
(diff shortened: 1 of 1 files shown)


packages/dd-trace/src/exporters/agentless/index.js
```diff
@@ -0,0 +1,89 @@
...
+      this._url = null
+    }
+
+    this._writer = new Writer({
+      url: this._url,
+      site,
+    })
+
+    const ddTrace = globalThis[Symbol.for('dd-trace')]
+    if (ddTrace?.beforeExitHandlers) {
```
pabloerhard, line 45: Doubt: Was this occasionally throwing an error? If so, then it should probably be fixed by #6910.

packages/dd-trace/src/exporters/agentless/writer.js
```diff
@@ -0,0 +1,184 @@
...
+      log.error('DD_API_KEY is required for agentless span intake. Set DD_API_KEY. Spans will not be sent.')
+    }
+  }
+
+  /**
+   * Flushes the current trace. Since we flush after each trace, this sends
+   * a single request.
+   * @param {Function} [done] - Callback when send completes
+   */
+  flush (done = () => {}) {
```
pabloerhard, line 51: This flush seems mostly on par to the one in BaseWriter, any reason on why we overwrite it?

PR #7645, opened 2026-03-02
feat(bundler): add webpack support via unplugin

tlhunter:
### What does this PR do?

Adds webpack bundler support to dd-trace using the [unplugin](https://unplugin.unjs.io/) package, while keeping the existing esbuild plugin interface unchanged.

**New entry point:** `require('dd-trace/webpack')` returns a webpack plugin class (`new DatadogPlugin()`) that instruments bundled applications.

**Unchanged:** `require('dd-trace/esbuild')` continues to return `{ name, setup }` exactly as before.

### Motivation

Users building Node.js applications with webpack had no way to use dd-trace instrumentation in their bundles. The esbuild plugin has existed for some time; this PR brings equivalent support to webpack 5.

### Architecture

A new `packages/datadog-unplugin/` package acts as a unified adapter:

- **esbuild**: delegates entirely to the existing `packages/datadog-esbuild/` plugin via the `esbuild.setup` escape hatch (unplugin's own resolveId/load hooks are disabled via `onResolveFilter: /^$/` to avoid conflicts)
- **webpack (npm packages)**: `resolveId` + `load` hooks generate virtual CJS wrapper modules that require the original package, then publish to `dd-trace:bundler:load` so the tracer instruments it
- **webpack (Node.js built-ins)**: a `webpack(compiler)` hook intercepts built-in requests in `NMF.hooks.beforeResolve` (before webpack's `NodeTargetPlugin` externalizes them) and redirects to virtual wrapper files via `webpack-virtual-modules`

### Timing fix

Built-in module wrappers can load before `plugin_manager.js` is initialized, causing `dd-trace:instrumentation:load` events to fire with no subscriber, leaving `pluginClasses` empty. Two changes address this:

1. Every virtual wrapper starts with `require(bundlerRegisterPath)` to guarantee `bundler-register.js` is subscribed before the channel publish fires.
2. `bundler-register.js` exports a `processedPackages` Set. `plugin_manager.configure()` iterates it and calls `maybeEnable()` for each package, ensuring plugin classes are registered even when their `loadChannel` events arrived before `plugin_manager.js` was loaded.

### Additional Notes

- `unplugin@^1.16.1` added as a regular `dependency` (Node ≥ 14 compatible; v2.x requires ≥ 18.12)
- `webpack-virtual-modules` is a `peerDependency` / optional — only needed at webpack build time
- Integration test in `integration-tests/webpack/` mirrors the esbuild test: installs webpack, builds a bundle with express + http, runs it, asserts the express span is active

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,6 +138,14 @@
     "import-in-the-middle": "^2.0.6",
     "unplugin": "^1.16.1"
   },
+  "peerDependencies": {
+    "webpack-virtual-modules": ">=0.5.0"
+  },
+  "peerDependenciesMeta": {
+    "webpack-virtual-modules": {
+      "optional": true
+    }
+  },
   "optionalDependencies": {
     "@datadog/libdatadog": "0.8.1",
     "@datadog/native-appsec": "11.0.1",
--- a/packages/datadog-unplugin/index.js
+++ b/packages/datadog-unplugin/index.js
@@ -63,8 +63,8 @@ module.exports = createUnplugin(() => {
      * virtual-module resolveId/load hooks and the ESBuild-specific plugin logic.
      */
     esbuild: {
-      onResolveFilter: /^$/,  // never matches — disables unplugin's onResolve wrapper
-      onLoadFilter: /^$/,     // never matches — disables unplugin's onLoad wrapper
+      onResolveFilter: /^$/, // never matches — disables unplugin's onResolve wrapper
+      onLoadFilter: /^$/, // never matches — disables unplugin's onLoad wrapper
       setup (build) {
         require('../datadog-esbuild/index.js').setup(build)
       },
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -132,7 +135,16 @@
...
+    "import-in-the-middle": "^2.0.6",
+    "unplugin": "^1.16.1"
+  },
+  "peerDependencies": {
+    "webpack-virtual-modules": ">=0.5.0"
+  },
+  "peerDependenciesMeta": {
+    "webpack-virtual-modules": {
+      "optional": true
+    }
```
tlhunter, line 147: This should all move to vendor/

esbuild.js
```diff
@@ -1,3 +1,3 @@
 'use strict'
 
-module.exports = require('./packages/datadog-esbuild/index.js')
+module.exports = require('./packages/datadog-unplugin/index.js').esbuild()
```
tlhunter, line 3: It looks like `./packages/datadog-esbuild` should be deleted too

.github/workflows/platform.yml
```diff
@@ -415,6 +426,7 @@ jobs:
       - run: sudo sysctl -w kernel.core_pattern='|/bin/false'
       - run: yarn test:integration
       - run: yarn test:integration:esbuild
+      - run: yarn test:integration:webpack
```
tlhunter, line 429: I don't think either webpack or esbuild should be listed in this section...

PR #7647, opened 2026-03-02
feat(bundler): experimental webpack bundling support

tlhunter:
Example Usage:

```js
const webpack = require('webpack')
const DDWebpack = require('dd-trace/webpack')

const compiler = webpack({
  mode: 'development',
  entry: 'foo.js',
  target: 'node',
  externalsType: 'commonjs',
  output: {
    filename: 'bundle.js',
    hashFunction: 'sha256',
  },
  externals: [
    // Node built-in not in webpack's default list for target: 'node'
    'diagnostics_channel',
    // optional native dd-trace modules
    '@datadog/native-appsec',
    '@datadog/native-iast-taint-tracking',
    '@datadog/native-metrics',
    '@datadog/pprof',
    '@datadog/libdatadog',
  ],
  plugins: [
    new DDWebpack(),
  ],
})
```


packages/datadog-webpack/index.js
```diff
@@ -0,0 +1,197 @@
...
+
+  /**
+   * @param {object} compiler
+   */
+  apply (compiler) {
+    // optimization.minimize is not yet set when apply() is called in webpack 5.54.0+
+    // (applyWebpackOptionsDefaults runs after plugins), so we defer the check to the
+    // environment hook which fires synchronously after defaults are applied.
+    compiler.hooks.environment.tap(PLUGIN_NAME, () => {
+      if (compiler.options.optimization?.minimize && !this.options.keepNames) {
```
tlhunter: Looks like a hallucination... `keepNames` is an ESBuild concept (which this code is trained on). Webpack has a `keep_classnames` setting but no `keepNames`.

.github/CODEOWNERS
```diff
@@ -36,9 +36,11 @@
 /integration-tests/esbuild/esm.integration.spec.js @DataDog/apm-idm-js
 /integration-tests/esbuild/index.spec.js @DataDog/apm-idm-js
 /integration-tests/esbuild/openfeature.spec.js @DataDog/apm-idm-js
+/integration-tests/webpack/index.spec.js @DataDog/apm-idm-js
```
watson: Any reason why it's not the entire folder?

```suggestion
/integration-tests/webpack/ @DataDog/apm-idm-js
```

watson: Unrelated to this PR, I recommend that @DataDog/apm-idm-js owns the entirety of `/integration-tests/esbuild/` as well.

tlhunter: I made the esbuild change in this PR too since they're tangentially related.

package.json
```diff
@@ -36,6 +36,8 @@
     "test:trace:guardrails:ci": "nyc -- npm run test:trace:guardrails",
     "test:esbuild": "mocha packages/datadog-esbuild/test/**/*.spec.js",
     "test:esbuild:ci": "nyc -- npm run test:esbuild",
+    "test:webpack": "mocha packages/datadog-webpack/test/**/*.spec.js",
+    "test:webpack:ci": "nyc -- npm run test:webpack",
```
IlyasShabi, line 41: Is it possible to have one script for all bundlers?

webpack.js
```diff
@@ -0,0 +1,3 @@
+'use strict'
```
IlyasShabi, line 1: what about bundlers/webpack.js for better structure

packages/datadog-webpack/index.js
```diff
@@ -0,0 +1,193 @@
...
+          return
+        }
+
+        if (!packageNamesOfInterest.has(pkg)) {
+          return
+        }
+
+        const request = resolveData.request
+
+        if (!modulesOfInterest.has(request) && !modulesOfInterest.has(`${pkg}/${modulePath}`)) {
```
IlyasShabi, line 145: I think we could remove `packageNamesOfInterest` based on this check, no?

package.json
tlhunter: I tried that approach but it was pretty messy. There's a tool called unplugin we can use but I wasn't able to get the code any cleaner.

https://github.com/DataDog/dd-trace-js/pull/7645

webpack.js
tlhunter: The location of this file is ultimately the way that the user consumes it. In this case they do `require('dd-trace/webpack')` which is pretty clean. Moving it to a bundlers/ directory makes the structure cleaner for us project devs but dirtier for the users who would need to do `require('dd-trace/bundlers/webpack')`.

And at the end of the day this file's location is just to avoid the user having to do `require('dd-trace/packages/datadog-webpack')`.

tlhunter:
I'm fine with sussing out if the IAST issue still exists after this gets merged. It's marked as experimental after all.

PR #7661, opened 2026-03-04
test(integration): use stopProc helper for process teardown

watson:
### What does this PR do?

Integration tests that spawn child processes now use the shared `stopProc` helper from `integration-tests/helpers` in their `after`/`afterEach` teardown instead of calling `proc.kill()` directly.

### Motivation

Using `proc.kill()` alone can leave processes or resources in an inconsistent state before the next test runs. Awaiting `stopProc(proc)` ensures child processes are fully stopped and cleaned up before the next test, which should reduce flakiness caused by one test leaking state into another.

### DIFF
--- a/integration-tests/aiguard/index.spec.js
+++ b/integration-tests/aiguard/index.spec.js
@@ -5,7 +5,7 @@ const path = require('path')
 
 const { after, afterEach, before, beforeEach, describe, it } = require('mocha')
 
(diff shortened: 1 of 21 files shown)


BridgeAR (commented):
LGTM, while I believe we should lower the default timeout to something much much smaller than 2 seconds.

szegedi (approved):

PR #7663, opened 2026-03-04
chore(ci) : add final_status property on junit XML [APMSP-2610]

cbeauchesne:
### What does this PR do?

Add a `<property name="dd_tags[test.final_status]" value="pass" />` on every test case reported through junit

### Motivation

Some repositories intentionally use mechanisms that ignore test failures (there are valid reasons to still execute those tests behind the scenes).

In these cases, the reported test status cannot be relied upon by Test Optimization to monitor repository health or trigger team notifications.

To ensure a consistent and unified approach, we will instead use final_status — the property specifically designed and used by the Test Optimization integration for this purpose.

### Important note

I'm using the not-that-ideal XSLT here as this change, because it must be done over [17+ repos](https://datadoghq.atlassian.net/browse/APMSP-2610), and it's a workaround until this gets integrated into a better place, such as datadog-ci or the backend. 

We're trying to work with the team on this, please have a look on this [RFC](https://docs.google.com/document/d/1OaX_h09fCXWmK_1ADrwvilt8Yt5h4WjC7UUAdS3Y3uw/edit?pli=1&tab=t.0#heading=h.tfy5viz7rz2) for more context.


.github/actions/instrumentations/test/action.yml
```diff
@@ -17,8 +17,7 @@ runs:
     - uses: ./.github/actions/coverage
       with:
         flags: instrumentations-${{ github.job }}
-    - uses: DataDog/junit-upload-github-action@055560f63c405095e9228ba443eee7987e22bb94  # v2.1.1
-      if: always() && github.actor != 'dependabot[bot]'
+    - uses: ./.github/actions/push_to_test_optim
```
BridgeAR: Nit: I think it would be better to use `push_to_test_optimization`. For someone who might not be familiar with this, it is likely that they have to briefly think what it stands for instead of knowing right away with the full name.

cbeauchesne: Totally agree, done in [0037831](https://github.com/DataDog/dd-trace-js/pull/7663/commits/00378315c7fe1e4812b411cd04b920c46897f7ec)

PR #7674, opened 2026-03-04
feat(logging): add hook-based log capture with HTTP forwarding

litianningdatadog:
[Background](https://docs.google.com/document/d/1XwYcUvXN5pJ0UBnduy7N8y6_xxLBFx7w1KGLQ4HWaP8/edit?tab=t.ibfcdqcri6k):

https://datadoghq.atlassian.net/browse/SVLS-8572

### What does this PR do?

Introduces a new log capture feature that forwards log records from Winston, Bunyan, and Pino to a configurable HTTP/HTTPS intake endpoint, without requiring per-instance transport injection.

**How it works:**
- Winston and Bunyan: captured via the existing `apm:<logger>:log` diagnostic channels (already firing for trace injection) inside the generic `LogPlugin` base class
- Pino: captured via a new `apm:pino:json` channel that publishes the complete serialized JSON line from `asJsonSym` for all Pino versions, bypassing the partial mixin data limitation of ≥5.14. This subscription lives in `PinoPlugin` (not `LogPlugin`) to keep the generic base free of pino-specific knowledge
- A new `log-capture/sender` module manages a batched NDJSON buffer with a one-shot `setTimeout` flush (armed only when records are buffered) and configurable max buffer size and HTTP/HTTPS transport

**Architecture:**
- `LogPlugin._captureEnabled` is a protected getter (returns `!!config.logCaptureEnabled`) that `PinoPlugin` overrides to return `false`, suppressing the generic capture path for pino while leaving trace injection intact
- `LogPlugin._configureSender(config)` is a protected method that configures the shared sender; it is a no-op when `logCaptureEnabled` is false — the sender goes dormant naturally once no more records are added
- `PINO_JSON_CHANNEL` is defined once in the pino instrumentation (the publisher) and imported by `PinoPlugin` (the subscriber)

**Configuration (env / init options):**
```
DD_LOG_CAPTURE_ENABLED / logCaptureEnabled       (default: true in Lambda Lite, false elsewhere)
DD_LOG_CAPTURE_HOST    / logCaptureHost           (default: localhost)
DD_LOG_CAPTURE_PORT    / logCapturePort           (default: 10517)
DD_LOG_CAPTURE_PATH    / logCapturePath           (default: /logs)
DD_LOG_CAPTURE_PROTOCOL / logCaptureProtocol      (default: http:)
DD_LOG_CAPTURE_MAX_BUFFER_SIZE / logCaptureMaxBufferSize (default: 1000)
DD_LOG_CAPTURE_FLUSH_INTERVAL_MS / logCaptureFlushIntervalMs (default: 5000)
DD_LOG_CAPTURE_TIMEOUT_MS / logCaptureTimeoutMs   (default: 5000)
```

### Motivation

Alternative to the transport-injection approach (PR #7522). By hooking into channels that already fire on every log write, this avoids per-instance injection, WeakSet de-duplication, and the duplicate-injection issues observed with Winston's `createLogger`.

### Additional Notes

**Key design decisions:**
- `logInjection` and `logCaptureEnabled` are independent — enabling only capture does not inject `dd` into the actual log output, but `dd` trace context is always included in the captured JSON sent to the intake
- Pino ≥5.14: `mixinSym` (used for trace injection) only provides partial mixin fields. A separate `wrapAsJsonForCapture` wrapper on `asJsonSym` captures the complete JSON line post-serialization and publishes to `apm:pino:json`
- The sender uses a one-shot `setTimeout` (re-armed only when records are buffered) rather than `setInterval`, avoiding idle event-loop wakeups. There is no `stop()` — when capture is disabled, records simply stop being added and the sender goes dormant
- `flush` is registered in `beforeExitHandlers` once when first configured and never removed, ensuring buffered records are flushed on process exit regardless of runtime config changes
- NDJSON body trims trailing `\n` from each record before joining (Pino's `asJson` appends `\n`; naïve join would produce double blank lines)
- HTTP response body is drained via `res.resume()` to release sockets back to the pool
- All `logCapture*` config fields are always forwarded through `plugin_manager._getSharedConfig()` (no conditional guard)

**Performance:** Pino overhead is ~200–400 ns/write (no extra serialization). Winston/Bunyan overhead is ~3–25 µs/write dominated by `JSON.stringify`.


packages/dd-trace/src/log-capture/sender.js
```diff
@@ -0,0 +1,132 @@
...
+ */
+function add (jsonLine) {
+  if (opts === undefined) return
+  if (buffer.length >= opts.maxBufferSize) {
+    flush()
+  }
+  buffer.push(jsonLine)
+  if (timer === undefined) {
+    timer = setTimeout(flushAndReschedule, opts.flushIntervalMs)
+    timer.unref()
```
rochdev:
```suggestion
    timer.unref?.()
```

packages/dd-trace/src/log-capture/sender.js
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * Flush buffered records and re-arm the timer if more records arrived during flush.
+ */
+function flushAndReschedule () {
+  timer = undefined
+  flush()
+  if (buffer.length > 0) {
+    timer = setTimeout(flushAndReschedule, opts.flushIntervalMs)
+    timer.unref()
```
rochdev:
```suggestion
    timer.unref?.()
```

litianningdatadog: Acked

packages/dd-trace/src/log-capture/sender.js
litianningdatadog: Acked

packages/datadog-plugin-pino/src/index.js
```diff
@@ -1,9 +1,82 @@
 'use strict'
 
+const { storage } = require('../../datadog-core')
+const { LOG } = require('../../../ext/formats')
 const LogPlugin = require('../../dd-trace/src/plugins/log_plugin')
+const { Subscription } = require('../../dd-trace/src/plugins/plugin')
+const log = require('../../dd-trace/src/log')
+const { PINO_JSON_CHANNEL } = require('../../datadog-instrumentations/src/pino')
```
rochdev: Instrumentation and plugins should never interact other than through a channel.

packages/dd-trace/test/plugins/plugin.spec.js
```diff
@@ -17,8 +17,47 @@ const log = {
 const Plugin = proxyquire('../../src/plugins/plugin', {
   '../log': log,
 })
+const { Subscription } = Plugin
 const { storage } = require('../../../datadog-core')
 
+describe('Subscription', () => {
```
rochdev: I think this is only needed because `Subscription` was used directly. Instead, `this.addSub()` should be used.

packages/dd-trace/src/plugins/plugin.js
```diff
@@ -191,3 +203,6 @@ module.exports = class Plugin {
     }
   }
 }
+
+module.exports = Plugin
+module.exports.Subscription = Subscription
```
rochdev: This should not be needed, all subscriptions should be done through `addSub`.

packages/dd-trace/src/plugins/log_plugin.js
```diff
@@ -37,18 +38,84 @@ module.exports = class LogPlugin extends Plugin {
...
+      }
+
+      // Forward to capture sender.
+      // Subclasses may override _captureEnabled to suppress this path
+      // and handle capture via their own channel instead.
+      if (this._captureEnabled) {
+        try {
+          // Always include dd trace context in captured records even if logInjection is off.
+          const msg = this.config.logInjection ? arg.message : messageProxy(arg.message, holder)
+          require('../log-capture/sender').add(JSON.stringify(msg))
```
rochdev: Why require here instead of at the top of the file? This could be a hot path and it's better to avoid unnecessary abstractions.

packages/datadog-plugin-pino/src/index.js
```diff
@@ -1,9 +1,83 @@
...
+        }
+      }
+      sender.add(json)
+    })
+    // Subscription starts disabled (_enabled = false). configure() manages its lifecycle.
+  }
+
+  /**
+   * Override to prevent the generic apm:${id}:log capture path from running for pino.
+   * Pino capture is handled via the apm:pino:json channel in #jsonCaptureSub.
```
rochdev: What about cases where `json` is not used at all? Do we just skip these logs entirely? If we do, then that implementation seems incomplete? And if we don't, then why even bother with `log`? We could just use `json` for everything at that point.

packages/dd-trace/src/plugins/log_plugin.js
```diff
@@ -37,18 +38,84 @@ module.exports = class LogPlugin extends Plugin {
...
+   * once no more records are added to it.
+   *
+   * @param {object} config Plugin configuration object.
+   * @returns {void}
+   */
+  _configureSender (config) {
+    if (!config.logCaptureEnabled) return
+
+    const captureSender = require('../log-capture/sender')
+    captureSender.configure({
```
rochdev: This should happen outside of plugins. Right now the tracer gets these options, passes them to the plugins so that the plugins can then go back to the tracer to configure it. This not only requires unnecessary code but it also technically broken because it means 2 plugins could be configured differently and then the last one wins which is broken.

packages/datadog-plugin-pino/src/index.js
litianningdatadog: Good question. apm:pino:json actually fires for all standard pino output, including pino-pretty, because pino always serializes internally via asJson before writing to any estination. So no logs are silently dropped.  The reason we can't use apm:pino:log instead is that for pino ≥5.14 it fires from the mixin hook, which only carries partial data (the mixin fields, not the full log record). apm:pino:json fires from the asJson wrapper and always has the complete record, which is why we use it exclusively for capture.

packages/dd-trace/src/plugins/log_plugin.js
litianningdatadog: acked

packages/dd-trace/src/plugins/plugin.js
litianningdatadog: acked

packages/dd-trace/test/plugins/plugin.spec.js
litianningdatadog: acked

packages/dd-trace/src/plugins/log_plugin.js
litianningdatadog: acked

packages/dd-trace/test/plugins/plugin.spec.js
```diff
@@ -19,6 +19,53 @@ const Plugin = proxyquire('../../src/plugins/plugin', {
...
+
+  class SubTestPlugin extends Plugin {
+    static id = 'subTest'
+  }
+
+  afterEach(() => {
+    plugin.configure({ enabled: false })
+  })
+
+  it('should not subscribe twice when enabled multiple times', () => {
```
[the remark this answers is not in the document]
litianningdatadog: acked

PR #7675, opened 2026-03-04
refactor: drop unused noop scope file

watson:
### What does this PR do?

Removes the unused noop scope implementation at `packages/dd-trace/src/scope/noop/scope.js`.

### Motivation

`packages/dd-trace/src/scope/noop/scope.js` is not referenced anywhere in the codebase.


BridgeAR (changes requested):
I find a reference in `packages/dd-trace/src/noop/tracer.js`? That seems to be used in the cypress plugin.

I am wondering about the tests not running. They should fail right away (noop.spec.js)

PR #7685, opened 2026-03-05
[test optimization] Update extraction of `ci.job.url` in github actions

cbasitodx:
### What does this PR do?
Extracts the Github job numeric identifier through a parse of the runner diagnostic files. 

### Motivation
Test Optimization libraries tag test events with a `ci.job.url` tag. This job url should be constructed as `https://github.com/REPO_NAME/actions/runs/RUN_ID/job/JOB_ID`. Since `JOB_ID` is not accesible through the enviroment, the `ci.job.url` tag was forced to be a generic one.

### DIFF
--- a/packages/dd-trace/src/opentracing/propagation/text_map.js
+++ b/packages/dd-trace/src/opentracing/propagation/text_map.js
@@ -667,6 +667,7 @@ class TextMapPropagator {
   }
 
   _extractBaggageItems (carrier, spanContext) {
+    removeAllBaggageItems()
     if (!this._hasPropagationStyle('extract', 'baggage')) return
     if (!carrier?.baggage) return
     const baggages = carrier.baggage.split(',')
--- a/packages/dd-trace/test/opentracing/propagation/text_map.spec.js
+++ b/packages/dd-trace/test/opentracing/propagation/text_map.spec.js
@@ -750,6 +750,37 @@ describe('TextMapPropagator', () => {
       assert.deepStrictEqual(getAllBaggageItems(), { foo: 'bar' })
     })
 
+    it('should clear pre-existing baggage items before extracting new ones', () => {
+      removeAllBaggageItems()
+      setBaggageItem('stale', 'leftover')
+      setBaggageItem('foo', 'old-value')
+      assert.deepStrictEqual(getAllBaggageItems(), { stale: 'leftover', foo: 'old-value' })
+
+      const carrier = {
+        'x-datadog-trace-id': '123',
+        'x-datadog-parent-id': '456',
+        baggage: 'foo=new-value,fresh=added',
+      }
+      propagator.extract(carrier)
+
+      assert.deepStrictEqual(getAllBaggageItems(), { foo: 'new-value', fresh: 'added' })
+      assert.strictEqual(getBaggageItem('stale'), undefined)
+    })
+
+    it('should clear pre-existing baggage items when carrier has no baggage header', () => {
+      removeAllBaggageItems()
+      setBaggageItem('stale', 'leftover')
+      assert.deepStrictEqual(getAllBaggageItems(), { stale: 'leftover' })
+
+      const carrier = {
+        'x-datadog-trace-id': '123',
+        'x-datadog-parent-id': '456',
+      }
+      propagator.extract(carrier)
+
+      assert.deepStrictEqual(getAllBaggageItems(), {})
+    })
+
     it('should convert signed IDs to unsigned', () => {
       textMap['x-datadog-trace-id'] = '-123'
       textMap['x-datadog-parent-id'] = '-456'


packages/dd-trace/src/plugins/util/ci.js
```diff
@@ -102,8 +103,106 @@ function getGitHubEventPayload () {
   return JSON.parse(readFileSync(path, 'utf8'))
 }
 
+function getJobIDFromDiagFile (runnerTemp) {
+  // There should be an if statement here for checking wether the env job id var is available or not.
+  // If it is available, just return that... This could be added whenever Github decides to (check out https://github.com/actions/runner/pull/4053)
+
+  // Since said env variable is not available, we'll extract the Job ID from a diagnostics file
```
juan-fernandez:
```suggestion
  // TODO: read job id directly from the environment.
  // Check https://github.com/actions/runner/pull/4053 for updates
```

packages/dd-trace/src/plugins/util/ci.js
```diff
@@ -102,8 +103,106 @@ function getGitHubEventPayload () {
...
+      path.join(runnerRoot, '_diag')
+    )
+  } else {
+    dirs.push(
+      path.join(runnerRoot, 'cached', '_diag'),
+      path.join(runnerRoot, '_diag'),
+      path.join(runnerRoot, 'actions-runner', 'cached', '_diag'),
+      path.join(runnerRoot, 'actions-runner', '_diag')
+    )
+  }
```
juan-fernandez: you can define `dirs` here since it's not used before and save some lines. Something like: 
```suggestion
  const dirs = [
    path.join(runnerRoot, 'cached', '_diag')
    path.join(runnerRoot, '_diag')
  ]
  if (!isWin) {
    dirs.push(
      path.join(runnerRoot, 'actions-runner', 'cached', '_diag'),
      path.join(runnerRoot, 'actions-runner', '_diag')
    )
  }
```

packages/dd-trace/src/plugins/util/ci.js
```diff
@@ -102,8 +103,106 @@ function getGitHubEventPayload () {
...
+  let chosenDiagPath = ''
+
+  for (const diagPath of possibleDiagsPaths) {
+    try {
+      // Obtain a list of fs.Dirent objects of the files in diagPath
+      const files = readdirSync(diagPath, { withFileTypes: true })
+
+      // Check if there are valid pontential log files
+      const potentialLogs = files
+        .filter((file) => file.isFile() && file.name.startsWith('Worker_')) // && file.name.endsWith('.log'))
```
juan-fernandez:
```suggestion
        .filter((file) => file.isFile() && file.name.startsWith('Worker_'))
```

packages/dd-trace/src/plugins/util/ci.js
```diff
@@ -102,8 +103,106 @@ function getGitHubEventPayload () {
...
+
+  // This will hold the chosen diagnostics path (between the ones that are contemplated in possibleDiagsPath)
+  let chosenDiagPath = ''
+
+  for (const diagPath of possibleDiagsPaths) {
+    try {
+      // Obtain a list of fs.Dirent objects of the files in diagPath
+      const files = readdirSync(diagPath, { withFileTypes: true })
+
+      // Check if there are valid pontential log files
```
juan-fernandez:
```suggestion
      // Check if there are valid potential log files
```

packages/dd-trace/src/plugins/util/ci.js
juan-fernandez: on a second look: `datadog-ci` checks the extension. Why not us? https://github.com/DataDog/datadog-ci/blob/d05ee839b8ce6fc96adb892efa5e898c96aedfce/packages/base/src/helpers/ci.ts#L1132

packages/dd-trace/src/plugins/util/ci.js
```diff
@@ -102,8 +103,106 @@ function getGitHubEventPayload () {
...
+      return null
+    }
+  }
+
+  // Get the job ID via regex
+  for (const logFile of workerLogFiles) {
+    const filePath = path.posix.join(chosenDiagPath, logFile)
+    const content = readFileSync(filePath, 'utf8')
+
+    const match = content.match(/"job":\s*{[\s\S]*?"v"\s*:\s*(\d+)(?:\.0)?/)
```
juan-fernandez, line 180: `datadog-ci` uses a different regex: shouldn't we use the same? https://github.com/DataDog/datadog-ci/blob/d05ee839b8ce6fc96adb892efa5e898c96aedfce/packages/base/src/helpers/ci.ts#L81

packages/dd-trace/src/plugins/util/ci.js
```diff
@@ -102,8 +103,106 @@ function getGitHubEventPayload () {
...
+
+    // match[1] is the captured group with the display name
+    if (match && match[1]) { return match[1] }
+  }
+
+  return null // Both of the directories were empty
```
juan-fernandez: it's not just two now, is it? 
```suggestion
  return null
```

packages/dd-trace/test/plugins/util/fixtures/.gitignore
```diff
@@ -0,0 +1 @@
+!*.log
```
juan-fernandez, line 2: why do we need this `.gitignore`?

packages/dd-trace/test/plugins/util/test-environment.spec.js
```diff
@@ -112,3 +112,20 @@ describe('test environment data', () => {
...
+    { runnerTemp: null, expected: null },                                                                                                                                                                     
+    { runnerTemp: undefined, expected: null },
+  ]
+
+  for (const { runnerTemp, expected } of runnerTempPaths) {
+    it(`returns ${expected} for runnerTemp: ${runnerTemp}`, () => {
+      assert.strictEqual(getJobIDFromDiagFile(runnerTemp), expected)
+    })
+  }
+})
```
juan-fernandez: test looks good! Let's review the linter though

juan-fernandez (commented):
great job! I left some minor comments. We're almost there 😄

packages/dd-trace/test/plugins/util/fixtures/.gitignore
cbasitodx: I've added it because I created a new file for testing (`Worker_20240115-102345-12345.log`) and, because of the hierarchy of the `.gitignore`s, it was not being tracked. Therefore, I added that `.gitignore` so mock diagnostic files in the `fixtures` directory are not ignored (thus the `!`). However, I think I am going to move it inside the `runner` directory, that way there is no chance it affects other files in `fixtures` :)

packages/dd-trace/src/plugins/util/ci.js
cbasitodx: Based on the discussion held on the RFC, apparently the part we are looking for is available in all `Worker_*` files. I don't know if by adding the extension checking we could be missing on some files? 🤔

packages/dd-trace/test/plugins/util/fixtures/.gitignore
juan-fernandez: that makes sense, thanks!

juan-fernandez, line 2: I'd add a small comment explaining why
```suggestion
# Some fixtures are log files and logs are ignored in the global `.gitignore`
!*.log
```

packages/dd-trace/src/plugins/util/ci.js
juan-fernandez: if we are not using the extension, let's remove the comment

packages/dd-trace/test/plugins/util/fixtures/.gitignore
juan-fernandez, line 2: should we remove this one? Now we have 2 gitignores

packages/dd-trace/test/plugins/util/test-environment.spec.js
```diff
@@ -112,3 +112,19 @@ describe('test environment data', () => {
...
   })
 })
+
+describe('test getJobIDFromDiagFile function', () => {
+  const TEST_HOME = 'packages/dd-trace/test/plugins/util/fixtures'
+
+  const runnerTempPaths = [
+    { runnerTemp: path.join(TEST_HOME, '/runner/work/_temp'), expected: '9876543210' },
+    { runnerTemp: null, expected: null },
+    { runnerTemp: undefined, expected: null },
```
juan-fernandez, line 122: let's also add a case where the folders do not exist, something like 
```javascript
{ runnerTemp: path.join(TEST_HOME, '/runner/work/_do_not_exist'), expected: null },
```
and another one for a log file that does not include what we want
```javascript
{ runnerTemp: path.join(TEST_HOME, '/runner/work/Worker_empty.log'), expected: null },
```

you'd have to add an empty `Worker_empty.log` fixture file

packages/dd-trace/test/plugins/util/fixtures/runner/.gitignore
```diff
@@ -0,0 +1 @@
+!*.log
```
cbasitodx: my bad. I missed it. I'm removing the other one :)

packages/dd-trace/test/plugins/util/test-environment.spec.js
```diff
@@ -112,3 +112,21 @@ describe('test environment data', () => {
     })
   })
 })
+
+describe('test getJobIDFromDiagFile function', () => {
+  const TEST_HOME = 'packages/dd-trace/test/plugins/util/fixtures'
```
juan-fernandez:
```suggestion
  const TEST_HOME = path.join(__dirname, 'fixtures')
```

PR #7692, opened 2026-03-05
feat(eslint): add no-does-not-throw rule with auto-fix

bm1549:
## Summary

- Enforces the project convention of not using `assert.doesNotThrow()` in test files (per CLAUDE.md)
- Uses the built-in ESLint `no-restricted-syntax` rule with an AST selector — no custom rule needed
- Fixed all ~50 existing violations across 25 test files (callback bodies extracted and called directly)
- Cascading cleanups: removed 3 now-unused `assert` imports, converted 2 `let` → `const`

### Fix patterns applied

| Before | After |
|--------|-------|
| `assert.doesNotThrow(() => expr)` | `expr` |
| `assert.doesNotThrow(() => { stmt1; stmt2 })` | `stmt1`<br>`stmt2` |
| `assert.doesNotThrow(fn)` | `fn()` |
| `assert.doesNotThrow(obj.method)` | `obj.method()` |

## Test plan

- [x] `npm run lint` passes clean
- [x] Verified all 25 fixed test files have semantically equivalent transformations
- [x] Verified cascading cleanups are correct (unused imports, let→const)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -115,8 +115,6 @@ Use `node:assert/strict` for standard assertions. For partial deep object checks
 
 Favor fewer `assert.deepStrictEqual`/`assertObjectContains` calls over many `assert.strictEqual` calls. Combine existing calls, when touching test files.
 
-Never use the `doesNotThrow()` assertion. Instead, execute the method directly.
-
 ### Time-Based Testing
 
 **Never rely on actual time passing in unit tests.** Use sinon's fake timers to mock time and make tests deterministic and fast.


packages/dd-trace/test/appsec/iast/taint-tracking/rewriter.spec.js
```diff
@@ -16,10 +16,7 @@ const iastEnabledConfig = {
 
 describe('IAST Rewriter', () => {
   it('Addon should return a rewritter instance', () => {
-    let rewriter = null
-    assert.doesNotThrow(() => {
-      rewriter = require('@datadog/wasm-js-rewriter')
-    }, Error)
+    const rewriter = require('@datadog/wasm-js-rewriter')
     assert.notStrictEqual(rewriter, null)
```
BridgeAR: Could we change this to
```suggestion
    assert.ok(rewriter)
```

That way it is a safer check :)

PR #7700, opened 2026-03-06
[test optimization] Allow keeping user coverage configuration even if TIA is enabled

juan-fernandez:
### What does this PR do?

Add `DD_TEST_TIA_KEEP_COV_CONFIG` support to Jest TIA coverage handling and refines behavior based on whether coverage was enabled by dd-trace or by the user.

### Motivation

Users need a way to preserve their Jest coverage configuration during TIA runs, while still allowing dd-trace to optimize coverage work when coverage was only enabled internally.

### Additional Notes

Add `DD_TEST_TIA_KEEP_COV_CONFIG` to supported configurations.
Behavior now:
  - If set to true: only enforce `collectCoverage` + `passWithNoTests`.
  - If not set to true: apply `coverageReporters: ['none']` and skip `_addUntestedFiles` only when the user had **not** already configured `jest` to run with code coverage.


packages/datadog-instrumentations/src/jest.js
```diff
@@ -109,6 +111,7 @@ const testSuiteJestObjects = new Map()
 const BREAKPOINT_HIT_GRACE_PERIOD_MS = 200
 const ATR_RETRY_SUPPRESSION_FLAG = '_ddDisableAtrRetry'
 const atrSuppressedErrors = new Map()
+const isKeepingCoverageConfiguration = getEnvironmentVariable('DD_TEST_TIA_KEEP_COV_CONFIG') === 'true'
```
BridgeAR:
```suggestion
const isKeepingCoverageConfiguration = isTrue(getEnvironmentVariable('DD_TEST_TIA_KEEP_COV_CONFIG'))
```

Please move this into the config file, since it would only receive the proper telemetry when being in there right now.

packages/dd-trace/src/config/supported-configurations.json
```diff
@@ -1770,6 +1770,13 @@
         ]
       }
     ],
+    "DD_TEST_TIA_KEEP_COV_CONFIG": [
+      {
+        "implementation": "A",
+        "type": "boolean",
+        "default": "false"
```
BridgeAR: Please add the configuration name as well (it will soon be renamed to `internalPropertyName` in case it is not exposed as option, but for now using the `configurationNames` would be sufficient.

packages/dd-trace/src/config/index.js
```diff
@@ -1126,6 +1126,7 @@ class Config {
       setBoolean(calc, 'isIntelligentTestRunnerEnabled', isTrue(isCiVisibilityItrEnabled()))
       setBoolean(calc, 'isManualApiEnabled', !isFalse(getEnv('DD_CIVISIBILITY_MANUAL_API_ENABLED')))
       setString(calc, 'ciVisibilityTestSessionName', getEnv('DD_TEST_SESSION_NAME'))
+      setBoolean(calc, 'isKeepingCoverageConfiguration', isTrue(getEnv('DD_TEST_TIA_KEEP_COV_CONFIG')))
```
BridgeAR: This seems like it is the wrong spot. It would be best to be placed inside of the environment part. Calculated values are only about combining or dependent configs. The user does define the value one way or the other, no matter if ci vis is enabled or not, so having that data would be good :)

packages/dd-trace/src/config/supported-configurations.json
juan-fernandez: https://github.com/DataDog/dd-trace-js/pull/7700/changes/7fbb9cb6045f7747096943fd90016a89a569038c

packages/dd-trace/src/config/index.js
juan-fernandez: > It would be best to be placed inside of the environment part 

where's that part?

juan-fernandez: do you mean `applyConfigValues` ? https://github.com/DataDog/dd-trace-js/blob/44733b7389fc5e06db3d6abf10f38ef13678e95b/packages/dd-trace/src/config/index.js#L246

BridgeAR: Yes, exactly that one :)

juan-fernandez: done in https://github.com/DataDog/dd-trace-js/pull/7700/changes/9dfa08c2ba61d26193bd89112816558d7a56055f

anmarchenko (approved):

PR #7701, opened 2026-03-06
More test optimization fixes to CODEOWNERS

juan-fernandez:
### What does this PR do?
Remove manual api test path as it was wrong and the actual path is covered by the `/integration-tests/ci-visibility/` entry already. 

### Motivation
I missed this one too

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -114,7 +114,6 @@
 /integration-tests/cypress/cypress.spec.js @DataDog/ci-app-libraries
 /integration-tests/vitest/vitest.spec.js @DataDog/ci-app-libraries
 /integration-tests/vitest.config.mjs @DataDog/ci-app-libraries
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -114,7 +114,7 @@
 /integration-tests/cypress/cypress.spec.js @DataDog/ci-app-libraries
 /integration-tests/vitest/vitest.spec.js @DataDog/ci-app-libraries
 /integration-tests/vitest.config.mjs @DataDog/ci-app-libraries
-/integration-tests/test-api-manual.spec.js @DataDog/ci-app-libraries
+/integration-tests/ci-visibility/test-api-manual.spec.js @DataDog/ci-app-libraries
```
watson: Actually, we already have a catch all for everything under `/integration-tests/ci-visibility/`, so this should just be deleted

juan-fernandez: yeah good catch, already pushed the fix

PR #7705, opened 2026-03-06
feat(eslint): add prefer-private-class-fields rule with auto-fix

bm1549:
## Summary

- Adds `eslint-prefer-private-class-fields`, a custom ESLint rule that detects underscore-prefixed class properties (`_foo`) which are only used within a single class, and enforces conversion to private class fields (`#foo`)
- Includes an auto-fix that renames all occurrences (field declarations, method definitions, and member-expression accesses) in a single lint pass
- Ran the auto-fix across the entire codebase, converting **62 source files** and renaming **585** underscore-prefixed properties to runtime-enforced private fields

## Rule design

The rule is intentionally conservative to avoid false positives:

| Guard | Reason |
|---|---|
| Requires an explicit `PropertyDefinition` or `MethodDefinition` | Avoids synthesizing new `#foo` class field declarations where none existed |
| Requires at least one `MemberExpression` access | Skips definition-only properties that may be a public API consumed from other files |
| Blocks on `super._foo` | `super.#foo` is syntactically invalid; private fields cannot be accessed via `super` |
| Blocks on destructuring keys (`{ _foo } = this`) | Private fields cannot be destructured |
| Blocks when accessed from more than one class in the file | Cross-class access is semantically external |

Each occurrence is reported as a **separate problem with a single-node fix**, which allows ESLint to apply all fixes for a file in one pass without range conflicts (a known issue with generator-based multi-fix approaches where ESLint merges yields into a wide range that can overlap with other problems).

## Test plan

- [x] All ESLint rule unit tests pass (`npm run test:eslint-rules`)
- [x] `npm run lint` reports zero violations after the auto-fix
- [x] Verified that `super._foo` and destructuring cases are correctly excluded
- [x] Verified that definition-only methods (no call sites) are not privatized

### DIFF
--- a/eslint-rules/eslint-prefer-private-class-fields.mjs
+++ b/eslint-rules/eslint-prefer-private-class-fields.mjs
@@ -6,10 +6,12 @@
  * while underscore convention (_prop) is advisory only.
  *
  * The rule is intentionally conservative:
- * - Only flags when an explicit class field declaration (`_foo = value`) or method definition
- *   (`_foo() {}`) exists. Constructor-only assignments (`this._foo = 1`) without a class
- *   field declaration are not flagged, since the auto-fix would produce invalid JavaScript
- *   without also synthesizing a new `#foo` field declaration.
+ * - Only flags when an explicit class field declaration (`_foo = value`) exists.
+ *   Constructor-only assignments (`this._foo = 1`) and method definitions (`_foo() {}`)
+ *   are not flagged. For constructor-only assignments, the auto-fix would produce invalid
+ *   JavaScript without also synthesizing a new `#foo` field declaration. For methods,
+ *   test files often use sinon.stub(instance, '_methodName') which cannot stub private
+ *   methods — privatizing methods would break test stubs.
  * - Only flags when ALL accesses to that name in the file are within the body of exactly
  *   one class. Any external access, access in a different class, or destructuring use
  *   (incompatible with private fields) suppresses the warning for that property name.
@@ -117,21 +119,6 @@ export default {
       },
 
       /**
-       * Track method definitions: _foo() {}, get _foo() {}, set _foo(v) {}.
-       *
-       * @param {import('eslint').Rule.Node} node
-       */
-      MethodDefinition (node) {
-        if (node.computed) return
-        if (node.key.type !== 'Identifier') return
-        const { name } = node.key
-        if (!name.startsWith('_') || name === '_') return
-
-        const enclosingClass = getEnclosingClass(node)
-        recordSource(name, node, enclosingClass)
-      },
-
-      /**
        * Block properties that appear as destructuring keys.
        * Private fields cannot be destructured, so `const { _foo } = this` cannot be
        * automatically fixed to `const { #foo } = this`.
@@ -161,12 +148,12 @@ export default {
           const classes = new Set(sources.map(s => s.enclosingClass))
           if (classes.size !== 1) continue
 
-          // Only flag when an explicit class field or method definition exists.
+          // Only flag when an explicit class field declaration exists (PropertyDefinition).
           // Without a PropertyDefinition, replacing this._foo with this.#foo produces a
           // SyntaxError because private fields must be declared in the class body.
-          const hasDefinition = sources.some(
-            s => s.node.type === 'PropertyDefinition' || s.node.type === 'MethodDefinition'
-          )
+          // MethodDefinition (methods) are intentionally excluded: test files often use
+          // sinon.stub(instance, '_methodName') which cannot stub private methods.
+          const hasDefinition = sources.some(s => s.node.type === 'PropertyDefinition')
           if (!hasDefinition) continue
 
           // Only flag when there is at least one MemberExpression access (this._foo, obj._foo).
--- a/eslint-rules/eslint-prefer-private-class-fields.mjs
+++ b/eslint-rules/eslint-prefer-private-class-fields.mjs
@@ -1,0 +1,202 @@
+/**
+ * Enforce use of private class fields (#prop) over underscore-prefixed properties (_prop)
+ * when the property is only accessed within a single class in the same file.
+ *
+ * Private fields (#prop) are enforced by the runtime (SyntaxError if accessed externally),
+ * while underscore convention (_prop) is advisory only.
+ *
+ * The rule is intentionally conservative:
+ * - Only flags when an explicit class field declaration (`_foo = value`) or method definition
+ *   (`_foo() {}`) exists. Constructor-only assignments (`this._foo = 1`) without a class
+ *   field declaration are not flagged, since the auto-fix would produce invalid JavaScript
+ *   without also synthesizing a new `#foo` field declaration.
+ * - Only flags when ALL accesses to that name in the file are within the body of exactly
+ *   one class. Any external access, access in a different class, or destructuring use
+ *   (incompatible with private fields) suppresses the warning for that property name.
+ *
+ * Each occurrence (definition + every access) is reported as a separate problem, each with
+ * its own single-node fix. This lets ESLint apply all fixes in one pass without range conflicts.
+ */
+
+/**
+ * @typedef {{ enclosingClass: import('eslint').Rule.Node | null, node: import('eslint').Rule.Node }} PropSource
+ */
+
+export default {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description:
+        'Prefer private class fields (#prop) over underscore-prefixed properties (_prop) ' +
+        'when the property is only accessed within a single class',
+      recommended: false,
+    },
+    fixable: 'code',
+    messages: {
+      preferPrivate:
+        'Property "{{name}}" is only used within class "{{className}}". ' +
+        'Use private field "#{{privateName}}" instead of the underscore convention.',
+    },
+    schema: [],
+  },
+
+  create (context) {
+    /** @type {Map<string, PropSource[]>} */
+    const propSources = new Map()
+
+    /** @type {Set<string>} Property names that cannot be privatized (e.g. destructured) */
+    const blocked = new Set()
+
+    /**
+     * Walk up the parent chain to find the nearest enclosing ClassDeclaration or ClassExpression.
+     *
+     * @param {import('eslint').Rule.Node} node
+     * @returns {import('eslint').Rule.Node | null}
+     */
+    function getEnclosingClass (node) {
+      let current = node.parent
+      while (current) {
+        if (current.type === 'ClassBody') {
+          return current.parent
+        }
+        current = current.parent
+      }
+      return null
+    }
+
+    /**
+     * @param {string} propName
+     * @param {import('eslint').Rule.Node} node
+     * @param {import('eslint').Rule.Node | null} enclosingClass
+     */
+    function recordSource (propName, node, enclosingClass) {
+      if (!propSources.has(propName)) {
+        propSources.set(propName, [])
+      }
+      propSources.get(propName).push({ enclosingClass, node })
+    }
+
+    return {
+      /**
+       * Track all member-expression accesses: this._foo, obj._foo, etc.
+       *
+       * @param {import('eslint').Rule.Node} node
+       */
+      MemberExpression (node) {
+        if (node.computed) return
+        if (node.property.type !== 'Identifier') return
+        const { name } = node.property
+        if (!name.startsWith('_') || name === '_') return
+
+        // `super._foo` means this property is defined on the parent class.
+        // Converting it to `super.#foo` is syntactically invalid — private fields
+        // cannot be accessed via `super`. Block the property entirely.
+        if (node.object.type === 'Super') {
+          blocked.add(name)
+          return
+        }
+
+        const enclosingClass = getEnclosingClass(node)
+        recordSource(name, node, enclosingClass)
+      },
+
+      /**
+       * Track class field declarations: _foo = value.
+       * These are `PropertyDefinition` nodes in the class body.
+       *
+       * @param {import('eslint').Rule.Node} node
+       */
+      PropertyDefinition (node) {
+        if (node.computed) return
+        if (node.key.type !== 'Identifier') return
+        const { name } = node.key
+        if (!name.startsWith('_') || name === '_') return
+
+        const enclosingClass = getEnclosingClass(node)
+        recordSource(name, node, enclosingClass)
+      },
+
+      /**
+       * Track method definitions: _foo() {}, get _foo() {}, set _foo(v) {}.
+       *
+       * @param {import('eslint').Rule.Node} node
+       */
+      MethodDefinition (node) {
+        if (node.computed) return
+        if (node.key.type !== 'Identifier') return
+        const { name } = node.key
+        if (!name.startsWith('_') || name === '_') return
+
+        const enclosingClass = getEnclosingClass(node)
+        recordSource(name, node, enclosingClass)
+      },
+
+      /**
+       * Block properties that appear as destructuring keys.
+       * Private fields cannot be destructured, so `const { _foo } = this` cannot be
+       * automatically fixed to `const { #foo } = this`.
+       *
+       * @param {import('eslint').Rule.Node} node
+       */
+      Property (node) {
+        if (node.parent?.type !== 'ObjectPattern') return
+        if (node.computed) return
+        if (node.key.type !== 'Identifier') return
+        const { name } = node.key
+        if (!name.startsWith('_') || name === '_') return
+
+        blocked.add(name)
+      },
+
+      'Program:exit' () {
+        for (const [propName, sources] of propSources) {
+          // Cannot be privatized: appears as a destructuring key
+          if (blocked.has(propName)) continue
+
+          // Cannot be privatized: accessed outside any class body
+          const hasExternalAccess = sources.some(s => !s.enclosingClass)
+          if (hasExternalAccess) continue
+
+          // Cannot be privatized: referenced in more than one class
+          const classes = new Set(sources.map(s => s.enclosingClass))
+          if (classes.size !== 1) continue
+
+          // Only flag when an explicit class field or method definition exists.
+          // Without a PropertyDefinition, replacing this._foo with this.#foo produces a
+          // SyntaxError because private fields must be declared in the class body.
+          const hasDefinition = sources.some(
+            s => s.node.type === 'PropertyDefinition' || s.node.type === 'MethodDefinition'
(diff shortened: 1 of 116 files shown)


eslint-rules/eslint-prefer-private-class-fields.mjs
```diff
@@ -0,0 +1,278 @@
...
+ *   JavaScript without also synthesizing a new `#foo` field declaration. For methods,
+ *   test files often use sinon.stub(instance, '_methodName') which cannot stub private
+ *   methods — privatizing methods would break test stubs.
+ * - Only flags when ALL accesses to that name in the file are within the body of exactly
+ *   one class. Any external access, access in a different class, or destructuring use
+ *   (incompatible with private fields) suppresses the warning for that property name.
+ * - Never flags fields in exported classes. Exported classes may be subclassed in other
+ *   files, or their fields may be set/read directly by test code (e.g., `instance._field = x`
+ *   or `sinon.stub(instance, '_field')`). Cross-file access is invisible to this single-file
+ *   analysis rule, so exported classes are skipped entirely.
```
BridgeAR, line 21: I am not sure if this is useful as eslint rule in that case (while it does not hurt either). Maybe it is best to replace as much as possible with a eslint rule? That way the LLM and users will see the private properties more and will use that as reference instead of needing the LLM.

We just have to refactor more code to get there.

packages/dd-trace/src/ci-visibility/exporters/ci-visibility-exporter.js
```diff
@@ -414,6 +419,14 @@ class CiVisibilityExporter extends BufferingExporter {
       evpProxyPrefix: this.evpProxyPrefix,
     }, callback)
   }
+
+  // Exposed for test access
```
BridgeAR, line 423:
```suggestion
  // Exposed for test access
  // TODO: Rewrite to not need to expose things for testing
```

packages/dd-trace/src/datastreams/writer.js
```diff
@@ -55,10 +57,13 @@ class DataStreamsWriter {
     })
   }
 
+  // Exposed for test access
```
BridgeAR, line 60:
```suggestion
  // Exposed for test access
  // TODO: Rewrite to not need to expose things for testing. E.g., by using proxyquire
```

packages/dd-trace/src/encode/0.4.js
```diff
@@ -331,11 +338,15 @@ class AgentEncoder {
...
 
   _writeTraces (buffer, offset = 0) {
-    offset = this._writeArrayPrefix(buffer, offset, this._traceCount)
+    offset = this._writeArrayPrefix(buffer, offset, this.#traceCount)
     offset += this._traceBytes.buffer.copy(buffer, offset, 0, this._traceBytes.length)
 
     return offset
   }
+
+  // Exposed for test access
```
BridgeAR, line 347:
```suggestion
  // Exposed for test access
  // TODO: Rewrite to not need to expose things for testing. E.g., by using proxyquire
```

packages/dd-trace/src/exporters/agent/index.js
```diff
@@ -30,26 +33,36 @@ class AgentExporter {
...
 
+  // Exposed for external access by opentracing/tracer.js and opentelemetry/tracer_provider.js
+  get _url () {
+    return this.#url
+  }
+
+  // Exposed for external access by opentelemetry/tracer_provider.js and cypress plugin
+  get _writer () {
+    return this.#writer
+  }
```
BridgeAR: I am on the edge about doing that. I would probably just skip things that are indeed used throughout the code.

packages/dd-trace/test/startup-log.spec.js
```diff
@@ -109,9 +109,9 @@ describe('startup logging', () => {
       debug: true,
       sample_rate: 1,
       sampling_rules: [
-        { matchers: [{ pattern: 'rule1' }], _sampler: { _rate: 0.4 } },
+        { matchers: [{ pattern: 'rule1' }] },
         'rule2',
-        { matchers: [{ pattern: 'rule3' }], _sampler: { _rate: 1 } },
```
BridgeAR, line 114: I believe we actually would like to expose this information. I guess we should evaluate closely, if we should expose the property as a regular one or not.

packages/dd-trace/test/sampling_rule.spec.js
```diff
@@ -391,7 +391,6 @@ describe('sampling rule', () => {
         maxPerSecond: 123,
       })
 
-      assert.notStrictEqual(rule._limiter, undefined)
```
BridgeAR, line 394: The test seems to loose some value with this assertion not being possible anymore. I guess we could use a spy instead to determine things better?

packages/dd-trace/src/tracer.js
```diff
@@ -46,15 +49,19 @@ class DatadogTracer extends Tracer {
...
   }
 
   setOffset (offsetData) {
-    return this._dataStreamsProcessor.setOffset(offsetData)
+    return this.#dataStreamsProcessor.setOffset(offsetData)
+  }
+
+  get _dataStreamsProcessor () {
+    return this.#dataStreamsProcessor
   }
```
BridgeAR, line 65: This has no comment why it is exposed. I would actually not touch anything that needs a getter that is not exposed for testing only.

packages/dd-trace/src/span_processor.js
```diff
@@ -166,6 +174,10 @@ class SpanProcessor {
     trace.started = active
     trace.finished = []
   }
+
+  get _processTags () {
+    return this.#processTags
+  }
```
BridgeAR, line 180: This has no comment why it is exposed. I would actually not touch anything that needs a getter that is not exposed for testing only.

packages/dd-trace/src/rate_limiter.js
```diff
@@ -3,17 +3,28 @@
...
-    this._prevTokensRequested = 0
+    this.#limiter = new limiter.RateLimiter(this.#rateLimit, interval)
+  }
+
+  /**
+   * Maximum allowed units per interval.
+   * @returns {number}
+   */
+  get rateLimit () {
+    return this.#rateLimit
```
BridgeAR, line 28: A comment where it is used / why would be good :)

Otherwise exposing `internals` always seems meh.

packages/dd-trace/src/proxy.js
```diff
@@ -346,6 +354,20 @@ class Tracer extends NoopProxy {
   get TracerProvider () {
     return require('./opentelemetry/tracer_provider')
   }
+
+  // Exposed for external access by test infrastructure and internal modules (startup-log, telemetry, llmobs)
+  get _pluginManager () {
+    return this.#pluginManager
+  }
+
+  // Exposed for test access (e.g. to force re-initialization with new env vars)
```
BridgeAR, line 362:
```suggestion
  // Exposed for test access (e.g. to force re-initialization with new env vars)
  // TODO: Rewrite the code to use e.g., proxyquire or sinon instead of exposing information
```

packages/dd-trace/src/proxy.js
```diff
@@ -346,6 +354,20 @@ class Tracer extends NoopProxy {
   get TracerProvider () {
     return require('./opentelemetry/tracer_provider')
   }
+
+  // Exposed for external access by test infrastructure and internal modules (startup-log, telemetry, llmobs)
+  get _pluginManager () {
+    return this.#pluginManager
+  }
```
BridgeAR: I guess the property should just be marked public in that case?

@rochdev what do you think?

packages/dd-trace/src/plugin_manager.js
```diff
@@ -120,21 +126,26 @@ module.exports = class PluginManager {
...
 
     for (const name in pluginClasses) {
       this.loadPlugin(name)
     }
   }
 
+  // Exposed for external access by startup-log, telemetry, llmobs, and test infrastructure
+  get _pluginsByName () {
+    return this.#pluginsByName
+  }
```
BridgeAR: Same as below: this should likely just be a regular property?

packages/dd-trace/src/dogstatsd.js
```diff
@@ -23,24 +23,36 @@ const TYPE_HISTOGRAM = 'h'
...
-    this._udp4 = this._socket('udp4')
-    this._udp6 = this._socket('udp6')
+    this.#host = options.host || defaults['dogstatsd.hostname']
+    this.#family = isIP(this.#host)
+    this.#port = options.port || defaults['dogstatsd.port']
+    this.#prefix = options.prefix || ''
+    this.#tags = options.tags || []
+    this.#queue = []
+    this.#buffer = ''
+    this.#offset = 0
```
BridgeAR: At least these three can definitely be moved to being defaults during assignment above. I think we should check __all__ our new private properties if default assignment is possible or not. Otherwise the new code feels a bit incomplete to me 😅

packages/dd-trace/src/profiling/exporters/agent.js
```diff
@@ -87,16 +87,25 @@ function computeRetries (uploadTimeout) {
...
 
-    this._backoffTime = backoffTime
-    this._backoffTries = backoffTries
+    this.#backoffTime = backoffTime
+    this.#backoffTries = backoffTries
+  }
+
+  get _url () {
+    return this.#url
   }
```
BridgeAR, line 111: Could be add why this is still exposed? Should this just be made public? Is it exposed for testing only?

packages/dd-trace/src/plugins/plugin.js
```diff
@@ -160,6 +182,21 @@ module.exports = class Plugin {
...
+  get _enabled () {
+    return this.#enabled
+  }
+
+  // Exposed for test access
+  get _subscriptions () {
+    return this.#subscriptions
+  }
+
+  // Exposed for test access
```
BridgeAR, line 198:
```suggestion
  // Exposed for test access
  // TODO: Refactor to use proxyquire or sinon
```

packages/dd-trace/src/plugins/plugin.js
```diff
@@ -160,6 +182,21 @@ module.exports = class Plugin {
     }
   }
 
+  // Exposed for external access by telemetry and llmobs modules
+  get _enabled () {
+    return this.#enabled
+  }
+
+  // Exposed for test access
```
BridgeAR, line 192:
```suggestion
  // Exposed for test access
  // TODO: Refactor to use proxyquire or sinon
```

packages/dd-trace/src/plugins/plugin.js
```diff
@@ -160,6 +182,21 @@ module.exports = class Plugin {
     }
   }
 
+  // Exposed for external access by telemetry and llmobs modules
+  get _enabled () {
+    return this.#enabled
+  }
```
BridgeAR: Should we just mark this being public? I think for `.enabled`, that makes sense (we could keep a getter to prevent external manipulation, while I have no strong opinion on that).

packages/dd-trace/src/plugins/plugin.js
```diff
@@ -58,26 +74,32 @@ class StoreBinding {
     }
   }
 
+  // Exposed for test access
```
BridgeAR, line 79:
```suggestion
  // Exposed for test access
  // TODO: Refactor to use proxyquire or sinon
```

packages/dd-trace/src/plugins/plugin.js
```diff
@@ -25,31 +25,47 @@ const { storage } = require('../../../datadog-core')
...
       }
     }
   }
 
+  // Exposed for test access
+  get _channel () {
+    return this.#channel
+  }
+
+  // Exposed for test access
```
BridgeAR, line 47:
```suggestion
  // Exposed for test access
  // TODO: Refactor to use proxyquire or sinon
```

packages/dd-trace/src/plugins/plugin.js
```diff
@@ -25,31 +25,47 @@ const { storage } = require('../../../datadog-core')
...
+    this.#channel = dc.channel(event)
+    this.#handler = (message, name) => {
       const store = storage('legacy').getStore()
       if (!store || !store.noop) {
         handler(message, name)
       }
     }
   }
 
+  // Exposed for test access
```
BridgeAR, line 41:
```suggestion
  // Exposed for test access
  // TODO: Refactor to use proxyquire or sinon
```

packages/dd-trace/src/opentelemetry/tracer_provider.js
```diff
@@ -69,12 +74,15 @@ class TracerProvider {
...
-    return this._activeProcessor.forceFlush()
+    return this.#activeProcessor.forceFlush()
   }
 
   shutdown () {
-    return this._activeProcessor.shutdown()
+    return this.#activeProcessor.shutdown()
   }
+
+  // Exposed for test access
```
BridgeAR, line 84:
```suggestion
  // Exposed for test access
  // TODO: Refactor to use proxyquire or sinon
```

packages/dd-trace/src/openfeature/noop.js
```diff
@@ -75,18 +77,21 @@ class NoopFlaggingProvider {
     })
   }
 
+  // Exposed for test access
```
BridgeAR, line 79:
```suggestion
  // Exposed for test access
  // TODO: Refactor to use proxyquire or sinon
```

packages/dd-trace/src/openfeature/noop.js
```diff
@@ -8,12 +8,14 @@ const { NOOP_REASON } = require('./constants/constants')
...
 class NoopFlaggingProvider {
+  #config
+
   /**
    * @param {object} [noopTracer] - Optional noop tracer instance
    */
   constructor (noopTracer) {
     this._tracer = noopTracer
-    this._config = {}
+    this.#config = {}
```
BridgeAR: This should be moved to the private property definition. The same applies to other places that I have not commented on :)

packages/dd-trace/src/openfeature/writers/exposures.js
```diff
@@ -60,21 +64,24 @@ class ExposuresWriter extends BaseFFEWriter {
...
     })
-    this._enabled = false // Start disabled until agent strategy is set
-    this._pendingEvents = [] // Buffer events until enabled
-    this._context = this._buildContext()
+    this.#enabled = false // Start disabled until agent strategy is set
+    this.#pendingEvents = [] // Buffer events until enabled
+    this.#context = this._buildContext()
   }
 
+  // Exposed for test access
```
BridgeAR, line 72:
```suggestion
  // Exposed for test access
  // TODO: Refactor to use proxyquire or sinon
```

packages/dd-trace/src/openfeature/writers/base.js
```diff
@@ -142,17 +156,44 @@ class BaseFFEWriter {
...
 
-      if (this._droppedEvents > 0) {
-        log.warn('%s dropped %d events due to buffer overflow', this.constructor.name, this._droppedEvents)
+      if (this.#droppedEvents > 0) {
+        log.warn('%s dropped %d events due to buffer overflow', this.constructor.name, this.#droppedEvents)
       }
     }
   }
 
+  // Exposed for test access
```
BridgeAR, line 171:
```suggestion

  // TODO: Refactor all of these getters to use proxyquire or sinon
  // Exposed for test access
```

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -28,19 +28,32 @@ class LLMObs extends NoopLLMObs {
...
-    return this._config.llmobs.enabled
+    return this.#config.llmobs.enabled
   }
 
+  // Exposed for test access
+  get _tracer () { return this.#tracer }
+  get _tagger () { return this.#tagger }
+  get _config () { return this.#config }
+
+  // Exposed for test access
```
BridgeAR, line 55:
```suggestion
  // Exposed for test access
  // TODO: Refactor to use proxyquire or sinon
  get _tracer () { return this.#tracer }
  get _tagger () { return this.#tagger }
  get _config () { return this.#config }
```

packages/dd-trace/src/llmobs/plugins/base.js
```diff
@@ -8,10 +8,16 @@ const TracingPlugin = require('../../plugins/tracing')
...
+
   constructor (...args) {
     super(...args)
 
-    this._tagger = new LLMObsTagger(this._tracerConfig, true)
+    this.#tagger = new LLMObsTagger(this._tracerConfig, true)
+  }
+
+  get _tagger () {
+    return this.#tagger
```
BridgeAR, line 21: Is the exposure needed? Is it used for testing only?

BridgeAR (commented):
I just did a partial review. A lot of the new getters should later on be removed, so adding TODO entries there would be nice.

What we should always do is to define the default value directly at definition time, when possible instead of at construction time.

A few properties should likely just regularly be exposed instead of being a getter now. I think we should review these very closely one by one. So refactoring the unclear ones seems like it might be too early.

packages/dd-trace/src/llmobs/plugins/base.js
bm1549: Yes, the exposure is needed — `_tagger` is used extensively in production code by subclasses (langchain, openai, anthropic, vertexai, bedrockruntime, genai, ai plugins). It's not test-only. Added a comment to make that clear.

packages/dd-trace/src/encode/0.4.js
```diff
@@ -24,32 +24,39 @@ function formatSpan (span, config) {
...
+  #limit
+  #writer
+  #debugEncoding
+  #config
+  #traceCount
+
   constructor (writer, limit = SOFT_LIMIT) {
-    this._msgpack = new MsgpackEncoder()
-    this._limit = limit
+    this.#msgpack = new MsgpackEncoder()
```
BridgeAR: Please assign at definition time

packages/dd-trace/src/encode/agentless-ci-visibility.js
```diff
@@ -41,7 +45,7 @@ class AgentlessCiVisibilityEncoder extends AgentEncoder {
 
     // Used to keep track of the number of encoded events to update the
     // length of `payload.events` when calling `makePayload`
-    this._eventCount = 0
+    this.#eventCount = 0
```
BridgeAR: Please assign at definition time

BridgeAR:
This has a ton of conflicts and I think we can get back to this in a few month again porting more things to private properties. As soon as we have ported enough, a lint rule can be added. That will just likely still take a year or so.

PR #7706, opened 2026-03-06
fix(ci): address flaky test infrastructure issues

bm1549:
## Summary

- **Debugger / busboy**: Added a readiness health-check after `docker compose up -d testagent`. Previously, tests could start sending multipart form data to the test agent container before it was ready to accept connections, causing busboy to receive truncated requests and throw `Error: Unexpected end of form`.

- **Azure Service Bus / queue not found**: The `azure-service-bus` CI job was missing the config-copy + emulator-restart steps that the `azure-functions` job already has. GitHub Actions services don't support volume mounts, so the servicebus emulator was starting with its default (empty) config — no `queue.1` or `topic.1` existed. Added `docker cp` to inject `servicebus-emulator-config.json` and a `docker restart` to apply it.

> **Note:** The AI Guard / macOS \"Timeout in beforeEach\" was also investigated. The root cause is unclear (likely slow proxyquire-based module loading on macOS runners, not a fixable infrastructure issue). That failure is tracked separately.

### testagent health check
- **Blast radius is wide**: this action is used by `lambda`, `aws-sdk`, and everything routed through `plugins/test`. If the `/info` endpoint doesn't exist or returns a non-2xx while booting, `curl -sf` will keep retrying until the 30s timeout and then fail — turning a flake into a hard breakage across many jobs.
- **Mitigation**: the `/info` endpoint is a standard diagnostic endpoint on the DD APM test agent; this pattern is well-established. 30s timeout should be sufficient on normal runners, though a cold-pull on a slow runner could hit it.
- **Confidence**: high that it fixes the race condition; moderate confidence in the specific endpoint choice.

### Azure Service Bus config copy + restart
- **`docker cp` timing**: GitHub Actions marks services "running" without a healthcheck, so the emulator may not have created `/ServiceBus_Emulator/ConfigFiles/` yet when `docker cp` runs. A failure here would be a hard error rather than a flake.
- **No explicit readiness wait after `docker restart`**: relies on the implicit time taken by `testagent/start` + `node` setup + `install` inside `plugins/test`. The `azure-functions` job uses the same pattern and works in practice, which is the main evidence this is safe.
- **Confidence**: high that the missing config was the root cause of \"Queue not found\"; moderate confidence in the timing assumptions.

## Test plan

- [x] Debugger tests no longer show `Error: Unexpected end of form from busboy`
- [x] `azure-service-bus` job no longer fails with \"Queue not found\"
- [x] No regressions in other jobs that use the testagent start action

### DIFF
--- a/.github/workflows/serverless.yml
+++ b/.github/workflows/serverless.yml
@@ -258,13 +258,6 @@ jobs:
       SERVICES: azureservicebusemulator,azuresqledge
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-      - name: Copy emulator config file
-        run: |
-          docker cp \
-          ${{ github.workspace }}/packages/datadog-plugin-azure-functions/test/fixtures/servicebus-emulator-config.json \
-          ${{ job.services.azureservicebusemulator.id }}:/ServiceBus_Emulator/ConfigFiles/Config.json
-      - name: Restart emulator to pick up config changes
-        run: docker restart ${{ job.services.azureservicebusemulator.id }}
       - name: Wait for Service Bus emulator to be ready
         run: timeout 120 bash -c 'until nc -z localhost 5672; do sleep 3; done'
       - uses: ./.github/actions/plugins/test


.github/workflows/serverless.yml
```diff
@@ -244,11 +247,26 @@ jobs:
...
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Copy emulator config file
+        run: |
+          docker cp \
+          ${{ github.workspace }}/packages/datadog-plugin-azure-functions/test/fixtures/servicebus-emulator-config.json \
+          ${{ job.services.azureservicebusemulator.id }}:/ServiceBus_Emulator/ConfigFiles/Config.json
+      - name: Restart emulator to pick up config changes
+        run: docker restart ${{ job.services.azureservicebusemulator.id }}
+      - name: Wait for Service Bus emulator to be ready
+        run: timeout 120 bash -c 'until nc -z localhost 5672; do sleep 3; done'
```
jcstorms1: Although it won't break anything, we don't need to copy the config file for this action. The Service bus tests use the generic `queue.1` and `topic.1` that is built into the OOB emulator config. Azure functions use additional queues and topics, which is why it has a config file in the setup.

I'd suggest removing it to avoid any additional issues, but I'll leave the decision up to you.

jcstorms1 (dismissed):
Just one comment for Service Bus but LGTM otherwise.

PR #7710, opened 2026-03-06
fix(http): prevent Node.js 24 stream crash in HTTP response instrumentation

bm1549:
### What does this PR do?

Wraps the `originalRead.apply()` call inside the HTTP response `read()` wrapper with a scoped try-catch to handle a Node.js 24 crash in `fromList` during stream mode transitions.

When a consumer adds a `data` listener while a `readable`+`read()` loop is active, Node.js 24 transitions the stream from paused to flowing mode. Calling `read()` synchronously during this transition crashes inside `fromList` with `TypeError: Cannot read properties of undefined (reading 'length')`. The catch block:

- Only catches `TypeError` with message matching `"Cannot read prop"` — all other errors are re-thrown
- Logs a warning via `log.warn()` for diagnosability with `DD_TRACE_DEBUG`
- Restores `originalRead` and returns `null` to safely exit the caller's read loop

On Node.js 18–23, the try-catch never fires and behavior is identical to master.

### Motivation

Node.js 24 changed the internals of `fromList` in `node:internal/streams/readable`. Our `setupResponseInstrumentation` wraps `res.read` to collect body chunks for AppSec response body inspection. When a consumer mixes `readable`+`read()` with a `data` listener (an unusual but valid pattern), the wrapped `read()` calls `originalRead.apply()` during the stream mode transition, triggering the crash.

### Additional Notes

- The existing `dataReadStarted` / `dataListenerAdded` guards from master are preserved unchanged — only the try-catch around `originalRead.apply()` is new
- Added a test that mocks `originalRead` to throw the expected `TypeError`, verifying the wrapper recovers and the response completes without crashing (HTTPS only — HTTP's 301 redirect response is too small to exercise the multi-chunk crash scenario)
- The crash only manifests when there is buffered data in the stream's internal buffer at the moment of the mode transition, which is why it's intermittent in practice (depends on network timing and response size)

**File changed:** `packages/datadog-instrumentations/src/http/client.js`, `packages/datadog-instrumentations/test/http.spec.js`


BridgeAR:
I think we should not do this for now, while I think it would be great to discuss this again with product side about such issues.

PR #7717, opened 2026-03-09
feat(process tags): add tags to runtime metrics

tlhunter:
### What does this PR do?
- adds process tags to runtime metrics
- ~adds process tags to process discovery~
- ~enables process tags by default~

### Motivation
- process tags are key/value pairs about the current application useful for renaming processes
- includes things like cwd, script name, etc.

### Additional Notes
- search the commit history for "process tags" to see previous commits

### DIFF
--- a/packages/dd-trace/src/config/supported-configurations.json
+++ b/packages/dd-trace/src/config/supported-configurations.json
@@ -744,7 +744,7 @@
     ],
     "DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED": [
       {
-        "implementation": "B",
+        "implementation": "A",
         "type": "boolean",
         "default": "false",
         "configurationNames": [


packages/dd-trace/src/runtime_metrics/runtime_metrics.js
```diff
@@ -38,6 +38,13 @@ module.exports = {
     this.stop()
     const clientConfig = DogStatsDClient.generateClientConfig(config)
 
+    if (config.propagateProcessTags?.enabled) {
+      const processTags = require('../process-tags')
```
tlhunter: This should be hoisted to the top of the file

packages/dd-trace/src/tracer_metadata.js
```diff
@@ -11,13 +12,18 @@ function storeConfig (config) {
       return
     }
 
+    const processTags = require('./process-tags')
```
tlhunter: This should be hoisted to the top of the file

packages/dd-trace/test/tracer_metadata.spec.js
```diff
@@ -0,0 +1,128 @@
...
+  let maybeLoad
+
+  function makeStoreConfig (options = {}) {
+    TracerMetadata = function (...args) { this.args = args }
+    storeMetadata = function (metadata) { return metadata }
+    maybeLoad = function () { return { TracerMetadata, storeMetadata } }
+
+    return proxyquire('../src/tracer_metadata', {
+      '@datadog/libdatadog': { maybeLoad },
+      './exporters/common/docker': { containerId: options.containerId ?? undefined },
```
CarlesDD: Nit: `options.containerId ?? undefined` is redundant here, there is no case where `options.containerId = null` in the test suite

CarlesDD (dismissed):
LGTM from a language perspective

wantsui (approved):
LGTM in terms of functionality.

PR #7732, opened 2026-03-10
use dd-octo-sts for tag creation

bengl:


.github/chainguard/release.sts.yaml
```diff
@@ -0,0 +1,13 @@
+issuer: https://token.actions.githubusercontent.com
+
+subject_pattern: "repo:DataDog/dd-trace-js:ref:refs/heads/(v[345]\\.x|master)"
+
+claim_pattern:
+  event_name: (push|workflow_dispatch)
+  ref: refs/heads/(v[345]\.x|master)
+  ref_protected: "true"
```
xopham: this has turned out to be an unreliable and misleading field, so let's drop it. The plugin is technically instructed to not add it, so i am taking note

.github/chainguard/self.github.release.push-tags.sts.yaml
```diff
@@ -0,0 +1,13 @@
+issuer: https://token.actions.githubusercontent.com
```
xopham, line 1: could we name it `self.github.release.push-tags.sts.yaml` to comply with naming scheme?

.github/workflows/release.yml
```diff
@@ -26,6 +26,11 @@ jobs:
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
+      - uses: DataDog/dd-octo-sts-action@acaa02eee7e3bb0839e4272dacb37b8f3b58ba80 # v1.0.3
+        id: octo-sts
+        with:
+          scope: DataDog/dd-trace-js
+          policy: release
```
xopham: if adjusted above:
```suggestion
          policy: self.github.release.push-tags
```

.github/workflows/release.yml
```diff
@@ -26,6 +26,11 @@ jobs:
...
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
xopham, line 32: better than injecting below, you could do:

```suggestion
      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
        with:
          token: ${{ steps.octo-sts.outputs.token }}
```

.github/workflows/release.yml
```diff
@@ -35,7 +40,7 @@ jobs:
           echo "json=$content" >> $GITHUB_OUTPUT
       - run: |
           git tag v${{ fromJson(steps.pkg.outputs.json).version }}
-          git push origin v${{ fromJson(steps.pkg.outputs.json).version }}
+          git push https://x-access-token:${{ steps.octo-sts.outputs.token }}@github.com/${{ github.repository }}.git v${{ fromJson(steps.pkg.outputs.json).version }}
```
xopham, line 41: if adjusting the previous comment, you can go back to:

```suggestion
          git push origin v${{ fromJson(steps.pkg.outputs.json).version }}
```

.github/workflows/release.yml
xopham: I am not sure why there ever was a `pull-requests: read`, but I assume this and the `contents: write` can be removed:

```suggestion
```
Unless the npm publish action needs it in which case only drop the `contents: write`

xopham (dismissed):
I made a few comments to improve this on the `publish-v3` job that should if accepted be used for all others as well. Made a brief [self-test workflow](https://github.com/DataDog/sdlc-security-playground/blob/main/.github/workflows/release.yml) and [run](https://github.com/DataDog/sdlc-security-playground/actions/runs/22971830997/job/66690586810) for reference :)

Thanks for fixing!

.github/workflows/release.yml
xopham: Actually, let's not do this. It over-privileges other steps and we should not do that

.github/workflows/release.yml
xopham: see previous comment...this is not a good idea. Sorry for the confusion

xopham (approved):
lgtm. thank you for the patience and sorry for the detour

.github/workflows/release.yml
```diff
@@ -71,11 +77,14 @@ jobs:
...
-      pull-requests: read
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
+      - uses: DataDog/dd-octo-sts-action@acaa02eee7e3bb0839e4272dacb37b8f3b58ba80 # v1.0.3
+        id: octo-sts
+        with:
+          scope: DataDog/dd-trace-js
+          policy: self.github.release.push-tags
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
xopham, line 88: the below push will fail if you do not drop the creds here:

```suggestion
      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
```
you will have to check though wether any of the below commands requires local github access. I have tested the different options and problems [here](https://github.com/DataDog/sdlc-security-playground/actions/runs/23645110738/workflow) for reference

xopham (commented):
one small change needed

BridgeAR:
I believe this got resolved by https://github.com/DataDog/dd-trace-js/pull/7959. Closing

PR #7764, opened 2026-03-12
feat: tag trace and current span with feature flag metadata

petzel:
### What does this PR do?
Attaches feature flag evaluation metadata to APM spans, enabling correlation between flag evaluations and traces.

When a flag is evaluated via the OpenFeature provider, we add tags on both the active span and trace with flag metadata (`feature_flags.[flag key]: [variant key]`). 

Falls back to `String(result.value)` as the variant when the provider returns no explicit variant (e.g., flag not found, default returned). 

Also enforces a configurable per-span limit on the number of flag tags to prevent bandwidth issues in traces with many evaluations

### Motivation

Users of Datadog Feature Flagging + APM want to be able to filter APM traces by feature flag variant. These changes aim to enable that use case. 

### Additional Notes
Added config option:
  - `DD_EXPERIMENTAL_FLAGGING_PROVIDER_MAX_FLAG_TAGS` (default: 300) — maximum number of unique flag tags per span. Once reached, further evaluations still tag the current span, but skip tagging the trace.
    - The default maximum for feature flags added to RUM/Product Analytics events is 300, but that is across an entire user session. I chose to use 300 here to align with RUM.

### Testing
* Installed the tracer on a local NodeJS application
* Created an endpoint which evaluated N (as input) number of feature flags
* Set `DD_EXPERIMENTAL_FLAGGING_PROVIDER_MAX_FLAG_TAGS=3`
* Invoked the endpoint with N=10
* Ensured span tags were added to the current span
* Ensured 3 trace tags were added

image

* Set `DD_EXPERIMENTAL_FLAGGING_PROVIDER_MAX_FLAG_TAGS=50` 
* Invoked the endpoint with N=50
* Ensured span tags were added to the current span
* Ensured 50 trace tags were added

image

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -802,6 +802,16 @@ declare namespace tracer {
          * Programmatic configuration takes precedence over the environment variables listed above.
          */
         initializationTimeoutMs?: number
+        /**
+         * Maximum number of feature flag tags to attach to a single trace.
+         * Limits bandwidth when many flags are evaluated in one trace. Capped at 1000.
+         * Can be configured via DD_EXPERIMENTAL_FLAGGING_PROVIDER_MAX_FLAG_TAGS environment variable.
+         *
+         * @default 300
+         * @env DD_EXPERIMENTAL_FLAGGING_PROVIDER_MAX_FLAG_TAGS
+         * Programmatic configuration takes precedence over the environment variables listed above.
+         */
+        maxFlagTags?: number
       }
     };
 
--- a/packages/dd-trace/src/config/index.js
+++ b/packages/dd-trace/src/config/index.js
@@ -415,6 +415,7 @@ class Config {
       OTEL_TRACES_SAMPLER_ARG,
       DD_EXPERIMENTAL_FLAGGING_PROVIDER_ENABLED,
       DD_EXPERIMENTAL_FLAGGING_PROVIDER_INITIALIZATION_TIMEOUT_MS,
+      DD_EXPERIMENTAL_FLAGGING_PROVIDER_MAX_FLAG_TAGS,
       OTEL_EXPORTER_OTLP_LOGS_ENDPOINT,
       OTEL_EXPORTER_OTLP_LOGS_HEADERS,
       OTEL_EXPORTER_OTLP_LOGS_PROTOCOL,
@@ -602,6 +603,10 @@ class Config {
     if (DD_EXPERIMENTAL_FLAGGING_PROVIDER_INITIALIZATION_TIMEOUT_MS != null) {
       target['experimental.flaggingProvider.initializationTimeoutMs'] =
         maybeInt(DD_EXPERIMENTAL_FLAGGING_PROVIDER_INITIALIZATION_TIMEOUT_MS)
+    }
+    if (DD_EXPERIMENTAL_FLAGGING_PROVIDER_MAX_FLAG_TAGS != null) {
+      target['experimental.flaggingProvider.maxFlagTags'] =
+        Math.min(maybeInt(DD_EXPERIMENTAL_FLAGGING_PROVIDER_MAX_FLAG_TAGS), 1000)
     }
     setBoolean(target, 'traceEnabled', DD_TRACE_ENABLED)
     setBoolean(target, 'experimental.aiguard.enabled', DD_AI_GUARD_ENABLED)
--- a/packages/dd-trace/src/config/supported-configurations.json
+++ b/packages/dd-trace/src/config/supported-configurations.json
@@ -742,6 +742,16 @@
         "default": "false"
       }
     ],
+    "DD_EXPERIMENTAL_FLAGGING_PROVIDER_MAX_FLAG_TAGS": [
+      {
+        "implementation": "A",
+        "type": "int",
+        "configurationNames": [
+          "experimental.flaggingProvider.maxFlagTags"
+        ],
+        "default": "300"
+      }
+    ],
     "DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED": [
       {
         "implementation": "A",
--- a/packages/dd-trace/src/openfeature/span_tagger.js
+++ b/packages/dd-trace/src/openfeature/span_tagger.js
@@ -1,0 +1,65 @@
+'use strict'
+
+const { FF_TAG_PREFIX } = require('./constants/constants')
+
+/**
+ * Builds a feature flag tag mapping flag key to variant.
+ *
+ * @param {string} flagKey - The feature flag key
+ * @param {string} [variantKey] - The variant key from the evaluation result
+ * @returns {{[key: string]: string}} Tag key-value pair
+ */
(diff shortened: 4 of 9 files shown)


packages/dd-trace/src/config/index.js
```diff
@@ -603,6 +604,10 @@ class Config {
       target['experimental.flaggingProvider.initializationTimeoutMs'] =
         maybeInt(DD_EXPERIMENTAL_FLAGGING_PROVIDER_INITIALIZATION_TIMEOUT_MS)
     }
+    if (DD_EXPERIMENTAL_FLAGGING_PROVIDER_MAX_FLAG_TAGS != null) {
```
bantonsson, line 607: Not sure how things are normally done in the JS tracer, but relying on `undefined` feels weird.

packages/dd-trace/src/config/supported-configurations.json
```diff
@@ -742,6 +742,16 @@
...
       }
     ],
+    "DD_EXPERIMENTAL_FLAGGING_PROVIDER_MAX_FLAG_TAGS": [
+      {
+        "implementation": "A",
+        "type": "int",
+        "configurationNames": [
+          "experimental.flaggingProvider.maxFlagTags"
+        ],
+        "default": "300"
```
bantonsson, line 752: There is never a default of `300`, but a cap of `1000`.

packages/dd-trace/src/openfeature/span_tagger.js
```diff
@@ -0,0 +1,65 @@
...
+}
+
+/**
+ * Counts the number of feature flag tags in the trace-level tags.
+ *
+ * @param {import('../opentracing/span')} span
+ * @returns {number}
+ */
+function countFlagTags (span) {
+  const traceTags = span.context()._trace.tags
```
bantonsson, line 24: Can't you have the counter stored on the `_trace`? It feels like this could become costly if you have a fairly large cap but way more evaluations.

packages/dd-trace/src/openfeature/span_tagger.js
```diff
@@ -0,0 +1,65 @@
...
+function tagSpansForEvaluation (tracer, { flagKey, variantKey, maxFlagTags }) {
+  const activeSpan = tracer.scope().active()
+  if (!activeSpan) return
+
+  if (maxFlagTags !== undefined && countFlagTags(activeSpan) >= maxFlagTags) return
+
+  const tags = buildFeatureFlagTags(flagKey, variantKey)
+
+  activeSpan.addTags(tags)
+  Object.assign(activeSpan.context()._trace.tags, tags)
```
bantonsson, line 58: So why are these added to both the active span and the _trace segment_? It will double the data 🙁 Also, what if the same `flagKey` is evaluated multiple times in the same span or _trace segment_, won't the keys overwrite each other?

packages/dd-trace/src/openfeature/span_tagger.js
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @param {import('../tracer')} tracer - Datadog tracer instance
+ * @param {object} params
+ * @param {string} params.flagKey - The feature flag key
+ * @param {string} [params.variantKey] - The variant key from the evaluation result
+ * @param {number} [params.maxFlagTags] - Maximum number of flag tags per trace
+ */
+function tagSpansForEvaluation (tracer, { flagKey, variantKey, maxFlagTags }) {
+  const activeSpan = tracer.scope().active()
+  if (!activeSpan) return
```
bantonsson, line 51: I think that if the active span or _trace segment_ is marked as drop, the we shouldn't add evaluation tags to it. They will never reach the backend then (except for some error cases I guess).

index.d.ts
```diff
@@ -802,6 +802,16 @@ declare namespace tracer {
          * Programmatic configuration takes precedence over the environment variables listed above.
          */
         initializationTimeoutMs?: number
+        /**
+         * Maximum number of feature flag tags to attach to a single trace.
+         * Limits bandwidth when many flags are evaluated in one trace. Capped at 1000.
+         * Can be configured via DD_EXPERIMENTAL_FLAGGING_PROVIDER_MAX_FLAG_TAGS environment variable.
+         *
+         * @default 300
```
bantonsson, line 810: There is no default, since the default is to not tags.

bantonsson (commented):
I would like to know if there are any numbers on how many evaluations are common for end users, and how much data one would expect on them. Having a large number of tags can have a huge performance impact, not only on the tracer itself but also on the Datadog Agent and the amount of data being de/serialized at all stages.

The other thing is adding the tags to the _trace segment_ as well as the span itself. What is the rationale for this?

What data is already being sent to the backend for the evaluations, and is there some way to combine that with the spans to get the search/filter capability without adding all this data to the spans.

packages/dd-trace/src/config/index.js
petzel: This is following the existing pattern of configuration for this tracer: `packages/dd-trace/src/config/supported-configurations.json` contains defaults as well as the env var which can override its value. Only presence of the value will override the configuration value.

packages/dd-trace/src/openfeature/span_tagger.js
petzel: With a counter, I'm not sure how we'd be able to handle cases where a flag is evaluated multiple times in the same trace. 

The counter would work in cases where flag1, flag2, ..., flagN are evaluated, with no duplicate flags. However if a trace contains evaluations for flag1, flag1, ..., flag1 that should still be a single span tag that is added, and a counter would not be accurate.

packages/dd-trace/src/openfeature/span_tagger.js
petzel: I didn't know about drop - agreed, will update.

packages/dd-trace/src/openfeature/span_tagger.js
petzel: This was a [suggestion](https://dd.slack.com/archives/C08C0CC8BNG/p1773080706408389?thread_ts=1772828446.594299&cid=C08C0CC8BNG) from @djova. I'd be open to leaving tags out of the current span for now, as being able to filter on the trace level is the most important. 

In the situation you described, the last evaluation within a span would win. I think that is acceptable.

packages/dd-trace/src/config/supported-configurations.json
petzel: I believe the value here is the actual default, not just documentation. It is applied here: https://github.com/DataDog/dd-trace-js/blob/c53d4194633f895a340e2685e2ade6920030d056/packages/dd-trace/src/config/defaults.js#L94

index.d.ts
petzel: Responded in the other comment - I do believe this default is applied.

BridgeAR:
@petzel is this still worked upon? :)

petzel:
> @petzel is this still worked upon? :)

Nope, closing!

PR #7778, opened 2026-03-13
WORK-IN-PROGRESS feat(lambda): migrate datadog-lambda-js into dd-trace-js as a plugin

joeyzhao2018:
Consolidates all lambda tracing functionality from the standalone datadog-lambda-js repo into dd-trace-js as a proper plugin, eliminating the double-wrapping problem and using AWS_LAMBDA_EXEC_WRAPPER for clean auto-instrumentation.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


BridgeAR:
@joeyzhao2018 I guess if we want to do this, we likely have to do it in a concentrated effort, otherwise there are likely conflicts pilling up fast 😅

PR #7783, opened 2026-03-16
feat(aap): In App WAF support for lambda

CarlesDD:
### What does this PR do?
Adds AppSec support for AWS lambda to dd-trace-js by introducing DC handlers that allow the `datadog-lambda-js` layer to delegate WAF execution to the tracer.

inappwafport

Changes:
- New `datadog:lambda:start-invocation` and `datadog:lambda:end-invocation` diagnostic channels with its subscribers (`appsec/lambdja.js`), which maps extracted HTTP data to WAF addresses, run the WAF and reports the results.
- Reporter adaptation: `reportMetrics`, `reportAttack`, `reportAttributes`, and `finishRequest` now accept an optional `rootSpan` parameter, falling back to `web.root(req)` when not provided. This allows the Lambda handler to pass the span directly since there is no `req` object in Lambda
- WAF `rootSpan` threading: `waf/index.js` and `waf_context_wrapper.js` thread `rootSpan` through the call chain so that `keepTrace` and reporter calls work correctly for Lambda

### Motivation
Porting the In-App WAF security product to AWS Lambda for the Node.js runtime. The Lambda layer extracts HTTP data from the event and dispatches it to the tracer's AppSec domain via DC.

This approach keeps all security logic (WAF execution, attack reporting, trace keep decisions) inside dd-trace-js, and the Lambda layer is only responsible for extracting raw HTTP data and publishing it.

### Additional Notes
- The Lambda channels are always subscribed when AppSec is enabled. They are no-ops unless `datadog-lambda-js` publishes to them.
- No new configuration is needed;`DD_APPSEC_ENABLED` env var controls everything.
- Serverless defaults in dd-trace-js already neutralize telemetry, remote config, and error logging, so there are no side effects to enabling AppSec in Lambda beyond the WAF initialization cold start cost.
- This is a monitoring-only first iteration. Blocking, Remote Config, and telemetry are out of scope.
- Companion PR in datadog-lambda-js: https://github.com/DataDog/datadog-lambda-js/pull/744

[APPSEC-60752](https://datadoghq.atlassian.net/browse/APPSEC-60752)

[APPSEC-60752]: https://datadoghq.atlassian.net/browse/APPSEC-60752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dd-trace/src/appsec/reporter.js
```diff
@@ -552,6 +560,8 @@ function finishRequest (req, res, storedResponseHeaders, requestBody) {
 
   incrementWafRequestsMetric(req)
 
+  if (!req) return
```
simon-id:
```suggestion
  if (!req) return

  incrementWafRequestsMetric(req)
```

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -335,12 +337,14 @@ function reportTruncationMetrics (rootSpan, metrics) {
   }
 }
 
-function reportAttack ({ events: attackData, actions }, req) {
+function reportAttack ({ events: attackData, actions }, req, rootSpan) {
   if (!req) {
     req = storage('legacy').getStore()?.req
   }
```
simon-id, line 377: i'm worried that in this function, if we don't pass the empty object fake "req", it will try to get it from the store, which would make this req variable here undefined, and crash later down the function because it's trying to access props in req (req.socket, req.body, etc).

The safety mecanism that was in place before was the rootSpan check. If req was missing, then rootSpan would also be missing, and thus the `if(!rootSpan) return` would protect us from undefined req.
But now that we can pass the rootSpan separately from req, this safety check is inoperant. Does that make sense ?

Can you prove to me that req can never be empty here ?

packages/dd-trace/src/appsec/reporter.js
simon-id: (incrementWafRequestsMetric() is a noop with empty req anyway)

simon-id: also reportRequestBody() can be used without req no ? here you'd be skipping it

simon-id: and for header collection, is there no headers we should collect in lambda ?

packages/dd-trace/src/appsec/lambda.js
```diff
@@ -0,0 +1,129 @@
...
+      persistent[addresses.HTTP_INCOMING_URL] = path
+    }
+
+    if (method) {
+      persistent[addresses.HTTP_INCOMING_METHOD] = method
+    }
+
+    if (headers) {
+      persistent[addresses.HTTP_INCOMING_HEADERS] = headers
+    }
```
simon-id, line 51: are you not removing cookies from this ?

packages/dd-trace/src/appsec/reporter.js
CarlesDD: This is a monitoring only first iteration, so there are deliberate gaps to address in next iterations:

- `incrementWafRequestMetric`: your point is correct
- `reportRequestBody`: Right, it does not need `req`. Lambda path calls `finishRequest` with an unefined `requestBody` so `reportRequestBody` will return early. Extended data collection is something that will be revisited in a future iteration.
- for header collection, it is a known gap taht will be addressed in a future iteration.

packages/dd-trace/src/appsec/reporter.js
CarlesDD: Yep, your concern is legit: decoupling rootSpan from req breaks the implicit safety mechanism. But in the lambda path, req is never `null` or `undefined` in `reportAttack` (it's an invocationKey ({}), a truthy object flowing from lambda.js). The only place where `req` is `null` is in `finishRequest`.

I've added a test to check the WAF path with a strict req object that throws on any unaudited property access, making it fail if someone adds `req.something `without a guard in this path.

packages/dd-trace/src/appsec/lambda.js
CarlesDD: cookies are already stripped upstream in lambda's layer. Added an inline comment to clarify this.

CarlesDD:
> Overall this PR is a bit scary because it invalidates assumptions that were made during the entire evolution of our codebase. We should really be careful about these changes, a strong test suite, and proactive monitoring. For now I feel like the test suite is a bit light no ?

Agree this needs careful handling. I've pushed additional changes to strengthen the safety net:

- More tests for Lambda handler
- A proxy based WAF path safety test: it fails if anyone adds req.something with no guard in the future
- Reporter lambda specific test: covering finishRequest and reportAttack with null/empty req
- JSDocs to make the contract more explicit.

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -301,11 +301,19 @@ function reportWafConfigUpdate (product, rcConfigId, diagnostics, wafVersion) {
...
+function reportMetrics (metrics, raspRule, req, rootSpan) {
   if (!req) {
     req = storage('legacy').getStore()?.req
   }
-  const rootSpan = req && web.root(req)
+  if (!rootSpan) {
+    rootSpan = req && web.root(req)
+  }
 
   if (!rootSpan) return
```
simon-id:
```suggestion
  if (!req || !rootSpan) return
```

simon-id: from our previous thread https://github.com/DataDog/dd-trace-js/pull/7783#discussion_r2952183261

packages/dd-trace/src/appsec/lambda.js
```diff
@@ -0,0 +1,130 @@
...
+    span.setTag('_dd.appsec.enabled', 1)
+
+    if (clientIp) {
+      span.setTag(HTTP_CLIENT_IP, clientIp)
+    }
+
+    const persistent = {}
+
+    if (path) {
+      persistent[addresses.HTTP_INCOMING_URL] = path
```
uurien, line 41: `HTTP_INCOMING_URL` is expected to be an URL not only the path. There is no way to get the whole URL in lambda?

packages/dd-trace/src/appsec/lambda.js
```diff
@@ -0,0 +1,130 @@
...
+  try {
+    const { span, headers, method, path, query, body, clientIp, pathParams, cookies } = data
+
+    if (!span) {
+      log.warn('[ASM] No span provided in Lambda start invocation')
+      return
+    }
+
+    const invocationKey = {}
+    span._lambdaAppsecKey = invocationKey
```
uurien: I don't think we should store extra data in span, we usually create a WeakMap<span,data>.

And I think that you even could use the span itself as "invocationKey"

packages/dd-trace/src/appsec/lambda.js
```diff
@@ -0,0 +1,130 @@
...
+      persistent[addresses.HTTP_INCOMING_RESPONSE_HEADERS] = filteredHeaders
+    }
+
+    if (Object.keys(persistent).length > 0) {
+      waf.run({ persistent }, invocationKey, undefined, span)
+    }
+
+    waf.disposeContext(invocationKey)
+
+    Reporter.finishRequest(null, null, {}, undefined, span)
```
uurien: shouldn't we send `invocationKey` as `req` here to be able to report metrics correctly?

packages/dd-trace/src/appsec/lambda.js
```diff
@@ -0,0 +1,130 @@
...
+      persistent[addresses.HTTP_INCOMING_METHOD] = method
+    }
+
+    if (headers) {
+      // Cookie header is already stripped by the Lambda layer's event-data-extractor
+      persistent[addresses.HTTP_INCOMING_HEADERS] = headers
+    }
+
+    if (clientIp) {
+      persistent[addresses.HTTP_CLIENT_IP] = clientIp
```
IlyasShabi, line 55: add `span.setTag(HTTP_CLIENT_IP, clientIp)` here and remove first block

packages/dd-trace/src/appsec/lambda.js
```diff
@@ -0,0 +1,130 @@
...
+
+    if (pathParams) {
+      persistent[addresses.HTTP_INCOMING_PARAMS] = pathParams
+    }
+
+    if (cookies) {
+      persistent[addresses.HTTP_INCOMING_COOKIES] = cookies
+    }
+
+    waf.run({ persistent }, invocationKey, undefined, span)
```
IlyasShabi: `invocationKey` gonna be always {} no?

packages/dd-trace/src/appsec/lambda.js
IlyasShabi: I agree

packages/dd-trace/src/appsec/lambda.js
```diff
@@ -0,0 +1,130 @@
...
+      persistent[addresses.HTTP_INCOMING_RESPONSE_CODE] = String(statusCode)
+    }
+
+    if (responseHeaders) {
+      const filteredHeaders = { ...responseHeaders }
+      delete filteredHeaders['set-cookie']
+      persistent[addresses.HTTP_INCOMING_RESPONSE_HEADERS] = filteredHeaders
+    }
+
+    if (Object.keys(persistent).length > 0) {
```
IlyasShabi: let's add a flag and make it true when we have something to add to persistent and then avoid `Object.keys(persistent).length > 0`.
We should fix this in all our appsec code to avoid the overhead of creating unnecessary objects

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -353,12 +361,24 @@ function reportTruncationMetrics (rootSpan, metrics) {
...
+function reportAttack ({ events: attackData, actions }, req, rootSpan) {
   if (!req) {
     req = getActiveRequest()
   }
 
-  const rootSpan = web.root(req)
+  if (!rootSpan) {
+    rootSpan = web.root(req)
+  }
   if (!rootSpan) return
```
IlyasShabi:
```suggestion
  if (!req || !rootSpan) return
```

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -380,12 +400,14 @@ function reportAttack ({ events: attackData, actions }, req) {
     ? currentJson.slice(0, -2) + ',' + attackDataStr.slice(1) + '}'
     : '{"triggers":' + attackDataStr + '}'
 
-  if (req.socket) {
+  if (req?.socket) {
```
IlyasShabi: No need to change it if we check !req before
```suggestion
  if (req.socket) {
```

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -380,12 +400,14 @@ function reportAttack ({ events: attackData, actions }, req) {
...
     newTags[NETWORK_CLIENT_IP] = req.socket.remoteAddress
   }
 
   rootSpan.addTags(newTags)
 
+  if (!req) return
```
IlyasShabi: No need for this it if we check !req before

packages/dd-trace/src/appsec/reporter.js
```diff
@@ -494,14 +516,21 @@ function isSchemaAttribute (attribute) {
...
   if (!req) {
     req = getActiveRequest()
   }
 
-  const rootSpan = web.root(req)
+  if (!rootSpan) {
+    rootSpan = web.root(req)
+  }
 
   if (!rootSpan) return
```
IlyasShabi, line 534: Ditto

packages/dd-trace/src/appsec/lambda.js
CarlesDD: `HTTP_INCOMING_URL` maps to `server.request.uri.raw`, which is expected to be relative URI (i.e starting at path component, without scheme nor authority components)

packages/dd-trace/src/appsec/reporter.js
CarlesDD: `reportAttributes` doesn't use `req` after getting `rootSpan`, so the `!req` guard isn't needed here (unlike `reportAttack`, which accesses `req.socket`, `web.getContext(req)`, and `req.body` after the guard.

There is also an [existing test](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/test/appsec/reporter.spec.js#L795L798) that calls `reportAttributes` without a `req` and expects `addTags` to be called, relying on `web.root(undefined)` resolving to a span. Adding `!req` would break that path.

packages/dd-trace/src/appsec/lambda.js
CarlesDD: Replaced `span._lambdaAppsecKey` with a `WeakSet` to track active invocations.

packages/dd-trace/src/appsec/lambda.js
CarlesDD: Fixed.

packages/dd-trace/src/appsec/lambda.js
CarlesDD: Done. The two separate conditionals are now merged into one.

packages/dd-trace/src/appsec/lambda.js
CarlesDD: Addressed

packages/dd-trace/src/appsec/lambda.js
CarlesDD: Fixed with a `hasPersistentData` boolean flag

packages/dd-trace/src/appsec/reporter.js
CarlesDD: Applied

packages/dd-trace/src/appsec/reporter.js
CarlesDD: Applied

packages/dd-trace/src/appsec/reporter.js
CarlesDD: Fixed

uurien (approved):

packages/dd-trace/src/appsec/index.js
```diff
@@ -99,11 +102,15 @@ function enable (_config) {
...
     stripeConstructEvent.subscribe(onStripeConstructEvent)
+    lambdaStartInvocation.subscribe(lambda.onLambdaStartInvocation)
+    lambdaEndInvocation.subscribe(lambda.onLambdaEndInvocation)
 
     isEnabled = true
     config = _config
   } catch (err) {
-    if (!IS_SERVERLESS) {
+    if (IS_SERVERLESS) {
+      log.debug('[ASM] Serverless mode: suppressing error log, calling disable()')
```
IlyasShabi, line 112: I think it's helpful to attach the error here for later investigations.

IlyasShabi (approved):

PR #7803, opened 2026-03-17
fix(appsec): isolate fs state from tracing context

watson:
### What does this PR do?

Isolates AppSec `fs` nesting state from the shared tracing context by moving it into its own async storage namespace.

Today both tracing `fs` and AppSec `fs` tracking try to use the same `legacy` store on `apm:fs:operation:*` channels. This PR makes AppSec keep its `root` / `opExcluded` state in a dedicated AppSec `fs` store instead, and updates the RASP/IAST consumers and tests to read from that store.

### Motivation

This is a small stack step to remove a real production coupling between AppSec and tracing `fs` state management.

Using the same store for both concerns makes behavior depend on channel store ownership and load order. By separating AppSec `fs` state from the main tracing context, we make the current behavior more robust on its own and unblock the follow-up PR that makes `fs` a real configurable plugin target.

### Additional Notes

This PR is intentionally limited to the AppSec side of the issue and does not change `fs` plugin registry/configurability behavior by itself (that will follow in: https://github.com/DataDog/dd-trace-js/pull/7596).

### DIFF
--- a/packages/dd-trace/src/appsec/iast/analyzers/path-traversal-analyzer.js
+++ b/packages/dd-trace/src/appsec/iast/analyzers/path-traversal-analyzer.js
@@ -2,6 +2,7 @@
 
 const path = require('path')
 
+const { APPSEC_FS_STORAGE } = require('../../rasp/fs-plugin')
 const { getIastContext } = require('../iast-context')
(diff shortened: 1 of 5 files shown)


BridgeAR:
This seems to be a good idea. Are you still working on this?

PR #7808, opened 2026-03-17
feat(llmobs): add support for OpenAI Agents

crysmags:
### What does this PR do?
Adds APM tracing and LLMObs instrumentation for the OpenAI Agents SDK (@openai/agents).

  Tracing — instruments the following operations via Orchestrion (rewriter-based):
  - run() — agent execution span (openai-agents.run)
  - getResponse() / getStreamedResponse() — model request spans (openai-agents.getResponse, openai-agents.getStreamedResponse)
  - invokeFunctionTool() — tool call spans (openai-agents.invokeFunctionTool)
  - onInvokeHandoff() — agent handoff spans (openai-agents.onInvokeHandoff)
  - runInputGuardrails() / runOutputGuardrails() — guardrail spans

  LLMObs — adds an LLMObs plugin (packages/dd-trace/src/llmobs/plugins/openai-agents/) that enriches spans with LLM observability tags (model, provider, token usage, input/output).

### Motivation
The OpenAI Agents SDK is a first-party framework from OpenAI for building multi-agent systems in Node.js. It reached a stable API in >=0.7.0. Instrumenting it gives Datadog users distributed tracing and LLM observability for agent workflows without any code changes.

### Additional Notes
  - Instrumentation targets @openai/agents-core and @openai/agents-openai (the sub-packages that contain the actual implementation); @openai/agents is the umbrella re-export.
  - Uses the Orchestrion rewriter pattern (same as ai, langchain) rather than traditional addHook — patches are applied at the function level via versionRange: '>=0.7.0'.
  - 14 tracing tests and 6 LLMObs tests added; both suites run with PLUGINS=openai-agents.


.github/workflows/apm-integrations.yml
```diff
@@ -969,6 +969,17 @@ jobs:
...
 
+  openai-agents:
+    runs-on: ubuntu-latest
+    env:
+      PLUGINS: openai-agents
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: ./.github/actions/plugins/test
+        with:
+          dd_api_key: ${{ secrets.DD_API_KEY }}
```
wconti27: Need to move the llmobs test suite. Side note, we should update the toolkit to generate the llmobs test suite instead of this change if we are writing an llm integration.

index.d.ts
```diff
@@ -2716,6 +2717,12 @@ declare namespace tracer {
      * This plugin automatically instruments the
      * [opensearch](https://github.com/opensearch-project/opensearch-js) module.
      */
+    /**
+     * This plugin automatically instruments the
+     * [openai-agents](https://github.com/npmjs/package/@openai/agents) library.
+     */
+    interface openai_agents extends Instrumentation {}
+
     interface opensearch extends elasticsearch {}
```
wconti27, line 2945: need to fix the comment here. also something thats fixable on the toolkit side to keep formatting correct.

packages/datadog-plugin-openai-agents/src/client.js
```diff
@@ -0,0 +1,127 @@
...
+    if (usage) {
+      if (usage.inputTokens !== undefined) {
+        span.setTag('openai.response.usage.prompt_tokens', usage.inputTokens)
+      }
+      if (usage.outputTokens !== undefined) {
+        span.setTag('openai.response.usage.completion_tokens', usage.outputTokens)
+      }
+      if (usage.totalTokens !== undefined) {
+        span.setTag('openai.response.usage.total_tokens', usage.totalTokens)
+      }
```
wconti27: AFAIK we only tag token usage on the actual llm event spans, and not the apm tracing spans

packages/datadog-plugin-openai-agents/test/index.spec.js
```diff
@@ -0,0 +1,353 @@
...
+  describe('run() - agent.run', () => {
+    // run() produces multiple spans (run + getResponse); use assertSomeTraces
+    // to find the run span regardless of which span arrives first.
+    it('should generate span with correct tags (happy path)', async () => {
+      const traceAssertion = agent.assertSomeTraces((traces) => {
+        const runSpan = traces.flat().find(s => s.name === 'openai-agents.run')
+        assertObjectContains(runSpan, {
+          name: 'openai-agents.run',
+          service: ANY_STRING,
+          resource: ANY_STRING,
```
wconti27: we should assert the actual values here.

wconti27: as well as the rest of the file for values that are constant

packages/datadog-plugin-openai-agents/test/test-setup.js
```diff
@@ -0,0 +1,333 @@
...
+          },
+        }
+      },
+    })
+
+    // Load @openai/agents-openai from its own version directory to get the actual
+    // OpenAIResponsesModel class (instrumented by orchestrion on the prototype)
+    const agentsOpenaiVersionDir = path.join(
+      __dirname, '..', '..', '..', 'versions', '@openai', 'agents-openai@>=0.7.0'
+    )
```
wconti27: the module is already provided from the `this.module` variable.

packages/datadog-plugin-openai-agents/test/test-setup.js
```diff
@@ -0,0 +1,333 @@
...
+            ],
+            usage: {
+              input_tokens: 10,
+              output_tokens: 5,
+              total_tokens: 15,
+            },
+          }
+        },
+      },
+    }
```
wconti27: can we instead use cassettes? I would think thats doable given we can point the base url at the test agent to record them.

packages/dd-trace/src/config/supported-configurations.json
```diff
@@ -25,6 +25,7 @@
         "default": null
       }
     ],
+    "DD_TRACE_OPENAI_AGENTS_ENABLED": ["A"],
```
wconti27: seems like our code generator is generating the wrong format for this, should match the below structure.

packages/dd-trace/src/llmobs/plugins/openai-agents/index.js
```diff
@@ -0,0 +1,288 @@
...
+    const request = ctx.arguments?.[0]
+    const inputMessages = extractInputMessages(request)
+
+    // Streaming spans finish before iteration; output is not available
+    this._tagger.tagLLMIO(span, inputMessages, [{ content: '', role: '' }])
+
+    const metadata = extractMetadata(request)
+    metadata.stream = true
+    this._tagger.tagMetadata(span, metadata)
+  }
```
wconti27, line 391: it seems to have skipped tagging streaming output, which we should collect by wrapping the returned async iterator.

packages/dd-trace/src/llmobs/plugins/openai-agents/index.js
```diff
@@ -0,0 +1,288 @@
...
+/**
+ * Determines the model provider from the OpenAI client base URL.
+ *
+ * @param {string} baseURL - The base URL of the OpenAI client
+ * @returns {string} The model provider name
+ */
+function getModelProvider (baseURL) {
+  if (baseURL.includes('azure')) return 'azure_openai'
+  if (baseURL.includes('deepseek')) return 'deepseek'
+  return 'openai'
```
wconti27: can we parse the url instead of hardcoding?

packages/dd-trace/src/llmobs/plugins/openai-agents/index.js
```diff
@@ -0,0 +1,288 @@
...
+        }
+      } else if (item.type === 'function_call') {
+        let args = item.arguments
+        if (typeof args === 'string') {
+          try {
+            args = JSON.parse(args)
+          } catch {
+            args = {}
+          }
+        }
```
wconti27: can we break this function into some helpers? Would help to improve readability.

packages/dd-trace/test/llmobs/plugins/openai-agents/index.spec.js
```diff
@@ -0,0 +1,340 @@
...
+              {
+                type: 'message',
+                role: 'assistant',
+                content: [{ type: 'output_text', text: 'Hello! How can I help you?' }],
+              },
+            ],
+            usage: { input_tokens: 12, output_tokens: 8, total_tokens: 20 },
+          }),
+        },
+      }
```
wconti27: lets use actual model calls and record with VCR cassettes.

index.d.ts
crysmags: yea we need to change something because this is a consistent issue.

packages/datadog-instrumentations/src/helpers/rewriter/instrumentations/openai-agents.js
```diff
@@ -0,0 +1,215 @@
+'use strict'
+
+// TODO: Add agent-level LLMObs span (kind: 'agent') wrapping per-agent async execution.
+// Python achieves this via add_trace_processor(LLMObsTraceProcessor) which hooks
+// Span.start() / Span.end() on the SDK's internal Span class (dist/tracing/spans.js).
+// The equivalent here would be hooking Span.prototype.start / Span.prototype.end via
+// orchestrion. Requires team sign-off before implementation.
```
sabrenner: does this comment still apply? looks like we patch the `run` method below, this should allow us to capture the agent-level spans i think (but, correct me if i'm wrong, and i'll also run through this locally after giving a first review)

packages/datadog-plugin-openai-agents/src/client.js
```diff
@@ -0,0 +1,127 @@
...
+      tags['ai.request.model'] = modelName
+      tags['openai.request.model'] = modelName
+      tags['resource.name'] = modelName
+    }
+
+    const baseURL = ctx.self?._client?.baseURL
+    if (baseURL) {
+      const host = this.getHostFromBaseURL(baseURL)
+      if (host) {
+        tags['out.host'] = host
```
sabrenner: we typically haven't done stuff like this for the APM side of llm-type or agentic integrations, any reason we're including it here? maybe we're good to just tag model name and provider

packages/datadog-plugin-openai-agents/src/client.js
```diff
@@ -0,0 +1,127 @@
...
+    } catch (e) {
+      log.error('openai-agents: failed to parse baseURL %s', baseURL, e)
+    }
+  }
+
+  getTags (ctx) {
+    const tags = {
+      component: 'openai-agents',
+      'span.kind': 'client',
+      'ai.request.model_provider': 'openai',
```
sabrenner: i actually don't think we set any APM tags for the openai-agents package in the Python integration, we're probably good to just not set any tags

packages/datadog-plugin-openai-agents/src/client.js
sabrenner: yeah, we don't need to add really any tags on the APM spans here

packages/datadog-plugin-openai-agents/src/client.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+}
+
+class GetStreamedResponsePlugin extends BaseOpenaiAgentsClientPlugin {
+  static prefix = 'tracing:orchestrion:@openai/agents-openai:getStreamedResponse'
+  static spanName = 'openai-agents.getStreamedResponse'
+
+  getTags (ctx) {
+    const tags = super.getTags(ctx)
+    tags['openai.request.stream'] = 'true'
```
sabrenner: i think this is maybe the one tag we might wanna keep, but we could also just get rid of it too. all tagging/metadata can just be done on the LLMObs spans

packages/datadog-plugin-openai-agents/src/index.js
```diff
@@ -0,0 +1,20 @@
...
+class OpenaiAgentsPlugin extends CompositePlugin {
+  static id = 'openai-agents'
+  // LLMObs plugins must be registered before tracing plugins for the same operations.
+  // The span-finish diagnostic channel fires synchronously when span.finish() is called,
+  // so setLLMObsTags must run before the tracing plugin's asyncEnd calls span.finish().
+  static plugins = {
+    ...llmobsPlugin,
+    ...internalPlugin,
+    ...clientPlugin,
+  }
```
sabrenner: we have a couple of different approaches at this point - the spreading here, and then the [object accumulation via array exports](https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-plugin-langgraph/src/index.js). would be cool to standardize on how we do it (i guess probably the way we already do it?)

packages/datadog-plugin-openai-agents/src/internal.js
```diff
@@ -0,0 +1,100 @@
+'use strict'
+
+const TracingPlugin = require('../../dd-trace/src/plugins/tracing')
+
+class BaseOpenaiAgentsInternalPlugin extends TracingPlugin {
```
sabrenner, line 5: i think we should not give the static prefixes below in this base class, and instead let all implementers define them (for example, the `RunPlugin` below would define these fields, as the other implementing Plugins here do

packages/dd-trace/src/llmobs/plugins/openai-agents/index.js
```diff
@@ -0,0 +1,591 @@
...
+  }
+}
+
+/**
+ * Determines the model provider from the OpenAI client base URL.
+ *
+ * @param {string} baseURL - The base URL of the OpenAI client
+ * @returns {string} The model provider name
+ */
+function getModelProvider (baseURL) {
```
sabrenner: we just landed a change which updates this logic elsewhere: https://github.com/DataDog/dd-trace-js/commit/a7de9c024efee6e61384207801a740df6f60a32b

i wonder if we can refactor both here and that instance into a shared `getModelProviderFromOpenAIBaseUrl` function, or something like that, so that any logic updates are shared,

packages/dd-trace/src/llmobs/plugins/openai-agents/index.js
sabrenner, line 1: i think for all of the inlined-functions here, we can move them to a `util.js` file in this folder

packages/dd-trace/test/llmobs/plugins/openai-agents/index.spec.js
```diff
@@ -0,0 +1,627 @@
...
+    const savedAgentPort = process.env.DD_TRACE_AGENT_PORT
+    before(() => {
+      delete process.env.DD_TRACE_AGENT_URL
+      delete process.env.DD_TRACE_AGENT_PORT
+    })
+    after(() => {
+      if (savedAgentUrl !== undefined) process.env.DD_TRACE_AGENT_URL = savedAgentUrl
+      if (savedAgentPort !== undefined) process.env.DD_TRACE_AGENT_PORT = savedAgentPort
+    })
+
```
sabrenner: i believe this setup should not be needed, it works fine with all other llmobs tests without this change. are we able to remove these blocks?

packages/dd-trace/test/llmobs/plugins/openai-agents/index.spec.js
```diff
@@ -0,0 +1,627 @@
...
+    let testTool
+    let errorTool
+
+    before(() => {
+      // Clear require cache so modules load fresh through orchestrion rewriter
+      for (const key of Object.keys(require.cache)) {
+        if (key.includes('@openai/agents')) {
+          delete require.cache[key]
+        }
+      }
```
sabrenner: this approach also isn't needed for the `langchain` or `langgraph` suites which also use orchestrion, can we try getting rid of this and follow the same patterns we use in those test suites?

sabrenner (commented):
did a first pass - as @PROFeNoM referenced in his comment, we can try and guide the toolkit to follow the existing tracing results (producing the same spans) that the Python integration has.

packages/datadog-instrumentations/src/helpers/rewriter/instrumentations/openai-agents.js
crysmags: For the simple single-agent case, run() already has everything we need to emit an agent span — we know the agent name, input, and output, and the hook wraps the full execution. The gap is multi-agent handoff scenarios: run() only gives us the starting agent, so we can't derive per-agent execution boundaries mid-run without hooking something lower-level like prepareAgentArtifacts.                                            
                                                                                                               
That said, onInvokeHandoff is already instrumented separately, so the combination of run() + onInvokeHandoff covers most handoff observability. The missing piece is strictly the parent relationship — having the agent span wrap its own LLM calls in a handoff chain.                                                              
                                                                                                               
For the simple case, would it make sense to emit an agent span from the existing run() hook and update the TODO to note the handoff limitation? Or are you looking for full Python parity on the parent hierarchy, which would require a new hook point?

packages/datadog-plugin-openai-agents/src/client.js
crysmags: `client.js` now only sets `component` + `span.kind`. All token/metadata tagging is on LLMObs spans.

packages/datadog-plugin-openai-agents/test/index.spec.js
crysmags: Switched to VCR cassettes; assertions now use recorded values instead of stubs.

packages/dd-trace/src/config/supported-configurations.json
crysmags: Regenerated via the updated script in `34073c2e7`.

packages/dd-trace/src/llmobs/plugins/openai-agents/index.js
crysmags: Commit `45b42c3de` added `GetStreamedResponsePlugin` + `GetStreamedResponseNextPlugin` to wrap the returned async iterator and finish the span when it's exhausted.

packages/dd-trace/src/llmobs/plugins/openai-agents/index.js
crysmags: Uses `getOpenAIModelProvider(baseUrl)` from the shared `utils.js` helper instead of a hardcoded match.

packages/dd-trace/src/llmobs/plugins/openai-agents/index.js
crysmags: Extracted helpers into `packages/dd-trace/src/llmobs/plugins/openai-agents/utils.js`.

packages/datadog-plugin-openai-agents/src/client.js
crysmags: `client.js` reduced to `component` + `span.kind`; all metadata tagging is on LLMObs spans.

packages/datadog-plugin-openai-agents/src/client.js
crysmags: No APM tags set on `client.js` anymore.

packages/datadog-plugin-openai-agents/src/client.js
crysmags: Removed; tagging is entirely on LLMObs spans.

packages/datadog-plugin-openai-agents/src/index.js
crysmags: Switched to the langgraph pattern: each sub-plugin file exports an array, `src/index.js` iterates and registers via `CompositePlugin`.

packages/datadog-plugin-openai-agents/src/internal.js
crysmags: `BaseOpenaiAgentsInternalPlugin` now only has `static id`; each subclass (`RunPlugin`, `InvokeFunctionToolPlugin`, etc.) defines its own `prefix` + `spanName`.

packages/dd-trace/src/llmobs/plugins/openai-agents/index.js
crysmags: Extracted to `utils.js#getOpenAIModelProvider`; the `openai` LLMObs plugin now uses the same helper.

packages/dd-trace/src/llmobs/plugins/openai-agents/index.js
crysmags: Inlined helpers extracted to `packages/dd-trace/src/llmobs/plugins/openai-agents/utils.js`.

packages/dd-trace/test/llmobs/plugins/openai-agents/index.spec.js
crysmags: Test setup now follows the langchain/langgraph pattern using `withVersions` from `setup/mocha`; the extra setup blocks are removed.

packages/dd-trace/test/llmobs/plugins/openai-agents/index.spec.js
crysmags: Test setup now follows the langchain/langgraph pattern using `withVersions` from `setup/mocha`; the extra setup blocks are removed.

PR #7838, opened 2026-03-23
fix(datastreams): preserve extracted schema field descriptions

watson:
### What does this PR do?

This PR fixes several issues that were surfaced while exploring an `@eslint/js` upgrade, but it does **not** include the ESLint upgrade itself.

While working through the lint findings, this change set ended up fixing a mix of test-only bugs, helper error-reporting issues, and a small number of shipped plugin behaviors.

The changes fall into 3 groups:

- Test-only bug fixes and cleanups:
  - `integration-tests/ci-visibility-intake.js`
  - `integration-tests/ci-visibility/intake.spec.js`
  - `packages/datadog-plugin-confluentinc-kafka-javascript/test/index.spec.js`
  - `packages/datadog-plugin-express-mongo-sanitize/test/integration-test/client.spec.js`
  - `packages/datadog-plugin-microgateway-core/test/index.spec.js`
  - `packages/dd-trace/test/telemetry/index.spec.js`

- Error-reporting fixes surfaced during that linting work:
  - `benchmark/e2e-test-optimization/benchmark-run.js`
  - `integration-tests/helpers/index.js`
  - both now preserve the original caught error as `cause` when throwing a new error from a `catch` block

- Shipped plugin/runtime changes:
  - `packages/datadog-plugin-avsc/src/schema_iterator.js`
  - `packages/datadog-plugin-protobufjs/src/schema_iterator.js`

The shipped plugin changes are the ones that deserve the closest review:

- `datadog-plugin-avsc`
  - Behavior change / bug fix:
    - extracted schema property descriptions now use `field.doc`
  - Syntax/static-analysis only:
    - `this.constructor.extractSchema(...)` -> `SchemaExtractor.extractSchema(...)`

- `datadog-plugin-protobufjs`
  - Behavior change / bug fix:
    - extracted schema property descriptions now use `field.comment`
  - Syntax/static-analysis only:
    - `this.constructor.extractSchema(...)` -> `SchemaExtractor.extractSchema(...)`

Everything else in this PR is either test-only, benchmark-only, helper-only, or static-analysis cleanup.

### Motivation

The original motivation was upgrading `@eslint/js`.

That work surfaced several `no-unassigned-vars` findings, along with two places where caught errors were discarded instead of being preserved as the `cause` of a newly thrown error.

Some of the findings were harmless placeholders, but some appeared to be incomplete implementations or real test bugs, and a small number exposed real shipped behavior fixes in plugin code.

The actual ESLint upgrade was intentionally left out of this PR after review. Bumping `@eslint/js` to v10 would also require a coordinated ESLint 10 migration across the repo's lint stack and custom ESLint rules, so this PR keeps the functional fixes but does not include the dependency upgrade itself.

### Additional Notes

Please review the plugin source changes carefully, since those affect shipped code:

- `packages/datadog-plugin-avsc/src/schema_iterator.js`
  - likely real product fix
  - schema descriptions now preserve Avro field docs

- `packages/datadog-plugin-protobufjs/src/schema_iterator.js`
  - likely real product fix
  - schema descriptions now preserve protobuf field comments

The remaining modified plugins are test-only changes.


integration-tests/ci-visibility-intake.js
```diff
@@ -409,37 +409,28 @@ class FakeCiVisIntake extends FakeAgent {
...
 
   assertPayloadReceived (fn, messageMatch, timeout) {
-    let resultResolve
-    let resultReject
-    let error
-
-    const timeoutObj = setTimeout(() => {
-      resultReject([error, new Error('timeout')])
-    }, timeout || 15000)
+    return /** @type {Promise<void>} */ (new Promise((resolve, reject) => {
```
BridgeAR, line 434: This file could be simplified a lot. I did work on this as well before: https://github.com/DataDog/dd-trace-js/pull/7686/changes

packages/datadog-plugin-avsc/src/schema_iterator.js
```diff
@@ -43,9 +43,9 @@ class SchemaExtractor {
     let type
     let format
     let enumValues
-    let description
     let ref
 
+    const description = field.doc
```
BridgeAR, line 48: Nice! This was something I felt unsure about, this seems like the correct fix.

packages/datadog-plugin-confluentinc-kafka-javascript/test/index.spec.js
```diff
@@ -172,27 +172,30 @@ describe('Plugin', () => {
...
+                    resolve()
+                  } catch (e) {
+                    reject(e)
+                  }
                 }
-              }
 
-              consumer.run({ eachMessage: (...args) => eachMessage(...args) })
-                .then(() => sendMessages(kafka, testTopic, messages))
-                .then(() => consumerReceiveMessagePromise)
```
BridgeAR, line 194: I believe this can just be removed. That way it should be simpler

packages/dd-trace/test/telemetry/index.spec.js
```diff
@@ -290,7 +290,6 @@ describe('telemetry', () => {
 describe('telemetry app-heartbeat', () => {
   const HEARTBEAT_INTERVAL = 60000
   let telemetry
-  let pluginsByName
```
BridgeAR, line 301: I guess this works, while it would prevent a use case where this is filled at some point. I went the direction of just assigning an empty object. I have no strong opinion either way

packages/datadog-plugin-confluentinc-kafka-javascript/test/index.spec.js
watson: The line you highlighted is already removed, so I'm not sure exactly what you mean?

packages/dd-trace/test/telemetry/index.spec.js
watson: Done 👍

integration-tests/ci-visibility-intake.js
watson: Simplified 👍

packages/datadog-plugin-express-mongo-sanitize/test/integration-test/client.spec.js
```diff
@@ -34,7 +34,7 @@ withVersions('express-mongo-sanitize', 'express-mongo-sanitize', version => {
 
     for (const variant of varySandbox.VARIANTS) {
       it(`is instrumented loaded with ${variant}`, async () => {
-        const proc = await spawnPluginIntegrationTestProc(sandboxCwd(), variants[variant], agent.port)
+        proc = await spawnPluginIntegrationTestProc(sandboxCwd(), variants[variant], agent.port)
```
BridgeAR, line 37: I think it would be nicer to only define the variable inside of the test

packages/datadog-plugin-microgateway-core/test/index.spec.js
```diff
@@ -20,21 +20,18 @@ describe('Plugin', () => {
   let tracer
   let api
   let gatewayPort
-  let proxyPort
-  let apiPort
 
   const startGateway = (cb) => {
-    const api = http.createServer((req, res) => res.end('OK'))
+    api = http.createServer((req, res) => res.end('OK'))
```
BridgeAR, line 25: Should `api` not stay local?

packages/datadog-plugin-express-mongo-sanitize/test/integration-test/client.spec.js
watson: The whole point of doing it this way is so that `afterEach` callback can actually stop the process (which was the bug that we had before).

packages/datadog-plugin-microgateway-core/test/index.spec.js
watson: Same reason here.

BridgeAR (dismissed):
LGTM

I think we could improve the description test case a bit further by adding a description to our end to end tests. That way we would catch more issues (in case the library would ever rename the property)

BridgeAR (approved):
Great work! Very solid improvement :)

PR #7845, opened 2026-03-23
feat(llmobs): add support for Model Context Protocol SDK

crysmags:
<!-- !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! -->
<!-- Please make sure your changes are properly tested -->
<!-- !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! -->

### What does this PR do?
Adds tracing and LLM Observability instrumentation for the @modelcontextprotocol/sdk client.
                                                                                                                             
  Two spans are produced per MCP client interaction:                                                                         
                                                                                                                             
  - mcp.tool.call — created when Client.callTool() is invoked, tagged with mcp.tool.name, mcp.tool.arguments, mcp.operation, 
  and mcp.server.name
  - mcp.request — created for every underlying Protocol.request() call (e.g. listTools), tagged with rpc.system, mcp.method, 
  mcp.request.id, and mcp.server.name                                                                                        
   
  The LLMObs integration wraps mcp.tool.call spans as tool kind spans, capturing tool input arguments and response content. 

### Motivation
The Model Context Protocol is rapidly becoming the standard interface for connecting LLM agents to external tools and data sources. Instrumenting the MCP client gives users automatic visibility into tool calls made by their AI agents — latency,  errors, which tools are being called, and with what arguments — without any manual instrumentation.   

### Additional Notes
- Uses orchestrion-based diagnostic channels (tracing:orchestrion:@modelcontextprotocol/sdk:Client_callTool and            
  Protocol_request) for zero-overhead instrumentation
  - mcp.server.name is used as the peer.service precursor for service topology mapping                                       
  - Supports @modelcontextprotocol/sdk >= 1.27.1                                                                             
  - LLMObs integration is tested separately in packages/dd-trace/test/llmobs/plugins/modelcontextprotocol-sdk/


packages/datadog-plugin-modelcontextprotocol-sdk/src/client.js
```diff
@@ -0,0 +1,81 @@
+'use strict'
+
+const ClientPlugin = require('../../dd-trace/src/plugins/client')
+
+class BaseModelcontextprotocolSdkClientPlugin extends ClientPlugin {
```
sabrenner:
```suggestion
class BaseModelContextProtocolSdkClientPlugin extends ClientPlugin {
```

ever so small aesthetic/naming nit lol

packages/datadog-plugin-modelcontextprotocol-sdk/src/client.js
```diff
@@ -0,0 +1,81 @@
...
+    const meta = {
+      'ai.operation': 'tool_call',
+      'mcp.tool.name': toolName,
+      'mcp.operation': 'tools/call',
+      'mcp.server.name': serverName,
+    }
+
+    if (toolArguments !== undefined) {
+      meta['mcp.tool.arguments'] = JSON.stringify(toolArguments)
+    }
```
sabrenner: we don't need to tag any of this on the apm span - we can just leave it for the llmobs span

packages/datadog-plugin-modelcontextprotocol-sdk/src/client.js
```diff
@@ -0,0 +1,81 @@
...
+
+    return ctx.currentStore
+  }
+
+  asyncEnd (ctx) {
+    super.finish(ctx)
+  }
+}
+
+class ProtocolRequestPlugin extends BaseModelcontextprotocolSdkClientPlugin {
```
sabrenner: i don't think this needs to extend the other class if it's just re-using the `asyncEnd` function. these can be two separate classes that each extend `ClientPlugin` and each define `asyncEnd`.

packages/datadog-plugin-modelcontextprotocol-sdk/src/client.js
sabrenner: another question: any reason we're using `ClientPlugin` here vs just a generic `TracingPlugin`? i guess technically all of the LLM-space SDKs are clients, but we haven't been labeling them as such

packages/datadog-plugin-modelcontextprotocol-sdk/src/client.js
```diff
@@ -0,0 +1,81 @@
...
+    const meta = {
+      'rpc.system': 'jsonrpc',
+      'rpc.jsonrpc.version': '2.0',
+      'mcp.method': method,
+      'mcp.server.name': serverName,
+    }
+
+    if (requestId !== undefined) {
+      meta['mcp.request.id'] = String(requestId)
+    }
```
sabrenner: likewise here - don't need to do tagging for apm spans, just for llmobs spans

packages/datadog-plugin-modelcontextprotocol-sdk/src/client.js
```diff
@@ -0,0 +1,81 @@
...
+    }, ctx)
+
+    return ctx.currentStore
+  }
+}
+
+// module.exports = {
+//   BaseModelcontextprotocolSdkClientPlugin,
+//   ProtocolRequestPlugin,
+// }
```
sabrenner:
```suggestion
```

packages/datadog-plugin-modelcontextprotocol-sdk/src/client.js
sabrenner, line 1: to make things less confusing, i would name this file `tracing.js`, so it's a bit more distinct from our `llmobs` plugins.

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
```diff
@@ -0,0 +1,89 @@
...
+  static integration = 'modelcontextprotocol-sdk'
+  static prefix = 'tracing:orchestrion:@modelcontextprotocol/sdk:Client_callTool'
+
+  getLLMObsSpanRegisterOptions (ctx) {
+    const params = ctx.arguments?.[0]
+    const toolName = params?.name || 'tool'
+
+    return {
+      kind: 'tool',
+      name: `mcp.tool.${toolName}`,
```
sabrenner: i would format this similarly to how we do it in our python integration: https://github.com/DataDog/dd-trace-py/blob/dc8ccc1c31740d024ea66949d3ffacea3eda047a/ddtrace/llmobs/_integrations/mcp.py#L136-L137

```python
tool_name = args[0] if len(args) > 0 else kwargs.get("name", "unknown_tool")
span_name = "MCP Client Tool Call: {}".format(tool_name)
```

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
sabrenner, line 1: looks like we don't have an llmobs plugin here for the `request` function

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+    this._tagger.tagTextIO(span, input, output)
+
+    const metadata = {}
+    if (toolName) metadata['mcp.tool.name'] = toolName
+    if (serverName) metadata['mcp.server.name'] = serverName
+
+    if (Object.keys(metadata).length > 0) {
+      this._tagger.tagMetadata(span, metadata)
+    }
```
sabrenner: we don't need to tag this as part of the llmobs span for now, just the input and output above is good!

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+}
+
+/**
+ * Formats tool call input as a text string.
+ * @param {string} toolName - The name of the tool being called
+ * @param {object} toolArguments - The arguments passed to the tool
+ * @returns {string} Formatted input string
+ */
+function formatInput (toolName, toolArguments) {
```
sabrenner: let's move this and the `formatOutput` function to a utility file

sabrenner (commented):
code mostly looks good to me! i'm not sure what the scope is for this first go, i assume it's not to be in complete parity with our Python integration, which is OK, but if it's easy enough to include everything we do have, I would say these are good refs:

1. [Patching code](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/contrib/internal/mcp/patch.py) - what functions we patch
2. [Integration code](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/llmobs/_integrations/mcp.py#L136-L137) - how we tag the llmobs spans.

it could be cool to point the toolkit at those and see what it makes of that.

packages/datadog-plugin-modelcontextprotocol-sdk/src/client.js
crysmags: Fixed — renamed to `McpToolCallPlugin` (dropped the `Base` prefix too since `ProtocolRequestPlugin` was removed and there's no longer a base class).

packages/datadog-plugin-modelcontextprotocol-sdk/src/client.js
crysmags: Fixed — removed all meta tags from `bindStart`. APM span now only carries `resource`, `type`, and `kind`.

packages/datadog-plugin-modelcontextprotocol-sdk/src/client.js
crysmags: Fixed — removed `ProtocolRequestPlugin` entirely (see reply on the Protocol.request comment). Both classes now independently extend `TracingPlugin` with their own `asyncEnd`.

packages/datadog-plugin-modelcontextprotocol-sdk/src/client.js
crysmags: Fixed — switched to `TracingPlugin` to align with openai and anthropic integrations.

packages/datadog-plugin-modelcontextprotocol-sdk/src/client.js
crysmags: Fixed — `ProtocolRequestPlugin` was removed entirely, so this is gone with it.

packages/datadog-plugin-modelcontextprotocol-sdk/src/client.js
crysmags: Fixed — removed the commented-out block.

packages/datadog-plugin-modelcontextprotocol-sdk/src/client.js
crysmags: Fixed — renamed to `tracing.js` and updated `require('./client')` → `require('./tracing')` in `index.js`.

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
crysmags: Fixed — span name is now `MCP Client Tool Call: ${toolName}` with fallback to `unknown_tool`.

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
crysmags: `ProtocolRequestPlugin` was removed. `Protocol.request` is the right hook for distributed tracing header injection (injecting DD trace context into MCP request `_meta` so a server-side span can be linked), but that only makes sense once we have server-side instrumentation to receive it. Left a note in the orchestrion rewriter — follow-up for when server-side coverage is added.

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
crysmags: Fixed — removed the metadata tagging block. `setLLMObsTags` now only calls `tagTextIO`.

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
crysmags: Fixed — `formatInput` and `formatOutput` extracted to `utils.js` in the same directory.

packages/datadog-plugin-modelcontextprotocol-sdk/src/tracing.js
```diff
@@ -0,0 +1,69 @@
...
+
+class McpToolCallPlugin extends TracingPlugin {
+  static id = 'modelcontextprotocol_client'
+  static prefix = 'tracing:orchestrion:@modelcontextprotocol/sdk:Client_callTool'
+
+  bindStart (ctx) {
+    const params = ctx.arguments?.[0]
+    const toolName = params?.name
+
+    this.startSpan('mcp.tool.call', {
```
sabrenner:
```suggestion
    this.startSpan('mcp.client.tool.call', {
```

more of an open question - should we be doing this for apm spans? python does something like [`client_tool_call`](https://github.com/DataDog/dd-trace-py/blob/8d7138a353a968ad6491152dd7c0137f4ee593c7/ddtrace/llmobs/_integrations/mcp.py#L28), i'm fine to differ here, but wondering if we want to include `client` in the span name somehow (even though its span kind is `client`).

packages/datadog-plugin-modelcontextprotocol-sdk/src/tracing.js
```diff
@@ -0,0 +1,69 @@
...
+    super.finish(ctx)
+  }
+}
+
+class McpListToolsPlugin extends TracingPlugin {
+  static id = 'modelcontextprotocol_list_tools'
+  static prefix = 'tracing:orchestrion:@modelcontextprotocol/sdk:Client_listTools'
+
+  bindStart (ctx) {
+    this.startSpan('mcp.list_tools', {
```
sabrenner:
```suggestion
    this.startSpan('mcp.tools.list', {
```

maybe just to be more like the above? (again, this would differ from python, but i think small naming differentiations is ok)

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
```diff
@@ -0,0 +1,79 @@
...
+    if (!span) return
+
+    const params = ctx.arguments?.[0]
+    const toolName = params?.name
+    const toolArguments = params?.arguments
+
+    const input = formatInput(toolName, toolArguments)
+    const output = ctx.error ? undefined : formatOutput(ctx.result)
+
+    this._tagger.tagTextIO(span, input, output)
```
sabrenner, line 44: let's also set the `mcp_tool_kind: "client"` tag on this span ([python ref](https://github.com/DataDog/dd-trace-py/blob/8d7138a353a968ad6491152dd7c0137f4ee593c7/ddtrace/llmobs/_integrations/mcp.py#L139))

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/utils.js
```diff
@@ -0,0 +1,48 @@
...
+}
+
+/**
+ * Formats MCP tool call result as a text string.
+ * MCP tool results contain a `content` array with items like:
+ * `[{ type: 'text', text: '...' }, { type: 'image', data: '...', mimeType: '...' }]`
+ * @param {object} result - The MCP CallToolResult
+ * @returns {string} Formatted output string
+ */
+function formatOutput (result) {
```
sabrenner, line 30: i think we can/should stick more closely to the [python logic](https://github.com/DataDog/dd-trace-py/blob/8d7138a353a968ad6491152dd7c0137f4ee593c7/ddtrace/llmobs/_integrations/mcp.py#L154-L157) if it's applicable. specifically, formatting each [text part ](https://github.com/DataDog/dd-trace-py/blob/8d7138a353a968ad6491152dd7c0137f4ee593c7/ddtrace/llmobs/_integrations/mcp.py#L99-L113).

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
```diff
@@ -0,0 +1,79 @@
...
+  setLLMObsTags (ctx) {
+    const span = ctx.currentStore?.span
+    if (!span) return
+
+    const params = ctx.arguments?.[0]
+    const toolName = params?.name
+    const toolArguments = params?.arguments
+
+    const input = formatInput(toolName, toolArguments)
+    const output = ctx.error ? undefined : formatOutput(ctx.result)
```
sabrenner: let's have the output annotation be an object of `content` and `isError`, [as done in Python](https://github.com/DataDog/dd-trace-py/blob/8d7138a353a968ad6491152dd7c0137f4ee593c7/ddtrace/llmobs/_integrations/mcp.py#L158)

packages/datadog-instrumentations/src/helpers/rewriter/instrumentations/modelcontextprotocol-sdk.js
```diff
@@ -0,0 +1,85 @@
...
+      name: '@modelcontextprotocol/sdk',
+      versionRange: '>=1.27.1',
+      filePath: 'dist/cjs/client/index.js',
+    },
+    functionQuery: {
+      methodName: 'connect',
+      className: 'Client',
+      kind: 'Async',
+    },
+    channelName: 'Client_connect',
```
sabrenner, line 83: (preface) i would maybe leave this for last, as i'm not 100% sure about this, even though it matches the behavior we have in python. if it ends up completely upending the PR, we can punt this for another PR

the way the python integration treats client sessions/connections is by encapsulating the whole session as one span. would that be possible at all to do here? if not, we can follow up with it. so like

```js
const client = new Client({ name: 'my-client', version: '1.0.0' });
const transport = new StreamableHTTPClientTransport(baseUrl);
await client.connect(transport); // <-- start client span

// all mcp operations for this client session

await transport.terminateSession();
await client.close(); // <-- end client span
```

see how this is [accomplished in python](https://github.com/DataDog/dd-trace-py/blob/8d7138a353a968ad6491152dd7c0137f4ee593c7/ddtrace/contrib/internal/mcp/patch.py#L291-L292), for reference. maybe we could just keep a weakmap of client sessions to context objects in the plugins or something, and instrument both `connect` and `close`.

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
```diff
@@ -0,0 +1,79 @@
...
+
+class McpListToolsLLMObsPlugin extends LLMObsPlugin {
+  static id = 'llmobs_mcp_list_tools'
+  static integration = 'modelcontextprotocol-sdk'
+  static prefix = 'tracing:orchestrion:@modelcontextprotocol/sdk:Client_listTools'
+
+  getLLMObsSpanRegisterOptions () {
+    return {
+      kind: 'task',
+      name: 'MCP Client list Tools',
```
sabrenner:
```suggestion
      name: 'MCP Client List Tools',
```

lol i checked and we do lowercase in python but it's probs a typo, we can just uppercase it here

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
```diff
@@ -0,0 +1,79 @@
...
+
+  getLLMObsSpanRegisterOptions () {
+    return {
+      kind: 'workflow',
+      name: 'MCP Client Session',
+    }
+  }
+
+  setLLMObsTags () {
+    // No meaningful I/O to capture for a connection lifecycle span
```
sabrenner, line 78: we should tag `mcp_server_name`, `mcp_server_version`, and `mcp_server_title` (from the client info passed in during `new Client` initialization).

packages/datadog-plugin-modelcontextprotocol-sdk/test/test-setup.js
```diff
@@ -0,0 +1,87 @@
...
+    this._InMemoryTransport = InMemoryTransport
+    this._CallToolRequestSchema = CallToolRequestSchema
+    this._ListToolsRequestSchema = ListToolsRequestSchema
+
+    this._server = new Server(
+      { name: 'test-server', version: '1.0.0' },
+      { capabilities: { tools: {} } }
+    )
+
+    this._server.setRequestHandler(CallToolRequestSchema, async (request) => {
```
sabrenner: could we instead register tools via the actual `registerTool` API? ie as is officially documented [here](https://github.com/modelcontextprotocol/typescript-sdk/blob/7ba58dac512ca34a699243458fb95b882362d69b/docs/server.md)

sabrenner (commented):
looking a lot better for the first pr! thanks for separating out the client features first before supporting server side requests 😄

packages/datadog-plugin-modelcontextprotocol-sdk/src/tracing.js
crysmags: Fixed — updated span name to `mcp.client.tool.call` to include `client` in the name.

packages/datadog-plugin-modelcontextprotocol-sdk/src/tracing.js
crysmags: Fixed — updated to `mcp.tools.list` to stay consistent with the naming above.

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
crysmags: Fixed — added `mcp_tool_kind: 'client'` tag in `setLLMObsTags`.

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/utils.js
crysmags: Fixed — `formatOutput` now filters to `type === 'text'` content blocks and extracts `text`, `annotations`, and `meta` fields for each, matching the Python logic.

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
crysmags: Fixed — output annotation is now a `{ content, isError }` object, matching the Python format.

packages/datadog-instrumentations/src/helpers/rewriter/instrumentations/modelcontextprotocol-sdk.js
crysmags: ~~Added a `McpConnectPlugin` that instruments `Client.connect` as a starting point. It doesn't yet span the full session lifecycle (connect → close) — that would need a WeakMap to hold context across the two calls. Happy to punt the full lifecycle span to a follow-up PR to keep this one scoped.~~
will address in followup pr if needed.

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
crysmags: Fixed — updated to `MCP Client List Tools` (uppercase).

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
crysmags: Fixed — now tagging `mcp_server_name`, `mcp_server_version`, and `mcp_server_title` from `ctx.self._serverVersion` (populated after the client connects).

packages/datadog-plugin-modelcontextprotocol-sdk/test/test-setup.js
crysmags: Fixed — test setup now uses the `McpServer.registerTool` API instead of manually wiring handlers.

packages/datadog-plugin-modelcontextprotocol-sdk/test/test-setup.js
```diff
@@ -0,0 +1,72 @@
...
+    const { Client } = clientModule
+    // Use versionMod.getPath to resolve the SDK root since the package exports map
+    // remaps @modelcontextprotocol/sdk/package.json to dist/cjs/package.json
+    const clientEntryPath = versionMod.getPath('@modelcontextprotocol/sdk/client')
+    const sdkDir = path.resolve(path.dirname(clientEntryPath), '..', '..', '..')
+    const { McpServer } = require(path.join(sdkDir, 'dist/cjs/server/mcp.js'))
+    const { InMemoryTransport } = versionMod.get('@modelcontextprotocol/sdk/inMemory.js')
+
+    this._McpServer = McpServer
+    this._InMemoryTransport = InMemoryTransport
```
sabrenner: nit: idt we use these in the class at all?

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
```diff
@@ -0,0 +1,65 @@
...
+
+  getLLMObsSpanRegisterOptions () {
+    return {
+      kind: 'task',
+      name: 'MCP Client List Tools',
+    }
+  }
+
+  setLLMObsTags () {
+    // No meaningful I/O to capture for a list tools span
```
sabrenner: sorry, i thought i had left a comment here before, but can we capture the returned list of tools? ie i think just doing
```js
this._tagger.tagTextIO(span, null, JSON.stringify(ctx.result)
```

packages/dd-trace/test/llmobs/plugins/modelcontextprotocol-sdk/index.spec.js
```diff
@@ -0,0 +1,219 @@
...
+        const versionModule = require(`../../../../../../versions/@modelcontextprotocol/sdk@${version}`)
+
+        // Require the client submodule first so RITM patches it before the server loads it transitively
+        Client = versionModule.get('@modelcontextprotocol/sdk/client').Client
+
+        // The package exports map remaps package.json to dist/cjs/package.json, so navigate
+        // up from the resolved client entry path to find the SDK root directory
+        const clientEntryPath = versionModule.getPath('@modelcontextprotocol/sdk/client')
+        const sdkDir = path.resolve(path.dirname(clientEntryPath), '..', '..', '..')
+        McpServer = require(path.join(sdkDir, 'dist/cjs/server/mcp.js')).McpServer
```
sabrenner, line 36: general toolkit question / not blocking the pr - should we have it re-use the `test-setup` from the APM plugin tests? i think it looks like the same setup

sabrenner (commented):
lgtm!! only the ask on annotating the list tools output is actionable, i thought i had left a comment previously so that's my bad, but otherwise the other two are just notes to maybe apply in the next pr

packages/datadog-plugin-modelcontextprotocol-sdk/test/test-setup.js
crysmags: Fixed — removed the unused `_McpServer`, `_clientTransport`, and `_serverTransport` fields from `setup()`.

packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js
crysmags: Fixed — `McpListToolsLLMObsPlugin.setLLMObsTags` now tags `JSON.stringify(ctx.result)` as the output. Updated the spec to assert on `outputValue` as well.

packages/dd-trace/test/llmobs/plugins/modelcontextprotocol-sdk/index.spec.js
crysmags: Good call — I kept them separate for now since the LLMObs spec currently registers additional tools (`multi-content-tool`) that the APM spec doesn't need. Happy to consolidate into a single shared `test-setup` as a follow-up once the shape stabilizes.

sabrenner: makes sense to me! yeah maybe we can consolidate in a follow-up

sabrenner (approved):

PR #7877, opened 2026-03-27
Fix/rewriter this property name

crysmags:
### What does this PR do?
1. Add thisPropertyName query type to the transformer

    The existing transformer only supported functionName, methodName, and expressionName for targeting functions. This
     adds thisPropertyName, which matches arrow functions assigned to instance properties inside constructors:

    // Matches: this._queryPromise = (sql, values) => { ... }
    functionQuery: { thisPropertyName: '_queryPromise', kind: 'Async' }

    The AST query used is:
    AssignmentExpression[left.object.type=ThisExpression][left.property.name="${thisPropertyName}"] > [async]

    2. Fix traceFunction to convert ArrowFunctionExpression to FunctionExpression

    Arrow functions have no own arguments or this binding — they capture both from their lexically enclosing scope.
    When the rewriter wraps an arrow function, the generated wrapper references arguments to forward call-site
    arguments to the inner function. If the outer function remains an ArrowFunctionExpression, arguments resolves to
    the enclosing constructor's arguments (e.g., the connection options object) rather than the actual call-site
    arguments (e.g., [sql, values]).

    Converting to FunctionExpression gives the wrapper its own arguments and a dynamic this binding, making span
    resource and argument forwarding work correctly.

### Additional Notes
  Test plan

    - New test case in packages/datadog-instrumentations/test/helpers/rewriter/index.spec.js: should auto instrument
    arrow function instance properties (thisPropertyName)
    - Verify mariadb v2 query spans have correct resource (SQL string, not connection options)

### DIFF
--- a/packages/datadog-instrumentations/src/helpers/rewriter/orchestrion/transformer.js
+++ b/packages/datadog-instrumentations/src/helpers/rewriter/orchestrion/transformer.js
@@ -86,7 +86,7 @@ class Transformer {
   }
 
   #fromFunctionQuery (functionQuery) {
-    const { functionName, expressionName, className } = functionQuery
+    const { functionName, expressionName, className, thisPropertyName } = functionQuery
     const method = functionQuery.methodName || functionQuery.privateMethodName
     const type = functionQuery.privateMethodName ? 'PrivateIdentifier' : 'Identifier'
     const queries = []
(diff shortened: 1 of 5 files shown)


rochdev (changes requested):
This should be brought to Orchestrion-JS.

Also, why have this work specifically for `this`? Why not pass an object name which could be `this` or anything else so that something like `obj.foo = ` would also work?

BridgeAR:
I believe this moved to Orchestrion-JS, closing

PR #7881, opened 2026-03-29
feat(span): internal API split for Span

pabloerhard:
### What does this PR do?

This PR adds an internal API split for Span which allows a distinction between the internal use of Span and the user use of Span. This internal API split consists of adding a wrapper to the user exposed Span through the proxy. 

### Motivation

Having a distinction between internal and external use of Span, and acting accordingly.


packages/datadog-plugin-aerospike/test/index.spec.js
```diff
@@ -247,7 +247,7 @@ describe('Plugin', () => {
             aerospike.connect(config).then(client => {
               tracer.scope().activate(obj, () => {
                 client.put(key, { i: 123 }, () => {
-                  assert.strictEqual(tracer.scope().active(), obj)
+                  assert.strictEqual(tracer.scope().active()._span, obj)
```
rochdev: The problem is that we're using the API incorrectly to begin with. Instead of accessing the internal `_span` property, the activated span should be an actual span, so instead of `const obj = {}` it should be `const span = tracer.startSpan()`. This is also true for all the other tests.

packages/datadog-plugin-avsc/test/index.spec.js
```diff
@@ -81,7 +81,7 @@ describe('Plugin', () => {
             const buf = type.toBuffer({ name: 'Alyssa', favorite_number: 256, favorite_color: null })
             fs.writeFileSync(filePath, buf)
 
-            assert.strictEqual(span._name, 'user.serialize')
+            assert.strictEqual(span.context()._name, 'user.serialize')
```
rochdev, line 84: I don't remember exactly the history behind it, but I remember `span._ name` having a different semantic than `span._context._name`. In this case the latter is probably correct though so I think we're fine.

packages/datadog-plugin-cypress/src/cypress-plugin.js
```diff
@@ -1035,11 +1035,11 @@ class CypressPlugin {
         if (isQuarantinedFromSupport) {
           this.activeTestSpan.setTag(TEST_MANAGEMENT_IS_QUARANTINED, 'true')
         }
-
+        const activeSpan = this.activeTestSpan._span || this.activeTestSpan
         const finishedTest = {
           testName,
           testStatus,
-          finishTime: this.activeTestSpan._getTime(), // we store the finish time here
+          finishTime: activeSpan._getTime(), // we store the finish time here
```
rochdev: What should be on the store is an internal span, not a public span, so this change shouldn't be needed.

packages/datadog-plugin-dd-trace-api/test/index.spec.js
```diff
@@ -92,40 +92,48 @@ describe('Plugin', () => {
       describe('scope:active', () => {
         it('should call underlying api', () => {
           scope = tracer.scope()
-          sinon.spy(scope, 'active')
+          const internalScope = scope._scope || scope
```
rochdev: Is this needed? Is the scope ever used internally other than in tests where the public API is used anyway?

packages/dd-trace/src/noop/proxy.js
```diff
@@ -4,6 +4,9 @@ const NoopAppsecSdk = require('../appsec/sdk/noop')
 const NoopLLMObsSDK = require('../llmobs/noop')
 const NoopFlaggingProvider = require('../openfeature/noop')
 const NoopAIGuardSDK = require('../aiguard/noop')
+const PublicSpan = require('../opentracing/public/span')
```
rochdev: Nit pick: I'd probably rename the internal one just so that stack traces show `Span` for the public one. If that's too big a change, then we can keep `PublicSpan`. We can always use an alias as well.

packages/dd-trace/src/opentelemetry/context_manager.js
```diff
@@ -16,7 +16,8 @@ class ContextManager {
   active () {
     const store = this._store.getStore()
     const baseContext = store || ROOT_CONTEXT
-    const activeSpan = tracer.scope().active()
+    const publicSpan = tracer.scope().active()
+    const activeSpan = publicSpan?._span || publicSpan
```
rochdev: Why does OTel need to know about the internal span? Not that it's necessarily a problem, just wondering. Also, in what cases would `publicSpan?._span` be undefined?

packages/dd-trace/src/opentracing/public/scope.js
```diff
@@ -0,0 +1,24 @@
+'use strict'
+
+const PublicSpan = require('./span')
+
+class Scope {
```
rochdev: I don't think this is needed as we always use the store directly internally, so the existing API is already only public-facing.

packages/dd-trace/src/opentracing/tracer.js
```diff
@@ -81,7 +82,7 @@ class DatadogTracer {
   }
 
   inject (context, format, carrier) {
-    if (context instanceof Span) {
+    if (context instanceof Span || context instanceof PublicSpan) {
```
rochdev: Should it be that the public API only accepts `PublicSpan` and the internal one only accepts `Span`?

packages/dd-trace/src/opentracing/public/span.js
```diff
@@ -0,0 +1,80 @@
...
+ * external usage and acting accordingly.
+ */
+class PublicSpan {
+  constructor (span) {
+    this._span = span
+  }
+
+  // A WeakMap cache ensures the same wrapper instance is returned for the same
+  // underlying span, so reference equality checks (===) in user code remain stable.
+  static wrap (span) {
```
rochdev: If this is ever used in `tracer.js`, should it live there?

packages/dd-trace/src/opentracing/public/span.js
```diff
@@ -0,0 +1,80 @@
...
+    if (span instanceof PublicSpan) {
+      return span
+    }
+    const cached = cache.get(span)
+    if (cached !== undefined) {
+      return cached
+    }
+    const wrapper = new PublicSpan(span)
+    try {
+      cache.set(span, wrapper)
```
rochdev: Why would that ever throw?

rochdev: Especially with an empty catch?

packages/dd-trace/src/opentracing/public/span.js
```diff
@@ -0,0 +1,80 @@
...
+ */
+class PublicSpan {
+  constructor (span) {
+    this._span = span
+  }
+
+  // A WeakMap cache ensures the same wrapper instance is returned for the same
+  // underlying span, so reference equality checks (===) in user code remain stable.
+  static wrap (span) {
+    if (span instanceof PublicSpan) {
```
rochdev: We never use `tracer.trace` or `tracer.wrap` internally, so this check is not needed.

packages/dd-trace/src/noop/proxy.js
```diff
@@ -51,8 +54,17 @@ class NoopProxy {
...
+        ...options.tags,
+        [SVC_SRC_KEY]: 'm',
+      }
+    }
 
-    return this._tracer.trace(name, options, fn)
+    const callback = fn.length > 1
+      ? function (span, done) { return fn(PublicSpan.wrap(span), done) }
+      : function (span) { return fn(PublicSpan.wrap(span)) }
+    return this._tracer.trace(name, options, callback)
```
rochdev: These methods (trace and wrap) only exist publicly, so they should always be using PublicSpan.

packages/dd-trace/src/opentracing/public/span.js
```diff
@@ -0,0 +1,80 @@
...
+      this._span.setTag(SVC_SRC_KEY, 'm')
+    }
+    this._span.addTags(tags)
+    return this
+  }
+}
+
+// Whenever a method needs to be modified to have a unique public behavior, it
+// should be removed from this list.
+for (const method of [
```
rochdev: Why not just loop through the prototype instead of maintaining a list?

rochdev (changes requested):

packages/dd-trace/src/opentracing/public/span.js
pabloerhard: stub tests that return random stuff

packages/dd-trace/src/opentracing/public/span.js
pabloerhard: Yeah this would probably be better, but we would still need to know which methods to skip so we don't overwrite them

packages/dd-trace/src/opentelemetry/context_manager.js
pabloerhard: It's needed for this comparison: storedSpan._ddSpan === activeSpan

packages/dd-trace/src/opentracing/tracer.js
pabloerhard: Inject does not live on the Public Span api though

packages/dd-trace/src/constants.js
```diff
@@ -30,6 +30,7 @@ module.exports = {
   ERROR_STACK: 'error.stack',
   IGNORE_OTEL_ERROR: Symbol('ignore.otel.error'),
   COMPONENT: 'component',
+  SVC_SRC_KEY: '_dd.svc_src',
```
pabloerhard: This file change will not be needed once https://github.com/DataDog/dd-trace-js/pull/7824 is merged.
This should be resolved depending on what PR gets merged first.

packages/datadog-plugin-cypress/src/cypress-plugin.js
pabloerhard: This does not come from the store, it uses the tracer to start a span which then returns a wrapped a span.

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -285,6 +289,7 @@ class LLMObs extends NoopLLMObs {
 
   exportSpan (span) {
     span = span || this._active()
+    span = span?._span || span
```
sabrenner: i think to make this just a little cleaner, having a global helper like `getInternalSpan(span)` with a comment explaining why it's needed would help clean this up a bit. because then this line could be come

```js
span = getInternalSpan(span || this._active())
```

and so on - even outside of this file. definitely not a hard blocker, i just think it would be a bit cleaner & the method could document what it's accomplishing

sabrenner (commented):
lgtm from an mlobs perspective (just checked `llmobs/sdk.js` and `llmobs/sdk/index.spec.js`)! left one non-blocking comment that i think could just help clean up implementation and readability ever so slightly.

i'll leave the final approval for a more relevant team 😄

packages/dd-trace/src/llmobs/sdk.js
pabloerhard: I'm still debating this exact change, ideally public facing API's should only receive an instance of PublicSpan, but since it seems like llmobs makes use of this functions internally it ends up handling both type of spans.

pabloerhard:
@codex review

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -199,7 +199,7 @@ class AIGuard extends NoopAIGuard {
       const metaStruct = {
         messages: this.#buildMessagesForMetaStruct(messages),
       }
-      span.meta_struct = {
+      span._span.meta_struct = {
```
rochdev: This should be done at the very beginning of the callback so that the entire function gets the internal span.

packages/dd-trace/src/opentracing/public/span.js
```diff
@@ -0,0 +1,73 @@
...
+
+const cache = createPrivateMap('dd.publicSpan')
+
+/**
+ * This is a public wrapper of Span, this allows distinguishing internal usage from
+ * external usage and acting accordingly.
+ */
+class PublicSpan {
+  constructor (span) {
+    if (span instanceof PublicSpan) {
```
rochdev: This should not be possible.

packages/dd-trace/src/tracer.js
```diff
@@ -59,8 +60,11 @@ class DatadogTracer extends Tracer {
 
   trace (name, options, fn) {
     options = { childOf: this.scope().active(), ...options }
+    if (options?.childOf instanceof PublicSpan) {
```
rochdev: This check is not needed, `tracer.trace` cannot receive an internal span.

packages/dd-trace/src/util.js
```diff
@@ -2,6 +2,39 @@
 
 const path = require('path')
 
+const createPrivateSymbol = probeCreatePrivateSymbol()
+
+function probeCreatePrivateSymbol () {
+  try {
+    // eslint-disable-next-line no-new-func
+    return new Function('name', 'return %CreatePrivateSymbol(name)')
+  } catch {}
```
rochdev: Why not put the rest of the function in the `catch` instead of just swallowing the error? It would make the relationship more obvious.

rochdev: It would also help to add comment between each step to explain what it's doing to a reader without any knowledge of `CreatePrivateSymbol`.

packages/dd-trace/src/opentracing/public/span.js
```diff
@@ -0,0 +1,73 @@
...
+function uncachedWrapper (span) {
+  // Skips the cache entirely — intentional for startSpan.
+  const wrapper = Object.create(PublicSpan.prototype)
+  wrapper._span = span
+  return wrapper
+}
+
+function cacheWrapper (wrapper) {
+  if (!cache.has(wrapper._span)) {
+    cache.set(wrapper._span, wrapper)
```
rochdev: Is this now fast enough to avoid the wrapper functions? Now that it's a symbol it should just be an assignment which is very low cost.

packages/dd-trace/src/opentracing/public/span.js
```diff
@@ -0,0 +1,73 @@
...
+  addTags (tags) {
+    if (tags && (tags[SERVICE_KEY] || tags[SERVICE_NAME_KEY])) {
+      this._span.setTag(SVC_SRC_KEY, 'm')
+    }
+    this._span.addTags(tags)
+    return this
+  }
+}
+
+// This is only used for startSpan which is guaranteed to not be activated.
```
rochdev: Where is this used? I can't seem to find it. Also, I'm not sure to understand why it would be needed in the first place.

rochdev (changes requested):
A few small things, but looks good for the most part!

packages/dd-trace/src/opentracing/public/span.js
pabloerhard: This is used for startSpan, the startSpan we expose to users through the proxy.

packages/dd-trace/src/opentracing/public/span.js
pabloerhard: We can test it out

packages/dd-trace/src/util.js
pabloerhard: In my mind it looked cleaner this way, instead of having it nested.

packages/dd-trace/src/tracer.js
pabloerhard: I added this because childOf can be either SpanContext or PublicSpan

rochdev (dismissed):

packages/dd-trace/src/noop/proxy.js
```diff
@@ -55,6 +57,12 @@ class NoopProxy {
     if (typeof fn !== 'function') return
 
     options = options || {}
+    if (options.service || options?.tags?.service || options?.tags?.['service.name']) {
```
BridgeAR: Options are guaranteed to be truthy with the line above.

Are options coming from the user or are they from the config? If they come from the config, tags is also always truthy and service is always falling back to tags.service already, so no need to do that.

packages/dd-trace/src/noop/proxy.js
```diff
@@ -68,7 +76,12 @@ class NoopProxy {
     if (typeof fn !== 'function') return fn
 
     options = options || {}
-
+    if (options.service || options?.tags?.service || options?.tags?.['service.name']) {
```
BridgeAR: Same as above

packages/dd-trace/src/noop/proxy.js
```diff
@@ -55,6 +57,12 @@ class NoopProxy {
     if (typeof fn !== 'function') return
 
     options = options || {}
+    if (options.service || options?.tags?.service || options?.tags?.['service.name']) {
+      options.tags = {
+        ...options.tags,
+        [SVC_SRC_KEY]: 'm',
+      }
```
BridgeAR: I am not sure I understand why we spread.
We mutate the options, so it is likely not a user object (by writing a new tags object), while we do not want to mutate the original tags object?

The same applies to other cases here.

packages/dd-trace/src/noop/proxy.js
pabloerhard: options are coming from the user

packages/dd-trace/src/util.js
```diff
@@ -2,6 +2,54 @@
 
 const path = require('path')
 
+let createPrivateSymbol = probeCreatePrivateSymbol()
+
+// Attempts to obtain a factory for V8's %CreatePrivateSymbol, a native runtime function that
+// produces a property key invisible to all reflection APIs (Object.getOwnPropertySymbols, Proxy, etc.).
+function probeCreatePrivateSymbol () {
```
BridgeAR: Why is that needed?

pabloerhard: This is needed to avoid WeakMap lookups on a hot path while also avoiding exposing anything, since, as mentioned in the comment, it makes the key invisible to reflection APIs. This was microbenchmarked and is up to 2x faster.

packages/dd-trace/src/noop/proxy.js
pabloerhard: Actually spreading should be done for both options and tags.

packages/dd-trace/src/util.js
pabloerhard: As for the cache, it's needed because a user could, for example, do something like the following:

```
const span1 = tracer.scope().active()
const span2 = tracer.scope().active()

if (span1 === span2) { do () }
```

if active is an internal span, it will be wrapped by a PublicSpan, with out the cache the comparison would've failed as the same internal span would've been wrapped by 2 different instances of PublicSpan.

BridgeAR (dismissed):
Having a separation is a good idea. This one just leaks internals due to the way our code is structured and adds additional overhead.

For this particular thing, I believe we need to go back and find alternatives. I did look into this with @pabloerhard earlier in a call and suggested a simpler implementation.

packages/dd-trace/src/opentracing/public/span.js
```diff
@@ -0,0 +1,88 @@
...
+    return this.#span.logEvent.apply(this.#span, arguments)
+  }
+
+  finish () {
+    return this.#span.finish.apply(this.#span, arguments)
+  }
+
+  static unwrap (value) {
+    return value instanceof PublicSpan ? value.#span : value
+  }
```
BridgeAR: This looses the encapsulation, similar to the underscored ones. A user can use this just the same.

pabloerhard: What about using something like the following: 

```
static {
    unwrap = (value) => value instanceof PublicSpan ? value.#span : value
  }
```

pabloerhard: This was the original approach, this completely removes access through the constructor and a user wouldn't have access through the PublicSpan instances we expose.

packages/dd-trace/src/datastreams/checkpointer.js
```diff
@@ -64,7 +65,8 @@ class DataStreamsCheckpointer {
    */
   trackTransaction (transactionId, checkpointName, span = null) {
     if (!this.config.dsmEnabled) return
-    const activeSpan = span ?? this.tracer.scope().active()
+    const activeSpan = span ?? storage?.('legacy').getStore()?.span ?? null
```
BridgeAR: We could just move the `storage('legacy')` part to the top. That would now change.

The same for all other spots where the same is used.

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -116,12 +119,12 @@ class LLMObs extends NoopLLMObs {
 
     if (fn.length > 1) {
       return this._tracer.trace(name, spanOptions, (span, cb) =>
-        this.#activate(span, { kind, ...llmobsOptions }, () => fn(span, cb))
+        this.#activate(unwrap(span), { kind, ...llmobsOptions }, () => fn(span, cb))
```
BridgeAR: How do I know where to unwrap or where not to unwrap? That is something that seems quite difficult with the design

pabloerhard: Unwrap is only needed when a public facing API ingests a PublicSpan. LLMObs/aiguard are a special cases because its tracer are built on top of the tracer we expose to users. That said, AFAIK we don’t add user facing APIs often, so this scenario is unlikely.

We could add a comment explaining when it's appropiate to use it to prevent bad usage of it.

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -187,6 +188,7 @@ class AIGuard extends NoopAIGuard {
     }
     const { block = true } = opts ?? {}
     return this.#tracer.trace(AI_GUARD_RESOURCE, {}, async (span) => {
+      span = unwrap(span)
```
BridgeAR:
```suggestion
return this.#tracer.trace(AI_GUARD_RESOURCE, {}, async (publicSpan) => {
      const span = unwrap(publicSpan)
```

packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -293,8 +261,50 @@ class LLMObs extends NoopLLMObs {
     }
   }
 
+  _annotate (span, options) {
```
BridgeAR: Please use 
```suggestion
  #annotate (span, options) {
```

packages/dd-trace/src/opentracing/public/service-source.js
```diff
@@ -0,0 +1,38 @@
...
+ * when the caller supplied a user-controlled service name. Returns the input
+ * unchanged when there's nothing to mark.
+ *
+ * @template {{ service?: unknown, tags?: Record<string, unknown> } | null | undefined} T
+ * @param {T} options
+ * @returns {T}
+ */
+function markManualService (options) {
+  if (options == null) return options
+  if (!options.service && !tagsHaveManualService(options.tags)) return options
```
BridgeAR:
```suggestion
  if (options == null || !options.service && !tagsHaveManualService(options.tags)) return options
```

Nit

packages/dd-trace/test/opentelemetry/context_manager.spec.js
```diff
@@ -209,10 +228,8 @@ describe('OTel Context Manager', () => {
   })
 
   describe('with an active Datadog span', () => {
-    const ddTracer = require('../../')
-
     it('exposes the active span via trace.getActiveSpan() and forwards writes', () => {
-      ddTracer.trace('dd-active', (ddSpan) => {
+      withActiveDatadogSpan('dd-active', (ddSpan) => {
```
BridgeAR: I believe we should look into rewriting the tests to still verify their behavior but in a more robust way that does not touch the internals as such instead of hacking our way into returning the internal span.

pabloerhard: This all can also be rewritten to something like this:
```
      ddTracer.trace('dd-active', () => {
        const active = trace.getActiveSpan()
        const ddSpan = active._ddSpan
```
This removes the need of the helper, while it doesn't remove of accessing internals it's a cleaner solutions.

pabloerhard:
Closing in favor of #8557

PR #7943, opened 2026-04-07
feat(azure/cosmos): add Azure CosmosDB integration

rithikanarayan:
### What does this PR do?

Adds support for distributed tracing in certain Azure environments

Instruments the CosmosClient in the Azure CosmosDB SDK
Instruments the cosmosdb consumer triggers in the Azure Function SDK

New integration for Azure Cosmoss configurable by:
DD_TRACE_AZURE_COSMOS_ENABLED (default true) enables/disables the integration

### Motivation

Improving distributed tracing in Azure.


packages/datadog-plugin-azure-functions/test/integration-test/cosmosdb-test/cosmosdb.spec.js
```diff
@@ -0,0 +1,120 @@
...
+
+describe('esm', () => {
+  let agent
+  let proc
+
+  withVersions('azure-functions', '@azure/functions', version => {
+    useSandbox([
+      `@azure/functions@${version}`,
+      'azure-functions-core-tools@4',
+      '@azure/event-hubs@6.0.0',
```
pablomartinezbernardo, line 27: Do we need this?

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,160 @@
...
+    const resource = this.getResource(requestContext)
+    const { dbName, containerName } = this.getDbInfo(requestContext)
+    const connectionMode = this.getConnectionMode(requestContext)
+    const { outHost, userAgent } = this.getHttpInfo(requestContext)
+    const pluginOn = ctx.arguments?.[3]
+
+    // getting really specific here but otherwise we get doubled up read spans
+    // for the most part, only trace operations not requests (pluginOn)
+    // trace requests only if they are read or query operations not on docs
+    if (pluginOn === 'request' && ((!resource.includes('read') && !resource.includes('query')) ||
```
pablomartinezbernardo: Should we use `requestContext.operationType` instead of `resource.includes`?

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,160 @@
...
+    const requestContext = ctx.arguments?.[1]
+    const resource = this.getResource(requestContext)
+    const { dbName, containerName } = this.getDbInfo(requestContext)
+    const connectionMode = this.getConnectionMode(requestContext)
+    const { outHost, userAgent } = this.getHttpInfo(requestContext)
+    const pluginOn = ctx.arguments?.[3]
+
+    // getting really specific here but otherwise we get doubled up read spans
+    // for the most part, only trace operations not requests (pluginOn)
+    // trace requests only if they are read or query operations not on docs
```
pablomartinezbernardo, line 63: Why do we want to trace requests if they are read/query on something other than docs?

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,160 @@
...
+        }
+        if (result.substatus !== undefined) {
+          span.setTag('http.status_subcode', result.substatus)
+        }
+      }
+      span.finish()
+    }
+  }
+
+  error(ctx) {
```
pablomartinezbernardo, line 38: What happens if this is a `pluginOn === 'request'` that based on the skip logic in bindStart didn't create a span?

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,160 @@
...
+  // Shimmer + tracingChannel: static prefix = 'tracing:apm:<name>:<operation>'
+  // Shimmer + manual channels: omit prefix — defaults to `apm:${id}:${operation}`
+  static prefix = 'tracing:orchestrion:@azure/cosmos:executePlugins'
+  static peerServicePrecursors = ['db.name']
+
+  operationName() {
+    return 'cosmosdb.query'
+  }
+
+  asyncEnd(ctx) {
```
pablomartinezbernardo, line 22: Similar comment to https://github.com/DataDog/dd-trace-js/pull/7943/changes#r3110691960

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,160 @@
+'use strict'
+
+const { storage } = require('../../datadog-core')
+
+const DatabasePlugin = require('../../dd-trace/src/plugins/database')
+
+class AzureCosmosPlugin extends DatabasePlugin {
```
pablomartinezbernardo, line 7: Is this base class appropriate? Looking at it it looks like it's related to dbm

pablomartinezbernardo:
Are snapshot tests expected for new integrations? I was surprised to not see any, but maybe that's how node does things

rithikanarayan:
> Are snapshot tests expected for new integrations? I was surprised to not see any, but maybe that's how node does things

No, it doesn't seem that node has snapshot tests, at least not following https://github.com/DataDog/dd-trace-js/blob/master/.agents/skills/apm-integrations/references/testing.md

packages/datadog-plugin-azure-cosmos/src/index.js
rithikanarayan: This is mostly to take into account the `createIfNotExists` functions that exist for databases and containers. Without the additional exception, if a customer uses the `createIfNotExists` function they would see a [double read span ](https://ddserverless.datadoghq.com/apm/traces?query=env%3Arithika-sadge&agg_m=count&agg_m_source=base&agg_t=count&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=false&graphType=flamegraph&historicalData=false&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=desc&spanID=5666629422366022984&spanType=all&storage=hot&timeHint=1776695573403&trace=56666294223660229845666629422366022984&traceID=5666629422366022984&traceQuery=&view=spans&start=1776695401389&end=1776696301389&paused=false)for accessing their existing databases and containers. Doesn't seem to affect other operations.

packages/datadog-plugin-azure-cosmos/src/index.js
rithikanarayan: I think that the check for the existence of the span in lines 24 and 40 should address that? Not certain though.

packages/datadog-plugin-azure-cosmos/src/index.js
rithikanarayan: I suppose I could use the generic `TracingPlugin` class but based on [the guide](https://github.com/DataDog/dd-trace-js/blob/master/.agents/skills/apm-integrations/references/new-integration-guide.md#choosing-the-right-base-class), it seems that this is the correct class to use `db.*` tags with.

packages/datadog-plugin-azure-functions/test/integration-test/cosmosdb-test/cosmosdb.spec.js
rithikanarayan: It is used in testing the [eventhubs integration](https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-plugin-azure-functions/test/integration-test/eventhubs-test/eventhubs.spec.js#L22) and [the servicebus integration](https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-plugin-azure-functions/test/integration-test/servicebus-test/servicebus.spec.js#L22) so I used it here as well.

packages/datadog-plugin-azure-cosmos/src/index.js
rithikanarayan: Oops I misread the question. We have that middle clause to avoid traces[ like this](https://ddserverless.datadoghq.com/apm/traces?query=env%3Arithika-sadge&agg_m=count&agg_m_source=base&agg_t=count&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=false&graphType=flamegraph&historicalData=false&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=desc&spanID=6537361663688705071&spanType=all&storage=driveline&timeHint=1776708795334&trace=AQAAAZ2sGLvG9GwELwAAAAA2NTM3MzYxNjYzNjg4NzA1MDcx&traceID=4376277917371952657&traceQuery=&view=spans&start=1776708033386&end=1776708933386&paused=false) where there are doubled upsert and delete spans.

pablomartinezbernardo: Yeah, my question could have been phrased better, in other words: Do we ever want to trace `request` calls at all? What would we be losing if the condition was just `if (pluginOn === 'request') return ctx.currentStore`?

packages/datadog-plugin-azure-cosmos/src/index.js
pablomartinezbernardo: Let's leave this open, see if someone more experienced in this tracer can confirm if this is an issue.

But either way, the `error` function is doing `this.addError(ctx.error, span)` unconditionally

packages/datadog-plugin-azure-cosmos/src/index.js
pablomartinezbernardo: You're right, I initially saw [the class](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/plugins/database.js) that contains only dbm-related stuff, but doesn't really impact anything if you don't use them

packages/datadog-plugin-azure-cosmos/src/index.js
rithikanarayan: True, moving that within the null check 🫡

packages/datadog-plugin-azure-cosmos/src/index.js
rithikanarayan: If we just use the condition `pluginOn === 'request`, we get traces [like this](https://ddserverless.datadoghq.com/apm/traces?query=env%3Arithika-sadge&agg_m=count&agg_m_source=base&agg_t=count&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=false&graphType=flamegraph&historicalData=false&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=desc&spanID=1245959518430818318&spanType=all&storage=driveline&timeHint=1776779167809&trace=AQAAAZ2wSohBpJXwDgAAAAAxMjQ1OTU5NTE4NDMwODE4MzE4&traceID=4879596388093163269&traceQuery=&view=spans&start=1776778283901&end=1776779183901&paused=false), where there are no spans for read or query operations on the items.

packages/datadog-plugin-azure-cosmos/test/index.spec.js
```diff
@@ -0,0 +1,125 @@
...
+          const span = traces[0][0]
+
+          assert.strictEqual(span.name, 'cosmosdb.query')
+          assert.strictEqual(span.service, 'test-azure-cosmos')
+          assert.strictEqual(span.type, 'cosmosdb')
+          assert.strictEqual(span.meta.component, 'azure_cosmos')
+          assert.strictEqual(span.meta['db.system'], 'cosmosdb')
+          assert.strictEqual(span.meta['db.name'], 'testDatabase')
+          assert.strictEqual(span.meta['cosmosdb.container'], 'testContainer')
+          assert.strictEqual(span.meta['cosmosdb.connection.mode'], 'gateway')
```
pabloerhard: This could be simplified to something like the following:
```suggestion
  assertObjectContains(span, {
    name: 'cosmosdb.query',
    service: 'test-azure-cosmos',
    type: 'cosmosdb',
    meta: {
      component: 'azure_cosmos',
      'db.system': 'cosmosdb',
      'db.name': 'testDatabase',
      'cosmosdb.container': 'testContainer',
      'cosmosdb.connection.mode': 'gateway'
    }
  })
```

docs/test.ts
```diff
@@ -57,10 +57,10 @@ tracer.init({
...
   hostname: 'agent',
   logger: {
-    error (message: string | Error) {},
-    warn (message: string) {},
-    info (message: string) {},
-    debug (message: string) {}
+    error(message: string | Error) { },
+    warn(message: string) { },
+    info(message: string) { },
+    debug(message: string) { }
```
BridgeAR, line 58: Can you please undo the changes that are not needed? I guess you have a formatter installed that is not using the repository rules? :)

rithikanarayan: I have tried to use multiple linters (running `npm run preprare; npm run lint:fix` and `./node_modules/.bin/eslint --fix <filename>`) but neither have worked on `docs/test.ts` and `index.d.ts`. Do you know of any other ways to lint? If I manually undo the changes and try to save the file, it automatically reverts to the state it's in now.

packages/datadog-plugin-azure-cosmos/test/integration-test/client.spec.js
```diff
@@ -0,0 +1,59 @@
...
+      agent = await new FakeAgent().start()
+      spawnEnv = { NODE_OPTIONS: '--experimental-global-webcrypto' }
+    })
+
+    afterEach(async () => {
+      await stopProc(proc)
+      await agent.stop()
+    })
+
+    for (const variant of ['star', 'destructure']) {
```
BridgeAR, line 40: Is there no default export? :)

packages/datadog-plugin-azure-cosmos/test/index.spec.js
```diff
@@ -0,0 +1,134 @@
...
+                'db.name': 'testDatabase',
+                'cosmosdb.container': 'testContainer',
+                'cosmosdb.connection.mode': 'gateway',
+              },
+            })
+
+            assert.strictEqual(span.error, 1)
+            assert.strictEqual(span.meta['error.message'], 'The document already exists in the collection.')
+            assert.strictEqual(span.meta['error.type'], 'Error')
+            assert.strictEqual(span.meta['http.status_code'], '409')
```
BridgeAR: Using `assertObjectContains` would be nice for these four lines

packages/datadog-plugin-azure-functions/src/index.js
```diff
@@ -127,6 +128,11 @@ function getMetaForTrigger ({ functionName, methodName, invocationContext }) {
       'resource.name': `EventHubs ${functionName}`,
       'span.kind': 'consumer',
     }
+  } else if (triggerMap[methodName] === 'CosmosDB') {
+    meta = {
+      ...meta,
+      'resource.name': `CosmosDB ${functionName}`,
+    }
```
BridgeAR:
```suggestion
    meta['resource.name'] = `CosmosDB ${functionName}`
```

packages/datadog-plugin-azure-functions/test/integration-test/cosmosdb-test/cosmosdb.spec.js
```diff
@@ -0,0 +1,140 @@
...
+  let proc
+  let setup
+  let teardown
+
+  withVersions('azure-functions', '@azure/functions', version => {
+    useSandbox([
+      `@azure/functions@${version}`,
+      'azure-functions-core-tools@4',
+      '@azure/event-hubs@6.0.0',
+      '@azure/cosmos@4.9.2',
```
BridgeAR: Could we keep the version open ended for these two?

docker-compose.yml
```diff
@@ -6,9 +6,19 @@ services:
   azurite:
     image: mcr.microsoft.com/azure-storage/azurite:3.35.0
     ports:
-      - "127.0.0.1:10000:10000"
-      - "127.0.0.1:10001:10001"
-      - "127.0.0.1:10002:10002"
+        - "127.0.0.1:10000:10000"
```
BridgeAR: The former indentation seems to be the one used in other spots :)

index.d.ts
```diff
@@ -3738,7 +3745,7 @@ declare namespace tracer {
        * @param key The key of the tag to get.
        * @returns The value of the tag, or `undefined` if the tag does not exist.
        */
-      getTag (key: string): string | undefined
+      getTag(key: string): string | undefined
```
BridgeAR: Please undo all formatter related changes

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,167 @@
...
+  getHttpInfo (requestContext) {
+    let outHost = null
+    let userAgent = null
+    if (requestContext != null) {
+      outHost = requestContext.client?.cosmosClientOptions?.endpoint
+      const headers = requestContext.headers
+      if (headers != null) {
+        userAgent = headers['User-Agent']
+      }
+    }
```
BridgeAR:
```suggestion
    const outHost = requestContext?.client?.cosmosClientOptions?.endpoint
    const userAgent = requestContext?.headers?.['User-Agent']
```

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,167 @@
...
+    if (requestContext != null) {
+      const mode = requestContext.client?.connectionPolicy?.connectionMode
+      if (mode === 0) {
+        return 'gateway'
+      } else if (mode === 1) {
+        return 'direct'
+      }
+      return 'other'
+    }
+    return null
```
BridgeAR: Nit

```suggestion
    if (!requestContext) {
      return null
    }
    const mode = requestContext.client?.connectionPolicy?.connectionMode
    if (mode === 0) {
      return 'gateway'
    } else if (mode === 1) {
      return 'direct'
    }
    return 'other'
```

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,167 @@
...
+    if (span != null) {
+      const result = ctx.result
+      if (result != null) {
+        if (result.code != null) {
+          span.setTag('http.status_code', result.code)
+        }
+        if (result.substatus !== undefined) {
+          span.setTag('http.status_subcode', result.substatus)
+        }
+      }
```
BridgeAR: Nit

```suggestion
      if (result?.code != null) {
        span.setTag('http.status_code', result.code)
      }
      if (result?.substatus != null) {
        span.setTag('http.status_subcode', result.substatus)
      }
```

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,167 @@
...
+      this.addError(ctx.error, span)
+      const error = ctx.error
+      if (error != null) {
+        if (error.code != null) {
+          span.setTag('http.status_code', error.code)
+        }
+        if (error.substatus !== undefined) {
+          span.setTag('http.status_subcode', error.substatus)
+        }
+      }
```
BridgeAR: Nit

```suggestion
      if (error?.code != null) {
        span.setTag('http.status_code', error.code)
      }
      if (error?.substatus != null) {
        span.setTag('http.status_subcode', error.substatus)
      }
```

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,167 @@
...
+    return ctx.currentStore
+  }
+
+  getResource (requestContext) {
+    if (requestContext != null) {
+      const operationType = requestContext.operationType
+      const resourceLink = requestContext.path
+      return operationType + ' ' + resourceLink
+    }
+    return null
```
BridgeAR: Nit


```suggestion
    return requestContext
      ? `${requestContext.operationType} ${requestContext.path}`
      : null
```

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,167 @@
...
+
+    // only trace operations not requests (pluginOn)
+    // trace requests only if they are read or query operations not on docs
+    // prevents doubled read spans for createIfNotExists calls
+    if (pluginOn != null && requestContext.operationType != null && requestContext.resourceType != null) {
+      const operationType = requestContext.operationType
+      const resourceType = requestContext.resourceType
+      if (pluginOn === 'request' && ((operationType !== 'read' && operationType !== 'query') ||
+        (operationType === 'read' && resourceType !== 'docs'))) {
+        ctx.currentStore = { ...storage('legacy').getStore() }
```
BridgeAR: This creates a new object here, which is confusing for me.

@rochdev PTAL

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,167 @@
...
+        }
+        const parts = resourceLink.split('/')
+        if (parts.length > 0 && parts[0].toLowerCase() === 'dbs' && parts.length >= 2) {
+          dbName = parts[1]
+          if (parts.length >= 4 && parts[2].toLowerCase() === 'colls' && parts[3].toLowerCase() !== '') {
+            containerName = parts[3]
+          }
+        }
+      }
+    }
```
BridgeAR: This could be further optimized by directly checking the string, but I guess it's alright for now.

```suggestion
      if (resourceLink?.length > 1 && resourceLink.startsWith('/')) {
        resourceLink = resourceLink.slice(1)
        const parts = resourceLink.split('/')
        if (parts.length > 0 && parts[0].toLowerCase() === 'dbs' && parts.length >= 2) {
          dbName = parts[1]
          if (parts.length >= 4 && parts[2].toLowerCase() === 'colls' && parts[3] !== '') {
            containerName = parts[3]
          }
        }
      }
```

packages/datadog-plugin-azure-cosmos/src/index.js
pabloerhard: I guess this should be changed to something like the following:
```
 const store = storage('legacy').getStore() || {}
 ctx.currentStore = { ...store, span }
```
This is on par with most instrumentations and covers the case in which storage('legacy').getStore() returns undefined.

packages/datadog-plugin-azure-cosmos/test/integration-test/client.spec.js
rithikanarayan: I believe that the @azure/cosmos package only includes named exports, no default exports.

.github/workflows/serverless.yml
```diff
@@ -169,6 +169,7 @@ jobs:
           - eventhubs
           - client
           - servicebus
+          - cosmosdb
```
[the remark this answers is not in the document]
rithikanarayan: The other emulators are not waited on, so I believe this should be ok.

packages/datadog-plugin-azure-cosmos/src/index.js
rithikanarayan: How would this code snippet work when the span is not defined yet (second line)?

pabloerhard: Same way, just don't send through the span (span is only there because I took this example from another plugin)

```
const store = storage('legacy').getStore() || {}
 ctx.currentStore = { ...store  }
```

Edit: This is purely visual, doing the spread op on undefined does not throw

packages/datadog-plugin-bullmq/test/index.spec.js
```diff
@@ -172,7 +172,7 @@ createIntegrationTestSuite('bullmq', 'bullmq', {
           'messaging.system': 'bullmq',
           'error.type': 'TypeError',
           'error.message': 'Converting circular structure to JSON\n    --> ' +
-           "starting at object with constructor 'Object'\n    --- property 'self' closes the circle",
+            "starting at object with constructor 'Object'\n    --- property 'self' closes the circle",
```
rithikanarayan: I will relint this along with the next round of review feedback

docs/test.ts
BridgeAR: We do not use a formatter, so the formatting happens due to the editor defaults you have set :)
The auto formatting has to be deactivated in your editor.

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,153 @@
...
+    if (span != null) {
+      const result = ctx.result
+      if (result != null) {
+        if (result.code != null) {
+          span.setTag('http.status_code', result.code)
+        }
+        if (result.substatus !== undefined) {
+          span.setTag('http.status_subcode', result.substatus)
+        }
+      }
```
pabloerhard: Nit: We could simplify this to something like the following:

```suggestion
    if (span) {
      const result = ctx.result
        if (result?.code) span.setTag('http.status_code', result.code)
        if (result?.substatus) span.setTag('http.status_subcode', result.substatus)
      }
```

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,153 @@
...
+    const span = ctx.currentStore?.span
+    if (span != null) {
+      this.addError(ctx.error, span)
+      const error = ctx.error
+      if (error?.code != null) {
+        span.setTag('http.status_code', error.code)
+      }
+      if (error?.substatus != null) {
+        span.setTag('http.status_subcode', error.substatus)
+      }
```
pabloerhard: Same here
```suggestion
      const error = ctx.error
      if (error?.code) span.setTag('http.status_code', error.code)
      if (error?.substatus) span.setTag('http.status_subcode', error.substatus)
```

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,153 @@
...
+        'out.host': outHost,
+      },
+    }, ctx)
+
+    ctx.span = span
+    return ctx.currentStore
+  }
+
+  getResource (requestContext) {
+    return requestContext
```
pabloerhard: Correct me if I’m wrong, but isn’t requestContext guaranteed to exist? If so couldn’t many of the requestContext conditionals be simplified?

rithikanarayan: The checks were included in case the context or function arguments were somehow malformed. I checked some other packages and it looks like if the argument is being used extensively that the null checks are appropriate (ex. /packages/dd-trace/src/llmobs/plugins/modelcontextprotocol-sdk/index.js). Open to removing it if you don't think this should be a concern :)

packages/datadog-plugin-azure-cosmos/src/index.js
rithikanarayan: Given that the suggestion above is visual, is there any other suggested change to make here? Or is it acceptable as is.

packages/datadog-plugin-azure-cosmos/src/index.js
pabloerhard: We could move the check to bindStart and this would avoid adding doing the same check in getResource, getDbInfo, getConnectionMode and getHttpInfo

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,147 @@
...
+      if (error?.substatus) span.setTag('http.status_subcode', error.substatus)
+    }
+  }
+
+  bindStart (ctx) {
+    const requestContext = ctx.arguments?.[1]
+    const resource = requestContext ? this.getResource(requestContext) : null
+    const { dbName, containerName } = requestContext ? this.getDbInfo(requestContext) : null
+    const connectionMode = requestContext ? this.getConnectionMode(requestContext) : null
+    const { outHost, userAgent } = requestContext ? this.getHttpInfo(requestContext) : null
```
pabloerhard: I think the nature of the function being instrumented (executePlugins) allows for further simplifications, for example most arguments are guaranteed to exist (if not all) as they are built internally by the sdk. 

This will not execute unless that internal function is called so we could further simplify to something like the following:

```suggestion
    const requestContext = ctx.arguments[1]
    const resource = his.getResource(requestContext)
    const { dbName, containerName } = this.getDbInfo(requestContext)
    const connectionMode = this.getConnectionMode(requestContext)
    const { outHost, userAgent } = this.getHttpInfo(requestContext)
```

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,147 @@
...
+
+    ctx.span = span
+    return ctx.currentStore
+  }
+
+  getResource (requestContext) {
+    const path = requestContext.path
+    const parts = path.split('/')
+    let modified = false
+    for (let i = 0; i < parts.length; i += 2) {
```
pablomartinezbernardo:
```suggestion
    for (let i = 1; i < parts.length; i += 2) {
```

I think this should be like this, but may be worth testing the function

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,147 @@
...
+  asyncEnd (ctx) {
+    if (!ctx.span) return
+    const span = ctx.currentStore?.span
+    if (span) {
+      const result = ctx.result
+      if (result?.code) span.setTag('http.status_code', result.code)
+      if (result?.substatus) span.setTag('http.status_subcode', result.substatus)
+    }
+    span.finish()
+  }
```
pablomartinezbernardo, line 31:
```suggestion
  asyncEnd (ctx) {
    if (!ctx.span) return
    const span = ctx.currentStore?.span
    if (span) {
      const result = ctx.result
      if (result?.code) span.setTag('http.status_code', result.code)
      if (result?.substatus) span.setTag('http.status_subcode', result.substatus)
      span.finish()
    }
  }
```

packages/datadog-plugin-azure-cosmos/src/index.js
rithikanarayan: Actually had to make it `i = 2`, the split method leaves a `''` as the first element of the parts array so there's an extra element compared to what we expected.

pablomartinezbernardo (dismissed):
I would love to see a couple of unit tests for `getResource`, otherwise LGTM from the serverless point of view

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,147 @@
...
+  operationName () {
+    return 'cosmosdb.query'
+  }
+
+  asyncEnd (ctx) {
+    if (!ctx.span) return
+    const span = ctx.currentStore?.span
+    if (span) {
+      const result = ctx.result
+      if (result?.code) span.setTag('http.status_code', result.code)
```
rochdev: Are those actually related to HTTP?

packages/datadog-plugin-azure-cosmos/src/index.js
rochdev: In the case where we don't create a span, why even create a new store?

packages/datadog-plugin-azure-cosmos/test/get-resource.spec.js
```diff
@@ -0,0 +1,49 @@
...
+const assert = require('node:assert/strict')
+
+const AzureCosmosPlugin = require('../src')
+
+describe('azure-cosmos', () => {
+  describe('getResource', () => {
+    let plugin
+
+    before(() => {
+      plugin = new AzureCosmosPlugin({}, {})
```
rochdev, line 12: Why is this tested this way? We don't typically have unit tests for plugins, we instead test real scenarios with the entire library.

docker-compose.yml
```diff
@@ -9,6 +9,16 @@ services:
       - "127.0.0.1:10000:10000"
       - "127.0.0.1:10001:10001"
       - "127.0.0.1:10002:10002"
+  azurecosmosemulator:
+    image: mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:vnext-preview
```
rochdev: Using MCR is now forbidden, see https://github.com/DataDog/dd-trace-js/pull/8215. You can use the `Mirror image to GHCR` action to re-publish it on our mirror.

packages/datadog-plugin-azure-cosmos/src/index.js
pablomartinezbernardo: They actually are in the success case, but not in the error case. That being said

The dotnet implementation is actually using https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-response-status-code 

And the sub status code should not be part of the http "group", it is instead cosmos-specific which dotnet sets in `cosmosdb.response.sub_status_code`

packages/datadog-plugin-azure-cosmos/test/get-resource.spec.js
rithikanarayan: @pablomartinezbernardo requested unit tests for the `getResource` function, so I added them in addition to the plugin integration tests.

packages/datadog-plugin-azure-cosmos/src/index.js
rithikanarayan: Fixed the substatus, thanks for pointing that out! For the error case, there weren't uses of `db.response.status_code` in the Python or Node tracers so I just used http.status_code for those.

rithikanarayan: Also going to fix the status code to be `db.response.status_code` 🙏

packages/datadog-plugin-azure-functions/test/integration-test/cosmosdb-test/cosmosdb.spec.js
```diff
@@ -0,0 +1,139 @@
...
+  hookFile,
+  sandboxCwd,
+  useSandbox,
+  curlAndAssertMessage,
+  assertObjectContains,
+  stopProc,
+} = require('../../../../../integration-tests/helpers')
+const { withVersions } = require('../../../../dd-trace/test/setup/mocha')
+
+describe('esm', () => {
```
rochdev, line 19: This is typically a test file next to the plugin, why have it in integration tests for CosmosDB? Basically, plugin tests are already pretty close to integration tests, with the only difference that they don't use a sandbox, so they're much simpler.

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,146 @@
...
+        (operationType === 'read' && resourceType !== 'docs'))) {
+        return storage('legacy').getStore()
+      }
+
+      // separately, skip tracing read requests without a path, these don't
+      // represent CRUD operations on a resource we care about
+      // not returning current store because we don't want the child http.request spans
+      // to be created
+      if (operationType === 'read' && requestContext.path === '') {
+        return
```
rochdev: This should return the active store. Returning no store will effectively create a new root store, so child spans will still be created but orphaned. If the goal is to completely disable tracing for the scope of the caller, then you can return `{ noop: true }` which will activate a store that tells the tracer to become inert for that scope. I would typically discourage this though as that would hide things that the process is doing.  For the most part, `noop` was added so that we don't trace ourselves, it's not really meant to hide what libraries are doing.

Since tests are passing, it also seems like a test is missing because this should break.

rochdev (changes requested):

packages/datadog-plugin-azure-functions/test/integration-test/cosmosdb-test/cosmosdb.spec.js
rithikanarayan: This is meant to test that an Azure Function triggered by a CosmosDB update is properly traced. It mirrors the existing tests under the `integration-test/eventhubs-test`, `integration-test/http-test`, and `integration-test/servicebus-test` folders, which test that those three types of Azure Functions triggers are also properly traced. The tests use sandboxes to spin up an Azure Function App and the relevant Azure managed service.

packages/datadog-plugin-azure-cosmos/src/index.js
rithikanarayan: Updated to return `{ noop : true }` and added a test `does not create cosmosdb or http spans for empty-path read requests` checking this logic.

pabloerhard (commented):
Note: We might want to ensure tests pass with Node 26 before merging this, as we have ongoing work to fully support Node 26 and some azure integrations fail when using Node 26

Note: I've tested running azure-cosmos with Node 26 and it doesn't fail.

packages/datadog-plugin-azure-cosmos/src/index.js
```diff
@@ -0,0 +1,144 @@
...
+  // Shimmer + tracingChannel: static prefix = 'tracing:apm:<name>:<operation>'
+  // Shimmer + manual channels: omit prefix — defaults to `apm:${id}:${operation}`
+  static prefix = 'tracing:orchestrion:@azure/cosmos:executePlugins'
+  static peerServicePrecursors = ['db.name']
+
+  operationName () {
+    return 'cosmosdb.query'
+  }
+
+  asyncEnd (ctx) {
```
rochdev, line 22: Why does this contain 2 checks for `span` in 2 different places?

rochdev (approved):
Just one question for code that I think could be simplified of better explained, but that's non-blocking and otherwise LGTM!

packages/datadog-plugin-azure-cosmos/src/index.js
rithikanarayan: I kept both checks because the first one is for `ctx.span` and the second is for `ctx.currentStore.span` which at the time I was not sure were the same. Happy to update this in a follow up PR if the second check is unneccessary.

PR #7974, opened 2026-04-10
feat(claude-agent-sdk): add instrumentation for @anthropic-ai/claude-agent-sdk

mr-lee:
### What does this PR do?

Adds support for tracing top-level query, tool, LLM, and subagent calls from the Claude Agent SDK.

This is accomplished by parsing the event stream into tracing-lifecycle events after the fact.

### Motivation

Provide support for the Claude Agent SDK package.


benchmark/sirun/plugin-claude-agent-sdk/index.js
sabrenner, line 1: i think ideally we would test against the real package directly (pointing it to the vcr directory so we can re-use recorded request-responses). the usual pattern is to not export things from instrumentation files and instead test through using the package its instrumenting.

i do like having the benchmark here though!

packages/datadog-instrumentations/src/claude-agent-sdk.js
```diff
@@ -0,0 +1,234 @@
+'use strict'
+
+// Shimmer is used here because we must merge user-provided Claude hook matchers
+// into `options.hooks` at runtime before calling `query()`, which orchestrion
+// cannot express without custom argument mutation logic.
```
sabrenner: i think we can still use orchestrion - we have other examples of [argument mutation from the vercel ai sdk plugin](https://github.com/DataDog/dd-trace-js/blob/b4740a3636892ef7c0308d1ddbc1bc9c3215afa0/packages/datadog-instrumentations/src/ai.js#L182-L193). i think as we're trying to move the repo to use orchestrion-js, we can probably implement practices like this!

packages/datadog-instrumentations/src/claude-agent-sdk.js
```diff
@@ -0,0 +1,234 @@
...
+  name: '@anthropic-ai/claude-agent-sdk',
+  file: 'sdk.mjs',
+  versions: ['>=0.2.0'],
+}, (exports) => {
+  shimmer.wrap(exports, 'query', wrapQuery)
+  return exports
+})
+
+// Exported for unit testing
+module.exports = { mergeHooks, buildTracerHooks, wrapQuery }
```
sabrenner: tests should ideally rely on using the package directly. we (should 🙏) be able to set the base url (either programatically, or maybe via env var?) to `localhost:9126/vcr/anthropic` to record the requests and responses in each test locally so they can be served back in CI and subsequent test runs

packages/datadog-plugin-claude-agent-sdk/src/tracing.js
```diff
@@ -0,0 +1,122 @@
...
+    span?.finish()
+  }
+
+  static computeSpanName () {
+    return 'claude-agent-sdk'
+  }
+
+  static extractTags () {
+    return null
+  }
```
sabrenner: we don't need to/shouldn't set any tags on APM spans. we can get rid of this

packages/datadog-plugin-claude-agent-sdk/test/index.spec.js
sabrenner, line 1: this should be testing against the actual package, using the vcr capabilities of the test agent so we can do real queries one time and then re-use them in CI.

packages/dd-trace/src/llmobs/plugins/claude-agent-sdk/index.js
```diff
@@ -0,0 +1,164 @@
...
+  if (typeof value === 'string') return value
+  try { return JSON.stringify(value) } catch { return '[unserializable]' }
+}
+
+function splitModel (model) {
+  if (!model) return { modelName: undefined, modelProvider: 'anthropic' }
+  const idx = model.indexOf('/')
+  if (idx === -1) return { modelName: model, modelProvider: 'anthropic' }
+  return { modelName: model.slice(idx + 1), modelProvider: model.slice(0, idx) }
+}
```
sabrenner: we can move these to a separate utils file

packages/dd-trace/test/llmobs/plugins/claude-agent-sdk/index.spec.js
sabrenner, line 1: similar comment about testing using the actual library

packages/dd-trace/test/llmobs/plugins/claude-agent-sdk/index.spec.js
```diff
@@ -0,0 +1,445 @@
...
+const toolCh = tracingChannel('apm:claude-agent-sdk:tool')
+const subagentCh = tracingChannel('apm:claude-agent-sdk:subagent')
+
+describe('Plugin', () => {
+  const { getEvents } = useLlmObs({ plugin: 'claude-agent-sdk' })
+
+  // The Claude Agent SDK is pure ESM and can't be CJS-required in tests.
+  // Simulate the module load event so the plugin manager activates the
+  // plugin's channel subscriptions (same event register.js publishes
+  // when a real SDK module is loaded and version-matched).
```
sabrenner: i am able to `require` it locally, so i think we should be OK to do it in tests

sabrenner (commented):
high-level comments, two main callouts:
1. we should still be able to use orchestrion, even when doing argument mutations. orchestrion is great for tracing the top-level things like `query` directly with lifecycle events, but otherwise we can intercept arguments and patch them there. ideally, we could also patch the code where hooks are called and do a tracing channel on that function.
2. we _should_ be able to actually test against the library vs using hooks & mocking calls to the library. i was able to `require` it in a CJS file, and even if it doesn't work here, we can write some tests in our `integration-test` suite to run ESM files.
3. looks like there are a couple of `debugger` changes in this PR that are unrelated and are probs just from a bad rebase. for some reason i couldnt comment on them directly

i think these points will change a decent chunk of the code, so i'll do a further in-depth review after!

packages/datadog-plugin-claude-agent-sdk/src/tracing.js
mr-lee: Done — stripped all custom APM tags. Tracing plugin now matches the anthropic pattern: just span name + resource.name, no custom tags. All meaningful data lives in LLM Obs metadata.

packages/dd-trace/src/llmobs/plugins/claude-agent-sdk/index.js
mr-lee: Done — extracted `safeStringify`, `splitModel`, `SOURCE_TO_TRIGGER` to `utils.js`.

benchmark/sirun/plugin-claude-agent-sdk/index.js
mr-lee: The SDK's default entry point (`sdk.mjs`) spawns a Claude CLI child process via `child_process.spawn()` for `query()` — there are no HTTP calls to VCR. The subprocess communicates over stdin/stdout JSON streams, not HTTP.

The `bridge` and `assistant` entry points do use `globalThis.fetch` to the Anthropic API, but they have a completely different API surface (`createCodeSession`, `runAssistantWorker`) — not `query()`.

For the default `query()` path, channel-based testing is the right approach since there's no HTTP to record. Happy to add VCR tests for the bridge/assistant entry points as a follow-up if we want to cover those paths too.

Re: `require` — you're right it's requireable, but the `query()` function needs a running Claude CLI binary to do anything useful (it spawns it as a child process). So even with a real require, we can't make meaningful API calls without that binary.

packages/datadog-instrumentations/src/claude-agent-sdk.js
mr-lee: Same reasoning as above — `query()` uses subprocess transport, not HTTP. No HTTP calls to VCR.

packages/datadog-plugin-claude-agent-sdk/test/index.spec.js
mr-lee: Same as above.

packages/dd-trace/test/llmobs/plugins/claude-agent-sdk/index.spec.js
mr-lee: Same as above.

packages/dd-trace/test/llmobs/plugins/claude-agent-sdk/index.spec.js
mr-lee: You're right it's requireable — but `query()` needs a running Claude CLI binary (it spawns it via `child_process.spawn()`). So even with a real require, we can't make meaningful API calls in tests without that binary present. Channel-based testing lets us exercise the full hook lifecycle without the subprocess dependency.

packages/datadog-instrumentations/src/claude-agent-sdk.js
mr-lee: Done — migrated to orchestrion. Added `packages/datadog-instrumentations/src/helpers/rewriter/instrumentations/claude-agent-sdk.js` with `isExportAlias: true` to handle the minified ESM bundle (`export { Ns as query }`). The shimmer `addHook` path is kept as a fallback for environments where the rewriter is not active. The orchestrion channel subscribes to `orchestrion:@anthropic-ai/claude-agent-sdk:query` and mutates `ctx.arguments` to inject tracer hooks.

packages/datadog-plugin-claude-agent-sdk/test/index.spec.js
```diff
@@ -0,0 +1,268 @@
+'use strict'
+
+const assert = require('node:assert')
+const { describe, before, after, it } = require('mocha')
+const { tracingChannel, channel } = require('dc-polyfill')
```
mr-lee, line 5: Tried setting ANTHROPIC_BASE_URL=http://127.0.0.1:19126 and running query() — the subprocess spawns and returns real messages, but zero HTTP requests hit the proxy. The CLI binary (cli.js) reads ANTHROPIC_BASE_URL but has a host validation (function cH) that checks if the host is api.anthropic.com. Non-standard hosts get filtered. How did you get it working locally? Were you using the bridge entry point or the default query() path?

sabrenner: you can also tried setting `CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST: true` and see if that helps, that's what i also had to include locally as otherwise i ran into what you're describing

mr-lee, line 5: That worked! CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST=true makes the subprocess send HTTP through the configured ANTHROPIC_BASE_URL. Proxy captured POST /v1/messages?beta=true requests.

One issue for cassette recording: the CLI uses Vault OIDC auth (not raw ANTHROPIC_API_KEY), so the requests 401 when proxied. Would be great to pair on getting a successful recording for the cassette — do you have a working cassette from your local testing?

packages/datadog-instrumentations/src/claude-agent-sdk.js
```diff
@@ -0,0 +1,322 @@
...
+const sessionCh = tracingChannel('apm:claude-agent-sdk:session')
+const turnCh = tracingChannel('apm:claude-agent-sdk:turn')
+const toolCh = tracingChannel('apm:claude-agent-sdk:tool')
+const subagentCh = tracingChannel('apm:claude-agent-sdk:subagent')
+
+function mergeHooks (userHooks, tracerHooks) {
+  const merged = {}
+
+  for (const event of Object.keys(tracerHooks)) {
+    const userMatchers = (userHooks && userHooks[event]) || []
```
sabrenner:
```suggestion
    const userMatchers = userHooks?.[event] || []
```

small nit

packages/datadog-instrumentations/src/claude-agent-sdk.js
```diff
@@ -0,0 +1,262 @@
...
+
+const turnCh = tracingChannel('apm:claude-agent-sdk:turn')
+const toolCh = tracingChannel('apm:claude-agent-sdk:tool')
+const subagentCh = tracingChannel('apm:claude-agent-sdk:subagent')
+
+const mergeHooks = function mergeHooks (userHooks, tracerHooks) {
+  const merged = {}
+
+  for (const event of Object.keys(tracerHooks)) {
+    const userMatchers = (userHooks && userHooks[event]) || []
```
sabrenner:
```suggestion
    const userMatchers = userHooks?.[event] || []
```

packages/datadog-instrumentations/src/claude-agent-sdk.js
```diff
@@ -0,0 +1,262 @@
...
+    toolCh.asyncEnd.publish(toolCtx)
+  }
+  sessionCtx.pendingTools.clear()
+  for (const subCtx of sessionCtx.pendingSubagents.values()) {
+    subagentCh.asyncEnd.publish(subCtx)
+  }
+  sessionCtx.pendingSubagents.clear()
+}
+
+const onQueryStart = function onQueryStart (ctx) {
```
sabrenner: likewise
```suggestion
function onQueryStart (ctx) {
```

packages/datadog-instrumentations/src/claude-agent-sdk.js
```diff
@@ -0,0 +1,262 @@
...
+      source: sessionCtx.source,
+      cwd: sessionCtx.cwd,
+      agentType: sessionCtx.agentType,
+      permissionMode: sessionCtx.permissionMode,
+      transcriptPath: sessionCtx.transcriptPath,
+    }
+
+    sessionCtx.currentTurn = turnCtx
+    turnCh.start.runStores(turnCtx, () => {
+      turnCh.end.publish(turnCtx)
```
sabrenner: move this to the `onStop` hook, otherwise we'll end the context scope too early

packages/datadog-instrumentations/src/claude-agent-sdk.js
```diff
@@ -0,0 +1,262 @@
...
+    const toolCtx = {
+      sessionId: input.session_id,
+      toolName: input.tool_name,
+      toolInput: input.tool_input,
+      toolUseId: id,
+    }
+
+    sessionCtx.pendingTools.set(id, toolCtx)
+    toolCh.start.runStores(toolCtx, () => {
+      toolCh.end.publish(toolCtx)
```
sabrenner: move this to the `onPostToolUse` hook, similarly

packages/datadog-instrumentations/src/claude-agent-sdk.js
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  const onPostToolUse = function onPostToolUse (input, toolUseId) {
+    const id = toolUseId || input.tool_use_id
+    const toolCtx = sessionCtx.pendingTools.get(id)
+    if (toolCtx) {
+      toolCtx.toolResponse = input.tool_response
+      toolCtx.toolName = toolCtx.toolName || input.tool_name
+      sessionCtx.pendingTools.delete(id)
+      toolCh.asyncEnd.publish(toolCtx)
```
sabrenner: publish `end` here as well for context management

packages/datadog-instrumentations/src/claude-agent-sdk.js
```diff
@@ -0,0 +1,262 @@
...
+
+    const subagentCtx = {
+      sessionId: input.session_id,
+      agentId,
+      agentType: input.agent_type,
+    }
+
+    sessionCtx.pendingSubagents.set(agentId, subagentCtx)
+    subagentCh.start.runStores(subagentCtx, () => {
+      subagentCh.end.publish(subagentCtx)
```
sabrenner: move this to the `onSubagentStop` hook, similarly

packages/datadog-instrumentations/src/claude-agent-sdk.js
sabrenner, line 1: all of the `const foo = function foo () {}` statements in this file can just be normal functions i believe. no need for variable assignment as well.

packages/dd-trace/src/llmobs/plugins/claude-agent-sdk/index.js
```diff
@@ -0,0 +1,115 @@
...
+    const output = ctx.error
+      ? safeStringify(ctx.error)
+      : safeStringify(ctx.toolResponse)
+
+    this._tagger.tagTextIO(span, input, output)
+
+    const metadata = {}
+    if (ctx.toolName) metadata.tool_name = ctx.toolName
+    if (ctx.toolUseId) metadata.tool_use_id = ctx.toolUseId
+    if (ctx.sessionId) metadata.session_id = ctx.sessionId
```
sabrenner: this should be a tag (not metadata), otherwise it is not parsed properly

packages/datadog-plugin-claude-agent-sdk/src/tracing.js
sabrenner, line 1: nit: we could simplify this a bit with a base class `ClaudeCodePlugin` which defines `bindStart` and `asyncEnd`, and calls a per-implementer `getName` function that returns the proper name for the span. could help DRY up this file a bit

packages/dd-trace/src/llmobs/plugins/claude-agent-sdk/index.js
```diff
@@ -0,0 +1,115 @@
...
+  setLLMObsTags (ctx) {
+    const span = ctx.currentStore?.span
+    if (!span) return
+
+    const input = ctx.prompt || ''
+    const output = ctx.lastAssistantMessage || ctx.stopReason || ''
+    this._tagger.tagTextIO(span, input, output)
+
+    const metadata = {}
+    if (ctx.sessionId) metadata.session_id = ctx.sessionId
```
sabrenner: this should be a tag instead of metadata

packages/dd-trace/src/llmobs/plugins/claude-agent-sdk/index.js
```diff
@@ -0,0 +1,115 @@
...
+    if (!span) return
+
+    const input = ctx.agentType || ctx.agentId || ''
+    const output = ctx.lastAssistantMessage || ''
+    this._tagger.tagTextIO(span, input, output)
+
+    const metadata = {}
+    if (ctx.agentId) metadata.agent_id = ctx.agentId
+    if (ctx.agentType) metadata.agent_type = ctx.agentType
+    if (ctx.sessionId) metadata.session_id = ctx.sessionId
```
sabrenner: this should be a tag, not metadata

packages/datadog-instrumentations/src/claude-agent-sdk.js
sabrenner, line 1: while the `end` event comments i left still apply, there is a bigger fundamental problem in that the APM tracing plugin interacting with this, which is what creates trace contexts, is not properly set up. as of now, each of the hook lifecycles create their own traces. this is because the context is scoped to each `runStores` call, and is dropped after the function it's wrapping stops. 

this works fine for LLM Observability parentage, but as far as APM is concerned, which creates trace contexts, for an agent turn with two tool spans, it creates three entirely different traces.

i recommend seeing if we can instead do a bit of manual context management. it's a bit ugly, but seeing as this SDK just spawns a subprocess under the hood we have no control over, I don't see a better option.

sabrenner (commented):
one main concern - currently, if a query spawns two tool calls, we'll have one agent (turn) span, and two tool calls. while the tool calls will have the proper LLM Observability parentage, not only will they have the wrong APM parentage, but for both, each one will be its own trace. this is because

```js
ch.runStores(ctx, () => {})
```

immediately enters and then exits that context. it kinda sucks, but let's do some manual context management here, probably by doing things like taking the turn, tool call, subagent contexts we track on the session context and using those with manual calls to storage `enterWith` calls on the plugin side. my bad for not really catching this before, as we were focused on the testing strategy.

otherwise, the LLMObs tagging logic looks good, just a couple of comments on making `session_id` a tag instead of metadata.

sabrenner:
we can more easily fix this (the trace context issue), actually, with something like
```js
const turnStore = sessionCtx.currentTurn?.currentStore
if (turnStore) {
  legacyStorage.run(turnStore, () => toolCh.start.runStores(toolCtx, () => {}))
} else {
  toolCh.start.runStores(toolCtx, () => {})
}
```

and

```js
const turnStore = sessionCtx.currentTurn?.currentStore
if (turnStore) {
  legacyStorage.run(turnStore, () => subagentCh.start.runStores(subagentCtx, () => {}))
} else {
  subagentCh.start.runStores(subagentCtx, () => {})
}
```

ie manually activating context. i think this is the cleaner approach, although not standard in our instrumentation, but then again this is a non-standard library to instrument. this seems to work locally.

aside from that, my one other request is if we can more closely mirror the (LLMObs) span names from the Python integration, for consistency

packages/datadog-instrumentations/src/claude-agent-sdk.js
```diff
@@ -0,0 +1,262 @@
...
+      turnCtx.stopReason = input.stop_reason
+      turnCtx.lastAssistantMessage = input.last_assistant_message
+      turnCh.asyncEnd.publish(turnCtx)
+      sessionCtx.currentTurn = null
+    }
+    sessionCtx.lastAssistantMessage = input.last_assistant_message
+    return {}
+  }
+
+  const onPreToolUse = function onPreToolUse (input, toolUseId) {
```
sabrenner: commenting here, but to be implemented in each of the subscribers - we should probably not create a tool span(s) when it is the tool spawning a subagent. it shows with a redundant span currently (ie, the `Agent` tool span is not really needed)

Image

packages/dd-trace/test/plugins/versions/package.json
```diff
@@ -8,6 +8,7 @@
     "@ai-sdk/anthropic": "3.0.85",
     "@ai-sdk/google": "3.0.83",
     "@ai-sdk/openai": "3.0.73",
+    "@anthropic-ai/claude-agent-sdk": "0.2.98",
```
sabrenner:
```suggestion
    "@anthropic-ai/claude-agent-sdk": "0.3.195",
```

sabrenner:
converting this to a draft while i push up some refactor commits and tests, to close coverage gap for LLM spans and proper parentage based on the event stream rather than hooks. tests will also be more streamlined

docker-compose.yml
```diff
@@ -228,6 +228,10 @@ services:
       - ENABLED_CHECKS=trace_stall,meta_tracer_version_header,trace_count_header,trace_peer_service
       - PORT=9126
       - VCR_CASSETTES_DIRECTORY=/vcr-cassettes
+      - VCR_PROVIDER_MAP=claude-agent-sdk=https://api.anthropic.com
+      - VCR_JSON_BODY_NORMALIZERS=metadata.user_id,output_config.effort
+      - >-
+        VCR_BODY_REGEX_NORMALIZERS=agentId: [a-f0-9]+,SendMessage with to: '[a-f0-9]+',<usage>[\s\S]*?</usage>,"events":\[\{"event_type":"ClaudeCodeInternalEvent"[^\]]*\],cc_version=[^;]+,cch=[^;]+,<system-reminder>[\s\S]*?</system-reminder>,<env>[\s\S]*?</env>,(?<="input":\{"description":")[^"]*,(?:You are an agent for Claude Code|Write the title|Report the result concisely\.|Async agent launched successfully)(?:\\.|[^"\\])*
```
sabrenner, line 234: lots of weird normalization had to happen between the linux GHA runners, local macos development, etc., but these normalizations on things that are non-query specific seem to do the trick. this might be finicky for versions going forward, but work here

sabrenner:
@codex review

mr-lee:
was there a strong reason to swap away from hook-based instrumentation?

sabrenner:
> was there a strong reason to swap away from hook-based instrumentation?

we would not get LLM spans while using hooks, and parentage was still very finicky. Parsing the event stream seems to be more stable to me

mr-lee:
>we would not get LLM spans while using hooks, and parentage was still very finicky. Parsing the event stream seems to be more stable to me

This makes sense for the LLM spans - but also - the hooks are a direct ancestor to the event stream, in a way.  IMO it's a more stable surface to instrument to get reliable span creation than the event stream in isolation.  Perhaps consider a joint approach?

sabrenner:
> Perhaps consider a joint approach?

i had thought of that, but parentage gets really tricky and i'd rather not be storing contexts and re-activating them when we can "reconstruct" the lifecycle events with tracing channels and make it a bit cleaner. this is sorta similar to our approach with tracing the python version as well.

mr-lee:
>i had thought of that, but parentage gets really tricky and i'd rather not be storing contexts and re-activating them when we can "reconstruct" the lifecycle events with tracing channels and make it a bit cleaner. this is sorta similar to our approach with tracing the python version as well.

Wait - does this need to be stored in memory?  My argument is to have this be the trigger (e.g. the "Stop" hook) because that's the cleanest injection point.  As and when progressive tracing (i.e. be able to publish incomplete traces) becomes a requirement - additional hook points can be included.

sabrenner:
superseded by https://github.com/DataDog/dd-trace-js/pull/9202

PR #7975, opened 2026-04-10
release: pass github token from sts to release notes

IlyasShabi:
### What does this PR do?
The node release notes [step](https://github.com/DataDog/dd-trace-js/actions/runs/24243280946/job/70782946176) was failing with HTTP 403 because `contents: write` was removed from GITHUB_TOKEN permissions as part of the dd-octo-sts migration. 

### Motivation
Release notes step is failing https://github.com/DataDog/dd-trace-js/actions/runs/24243280946/job/70782946176 due to miss of GH token

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -98,6 +98,8 @@ jobs:
           git tag v${{ fromJson(steps.pkg.outputs.json).version }}
           git push https://x-access-token:${{ steps.octo-sts.outputs.token }}@github.com/${{ github.repository }}.git v${{ fromJson(steps.pkg.outputs.json).version }}
       - run: node scripts/release/notes --latest
+        env:
+          GH_TOKEN: ${{ steps.octo-sts.outputs.token }}
 
   docs:
     needs: "publish-latest"


.github/workflows/release.yml
```diff
@@ -98,6 +98,8 @@ jobs:
           git tag v${{ fromJson(steps.pkg.outputs.json).version }}
           git push https://x-access-token:${{ steps.octo-sts.outputs.token }}@github.com/${{ github.repository }}.git v${{ fromJson(steps.pkg.outputs.json).version }}
       - run: node scripts/release/notes --latest
+        env:
+          GH_TOKEN: ${{ steps.octo-sts.outputs.token }}
```
xopham: ah, side effects 😒 can also add the `contents: write` above and have the `GITHUB_TOKEN` do it

xopham (dismissed):
lgtm. Minor comment, but should be fine

.github/workflows/release.yml
IlyasShabi: yes, this is simpler

.github/workflows/release.yml
xopham, line 100: my bad, just noticed from the script that you'll now have to inject the `GITHUB_TOKEN` here as you did the dd-octo-sts token before. Sorry!

xopham (commented):
one small fix that I just noticed when reviewing the script

BridgeAR (dismissed):

xopham (approved):
lgtm

PR #7976, opened 2026-04-10
feat(next): add dd-trace/next entry point for RUM<>APM trace correlation

Miz85:
### What does this PR do?

Extends the existing `datadog-plugin-next` package with new helpers for Next.js
App Router, exposed via a `dd-trace/next` entry point. All new files live within
the existing `packages/datadog-plugin-next/src/` structure alongside the current
`next.request` server plugin.

The existing `NextPlugin` class in `src/index.js` (automatic HTTP request
tracing, page load events) is unchanged.

### Motivation

The existing Next.js plugin provides automatic `next.request` spans for HTTP
handling, but App Router introduced three gaps:

1. **RUM↔APM correlation** — Server-rendered pages had no mechanism to pass
   trace context to the browser. `getDatadogTraceMetadata()` bridges this via
   Next.js's `generateMetadata` API.
2. **Server Actions** — These run inside the HTTP request span but have no
   distinct identity in traces. `withDatadogServerAction()` adds named spans.
3. **Error attribution** — Next.js's `onRequestError` hook provides route-level
   error context that the generic HTTP instrumentation misses.

### Additional Notes

- The RSC origin check (`context._trace.origin === 'rum'`) prevents a subtle
  bug: during client-side `<Link>` navigation, the RSC response embeds metadata
  that Next.js applies to the document, overwriting the original page's trace
  context and creating duplicate root spans.
- This is the server-side counterpart to `@datadog/browser-rum-nextjs` in the
  Browser SDK, which reads the meta tags and propagates trace context on RUM
  resource events.
- All helpers gracefully no-op when dd-trace is not initialized (`global._ddtrace`
  is undefined), making them safe to call unconditionally.

### Test instructions

**Automated (extend existing integration tests):**

Add tests to `packages/datadog-plugin-next/test/index.spec.js` that:

1. Verify `GET /appDir/test` response contains `<meta name="dd-trace-id">` and
   `<meta name="dd-root-span-id">` (requires adding `getDatadogTraceMetadata()`
   to `test/app/layout.js`'s `generateMetadata`)
2. Verify the `web.request` span's `parentId` matches the `dd-root-span-id`
   value (re-parenting works)
3. Send a request with `RSC: 1` and `x-datadog-origin: rum` headers — verify
   the response does NOT contain `dd-trace-id` meta tags

**Manual (full end-to-end with RUM):**

The existing test app at `packages/datadog-plugin-next/test/` can be used:

```bash
# Build the test app
cd packages/datadog-plugin-next/test
npx next build

# Start with dd-trace preloaded
NODE_OPTIONS="--require ./datadog.js" PORT=3000 HOSTNAME=localhost node server.js
```

Then test:

1. **Initial page load:** Open `http://localhost:3000/appDir/test` — view page
   source and confirm `<meta name="dd-trace-id">`, `<meta name="dd-trace-time">`,
   and `<meta name="dd-root-span-id">` are present. In Datadog, the RUM document
   resource should link to an APM trace with `web.request` and `next.request`
   spans nested under the browser span.

2. **Client-side navigation:** Add a `<Link>` between two App Router pages.
   Navigate via the link, then check the trace — there should be a single root
   browser span (no duplicates). The `dd-trace-id` meta tag in the document
   should still contain the original page load's trace ID.

3. **Server Action:** Wrap an action with `withDatadogServerAction('name', ...)`
   and trigger it — verify a named child span appears under the HTTP span in APM.

4. **Error handling:** Export `datadogOnRequestError` from `instrumentation.ts`,
   trigger a server error, and verify a `nextjs.server_error` span with
   `nextjs.route_path` and `error.message` tags.


packages/datadog-plugin-next/src/utils/parse-session-cookie.js
```diff
@@ -0,0 +1,23 @@
...
+
+  const entries = cookieValue.split(SESSION_ENTRY_SEPARATOR)
+  for (const entry of entries) {
+    const match = SESSION_ENTRY_REGEXP.exec(entry)
+    if (match !== null) {
+      const [, key, value] = match
+      if (key === 'id') return value
+    }
+  }
+}
```
tlhunter: Am I reading this function right that it splits a query string on all the `&` characters then runs the regex on every single pair then returns once it finds the one beginning with an `id`? If so it seems rather inefficient.

There's a built in module / function for parsing strings formatting in this manner, `querystring.parse()`:
https://nodejs.org/api/querystring.html#querystringparsestr-sep-eq-options

If the key/value pairs are expected to be long and have many pairs then I suppose you could use a single regex applied to the entire string (not applied for every ampersand-separated component). The regex would look something like this:

```js
/id=([a-z0-9-]+)/
```

packages/datadog-plugin-next/src/request-error.js
```diff
@@ -0,0 +1,70 @@
...
+  if (error?.digest) {
+    span.setTag('nextjs.error_digest', error.digest)
+  }
+
+  // Extract RUM session ID from _dd_s cookie for client<>server correlation
+  const cookieHeader = request.headers?.cookie
+  if (cookieHeader) {
+    const ddsCookieMatch = cookieHeader.match(/(?:^|;\s*)_dd_s=([^;]*)/)
+    if (ddsCookieMatch) {
+      const sessionId = parseRumSessionId(ddsCookieMatch[1])
```
tlhunter: This part also seems inefficient too since we're searching the string twice. Line 58 does a regex and line 60 calls a function to do one or more regexes as well. Is there a way to do this in a single operation?

packages/datadog-plugin-next/src/trace-metadata.js
```diff
@@ -0,0 +1,97 @@
...
+  const context = activeSpan.context()
+
+  // Skip meta tag emission for RSC fetches that already have a RUM-injected parent.
+  // The RUM SDK sets x-datadog-origin: 'rum' which dd-trace stores as _trace.origin.
+  // For these requests the browser already has the trace context — emitting meta tags
+  // would overwrite the document's original trace data during client-side navigation.
+  const origin = context._trace && context._trace.origin
+  if (origin === 'rum') {
+    return {}
+  }
```
tlhunter: Might as well be consistent ;)

```suggestion
  if (origin === 'rum') return {}
```

tlhunter:
@codex review

tlhunter (changes requested):
I commented with a few perf concerns

packages/datadog-plugin-next/src/utils/parse-session-cookie.js
Miz85: yeah you're right, good catch! will fix.

BridgeAR:
I am currently looking at old PRs and I believe this is outdated. I am closing it as such, please feel free to reopen in case that is wrong!

PR #7977, opened 2026-04-10
chore: simplify trace propagation env and logger implementation

BridgeAR:
The trace propagation can rely on our fallback env handling. This works out of the box, so the code special handling is not needed anymore and simplifies the configuration code.

The memoize method was only used in a single spot and the deprecate method was never called anywhere in the code. So removing that seemed like the right thing. The memoize function was replaced with a single set that does the lookup in place.

As drive-by remove optional chaining in tests due to knowing the types for these with the generated configuration types.


eslint-rules/eslint-env-aliases.mjs
```diff
@@ -22,6 +23,12 @@ for (const [canonical, entries] of Object.entries(supportedConfigurations)) {
   }
 }
 
+for (const [alias, canonicals] of Object.entries(aliasToCanonical)) {
+  if (canonicals.length > 1 && Object.hasOwn(supportedConfigurations, alias)) {
```
szegedi, line 27: Having reviewed this all, I now understand why is there a `canonicals.length > 1 ` here, but it was not intuitive at first. I feel adding a bit of a comment here would help.

packages/dd-trace/src/config/index.js
```diff
@@ -712,7 +696,7 @@ function warnWrongOtelSettings () {
     // eslint-disable-next-line eslint-rules/eslint-env-aliases
     ['OTEL_LOG_LEVEL', 'DD_TRACE_LOG_LEVEL', 'logLevel'],
     // eslint-disable-next-line eslint-rules/eslint-env-aliases
-    ['OTEL_PROPAGATORS', 'DD_TRACE_PROPAGATION_STYLE'],
+    ['OTEL_PROPAGATORS', 'DD_TRACE_PROPAGATION_STYLE', 'DD_TRACE_PROPAGATION_STYLE'],
```
szegedi, line 699: This still looks weird, tho'.

packages/dd-trace/src/encode/0.4.js
```diff
@@ -338,11 +337,13 @@ class AgentEncoder {
...
-  // return something to store in memoize cache
-  return true
-})
+const seenKeys = new Set()
+const memoizedLogDebug = (key, message) => {
+  if (!seenKeys.has(key)) {
+    seenKeys.add(key)
+    log.debug(message)
+  }
+}
```
szegedi, line 346: Big +1 on reworking this. Memoization here really felt like an abuse of the mechanism :-)

szegedi (approved):

eslint-rules/eslint-env-aliases.mjs
BridgeAR: I can do that, while this is just a lint rule that should ideally never be read 😅

packages/dd-trace/src/config/index.js
BridgeAR: This is indeed somewhat weird. I personally think the whole special handling for Otel is not great and I would rather have a generic config warning (which we already have in the telemetry, it is just ignored in the backend).

This is about a) warning on duplicate envs, which it does and b) report invalid values, which it also does. In this case, if both are set, it will report it both values are invalid. I think this is alright for that particular case.

PR #7978, opened 2026-04-10
chore: refactor profiler start

BridgeAR:
The profiler start code used to copy the configs and to have a separate logger behavior. That is now changed to use the generic logger the same way as other code.

This simplifies a lot of code overall.

### DIFF
--- a/benchmark/sirun/profiler/index.js
+++ b/benchmark/sirun/profiler/index.js
@@ -2,8 +2,6 @@
 
 const {
   profiler,
-  WallProfiler,
-  SpaceProfiler,
 } = require('../../../packages/dd-trace/src/profiling')
 
 const { PROFILER } = process.env
@@ -11,33 +9,16 @@ const { PROFILER } = process.env
 const profilers = []
 
 if (PROFILER === 'wall' || PROFILER === 'all') {
-  profilers.push(new WallProfiler())
+  profilers.push('wall')
 }
 if (PROFILER === 'space' || PROFILER === 'all') {
-  profilers.push(new SpaceProfiler())
+  profilers.push('space')
 }
 
-if (profilers.length === 0) {
-  // Add a no-op "profiler"
-  profilers.push({
-    start: () => {},
-    stop: () => {},
-    profile: () => { return true },
-    encode: () => { Promise.resolve(true) },
-  })
-}
+const exporters = ['none']
 
-const exporters = [{
-  export () {
-    profiler.stop()
-    return Promise.resolve()
-  },
-}]
-
-profiler._start({
-  profilers,
-  exporters,
-  interval: 0,
-}).then(() => {
-  profiler._timer.ref()
-})
+profiler.start(/** @type {import('../../../packages/dd-trace/src/config/config-base')} */ ({
+  DD_PROFILING_PROFILERS: profilers,
+  DD_PROFILING_EXPORTERS: exporters,
+  DD_PROFILING_HEAP_SAMPLING_INTERVAL: 0,
+}))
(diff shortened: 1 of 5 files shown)


packages/dd-trace/src/profiling/profiler.js
```diff
@@ -191,41 +154,36 @@ class Profiler extends EventEmitter {
   /**
    * @param {import('../config/config-base')} options - Tracer configuration
    */
-  _start (options) {
+  start (options) {
     if (this.enabled) return true
+    this.#enabled = true
```
IlyasShabi, line 159: IMO this should be move after `new Config(options)` if the config fails

packages/dd-trace/src/profiling/profiler.js
```diff
@@ -191,41 +154,36 @@ class Profiler extends EventEmitter {
...
-              })
-            }
-          })
-          .catch((err) => {
-            this.#logError(err)
-          })
-      }
-    } catch (err) {
-      this.#logError(err)
+    const { setLogger, SourceMapper } = require('@datadog/pprof')
```
IlyasShabi, line 168: Why removing the try/catch? this probably could fail on serverless

packages/dd-trace/src/profiling/profiler.js
BridgeAR: I have another perspective: if that fails, we do not want to retry starting it again. So I would keep it where it is.

packages/dd-trace/src/profiling/profiler.js
BridgeAR: It is used in the proxy and this is wrapping it in another try/catch. So having this as additional safe guard seems not needed.

packages/dd-trace/src/profiling/profiler.js
IlyasShabi: Yes we shouldn't retry after a fail and we should not set enabled to true because it becomes misleading. 
I think that we should keep enabled for a successful start, and handle the no-retry case separately.

BridgeAR: In this case, that variable is used for that purpose. We could have another state (setting it to undefined or similar), while I wonder if we overcomplicate things for now.

I can move it below the config part as well, but as I said: repeating that seems confusing to me (besides the fact that start should normally never be called more than once anyway)

IlyasShabi: LGTM, I'd like to have another state later

IlyasShabi (dismissed):
@szegedi Could you please take a look

benchmark/sirun/profiler/index.js
```diff
@@ -2,42 +2,23 @@
...
 
-if (profilers.length === 0) {
-  // Add a no-op "profiler"
-  profilers.push({
-    start: () => {},
-    stop: () => {},
-    profile: () => { return true },
-    encode: () => { Promise.resolve(true) },
-  })
-}
```
szegedi, line 28: This was presumably here because of at least [this check](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/profiling/profiler.js#L343) in `profiler.js`. I presume this now happens to work due to the fact that when no profilers are specified, the defaults are used by the config.

That does change the behavior of `PROFILER` env var for the benchmarks though; previously you could specify `PROFILER=` (or `PROFILER=none`, or really any value that's not `wall`, `space`, or `all`) to have this no-op profiler configured so you could benchmark the overhead of profiling itself being on but no actual profilers running. That can be used as a baseline for checking individual profiler overheads. 

This definitely changes the semantics of `PROFILER=` as now it'll enable all profilers instead. To restore previous behavior, it seems to me you'd have to introduce a new recognized value in the config named e.g. `none` or `noop` and specify it here.

szegedi (changes requested):

benchmark/sirun/profiler/index.js
BridgeAR: That should still work as before.

With the new setup, we just pass things through differently. I guess the baseline should just be re-added that I removed?

Update: I did that. The change here should actually be a fix for the benchmark which is broken without this PR.

szegedi (approved):
We'll later create the ability to run the top-level profiler with no configured profilers, for benchmark baselining.

PR #7979, opened 2026-04-10
Apm ai toolkit/new integration/aws durable execution sdk js/20260410 013300

joeyzhao2018:


.github/workflows/apm-integrations.yml
```diff
@@ -147,6 +147,17 @@ jobs:
...
 
+  aws-durable-execution-sdk-js:
+    runs-on: ubuntu-latest
+    env:
+      PLUGINS: aws-durable-execution-sdk-js
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: ./.github/actions/plugins/test
+        with:
+          dd_api_key: ${{ secrets.DD_API_KEY }}
```
wconti27, line 158: I think serverless has its own CI file so this will need to be changed in the toolkit (and on this PR ofc)

index.d.ts
```diff
@@ -2291,6 +2292,12 @@ declare namespace tracer {
      * This plugin automatically instruments the
      * [aws-sdk](https://github.com/aws/aws-sdk-js) module.
      */
+    /**
+     * This plugin automatically instruments the
+     * [aws-durable-execution-sdk-js](https://github.com/npmjs/package/@aws/durable-execution-sdk-js) library.
+     */
+    interface aws_durable_execution_sdk_js extends Instrumentation {}
```
wconti27, line 2299: seems like this was injected between the comment and interface for aws-sdk, need to fix on the toolkit side

.github/workflows/apm-integrations.yml
joeyzhao2018: yes. i'll remove this.

package.json
```diff
@@ -140,6 +140,7 @@
     "version.js"
   ],
   "dependencies": {
+    "@aws/durable-execution-sdk-js": "1.1.0",
```
wconti27, line 143: The agent consistently struggles and tries to change package.json while trying to get test infrastructure running correctly, maybe a blocking hook from changing this file. For Joey, we will need to remove this change.

packages/datadog-plugin-aws-durable-execution-sdk-js/src/client.js
```diff
@@ -0,0 +1,61 @@
...
+   * @param {{ arguments?: ArrayLike<unknown> }} ctx
+   * @returns {Record<string, string>}
+   */
+  getTags (ctx) {
+    const args = ctx.arguments || []
+    const functionname = args[1] ? String(args[1]) : undefined
+
+    const tags = {
+      component: 'aws-durable-execution-sdk-js',
+      'span.kind': 'client',
```
wconti27, line 35: The toolkit pulls a list of tags to set according to our APM Semantics, but we don't currently have any defined semantics for orchestration category integrations, so no tags were added :(. I would assume there's more tags that should be defined for all these spans.

PR #7992, opened 2026-04-14
fix(openfeature): clear flags config when RC tells us to

dd-oleksii:
### What does this PR do?

Clear flags configuration when Remote Config tells us to unapply the config.

### Motivation

This is required to make behavior consistent with other tracers and pass new eval metrics system tests.

### DIFF
--- a/packages/dd-trace/src/openfeature/remote_config.js
+++ b/packages/dd-trace/src/openfeature/remote_config.js
@@ -19,9 +19,14 @@ function enable (rc, config, getOpenfeatureProxy) {
 
   // Set product handler for FFE_FLAGS
   rc.setProductHandler('FFE_FLAGS', (action, conf) => {
-    // Feed UFC config directly to OpenFeature provider
     if (action === 'apply' || action === 'modify') {
+      // Feed UFC config directly to OpenFeature provider
       getOpenfeatureProxy()._setConfiguration(conf)
+    } else if (action === 'unapply') {
+      // Clear the configuration so evaluations return PROVIDER_NOT_READY,
+      // consistent with Go and Python which also set config to null on RC deletion.
+      // The evaluator returns PROVIDER_NOT_READY when config is null/undefined.
+      getOpenfeatureProxy()._setConfiguration(null)
     }
   })
 }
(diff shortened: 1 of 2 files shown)


simon-id:
WARNING: there is no RC protocol guarantee that a RC product like `FFE_FLAGS`, will always provide one single configuration.
This means you could technically receive multiple different configurations for FFE_FLAGS at different times. For example:
1. Apply conf A
2. Apply conf B
3. Unapply conf A
4. Unapply conf B

This means you usually need to keep track of them by their ID, and unapply them selectively. With your changes here, step 3 above would remove both conf A and conf B.

If you assume you will only ever receive one FFE_FLAGS configuration at a time, this needs to be enforced by the backend, as it is not enforced by the protocol/data model itself. Make sure your backend assures you will only ever receive one at a time.

Otherwise LGTM, but would be nice to have a more integration-y test, to make sure it actually clears the config, instead of just having one unit test that checks a function is called.

BridgeAR (approved):

dd-oleksii:
@simon-id thanks for the warning 🙏 Our team is aware of that: one-config-at-a-time is enforced by our backend and we currently rely on this in all tracers.

The change here is mostly for testing, so we can test different configs / no config scenarios in system tests. I've also tested this change against system tests locally, so I can confirm that it's clearing the config correctly. (Although I agree that another local test would be nice to have — will try adding one shortly)

simon-id:
System tests definitely count as integ tests! thank you for testing locally, but it's way safer to open another PR on the system tests repo with the new tests enabled, and that points to this branch, so you can have the full CI run :+1:

simon-id (approved):

sameerank (approved):

greghuels (approved):

dd-oleksii:
> but it's way safer to open another PR on the system tests repo with the new tests enabled, and that points to this branch, so you can have the full CI run 👍

No config scenario is covered by feature_flags_eval_metrics.Test_FFE_Eval_No_Config_Loaded test but it also dependents on https://github.com/DataDog/dd-trace-js/pull/7993 to actually send eval metrics. (I've been working on the feature and decided to split PRs so they are easier to review.) So CI run will not pass yet.

We don't have a lot of system tests, so I did have the full FFE suite pass. So between my local run and the small unit test here, I'm confident enough that it's working as intended, so I'm going to merge this PR and retest on #7993 branch

PR #7993, opened 2026-04-14
feat(openfeature): implement flag evaluation metrics

dd-oleksii:
### What does this PR do?

Add an OpenFeature `finally` hook that emits a `feature_flag.evaluations` OTel counter for every flag evaluation, including error and short-circuit cases. The hook is a no-op when `otelMetricsEnabled` is false.

### Motivation

[FFL-1944](https://datadoghq.atlassian.net/browse/FFL-1944)

### Additional Notes

Tested locally against system-tests and it passes except a couple of cases affected by [FFL-2064](https://datadoghq.atlassian.net/browse/FFL-2064).

[FFL-1944]: https://datadoghq.atlassian.net/browse/FFL-1944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FFL-2064]: https://datadoghq.atlassian.net/browse/FFL-2064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/src/openfeature/eval-metrics-hook.js
+++ b/packages/dd-trace/src/openfeature/eval-metrics-hook.js
@@ -1,0 +1,103 @@
+'use strict'
+
+const log = require('../log')
+
+const METER_NAME = 'dd-trace-js/openfeature'
+const COUNTER_NAME = 'feature_flag.evaluations'
+const COUNTER_DESCRIPTION = 'Number of feature flag evaluations'
+const COUNTER_UNIT = '{evaluation}'
+
+/**
+ * OpenFeature hook that tracks feature flag evaluation metrics using an
+ * OpenTelemetry counter.
+ *
+ * Implements the OpenFeature `finally` hook interface so it can be pushed
+ * directly onto a provider's `hooks` array. We use the `finally` stage
+ * (not diagnostic channels inside the provider's `resolve*` methods) because
+ * the OpenFeature SDK short-circuits before calling the provider when it is in
+ * NOT_READY state; the `finally` hook still fires, ensuring all evaluations are
+ * captured. It also catches type-mismatch errors detected by the SDK client
+ * after the provider returns.
+ *
+ * The counter is created lazily on the first successful `finally()` call rather
+ * than in the constructor. This is necessary because `FlaggingProvider` is
+ * constructed eagerly by `proxy.js#updateTracing()`, which runs *before*
+ * `initializeOpenTelemetryMetrics()` sets the global OTel meter provider.
+ * Calling `getMeter()` in the constructor would return the noop meter and
+ * produce a noop counter that silently discards all measurements. By deferring
+ * to `finally()` time we give the meter provider a chance to be set up first.
+ *
+ * If counter creation fails (e.g. the OTel API is not yet available), the call
+ * is silently skipped and retried on the next `finally()` invocation.
+ *
+ * When `config.otelMetricsEnabled` is false, `finally()` is always a no-op.
+ */
+class EvalMetricsHook {
+  #enabled = false
+  #counter = null
+
+  /**
+   * @param {import('../config')} config - Tracer configuration object
+   */
+  constructor (config) {
+    this.#enabled = config.otelMetricsEnabled === true
+  }
+
+  /**
+   * Returns the OTel counter, creating it on first successful call.
+   * Returns `null` if counter creation fails; will retry on next call.
+   *
+   * @returns {import('@opentelemetry/api').Counter | null}
+   */
+  #getCounter () {
+    if (this.#counter) return this.#counter
+
+    try {
+      const { metrics } = require('@opentelemetry/api')
+      const meter = metrics.getMeter(METER_NAME)
+      this.#counter = meter.createCounter(COUNTER_NAME, {
+        description: COUNTER_DESCRIPTION,
+        unit: COUNTER_UNIT,
+      })
+    } catch (e) {
+      log.warn('EvalMetricsHook: failed to create counter: %s', e.message)
+    }
+
+    return this.#counter
+  }
+
+  /**
+   * Called by the OpenFeature SDK after every flag evaluation (success or error).
+   *
+   * @param {{ flagKey: string }} hookContext - Hook context containing the flag key
+   * @param {{ variant?: string, reason?: string, errorCode?: string, flagMetadata?: object }} evaluationDetails
+   *   - Full evaluation details
+   * @returns {void}
+   */
+  finally (hookContext, evaluationDetails) {
+    if (!this.#enabled) return
+
+    const counter = this.#getCounter()
+    if (!counter) return
+
+    const attributes = {
+      'feature_flag.key': hookContext?.flagKey ?? '',
+      'feature_flag.result.variant': evaluationDetails?.variant ?? '',
+      'feature_flag.result.reason': evaluationDetails?.reason?.toLowerCase() ?? 'unknown',
+    }
+
+    const errorCode = evaluationDetails?.errorCode
+    if (errorCode) {
+      attributes['error.type'] = errorCode.toLowerCase()
+    }
+
+    const allocationKey = evaluationDetails?.flagMetadata?.allocationKey
+    if (allocationKey) {
+      attributes['feature_flag.result.allocation_key'] = allocationKey
+    }
+
+    counter.add(1, attributes)
+  }
+}
+
+module.exports = EvalMetricsHook
--- a/packages/dd-trace/src/openfeature/flagging_provider.js
+++ b/packages/dd-trace/src/openfeature/flagging_provider.js
@@ -4,6 +4,7 @@ const { DatadogNodeServerProvider } = require('@datadog/openfeature-node-server'
 const { channel } = require('dc-polyfill')
 const log = require('../log')
 const { EXPOSURE_CHANNEL } = require('./constants/constants')
+const EvalMetricsHook = require('./eval-metrics-hook')
 
 /**
  * OpenFeature provider that integrates with Datadog's feature flagging system.
@@ -23,6 +24,8 @@ class FlaggingProvider extends DatadogNodeServerProvider {
 
     this._tracer = tracer
     this._config = config
+
+    this.hooks.push(new EvalMetricsHook(config))
 
     log.debug('%s created with timeout: %dms', this.constructor.name,
       config.experimental.flaggingProvider.initializationTimeoutMs)
--- a/packages/dd-trace/test/openfeature/eval-metrics-hook.spec.js
+++ b/packages/dd-trace/test/openfeature/eval-metrics-hook.spec.js
@@ -1,0 +1,213 @@
+'use strict'
+
+const assert = require('node:assert/strict')
+
+const { describe, it, beforeEach } = require('mocha')
+const sinon = require('sinon')
+const proxyquire = require('proxyquire')
+
+require('../setup/core')
+
+describe('EvalMetricsHook', () => {
+  let EvalMetricsHook
+  let mockCounter
+  let mockMeter
+  let mockOtelApi
+  let log
+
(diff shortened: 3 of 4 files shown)


packages/dd-trace/src/openfeature/eval-metrics-hook.js
```diff
@@ -0,0 +1,103 @@
...
+    if (errorCode) {
+      attributes['error.type'] = errorCode.toLowerCase()
+    }
+
+    const allocationKey = evaluationDetails?.flagMetadata?.allocationKey
+    if (allocationKey) {
+      attributes['feature_flag.result.allocation_key'] = allocationKey
+    }
+
+    counter.add(1, attributes)
```
sameerank, line 99: In Python, `counter.add` could throw and therefore needed [to be wrapped in a try/catch](https://github.com/DataDog/dd-trace-py/blob/62e5bae1b94348d4041515a077e311af1c527969/ddtrace/internal/openfeature/_flageval_metrics.py#L123-L125).

If `counter.add` in JS behaves similarly, then you may want to add the same error handling protection here

packages/dd-trace/src/openfeature/eval-metrics-hook.js
```diff
@@ -0,0 +1,103 @@
+'use strict'
+
+const log = require('../log')
+
+const METER_NAME = 'dd-trace-js/openfeature'
```
sameerank, line 5: Just sharing an observation: every SDK uses a different name here. I don't know if that has any consequences. I assume it's fine

- JS (this PR): `dd-trace-js/openfeature`
- .NET: `Datadog.FeatureFlags.OpenFeature`
- Go: `github.com/DataDog/dd-trace-go/openfeature`
- Python: `ddtrace.openfeature`

sameerank (approved):
FYI we are currently ignoring this following system test for node: `Test_FFE_Eval_Targeting_Key_Optional`

https://github.com/DataDog/system-tests/blob/7d79f7af78d86a8e43723e6c27854ca368cfd70c/manifests/nodejs.yml#L1579

At some point we may want to think about making it match the other tracers

packages/dd-trace/src/openfeature/eval-metrics-hook.js
dd-oleksii: OpenFeature SDK has [its own try/catch wrapper](https://github.com/open-feature/js-sdk/blob/431f899191a2f7693fea9fab40f7cfb1f22f61f8/packages/server/src/client/internal/open-feature-client.ts#L392-L402) (I believe required by the spec), so there should be no need

dd-oleksii:
> FYI we are currently ignoring this following system test for node: Test_FFE_Eval_Targeting_Key_Optional

Yeah. My understanding is that this should have been fixed in FFL-1730 but it was not

packages/dd-trace/src/openfeature/eval-metrics-hook.js
```diff
@@ -0,0 +1,103 @@
...
+   * Returns the OTel counter, creating it on first successful call.
+   * Returns `null` if counter creation fails; will retry on next call.
+   *
+   * @returns {import('@opentelemetry/api').Counter | null}
+   */
+  #getCounter () {
+    if (this.#counter) return this.#counter
+
+    try {
+      const { metrics } = require('@opentelemetry/api')
```
rochdev, line 56: This would mean there is now a hard dependency from OpenFeature to OpenTelemetry. Can the API be done in a way that they can be externally combined instead?

rochdev (dismissed):

packages/dd-trace/src/openfeature/eval-metrics-hook.js
dd-oleksii: This code is only evaluated when otel is enabled, so it is a soft dependency — not a hard one.

dd-oleksii:
(rebased on latest master to run system-tests)

leoromanovsky (approved):
Solid tests.

packages/dd-trace/src/openfeature/eval-metrics-hook.js
dd-oleksii: Heads up. This comment is the last thing blocking the merge and @rochdev is unresponsive on both github and slack.

FFE team's current consensus is that it is a misreading of the code and does not warrant the block. We also use the same approach in [Go](https://github.com/DataDog/dd-trace-go/pull/4489), [Java](https://github.com/DataDog/dd-trace-java/pull/11040), [Python](https://github.com/DataDog/dd-trace-py/pull/17029), and [.NET](https://github.com/DataDog/dd-trace-dotnet/pull/8367) tracers.

I'm dismissing the review to unblock the progress. I'll wait for Monday eve before merging unless I hear any objections 🙏

rochdev: It's a hard dependency because bundlers require that all requires have modules that can be found. I am planning to address this issue with dd-trace usage, but this introduces the same problem for OpenFeature. If it's expected that OpenFeature depends on OpenTelemetry, with the understanding that it also means we can never use it internally in dd-trace, then that's fine. If that was done by mistake, then this needs to be revisited.

rochdev (dismissed):

packages/dd-trace/src/openfeature/eval-metrics-hook.js
dd-oleksii: @rochdev isn't `@opentelemetry/api` already a hard dependency when using bundlers?

It is already `require`d through importing `dd-trace`:
https://github.com/DataDog/dd-trace-js/blob/873af1cac4043e99fcaeca47e058cd86f26e6933/packages/dd-trace/src/proxy.js#L233-L236
https://github.com/DataDog/dd-trace-js/blob/873af1cac4043e99fcaeca47e058cd86f26e6933/packages/dd-trace/src/opentelemetry/metrics/index.js#L31

We also have a similar situation with a bunch of other optional dependencies, including `@openfeature/core`.

It looks like we have some handling here and there but I'm not sure if that's enough.
https://github.com/DataDog/dd-trace-js/blob/873af1cac4043e99fcaeca47e058cd86f26e6933/vendor/rspack.config.js#L68-L71
https://github.com/DataDog/dd-trace-js/blob/873af1cac4043e99fcaeca47e058cd86f26e6933/packages/datadog-esbuild/index.js#L131-L137

So... I'm not sure how this PR changes the situation?

In terms of solving it, it actually looks like something that should be fixed in bundler's configuration. Most bundlers have ways to exclude dependencies from bundling. There are also ways to force dynamic import like using string interpolations for import path (`'@' + 'opentelemetry' + '/api'`) but that's rather hacky.

Another option is to export this hook as a separate entry point and require our users to register it (I believe that's what you mean by "externally combined?") — but that makes initialization more complicated and it's highly likely that the majority of the users will skip it (and will later wonder why they have empty dashboards). It's also not entirely equivalent because it will cause our hook to trigger for events emitted by other providers as well.

I'm also not sure if that's something that we need to fix _right now_ or we can revisit it when someone wants to use openfeature inside dd-trace. Given we're still under an experimental env var, I believe revisiting later should be fine.

So, if my assessment that it's a pre-existing issue is correct, I'm leaning to merging this as is and postpone internal dd-trace usage for later. But will chat a bit more with the team to see if we can find a good solution

rochdev: > So... I'm not sure how this PR changes the situation?

It worsens it, which is never great in the first place, but we plan to address the OTel issue this quarter so then only OpenFeature will still be a problem.

> In terms of solving it, it actually looks like something that should be fixed in bundler's configuration.

That's fine for an application, not for a library with thousands of users. It's a recurrent issue which we have mostly solved and OTel is the last remaining part which I'm working on right now.

> Another option is to export this hook as a separate entry point and require our users to register it (I believe that's what you mean by "externally combined?") — but that makes initialization more complicated and it's highly likely that the majority of the users will skip it (and will later wonder why they have empty dashboards)

That's where the main question becomes relevant: is OpenFeature dependent on OTel or not? If it is, then having a hard dependency makes sense, because OTel _must_ be present for OpenFeature to function. If it isn't, then I don't see why it wouldn't make sense to register something separate when using OTel. OTel itself is already full of programmatic setup requirements in part for this reason. Any user of OTel already needs to install and setup something else along with any necessary glue between components.

> I'm also not sure if that's something that we need to fix right now or we can revisit it when someone wants to use openfeature inside dd-trace. Given we're still under an experimental env var, I believe revisiting later should be fine.

Yes and no. I wasn't planning to remove OpenFeature from the main package in my refactor, but with this change I would be forced to. It's possible I'll still need to do that since OpenFeature also shares some of the problems of OTel to begin with, but I was hoping to avoid it. With this change that would be guaranteed to be needed.

> So, if my assessment that it's a pre-existing issue is correct, I'm leaning to merging this as is and postpone internal dd-trace usage for later. But will chat a bit more with the team to see if we can find a good solution

To be clear I am not blocking the PR on this. I only used "requested change" to make sure everyone is on the same page about the implications before merging. If you think this is the way to go with all the above said, then I'll revert my review and start working on moving OpenFeature as a separate package.

dd-oleksii: @rochdev so what's the current plan for handling `@opentelemetry/api` in other places? Can we use the same approach?

And what's the plan for `@datadog/openfeature-node-server`, `@openfeature/core`, and `@openfeature/server-sdk`? These three are optional dependencies in dd-trace which are _required_ for feature flagging to work. If we have a way to handle these three, we should be able to use the same approach for our use of `@opentelemetry/api`. And if we don't have a way to handle these three and we have to evict feature flagging from dd-trace, we'd have to do it regardless of our use of `@opentelemetry/api`.

rochdev: Unfortunately, the design of OpenTelemetry means that in order to implement an SDK, we have to explicitly depend on `@opentelemetry/api` instead of receiving the one that the user is using explicitly. So the "fix" (or rather workaround for a design flaw we do not control) will be to take our own code that depends on it out of the main library, and have it be registered manually. The top level package will continue to do that automatically to avoid a breaking change but we will use the new strategy whenever we can.

This does mean that OpenTelemetry will not be available out of the box in those environment, but that's fine because it will be an `npm install` away and it already requires registration of the TracerProvider anyway.

I looked a bit into OpenFeature a few months ago and I remember seeing similar design flaws, but I haven't looked into it enough to know if we can do better for it. However, if it depends on OpenTelemetry then we are guaranteed to have to do the same for OpenFeature by extension, so it will become available through an external install instead of being always included with the SDK.

rochdev: I think at this point enough context has been shared about this for you to make the call which is what I wanted to make sure of with the "requested changes", so I will lift it from the PR.

dd-oleksii:
(rebased to fix merge conflict in tests — no code changes)

packages/dd-trace/src/openfeature/eval-metrics-hook.js
dd-oleksii: openfeature is really in the same situation as opentelemetry here — we need to import openfeature sdk to implement the provider interface. We also have a bunch of other dependencies that we might not want in the core (stuff in [DataDog/openfeature-js-client](https://github.com/DataDog/openfeature-js-client) and spark-md5). So unless openfeature splits up the SDK from provider interface, I don't think we can be included in core. For these reasons, I'm going ahead with merging this PR as is

typotter (approved):

sameerank (approved):

PR #8000, opened 2026-04-14
fix: handle esm and cjs in rewriter

### LINKED ISSUE #7991: dd-trace rewriter injects CJS require() into ESM modules, breaking @langchain/langgraph on Node.js v24
## Environment

- dd-trace: >= 5.91.0 (tested up to 5.97.0)
- Node.js: v24.14.0
- @langchain/langgraph: 1.2.8
- Module system: ESM (`"type": "module"`)

## Description

Starting from dd-trace 5.91.0, a rewriter instrumentation was added for `@langchain/langgraph` in `packages/datadog-instrumentations/src/helpers/rewriter/instrumentations/langgraph.js`. This rewriter injects a CJS `require()` call into the first line of `@langchain/langgraph/dist/pregel/index.js`:

```javascript
const {tracingChannel: tr_ch_apm_tracingChannel} = require("/path/to/node_modules/dc-polyfill/dc-polyfill.js");
```

However, `@langchain/langgraph` is a pure ESM package (`"type": "module"`), so `require()` is not available. This causes:

```
ReferenceError: require is not defined in ES module scope, you can use import instead
    at file:///app/node_modules/@langchain/langgraph/dist/pregel/index.js:1:52
```

## Root cause

The rewriter in `packages/datadog-instrumentations/src/helpers/rewriter/index.js` calls `transformer.transform(content, moduleType)` and passes the correct `moduleType` (`'esm'`), but the generated code always uses `require()` regardless of module type.

The relevant rewriter config in `instrumentations/langgraph.js`:

```javascript
module.exports = [
  {
    module: {
      name: '@langchain/langgraph',
      versionRange: '>=1.1.2',
      filePath: 'dist/pregel/index.js',  // ESM file
    },
    functionQuery: {
      methodName: 'stream',
      className: 'Pregel',
    },
    channelName: 'Pregel_stream',
    transform: 'traceAsyncIterator',
  },
  // ... also targets dist/pregel/index.cjs (CJS file)
]
```

The rewriter targets both `.js` (ESM) and `.cjs` (CJS) entry points. For the `.cjs` file, injecting `require()` is correct. For the `.js` file (ESM), it should use `import` instead.

## Reproduction

```json
{
  "name": "repro",
  "type": "module",
  "dependencies": {
    "dd-trace": "5.97.0",
    "@langchain/langgraph": "1.2.8",
    "@langchain/core": "1.1.39",
    "hono": "4.12.12",
    "@hono/node-server": "1.19.13"
  }
}
```

```javascript
// test.mjs
import tracer from "dd-trace";
tracer.init();
import { StateGraph } from "@langchain/langgraph";
console.log("OK:", typeof StateGraph);
```

```bash
node --import dd-trace/register.js test.mjs
```

**Result:**
```
ReferenceError: require is not defined in ES module scope
    at file:///node_modules/@langchain/langgraph/dist/pregel/index.js:1:52
```

## Workaround

Use dd-trace 5.90.0, which has the `node:` prefix normalization fix (PR #6910) but does not include the langgraph rewriter (added in 5.91.0).

## Expected behavior

The rewriter should use `import` instead of `require()` when transforming ESM modules, or skip rewriting ESM files entirely if ESM-compatible injection is not supported.

BridgeAR:
When handling ESM files, we may not introduce CJS in those. Instead use import instead.

Fixes: https://github.com/DataDog/dd-trace-js/issues/7991

We could also normalize somewhere else, but this should be a simple fix to accept both.

### DIFF
--- a/packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
+++ b/packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
@@ -19,7 +19,7 @@ const transforms = module.exports = {
       : `const {tracingChannel: tr_ch_apm_tracingChannel} = require("${dcModule}")`
 
     node.body.splice(index + 1, 0, parse(code, {
-      sourceType: isModuleSourceType(sourceType) ? 'module' : 'script',
+      isModule: isModuleSourceType(sourceType),
     }).body[0])
   },
 
--- a/packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
+++ b/packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
@@ -2,7 +2,7 @@
 
 // TODO: Move traceIterator to Orchestrion.
 
-const { isModuleSourceType, parse, query, traverse } = require('./compiler')
+const { parse, query, traverse } = require('./compiler')
 
 const tracingChannelPredicate = (node) => (
   node.specifiers?.[0]?.local?.name === 'tr_ch_apm_tracingChannel' ||
@@ -51,6 +51,13 @@ function traceAny (state, node, _parent, ancestry) {
   } else {
     traceFunction(state, node, program)
   }
+}
+
+/**
+ * @param {string} sourceType
+ */
+function isModuleSourceType (sourceType) {
+  return sourceType === 'module' || sourceType === 'esm'
 }
 
 function traceFunction (state, node, program) {
--- a/packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
+++ b/packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
@@ -19,7 +19,7 @@ const transforms = module.exports = {
       : `const {tracingChannel: tr_ch_apm_tracingChannel} = require("${dcModule}")`
 
     node.body.splice(index + 1, 0, parse(code, {
-      sourceType: isModuleSourceType(sourceType) ? 'module' : 'script'
+      sourceType: isModuleSourceType(sourceType) ? 'module' : 'script',
     }).body[0])
   },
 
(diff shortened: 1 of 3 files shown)


bengl (approved):

bengl (approved):

bengl (approved):

bengl (approved):

packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
```diff
@@ -14,11 +14,13 @@ const transforms = module.exports = {
     if (node.body.some(tracingChannelPredicate)) return
 
     const index = node.body.findIndex(child => child.directive === 'use strict')
-    const code = sourceType === 'module'
+    const code = isModuleSourceType(sourceType)
```
rochdev, line 17: Let's inline the check, the compiler is supposed to be equivalent to what would be exposed by a real compiler so that it's soon moved out.

rochdev (changes requested):

packages/datadog-instrumentations/src/helpers/rewriter/transforms.js
BridgeAR: In three spots? I would rather use the abstraction instead of duplicating the check multiple times

rochdev: Then make it a local function, it doesn't belong in the compiler.

BridgeAR: With our current structure, I think the compiler file is the most suiting for what this is doing. I could theoretically add a helper method, that seems a bit much for this though and having a local check twice removes the purpose of it being not repeated.

rochdev: It's not its responsibility to understand what every possible source can have as values. It has a well defined API, and the source should pass the right values. For example, other parsers have sometimes just a `module: true` flag, and others go with the file type. If a source receives `esm` then it should convert it to `module`. Then the conversion is only needed in one place, and the compiler is not polluted as it will eventually be moved out.

rochdev (approved):

PR #8001, opened 2026-04-14
feat(opentelemetry): OTLP trace metrics export

ida613:
### What does this PR do?
Export span stats as OTLP metrics (dd.trace.span.hits, errors, top_level_hits, duration) via a new OtlpStatsExporter alongside the existing Datadog /v0.6/stats exporter.

### Key changes
packages/dd-trace/src/span_stats.js
- creates OtlpStatsExporter when enabled
- `_serializeBuckets` -> `drainBuckets`: the old method did two things at once: it formatted the bucket data into the Datadog wire shape `({ Start, Duration, Stats:[...toJSON()] })` and cleared the map. The new `_drainBuckets()` only drains the raw data as `[{ timeNs, bucket }]` and clears. Formatting is deferred to `onInterval()` so each exporter can receive the data in whatever shape it needs.

### Additional Notes
- No mapping from DDSketch indices to OTLP ExponentialHistogram scale/bucket indices is defined. `DD_TRACE_OTEL_METRICS_HISTOGRAM_TYPE` defaults to `explicit`. Conversion to ExponentialHistorogram deferred until DDSketch→OTLP index mapping is specified.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1385,6 +1385,34 @@ declare namespace tracer {
     stats?: boolean
 
     /**
+     * Configuration for OTLP span metrics export.
+     * When enabled, span metrics (hits, errors, duration) are exported in OTLP format.
+     * Requires DD_TRACE_STATS_COMPUTATION_ENABLED or sets up its own span aggregation.
+     */
+    traceMetrics?: {
+      /**
+       * Whether to enable OTLP span metrics export.
+       * @default false
+       * @env DD_TRACE_OTEL_METRICS_ENABLED
+       */
+      enabled?: boolean
+
+      /**
+       * OTLP metrics endpoint URL.
+       * @default "http://localhost:4318/v1/metrics"
+       * @env DD_TRACE_OTEL_METRICS_URL, OTEL_EXPORTER_OTLP_METRICS_ENDPOINT, OTEL_EXPORTER_OTLP_ENDPOINT
+       */
+      url?: string
+
+      /**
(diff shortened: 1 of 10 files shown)


index.d.ts
```diff
@@ -1384,6 +1384,34 @@ declare namespace tracer {
...
+       */
+      url?: string
+
+      /**
+       * OTLP transport protocol. gRPC is not supported and will fall back to http/protobuf.
+       * @default "http/protobuf"
+       * @env DD_TRACE_OTEL_METRICS_PROTOCOL, OTEL_EXPORTER_OTLP_METRICS_PROTOCOL, OTEL_EXPORTER_OTLP_PROTOCOL
+       */
+      protocol?: 'http/protobuf' | 'http/json' | 'grpc'
+    }
```
BridgeAR, line 1413: I would prefer to not expose any new programmatic APIs, if we do not have to. Is this required?

packages/dd-trace/src/config/supported-configurations.json
```diff
@@ -3273,6 +3273,58 @@
         "default": "false"
       }
     ],
+    "DD_TRACE_OTEL_METRICS_ENABLED": [
+      {
+        "implementation": "A",
+        "type": "boolean",
+        "default": "false",
+        "internalPropertyName": "traceMetrics.enabled",
```
BridgeAR, line 3281: This is not needed, if already using `configurationNames`. The same for the other entries.

Could you please open a PR that actually adds a linter to detect these and prevents adding identical entries for `internalPropertyName` and the first `configurationNames` entry?

PR #8010, opened 2026-04-15
chore(scope) remove tracer.scope from tests and source files

pabloerhard:
### What does this PR do?

This PR removes the usage of tracer.scope() from test and source files and replaces it with storage. This will support upcoming efforts for the span API split.


BridgeAR (approved):
LGTM

Is it planned to remove/deprecate the public API later as well?

pabloerhard:
> LGTM
> 
> Is it planned to remove/deprecate the public API later as well?

No, but tracer.scope will be limited to external usage and will now return a PublicSpan, hence the need to separate usage: PublicSpans solely for external use and internal spans solely for internal use, with a clear separation between the two.

PR #8012, opened 2026-04-15
feat(aws): aws-durable-execution-sdk-js instrumentation

joeyzhao2018:
## Summary

Adds automatic instrumentation for the `@aws/durable-execution-sdk-js` library.

## Traced Operations

| Target | Span Name | Kind |
| ------ | --------- | ---- |
| `withDurableExecution()` wrapper | aws.durable.execute | internal |
| `DurableContext.step()` | aws.durable.step | internal |
| `DurableContext.invoke()` | aws.durable.invoke | client |
| `DurableContext.wait()` | aws.durable.wait | internal |
| `DurableContext.waitForCondition()` | aws.durable.wait_for_condition | internal |
| `DurableContext.waitForCallback()` | aws.durable.wait_for_callback | internal |
| `DurableContext.createCallback()` | aws.durable.create_callback | internal |
| `DurableContext.map()` | aws.durable.map | internal |
| `DurableContext.parallel()` | aws.durable.parallel | internal |
| `DurableContext.runInChildContext()` | aws.durable.child_context | internal |

### Lazy DurablePromise observation

`DurableContext` methods return `DurablePromise` instances that delay the actual execution until user code first awaits/chains them. The integration uses `kind: 'Sync'` and an `observeDurablePromise` helper that wraps `then`/`catch`/`finally`. The actual span-finishing chain only attaches after user code first awaits.

### Suspended (PENDING) workflows

When execution is suspended, and the SDK intentionally leaves the suspended operation's `DurablePromise` neither resolved nor rejected. So the handler plugin detects `PENDING` and finishes any still-open child spans owned by this integration before returning, so the trace processor can flush the partial trace.

### Checkpoint-driven retry finalization

On a step retry, the SDK doesn't reject the step's `DurablePromise`, it suspends execution and writes a checkpoint with `Action: 'RETRY'`. The `CheckpointManager.checkpoint` hook detects that, stamps `error.{type,message,stack}` on the still-open span using the checkpoint's error payload, and finishes the span on `asyncEnd`.

### map/parallel resource cardinality + child_context suppression

For `map()` / `parallel()` / `waitForCallback()` the SDK calls `runInChildContext` with internal `subType`s (`Map`, `Parallel`, `MapIteration`, `ParallelBranch`, `WaitForCallback`). Those internal spans are suppressed. Resource is also removed from direct step children of `aws.durable.map`/`aws.durable.parallel` to avoid unbounded resource cardinality from per-iteration operation names.


.github/workflows/apm-integrations.yml
```diff
@@ -147,6 +147,14 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - uses: ./.github/actions/plugins/test-and-upstream
 
+  aws-durable-execution-sdk-js:
```
joeyzhao2018, line 150: This is for tests added in [datadog-plugin-aws-durable-execution-sdk-js/test](/packages/datadog-plugin-aws-durable-execution-sdk-js/test/)

packages/datadog-plugin-aws-durable-execution-sdk-js/src/util.js
```diff
@@ -0,0 +1,80 @@
...
+ * DurableContextImpl is about to run, or undefined if unavailable. Mirrors the
+ * SDK's internal calculations
+ * @param {object} [ctxImpl]
+ * @returns {string|undefined}
+ */
+function getOperationId (ctxImpl) {
+  const stepId = ctxImpl?.getNextStepId?.()
+  if (!stepId) return
+  return createHash('md5').update(stepId).digest('hex').slice(0, 16)
+}
```
pablomartinezbernardo: Is replicating this acceptable? It's defined here, but doesn't seem like we can access it https://github.com/aws/aws-durable-execution-sdk-js/blob/main/packages/aws-durable-execution-sdk-js/src/utils/step-id-utils/step-id-utils.ts

.github/workflows/apm-integrations.yml
```diff
@@ -149,6 +149,16 @@ jobs:
...
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+    env:
+      PLUGINS: aws-durable-execution-sdk-js
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: ./.github/actions/plugins/test
```
wconti27: Should be added to `serverless.yml` to live with the other serverless integration tests

packages/datadog-instrumentations/src/aws-durable-execution-sdk-js.js
```diff
@@ -0,0 +1,5 @@
+'use strict'
+
+const { addHook } = require('./helpers/instrument')
+
+addHook({ name: '@aws/durable-execution-sdk-js', versions: ['>=1.1.0'] }, exports => exports)
```
wconti27: I would expect this file to be just registering the no-op hooks (since orchestrion is being used)

```
'use strict'

const { addHook, getHooks } = require('./helpers/instrument')

for (const hook of getHooks('@aws/durable-execution-sdk-js')) {
  addHook(hook, exports => exports)
}

```

packages/datadog-plugin-aws-durable-execution-sdk-js/src/checkpoint.js
```diff
@@ -0,0 +1,31 @@
...
+// On retries the SDK suspends execution without firing error/asyncEnd; finish the span here.
+class AwsDurableExecutionSdkJsCheckpointPlugin extends TracingPlugin {
+  static id = 'aws-durable-execution-sdk-js'
+  static prefix = 'tracing:orchestrion:@aws/durable-execution-sdk-js:CheckpointManager_checkpoint'
+
+  start (ctx) {
+    const data = ctx?.arguments?.[1]
+    if (data?.Action !== 'RETRY' || !data.Error) return
+
+    const span = this.activeSpan
```
wconti27, line 14:
```suggestion
    const span = ctx.currentStore?.span
```

packages/datadog-plugin-aws-durable-execution-sdk-js/src/util.js
wconti27: Should be fine I would think

scripts/install_plugin_modules.js
```diff
@@ -188,16 +190,16 @@ async function assertPeerDependencies (rootFolder, parent = '') {
...
       if (node && !semver.satisfies(process.versions.node, node)) {
         continue
       }
-      if (!pkgJsonPath) {
-        pkgJsonPath = require(folder).pkgJsonPath()
-        pkgJson = require(pkgJsonPath)
+      if (!pkgJson) {
+        const requireFromWorkspace = createRequire(join(folder, 'package.json'))
+        const nodeModulesPaths = requireFromWorkspace.resolve.paths(externalName)
+        pkgJson = requirePackageJson(externalName, { paths: nodeModulesPaths })
```
wconti27, line 204: This should be handled under the hood already and is unnecessary.

packages/datadog-plugin-aws-durable-execution-sdk-js/test/index.spec.js
```diff
@@ -0,0 +1,334 @@
...
+  category: 'orchestration',
+}, (meta) => {
+  const { agent } = meta
+
+  beforeEach(async () => setup(meta.mod, meta.versionMod))
+  afterEach(async () => teardown())
+
+  describe('withDurableExecution() - aws.durable.execute', () => {
+    it('happy: emits invocation_status=succeeded, replayed=false', async () => {
+      const trace = agent.assertSomeTraces(traces => assertSpanByName(traces, {
```
wconti27: nit, call this variable `tracePromise`, likewise for the remainder of the file

packages/datadog-plugin-aws-durable-execution-sdk-js/src/checkpoint.js
pablomartinezbernardo: `AwsDurableExecutionSdkJsCheckpointPlugin` doesn't create spans, it just enriches the span. As it doesn't create a span, `ctx.currentStore.span` will be undefined

scripts/install_plugin_modules.js
pablomartinezbernardo: Oh yeah forgot to mention this

aws-durable-execution-sdk-js [only exports](https://github.com/aws/aws-durable-execution-sdk-js/blob/f445772712d4bb40f53b1d0eed9e0954b30d0c7c/packages/aws-durable-execution-sdk-js/package.json#L18-L24) `.`

`pkgJsonPath` does `require.resolve('@aws/durable-execution-sdk-js/package.json')` which throws `ERR_PACKAGE_PATH_NOT_EXPORTED`.

From what I've seen the way to solve properly would be this https://nodejs.org/api/module.html#modulefindpackagejsonspecifier-base but... node 22

packages/datadog-plugin-aws-durable-execution-sdk-js/src/checkpoint.js
```diff
@@ -0,0 +1,31 @@
...
+    if (!span || span._spanContext?._tags?.error) return
+
+    const { ErrorMessage, ErrorType, StackTrace } = data.Error
+    span.setTag('error', 1)
```
rochdev, line 18: Why are we not creating a span? Otherwise the only options are overwriting an existing error or dropping it completely, both of which are not ideal.

packages/datadog-plugin-aws-durable-execution-sdk-js/src/checkpoint.js
```diff
@@ -0,0 +1,31 @@
...
+    if (ErrorMessage) span.setTag('error.message', ErrorMessage)
+    if (ErrorType) span.setTag('error.type', ErrorType)
+    if (Array.isArray(StackTrace)) span.setTag('error.stack', StackTrace.join('\n'))
+
+    ctx._ddRetryStepSpan = span
```
rochdev: Why use an underscore property on the context?

packages/datadog-plugin-aws-durable-execution-sdk-js/src/util.js
```diff
@@ -0,0 +1,80 @@
...
+ * Using `kind: 'Async'` in Orchestrion would side-chain `.then()` immediately on the returned
+ * thenable, prematurely triggering the SDK's `ensureExecution()` and `markOperationAwaited`.
+ * Callers pair `kind: 'Sync'` with this helper so `onSettle` only fires after user code first
+ * awaits / chains, preserving the SDK's lazy semantics.
+ * @param {object} dp - The returned DurablePromise instance.
+ * @param {(err: unknown) => void} onSettle - Called once with `undefined` on success or the
+ *   rejection reason on failure.
+ * @returns {void}
+ */
+function observeDurablePromise (dp, onSettle) {
```
rochdev: We should not be patching on the plugin side. Any way to instead patch the thenable directly at the source with a corresponding additional channel?

scripts/install_plugin_modules.js
rochdev: A lot of libraries do this, why are we only hitting this issue now?

packages/datadog-plugin-aws-durable-execution-sdk-js/src/client.js
```diff
@@ -0,0 +1,64 @@
...
+      }
+      this.finish(ctx)
+    })
+  }
+
+  error (ctxOrError) {
+    if (ctxOrError?._ddFinished) return
+    if (ctxOrError && typeof ctxOrError === 'object') ctxOrError._ddFinished = true
+    super.error(unwrapDurableError(ctxOrError))
+    super.finish(ctxOrError)
```
rochdev: Why call super for a different method?

scripts/install_plugin_modules.js
pablomartinezbernardo: Because this doesn't happen that often: `assertPeerDependencies` is only run for libraries with `dep: true` in [externals.js](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/test/plugins/externals.js). And from all those, this is the only case where `exports` is used without exporting the `package.json`, but could happen for other libraries as well

packages/datadog-plugin-aws-durable-execution-sdk-js/src/checkpoint.js
pablomartinezbernardo: > Why are we not creating a span? 

Because checkpointing is not an operation we really want to trace for the customer, the user doesn't run the checkpoint operation: the SDK checkpoints automatically after each step execution.

The error is first visible at the beginning of the checkpoint. Steps don't throw themselves when there is a user error (They could throw if there's a bug in the SDK when creating the promise, though). This condition shouldn't ever happen, no error should be set on the span prior to us seeing the checkpoint.

Do you think removing the condition would make it clearer?

packages/datadog-plugin-aws-durable-execution-sdk-js/src/checkpoint.js
pablomartinezbernardo: To be able to `finish` the span in the `asyncEnd`. We could also either
a) finish it in `start`, the difference in time wouldn't be that big, but it would finish earlier
b) actually do `this.activeSpan?.finish()` on `asyncEnd`, but we would need to check again for conditions like `data?.Action !== 'RETRY'`

Would you rather see any of these? Or maybe I'm misunderstanding your question?

packages/datadog-plugin-aws-durable-execution-sdk-js/src/util.js
pablomartinezbernardo: Thanks, makes sense, addressed in 5289e4d798d0ec467695d2df474aaee0619a51c6 can you have a look?

rochdev:
While investigating flaky CI jobs across the repo, I noticed this PR's tests are showing intermittent failures in the Serverless workflow.

The failures appear in `packages/datadog-plugin-aws-durable-execution-sdk-js/test/index.spec.js:21` with assertion errors like:

- `'pending' !== 'succeeded'` — execution status hasn't resolved to `succeeded` by the time the test checks it
- `'false' !== 'true'` — likely a `replayed` flag or similar field that isn't set yet

This looks like an async timing race where the durable execution hasn't finished before the span is asserted. Seen in runs [26032888063](https://github.com/DataDog/dd-trace-js/actions/runs/26032888063) and [26033817706](https://github.com/DataDog/dd-trace-js/actions/runs/26033817706).

_Comment generated with [Claude Code](https://claude.com/claude-code)_

joeyzhao2018:
> While investigating flaky CI jobs across the repo, I noticed this PR's tests are showing intermittent failures in the Serverless workflow.
> 
> The failures appear in `packages/datadog-plugin-aws-durable-execution-sdk-js/test/index.spec.js:21` with assertion errors like:
> 
> * `'pending' !== 'succeeded'` — execution status hasn't resolved to `succeeded` by the time the test checks it
> * `'false' !== 'true'` — likely a `replayed` flag or similar field that isn't set yet
> 
> This looks like an async timing race where the durable execution hasn't finished before the span is asserted. Seen in runs [26032888063](https://github.com/DataDog/dd-trace-js/actions/runs/26032888063) and [26033817706](https://github.com/DataDog/dd-trace-js/actions/runs/26033817706).
> 
> _Comment generated with [Claude Code](https://claude.com/claude-code)_

@rochdev @pablomartinezbernardo  
I think there's a problem with the aws sdk testing tools. With Claude-code's help i open this PR to the aws sdk https://github.com/aws/aws-durable-execution-sdk-js/pull/544.
In my branch, I'm just [skipping these tests for now with a note pointing to that fix PR.](https://github.com/DataDog/dd-trace-js/pull/8182/changes/8baa8ce12580c3686a1f367f7b78b62389b43c19)

pablomartinezbernardo:
@joeyzhao2018 I didn't get in the details of that specific issue, but in any case that would be unrelated to what was happening here: I was asserting on the first span found by name, in this case looking for the `execute` span. In durable executions there may be multiple execute spans, but we are looking for a specific one. I fixed it in c1bc5ce4b2d6f342ed9a95d9d80a1edd74356168

packages/datadog-plugin-aws-durable-execution-sdk-js/src/checkpoint.js
rochdev: We trace what the _process_ is doing, not what the _user_ is doing, so if checkpoints are happening shouldn't we be tracing them? Especially if it's an operation that can fail. Setting the error on another span is way more confusing.

packages/datadog-plugin-aws-durable-execution-sdk-js/src/checkpoint.js
rochdev: What I meant was why is the property underscored?

pablomartinezbernardo: Oh, because claude (seemingly incorrectly) chose it, and it looked reasonable. Seems like this PR is the first and only one that introduced this pattern, maybe we want to update it as well so AIs don't get confused? https://github.com/DataDog/dd-trace-js/pull/7669

In any case, updated the name in my PR.

packages/datadog-plugin-aws-durable-execution-sdk-js/src/checkpoint.js
pablomartinezbernardo: The `data.Error` here isn't a checkpoint failure, it's the user step's error object. On Action: `RETRY` the SDK doesn't throw or settle the step's promise, so the step's settle channel never fires. The checkpoint is where we can observe the user error. This is also the reason why we are finishing the span here.

Checkpoint operations themselves can fail only because of network/AWS errors, but they are already traced as a generic `aws.request`, they don't contain additional information relevant to durable functions other than the error on the step we are retrieving from here.

packages/datadog-plugin-aws-durable-execution-sdk-js/src/util.js
tlhunter: It's not ideal due to the CPU overhead.

Are we able to modify the source code of the file generating the hash to be able to access it? E.g. using orchestrion?

scripts/install_plugin_modules.js
tlhunter: I have a feeling this won't work properly when the customer bundles their application. Can you throw an acceptance test into the esbuild integration tests directory that installs this package? You should see other examples in there when we discover a tricky package.

packages/datadog-plugin-aws-durable-execution-sdk-js/src/handler.js
```diff
@@ -0,0 +1,61 @@
...
+  static prefix = 'tracing:orchestrion:@aws/durable-execution-sdk-js:withDurableExecution'
+
+  bindStart (ctx) {
+    const args = ctx.arguments || []
+    const event = args[0]
+    const durableExecutionMode = args[3]
+    const handler = args[5]
+
+    const meta = {
+      'aws.durable.replayed': String(durableExecutionMode === 'ReplayMode'),
```
BridgeAR: Nit
```suggestion
      'aws.durable.replayed': durableExecutionMode === 'ReplayMode' ? 'true' : 'false',
```

BridgeAR (commented):
Besides existing comments, this looks pretty solid! I did not review the tests closely though. I might have missed that but I believe we could also add an ESM test?

packages/datadog-plugin-aws-durable-execution-sdk-js/src/util.js
pablomartinezbernardo: Is the CPU overhead the issue? Even if we were able to access it using orchestrion (which doesn't seem to be the case for being private), and we could call it, we would just be running the same code.

This hash is used for 2 things by aws
* Accessing replayed operations, the unique hash is used to have direct access in a dict
* Showing it in the UI

There is no place where the hash is calculated for fresh (meaning non-replayed) operations, there's no place where we can hook onto to get this value for fresh operations.

The drawback I see from this is actually having to replicate code in our tracer, but I can't see a way to avoid it if we need to have the operation hash.

BridgeAR: With Orchestrion-JS we should be able to replace existing code of the library. Thus, we would be able to intercept the original implementation without additional overhead.

About CPU overhead: we are currently working hard to get the tracer to perform super fast.

scripts/install_plugin_modules.js
pablomartinezbernardo: Added the acceptance test e1a2f7cb40a528f9cfe2fff4b6b0ca0af092eb34 . The exports restriction shouldn't affect bundling, no? Both the esbuild plugin and the runtime resolver read package.json off disk, this fix is only for CI workspace generation.

pablomartinezbernardo:
> Besides existing comments, this looks pretty solid! I did not review the tests closely though. I might have missed that but I believe we could also add an ESM test?

@BridgeAR Added them a49360e0b504bc5a23685016b645507a10a42465 !

packages/datadog-plugin-aws-durable-execution-sdk-js/src/util.js
pablomartinezbernardo: The issue is that there is no good place to intercept this function call, it's not reliably called at any point in the process where we can correlate it with the span.

If this is unacceptable, the alternative may actually be removing this field. This will impact other components though

tlhunter (dismissed):

packages/datadog-plugin-aws-durable-execution-sdk-js/src/util.js
```diff
@@ -0,0 +1,48 @@
...
+ * @returns {{ error?: unknown } | unknown}
+ */
+function unwrapDurableError (ctxOrError) {
+  let err = ctxOrError?.error
+  if (!(err instanceof Error)) return ctxOrError
+
+  while (typeof err.errorType === 'string' && err.cause instanceof Error) {
+    err = err.cause
+  }
+  return { ...ctxOrError, error: err }
```
BridgeAR:
```suggestion
  return err
```

I think we only need the error :)

packages/datadog-plugin-aws-durable-execution-sdk-js/src/checkpoint.js
BridgeAR: Are we certain the activeSpan is the correct checkpoint part? Could the checkpoint run outside of original context? We would place the checkpoint on the wrong span in that case.

packages/datadog-instrumentations/src/helpers/hooks.js
```diff
@@ -2,6 +2,7 @@
 
 module.exports = {
   // Only list unprefixed node modules. They will automatically be instrumented as prefixed and unprefixed.
+  '@aws/durable-execution-sdk-js': () => require('../aws-durable-execution-sdk-js'),
```
BridgeAR: Would you mind moving this down to separate Node.js modules from the rest? :)

packages/datadog-plugin-aws-durable-execution-sdk-js/src/checkpoint.js
```diff
@@ -0,0 +1,31 @@
...
+
+const TracingPlugin = require('../../dd-trace/src/plugins/tracing')
+
+// On retries the SDK suspends execution without firing error/asyncEnd; finish the span here.
+class AwsDurableExecutionSdkJsCheckpointPlugin extends TracingPlugin {
+  static id = 'aws-durable-execution-sdk-js'
+  static prefix = 'tracing:orchestrion:@aws/durable-execution-sdk-js:CheckpointManager_checkpoint'
+
+  start (ctx) {
+    const data = ctx?.arguments?.[1]
```
BridgeAR: Is the context defined by us? I believe we could just drop the optional chaining in that case? 


```suggestion
    const data = ctx.arguments?.[1]
```

packages/datadog-instrumentations/src/aws-durable-execution-sdk-js.js
```diff
@@ -0,0 +1,69 @@
...
+  // instance-level wrapper. The promise can only settle once, so calling
+  // attachSpy at most once gives us exactly one onSettle invocation.
+  const attachSpy = () => {
+    if (attached) return
+    attached = true
+    proto.then.call(dp, () => onSettle(), err => onSettle(err))
+  }
+
+  shimmer.massWrap(dp, ['then', 'catch', 'finally'], original => function (...args) {
+    attachSpy()
```
BridgeAR: I think we could change the prototype of the promise and prevent rewrapping each promise afterwards. That should likely be more performant :)

packages/datadog-plugin-aws-durable-execution-sdk-js/src/util.js
```diff
@@ -0,0 +1,48 @@
...
+const { createHash } = require('node:crypto')
+
+/**
+ * Returns true if the op the DurableContextImpl is about to run will be served
+ * from the SDK's checkpoint (i.e. the next stepId already has a SUCCEEDED entry).
+ * @param {object} [ctxImpl]
+ * @returns {boolean}
+ */
+function isReplayedOp (ctxImpl) {
+  const stepId = ctxImpl?.getNextStepId?.()
```
BridgeAR, line 16: I believe the getNextStepId method is currently called twice in many cases where we could unify it to a single call for performance

packages/datadog-instrumentations/src/helpers/rewriter/instrumentations/aws-durable-execution-sdk-js.js
```diff
@@ -0,0 +1,41 @@
...
+  'step',
+  'invoke',
+  'runInChildContext',
+  'wait',
+  'waitForCondition',
+  'waitForCallback',
+  'createCallback',
+  'map',
+  'parallel',
+]
```
BridgeAR: Could we reuse this list to prevent the duplication? :)

packages/datadog-plugin-aws-durable-execution-sdk-js/src/checkpoint.js
pablomartinezbernardo: We are certain, checkpointing is done synchronously here

packages/datadog-instrumentations/src/helpers/hooks.js
pablomartinezbernardo: Of course 🤦  completely missed this

BridgeAR (dismissed):
LGTM! Good work :)

BridgeAR (approved):

PR #8014, opened 2026-04-15
feat(redis): split by instance

pabloerhard:
### What does this PR do?

This PR adds splitByInstance config into the Redis plugin and solves #7914.

The service naming layer already had full support for splitByInstance on the redis plugin, (shared redisConfig with ioredis), so no schema changes were needed. 

Fixes: #7914


packages/datadog-instrumentations/src/redis.js
```diff
@@ -47,6 +48,7 @@ function wrapCommandQueueClass (cls) {
         }
       }
       this._url = this._url || { host: 'localhost', port: 6379 }
+      this._connectionName = createClientName
```
BridgeAR: I would rather not manipulate the library in that way. Instead, could you please add a WeakMap with the instance as key and also add the url to that one?

packages/datadog-instrumentations/src/redis.js
```diff
@@ -134,12 +138,13 @@ addHook({ name: 'redis', versions: ['>=0.12 <2.6'] }, redis => {
   return redis
 })
 
-function getStartCtx (client, command, args, url = {}) {
+function getStartCtx (client, command, args, url = {}, connectionName) {
```
BridgeAR: No need for an additional argument here. We already have access to the instance and that is our key to access the property :)

packages/datadog-plugin-redis/test/client.spec.js
```diff
@@ -210,6 +209,63 @@ describe('Plugin', () => {
...
+          await promise
+        })
+
+        it('should set service source tag to split-by-instance', async () => {
+          const promise = agent.assertSomeTraces(traces => {
+            assert.strictEqual(traces[0][0].meta['_dd.svc_src'], 'opt.split_by_instance')
+          })
+
+          await client.get('foo')
+          await promise
```
BridgeAR: Nit: in theory these entries could resolve in the wrong order in which case this would end up as unhandled rejection.

The same applies to all other similar entries.

```suggestion
          await Promise.all([
            client.get('foo'),
            promise,
          ])
```

BridgeAR (commented):
Almost LGTM. I think we could just clean up some code a bit more :)

BridgeAR:
I updated the description so that the related issue will automatically close as resolved when this lands.

packages/datadog-instrumentations/src/redis.js
pabloerhard: Is your concern specifically about adding _url and _connectionName as properties on the subclassed instance?

BridgeAR (dismissed):
LGTM, thanks for the cleanup!

BridgeAR (approved):

PR #8016, opened 2026-04-16
chore(ci) : do not cancel jobs on master

cbeauchesne:
### What does this PR do?
Set `cancel-in-progress` to `false` when the `ref` is `master` on all workflows that previously had inconditionnal `cancel-in-progress: true`

Also simplifies the system-tests.yml concurrency block, which was using a more verbose conditional expression to achieve the same result.   

### Motivation

The concurrency cancellation mechanism is useful on feature branches to avoid running stale CI when new commits are pushed. However, on master it is harmful: a push that kicks off CI right after a merge should not cancel the CI from the previous merge. Every commit to master must have its workflows run to completion.


.github/workflows/system-tests.yml
```diff
@@ -10,10 +10,8 @@ on:
 
 concurrency:
   # this ensures that only one workflow runs at a time for a given branch on pull requests
-  # as the group key is the workflow name and the branch name
-  # for scheduled runs and pushes to master, we use the run id to ensure that all runs are executed
-  group: ${{ (github.event_name == 'pull_request' && format('{0}-{1}', github.workflow, github.ref)) || format('{0}-{1}', github.workflow, github.run_id) }}
```
BridgeAR, line 15: Is this not changing semantics for other branches?

BridgeAR (approved):
LGTM, while it would be good to look into the comment before landing

.github/workflows/system-tests.yml
cbeauchesne: The logic was already applied on system-tests, thanks to this hacky stuff (when the event is not `pull_request`, using `run_id`  in the group name, which is unique per pipeline).

I changed it to use the same logic than everywhere else : `group` does not carry any logic, only `cancel-in-progress` has a value that depends on the branch name, which sound way cleaner to me.

PR #8017, opened 2026-04-16
test(sampling): verify formatKnuthRate is locale-independent

bm1549:
### What does this PR do?

Exports `formatKnuthRate` from `priority_sampler.js` and adds direct unit tests that verify the function produces correct decimal-dot output for various sampling rates.

### Motivation

`formatKnuthRate` uses `Number.prototype.toFixed()`, which is locale-independent per ECMAScript spec (ECMA-262, sec 21.1.3.3) -- it always uses `.` as the decimal separator. This PR documents that contract with explicit tests covering normal decimals (0.3), integer rates (1.0), minimum precision (0.000001), and below-precision rounding (0.0000001).

### Additional Notes

- The named export (`module.exports.formatKnuthRate`) is additive and does not affect the default `PrioritySampler` export.
- All 57 tests in `priority_sampler.spec.js` pass, including the 4 new ones.

### DIFF
--- a/packages/dd-trace/src/priority_sampler.js
+++ b/packages/dd-trace/src/priority_sampler.js
@@ -393,3 +393,4 @@ class PrioritySampler {
 }
 
 module.exports = PrioritySampler
+module.exports.formatKnuthRate = formatKnuthRate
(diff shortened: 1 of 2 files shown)


packages/dd-trace/src/priority_sampler.js
```diff
@@ -393,3 +393,4 @@ class PrioritySampler {
 }
 
 module.exports = PrioritySampler
+module.exports.formatKnuthRate = formatKnuthRate
```
BridgeAR, line 396: Exporting internals for testing is something that I would rather prevent, if possible

PR #8019, opened 2026-04-16
chore: do not report isServiceNameInferred telemetry

BridgeAR:
It is an implementation detail that is not a configuration.


packages/dd-trace/test/config/index.spec.js
```diff
@@ -1518,11 +1518,21 @@ describe('Config', () => {
...
+      ['DD_TRACE_STATS_COMPUTATION_ENABLED', 'calculated'],
+    ])
+
+    for (const entry of updateConfig.getCall(0).args[0]) {
+      const origin = internalDetails.get(entry.name)
+      if (origin !== undefined) {
+        if (origin === null) {
+          assert.fail(`Expected ${entry.name} to have origin ${origin} but got ${entry.origin}`)
+        } else {
+          assert.notStrictEqual(origin, entry.origin)
```
szegedi, line 1532: Why not strict?

szegedi (approved):
LGTM, just curious about a non-strict assertion.

packages/dd-trace/test/config/index.spec.js
BridgeAR: This is for `DD_TRACE_STATS_COMPUTATION_ENABLED`. It should only be reported for default, if I am not mistaken. We could change the logic to check for that, while it would not work for other envs in case we do pass it through somehow. In that case we need multiple entries.

What I originally checked for was that calculated is not emitted.

PR #8020, opened 2026-04-16
fix(node) appsec and capabilities tests

pabloerhard:
### What does this PR do?

This PR fixes tests failing after node 24.15.0 were released.

Capabilities tests are failing as consequence of ( https://github.com/nodejs/node/pull/62030 ) which changed the way path defaults to '/' when not specified.

AppSec tests started failing as consequence of ( https://github.com/node/pull/61770 ) this change moved the registration of socket error listeners to an earlier step in the lifecycle. Combined with the test pattern of creating and rapidly destroying request, those listeners piled up on the same reused agent socket, excedding our low defaultMaxListeners threshold of 6.


BridgeAR (approved):
Thank you for the quick fixes!

PR #8029, opened 2026-04-19
fix: fix AppSec http memory retention

BridgeAR:
When AppSec is active, http calls have a long memory retention due
to a strong reference behind held. That reference is freed at the
end of a span for example, but it takes additional time and shows
in many memory profiles.

Fix that by using a weak reference and a specific handling to
circumvent that retention issue.

### DIFF
--- a/packages/dd-trace/src/appsec/index.js
+++ b/packages/dd-trace/src/appsec/index.js
@@ -40,6 +40,7 @@ const Reporter = require('./reporter')
 const appsecTelemetry = require('./telemetry')
 const apiSecuritySampler = require('./api_security_sampler')
 const { isBlocked, block, callBlockDelegation, setTemplates, getBlockingAction } = require('./blocking')
+const { getActiveRequest, getRequest } = require('./store')
 const UserTracking = require('./user_tracking')
 const graphql = require('./graphql')
 const rasp = require('./rasp')
@@ -117,7 +118,7 @@ function onRequestBodyParsed ({ req, res, body, abortController }) {
 
   if (!req) {
     const store = storage('legacy').getStore()
-    req = store?.req
+    req = getRequest(store)
   }
 
   const rootSpan = web.root(req)
@@ -258,8 +259,8 @@ function incomingHttpEndTranslator ({ req, res }) {
 }
 
 function onPassportVerify ({ framework, login, user, success, abortController }) {
-  const store = storage('legacy').getStore()
-  const rootSpan = store?.req && web.root(store.req)
+  const req = getActiveRequest()
+  const rootSpan = req && web.root(req)
 
   if (!rootSpan) {
     log.warn('[ASM] No rootSpan found in onPassportVerify')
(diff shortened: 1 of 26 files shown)


packages/dd-trace/src/appsec/index.js
```diff
@@ -117,7 +118,7 @@ function onRequestBodyParsed ({ req, res, body, abortController }) {
 
   if (!req) {
     const store = storage('legacy').getStore()
-    req = store?.req
+    req = getRequest(store)
```
IlyasShabi:
```suggestion
    req = getActiveRequest()
```

packages/dd-trace/src/appsec/index.js
```diff
@@ -309,7 +310,7 @@ function onRequestQueryParsed ({ req, res, query, abortController }) {
...
```
IlyasShabi:
```suggestion
    req = getActiveRequest()
```

packages/dd-trace/src/appsec/index.js
```diff
@@ -40,6 +40,7 @@ const Reporter = require('./reporter')
 const appsecTelemetry = require('./telemetry')
 const apiSecuritySampler = require('./api_security_sampler')
 const { isBlocked, block, callBlockDelegation, setTemplates, getBlockingAction } = require('./blocking')
+const { getActiveRequest, getRequest } = require('./store')
```
IlyasShabi: If we use `getActivRequest` we can remove `const { storage } = require('../../../datadog-core')` completely

packages/dd-trace/src/appsec/rasp/lfi.js
```diff
@@ -53,16 +55,18 @@ function analyzeLfiInResponseRender (ctx) {
   const store = storage('legacy').getStore()
   if (!store) return
```
IlyasShabi, line 56: Do we really need this?

packages/dd-trace/src/appsec/rasp/lfi.js
```diff
@@ -53,16 +55,18 @@ function analyzeLfiInResponseRender (ctx) {
...
 
-  analyzeLfiPath(ctx.view, ctx.req, store.res, ctx.abortController)
+  analyzeLfiPath(ctx.view, ctx.req, web.getContext(ctx.req)?.res, ctx.abortController)
 }
 
 function analyzeLfi (ctx) {
   const store = storage('legacy').getStore()
   if (!store) return
```
IlyasShabi: Ditto

packages/dd-trace/src/appsec/index.js
BridgeAR: Addressed

packages/dd-trace/src/appsec/index.js
BridgeAR: Addressed

packages/dd-trace/src/appsec/rasp/lfi.js
BridgeAR: We need the store for fs anyway, thus, I would keep it like that in the other place. Here, it is just a bit more future proof by working for any context.

packages/dd-trace/src/appsec/index.js
BridgeAR: Resolved

IlyasShabi (approved):

PR #8030, opened 2026-04-19
feat(otel): forward getActiveSpan() writes onto the active Datadog span

BridgeAR:
OTel consumers that retrieve a span via `trace.getActiveSpan()` and call
mutators on it expect those writes to land on whatever span is currently
active. When the active span was created by the Datadog tracer (i.e. all
DD-native instrumentations), the bridge previously returned a no-op span
proxy, so `getActiveSpan().setAttribute(...)` quietly went nowhere. The
pattern is documented in the OTel API and is used heavily by user code
that mixes manual OTel calls with auto-instrumentation, so this gap was
a frequent source of "my attribute didn't show up" reports.

Cache an `ActiveSpanProxy` per active context so `getActiveSpan()` hands
back the same proxy on every call, and forward all OTel mutators through
the same `span-helpers.js` surface that the bridge `Span` class uses.
Both bridge classes now extend a small `BridgeSpanBase` that owns the
shared method bodies (`setAttribute`, `setAttributes`, `addEvent`,
`addLink`, `addLinks`, `recordException`, `setStatus`, `isRecording`),
so the proxy and the OTel-created span cannot drift on the spec going
forward.

`updateName` is intentionally split between the two bridges:

- For OTel-created spans, `updateName` writes the DD operation name. The
  span exists because OTel asked for it, so the OTel name *is* the
  operation name.
- For the proxy wrapping a DD-native span, `updateName` writes
  `resource.name` instead. Touching the operation name on a span that an
  integration created would scramble metric aggregation in the backend
  (operation names drive the metric grouping); `resource.name` is the
  display field and is the correct target for "rename what I'm looking
  at" semantics.

The two paths share `span-helpers.js`'s `setOtelOperationName` and
`setOtelResource` so the gate logic still lives in one place.

Test coverage: the new proxy describe block in
`context_manager.spec.js` exercises the proxy lifecycle (caching across
calls, no-op once the underlying span has finished, the resource-name
path, attribute/event/link forwarding). `span-helpers.spec.js` gains
coverage for the two new name helpers. `span_context.js` records the
active OTel span on the DD span context (`_otelActiveSpan`) so the proxy
cache survives context propagation.

### DIFF
--- a/packages/dd-trace/src/opentelemetry/active-span-proxy.js
+++ b/packages/dd-trace/src/opentelemetry/active-span-proxy.js
@@ -1,0 +1,42 @@
+'use strict'
+
+const BridgeSpanBase = require('./bridge-span-base')
+const { setOtelResource } = require('./span-helpers')
+
+/**
+ * OTel `Span`-compatible proxy around an already-active Datadog span.
+ *
+ * Makes `trace.getActiveSpan()` forward attribute/link/event/status/exception writes onto
+ * the Datadog span. `end()` is intentionally a no-op: the span's lifecycle belongs to
+ * whoever created it. Mutation methods all bail out once the underlying Datadog span has
+ * finished (gated inside the helpers), matching OTel `Span` semantics.
+ */
+class ActiveSpanProxy extends BridgeSpanBase {
+  /** @type {import('./span_context')} */
+  #otelSpanContext
+
+  /**
+   * @param {import('../opentracing/span')} ddSpan
+   * @param {import('./span_context')} otelSpanContext
+   */
+  constructor (ddSpan, otelSpanContext) {
+    super(ddSpan)
+    this.#otelSpanContext = otelSpanContext
+  }
+
+  spanContext () {
+    return this.#otelSpanContext
+  }
+
+  /**
+   * @param {string} name
+   */
+  updateName (name) {
+    setOtelResource(this._ddSpan, name)
+    return this
+  }
+
+  end () {}
+}
+
+module.exports = ActiveSpanProxy
--- a/packages/dd-trace/src/opentelemetry/span-helpers.js
+++ b/packages/dd-trace/src/opentelemetry/span-helpers.js
@@ -268,6 +268,32 @@ function applyOtelStatus (ddSpan, currentCode, status) {
   return code
 }
 
+/**
+ * OTel `updateName` for OTel-created bridge spans: writes the DD operation name, matching
+ * the OTel SDK semantic that `updateName` updates the canonical span identifier.
+ *
+ * @param {import('../opentracing/span')} ddSpan
+ * @param {string} name
+ */
+function setOtelOperationName (ddSpan, name) {
+  if (!isWritable(ddSpan)) return
+
+  ddSpan.setOperationName(name)
+}
+
+/**
+ * OTel `updateName` for DD-native spans the bridge did not create: writes `resource.name`
+ * so the operation name (and the backend metric aggregation it drives) stays stable.
+ *
+ * @param {import('../opentracing/span')} ddSpan
+ * @param {string} name
+ */
+function setOtelResource (ddSpan, name) {
+  if (!isWritable(ddSpan)) return
+
+  ddSpan.setTag('resource.name', name)
+}
+
 module.exports = {
   addOtelEvent,
   addOtelLink,
@@ -277,4 +303,6 @@ module.exports = {
   recordException,
   setOtelAttribute,
   setOtelAttributes,
+  setOtelOperationName,
+  setOtelResource,
 }
--- a/packages/dd-trace/src/opentelemetry/span.js
+++ b/packages/dd-trace/src/opentelemetry/span.js
@@ -13,16 +13,9 @@ const { SERVICE_NAME, RESOURCE_NAME, SPAN_KIND } = require('../../../../ext/tags
 const kinds = require('../../../../ext/kinds')
 
 const id = require('../id')
+const BridgeSpanBase = require('./bridge-span-base')
 const SpanContext = require('./span_context')
-const {
-  addOtelEvent,
-  addOtelLink,
-  addOtelLinks,
-  applyOtelStatus,
-  recordException,
-  setOtelAttribute,
-  setOtelAttributes,
-} = require('./span-helpers')
+const { setOtelOperationName } = require('./span-helpers')
 
 const spanKindNames = {
   [api.SpanKind.INTERNAL]: kinds.INTERNAL,
@@ -124,10 +117,7 @@ function spanNameMapper (spanName, kind, attributes) {
  * OTel-bridge span backed by a `DatadogSpan`. `Tracer` constructs these on the OTel API
  * surface; the underlying DD span carries the lifecycle.
  */
-class Span {
-  // OTel SpanStatusCode: 0 = UNSET, 1 = OK, 2 = ERROR. Tracked for OK-is-final precedence.
-  #statusCode = 0
-
+class Span extends BridgeSpanBase {
   /**
    * @param {import('./tracer')} parentTracer
    * @param {import('@opentelemetry/api').Context} context
@@ -153,7 +143,7 @@ class Span {
     const hrStartTime = timeInputToHrTime(timeInput || (performance.now() + timeOrigin))
     const startTime = hrTimeToMilliseconds(hrStartTime)
 
-    this._ddSpan = new DatadogSpan(_tracer, _tracer._processor, _tracer._prioritySampler, {
+    const ddSpan = new DatadogSpan(_tracer, _tracer._processor, _tracer._prioritySampler, {
       operationName: spanNameMapper(spanName, kind, attributes),
       context: spanContext._ddContext,
       startTime,
@@ -166,6 +156,8 @@ class Span {
       },
       links,
     }, _tracer._debug)
+
+    super(ddSpan)
 
     if (attributes) {
       this.setAttributes(attributes)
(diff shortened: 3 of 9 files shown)


packages/dd-trace/src/opentelemetry/span-helpers.js
```diff
@@ -0,0 +1,222 @@
...
+}
+
+/**
+ * @param {import('../opentracing/span')} ddSpan
+ * @param {string} key
+ * @param {unknown} value
+ * @returns {void}
+ */
+function setOtelAttribute (ddSpan, key, value) {
+  if (key === 'http.response.status_code') {
```
mabdinur, line 146: In otel, spans should be immutable after span finish ([otel sdk example](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-sdk-trace-base/src/Span.ts#L147)). I don't think we do this as of now, so this fix is beyond the scope of this PR. We can address this in a follow up.

packages/dd-trace/src/opentelemetry/span-helpers.js
```diff
@@ -0,0 +1,222 @@
...
+}
+
+/**
+ * @param {import('../opentracing/span')} ddSpan
+ * @param {Record<string, unknown>} attributes
+ * @returns {void}
+ */
+function setOtelAttributes (ddSpan, attributes) {
+  if ('http.response.status_code' in attributes) {
+    attributes['http.status_code'] = String(attributes['http.response.status_code'])
```
mabdinur: Remapping `http.status_code` to `http.response.status_code` here feels a bit odd. Ideally this remapping should be handled by the backend. `http.status_code` is not special, we have tons of Datadog tags that do not align with otel semantics and should be remapped.

packages/dd-trace/src/opentelemetry/span-helpers.js
```diff
@@ -0,0 +1,222 @@
...
+  })
+
+  /** @type {Record<string, unknown>} */
+  const attributes = {}
+  if (exception.message) attributes['exception.message'] = exception.message
+  if (exception.type) attributes['exception.type'] = exception.type
+  if (exception.escaped) attributes['exception.escaped'] = exception.escaped
+  if (exception.stack) attributes['exception.stacktrace'] = exception.stack
+
+  eventTarget.addEvent(exception.name ?? 'Error', attributes, timeInput)
```
mabdinur: According to the otel name the name of this event should be "exception": https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-spans/#exception-event. 

This is another existing issue and might be beyond the scope of this PR.

```suggestion
  eventTarget.addEvent(;exception', attributes, timeInput)
```

packages/dd-trace/src/opentelemetry/span-helpers.js
```diff
@@ -0,0 +1,222 @@
...
+ * @param {import('../opentracing/span')} ddSpan
+ * @param {Record<string, unknown>} attributes
+ * @returns {void}
+ */
+function setOtelAttributes (ddSpan, attributes) {
+  if ('http.response.status_code' in attributes) {
+    attributes['http.status_code'] = String(attributes['http.response.status_code'])
+  }
+
+  ddSpan.addTags(attributes)
```
mabdinur, line 163: Do wee want to set both 'http.status_code' and 'http.response.status_code'? Is the duplication intentional

packages/dd-trace/src/opentelemetry/active-span-proxy.js
```diff
@@ -0,0 +1,89 @@
...
+    this._ddSpan = ddSpan
+    this._otelSpanContext = otelSpanContext
+  }
+
+  spanContext () {
+    return this._otelSpanContext
+  }
+
+  get ended () {
+    return this._ddSpan?._duration !== undefined
```
mabdinur: is catching undefined sufficient? Should we also look for null or 0?

packages/dd-trace/src/opentelemetry/active-span-proxy.js
```diff
@@ -0,0 +1,89 @@
...
+    this._ddSpan.addEvent(name, event.attributes, event.startTime)
+    return this
+  }
+
+  addLink (link, attrs) {
+    addOtelLink(this._ddSpan, link, attrs)
+    return this
+  }
+
+  addLinks (links) {
```
mabdinur: Can we add a test here

packages/dd-trace/src/opentelemetry/span-helpers.js
zacharycmontoya: +1, we should fix this but totally OK to do this fix in a separate PR

packages/dd-trace/src/opentelemetry/span-helpers.js
zacharycmontoya: We may need to do this since `http.status_code` is a special tag: It is used as part of APM trace metrics / client-side stats. We may need to keep this

packages/dd-trace/src/opentelemetry/span-helpers.js
zacharycmontoya: +1, if span is finished then we should skip the setting of tags

packages/dd-trace/src/opentelemetry/active-span-proxy.js
BridgeAR: Done

packages/dd-trace/src/opentelemetry/span-helpers.js
BridgeAR: This is indeed a DD tag that we should keep. I added a comment about that.

packages/dd-trace/src/opentelemetry/span-helpers.js
BridgeAR: Resolving as out of scope

packages/dd-trace/src/opentelemetry/span.js
```diff
@@ -244,16 +217,10 @@ class Span {
...
-    if (!this.ended && !this._hasStatus && code) {
-      this._hasStatus = true
-      if (code === 2) {
-        this._ddSpan.addTags({
-          [ERROR_MESSAGE]: message,
-          [IGNORE_OTEL_ERROR]: false,
-        })
-      }
-    }
+  setStatus (status) {
```
zacharycmontoya: The logic is slightly incorrect. I know this is mostly a refactor so up to you whether to include in this PR, but see the official SDK behavior [here](https://github.com/open-telemetry/opentelemetry-js/blob/a28f12fef5fbdfb3b77a1334175b752883ef708a/packages/opentelemetry-sdk-trace-base/src/Span.ts#L318) and API doc [here](https://opentelemetry.io/docs/specs/otel/trace/api/#set-status). The no-op fast path should be replaced with the following logic:
- No op if `this.ended` (unchanged)
- No op if `status` parameter is `UNSET`
- No op if the status has been previously set to `OK` (if previously set to `ERROR` this should still run since `OK` would override it)

packages/dd-trace/src/opentelemetry/active-span-proxy.js
```diff
@@ -0,0 +1,91 @@
...
+
+  setStatus (status) {
+    if (this.ended || this.#hasStatus || !status?.code) return this
+    this.#hasStatus = true
+    applyOtelStatus(this._ddSpan, status)
+    return this
+  }
+
+  updateName (name) {
+    this._ddSpan.setOperationName(name)
```
zacharycmontoya: This is the bug I spotted in https://github.com/DataDog/evalya/pull/524

Is your preference to fix that in this PR or separately?

```suggestion
    this._ddSpan.setTag('resource.name', name)
```

packages/dd-trace/src/opentelemetry/active-span-proxy.js
BridgeAR: It seems sufficient, it would either be undefined or a number :)

packages/dd-trace/test/opentelemetry/context_manager.spec.js
```diff
@@ -350,5 +350,69 @@ describe('OTel Context Manager', () => {
...
+        const active = trace.getActiveSpan()
+        active.updateName('renamed')
+
+        assert.deepStrictEqual({
+          operationName: ddSpan.context()._name,
+          resource: ddSpan.context()._tags['resource.name'],
+        }, {
+          operationName: 'dd-active',
+          resource: 'renamed',
+        })
```
zacharycmontoya:
```suggestion
        assert.strictEqual(ddSpan.context()._name, 'dd-active')
        assert.strictEqual(ddSpan.context()._tags['resource.name'], 'renamed')
```

packages/dd-trace/test/opentelemetry/span.spec.js
```diff
@@ -283,11 +283,27 @@ describe('OTel Span', () => {
...
-    assert.strictEqual(span.name, 'new name')
+    const ddContext = span._ddSpan.context()
+    assert.deepStrictEqual({
+      operationName: ddContext._name,
+      resource: ddContext._tags[RESOURCE_NAME],
+    }, {
+      operationName: 'name',
+      resource: 'new name',
+    })
+  })
```
zacharycmontoya:
```suggestion
    assert.strictEqual(ddContext._name, 'name')
    assert.strictEqual(ddContext._tags[RESOURCE_NAME], 'new name')
```

zacharycmontoya (dismissed):
LGTM, thanks for the updates!

mabdinur (approved):

PR #8034, opened 2026-04-20
set network and http client ip on ai guard

IlyasShabi:
### What does this PR do?

Set `http.client_ip` and `network.client.ip` on the local root span when an AI Guard span is created (resource_name:ai_guard), even when `DD_TRACE_CLIENT_IP_ENABLED=false` and AppSec is disabled.

### Motivation

[APPSEC-62220](https://datadoghq.atlassian.net/browse/APPSEC-62220)

### Additional Notes

[APPSEC-62220]: https://datadoghq.atlassian.net/browse/APPSEC-62220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dd-trace/src/aiguard/index.js
+++ b/packages/dd-trace/src/aiguard/index.js
@@ -8,7 +8,9 @@ let isEnabled = false
 let aiguard
 let block
 
-function onIncomingHttpRequestStart () {}
+function onIncomingHttpRequestStart () {
+  // No-op: subscribing ensures the HTTP plugin spreads req onto the store
+}
 
 function enable (tracer, config) {
   if (isEnabled) return
--- a/packages/dd-trace/src/aiguard/index.js
+++ b/packages/dd-trace/src/aiguard/index.js
@@ -1,14 +1,14 @@
 'use strict'
 
-const { channel } = require('dc-polyfill')
 const log = require('../log')
+const { incomingHttpRequestStart, aiguardChannel } = require('./channels')
 const AIGuard = require('./sdk')
-
-const aiguardChannel = channel('dd-trace:ai:aiguard')
 
 let isEnabled = false
 let aiguard
 let block
+
+function onIncomingHttpRequestStart () {}
 
 function enable (tracer, config) {
   if (isEnabled) return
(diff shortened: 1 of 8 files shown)


smola (dismissed):

packages/dd-trace/src/aiguard/index.js
```diff
@@ -1,22 +1,23 @@
...
+const { incomingHttpRequestStart, aiguardChannel } = require('./channels')
 const AIGuard = require('./sdk')
 
-const aiguardChannel = channel('dd-trace:ai:aiguard')
-
 let isEnabled = false
 let aiguard
 let block
 
+function onIncomingHttpRequestStart () {}
```
CarlesDD: This is worth adding a comment. A noop subscription handler could easily be removed during a refactor.

packages/dd-trace/src/aiguard/sdk.js
```diff
@@ -23,6 +26,7 @@ const {
 const appsecMetrics = telemetryMetrics.manager.namespace('appsec')
 
 const ALLOW = 'ALLOW'
+const NETWORK_CLIENT_IP = 'network.client.ip'
```
CarlesDD: Nit: `network.client.ip` is also used in the appsec reporter. It would be better to extract it into ext/tags as `HTTP_CLIENT_IP`

packages/dd-trace/src/aiguard/index.js
CarlesDD: And we should update the comment in http plugin
https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-plugin-http/src/server.js#L48L50

CarlesDD: Also, this empty subscriber pulls http plugin into a code path that has no trivial overhead per-req ([`store = { ...store, req, res } `](https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-plugin-http/src/server.js#L52)) even when no aiguard evaluation happens for the req.

With appsec disabled and aiguard enabled, it will introduce overhead in hot path.

WDYT?

IlyasShabi: IMO this is the same cost as we have AppSec or IAST enabled too. The most important thing as you mentioned is to clarify this in comments

CarlesDD (approved):

smola (approved):

PR #8039, opened 2026-04-20
move instrumentation to their own workflow

rochdev:
### What does this PR do?

Move instrumentation to their own workflow.

### Motivation

Instrumentations are not tied to platform.


pabloerhard (commented):
LGTM, lets just add light-my-request into the new instrumentation.yml

pabloerhard (approved):

PR #8042, opened 2026-04-21
feat(profiling): observe async zlib and crypto events in timeline profiler

szegedi:
### What does this PR do?

Extends the timeline events profiler to observe asynchronous compression (`zlib`) and encryption (`crypto`) operations offloaded to the libuv worker thread pool, in addition to the existing DNS, TCP connect, and filesystem events. Along the way, factors the DC-channel protocol shared across these callback-style libuv instrumentations into a single helper and applies it to `dns.js` as well.

Concretely:

- **Shared helper** `createCallbackInstrumentor(prefix, { captureResult })` in `packages/datadog-instrumentations/src/helpers/callback-instrumentor.js`. It owns the end-to-end protocol: bail when no subscribers or no callback, `runStores` on `:start`, wrap the callback to publish `:error` then `:finish`, publish `:error` on synchronous throw. The optional `captureResult` flag writes the callback's first non-error argument back onto the context before `:finish`, for plugins that tag spans from the call's return value.
- **`dns.js`** migrated to use the helper. Behavior is preserved bit-for-bit: `ctx.args` is built via `buildContext` (including the `rrtype` trailing append for resolve shorthands), and `captureResult` restores the `ctx.result` assignment consumed by the DNS lookup plugin. The migration commit doubles as the helper's proof of equivalence on code that's already production-tested.
- **New instrumentation**:
  -  `packages/datadog-instrumentations/src/zlib.js` publishing `apm:zlib:operation:{start,finish,error}` for the callback-based async zlib APIs (`deflate{,Raw}`, `inflate{,Raw}`, `gzip`, `gunzip`, `unzip`, `brotliCompress`, `brotliDecompress`).
  - `packages/datadog-instrumentations/src/crypto.js` extended with `apm:crypto:operation:{start,finish,error}` for `pbkdf2`, `scrypt`, `random{Bytes,Fill,Int}`, `generateKey{,Pair}`, `hkdf`, `sign`, `verify`, `generatePrime`, `checkPrime`. The existing AppSec-oriented `datadog:crypto:hashing:start` / `datadog:crypto:cipher:start` channels are preserved unchanged.
- **Two new `EventPlugin` / decorator pairs** (`event_plugins/zlib.js`, `event_plugins/crypto.js`) wired into `DatadogInstrumentationEventSource` so the profiler emits timeline samples with event type `zlib` / `crypto` and per-op labels.

The commits follow the above structure for easier review.

### Motivation

The events profiler lets us correlate libuv thread-pool contention with specific operations and spans. Today that correlation is only possible for DNS, TCP connect, and filesystem work. Compression and encryption are common sources of pool saturation (HTTP response encoding, PBKDF2 / scrypt password hashing, key generation), but we didn't support observing them before. Part of the reason was that the plugins for the already existing three integrations emit the requisite events in their channels, so it was easy to implement them. In contrast, the zlib and crypto plugins had no instrumentation to emit the events. This is now remedied. With this, we now cover all possible built-in Node.js workloads on the libuv thread and have the best possible visibility (without writing a native profiler for them.)

### Additional Notes

- **Scope**: only the one-shot callback-based async APIs. Streaming (`createGzip`, `crypto.createCipheriv`, etc.) and sync variants are intentionally out of scope for this initial pass — consistent with how `fs` / `net` plugins treat streams today. Adding them is a separate follow-up and we can decide per-target.
- **Observer fallback**: `NodeApiEventSource` has no zlib / crypto `perf_hooks` entry type to piggyback on, so when code hotspots are disabled these events are not captured. This matches the asymmetry the other event sources already have for their sub-events and can be filled in later if needed.
- **Tests**: unit tests in `packages/dd-trace/test/profiling/profilers/events.spec.js` drive the new channels directly and assert on sample labels. Integration fixtures `integration-tests/profiler/zlibtest.js` and `cryptotest.js` plus matching cases in `profiler.spec.js` exercise the full stack end-to-end through a forked tracer process.

### Potential future work

As mentioned in additional notes, we aren't instrumenting streaming variants.

Also, some 3rd party libraries also use the libuv worker pool; it might be a future expansion of this work to add support for them, especially for those for which we already have plugins. Some known examples of 3rd party libraries using `uv_queue_work` are:
- "bcrypt" and "argon2" (native) for password hashing; can saturate pool under auth load. bcryptjs (pure-JS) doesn't.
- "sharp" libvips image processing; heavy pool user in media pipelines.
- "better-sqlite3", "sqlite3", "pg-native", "oracledb", some mysql native clients run  queries on the pool
- "canvas", "puppeteer" (indirectly via native helpers).  

Jira: [PROF-14316]

[PROF-14316]: https://datadoghq.atlassian.net/browse/PROF-14316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/datadog-instrumentations/src/crypto.js
```diff
@@ -5,16 +5,43 @@ const {
...
+  generateKeyPair: ['type'],
+  generatePrime: ['size'],
+  hkdf: ['digest', null, null, null, 'keylen'],
+  pbkdf2: [null, null, 'iterations', 'keylen', 'digest'],
+  randomBytes: ['size'],
+  randomFill: [null, 'offset', 'size'],
+  randomInt: [],
+  scrypt: [null, null, 'keylen'],
+  sign: ['algorithm'],
+  verify: ['algorithm'],
```
BridgeAR, line 32: What about using maps instead so that the key is the arguments index and we do not need to iterate over `null` entries?

packages/datadog-instrumentations/src/crypto.js
```diff
@@ -30,3 +57,20 @@ function wrapCryptoMethod (channel) {
   }
   return wrapMethod
 }
+
+function buildAsyncContext (operation, paramNames) {
```
BridgeAR, line 61: This will have an overhead compared to writing the code manually. I think the performance wins over the simplification, no?

szegedi: I hardly think an additional level of a function call is a significant enough overhead worth worrying about. We're notoriously bad at intuiting the performance characteristics of code, that's exactly what we use profilers for. If this ever shows up as a significant performance detractor for a customer, I'd consider changing it. Otherwise it's a speculative optimization and I'd rather err towards the side of maintainability in the code.

packages/datadog-instrumentations/src/crypto.js
szegedi: I'd then still rather use e.g. `['digest',,,,'keylen']` and then an iteration would only include existing indices.

packages/datadog-instrumentations/src/zlib.js
```diff
@@ -0,0 +1,27 @@
...
+const asyncMethods = [
+  'brotliCompress',
+  'brotliDecompress',
+  'deflate',
+  'deflateRaw',
+  'gunzip',
+  'gzip',
+  'inflate',
+  'inflateRaw',
+  'unzip',
```
IlyasShabi, line 16: what about `zstdCompress` and `zstdDecompress`?

packages/datadog-instrumentations/src/crypto.js
IlyasShabi, line 1: Can we have a unit test file for crypto

packages/datadog-instrumentations/src/zlib.js
IlyasShabi, line 1: Can we have a unit test file for zlib

packages/datadog-instrumentations/src/zlib.js
szegedi: You're right, I can't believe I missed those as I specifically was working on switching profiling to using zstandard compression in the past! Added them now, thanks for pointing it out!

packages/datadog-instrumentations/src/crypto.js
szegedi: Yup. For completeness sake, I added tests for existing appsec events too, I think you're qualified to review those as well :-)

packages/datadog-instrumentations/src/zlib.js
szegedi: Yup, added it.

IlyasShabi (approved):
LGTM

PR #8044, opened 2026-04-21
feat(llmobs): add support for OpenAI Agents (trace-processor)

crysmags:
## What does this PR do?

Adds APM tracing and LLMObs instrumentation for the OpenAI Agents SDK (`@openai/agents`).

The integration is **trace-processor-driven** rather than orchestrion-wrapper-driven. agents-core exposes its own `TracingProcessor` interface (`addTraceProcessor(...)`) — we plug into that instead of patching individual SDK methods. Mirrors dd-trace-py's `openai_agents` integration architecture.

**What's instrumented:**

- `addHook` on `@openai/agents-core` registers a `DDOpenAIAgentsProcessor` via `addTraceProcessor(...)` once at module load. The processor maps each agents-core `Span` / `Trace` lifecycle event into a dd-trace span (APM + LLMObs annotated). Span hierarchy is resolved through the agents-core `parentId` chain.
- `addHook` on `@openai/agents-openai` wraps `OpenAIResponsesModel.prototype.getResponse` / `getStreamedResponse` to capture the client's `baseURL` and resolve `model_provider` for response spans (`openai` / `azure_openai` / `deepseek` / `unknown`). Single provider per process is the assumed deployment shape; concurrent runs against different providers see last-write-wins on this field.

**Span types handled:**

| oai-span type | dd-trace span name | LLMObs kind |
|---|---|---|
| Trace | `Agent workflow` (or `oaiTrace.name`) | `workflow` |
| `agent` | agent's `spanData.name` | `agent` |
| `response` | `${parent_agent_name} (LLM)` if child of top-level agent, else `openai_agents.response` | `llm` |
| `function` | tool's `spanData.name` | `tool` |
| `handoff` | `transfer_to_${to_agent}` | `tool` |
| `guardrail` | guardrail's `spanData.name` | `task` |
| `custom` | `spanData.name` | `task` |

## Why this architecture (vs #7808's orchestrion approach)

#7808 wrapped `run`, `getResponse`, `getStreamedResponse`, `invokeFunctionTool`, `onInvokeHandoff`, `runInputGuardrails`, `runOutputGuardrails` individually. That gave the integration full visibility into call arguments, but couldn't express multi-agent handoff parent/child relationships correctly — a reviewer flagged this on the original PR.

The trace-processor switch fixes hierarchy (we get `parentId` chains directly from agents-core) but trades off:

- **No access to user's `modelSettings`.** agents-openai stores only `request.input` on `spanData._input`; `temperature`/`top_p`/etc. are echoed back on the response. We read them from the response (matches Python's `OaiSpanAdapter.llmobs_metadata`).
- **No `systemInstructions` on error paths.** agents-openai gates `spanData._input/_response` on `request.tracing` and only sets them after a successful network call. On error, neither is on the span. The LLMObs error event therefore carries placeholder `input.messages` / `output.messages` (single source of truth: the `extractInputMessages` / `extractOutputMessages` helpers).
- **No stream marker.** Both streamed and non-streamed produce `type: 'response'` spans with no distinguishing field. Python doesn't tag `stream: true` either.
- **Direct-helper calls don't emit spans.** `invokeFunctionTool`, `onInvokeHandoff`, `runToolInputGuardrails`, `runToolOutputGuardrails` only create Spans when called by the SDK runner inside a `Runner.run()` flow. Direct calls to these helpers are no-ops on the tracing side. Same constraint applies to dd-trace-py.
- **`Trace.end()` not called when `withTrace`'s callback throws.** agents-core's lifecycle wrapper skips `Trace.end()` on error — we mitigate by finalizing the workflow span when any top-level oai-span ends (`_completeWorkflowSpan`).

The integration matches dd-trace-py's `openai_agents` line-by-line on these decisions; the test suite was rewritten to mirror Python's coverage and assertion shape.

## Critical-review cleanup (3ede0b85b)

A deep pass for dead, redundant, and inefficient code:

- Drop dead `_responseModelName` (always undefined at `onSpanStart` time).
- Drop the `_setResponseAttributes` early return that contradicted the "consistent error-path shape" comment; let the `extract*` helpers be the single source of truth for placeholder I/O messages. Closes Codex P2.
- Merge `endTrace` + `_finalizeTraceIfOrphaned` into one `_completeWorkflowSpan`.
- Cache a single `RESOLVED` Promise in the processor and drop `Promise.resolve()` allocations on every span lifecycle event.
- Make the processor read `getIntegration()` lazily so plugin re-init doesn't strand the registered processor against a stale reference.
- `clearState()` now finishes in-flight dd-trace spans on SDK shutdown.
- `extractMetrics` / `extractMetadata` return `undefined` when empty — callers skip `Object.keys()` allocations.
- Cache the first response span's `inputMessages` on `_traceInfo` so the workflow span's text I/O isn't recomputed at `Trace.end()`.
- Stash the top-level agent display name on `_traceInfo` and add a public `tagger.setName()`, closing the `_setTag(NAME)` / `context._name` encapsulation hole.
- Wire `getOpenAIModelProvider` through openai-agents (Azure / DeepSeek coverage) via the new `@openai/agents-openai` hook.
- `replaceAll` instead of `split.join.toLowerCase` for handoff names.

## Test coverage

- **APM plugin** (`packages/datadog-plugin-openai-agents/test/index.spec.js`): 7 tests — `Runner.run()` single-agent workflow (workflow/agent/response spans + error path), multi-agent handoff hierarchy, `getResponse` / `getStreamedResponse` direct calls (happy + error).
- **LLMObs spec** (`packages/dd-trace/test/llmobs/plugins/openai-agents/index.spec.js`): 6 tests covering `getResponse` / `getStreamedResponse` happy paths, `Runner.run()` single-agent workflow span, and tool-error-via-`Runner.run()`. The tool-error test uses a stateful mock client (1st call returns `function_call` to `error_tool`, 2nd call returns assistant text) so no recorded cassette is needed.

## Test plan

- [ ] `PLUGINS=openai-agents npm run test:plugins` — 7 APM tests pass.
- [ ] `./node_modules/.bin/mocha packages/dd-trace/test/llmobs/plugins/openai-agents/index.spec.js` — 6 LLMObs tests pass.
- [ ] `npm run lint` — clean.


packages/datadog-plugin-openai-agents/src/turn.js
```diff
@@ -0,0 +1,51 @@
...
+    const integration = getIntegration()
+    if (!integration?.enabled) return
+
+    const agent = ctx.arguments?.[0]
+    if (!agent) return
+
+    // The dd-trace span to tag is the most recently started `agent`-type span.
+    // agents-core's Span.parent chain guarantees it is still open when
+    // _runSingleTurn returns (withAgentSpan brackets the turn).
+    const ddSpan = integration.currentAgentSpan?.()
```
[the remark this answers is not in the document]
crysmags: Stale — `turn.js` was removed in c8a83bd5a (the orchestrion wrap on `AgentRunner._runSingleTurn` was eliminated as part of collapsing the plugin layout). Agent manifests now flow through the TracingProcessor itself: `_setAgentAttributes` reads `spanData.handoffs` / `spanData.tools` / `spanData.output_type` directly from the oai-span, so there's no longer a process-wide stack to confuse across concurrent runs. Closing as fixed by removal.

packages/datadog-plugin-openai-agents/src/integration.js
```diff
@@ -0,0 +1,459 @@
...
+      this._tagger.tagMetadata(activeSpan, { _dd: { agent_manifest: manifest } })
+    }
+  }
+
+  // ── Per-type attribute setters (match Python) ───────────────────────────────
+
+  _setResponseAttributes (ddSpan, oaiSpan) {
+    const response = oaiSpan.spanData?._response
+    const input = oaiSpan.spanData?._input
+    if (!response && input == null) return
```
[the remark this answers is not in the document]
crysmags: Addressed in 3ede0b85b. The early return `if (!response && input == null) return` is gone, and `_setResponseAttributes` now passes the raw `input` / `response` through to `extractInputMessages` and `extractOutputMessages` unconditionally. Those helpers are the single source of truth for placeholder messages — `extractInputMessages` returns `[{ role: 'user', content: '' }]` when there's nothing to capture, `extractOutputMessages` returns `[{ content: '', role: '' }]` — so error-path LLMObs events now carry the same `input.messages` / `output.messages` shape as the happy path.

packages/datadog-instrumentations/src/openai-agents.js
```diff
@@ -0,0 +1,57 @@
...
+)
+const shimmer = require('../../datadog-shimmer')
+const { addHook } = require('./helpers/instrument')
+
+/**
+ * @param {object} mod `@openai/agents-core` module exports.
+ * @returns {object} the same module exports, possibly with our processor registered.
+ */
+function registerProcessor (mod) {
+  if (mod?._datadogPatched) return mod
```
sabrenner: lets make this private or use a weakmap to not use an underscore

packages/datadog-instrumentations/src/openai-agents.js
```diff
@@ -0,0 +1,57 @@
...
+  if (typeof mod?.addTraceProcessor !== 'function') return mod
+
+  // Bind the lazy accessor so the processor follows plugin re-instantiation
+  // even though agents-core keeps the processor registration for the life
+  // of the module.
+  mod._datadogPatched = true
+  mod.addTraceProcessor(new DDOpenAIAgentsProcessor(getIntegration))
```
sabrenner:
```js
const modCh = channel('....')
modCh.publish({ mod })
```

packages/datadog-instrumentations/src/openai-agents.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Capture the OpenAI client's baseURL on each `getResponse` / `getStreamedResponse`
+ * call so the integration can tag `model_provider` (openai / azure_openai /
+ * deepseek / unknown). Last-write-wins; single-provider-per-process is the
+ * assumed deployment shape — matches dd-trace-py's openai-agents integration.
+ *
+ * @param {Function} original
+ * @returns {Function}
+ */
+function wrapResponseMethod (original) {
```
sabrenner, line 57: maybe this could be orchestrion and we do this in a subscriber

packages/datadog-plugin-openai-agents/src/index.js
```diff
@@ -0,0 +1,41 @@
...
+ * The integration's `enabled` flag follows this plugin's configure() lifecycle.
+ * The processor stays registered inside agents-core for the life of the
+ * process — re-enabling the plugin later flips the flag back on.
+ */
+class OpenaiAgentsPlugin extends Plugin {
+  static id = 'openai-agents'
+
+  constructor (tracer, tracerConfig) {
+    super(tracer, tracerConfig)
+    const tagger = new LLMObsTagger(tracerConfig, true)
```
sabrenner: move this & all llmobs logic as conditional inside of the OpenAIAgentsIntegration

sabrenner: `tracerConfig.llmobs.enabled`

packages/datadog-plugin-openai-agents/src/index.js
```diff
@@ -0,0 +1,41 @@
...
+    this._integration = new OpenAIAgentsIntegration({
```
sabrenner:
```suggestion
    this.#integration = new OpenAIAgentsIntegration({
```

packages/datadog-plugin-openai-agents/src/integration.js
sabrenner, line 1: any llmobs logic we should try and gate behind llmobs being enabled

packages/dd-trace/src/llmobs/plugins/utils.js
sabrenner, line 1: we can keep this under the openai plugin and import it into the openai-agents one

packages/datadog-plugin-openai-agents/src/integration.js
sabrenner, line 1: two things:
1. can swap out `_` for `#` where possible
2. any "helper" methods that don't need "this" can be in a `util.js` file

index.d.ts
sabrenner, line 1: might need to update `index.d.v5.ts` as well

packages/datadog-plugin-openai-agents/test/test-setup.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  /**
+   * Compose agents-core's span helpers directly to produce the hierarchy:
+   *   trace("handoff-test") → agent(agent_a) → handoff(agent_a→agent_b) → agent(agent_b)
+   * Bypasses the model call to avoid needing cassette support for handoff
+   * tool-call responses.
+   */
+  async multiAgentHandoff () {
+    return this.module.withTrace('handoff-test', async () => {
```
sabrenner: let's see if getting rid of `withTrace` things breaks anything

sabrenner (commented):
notes from sync review

packages/datadog-instrumentations/src/openai-agents.js
crysmags: Done in 950bd8eb1 — module-scoped `WeakSet` keyed by the loaded `mod`. Same dedupe semantics; no extra lifetime since `require.cache` keeps the mod alive anyway.

packages/datadog-instrumentations/src/openai-agents.js
crysmags: Done in 950bd8eb1 — both `addTraceProcessor` registration and the `OpenAIResponsesModel` client-baseURL capture now go through channels:
- `apm:openai-agents:agents-core:loaded` publishes the module on load; the plugin subscribes in its constructor and registers the trace processor.
- `apm:openai-agents:response:client` publishes the OpenAI client's `baseURL` on each `getResponse`/`getStreamedResponse` call; the plugin subscribes and pushes it to the integration.

The instrumentation file no longer imports anything from the plugin module. Timing: `register.js` publishes `loadChannel({ name })` before invoking the addHook callback, so the plugin (constructed by the plugin manager's `loadChannel` subscriber) is live when our publish fires. Edge case (another spec required `@openai/agents-core` before our plugin was constructed) is covered by a bounded `Set` of already-loaded mods that the plugin drains on construction — set holds at most one entry per process.

packages/datadog-instrumentations/src/openai-agents.js
crysmags: Tried orchestrion first, but the YAML-level operators it ships with are `tracePromise` / `traceSync` / `traceCallback`, and `OpenAIResponsesModel.getStreamedResponse` is an async generator (`async *getStreamedResponse`). `tracePromise` would force-wrap it into an async function and break callers expecting an `AsyncGenerator`; orchestrion's `kind: 'AsyncIterator'` matcher exists internally (see `packages/datadog-instrumentations/test/helpers/rewriter/index.spec.js`) but isn't surfaced through the YAML config.

Took the half of your suggestion that's not blocked on orchestrion: the wrap now publishes the client baseURL on `apm:openai-agents:response:client` and the plugin subscribes — same decoupling, no cross-package import. shimmer stays as the wrap mechanism only because of the async-generator limitation. Happy to revisit when orchestrion grows an async-iterator operator at the config layer. 950bd8eb1.

packages/datadog-plugin-openai-agents/src/index.js
crysmags: Done in 950bd8eb1 — `LLMObsTagger` construction moved into `OpenAIAgentsIntegration`'s constructor, gated on `config?.llmobs?.enabled`. When LLMObs is off, `#tagger` stays `undefined`, every tagger call site is gated, and the work that feeds those calls (`extractInputMessages` / `extractOutputMessages` / `extractMetrics` / `extractMetadata`, trace-info reconstruction, `#setResponseAttributes` / `#setAgentAttributes` / `#setTraceAttributes`) is skipped. APM span emission stays unconditional.

packages/datadog-plugin-openai-agents/src/index.js
crysmags: Done in 950bd8eb1 — `#integration` private field. The `currentIntegration` module-level singleton and the `getIntegration` export are gone too; the channel-subscription pattern from #3236453244 replaces them.

packages/datadog-plugin-openai-agents/src/integration.js
crysmags: Done in 950bd8eb1 — see reply on #3236465651. Constructor builds `#tagger = config?.llmobs?.enabled ? new LLMObsTagger(config, true) : undefined`; all `this.#tagger.xxx(...)` call sites are gated by an `if (this.#tagger)` (or moved inside one), so `extractInputMessages` / `extractOutputMessages` / `extractMetrics` / `extractMetadata` and the trace-info reconstruction stay off the hot path when LLMObs is disabled. APM tracing flows regardless.

packages/dd-trace/src/llmobs/plugins/utils.js
crysmags: Moved in 950bd8eb1 — `getOpenAIModelProvider` now lives in `packages/dd-trace/src/llmobs/plugins/openai/utils.js` and is re-imported by both `openai/index.js` and the openai-agents integration. Restored the `UNKNOWN_MODEL_PROVIDER` named constant import along the way (the earlier extraction had hardcoded the literal). `packages/dd-trace/src/llmobs/plugins/utils.js` is deleted.

packages/datadog-plugin-openai-agents/src/integration.js
crysmags: Done in 950bd8eb1.

1. All `_underscore` fields and internal methods inside `OpenAIAgentsIntegration` converted to `#private` (`#tracer`, `#enabled`, `#modelProvider`, `#tagger`, `#oaiToDdSpan`, `#traceInfo`, `#completeWorkflowSpan`, `#setResponseAttributes`, `#llmSpanParentAgentName`, `#setAgentAttributes`, `#setTraceAttributes`, `#updateTraceInfoInput`, `#updateTraceInfoOutput`, `#resolveParent`). Public surface (`enabled`, `setEnabled`, `setClientBaseURL`, `clearState`, `startTrace`, `endTrace`, `startSpan`, `endSpan`) stays public. Same for `OpenaiAgentsPlugin.#integration`.
2. The two stateless helpers — `deriveSpanName` and `applyError` — and the `AGENTS_ERROR_TYPE` constant moved into a new `packages/datadog-plugin-openai-agents/src/util.js`.

index.d.ts
crysmags: Done in 950bd8eb1 — added the corresponding lines to `index.d.v5.ts`:
- `"openai-agents": tracer.plugins.openai_agents;` in the plugins map.
- `interface openai_agents extends Instrumentation {}` with the matching docblock.

packages/datadog-plugin-openai-agents/test/test-setup.js
crysmags: Tried it and it does break, but only for the direct-`getResponse()` shape: agents-openai's `withResponseSpan` calls into agents-core's `getCurrentTrace()` and returns a `NoopSpan` when there's no enclosing Trace ([provider.js:79](https://github.com/openai/openai-agents-js/blob/v0.7.0/packages/agents-core/src/tracing/provider.ts) — `"No active trace. Make sure to start a trace with withTrace() first. Returning NoopSpan."`). The cassettes recover the HTTP response just fine, but no `response`-type oai-Span is ever emitted by the SDK, so our processor never sees a lifecycle event and the response-span assertions fail.

Path taken in 950bd8eb1: dropped the `getResponse` / `getStreamedResponse` test-setup helpers and the corresponding describe blocks (both APM and LLMObs specs). The `Runner.run` flows already cover response-span emission end-to-end, and I added a detailed LLM-event shape assertion under the `run` describe in the LLMObs spec so we keep input/output-message, metrics, and metadata coverage. `multiAgentHandoff` still uses `withTrace` because it intentionally bypasses `Runner.run` to test the dd-trace processor's parent-id resolution without needing recorded handoff cassettes — left it as-is with a comment.

packages/datadog-instrumentations/src/openai-agents.js
```diff
@@ -0,0 +1,70 @@
...
+// orchestrion grows an async-iterator operator at the YAML layer, shimmer
+// is the smaller surface; we keep the subscriber/channel half of the
+// suggestion so the wrap stays decoupled from the plugin module.
+const responseClientCh = channel('apm:openai-agents:response:client')
+
+addHook({ name: '@openai/agents-core', versions: ['>=0.7.0'] }, (mod) => {
+  if (patchedMods.has(mod)) return mod
+  if (typeof mod?.addTraceProcessor !== 'function') return mod
+  patchedMods.add(mod)
+  loadedAgentsCoreMods.add(mod)
```
sabrenner: i think these could be on the subscriber side, and we can just do
```js
agentsCoreLoadedCh.publish({ mod })
return mod
```
here

packages/datadog-instrumentations/src/openai-agents.js
sabrenner: i think `traceSync` might work? it would basically just not await the promise and pass it through. but we also have the stream iterator handlers as well we could try (like we use for langgraph). since we only care about the input and not the return type

packages/datadog-plugin-openai-agents/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+    const registerProcessor = (mod) => {
+      if (typeof mod?.addTraceProcessor !== 'function') return
+      mod.addTraceProcessor(new DDOpenAIAgentsProcessor(() => this.#integration))
+    }
+    // Drain any agents-core mods that loaded before this plugin was
+    // constructed (e.g. another plugin's tests triggered the require first),
+    // then keep listening for future loads. Set is bounded to one entry per
+    // process — agents-core is a singleton dep.
+    for (const mod of loadedAgentsCoreMods) registerProcessor(mod)
+    agentsCoreLoadedCh.subscribe(({ mod }) => registerProcessor(mod))
```
sabrenner: i think we can/should use `this.addSub`

packages/datadog-plugin-openai-agents/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+
+    responseClientCh.subscribe(({ baseURL }) => {
```
sabrenner: likewise here

packages/datadog-plugin-openai-agents/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+  constructor (tracer, tracerConfig) {
+    super(tracer, tracerConfig)
+    this.#integration = new OpenAIAgentsIntegration({
+      tracer: this.tracer,
+      config: tracerConfig,
+    })
+
+    const registerProcessor = (mod) => {
+      if (typeof mod?.addTraceProcessor !== 'function') return
+      mod.addTraceProcessor(new DDOpenAIAgentsProcessor(() => this.#integration))
```
sabrenner: can/should we add safeguards to make sure we somehow don't add more than one processor (not sure it's a possible case to hit, but to be safe)?

packages/datadog-plugin-openai-agents/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+    // Drain any agents-core mods that loaded before this plugin was
+    // constructed (e.g. another plugin's tests triggered the require first),
+    // then keep listening for future loads. Set is bounded to one entry per
+    // process — agents-core is a singleton dep.
+    for (const mod of loadedAgentsCoreMods) registerProcessor(mod)
```
sabrenner: i don't think we should have lines in here just to satisfy tests. i think it makes sense that the instrumentation triggers a call to the plugin via publisher-subscriber methods we have.

packages/datadog-plugin-openai-agents/test/test-setup.js
sabrenner: so is it expected that customers will need to do this too then?

sabrenner (commented):
nice just one question on the test and some cleanup things, but afaict it shouldn't change the logic that much, so i'll run through it locally!

packages/datadog-plugin-openai-agents/src/processor.js
```diff
@@ -0,0 +1,107 @@
...
+ * (APM + LLMObs-annotated) keyed off the agents-core spanId / traceId. Parent
+ * hierarchy is resolved through the agents-core parentId chain, which gives us
+ * correct multi-agent handoff nesting that ctx-argument capture cannot provide.
+ *
+ * agents-core awaits the lifecycle methods, so each one returns a settled
+ * Promise even though the work is synchronous.
+ */
+class DDOpenAIAgentsProcessor {
+  /**
+   * @param {() => (object | undefined)} getIntegration - Lazy accessor for the
```
sabrenner:
```suggestion
   * @param {() => (import('./integration').OpenAIAgentsIntegration | undefined)} getIntegration - Lazy accessor for the
```

typing nit

packages/datadog-instrumentations/src/helpers/hooks.js
```diff
@@ -19,6 +19,8 @@ module.exports = {
   '@apollo/gateway': () => require('../apollo'),
   '@langchain/langgraph': { esmFirst: true, fn: () => require('../langgraph') },
   '@modelcontextprotocol/sdk': () => require('../modelcontextprotocol-sdk'),
+  '@openai/agents-core': () => require('../openai-agents'),
```
sabrenner:
```suggestion
  '@openai/agents': () => require('../openai-agents'),
```

packages/datadog-instrumentations/src/openai-agents.js
```diff
@@ -0,0 +1,70 @@
...
+// publish below lands on a live subscriber.
+//
+// Late-construct path (e.g. unrelated test loads `@openai/agents-core`
+// before our plugin is constructed): the addHook callback fires first and
+// the publish has no subscriber. To survive that we also push the mod into
+// `loadedAgentsCoreMods` and the plugin drains the set on subscribe. Set
+// holds at most one entry per process (agents-core is a singleton dep) and
+// the mod is kept alive by `require.cache` anyway — no extra lifetime.
+const agentsCoreLoadedCh = channel('apm:openai-agents:agents-core:loaded')
+const loadedAgentsCoreMods = new Set()
```
sabrenner:
```suggestion
```

packages/datadog-instrumentations/src/openai-agents.js
```diff
@@ -0,0 +1,70 @@
...
+// orchestrion + a TracingPlugin subscriber, but orchestrion's YAML config
+// only exposes `tracePromise` / `traceSync` operators, and
+// `OpenAIResponsesModel.getStreamedResponse` is an async generator —
+// neither operator wraps async generators correctly today. Until
+// orchestrion grows an async-iterator operator at the YAML layer, shimmer
+// is the smaller surface; we keep the subscriber/channel half of the
+// suggestion so the wrap stays decoupled from the plugin module.
+const responseClientCh = channel('apm:openai-agents:response:client')
+
+addHook({ name: '@openai/agents-core', versions: ['>=0.7.0'] }, (mod) => {
```
sabrenner:
```suggestion
addHook({ name: '@openai/agents', versions: ['>=0.7.0'] }, (mod) => {
```

packages/datadog-instrumentations/src/openai-agents.js
sabrenner:
```suggestion
  patchedMods.add(mod)
```

packages/datadog-plugin-openai-agents/src/index.js
sabrenner: with some more comments i've left, i think we can remove this bit and just have the subscribe below

packages/dd-trace/src/plugins/index.js
```diff
@@ -33,6 +33,7 @@ const plugins = {
   get '@smithy/smithy-client' () { return require('../../../datadog-plugin-aws-sdk/src') },
   get '@vitest/runner' () { return require('../../../datadog-plugin-vitest/src') },
   get '@langchain/langgraph' () { return require('../../../datadog-plugin-langgraph/src') },
+  get '@openai/agents-core' () { return require('../../../datadog-plugin-openai-agents/src') },
```
sabrenner:
```suggestion
  get '@openai/agents' () { return require('../../../datadog-plugin-openai-agents/src') }
```

packages/datadog-instrumentations/src/openai-agents.js
```diff
@@ -0,0 +1,70 @@
...
+  if (patchedMods.has(mod)) return mod
+  if (typeof mod?.addTraceProcessor !== 'function') return mod
+  patchedMods.add(mod)
+  loadedAgentsCoreMods.add(mod)
+  agentsCoreLoadedCh.publish({ mod })
+  return mod
+})
+
+module.exports = { loadedAgentsCoreMods }
+
```
sabrenner:
```suggestion
```

sabrenner (commented):
a couple of suggestions from some local edits i made while trying this out ([see full diff](https://github.com/DataDog/dd-trace-js/compare/crysmags/openai-agents-trace-processor...sabrenner/openai-agents-modifications)). 

i had some other observations i will follow up with as well as far as like LLM Obs tagging goes, but wanted to throw these suggestions your way!

the gist of it is that we can just hook the higher-up package and not need to keep track of loaded modules/just subscribe to the one load channel and add our trace processor there.

packages/datadog-instrumentations/src/openai-agents.js
crysmags: Done in fc45a4818 — subscriber side handles everything now. The `@openai/agents` hook fires after all its dependencies have loaded, so the plugin's `addSub` subscription is already active when the publish lands. Removed `loadedAgentsCoreMods`, the `module.exports`, and the drain loop entirely.

packages/datadog-instrumentations/src/openai-agents.js
crysmags: Done in fc45a4818 — went with `traceAsyncIterator` (the langgraph pattern) for `getStreamedResponse` since it's a proper async generator. Added `instrumentations/openai-agents.js` with CJS + ESM variants and subscribed to the `:start` channel in the plugin to capture `ctx.self.client?.baseURL`. Kept shimmer only for `getResponse` since `traceAsyncIterator` mutates the resolved Promise's value (adds `.next`/`.return`/`.throw`), which is harmless but unclean for a regular async function.

packages/datadog-plugin-openai-agents/src/index.js
crysmags: Done in fc45a4818 — both channel subscriptions converted to `this.addSub`.

packages/datadog-plugin-openai-agents/src/index.js
crysmags: Done in fc45a4818 — likewise.

packages/datadog-plugin-openai-agents/src/index.js
crysmags: Done in fc45a4818 — added a `#registeredMods = new WeakSet()` class field. Before calling `mod.addTraceProcessor`, the handler checks `this.#registeredMods.has(mod)` and bails if it's already there, then marks it with `add(mod)`. Prevents double-registration if the channel fires more than once for the same module object.

packages/datadog-plugin-openai-agents/src/index.js
crysmags: Done in fc45a4818 — drain loop removed. Switching to `@openai/agents` (the user-facing package) resolves the timing issue entirely: `@openai/agents` loads after all its dependencies (`agents-core`, `agents-openai`), so our plugin is always constructed and its `addSub` is live before the hook fires. The drain set was only needed because tests were directly requiring `@openai/agents-core` before the plugin was constructed — that's no longer the case now that we key off the high-level package.

packages/datadog-plugin-openai-agents/test/test-setup.js
crysmags: The `withTrace` in `multiAgentHandoff` is intentional — that test exercises the processor's parent-id resolution without going through `Runner.run`, so we have to establish the trace context manually (same as the SDK's own unit tests do). For standard `Runner.run` usage customers don't need to call `withTrace` at all; the runner establishes the trace context internally. If a customer is calling `getResponse` directly outside of a runner, `withTrace` is the agents-core API contract for that (the SDK logs a warning when there's no enclosing trace and returns a `NoopSpan`).

packages/datadog-instrumentations/src/helpers/hooks.js
crysmags: Addressed in fc45a4818.

packages/datadog-instrumentations/src/openai-agents.js
crysmags: Addressed in fc45a4818 — `loadedAgentsCoreMods` Set removed.

packages/datadog-instrumentations/src/openai-agents.js
crysmags: Addressed in fc45a4818.

packages/datadog-instrumentations/src/openai-agents.js
crysmags: Addressed in fc45a4818 — `loadedAgentsCoreMods.add(mod)` line removed; only `patchedMods.add(mod)` + the publish remain.

packages/datadog-plugin-openai-agents/src/index.js
crysmags: Addressed in fc45a4818 — drain loop removed entirely (see fc45a4818 reply on #3284192791).

packages/dd-trace/src/plugins/index.js
crysmags: Addressed in fc45a4818.

packages/datadog-instrumentations/src/openai-agents.js
crysmags: Addressed in fc45a4818 — `module.exports` line removed; instrumentation file now exports nothing.

packages/datadog-plugin-openai-agents/src/processor.js
crysmags: Addressed in fc45a4818 — `@param` type updated to `{() => (import('./integration').OpenAIAgentsIntegration | undefined)}`.

packages/datadog-plugin-openai-agents/src/index.js
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * The integration's `enabled` flag follows this plugin's configure()
+ * lifecycle. Once registered, the processor stays inside agents-core for
+ * the life of the process — flipping `enabled` is what re-engages tracing.
+ */
+class OpenaiAgentsPlugin extends Plugin {
+  static id = 'openai-agents'
+
+  #integration
+  #registeredMods = new WeakSet()
```
sabrenner: we might even be able to just use a boolean i think, instead of a whole set object, unless i'm mistaken? we'll only ever add the processor one time i think. but lmk if i'm wrong!

sabrenner (commented):
one comment, but following up offline with something i noticed while QAing!

packages/datadog-plugin-openai-agents/src/index.js
crysmags: You're right! Simplified to `#processorRegistered = false` in 0e3227c34 — `patchedMods` in the instrumentation layer already prevents the channel from publishing the same mod twice, so the per-plugin guard only ever sees one mod and a boolean is sufficient.

packages/datadog-instrumentations/src/openai-agents.js
crysmags: @sabrenner one update after that reply: `traceAsyncIterator` did not keep the agent dd span active while each generator step executed, so the final head uses the same channel binding around `next` / `return` / `throw`. The streamed-parent regression test now pins that `openai.request` stays under the agent span.

packages/datadog-instrumentations/src/openai-agents.js
```diff
@@ -0,0 +1,109 @@
...
+// require (and without triggering n/no-missing-require on agents-core internals).
+let agentsMod
+
+// @openai/agents >=0.8.0 moved addTraceProcessor / getCurrentSpan out of the
+// top-level re-exports.  The new public surface uses getGlobalTraceProvider():
+//   provider.registerProcessor(processor)  (replaces addTraceProcessor)
+//   provider.getCurrentSpan()              (replaces getCurrentSpan)
+// Both APIs are tried so this file works across the full supported version range.
+// The plugin subscriber (index.js) handles processor registration via the channel.
+function getCurrentSpanId (mod) {
```
sabrenner, line 39: this probably doesnt need to take in an argument if it's only being called with the global `agentsMod`

packages/dd-trace/test/llmobs/cassettes/openai/openai_responses_post_1e503f45.yaml
sabrenner, line 1: these can probably be removed

sabrenner (approved):
we will follow up with any bug fixes in a subsequent pr - but everything looks safe and i had tested previously / manually as well

BridgeAR:
@crysmags @sabrenner we got quite a few comments by codex in the release branch: https://github.com/DataDog/dd-trace-js/pull/9523 about this change. Mind taking a look? :)

PR #8052, opened 2026-04-22
test(light-my-request): run suite under withVersions

BridgeAR:
The existing spec hard-required the root-installed copy of \`light-my-request\` and \`fastify\`, so it only ever exercised one version on CI and silently ignored the rest of the supported range. Every other instrumentation suite uses \`withVersions\` to sweep the version matrix maintained by \`install_plugin_modules\`.

Three coordinated changes restore that behavior:

1. The spec is wrapped in \`withVersions('light-my-request', ...)\` and the module \`require\`s now resolve through \`versions/light-my-request@<version>\` / \`versions/fastify\`, so each supported version actually runs.
2. \`packages/dd-trace/test/plugins/externals.js\` registers \`fastify\` (>=3) as an external dependency of the \`light-my-request\` suite so the matrix installer stages a working fastify alongside each \`light-my-request\` version.
3. \`.github/workflows/platform.yml\` adds the \`instrumentation-light-my-request\` job, mirroring the pattern used by the other \`instrumentation-*\` jobs, so CI actually invokes the newly-parameterized suite.

No production code changes, so the blast radius is test-only.

Depends on https://github.com/DataDog/dd-trace-js/pull/8051


.github/workflows/platform.yml
```diff
@@ -292,6 +292,16 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - uses: ./.github/actions/instrumentations/test
 
+  instrumentation-light-my-request:
```
pabloerhard, line 295: There is an open PR to move instrumentations into another workflow, I think we should land that before this PR. #8039

BridgeAR: I guess either needs to be updated, which ever lands first?

pabloerhard (approved):

PR #8059, opened 2026-04-22
test(openfeature): match printf-style log.debug args

BridgeAR:
The `FlaggingProvider` implementation was migrated to printf-style `log.debug('%s created with timeout: %dms', ...)` calls (cheaper in hot paths because the format string is only expanded when the log level is active). Two spec assertions still expected the pre-formatted interpolated strings and would fail once the production change lands.

Update the two `sinon.assert.calledWith` calls to match the new format-string + positional-args shape. Pure test catch-up.


leoromanovsky (approved):
Thank you

PR #8061, opened 2026-04-22
chore(scripts): flag unquoted `**` globs in npm scripts

BridgeAR:
POSIX sh — which is what both `npm run` and `yarn run` ultimately invoke — does not support globstar. An unquoted `**` in a package.json script therefore collapses to a single `*`, so the shell only passes one directory level through to mocha/glob and specs deeper than that are silently skipped. That is how a handful of suites were under-running without anyone noticing.

Two coordinated pieces:

1. A new `findUnquotedGlobstar` helper (quote-state-aware, so it doesn't false-positive on patterns that are already quoted) runs over every `scripts` entry during `verify-exercised-tests` and fails the check with a pointer to the offending column if any unquoted `**` is found. Doing this via string scanning instead of a glob-expansion probe is important because `globSync` itself will happily expand `**` recursively regardless of quoting, so it can't distinguish "works locally because glob is being nice" from "broken in CI because sh already collapsed it".

2. Wrap the existing offenders in double quotes so they actually pass the check: `test:aiguard`, `test:debugger`, `test:trace:guardrails`, `test:esbuild`, `test:webpack`, `test:instrumentations:misc`, `test:core`, `test:code-origin`, `test:lambda`, `test:openfeature`, and `test:profiler`.


packages/dd-trace/test/debugger/devtools_client/snapshot/collector-deadline.spec.js
```diff
@@ -61,13 +61,13 @@ describe('debugger -> devtools client -> snapshot collector deadline', function
   })
 
   it('should mark properties with timeout when deadline is exceeded', async function () {
-    // Override the hrtime stub to advance time on each call
-    // This simulates time passing during collection
+    // Drive process.hrtime.bigint() with a plain monotonic counter instead of reading
+    // `clock.now`. This works around a sinon fake-timers regression.
```
watson, line 67: it would be nice with a link to a github issue so it can later be reverted once it's fixed upstream

watson (dismissed):

packages/dd-trace/test/debugger/devtools_client/snapshot/collector-deadline.spec.js
BridgeAR:
```suggestion
    // `clock.now`. This works around a sinon fake-timers regression.
    // TODO: Revert when a fix for https://github.com/sinonjs/fake-timers/pull/549#pullrequestreview-4158432441 lands.
```

watson (dismissed):

packages/dd-trace/test/debugger/devtools_client/snapshot/collector-deadline.spec.js
```diff
@@ -107,13 +108,13 @@ describe('debugger -> devtools client -> snapshot collector deadline', function
   it('should cache deadline reached state in ctx', async function () {
     let hrtimeCallCount = 0
 
-    // Override the hrtime stub to track calls and advance time
+    // Same workaround as the previous test: avoid the fake-timers clock for `process.hrtime.bigint()`.
```
watson:
```suggestion
    // Override the hrtime stub to track calls and advance time
    // Drive process.hrtime.bigint() with a plain monotonic counter instead of reading
    // `clock.now`. This works around a sinon fake-timers regression.
watson marked this conversation as resolved.
    // TODO: Revert when a fix for https://github.com/sinonjs/fake-timers/pull/549#pullrequestreview-4158432441 lands.
```

packages/dd-trace/test/debugger/devtools_client/snapshot/collector-deadline.spec.js
```diff
@@ -61,13 +61,14 @@ describe('debugger -> devtools client -> snapshot collector deadline', function
   })
 
   it('should mark properties with timeout when deadline is exceeded', async function () {
-    // Override the hrtime stub to advance time on each call
-    // This simulates time passing during collection
+    // Drive process.hrtime.bigint() with a plain monotonic counter instead of reading
+    // `clock.now`. This works around a sinon fake-timers regression.
+    // TODO: Revert when a fix for https://github.com/sinonjs/fake-timers/pull/549#pullrequestreview-4158432441 lands.
```
watson, line 68:
```suggestion
    // Override the hrtime stub to advance time on each call
    // This simulates time passing during collection
    // Drive process.hrtime.bigint() with a plain monotonic counter instead of reading
    // `clock.now`. This works around a sinon fake-timers regression.
    // TODO: Revert when a fix for https://github.com/sinonjs/fake-timers/pull/549#pullrequestreview-4158432441 lands.
```

watson (dismissed):

watson (approved):

PR #8064, opened 2026-04-22
feat(test-optimization): create final_status tag on test event for vitest

cbasitodx:
What does this PR do?
Create a @test.final_status tag that is added to every vitest test event that represents the final execution of that test (there can be multiple executions of a single test due to retry mechanisms being active)

Motivation
Tracers send each test execution as a span to the backend. When retry mechanisms are enabled a single test can run several times with different outcomes. These mechanisms differ in behavior. Some stop once the test reaches the desired state such as ATR ending when a test passes. Others always run a fixed number of attempts such as EFD or Attempt to fix.

To ensure the correct final result for the customer some intermediate outcomes are suppressed and the testing framework receives an adjusted final status. For example if an early EFD attempt passes but later attempts fail the test is still reported as passed and the failing attempts are hidden to avoid failing CI pipelines.

Both external and internal users have requested a way to query tests in Datadog by this final status. Their main goal is to build monitors and alerts for hard failures on default branches.

To support this a new tag will be added to the last retry of a test named: @test.final_status:pass|fail|skip


packages/datadog-instrumentations/src/vitest.js
```diff
@@ -866,6 +866,9 @@ function wrapVitestTestRunner (VitestTestRunner) {
       } else {
         testPassCh.publish({ task, ...ctx.currentStore })
       }
+      if (shouldFlipStatus) {
+        task.result.state = 'pass'
```
juan-fernandez, line 870: is this an unrelated fix?

packages/datadog-instrumentations/src/vitest.js
```diff
@@ -1174,7 +1177,7 @@ addHook({
           })
         } else if (state === 'pass' && !isSwitchedStatus) {
           if (testCtx) {
-            testPassCh.publish({ task, ...testCtx.currentStore })
+            testPassCh.publish({ task, finalStatus: 'pass', ...testCtx.currentStore })
```
juan-fernandez: in this case, passing quarantined/disabled tests can be tagged with `test.final_status=pass`. A quarantined test that actually passes, or an ATF + disabled/quarantined test whose final attempt passes, reports `pass` instead of the expected management final status `skip`. Mocha/Cypress handle quarantined/disabled as skip regardless of actual execution result.

cbasitodx: Great catch! thanks

packages/datadog-instrumentations/src/vitest.js
cbasitodx: It was affecting the final status tag report, but it could pass as an unrelated fix :) 

This is doing the same as in line 848-855: forcefully reseting the status so 'pass' in the last test. This was done for intermediate executions, but not for the last one. That's why it was affecting my code (I only look at last executions)

packages/datadog-instrumentations/src/vitest.js
cbasitodx: Pushed a fix and a test case :)

packages/datadog-instrumentations/src/vitest.js
juan-fernandez: got it. Is this reflected in any test? This should've been caught by a test

cbasitodx: It fails on these test, for example `sets final_status tag to test status reported to test framework on last retry (EFD active only)`

juan-fernandez (approved):
great job!

PR #8065, opened 2026-04-22
chore(project): upgrade typescript for docs

sabrenner:
### What does this PR do?

Upgrades the version of Typescript used in our `docs/` directory. Makes a small modification to the `tsconfig.json` so that it is compatible with the new major version.

### Motivation

Updated project dependencies.


docs/package.json
```diff
@@ -4,13 +4,13 @@
   "main": "typedoc.js",
   "scripts": {
     "build": "typedoc ../index.d.ts",
-    "pretest": "tsc -p . && tsc --types node test",
```
BridgeAR, line 7: Is that not needed anymore?

sabrenner: from what i can tell, no, we just transpile in the pretest step and it works fine, i think in combination with changing the "baseUrl" property. if you look at the output, running `npm run type:doc:test` does run the transpile and test together and passes
```
npm run type:doc:test

> dd-trace@6.0.0-pre type:doc:test
> cd docs && yarn && yarn test

yarn install v1.22.22
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.04s.
yarn run v1.22.22
$ tsc -p .
$ node test
✨  Done in 1.89s.
```

BridgeAR (approved):

PR #8067, opened 2026-04-22
chore(ci): run all system-tests scenario even if empty when running on master/sc…

nccatoni:
### What does this PR do?
Runs all system-tests scenarios, even the empty ones, when running on master or on schedule.

### Motivation
This gives feedback on easy wins and makes it easier to monitor that no scenario was dropped by accident.

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -61,5 +61,5 @@ jobs:
       scenarios_groups: tracer-release
       excluded_scenarios: APM_TRACING_E2E_OTEL,APM_TRACING_E2E_SINGLE_SPAN # require AWS and datadog credentials
       parametric_job_count: 8
-      skip_empty_scenarios: true
+      skip_empty_scenarios: ${{ github.event_name != 'push' && github.event_name != 'schedule' }}
       push_to_test_optimization: ${{ github.actor != 'dependabot[bot]' }}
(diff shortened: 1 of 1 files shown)


BridgeAR (commented):
Can we not track this separately? I am confused that we have to look into it when we already know there are non to run?

nccatoni:
@BridgeAR I'm not sure what you mean. Would you rather have a second call to the workflow specifically for scheduled runs?

BridgeAR:
@nccatoni I meant why do we have to run the empty system tests? If non apply, we already know that, right?

So looking at the motivation:

> This gives feedback on easy wins and makes it easier to monitor that no scenario was dropped by accident.

Why do we not already know that without running them? Triggering an empty set means we know it is empty so we can just 'monitor' if something is an empty set in the system tests directly, no?

nccatoni:
@BridgeAR By “empty,” I mean that all tests are deactivated, but some still run when the scenario executes. Their results are simply ignored in the final status.

Running these scenarios allows us to monitor deactivated tests that are actually passing, which indicates they should likely be reactivated.

Additionally, we want to ensure that all scenarios are consistently executed in Datadog. Since the information from the manifest isn’t accessible there, running every scenario helps confirm that nothing was accidentally omitted.

simon-id (commented):
The AI comment looks reasonable, but idk if it's true, so double check that then LGTM

simon-id (approved):

BridgeAR (approved):

PR #8069, opened 2026-04-22
feat(llmobs): support `evalScope` on `submitEvaluation`

sabrenner:
### What does this PR do?
Adds support for specifying the evaluation scope (`trace` or `span`, defaulting to `span`) when submitting an evaluation via `submitEvaluation`.

### Motivation
Supporting trace-level evaluations in the Node.js SDK.


packages/dd-trace/src/llmobs/sdk.js
```diff
@@ -401,6 +401,10 @@ class LLMObs extends NoopLLMObs {
         err = 'invalid_metadata'
         throw new Error('metadata must be a JSON object')
       }
+      if (!['trace', 'span'].includes(evalScope)) {
```
BridgeAR: Should we move these two to the top and use a set instead? :)

cdfox:
hey Sam, a couple things:
- I think we should also add a method for calling the span export api, so that the user can easily download all spans for the trace to use in computing trace-level eval, similar to what we did on Python SDK: https://github.com/DataDog/dd-trace-py/pull/17582
- could you also mention what manual testing you did?

sabrenner:
converting back to draft until / when i can push up the get spans api

PR #8071, opened 2026-04-22
test: add integration coverage harness

BridgeAR:
[test(profiler): coerce numeric argv values in oom.js](https://github.com/DataDog/dd-trace-js/pull/8071/changes/1b54b2898236cf2be97898d7fa962612193f41c8)
`process.argv` slots arrive as strings; the OOM fixture compared and
multiplied them as numbers. A non-numeric `sizeQuantum` would flip the
`count < maxCount` comparison the wrong way once `maxCount` came
through as a string.

The worker spawn also hard-coded `maxOldGenerationSizeMb: 50` instead
of forwarding the caller-supplied value, so callers asking for a
larger heap silently got 50 MiB.

[test(integration): add integration coverage harness](https://github.com/DataDog/dd-trace-js/pull/8071/changes/177b7249483361b432f30ebc80dc95412d0eb214)
NYC's standard "wrap the test runner" model stops at the first `fork`,
so the integration suites — which spawn Node processes via `spawnProc`,
`fork`, and sandbox bring-ups — never count any production code
exercised in their descendants. The blind spot covers all of
`integration-tests/` and every plugin's `integration-test/`.

The harness pre-instruments dd-trace once at sandbox pack time
(per-child require-hook overhead is what makes the naive approach
unusable in CI), patches `child_process` and `worker_threads` so every
descendant inherits the bootstrap, and re-pushes the flush handler to
the tail of `'exit'` so plugin teardown that touches dd-trace files
inside its own `'exit'` handler still increments counters.

Three supporting changes:

1. `nyc.config.js` excludes `datadog-plugin-cypress/src/support.js`.
   Cypress's webpack bundles it browser-side, and istanbul's
   `compact: true` collapses it from ~300 lines to one, shifting
   every spec's line numbers in the emitted bundle.
2. `eslint.config.mjs` lints the harness (otherwise hidden by the
   global `coverage/` ignore) and forbids `require(cond ? 'a' : 'b')`:
   static-analysis bundlers only see literal `require` arguments.
   `packages/dd-trace/src/exporter.js` is reshaped to
   `cond ? require('a') : require('b')` to satisfy the rule.
3. `scripts/verify-{coverage,exercised-tests}.js` learn the
   `coverage/node-<version>[-<label>]` layout, the `.skipped` matrix
   sentinel, and `:coverage` sibling-script pairing.

[test(integration): wire :coverage scripts and CI workflows](https://github.com/DataDog/dd-trace-js/pull/8071/changes/f940d96b5c8dec9100c8009c0d5fa972d2257710)
Every `test:integration*` script gains a `:coverage` sibling, and
every workflow that runs one switches to the sibling and chains
`./.github/actions/coverage` to upload the lcov.

Two cases the script siblings alone do not cover:

1. The `apm-integrations.yml` restify job grows the same upload — its
   integration tests were uninstrumented. `plugins/upstream` drops
   its coverage upload: the upstream suite runs a third-party harness
   that produces no NYC coverage for dd-trace.
2. The four Azure-functions plugin specs that bypass `spawnProc`
   (they shell out to `func` through `/bin/sh`) drop their manual env
   helper. The unconditional spawn patch overlays `NODE_OPTIONS` on
   every command, so the inner Node started by `func` inherits the
   bootstrap automatically.

[test(integration): adjust specs for the coverage harness](https://github.com/DataDog/dd-trace-js/pull/8071/changes/8345bd19fc4c2dceac1f10e7bc040ad63b5b70bc)
The bootstrap adds enough child startup overhead to require two
opt-outs even after pre-instrumentation. While checking each
time-sensitive spec, we also caught two long-standing nuisances that
were exposed by the harness.

Four changes:

1. `mocha/mocha.spec.js` opts the DI retry test out via
   `_DD_TRACE_INTEGRATION_COVERAGE_DISABLE=1`. The bootstrap delays
   child startup enough that the single DI retry races the Inspector
   probe arming.
2. `code-origin.spec.js` opts out for the same reason istanbul's
   transform always did: NYC rewrites the compiled JS, so
   `--enable-source-maps` cannot map back to the original `.ts`
   regardless of whether the rewrite happens at pack time or require
   time.
3. `opentelemetry.spec.js` replaces an unconditional
   `setTimeout(..., 1000)` before the first HTTP probe with a
   `getWithRetry` helper. Polling is strictly better than a fixed
   delay regardless of the harness.
4. `cypress/*` and `playwright/*` drop the redundant
   `this.retries(2)`. The suite-wide retry already covers it; under
   coverage they stacked, tripling wall time for no benefit.

Drive-by fix:

* `profiler/profiler.spec.js` extracts the OOM heap budget into
  `OOM_HEAP_MB = 50` so the constant lives in one place.


rochdev (dismissed):
This makes an already borderline unacceptable CI run time completely unacceptable. To my knowledge these tests should actually be less costly than unit tests for adding coverage as the heavy part of integration tests is the setup, not the tests themselves, so I'm actually surprised by this. Hopefully it's a bug in the implementation and there is a quick fix for it, but otherwise this is absolutely not worth it just for better coverage. I'd even rather have 0 coverage information than a CI that takes that long.

.github/workflows/instrumentation.yml
```diff
@@ -412,7 +412,15 @@ jobs:
       # Disable core dumps since some integration tests intentionally abort and core dump generation takes around 5-10s
       - uses: ./.github/actions/install
       - run: sudo sysctl -w kernel.core_pattern='|/bin/false'
-      - run: yarn test:integration:${{ matrix.bundler }}
+      - if: matrix.bundler == 'esbuild'
+        run: yarn test:integration:esbuild:coverage
+      - if: matrix.bundler == 'webpack'
```
BridgeAR, line 414:
```suggestion
      # Getting webpack coverage is not yet supported. TODO: FIXME
      - if: matrix.bundler == 'webpack'
```

package.json
```diff
@@ -186,12 +202,15 @@
...
     "husky": "^9.1.7",
+    "istanbul-lib-report": "^3.0.0",
+    "istanbul-reports": "^3.0.2",
     "jszip": "^3.10.1",
     "mocha": "^11.6.0",
     "mocha-junit-reporter": "^2.2.1",
     "mocha-multi-reporters": "^1.5.1",
     "multer": "^2.1.1",
     "nock": "^13.5.6",
+    "node-preload": "^0.2.1",
```
BridgeAR, line 215: Check why we need this

packages/datadog-plugin-azure-durable-functions/test/integration-test/client.spec.js
BridgeAR: We could just instrument spawnProc instead

integration-tests/opentelemetry.spec.js
```diff
@@ -294,10 +309,9 @@ describe('opentelemetry', function () {
         DD_TELEMETRY_HEARTBEAT_INTERVAL: '1',
       },
     })
-    await new Promise(resolve => setTimeout(resolve, 1000)) // Adjust the delay as necessary
-    await axios.get(`http://localhost:${SERVER_PORT}/first-endpoint`)
+    await getWithRetry(`http://localhost:${SERVER_PORT}/first-endpoint`, 10_000 * COVERAGE_SLOWDOWN)
```
BridgeAR: This seems questionable. Retrying is probably not needed. Why do we need a delay before running the command?

integration-tests/coverage-child-process.spec.js
```diff
@@ -0,0 +1,542 @@
...
+      JSON.parse(fs.readFileSync(syncOut, 'utf8')),
+      { bootstrap: true, hasNycConfig: true }
+    )
+  })
+
+  // Regression: during sandbox bring-up we patch `execSync` (used by `bun add`, `cp`, etc.), so
+  // `resolveCoverageRoot` gets called with the sandbox folder _before_ dd-trace has been
+  // installed into it. We must NOT cache that miss — otherwise every later lookup (including
+  // the teardown one that locates the sandbox's NYC output) returns the cached miss and falls
+  // back to the repo root, and the sandbox's coverage JSON files are silently orphaned.
```
BridgeAR: Ask LLM to clean up tests and make more readable

integration-tests/profiler/oom.js
```diff
@@ -6,10 +6,12 @@ require('dd-trace').init()
 const { Worker, isMainThread, threadId } = require('worker_threads')
 
 const nworkers = Number(process.argv[2] || 0)
-const workerMaxOldGenerationSizeMb = process.argv[3]
+const workerMaxOldGenerationSizeMb = Number(process.argv[3]) || 0
```
BridgeAR:
```suggestion
const workerMaxOldGenerationSizeMb = Number(process.argv[3]) || 50
```

integration-tests/mocha/mocha.spec.js
```diff
@@ -3747,6 +3750,9 @@ describe(`mocha@${MOCHA_VERSION}`, function () {
               './dynamic-instrumentation/test-hit-breakpoint',
             ]),
             DD_CIVISIBILITY_FLAKY_RETRY_COUNT: '1',
+            // Opt out: NYC's require-hook slows child startup enough that the single DI retry
+            // finishes before the Inspector probe is armed.
+            DD_TRACE_INTEGRATION_COVERAGE_DISABLE: '1',
```
BridgeAR: Is this still true with the latest version where the instrumentation already happens before?

integration-tests/helpers/index.js
```diff
@@ -39,7 +46,11 @@ const defaultStopProcTimeoutMs = 2_000
  * @param {string} expectedSource
  */
 async function runAndCheckOutput (filename, cwd, expectedOut, expectedSource) {
-  const proc = spawn(process.execPath, [filename], { cwd, stdio: 'pipe' })
+  const proc = spawn(process.execPath, [filename], {
+    cwd,
+    env: applyCoverageEnv(undefined, { cwd, scriptPath: filename }),
```
BridgeAR: Do we still need this? We should be able to rely on the global child_process instrumentation? Same for the other parts

integration-tests/debugger/snapshot-time-budget.spec.js
```diff
@@ -7,6 +7,8 @@ const {
 } = require('../../packages/dd-trace/src/debugger/devtools_client/snapshot/constants')
 const { setup } = require('./utils')
 
+const FUZZ_TEST_TIMEOUT_MS = process.env.DD_TRACE_INTEGRATION_COVERAGE_ROOT ? 4000 : 2000
```
BridgeAR: Use generic multiplier

integration-tests/coverage/runtime.js
```diff
@@ -0,0 +1,248 @@
+'use strict'
+
+const { existsSync, mkdirSync, readFileSync, realpathSync, rmSync } = require('node:fs')
+const path = require('node:path')
+const { fileURLToPath } = require('node:url')
+
+const FLUSH_SIGNAL_KEY = '__ddCovFlush'
+const COLLECTOR_ENV = 'DD_TRACE_INTEGRATION_COVERAGE_COLLECTOR'
```
BridgeAR:
```suggestion
const COLLECTOR_ENV = '_DD_TRACE_INTEGRATION_COVERAGE_COLLECTOR'
```

integration-tests/coverage/runtime.js
```diff
@@ -0,0 +1,248 @@
...
+const DISABLE_ENV = 'DD_TRACE_INTEGRATION_COVERAGE_DISABLE'
+const ROOT_ENV = 'DD_TRACE_INTEGRATION_COVERAGE_ROOT'
```
BridgeAR:
```suggestion
const DISABLE_ENV = '_DD_TRACE_INTEGRATION_COVERAGE_DISABLE'
const ROOT_ENV = '_DD_TRACE_INTEGRATION_COVERAGE_ROOT'
```

integration-tests/coverage/runtime.js
```diff
@@ -0,0 +1,248 @@
...
+const REPO_ROOT = path.resolve(__dirname, '..', '..')
+const CHILD_BOOTSTRAP_PATH = path.join(__dirname, 'child-bootstrap.js')
+const BOOTSTRAP_REQUIRE_ARG = `--require=${CHILD_BOOTSTRAP_PATH}`
+
+const PRE_INSTRUMENTED_SENTINEL = '.nyc-pre-instrumented'
+// POSIX-style so `path.isAbsolute` matches on both POSIX and Windows.
+const PRE_INSTRUMENTED_ROOT = '/__DD_TRACE_PRE_INSTRUMENTED__'
+
+const COVERAGE_SLOWDOWN = process.env[ROOT_ENV] ? 2.5 : 1
```
BridgeAR: Consider 2 as factor and check if it is stable

integration-tests/coverage/runtime.js
```diff
@@ -0,0 +1,248 @@
...
+ * @param {NodeJS.ProcessEnv | undefined} env
+ * @param {object} [options]
+ * @param {string} [options.cwd]
+ * @param {string | URL} [options.scriptPath]
+ * @returns {NodeJS.ProcessEnv | undefined}
+ */
+function applyCoverageEnv (env, options = {}) {
+  if (!isCoverageActive()) return env
+  const baseEnv = env || process.env
+  if (baseEnv[DISABLE_ENV]) {
```
BridgeAR, line 213: Do we still need this now that we auto instrument all?

integration-tests/coverage/pre-instrumented-writer.js
```diff
@@ -0,0 +1,62 @@
...
+  let flushed = false
+  const flush = () => {
+    if (flushed) return
+    flushed = true
+    try {
+      writeFileSync(outFile, JSON.stringify(rawCoverage))
+    } catch {}
+  }
+
+  process.on('exit', flush)
```
BridgeAR: In theory: manipulate process.on to always be the last handler

integration-tests/coverage/child-bootstrap.js
```diff
@@ -0,0 +1,138 @@
...
+    }
+  })
+  const channel = /** @type {RefCountedChannel} */ (/** @type {unknown} */ (process.channel))
+  channel?.unrefCounted?.()
+}
+
+/**
+ * @returns {void}
+ */
+function patchWorkerThreads () {
```
BridgeAR: This should be consolidated with child_process patching

integration-tests/coverage/child-bootstrap.js
```diff
@@ -0,0 +1,138 @@
...
+    return true
+  }
+}
+
+function bootstrapCoverage () {
+  const coverageRoot = resolveCoverageRoot({ cwd: process.env[ROOT_ENV] || process.cwd() })
+  if (!coverageRoot) return
+
+  const preInstrumented = isPreInstrumentedSandbox(coverageRoot)
+  const foreignNyc = hasForeignNyc()
```
BridgeAR, line 46: Would it not be possible to also track that? We just accept the different NYC and we see what ever file is created at the end of the process and use that?

integration-tests/coverage/finalize-sandbox.js
```diff
@@ -0,0 +1,121 @@
...
+      throw err
+    }
+    if (content.length > 0) {
+      try {
+        return JSON.parse(content)
+      } catch {}
+    }
+    await setTimeout(READ_RETRY_DELAY_MS)
+  }
+}
```
BridgeAR, line 46: Is this really needed? Can we not guarantee a file is there or not? Or should we care if it fails? How big is the risk?

package.json
BridgeAR: It is needed for a custom pre-load

rochdev:
Thanks for addressing my performance concern! The slowdowns seems to still be present but small enough to be within acceptable range 😄 

However, going through the actual logic, this seems like a hell of a lot of code just to add coverage, it almost seems we're re-implementing nyc ourselves. We already have custom tooling for unit tests through mocha-parallel, and now this. Functionally it's great, but it adds a ton of maintenance burden when something goes wrong, as we've already seen in mocha-parallel which is also simpler.

Why can't we just run nyc in integration tests like we do anywhere else? Can that be clearly explained in the PR? I'd rather use tools as close as they are designed as possible to avoid maintaining thousands of lines of code like this with all the issues that will come with it.

integration-tests/helpers/index.js
```diff
@@ -332,8 +344,12 @@ function waitForProcExit (proc, timeoutMs) {
  * @returns {SpawnedProcess}
  */
 function spawnProcImpl (filename, options, stdioHandler, stderrHandler) {
-  // Cast to SpawnedProcess type - when stdio is 'pipe', stdout/stderr are guaranteed non-null
-  const proc = /** @type {SpawnedProcess} */ (fork(filename, { ...options, stdio: 'pipe' }))
+  // When stdio is 'pipe', stdout/stderr are guaranteed non-null.
+  const proc = /** @type {SpawnedProcess} */ (fork(filename, {
+    ...options,
+    execArgv: options.execArgv,
```
BridgeAR:
```suggestion
```

bengl (dismissed):

bengl (approved):

bengl (approved):

PR #8074, opened 2026-04-22
refactor(opentelemetry): simplify OTLP traces setup per PR #7531 review

BridgeAR:
refactor(opentelemetry): simplify OTLP traces setup per PR #7531 review

- Hardcoding defaults and DD_AGENT_HOST-based URLs diverged from the OTel
  spec, which defines a fixed localhost:4318 default. Moving defaults to
  the calculated section keeps the unified-agent model users rely on:
  setting DD_AGENT_HOST or DD_TRACE_AGENT_URL also steers OTLP traffic,
  matching what the parametric system tests expect.
- Having config pre-build signal subpaths for some exporters but not
  others made the layers inconsistent. Putting subpath handling in setUrl
  keeps the knowledge where it's already used.
- Building the OTLP exporter in proxy.js leaked OTLP concerns into code
  that shouldn't know about exporter selection; DatadogTracer already
  picks an exporter from config for the agent case.
- internalPropertyName and otelTracesEnabled added naming layers with no
  benefit; the latter also emitted misleading telemetry for a derived
  value. OTel header envs are now `type: map`, so parsing lives in one
  place (the config parser) and telemetry reports the parsed entries
  instead of a raw string.
- The OTel `key=value,key=value` header format belongs in the config
  parser, not in a per-exporter workaround. Removing
  `#parseAdditionalHeaders` from the exporter base lets the constructor
  take the already-parsed map directly.
- Declarative "allowed" validation is easier to maintain than a runtime
  warning block, and the alias machinery already handles generic→specific
  fallback, so the custom code was redundant.
- The sampler ternary was dead code — OTEL_TRACES_SAMPLER's default
  already covers the opt-in path.
- Aligning the OTel generic envs' implementation letters keeps the
  fallback semantics consistent across headers, endpoint, timeout, and
  protocol.

Trace transformer:
- deployment.environment.name is the stable OTel convention; keep up.
- error.type was being dropped during the DD→OTel transform, losing
  information users expect to see.
- Loose JSDoc types hid bugs at review time; allocating per span for
  root detection and copying bytes by hand was avoidable overhead.
- Added TODOs where follow-up work is known: batching (per-trace HTTP
  requests are unsuitable for production traffic) and base64 meta_struct
  (needs revisiting once protobuf lands).

Tests:
- process.setMaxListeners(50) worked around a safeguard rather than
  fixing the underlying listener leak. Building exporters directly
  avoids spinning up a tracer per test, which was the real source of
  the leak.
- The integration test could leave a promise unhandled on failure and
  used imperative transforms that hid what the payload actually looked
  like.
- Asserting `status.code === 0` on OTLP protobuf was too strict —
  default values get dropped on the wire. Checking "unset or zero"
  matches reality without being lax about anything meaningful.


packages/dd-trace/src/config/supported-configurations.json
```diff
@@ -3913,16 +3913,17 @@
       {
         "implementation": "A",
         "type": "string",
-        "default": null,
-        "internalPropertyName": "otelTracesUrl"
+        "default": "http://localhost:4318",
```
zacharycmontoya: FYI, there are 3 OTLP protocols with two default values. I'm not sure how to best encode this in our support-configurations:
- protocol: `grpc`, default: `http://localhost:4317`
- protocol: `http/protobuf`, default: `http://localhost:4318/v1/traces`
- protocol: `http/json`, default: `http://localhost:4318/v1/traces` (same as http/protobuf)

EDIT: I've updated the HTTP defaults to include the `/v1/traces` subpath. This differs from the `OTEL_EXPORTER_OTLP_ENDPOINT` configuration which does not include the `/v1/traces` subpath for the same protocols

packages/dd-trace/src/config/supported-configurations.json
```diff
@@ -3954,7 +3954,7 @@
       {
         "implementation": "A",
         "type": "string",
-        "default": null,
```
zacharycmontoya, line 3957: This has similar defaults to the [traces endpoint](https://github.com/DataDog/dd-trace-js/pull/8074/changes#r3127326739), except the HTTP subpath is `/v1/logs`

packages/dd-trace/src/config/supported-configurations.json
```diff
@@ -3999,7 +3999,7 @@
...
```
zacharycmontoya, line 4002: This has similar defaults to the [traces endpoint](https://github.com/DataDog/dd-trace-js/pull/8074/changes#r3127326739), except the HTTP subpath is `/v1/metrics`

packages/dd-trace/src/opentelemetry/trace/otlp_http_trace_exporter.js
```diff
@@ -8,13 +8,20 @@ const OtlpTraceTransformer = require('./otlp_transformer')
...
  *
- * This implementation follows the OTLP HTTP v1.7.0 specification:
+ * This implementation follows the OTLP HTTP specification:
  * https://opentelemetry.io/docs/specs/otlp/#otlphttp
  *
  * It receives DD-formatted spans (from span_format.js), transforms them
  * to OTLP ExportTraceServiceRequest JSON format, and sends them to the
  * configured OTLP endpoint via HTTP POST.
  *
+ * TODO: Add batch handling similar to the OpenTelemetry SDK Batch Processor
```
zacharycmontoya, line 18: +1, thanks for adding this

integration-tests/opentelemetry-traces.spec.js
```diff
@@ -112,14 +106,19 @@ describe('OTLP Trace Export', () => {
...
-    assert.strictEqual(dbSpan.name, 'db.query')
-    assert.strictEqual(errSpan.name, 'error.operation')
-
-    // Validate span kind (server=2, client=3 per OTLP proto SpanKind enum)
-    assert.strictEqual(webSpan.kind, 2, 'web.request should be SERVER kind')
-    assert.strictEqual(dbSpan.kind, 3, 'db.query should be CLIENT kind')
+    assertObjectContains(webSpan, {
+      name: 'GET /api/test',
+      kind: 2, // SERVER
+      status: { code: 0 },
```
zacharycmontoya: nit: We may actually consider omitting the status assertion. If no status is set, the protocol [says](https://github.com/open-telemetry/opentelemetry-proto/blob/85e63b1ad6d0667e48707e8c0a88f366e79a68ce/opentelemetry/proto/trace/v1/trace.proto#L301-L303) that it may either appear with `{ code: 0 }` or it may not be present. We only need to assert this if we want to assert that the JS tracer should always do one approach.

```suggestion
```

packages/dd-trace/src/config/index.js
```diff
@@ -605,21 +590,6 @@ class Config extends ConfigBase {
       }
     }
 
-    const DEFAULT_OTLP_PORT = '4318'
-    if (!this.otelLogsUrl) {
-      setAndTrack(this, 'otelLogsUrl', `http://${agentHostname}:${DEFAULT_OTLP_PORT}`)
```
zacharycmontoya, line 610: The reason we originally calculated individual `otelLogsUrl`, `otelMetricsUrl`, and `otelTracesUrl` (as well as analogous protocol counterparts) is because the configurations of the Otlp Exporter can be set with a signal-agnostic configuration but they can be overridden by the signal-specific one (e.g. `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` overrides `OTEL_EXPORTER_OTLP_PROTOCOL`).

Let's make sure we don't lose that resolution logic in these changes. Perhaps we'll need more unit tests as regression tests

BridgeAR: That still works as before. It is handled as fallback alias.

packages/dd-trace/src/config/supported-configurations.json
BridgeAR: This is handled in code a bit special at the moment that the path is added at runtime by checking if the url has one or not. Seemingly, the default may not be localhost as the spec says when there is a agent, so I am reverting parts for now (I guess that deviation from the spec is fine).

integration-tests/opentelemetry-traces.spec.js
BridgeAR: I addressed that by validating it is either unset or has a falsy code (so zero)

packages/dd-trace/src/config/index.js
```diff
@@ -605,19 +590,17 @@ class Config extends ConfigBase {
       }
     }
 
-    const DEFAULT_OTLP_PORT = '4318'
+    // Default OTLP endpoints follow the configured agent host so users who point DD at a custom
```
BridgeAR, line 594:
```suggestion
    // TODO: This could likely be moved to the base class and allow easier GRPC handling
    // Default OTLP endpoints follow the configured agent host so users who point DD at a custom
```

packages/dd-trace/src/config/parsers.js
```diff
@@ -13,6 +13,14 @@ const VALID_PROPAGATION_STYLES = new Set([
   'datadog', 'tracecontext', 'b3', 'b3 single header', 'b3multi', 'baggage', 'none',
 ])
 
+const RENAMED_OTEL_TAGS = new Map(
+  [
+    ['deployment.environment', 'env'],
```
BridgeAR, line 19:
```suggestion
    ['deployment.environment.name', 'env'],
    ['deployment.environment', 'env'],
```

zacharycmontoya (approved):
LGTM from the spec side of things 👍🏼

PR #8078, opened 2026-04-23
Pin datadog-ci for junit upload

juan-fernandez:
### What does this PR do?
Pins the `datadog-ci version` used by `DataDog/junit-upload-github-action` to [5.13.1](https://github.com/DataDog/datadog-ci/releases/tag/v5.13.1) in `.github/actions/push_to_test_optimization/action.yml and .github/workflows/platform.yml`.

### Motivation
Some jobs intermittently fail in the `push_to_test_optimization` step after the upgrade to `DataDog/junit-upload-github-action` v3.0.0. Example in https://github.com/DataDog/dd-trace-js/actions/runs/24780644771/job/72510523178?pr=8060.
The failure happens inside `DataDog/install-datadog-ci-github-action` while resolving the floating `datadog-ci` version v5, where `jq` receives a non-release JSON payload and crashes with `Cannot index string with string "prerelease"`. Pinning `datadog-ci` to 5.13.1 avoids that floating-version GitHub release lookup while preserving the v3 junit upload action behavior.

### DIFF
--- a/.github/actions/push_to_test_optimization/action.yml
+++ b/.github/actions/push_to_test_optimization/action.yml
@@ -17,4 +17,5 @@ runs:
       uses: DataDog/junit-upload-github-action@24449d01fc01e721fa36ccd2caa3caae6922f0e8 # v3.0.0
       with:
         api_key: ${{ inputs.dd_api_key }}
+        datadog-ci-version: 5.13.1
         service: dd-trace-js-tests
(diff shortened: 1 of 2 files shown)


BridgeAR (commented):
I can see that this fixes the immediate issue.

I believe it just causes a new issue, which is that the version would not automatically be updated anymore and we would for example not receive the latest security updates and dependabot would also not know to update this.

What about changing `junit-upload-github-action` to use a specific pinned version, if not specified and we just update that each time there is a new CI release with that version being the new default. That way the benefit would also exist for our users the same as for us :)

BridgeAR (approved):

PR #8079, opened 2026-04-23
feat(aws-sdk, llmobs): support Bedrock Converse and ConverseStream

PROFeNoM:
Adds LLMObs support for AWS Bedrock [`ConverseCommand`](https://github.com/aws/aws-sdk-js-v3/blob/d8fbfbc133a0d7e3bce3075b39b151c25ef82734/clients/client-bedrock-runtime/src/commands/ConverseCommand.ts#L1074) and [`ConverseStreamCommand`](https://github.com/aws/aws-sdk-js-v3/blob/d8fbfbc133a0d7e3bce3075b39b151c25ef82734/clients/client-bedrock-runtime/src/commands/ConverseStreamCommand.ts#L356), bringing Node.js to parity with dd-trace-py (PRs DataDog/dd-trace-py#12560, DataDog/dd-trace-py#12873, DataDog/dd-trace-py#13756).

👉 **[Datadog UI traces — Python (existing integration) vs Node.js (this PR), same operations side-by-side](https://dd.datad0g.com/llm/traces?query=%40ml_app%3Abedrock-converse-demo%20%40event_type%3Aspan%20%40parent_id%3Aundefined&agg_m=count&agg_m_source=base&agg_t=count&fromUser=true&refresh_mode=paused&selectedTab=overview&spanId=8016518324950164513&start=1779085380000&end=1779086220000&paused=true)**

image

Sample comparison (Left=Python; right=NodeJS)

> **Depends on #8082** (`feat(llmobs): add tool_definitions support to Tagger`) — merge that first.

### What gets emitted

One `llm`-kind LLMObs span per Converse call, named `bedrock-runtime.command`. Stream and non-stream produce the same span shape; the stream path assembles it from the [`ConverseStreamOutput`](https://github.com/aws/aws-sdk-js-v3/blob/d8fbfbc133a0d7e3bce3075b39b151c25ef82734/clients/client-bedrock-runtime/src/models/models_0.ts#L4655) event union ([`messageStart`](https://github.com/aws/aws-sdk-js-v3/blob/d8fbfbc133a0d7e3bce3075b39b151c25ef82734/clients/client-bedrock-runtime/src/models/models_0.ts#L4559) → [`contentBlockStart`](https://github.com/aws/aws-sdk-js-v3/blob/d8fbfbc133a0d7e3bce3075b39b151c25ef82734/clients/client-bedrock-runtime/src/models/models_0.ts#L4529) → [`contentBlockDelta`](https://github.com/aws/aws-sdk-js-v3/blob/d8fbfbc133a0d7e3bce3075b39b151c25ef82734/clients/client-bedrock-runtime/src/models/models_0.ts#L4382) → [`contentBlockStop`](https://github.com/aws/aws-sdk-js-v3/blob/d8fbfbc133a0d7e3bce3075b39b151c25ef82734/clients/client-bedrock-runtime/src/models/models_0.ts#L4547) → [`messageStop`](https://github.com/aws/aws-sdk-js-v3/blob/d8fbfbc133a0d7e3bce3075b39b151c25ef82734/clients/client-bedrock-runtime/src/models/models_0.ts#L4571) → [`metadata`](https://github.com/aws/aws-sdk-js-v3/blob/d8fbfbc133a0d7e3bce3075b39b151c25ef82734/clients/client-bedrock-runtime/src/models/models_0.ts#L4619)).

Example span for a Converse call with a system prompt + tool:

```json
{
  "name": "bedrock-runtime.command",
  "meta": {
    "span.kind": "llm",
    "model_name": "anthropic.claude-3-haiku-20240307-v1:0",
    "model_provider": "amazon_bedrock",
    "input": {
      "messages": [
        { "role": "system", "content": "You are an expert swe..." },
        { "role": "user", "content": "Explain distributed tracing." }
      ]
    },
    "output": {
      "messages": [
        {
          "role": "assistant",
          "tool_calls": [{
            "name": "fetch_concept",
            "arguments": { "concept": "distributed tracing" },
            "tool_id": "tooluse_...",
            "type": "toolUse"
          }]
        }
      ]
    },
    "tool_definitions": [
      { "name": "fetch_concept", "description": "...", "schema": { "json": {...} } }
    ],
    "metadata": {
      "temperature": 0.5,
      "max_tokens": 512,
      "stop_reason": "tool_use"
    }
  },
  "metrics": {
    "input_tokens": 364,
    "output_tokens": 55,
    "total_tokens": 419,
    "cache_read_input_tokens": 0,
    "cache_write_input_tokens": 0
  }
}
```

Content in both input and output follows Bedrock's [`ContentBlock`](https://github.com/aws/aws-sdk-js-v3/blob/d8fbfbc133a0d7e3bce3075b39b151c25ef82734/clients/client-bedrock-runtime/src/models/models_0.ts#L2832) tagged union. Supported variants (`text`, `toolUse`, `toolResult`) are decoded into the shape above. Unsupported ones (`image`, `document`, `video`, `reasoningContent`, `citationsContent`, `searchResult`, `cachePoint`) become inline markers on non-stream responses only — `[Unsupported content type: <type>]` for top-level blocks ([`dd-trace-py utils.py:279`](https://github.com/DataDog/dd-trace-py/blob/5b9c36eac44b767c3acfa36fda5b3bec7c614024/ddtrace/llmobs/_integrations/utils.py#L279)) and `[Unsupported content type(s): <keys>]` inside tool results ([`dd-trace-py utils.py:266`](https://github.com/DataDog/dd-trace-py/blob/5b9c36eac44b767c3acfa36fda5b3bec7c614024/ddtrace/llmobs/_integrations/utils.py#L266)). The stream aggregator silently ignores unhandled [`ContentBlockDelta`](https://github.com/aws/aws-sdk-js-v3/blob/d8fbfbc133a0d7e3bce3075b39b151c25ef82734/clients/client-bedrock-runtime/src/models/models_0.ts#L4253) variants, matching dd-trace-py.

<details>
<summary>Known disparities with dd-trace-py not addressed in this PR</summary>

Verified side-by-side against `dd-trace-py` on a demo app running both tracers end-to-end against the Datadog UI. The differences below were identified and noted, but ignored.

#### Service name

Python hardcodes `aws.bedrock-runtime` (ignores `DD_SERVICE`; uses `DD_BOTOCORE_SERVICE=aws` + endpoint-name suffix). Node respects `DD_SERVICE`, as shown in the screenshot above.

Why not acted on: pre-existing divergence, unrelated to Converse. Changing either side is a service-naming policy call, and a breaking change; not something that belongs to that PR. 

#### `model_name` / `model_provider`

Python splits via `parse_model_id` → `(anthropic, claude-3-haiku-20240307-v1:0)`; backend matches `claude-3-haiku` / `anthropic`. Node sends the full lowercased `modelId` + literal `amazon_bedrock` provider; the backend falls back to the raw string because `amazon_bedrock` isn't a recognized foundation-model provider.

Why not acted on: pre-existing Node behavior from PR #7952 (`fix(llmobs): fix missing estimated cost on Bedrock LLM spans`, commit `2bac2030d`), unchanged by this PR. 

#### `stop_reason` on `ConverseStream`

image

Python captures `stopReason` for non-stream `Converse` but drops it on the stream path (`bedrock.py:304-308`). Node emits `stop_reason` uniformly for both paths.

Why not acted on: Node is doing better parity than Python here tbh.

#### APM span name, resource case, tag prefix

image

| | Python | Node |
|---|---|---|
| Span name | `bedrock-runtime.command` | `aws.request` |
| Resource | `ConverseStream` (PascalCase) | `converseStream` (camelCase) |
| Bedrock tags prefix | `bedrock.request.*` | `aws.bedrock.request.*` |

Why not acted on: not introduced by this PR. Unchanged since the Bedrock plugin was introduced. A rename would be a breaking change.

</details>

### Test plan

- [x] `PLUGINS="aws-sdk" SPEC="bedrockruntime.spec.js" npm run test:plugins` — green
- [x] LLMObs spec: `packages/dd-trace/test/llmobs/plugins/aws-sdk/bedrockruntime.spec.js` — green
- [x] `npm run lint` on touched files — clean
- [x] End-to-end verified on Datadog UI via companion demo app (`demo-apps/bedrock-converse`) — see live-traces link at the top of this description
- [x] Full CI


packages/datadog-plugin-aws-sdk/test/bedrockruntime.spec.js
```diff
@@ -97,6 +97,44 @@ describe('Plugin', () => {
...
+              'aws.bedrock.request.model': converseRequest.modelId.split('.')[1],
+              'aws.bedrock.request.model_provider': converseRequest.provider.toLowerCase(),
+            },
+          })
+
+          const result = await bedrockRuntimeClient.send(command)
+          for await (const _event of result.stream) { // eslint-disable-line no-unused-vars
+            // drain
+          }
+          await tracesPromise
```
BridgeAR, line 137: This is a very small risk for having an unhandled rejection in case the stream fails.

This is unlikely and the fix is not nice: the tracePromise needs a catch handler above to not trigger the unhandled rejection.

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
```diff
@@ -476,12 +475,206 @@ function extractTextAndResponseReason (response, provider, modelName) {
...
+  for (const block of contentBlocks || []) {
+    if (block == null || typeof block !== 'object') continue
+    if (typeof block.text === 'string') {
+      content += block.text
+    } else if (block.toolUse) {
+      toolCalls.push(buildToolCall(block.toolUse))
+    } else if (block.toolResult) {
+      toolResults.push(buildToolResult(block.toolResult))
+    } else {
+      const type = Object.keys(block)[0] || 'unknown'
```
sabrenner: isn't there a block.type we can use? or am i getting this mixed up with anthropic 😅

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
```diff
@@ -476,12 +475,206 @@ function extractTextAndResponseReason (response, provider, modelName) {
...
 }
 
+/**
+ * Convert a Converse content-block array to an LLMObs message array.
+ *
+ * @param {string} role
+ * @param {Array<object>} contentBlocks
+ * @returns {Array<{ content?: string, role: string, toolCalls?: Array, toolResults?: Array }>}
+ */
+function extractMessagesFromConverseContent (role, contentBlocks) {
```
sabrenner, line 485: maybe this could just return the message instead of returning an array of length one, and let callers decide if it should be an array or not (as it seems in some places we spread the single result into a `push`, and in others we care about its length as an array)

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
```diff
@@ -476,12 +475,206 @@ function extractTextAndResponseReason (response, provider, modelName) {
...
+    } else if (chunk.messageStop?.stopReason) {
+      stopReason = chunk.messageStop.stopReason
+    } else if (chunk.metadata?.usage) {
+      usage = chunk.metadata.usage
+    } else if (chunk.contentBlockStart?.start?.toolUse) {
+      const { contentBlockIndex, start: { toolUse } } = chunk.contentBlockStart
+      blocksByIdx.set(contentBlockIndex, {
+        toolUse: { toolUseId: toolUse.toolUseId, name: toolUse.name, inputStr: '' },
+      })
+    } else if (chunk.contentBlockDelta) {
```
sabrenner, line 656: are tool calls included here? i see we have branches here for tool use but not tool calls

sabrenner (commented):
lgtm! just a couple of questions, then good to go

sabrenner:
also looks like a handful of lines are not covered by tests, which is in the codecov report. if you click on that job and click on the report, it should show you what lines are not being hit by tests

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
PROFeNoM: Well, I don't really see any "clear" primitive here unfortunately. 

Basically, what i'm leveraging in `Object.keys(block)[0]` is the fact that in the [ContentBlock interfaces](https://github.com/aws/aws-sdk-js-v3/blob/e55a38791745be6ed4be7009808d14a2a393adf9/clients/client-bedrock-runtime/src/models/models_0.ts#L2832-L3108), there is always exactly one key, all the other ones being marked as `never`.

So for instance, for a `VideoMember`, `Object.keys(block)` would be `['video']`, and hence `Object.keys(block)[0]` would be be `'video'`.

Right now, i've extracted it to a helper function so that it's perhaps a less bit cryptic... but i actually don't see a cleaner alternative here. Opened to suggestions though :+1:

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
PROFeNoM: Good question, but I believe it's a terminology thing: in the Bedrock Converse API there's no separate "tool call" concept. tool call == toolUse.

According to their [toolUse block documentation](https://github.com/aws/aws-sdk-js-v3/blob/e55a38791745be6ed4be7009808d14a2a393adf9/clients/client-bedrock-runtime/src/models/models_0.ts#L2798-L2802):
> A tool use content block. Contains information about a tool that the model is requesting be run., The model uses the result from the tool to generate a response.

So here what we're doing when both parts of a streamed tool call arrives are:
- `contentBlockStart.start.toolUse` (the branch just above) opens the block with `toolUseId`.+ `name` (since [toolUse](https://github.com/aws/aws-sdk-js-v3/blob/e55a38791745be6ed4be7009808d14a2a393adf9/clients/client-bedrock-runtime/src/models/models_0.ts#L4475) here is a [ToolUseBlockStart](https://github.com/aws/aws-sdk-js-v3/blob/e55a38791745be6ed4be7009808d14a2a393adf9/clients/client-bedrock-runtime/src/models/models_0.ts#L4433-L4454))
- `contentBlockDelta.delta.toolUse.input` (this branch) carries the arguments. Here [toolUse](https://github.com/aws/aws-sdk-js-v3/blob/e55a38791745be6ed4be7009808d14a2a393adf9/clients/client-bedrock-runtime/src/models/models_0.ts#L4259) is a [ToolUseBlockDelta](https://github.com/aws/aws-sdk-js-v3/blob/e55a38791745be6ed4be7009808d14a2a393adf9/clients/client-bedrock-runtime/src/models/models_0.ts#L4241-L4248), whose input is a `string` - unlike the non-stream [ToolUseBlock.input](https://github.com/aws/aws-sdk-js-v3/blob/e55a38791745be6ed4be7009808d14a2a393adf9/clients/client-bedrock-runtime/src/models/models_0.ts#L2802-L2825) which is a structured `__DocumentType`. We accumulate those string deltas (into our internal `inputStr` below) and JSON.parse the whole thing once the block ends.

In the end, this allows us to reassemble the block and `buildToolCall` turns it into the tool call.

So, unless i missunderstood what you meant, tool calls are covered: they're the `toolUse` branches (split across the two events the SDK emits for the identity and arguments respectively).

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
PROFeNoM: done

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
PROFeNoM: I've just quickly normalized the stream tool input so `buildToolCall ` sees one shape, so that `inputStr` becomes a private detail of the stream function in the hope of making it a tiny bit cleaner...

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
sabrenner: ah yeah ok no this is fine, thanks for the link and explanation!

packages/datadog-plugin-aws-sdk/src/services/bedrockruntime/utils.js
sabrenner: 🤦 i forgot about the terminology mis-match, this seems good then!

packages/dd-trace/test/llmobs/plugins/aws-sdk/bedrockruntime.converse.spec.js
sabrenner, line 1: is this file supposed to cover the missing coverage from the codecov report (i don't remember seeing it in the diff the last time i reviewed). if possible, we should put the tests directly in the `bedrockruntime.spec.js` file, as we don't want to spy on the tagger directly. but if i'm mistaken, lmk!

PROFeNoM: > is this file supposed to cover the missing coverage from the codecov report 

yes

> if possible, we should put the tests directly in the bedrockruntime.spec.js file, as we don't want to spy on the tagger directly. but if i'm mistaken, lmk!

ok i'll do it. I initially didn't to it because i followed the pattern found in `bedrockruntime.token-cache.spec.js` and i didnt want to manually "hand-write cassette responses"... I'll see what i can do

PROFeNoM: So i dropped the tagger-spy spec entirely. Moved those into two tests:
- A hand-authored VCR test whose response carries an unmapped `reasoningContent` block and an unknown tool-result item. It's plain JSON so it's ok enough i'd say to be handcrafted
- However, for the "malformed stream tool-use JSON" test... it's a bit more annoying to go through VCR cleanly since it'd mean hand-building a base64 body. It's not impossible at all, just super ugly and annoying to maintain imo. So i've instead done a small, direct unit test of the exported stream extractor.

I believe it's an ok compromise, right?

sabrenner: that seems reasonable to me, yeah some of these edge cases are hard to hit with actual generated cassettes instead of handcrafted, so i think this sgtm!

sabrenner (approved):

PR #8087, opened 2026-04-23
test: reduce CI flakiness via span-race filters

BridgeAR:
A real race fixes that stop assuming `traces[0][0]` is the span we're looking for across kafkajs, confluentinc-kafka-javascript, redis, ioredis, mariadb, mysql, mysql2, graphql (parse + validate), aws-sdk s3 span-pointers, elasticsearch, opensearch, mongodb-core, pg, prisma, iovalkey, and the IAST `testOutsideRequestHasVulnerability` and `checkVulnerabilityInRequest` helpers. Moleculer's `startBroker` now retries on EADDRINUSE to cover the documented port race, and the rhea `before` hook finally returns the `agent.load` promise so mocha actually waits for plugin registration.

### DIFF
--- a/packages/datadog-plugin-aws-sdk/test/s3.spec.js
+++ b/packages/datadog-plugin-aws-sdk/test/s3.spec.js
@@ -83,7 +83,8 @@ describe('Plugin', () => {
           it('should add span pointer for putObject operation', (done) => {
             agent.assertSomeTraces(traces => {
               try {
-                const span = traces[0][0]
+                const span = traces[0].find(s => s.meta?.['aws.operation'] === 'putObject')
+                assert.ok(span)
                 const links = JSON.parse(span.meta?.['_dd.span_links'] || '[]')
 
                 assert.strictEqual(links.length, 1)
@@ -113,7 +114,8 @@ describe('Plugin', () => {
           it('should add span pointer for copyObject operation', (done) => {
             agent.assertSomeTraces(traces => {
               try {
-                const span = traces[0][0]
+                const span = traces[0].find(s => s.meta?.['aws.operation'] === 'copyObject')
+                assert.ok(span)
                 const links = JSON.parse(span.meta?.['_dd.span_links'] || '[]')
 
                 assert.strictEqual(links.length, 1)
@@ -187,9 +189,8 @@ describe('Plugin', () => {
                 s3.completeMultipartUpload(completeParams, (err) => {
                   if (err) done(err)
                   agent.assertSomeTraces(traces => {
-                    const span = traces[0][0]
-                    const operation = span.meta?.['aws.operation']
-                    if (operation === 'completeMultipartUpload') {
+                    const span = traces[0].find(s => s.meta?.['aws.operation'] === 'completeMultipartUpload')
+                    if (span) {
                       try {
                         const links = JSON.parse(span.meta?.['_dd.span_links'] || '[]')
                         assert.strictEqual(links.length, 1)
--- a/packages/datadog-plugin-moleculer/test/index.spec.js
+++ b/packages/datadog-plugin-moleculer/test/index.spec.js
@@ -10,8 +10,8 @@ const { withNamingSchema, withPeerService, withVersions } = require('../../dd-tr
 const { expectedSchema, rawExpectedSchema } = require('./naming')
 const sort = trace => trace.sort((a, b) => Number(a.start - b.start))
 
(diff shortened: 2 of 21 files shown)


packages/dd-trace/test/plugins/agent.js
```diff
@@ -226,10 +226,16 @@ function handleTraceRequest (req, res, sendToTestAgent) {
...
+
+// Probe once per worker; the test agent doesn't come and go mid-run.
 function checkAgentStatus () {
-  return new Promise((resolve) => {
+  if (agentStatusPromise !== undefined) return agentStatusPromise
+
+  agentStatusPromise = new Promise((resolve) => {
     const agentUrl = process.env.DD_TRACE_AGENT_URL || 'http://127.0.0.1:9126'
-    const timeoutMs = 2000
+    const timeoutMs = 500
```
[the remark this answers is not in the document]
rochdev: @BridgeAR That comment seems to make sense at a high level, probably only the first _successful_ request should be cached. Why do we even need to query this to begin with?

wconti27 (commented):
LGTM tho i will let @rochdev  be the final approver

packages/dd-trace/test/plugins/agent.js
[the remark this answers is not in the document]
BridgeAR: I looked at it again and I think we should try to minimize the waiting time even for test-optimization code. I set it to 1 second now. Let's see if that is fine.

It might not matter at all anymore due to the way everything now works.

Re-probing does not help though. We need to rerun the test anyway.

packages/datadog-plugin-moleculer/test/index.spec.js
```diff
@@ -23,13 +23,29 @@ function getPort () {
   })
 }
 
+async function withPortRetry (fn, attempts = 5) {
```
rochdev: This has been proven times and times again to not work. Let's remove `get-port` usage instead.

rochdev: Looks like ServiceBroker supports `port: null` and it will get a random port from the kernel instead, and then it should be available on `broker.transport.options.port`.

BridgeAR: Fixed

packages/dd-trace/test/plugins/agent.js
```diff
@@ -226,35 +226,45 @@ function handleTraceRequest (req, res, sendToTestAgent) {
   })
 }
 
+/** @type {Promise<boolean> | undefined} */
+let agentStatusPromise
+
+// Probe once per worker, and only when a test agent is expected to be up
+// (signalled by `DD_TEST_AGENT_URL`). `testagent/start` already waits for
+// the agent to answer `/info` before tests run, so a single generous
```
rochdev: Not a blocker as it was already there, but why is this even needed? It's meant for when we don't know what version of the agent is available, but in tests we should know that already so this just slows down things.

rochdev (dismissed):

packages/dd-trace/test/plugins/agent.js
BridgeAR: I am working on removing it as a follow-up

BridgeAR:
I limited this to span race filters and I move the agent changes to a separate PR

packages/datadog-plugin-aws-sdk/test/s3.spec.js
```diff
@@ -83,7 +83,8 @@ describe('Plugin', () => {
           it('should add span pointer for putObject operation', (done) => {
             agent.assertSomeTraces(traces => {
               try {
-                const span = traces[0][0]
+                const span = traces[0].find(s => s.meta?.['aws.operation'] === 'putObject')
```
wconti27, line 86: all changes look good to me, tho im wondering if we should consider a helper function for matching the correct span assertion?

wconti27 (approved):

CarlesDD (approved):
LGTM from IAST prespective

packages/datadog-plugin-aws-sdk/test/s3.spec.js
BridgeAR: I think that makes sense, yes.
We have something like that already and improving our current mechanism is good!
That way the tests would generally run a tad faster because we generate less intermediate errors.

Right now, I would land it as is as stop gap.

PR #7757, opened 2026-03-12
refactor(graphql): migrate shimmer to orchestrion instrumentation

crysmags:
## What does this PR do?

Migrates the graphql plugin's instrumentation strategy from `shimmer.wrap` to [orchestrion](https://github.com/DataDog/orchestrion-js) AST rewriting. The plugin no longer dynamically wraps individual field resolvers at execute time; instead orchestrion rewrites `graphql`'s `execute`, `parse`, and `validate` entry points at module load to publish on diagnostic channels. The plugin subscribes to those channels and owns the per-execute / per-resolver work in `packages/datadog-plugin-graphql/src/execute.js`.

Also wires up two public API additions (`FieldContext` type + `hooks.resolve` user callback) and adds one benchmark variant for measuring shared dd-trace startup cost separately from graphql plugin overhead.

---

## Commits

| # | commit | what it does |
|---|---|---|
| 1 | `feat(graphql): migrate from shimmer to orchestrion instrumentation` | The full migration: orchestrion config, plugin restructure, IAST/AppSec correctness fixes, perf optimizations, depth=0 fast path |
| 2 | `feat(graphql): expose FieldContext and hooks.resolve public types` | Public TypeScript declarations for the new resolve hook |
| 3 | `bench(graphql-long): add tracer-loaded graphql-disabled baseline variant + microbench harness` | Adds a `with-tracer-graphql-disabled` variant for shared-cost subtraction, plus a microbench harness for local plugin work |

---

### Before (master)

- `packages/datadog-instrumentations/src/graphql.js` uses `shimmer.wrap` against `graphql.execute`, then walks the schema's type map *at execution time* to wrap every field resolver
- Per-field wrap fires for every resolver call, regardless of whether a span will be created
- Separate `GraphQLResolvePlugin` sub-class handles resolver spans via channel subscriptions

### After (this PR)

- **`packages/datadog-instrumentations/src/helpers/rewriter/instrumentations/graphql.js`** (new) — orchestrion config declaring which functions to rewrite (`execute`, `parse`, `validate`) for both `graphql` (CJS + ESM, versions ≥0.10) and `@graphql-tools/executor` (for graphql-yoga)
- **`packages/datadog-instrumentations/src/graphql.js`** — stripped to just `addHook` registrations + module-export captures for the IAST taint-tracking signature tools (`printer`, `visitor`, `utilities`)
[...]

PR #8089, opened 2026-04-23
bench: add plugin-graphql-long sirun benchmark

crysmags:
## Summary

Replaces `benchmark/sirun/plugin-graphql/` (6-query, startup-dominated) with `benchmark/sirun/plugin-graphql-long/` (100 queries per process), so per-query cost dominates the measurement rather than one-time setup cost.

Motivated by **#7757** (graphql plugin refactor), where the 6-query bench reports a regression that inverts into a large improvement under production-like workloads.

## Why

The existing `plugin-graphql` bench fired 6 queries per process then exited. At 6 queries, load was ~80% of the CPU budget — any change that traded startup cost for steady-state cost looked like a regression on this benchmark even when it made real applications faster. Keeping the 6-query bench preserved a misleading signal that reviewers had already learned to discount, so this PR removes it outright rather than running both.

### Case study: #7757 — master vs PR branch

Same sirun setup, same variants, two different process sizes (Node 22.10.0, `user.time` metric, median):

| variant | master @ 6q | #7757 @ 6q | Δ @ 6q | master @ 100q | #7757 @ 100q | Δ @ 100q |
|---|---|---|---|---|---|---|
| `with-depth-off` | 282.5 ms | 294.4 ms | 🟥 **`+4.2%`** | 1632.6 ms | **1210.3 ms** | 🟩 **`−25.9%`** |
| `with-depth-on-max` | 278.0 ms | 292.2 ms | 🟥 **`+5.1%`** | 1605.7 ms | **1169.4 ms** | 🟩 **`−27.2%`** |
| `with-depth-and-collapse-off` | 402.7 ms | 449.3 ms | 🟥 **`+11.6%`** | 2437.9 ms | **2358.2 ms** | 🟩 **`−3.3%`** |
| `with-depth-and-collapse-on` | 280.2 ms | 291.9 ms | 🟥 **`+4.2%`** | 1612.4 ms | **1187.6 ms** | 🟩 **`−26.3%`** |

Break-even: ~15–20 queries per process. Past that, #7757 is strictly faster. The 6-query bench was reporting only the 🟥 column.

### Why load share matters

| variant | load share @ 6q | load share @ 100q | Δ |
|---|---|---|---|
| `with-depth-on-max` | 🟥 **`83%`** | 🟩 **`23%`** | `−60pp` |
| `with-depth-and-collapse-off` | 🟥 **`77%`** | 🟩 **`16%`** | `−61pp` |

At `QUERIES=100`, load drops from 🟥 ~80% to 🟩 ~16–23% — per-query cost dominates. `QUERIES` is parametric so the same directory can be re-aimed at larger sizes (e.g. `QUERIES=500` for load <6%) without forking.

## What's in the change

- **Adds `benchmark/sirun/plugin-graphql-long/`**:
  - `index.js` — parametric on `QUERIES` (default 100); env-var truthy checks use `process.env.X` so `WITH_DEPTH=0` reads as configured (Codex review fix).
  - `meta.json` — same six variants as the old bench, `iterations: 5`.
  - `schema.js` — moved from the old `plugin-graphql/` location since it's the only consumer.
  - `README.md` — short sibling-style description.
- **Removes `benchmark/sirun/plugin-graphql/`** entirely.

## CI impact

`benchmark/sirun/runall.sh` auto-discovers subdirs (`DIRS=($(ls -d */ | sed 's:/$::'))`) and runs the 6 variants in parallel as backgrounded jobs pinned to separate cores (`(time node ../run-one-variant.js …) &`). Wall time per Node version equals the slowest variant, not the sum.

After this PR:
- ❌ `scenario:plugin-graphql-*` — removed.
- ✅ `scenario:plugin-graphql-long-*` — new steady-state signal.

| bench | per-Node wall (parallel-by-core) | bottleneck variant |
|---|---|---|
| old `plugin-graphql` (6q × 30 iters) | ~45 s | `with-depth-and-collapse-off` |
| new `plugin-graphql-long` (100q × 5 iters) | **~15 s** | `with-depth-and-collapse-off` |

Per iteration the long bench is ~3 s for the slowest variant; 5 iterations × 1 process per iteration ≈ 15 s on the busy core. Light variants (control, async-hooks) finish in ~3 s and idle.

## Test plan

- [ ] Verified locally with `sirun meta-temp.json` on Node 22.10.0 / Node 25 — all 6 variants run.
- [ ] Load/steady-state ratio measured; `QUERIES=100` keeps load 🟩 ≤23% across all variants (worst case: `with-depth-on-max`).
- [ ] `iterations=5` chosen as the smallest count where stddev/mean stays acceptable post-warmup; confirmed in local sirun output.
- [ ] Lint passes (eslint).

Follow-up on #7757: once this lands, #7757 will rebase to inherit the bench from master, and CI will produce comparable `plugin-graphql-long-*` numbers against the master baseline.


benchmark/sirun/plugin-graphql-long/index.js
```diff
@@ -0,0 +1,51 @@
+'use strict'
+
+// Long-workload graphql bench. Runs QUERIES sequential queries per process
+// (default 100) so the fixed startup cost doesn't dominate the measurement.
+// See ./README.md.
```
BridgeAR, line 5: I wanted to have another benchmark type as well where we can deduct the startup time. Right now, that is dominant in almost all our benchmarks and that is why they are not useful.

I asked @bengl a long time ago about implementing the type and he opened a PR in dd-trace-js (look at the oldest ones). It just needs to be tracked in our benchmark backend as a known type. That way this would also be solved while not needing to increase the runtime drastically. While having a combined view is also helpful.

benchmark/sirun/plugin-graphql-long/README.md
```diff
@@ -0,0 +1,49 @@
...
+bench (~45s per Node version) while dropping load share from ~80% to
+~20%.
+
+## Relationship to `plugin-graphql`
+
+The existing `plugin-graphql` benchmark (6 queries) is preserved
+unchanged — it catches regressions in startup / init / schema-walk
+paths. The new `plugin-graphql-long` complements it by catching
+regressions in steady-state per-query paths. Both run on every PR via
+`benchmark/sirun/runall.sh` automatic subdir discovery.
```
BridgeAR: This could be part of the PR description or the commit description, but keeping it in the repo as markdown seems not ideal to me.
It would get stale easily, depending on how the repo is changed over time.

What is kept could actually also be moved inside of the test file as comment. That way it is easier to follow.
```suggestion
```

benchmark/sirun/plugin-graphql-long/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+if (Number(process.env.WITH_TRACER)) {
+  const tracer = require('../../..').init()
+
+  if (Number(process.env.WITH_DEPTH)) {
```
BridgeAR:
```suggestion
  if (process.env.WITH_DEPTH) {
```

benchmark/sirun/plugin-graphql-long/index.js
```diff
@@ -0,0 +1,51 @@
+'use strict'
+
+// Long-workload graphql bench. Runs QUERIES sequential queries per process
+// (default 100) so the fixed startup cost doesn't dominate the measurement.
+// See ./README.md.
+
+if (Number(process.env.WITH_TRACER)) {
```
BridgeAR:
```suggestion
if (process.env.WITH_TRACER) {
```

benchmark/sirun/plugin-graphql-long/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+  if (Number(process.env.WITH_DEPTH)) {
+    tracer.use('graphql', { depth: Number(process.env.WITH_DEPTH) })
+  } else if (process.env.WITH_DEPTH_AND_COLLAPSE) {
+    const [depth, collapse] = process.env.WITH_DEPTH_AND_COLLAPSE.split(',')
+    tracer.use('graphql', { depth: Number(depth), collapse: Number(collapse) > 0 })
+  }
+}
+
+if (Number(process.env.WITH_ASYNC_HOOKS)) {
```
BridgeAR:
```suggestion
if (process.env.WITH_ASYNC_HOOKS) {
```

benchmark/sirun/plugin-graphql-long/index.js
```diff
@@ -0,0 +1,51 @@
...
+        name
+      }
+    }
+  }
+}
+`
+
+const variableValues = { who: 'world' }
+
+const queries = Number(process.env.QUERIES || 100)
```
BridgeAR:
```suggestion
const queries = process.env.QUERIES ? Number(process.env.QUERIES) : 100)
```

BridgeAR (commented):
Almost LGTM. The envs do not need the number parsing in most situations (we just need the existence check) and the README.md should likely be removed. I don't think it is good to land that in our repo. Instead, that could be part of the commit message :)

benchmark/sirun/plugin-graphql-long/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+// Long-workload graphql bench. Runs QUERIES sequential queries per process
+// (default 100) so the fixed startup cost doesn't dominate the measurement.
+// See ./README.md.
+
+if (Number(process.env.WITH_TRACER)) {
+  const tracer = require('../../..').init()
+
+  if (Number(process.env.WITH_DEPTH)) {
+    tracer.use('graphql', { depth: Number(process.env.WITH_DEPTH) })
```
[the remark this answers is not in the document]
crysmags: Addressed in 929427783: switched the four env-var truthy checks (`WITH_TRACER`, `WITH_DEPTH`, `WITH_ASYNC_HOOKS`, `QUERIES`) from `Number(process.env.X)` to plain `process.env.X`. `WITH_DEPTH="0"` now reads as "set" and the `with-depth-off` variant correctly calls `tracer.use('graphql', { depth: 0 })` instead of falling through to plugin defaults.

benchmark/sirun/plugin-graphql-long/index.js
crysmags: Thanks for the context — looked at @bengl's older work to understand the shape of the "deduct startup" approach. The two relevant ones I found:

- #5597 — original `shimmer-startup` benchmark
- #5613 — cut-down version so it fits in the bench platform

`shimmer-startup` doesn't deduct startup from another bench; it amortizes startup away by running 100,000 iterations of a trivial wrapped function so per-call cost dwarfs init. Same idea as what `plugin-graphql-long` does, just at a different scale because graphql's per-query work is much heavier than a single shimmer wrap (we top out at ~100 queries before per-Node CI time becomes a problem).

If you'd still rather see startup *subtracted* from the per-query measurement (e.g., a separate `plugin-graphql-startup` bench tracked alongside, with the bench backend computing `plugin-graphql-long − plugin-graphql-startup`), I'm happy to add that as a follow-up — let me know if that matches what you had in mind. For now this PR sticks with amortization, which I think gets us most of the way there: at `QUERIES=100`, startup load drops from ~80% on the 6-query bench to ~16–23% on the long bench, so per-query cost is the dominant signal even without a separate subtraction step.

benchmark/sirun/plugin-graphql-long/README.md
crysmags: Addressed in 5e6201d07: trimmed the README down to a 3-paragraph sibling-style description (matches `plugin-graphql/`, `shimmer-runtime/`, `encoding/`, etc.). The methodology + case-study tables it used to carry have been folded into the PR description above instead, where they belong.

benchmark/sirun/plugin-graphql-long/index.js
crysmags: Addressed in 929427783.

benchmark/sirun/plugin-graphql-long/index.js
crysmags: Addressed in 929427783.

benchmark/sirun/plugin-graphql-long/index.js
crysmags: Addressed in 929427783.

benchmark/sirun/plugin-graphql-long/index.js
crysmags: Addressed in 929427783.

benchmark/sirun/plugin-graphql-long/index.js
BridgeAR: I meant https://github.com/DataDog/dd-trace-js/pull/5606

BridgeAR (dismissed):
LGTM

BridgeAR (approved):

PR #8090, opened 2026-04-23
test: test reduced test runtime overhead

BridgeAR:
Remove the test agent and instead create the artifacts file statically. That way we have the upper and lower bounds visible directly in our repository.

The test agent added a tad of time to our CI runs and failed frequently.


packages/dd-trace/test/plugins/agent.js
```diff
@@ -452,9 +342,12 @@ module.exports = {
       next()
     })
 
+    // This is a workaround to ensure that the agent is ready to receive
+    // requests without being cleaned up properly.
+    // TODO: Fix the root cause.
     const innerAgent = agent
 
-    const useTestAgent = await checkAgentStatus()
+    await wait(1)
```
wconti27, line 350: whats the purpose of this wait here?

BridgeAR: See the comment above :)

wconti27 (approved):
The changes look good to me, I really like the new way of outputting our supported versions as well!

bengl (approved):

bengl (approved):

PR #8097, opened 2026-04-24
fix(llmobs): render BaseMessage in langgraph workflow I/O

### LINKED ISSUE #8096: [BUG]: LLM Obs LangGraph plugin renders BaseMessage state as raw JSON — formatIO lacks the BaseMessage handling the sibling LangChain plugin already has
## Tracer Version(s)
`5.98.0`

## Node.js Version(s)
`24.12.0`

## Operating System
`Darwin 24.6.0`

## Bundling
No Bundling

## Bug Report

When `createAgent(...)` from `langchain@1.x` (built on `@langchain/langgraph`) is invoked with LLM Observability enabled, the workflow-kind root span's Input/Output panels render the Pregel state as raw JSON — full `HumanMessage` / `AIMessage` class instances serialized with `additional_kwargs`, `response_metadata`, `id`, `tool_call_id`, etc. — instead of clean `{ content, role }` shapes.

The **child** `chat_model` LLM spans and `tool` spans render correctly. The issue is isolated to the **workflow** span emitted by the LangGraph LLM Obs plugin added in #7567.

### Root cause

An asymmetry between two plugins that should share formatting:

- **LangChain plugin**, `packages/dd-trace/src/llmobs/plugins/langchain/handlers/index.js:35-51` — `getContentFromMessage` already handles `BaseMessage` instances: reads `message.content`, extracts role via `message.getType() / _getType()`, emits `{ content, role }`.
- **LangGraph plugin**, `packages/dd-trace/src/llmobs/plugins/langgraph/index.js:8-32` — `formatIO` is a **separate, simpler implementation** that only recurses plain `Object` / `Array`. `BaseMessage` instances are not plain `Object` (their `constructor.name` is `HumanMessage` / `AIMessage` / etc.) so they fall through to the `JSON.stringify(data)` branch, serializing the entire class.

Because `createAgent(...).invoke({ messages: [...] })` input state and the last stream chunk both contain `BaseMessage` arrays, the workflow span's I/O is dominated by class-instance dumps.

### Expected behavior

The workflow span's I/O should represent `BaseMessage` instances the same way the LangChain plugin does — as `{ content, role }` objects — so the Input/Output panel is readable and consistent with child chat_model spans.

### Suggested fix

Ideally, share a single helper between the two plugins. The cleanest path is to lift `getContentFromMessage` out of `langchain/handlers/index.js` into a shared util (e.g. `llmobs/plugins/shared/messages.js`) and import it from both the LangChain chain/chat_model handlers and the LangGraph `formatIO` — avoiding the duplication that caused this drift in the first place.

If a full refactor is heavier than desired, a minimal duck-typed check inside `langgraph/index.js:formatIO` would restore parity without reaching across plugin boundaries or taking a hard dependency on `@langchain/core`:

```js
function formatIO (data) {
  if (data == null) return ''
  if (typeof data === 'string' || typeof data === 'number' || typeof data === 'boolean') return data

  // NEW: BaseMessage detection (duck-typed; mirrors getContentFromMessage in
  // packages/dd-trace/src/llmobs/plugins/langchain/handlers/index.js)
  if (typeof data?._getType === 'function' || typeof data?.getType === 'function') {
    const role = typeof data.getType === 'function' ? data.getType() : data._getType()
    return { content: data.content ?? '', role }
  }

  if (data.constructor?.name === 'Object') {
    const formatted = {}
    for (const [key, value] of Object.entries(data)) formatted[key] = formatIO(value)
    return formatted
  }
  if (Array.isArray(data)) return data.map(item => formatIO(item))
  try { return JSON.stringify(data) } catch { return String(data) }
}
```

## Reproduction Code

```js
// package.json: "type": "module"
// deps: dd-trace@5.98.0 @langchain/aws @langchain/core @langchain/langgraph langchain zod
import 'dd-trace/register.js';
import ddTrace from 'dd-trace';
import { ChatBedrockConverse } from '@langchain/aws';
import { createAgent } from 'langchain';

ddTrace.init({
  service: 'langgraph-repro',
  plugins: false,
  llmobs: {
    enabled: true,
    mlApp: 'langgraph-repro',
    agentlessEnabled: true, // requires DD_API_KEY + DD_SITE
  },
});
ddTrace.use('langchain');

const agent = createAgent({
  name: 'simple-demo',
  model: new ChatBedrockConverse({
    model: 'anthropic.claude-3-haiku-20240307-v1:0',
    region: 'us-east-1',
  }),
  tools: [],
  systemPrompt: 'You are a helpful assistant.',
});

await agent.invoke({
  messages: [{ role: 'user', content: 'What is OpenTelemetry? One sentence.' }],
});
```

Run with `DD_API_KEY=… DD_SITE=datadoghq.com AWS_REGION=us-east-1` and valid Bedrock credentials. The resulting LLM Obs trace has a workflow span named `simple-demo` whose Input/Output panels render verbose `HumanMessage` / `AIMessage` JSON dumps instead of the clean rendering seen on the child `chat_model` span.

## Tracer Config

```js
ddTrace.init({
  service: '<svc>',
  plugins: false,
  llmobs: { enabled: true, mlApp: '<ml-app>', agentlessEnabled: true },
});
ddTrace.use('langchain');
```

crysmags:
### What does this PR do?

Fixes #8096. The LangGraph LLM Obs plugin's workflow span now renders `BaseMessage` instances (`HumanMessage`, `AIMessage`, `SystemMessage`, …) as `{ content, role }` objects, matching the LangChain plugin's child `chat_model` spans — instead of dumping the full class with `additional_kwargs`, `response_metadata`, `id`, `tool_call_id`, etc. into the Input/Output panels.

### Motivation

The LangGraph plugin (added in #7567) shipped its own `formatIO` that only recursed plain `Object` / `Array`. Since `BaseMessage` subclasses have `constructor.name` of `HumanMessage` / `AIMessage` / etc., they fell through to `JSON.stringify(data)` and serialized the entire class shape. This made the workflow span emitted by `createAgent(...).invoke(...)` / `Pregel.stream` essentially unreadable, even though the child chat_model and tool spans rendered correctly.

### What changed

- Extracted `ROLE_MAPPINGS`, `getRole`, `getContentFromMessage`, and `formatIO` out of `packages/dd-trace/src/llmobs/plugins/langchain/handlers/index.js` into a new shared module at `packages/dd-trace/src/llmobs/plugins/shared/messages.js`.
- Rewired the LangChain handler to delegate to the shared helpers (`this.formatIO` / `this.getContentFromMessage` / `this.getRole` preserved as thin wrappers so `chain.js`, `chat_model.js`, `embedding.js`, `vectorstore.js` keep working unchanged).
- Rewired `packages/dd-trace/src/llmobs/plugins/langgraph/index.js` to use the shared `formatIO`, which now falls through to `getContentFromMessage` for non-plain-object values — handling `BaseMessage` the same way LangChain does.

### Testing

- New unit tests in `packages/dd-trace/test/llmobs/plugins/shared/messages.spec.js` covering `getRole` (explicit role, `_getType`, `getType`), `getContentFromMessage` (string passthrough, BaseMessage → `{ content, role }`, missing content default), and `formatIO` (null / primitive / plain object / array / BaseMessage / nested LangGraph-shape state).
- New regression test in `packages/dd-trace/test/llmobs/plugins/langgraph/index.spec.js` that drives a Pregel graph with real `HumanMessage` / `AIMessage` instances and asserts the workflow span's `input.value` / `output.value` render as `{ content, role }`.
- Unit tests pass locally (13/13). The existing langgraph plugin integration specs depend on a local test agent not present in my environment; they run in CI via the `langgraph` job in `.github/workflows/llmobs.yml`.

### Additional Notes

This is a strictly additive refactor: the LangChain handler's observable behavior is unchanged aside from `formatIO(null)` / `formatIO(undefined)` / `formatIO(primitive)` at the root (previously crashed or returned `{ content: '' }`, now returns `''` or the primitive). Existing LangChain call sites always pass objects, so no test fixtures need to change.


packages/dd-trace/src/llmobs/plugins/langchain/messages.js
```diff
@@ -0,0 +1,65 @@
...
+  if (data.constructor?.name === 'Object') {
+    const formatted = {}
+    for (const [key, value] of Object.entries(data)) {
+      formatted[key] = formatIO(value)
+    }
+    return formatted
+  }
+
+  if (Array.isArray(data)) {
+    return data.map(item => formatIO(item))
```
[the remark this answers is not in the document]
crysmags: Addressed in 5503957ba. `formatIO` now only routes duck-typed BaseMessage instances (`typeof data._getType === 'function' || typeof data.getType === 'function'`) to `getContentFromMessage`. Any other class instance falls back to `JSON.stringify(data)` — matching the pre-#8097 LangGraph behavior so `Document`-like or user-defined objects without a `content` property retain their serialized shape instead of collapsing to `{content: ''}`. Added a regression unit test in `packages/dd-trace/test/llmobs/plugins/shared/messages.spec.js` ("preserves non-message class instances via JSON.stringify...").

packages/dd-trace/src/llmobs/plugins/langchain/handlers/index.js
```diff
@@ -19,35 +19,15 @@ class LangChainLLMObsHandler {
...
-
-      return formatted
-    } else if (Array.isArray(messages)) {
-      return messages.map(message => this.formatIO(message))
-    } // either a BaseMesage type or a string
-    return this.getContentFromMessage(messages)
+    return formatIO(messages)
   }
 
   getContentFromMessage (message) {
```
sabrenner: we dont need this function here anymore since it just calls the utility version of itself. we can just replace the code that calls this function with calling the utility function instead,

packages/dd-trace/src/llmobs/plugins/langchain/handlers/index.js
```diff
@@ -19,35 +19,15 @@ class LangChainLLMObsHandler {
...
-      const role = this.getRole(message)
-      if (role) messageContent.role = role
+    return getContentFromMessage(message)
+  }
 
-      return messageContent
-    } catch {
-      return JSON.stringify(message)
-    }
+  getRole (message) {
```
sabrenner: ditto - we can get rid of this function and have the caller call the utility function

packages/dd-trace/src/llmobs/plugins/langchain/handlers/index.js
```diff
@@ -19,35 +19,15 @@ class LangChainLLMObsHandler {
   setMetaTags () {}
 
   formatIO (messages) {
```
sabrenner: ditto here

packages/dd-trace/src/llmobs/plugins/langchain/messages.js
sabrenner, line 1: nit: i would rename this file `langchain_messages`, just so its clear. tbh we could just keep it under `plugins/langchain` and have langgraph import from that, should be ok!

packages/dd-trace/test/llmobs/plugins/shared/messages.spec.js
sabrenner, line 1: this test file is probs not needed because we test in each of the langchain/langgraph tests

sabrenner (commented):
just some cleanup 🧹 things but lgtm!

packages/dd-trace/src/llmobs/plugins/langchain/handlers/index.js
crysmags: Addressed in 1f6e24687: dropped the `formatIO` wrapper. `chain.js`, `embedding.js`, and `vectorstore.js` now import `formatIO` from `../messages` and call it directly.

packages/dd-trace/src/llmobs/plugins/langchain/handlers/index.js
crysmags: Addressed in 1f6e24687: dropped the `getContentFromMessage` wrapper. It's no longer used outside `messages.js` (only `formatIO` calls it internally), so I also removed it from the public exports — only `formatIO` and `getRole` are exported now.

packages/dd-trace/src/llmobs/plugins/langchain/handlers/index.js
crysmags: Addressed in 1f6e24687: dropped the `getRole` wrapper. `chat_model.js` now imports `getRole` from `../messages` and calls it directly.

packages/dd-trace/src/llmobs/plugins/langchain/messages.js
crysmags: Addressed in 1f6e24687: moved the file to `packages/dd-trace/src/llmobs/plugins/langchain/messages.js` and deleted the empty `plugins/shared/` dir. LangGraph imports from `../langchain/messages`. (Kept the filename as plain `messages.js` since the path `plugins/langchain/` already conveys the scope, and the repo is kebab-case.)

packages/dd-trace/test/llmobs/plugins/shared/messages.spec.js
crysmags: Addressed in 1f6e24687: deleted `packages/dd-trace/test/llmobs/plugins/shared/messages.spec.js`. The new BaseMessage regression case in `packages/dd-trace/test/llmobs/plugins/langgraph/index.spec.js` ("renders BaseMessage input/output as clean { content, role }") plus the existing LangChain integration coverage exercise the same paths end-to-end.

sabrenner (approved):

PR #8098, opened 2026-04-24
split playwright and cypress tests in ci

rochdev:
### What does this PR do?

Split Playwright and Cypress tests in CI.

### Motivation

They're slow.

### Additional Notes

Total time for Test Optimization workflow is down from ~17min to ~6min.


BridgeAR (commented):
I did not review the code, while the coverage report says this is not identical anymore. I fear this might regress for a reason.

```
- Coverage   73.67%   72.45%   -1.22%     
```

rochdev:
> I did not review the code, while the coverage report says this is not identical anymore. I fear this might regress for a reason.

If you look at the report, it lists regression on hundreds of files completely unrelated to Test Optimization. I think the way it calculates might be wrong.

integration-tests/cypress/cypress-atr.spec.js
```diff
@@ -0,0 +1,593 @@
...
+    // TODO: Update tests files accordingly and test with different TS versions
+    useSandbox([`cypress@${version}`, 'cypress-fail-fast@7.1.0', 'typescript@5'], true)
+
+    before(async function () {
+      // Note: Cypress binary is already installed during useSandbox() via the postinstall script
+      // when the cypress npm package is installed, so no explicit install is needed here
+      cwd = sandboxCwd()
+    })
+
+    after(async () => {})
```
juan-fernandez, line 108: is this needed?

integration-tests/cypress/cypress-efd.spec.js
```diff
@@ -0,0 +1,695 @@
...
+    // TODO: Update tests files accordingly and test with different TS versions
+    useSandbox([`cypress@${version}`, 'cypress-fail-fast@7.1.0', 'typescript@5'], true)
+
+    before(async function () {
+      // Note: Cypress binary is already installed during useSandbox() via the postinstall script
+      // when the cypress npm package is installed, so no explicit install is needed here
+      cwd = sandboxCwd()
+    })
+
+    after(async () => {})
```
juan-fernandez, line 105: same as above

juan-fernandez (approved):
thanks! I've been meaning to do this for a while

integration-tests/cypress/cypress-atr.spec.js
rochdev: Hmm good catch, not sure where that came from. I'll merge it as-is since this is a pretty big PR and I want to get it merged as soon as possible, but I'll open a follow-up PR.

PR #8099, opened 2026-04-24
test: reduce agent.load overhead by reducing proxyquire calls

BridgeAR:
The previous \`agent.load\` re-ran a \`proxyquire\` cascade on every
call. The cascade re-evaluated \`datadog-instrumentations\` from the
top, accumulated extra entries in \`instrumentations[name]\`, layered
shimmer wraps on already-patched library functions, and — under
\`nyc\` on Node 20.20.2 — split \`dc.channel(name)\` into multiple
\`Channel\` singletons via V8's WeakRef-backed registry. The
\`child_process\` suite tripped on \`TypeError: Promise is not a
constructor\`after the second \`agent.load\`.

The new shape rebuilds the tracer only when a tracked env var
(\`DD_*\` / \`OTEL_*\` / \`_DD_*\` / serverless platform vars), the
\`tracerConfig\` argument, or an intervening \`agent.close\` differs
from the previous load. Integration files stay cached for the
process; \`agent.spec.js\` pins the single-evaluation and
channel-singleton invariants. \`agent.load\` resolves with the live
\`TracerProxy\` so specs can bind their local \`tracer\` after the
gate-fired rebuild rather than from a stale
\`require('../../dd-trace')\` capture.

The performance gain is roughly 20% in average tests and the
ergonomics are significantly improved by not needing to know
where to get the tracer instance or what to clean up after the
test run.


packages/dd-trace/test/llmobs/sdk/integration.spec.js
```diff
@@ -19,7 +19,10 @@ describe('end to end sdk integration tests', () => {
   const { getEvents, getEvaluationMetrics } = useLlmObs()
 
   before(() => {
-    tracer = require('../../../../dd-trace')
+    // `useLlmObs()` ran `agent.load` in a `before` registered above
+    // this one, so by the time we get here `global._ddtrace` is the
+    // live `TracerProxy`.
+    tracer = global._ddtrace
```
rochdev, line 25: Why not make `useLlmObs` return the tracer as well?

packages/dd-trace/test/plugins/agent.js
```diff
@@ -30,6 +42,86 @@ let plugins = []
...
 let dsmStats = []
 let currentIntegrationName = null
+let loaded = false
+
+// Non-prefix env vars `dd-trace` reads at module load. Hand-maintained: every
+// new read of a non-`DD_*` / non-`OTEL_*` / non-`_DD_*` env in `src/` adds an
+// entry here. The natural follow-up is to auto-populate this from
+// `getEnvironmentVariable` / `getValueFromEnvSources`, but that needs a path
+// that does not pay for the extra Set bookkeeping in production hot calls.
+const TRACKED_NON_PREFIX_ENV_NAMES = new Set([
```
rochdev, line 54: The comment explains the what but not the why. We need a very strong reason to have to maintain such a list, especially as nobody will know about it.

packages/dd-trace/test/plugins/agent.js
```diff
@@ -2,19 +2,31 @@
...
 const { assertObjectContains } = require('../../../../integration-tests/helpers')
 const { storage } = require('../../../datadog-core')
-const ritm = require('../../src/ritm')
+
+// Modules that close over the previous `Config` / `TracerProxy` singletons.
+// Evicted whenever `agent.load`'s gate decides the tracer must rebuild.
+// `datadog-instrumentations/*` and `plugin_manager.js` stay cached so RITM
+// hooks live for the whole process — see `agent.spec.js` for the regression
+// that pins the single-evaluation invariant.
+const RELOAD_EVICTION_IDS = [
```
rochdev, line 21: How was that list determined? Why are only these specific files evicted?

packages/dd-trace/test/llmobs/sdk/integration.spec.js
BridgeAR: You mean as a tracer promise from that call?

I agree that it seems cleaner to also get it from there, while this is the only test that wants access to a tracer using `useLlmObs()`. So it seems to be the exception, not the rule.

packages/dd-trace/test/plugins/agent.js
BridgeAR: I was also not that happy with this list. These values are manipulated for mostly serverless.

Thinking about it again, we can probably get a linter to complain relatively easily, in case a new entry is missing here. We can check `getEnvironmentVariable(s)` and `getValueFromEnvSources` for these. That way we would be safe.

Otherwise it is very hard to detect what is read and what not.

packages/dd-trace/test/plugins/agent.js
BridgeAR: These are the proxyquire calls from before and working through it by changing the tests and seeing the outcome (comparing states).

I can not think of any way to have a way to autodetect new entries here.

packages/dd-trace/test/plugins/agent.js
BridgeAR: I pushed another commit that adds the linter and adds all entries that were not yet in the list.

tlhunter (approved):

PR #8100, opened 2026-04-24
fix(hono): set resource name for single-handler routes

wconti27:
### What does this PR do?
Fixes resource name being set to just the HTTP method (e.g. `"GET"`) instead of `"GET /product"` for Hono routes with no middleware.

### Motivation
Hono's `#dispatch()` has a fast path that calls a matched handler directly when only one handler matches, bypassing `compose()`. Our `wrapCompose` hook never fires in that case, so the route is never published and the span resource name is just the HTTP method. This affects any Hono app where routes have no `app.use()` middleware.

Fix by pre-checking the router match in `wrapFetch` and publishing the route for single-match requests.

### Additional Notes
The existing test suite missed this because every test sets up a global `app.use()` middleware in `beforeEach`, which forces Hono through `compose()` and masked the bug. Added a dedicated test with a bare app (no middleware) to cover this path.

### DIFF
--- a/packages/datadog-instrumentations/src/hono.js
+++ b/packages/datadog-instrumentations/src/hono.js
@@ -14,6 +14,11 @@ const enterChannel = channel('apm:hono:middleware:enter')
 const exitChannel = channel('apm:hono:middleware:exit')
 const finishChannel = channel('apm:hono:middleware:finish')
 
+// Tracks handlers registered via `app.use()` so route-publishing wrappers
+// installed by `wrapRouterAdd` can skip middleware-only matches (a request
+// matching only middleware should keep the bare HTTP-method resource name).
+const middlewareHandlers = new WeakSet()
+
 // `app.request()` and non-node adapters call `app.fetch` without an `incoming`
 // IncomingMessage; the APM `web` helpers depend on one, so the wrappers below
 // skip publishing whenever it is missing.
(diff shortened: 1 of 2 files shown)


packages/dd-trace/test/plugins/agent.js
```diff
@@ -409,6 +518,13 @@ module.exports = {
       listener = server.listen(0, () => {
         const port = listener.address().port
 
+        // OTEL_TRACES_EXPORTER=otlp is commonly set by CI observability tooling. If inherited by
+        // the test process it causes the tracer to use OtlpHttpTraceExporter, which sends spans to
+        // an external OTLP endpoint instead of this mock agent, making assertSomeTraces() hang.
+        // Clear it around tracer.init() so plugin tests always use the standard AgentExporter.
+        const savedOtelTracesExporter = process.env.OTEL_TRACES_EXPORTER
+        delete process.env.OTEL_TRACES_EXPORTER
```
BridgeAR: This seems independent :)

BridgeAR (changes requested):
This definitely fixes an issue and that is good!

The overhead with the implementation is just that we now have to always do the special handling while we could do this at registration time.

We also do not yet handle the `app.all('/foo', () => {})` case.

BridgeAR (changes requested):
This will accidentally undo a recent hono fix.

Could you also check how to find a way to solve this from a less hot code path and a more robust way to detect the arguments in case they change? :)

packages/datadog-instrumentations/src/hono.js
```diff
@@ -27,6 +32,38 @@ function wrapFetch (fetch) {
...
+    if (typeof arg1 === 'function') middlewareHandlers.add(arg1)
+    for (const h of handlers) middlewareHandlers.add(h)
+    return originalUse.call(this, arg1, ...handlers)
+  }
+}
+
+function wrapRouterAdd (originalAdd) {
+  return function (method, path, handlerData) {
+    const handler = handlerData[0]
+    if (!middlewareHandlers.has(handler)) {
```
BridgeAR:
```suggestion
    const handler = handlerData?.[0]
    if (!middlewareHandlers.has(handler) && typeof handler === ''function') {
```

As additional safe guard against changes in the library :)

BridgeAR (commented):
Mainly LGTM!

Mind adding another guard for the `basePath` including a test?

```js
shimmer.wrap(hono.Hono.prototype, 'basePath', basePath => function (path) {
  const clone = basePath.call(this, path)
  shimmer.wrap(clone, 'use', wrapUse)
  return clone
})
```

BridgeAR (approved):
LGTM

PR #8102, opened 2026-04-24
reuse azure function process to lower tests run time

rochdev:
### What does this PR do?

Reuse Azure Function process to lower tests run time.

### Motivation

It takes a long time to start and tests use different paths, so side-effects between them should not occur regardless.

### DIFF
--- a/integration-tests/helpers/fake-agent.js
+++ b/integration-tests/helpers/fake-agent.js
@@ -178,6 +178,7 @@ module.exports = class FakeAgent extends EventEmitter {
     const errors = []
 
     const timeoutObj = setTimeout(() => {
+      this.removeListener('message', messageHandler)
       const errorsMsg = errors.length === 0 ? '' : `, additionally:\n${errors.map(e => e.stack).join('\n')}\n===\n`
       resultReject(new Error(`timeout${errorsMsg}`, { cause: { errors } }))
     }, timeout)
--- a/integration-tests/helpers/fake-agent.js
+++ b/integration-tests/helpers/fake-agent.js
@@ -279,6 +279,55 @@ module.exports = class FakeAgent extends EventEmitter {
     return resultPromise
   }
 
+  /**
+   * Collect span groups matching a predicate after firing a trigger.
+   *
+   * Attaches the listener before invoking `trigger`, so events emitted between request
+   * dispatch and listener registration are not missed.
+   *
+   * @param {object} options
+   * @param {() => Promise<unknown>} options.trigger Fired once the listener is in place.
+   * @param {(group: object[]) => boolean} options.predicate Group-level filter.
+   * @param {number} [options.expectedCount=1] Resolve after this many matching groups arrive.
+   * @param {number} [options.timeout=30_000] Timeout in milliseconds before the promise rejects.
+   * @returns {Promise<object[][]>} The matching groups, in arrival order.
+   */
+  collectGroups ({ trigger, predicate, expectedCount = 1, timeout = 30_000 }) {
+    const groups = []
+    let resolveResult
+    let rejectResult
+
+    const handler = ({ payload }) => {
+      for (const group of payload) {
+        if (predicate(group)) groups.push(group)
+      }
+      if (groups.length >= expectedCount) resolveResult(groups)
+    }
+
+    const timeoutObj = setTimeout(() => {
+      rejectResult(new Error(
+        `timed out waiting for ${expectedCount} matching span groups, got ${groups.length}`
+      ))
+    }, timeout)
+
+    const result = /** @type {Promise<object[][]>} */ (new Promise((resolve, reject) => {
+      resolveResult = (value) => {
+        clearTimeout(timeoutObj)
+        this.removeListener('message', handler)
+        resolve(value)
+      }
+      rejectResult = (error) => {
+        clearTimeout(timeoutObj)
+        this.removeListener('message', handler)
+        reject(error)
+      }
+    }))
+
+    this.on('message', handler)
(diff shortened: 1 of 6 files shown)


integration-tests/helpers/fake-agent.js
```diff
@@ -178,6 +178,7 @@ module.exports = class FakeAgent extends EventEmitter {
     const errors = []
 
     const timeoutObj = setTimeout(() => {
+      this.removeListener('message', messageHandler)
```
BridgeAR, line 181: ❤️

packages/datadog-plugin-azure-functions/test/integration-test/http-test/client.spec.js
```diff
@@ -27,11 +27,15 @@ describe('esm', () => {
     ['./packages/datadog-plugin-azure-functions/test/fixtures/*',
       './packages/datadog-plugin-azure-functions/test/integration-test/http-test/*'])
 
-    beforeEach(async () => {
+    before(async function () {
+      this.timeout(60000)
       agent = await new FakeAgent().start()
+      proc = await spawnPluginIntegrationTestProc(sandboxCwd(), 'func', ['start'], agent.port, undefined, {
+        PATH: `${sandboxCwd()}/node_modules/azure-functions-core-tools/bin:${process.env.PATH}`,
+      })
```
BridgeAR, line 35: ❤️

packages/datadog-plugin-azure-functions/test/integration-test/eventhubs-test/eventhubs.spec.js
```diff
@@ -28,366 +25,379 @@ describe('esm', () => {
...
-    }).timeout(60000)
-  })
-})
+function collectingAssert (fn) {
+  const allGroups = []
+  return ({ payload }) => {
+    allGroups.push(...payload)
+    fn(allGroups)
+  }
+}
```
BridgeAR: I find these helpers difficult to follow. Is there a reason the assertions needed to be changed?

BridgeAR (commented):
The change is mostly LGTM. I am just confused about the helper methods. Why do we need these? I find the former assertion easier to read.

packages/datadog-plugin-azure-functions/test/integration-test/eventhubs-test/eventhubs.spec.js
rochdev: A single function process means a single instance of dd-trace. This means that all the data is sending is shared across tests. These helpers handle that by ensuring that the only the data received for the current test is considered. I'm open to other suggestions as I also don't like them, but it seemed the easiest path forward.

BridgeAR (changes requested):
If we want to keep the helpers, let us move them to the FakeAgent to prevent copying them.

The handler are unsafer than the former behavior and could be replaced with:

```js
/**
 * @param {object} options
 * @param {() => Promise<unknown>} options.trigger Fired once the listener is in place.
 * @param {(group: object[]) => boolean} options.predicate Group-level filter.
 * @param {number} [options.expectedCount] Resolve after this many matching groups arrive.
 * @param {number} [options.timeout]
 * @returns {Promise<object[][]>} The matching groups, in arrival order.
 */
collectGroups ({ trigger, predicate, expectedCount = 1, timeout = 30_000 }) {
  const groups = []
  let resolveResult
  let rejectResult

  const handler = ({ payload }) => {
    for (const group of payload) {
      if (predicate(group)) groups.push(group)
    }
    if (groups.length >= expectedCount) resolveResult(groups)
  }

  const timeoutObj = setTimeout(() => {
    rejectResult(new Error(
      `timed out waiting for ${expectedCount} matching span groups, got ${groups.length}`
    ))
  }, timeout)

  const result = new Promise((resolve, reject) => {
    resolveResult = (value) => {
      clearTimeout(timeoutObj)
      this.removeListener('message', handler)
      resolve(value)
    }
    rejectResult = (error) => {
      clearTimeout(timeoutObj)
      this.removeListener('message', handler)
      reject(error)
    }
  })

  this.on('message', handler)
  trigger().catch(rejectResult)
  return result
}

// Plus a predicate in the file

const azureInvokeGroup = (resource) => (group) =>
  group.some(s => s.name === 'azure.functions.invoke' && s.resource === resource)
```

Result

```js
// Before

it('propagates eventdata through an event hub with a cardinality of one', async () => {
  return curlAndAssertMessage(
    agent,
    'http://127.0.0.1:7071/api/eh1-eventdata',
    collectingAssert(allGroups => {
      const ehGroups = allGroups.filter(g => isEhInvokeGroupFor(g, 'EventHubs eventHubTest1'))
      assert.strictEqual(ehGroups.length, 2)
      for (const group of ehGroups) {
        assertObjectContains(group[0], { /* … */ })
        assert.strictEqual(parseLinks(group[0]).length, 1)
      }
    })
  )
}).timeout(60000)

// After

it('propagates eventdata through an event hub with a cardinality of one', async () => {
  const groups = await agent.collectGroups({
    trigger: () => curl('http://127.0.0.1:7071/api/eh1-eventdata'),
    predicate: azureInvokeGroup('EventHubs eventHubTest1'),
    expectedCount: 2,
  })
  for (const group of groups) {
    assertObjectContains(group[0], { /* … */ })
    assert.strictEqual(parseLinks(group[0]).length, 1)
  }
}).timeout(60000)
```

rochdev:
> If we want to keep the helpers, let us move them to the FakeAgent to prevent copying them. The handler are unsafer than the former behavior and could be replaced with:

Made the request changes.

BridgeAR (approved):

PR #8103, opened 2026-04-24
test,ci: fix langchain being tested against Node.js 18

BridgeAR:
They dropped support for Node.js 18 in 2025 and the versions we support automatically pull in latest peer dependencies that now start to fail.

To fix it, a new action option is added that allows to adjust the lower bound of the testing range.


pabloerhard (approved):
LGTM! I was going to add why just not do what Pino is doing but this seems way cleaner.

mr-lee (approved):

PR #8104, opened 2026-04-25
split tracing-ubuntu job in ci

rochdev:
### What does this PR do?

Split `tracing-ubuntu` job in CI.

### Motivation

The tests are starting to take a significant enough amount of time to justify splitting them.

### DIFF
--- a/.github/workflows/apm-capabilities.yml
+++ b/.github/workflows/apm-capabilities.yml
@@ -48,12 +48,14 @@ jobs:
       id-token: write
     strategy:
       matrix:
-        node-version: [oldest-maintenance-lts, newest-maintenance-lts, active-lts, latest]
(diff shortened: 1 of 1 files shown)


BridgeAR (approved):
I think we could also aim at reducing the runtime of tests that take long instead of adding more machines.

rochdev:
> I think we could also aim at reducing the runtime of tests that take long instead of adding more machines.

Totally agree, but given the sheer amount of tests we have especially in this job, and that it only adds a few machines, this seemed the simplest short-term fix. One thing I want to try next week is to try switching a few of the workflows that we currently split on separate machines using the Mocha parallel script you added a few months ago instead. If that works well, it could save us dozens of machines.

PR #8107, opened 2026-04-26
reduce initial delay to 5 minutes

rochdev:
### What does this PR do?

Reduce initial delay to 5 minutes.

### Motivation

Our CI will be able to run in less than 7 minutes after all speed up PRs land.

### DIFF
--- a/.github/workflows/all-green.yml
+++ b/.github/workflows/all-green.yml
@@ -34,7 +34,7 @@ jobs:
           DELAY: ${{ github.run_attempt == 1 && '5' || '0' }} # 5 minutes on first attempt, no delay on reruns
           GITHUB_PR_NUMBER: ${{ github.event.pull_request.number }}
           GITHUB_TOKEN: ${{ github.token }}
-          POLLING_INTERVAL: 0.5
+          POLLING_INTERVAL: 1
           RETRIES: 30
       - name: Require vendor validation when vendor/ changes
         if: github.event_name == 'pull_request'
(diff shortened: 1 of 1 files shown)


.github/workflows/all-green.yml
```diff
@@ -31,10 +31,10 @@ jobs:
       - run: yarn add @actions/core @actions/github octokit
       - run: node scripts/all-green.mjs
         env:
-          DELAY: ${{ github.run_attempt == 1 && '7' || '0' }} # 7 minutes on first attempt, no delay on reruns
+          DELAY: ${{ github.run_attempt == 1 && '5' || '0' }} # 5 minutes on first attempt, no delay on reruns
           GITHUB_PR_NUMBER: ${{ github.event.pull_request.number }}
           GITHUB_TOKEN: ${{ github.token }}
-          POLLING_INTERVAL: 1
+          POLLING_INTERVAL: 0.5
```
BridgeAR: I don't think we need to poll faster than once a minute

rochdev: Why be slower when we can be faster? As explained above, now that CI will take 5 minutes, 1 minute becomes significant. There is a delay of 5 minutes anyway, so even at 30 seconds it should only poll a few times before finishing. Basically, we're able to shorten the polling interval _and_ reduce the number of API requests at the same time. I don't see how that is not a win.

rochdev: I rolled back the interval and only changed the delay to unblock, we can discuss any other option in a follow up PR.

CarlesDD (approved):

BridgeAR (approved):

PR #8110, opened 2026-04-27
[test optimization] [wip] playwright zero config tia 

juan-fernandez:


packages/dd-trace/src/ci-visibility/exporters/test-worker/index.js
```diff
@@ -41,6 +42,9 @@ function getInterprocessCoverageCode () {
   if (getEnvironmentVariable('JEST_WORKER_ID')) {
     return JEST_WORKER_COVERAGE_PAYLOAD_CODE
   }
+  if (getValueFromEnvSources('DD_PLAYWRIGHT_WORKER')) {
```
BridgeAR, line 45: I am currently trying hard to work on getting rid of the direct access to the env this way. Would it be possible to rework the code to use the config entry instead? Config is a singleton, so it can just be accessed easily and contains the already parsed value. Getting it here again is overhead for checking all the sources again.

I believe the hard part is to rewrite the tests. That is something we just have to do one way or the other.

juan-fernandez: you can ignore the contents of this PR for the moment, as it's purely a work in progress. This was just to test the viability of pushing these kinds of features with coding agents completely hands off 😄

PR #8114, opened 2026-04-27
test(profiler): fix Profiler telemetry flake

szegedi:
### What does this PR do?

Fixes the recurring `profiler > Profiler telemetry > sends profiler API telemetry` integration-test flake (32 of the 39 Profiling failures in the most recent 7-day flakiness report).

- Adds an optional `namespace` parameter to `FakeAgent#assertTelemetryReceived` that filters telemetry messages by their payload namespace at the same point `requestType` is filtered. Backward-compatible — no existing callers pass a 6th argument.
- Threads `namespace: 'profilers'` through the three profiler-telemetry assertions and drops the now-redundant inner `assert.strictEqual(pp.namespace, 'profilers')` checks.
- Loosens `assert.ok(requestCount <= 3)` to `<= 4` and rewords the explanatory comment.

### Motivation

Two distinct bugs combined to make the test flaky on all platforms:

1. The tracer's `NamespaceManager.send` (`packages/dd-trace/src/telemetry/metrics.js`) emits one `generate-metrics` payload **per namespace** (`tracers`, `profilers`, `appsec`, …). Because `assertTelemetryReceived` filtered only by `request_type`, the tracer-namespace message hit the test handler first and tripped the inner namespace assertion. Its error got pushed into the collected `errors[]` and surfaced as confusing `'tracers' !== 'profilers'` noise whenever the assertion ultimately timed out.

2. The `requestCount <= 3` upper bound was off-by-one: with `DD_PROFILING_UPLOAD_PERIOD=1` and `TEST_DURATION_MS=2500`, the periodic-vs-shutdown race the existing comment already acknowledges can produce 4 uploads, not just 3. When that happened, the legitimate `profilers` message also failed and the test timed out.

Filtering by namespace eliminates the noise entirely, and bumping the bound covers the actual race.

### Additional Notes

The remaining 7 of 39 Profiling flakes in the same report were Windows-only worker crashes (`exit=3221226505` / `STATUS_STACK_BUFFER_OVERRUN`) in `packages\dd-trace\test\profiling\config.spec.js` and one in `exporters\agent.spec.js`. All visible tests pass before the fast-fail, so it looks like a teardown issue in a native binding (likely `@datadog/pprof` / `@datadog/libdatadog`). That's a separate investigation and out of scope here.

Verified locally: the `Profiler telemetry` tests pass after the change, and `integration-tests/telemetry.spec.js` continues to pass (helper change is backward-compatible).

Jira: [PROF-14378]

[PROF-14378]: https://datadoghq.atlassian.net/browse/PROF-14378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


integration-tests/helpers/fake-agent.js
```diff
@@ -225,11 +225,14 @@ module.exports = class FakeAgent extends EventEmitter {
...
    * @param {number} [timeout=30_000] - The timeout in milliseconds.
    * @param {number} [expectedMessageCount=1] - The number of messages to expect.
+   * @param {boolean} [resolveAtFirstSuccess=false] - Resolve as soon as `fn` first runs without throwing.
+   * @param {string} [namespace] - If set, only consider messages whose payload namespace equals this value.
    * @returns {Promise<void>} A promise that resolves when the telemetry message of type `requestType` is received and
    *     the function `fn` has finished running. If `fn` throws an error, the promise will be rejected once `timeout`
    *     is reached.
    */
-  assertTelemetryReceived (fn, requestType, timeout = 30_000, expectedMessageCount = 1, resolveAtFirstSuccess = false) {
+  assertTelemetryReceived (fn, requestType, timeout = 30_000, expectedMessageCount = 1, resolveAtFirstSuccess = false,
```
BridgeAR: Just a thought: what about overloading `requestType` to become `filters` where filters is an object of things that must be defined on the payload?

For profiler it would look like:

```js
{
  request_type: 'generate-metrics',
  payload.namespace: 'profilers'
}
```

That way we do not have to add another argument, since six argument of which four are optional is quite challenging to use right.

IlyasShabi: I think we should rework assertTelemetryReceived params.. IMO this PR can land and we can refacto the options on another one

IlyasShabi (dismissed):
LGTM if @BridgeAR is okay with working on assertTelemetryReceived params separately

integration-tests/helpers/fake-agent.js
szegedi: I'll meet you halfway and replace the params with a param object, but I won't go the whole `filter` direction

BridgeAR (approved):

IlyasShabi (approved):

PR #8115, opened 2026-04-27
[test optimization] Update extraction of runner diagnostic dir using glob pattern matching

cbasitodx:
### What does this PR do?                                                                                                                 
                                                                                                                            
Updates the extraction of the GitHub Actions runner diagnostic directory by replacing hardcoded path patterns with glob pattern matching (using the glob package). The logic now derives candidate paths from the `RUNNER_TEMP` environment variable and expands them with bounded-depth glob patterns, accommodating runner layouts where diagnostic files are nested under versioned cache directories (e.g. actions-runner/cached/2.334.0/_diag/). Test fixtures were relocated accordingly to reflect the new directory structure.         
                                                                                                              
### Motivation                                                                                                                            
                                                                                               
Some GitHub Actions runner versions changed the layout of the runner installation directory, moving _diag/ files under a versioned cached/<version>/ subdirectory. The previous hardcoded path detection failed to find diagnostic files on updated runners, breaking CI visibility job ID extraction. Using glob patterns makes the detection resilient to future runner version changes without requiring code updates.


juan-fernandez:
@codex review

packages/dd-trace/src/plugins/util/ci.js
```diff
@@ -103,43 +103,138 @@ function getGitHubEventPayload () {
...
+
+  const runnerTemp = getValueFromEnvSources('RUNNER_TEMP')
+  if (runnerTemp) {
+    // RUNNER_TEMP is typically: <runnerRoot>/_work/_temp
+    const runnerRoot = path.resolve(runnerTemp, '..', '..')
+    // Bounded-depth patterns cover every runner layout we've observed
+    // (including cached/<version>/_diag) without assuming a `cached` wrapper
+    // and without walking the whole tree.
+    dirs.push(
+      `${runnerRoot}/*/_diag`, `${runnerRoot}/*/*/_diag`,
```
juan-fernandez: Mixed separators: template-literal patterns force `/` while `path.join` returns OS-native (`\` on Windows). On Windows `runnerRoot` will be `\`-separated, so you end up with `C:\x\foo/*/_diag` strings. The split regex `/[/\\]/` tolerates it, but it would be cleaner to commit to posix-style throughout (you already use `path.posix.join` later).

packages/dd-trace/src/plugins/util/ci.js
```diff
@@ -103,43 +103,138 @@ function getGitHubEventPayload () {
...
+  return str.includes('*') || str.includes('?')
+}
 
-  // RUNNER_TEMP usually looks like:
-  // Linux/mac hosted:   /home/runner/work/_temp
-  // Windows hosted:     C:\actions-runner\_work\_temp
-  // Self-hosted (unix): /opt/actions-runner/_work/_temp
+// Expands a glob pattern with only `*`/`?` at path-segment boundaries (no `**`)
+// into matching concrete paths using readdirSync — no external dependency needed.
+function expandGlobPattern (pattern) {
```
juan-fernandez, line 144: There is no direct unit test for `expandGlobPattern`. It's now load-bearing logic with a non-obvious contract (only `*`/`?` at segment boundaries — no `**`, no `[abc]`, no escapes). Please add a small spec covering: single-`*` match, double-`*/*` match, missing/unreadable directory, no-magic passthrough, and an explicit assertion that `**` is *not* expanded (so future maintainers don't add a pattern that silently no-ops).

packages/dd-trace/src/plugins/util/ci.js
```diff
@@ -103,43 +103,138 @@ function getGitHubEventPayload () {
...
+    } else {
+      expanded.push(candidate)
+    }
+  }
+
+  return uniq(expanded)
+}
+
+const githubWellKnownDiagnosticDirsUnix = [
+  '/home/runner/actions-runner/_diag',
```
juan-fernandez, line 198: `actions-runner/_diag` is still in the candidate list, but the fixture move means no test exercises it anymore. Suggest adding a `runner_legacy/actions-runner/_diag/Worker_…log` fixture (and a row in `runnerTempPaths`) to lock in backwards compatibility — older self-hosted runners still ship the flat layout.

packages/dd-trace/src/plugins/util/ci.js
```diff
@@ -103,43 +103,138 @@ function getGitHubEventPayload () {
...
+// since v2.334.0) and hypothetical future layouts that follow the same shape
+// without a `cached` wrapper (e.g. <runnerRoot>/<version>/_diag). Depth is
+// bounded on purpose: `*` matches a single segment, so no filesystem walk.
+const githubWellKnownDiagnosticDirPatternsUnix = [
+  '/home/runner/actions-runner/*/_diag',
+  '/home/runner/actions-runner/*/*/_diag',
+]
+const githubWellKnownDiagnosticDirPatternsWin = ['C:/actions-runner/*/_diag', 'C:/actions-runner/*/*/_diag']
+
+const githubJodIDRegex = /"job":\s*{[\s\S]*?"v"\s*:\s*(\d+)(?:\.0)?/
```
juan-fernandez: Typo: `githubJodIDRegex` → `githubJobIDRegex`. Same typo exists in datadog-ci#2291 — easy to fix here while it's only the one call site.

packages/dd-trace/src/plugins/util/ci.js
```diff
@@ -103,43 +103,138 @@ function getGitHubEventPayload () {
...
+
+function getJobIDFromDiagFile () {
+  const runnerTemp = getValueFromEnvSources('RUNNER_TEMP')
```
juan-fernandez, line 220: `RUNNER_TEMP` is read twice via `getValueFromEnvSources` — once here and once inside `getGithubDiagnosticDirsFromEnv`. Consider passing the value in (or have `getJobIDFromDiagFile` read it once and forward it).

packages/dd-trace/src/plugins/util/ci.js
```diff
@@ -103,43 +103,138 @@ function getGitHubEventPayload () {
...
 
   const isWin = process.platform === 'win32'
 
-  // Hardcoded fallbacks
+  let possibleDiagsPaths = expandDiagnosticDirCandidates([
+    ...getGithubDiagnosticDirsFromEnv(),
+    ...githubWellKnownDiagnosticDirPatternsUnix,
+    ...githubWellKnownDiagnosticDirsUnix,
+  ])
   if (isWin) {
```
juan-fernandez: This branch fully overwrites `possibleDiagsPaths` — the Unix list above is computed and discarded on Windows. Cleaner and avoids redundant work:

```js
const patterns = isWin ? githubWellKnownDiagnosticDirPatternsWin : githubWellKnownDiagnosticDirPatternsUnix
const literals = isWin ? githubWellKnownDiagnosticDirsWin    : githubWellKnownDiagnosticDirsUnix
const possibleDiagsPaths = expandDiagnosticDirCandidates([
  ...getGithubDiagnosticDirsFromEnv(), ...patterns, ...literals,
])
```

Also noticed the Windows path is uncovered by tests — would be good to add at least one Windows-pattern case alongside the cleanup.

packages/dd-trace/test/plugins/util/test-environment.spec.js
```diff
@@ -126,7 +126,23 @@ describe('test getJobIDFromDiagFile function', () => {
...
-      assert.strictEqual(getJobIDFromDiagFile(runnerTemp), expected)
+      const original = process.env.RUNNER_TEMP
+      try {
+        // getJobIDFromDiagFile reads RUNNER_TEMP from the environment
+        if (runnerTemp == null) {
+          delete process.env.RUNNER_TEMP
+        } else {
+          process.env.RUNNER_TEMP = runnerTemp
+        }
+        assert.strictEqual(getJobIDFromDiagFile(), expected)
```
juan-fernandez: Could simplify by stashing `RUNNER_TEMP` once in `before`/`after` hooks for the `describe` instead of try/finally per row — but more importantly, this test only exercises the new `cached/2.334.0/_diag` layout. Consider parameterizing across (a) the new nested layout, (b) the legacy `actions-runner/_diag` layout, and (c) the env-derived `${runnerRoot}/*/_diag` path so each candidate emitted by the production code is actually verified.

juan-fernandez (commented):
Behaviorally aligned with the datadog-ci side (DataDog/datadog-ci#2291) — same constants, patterns, and comments. Inline notes below; biggest gap vs. the datadog-ci PR is test coverage of the bespoke glob engine and of legacy layouts. None of these are blockers individually, but I'd push for the typo fix, the `isWin` cleanup, an `expandGlobPattern` unit test, and a legacy-layout fixture before merge.

juan-fernandez (approved):

PR #8116, opened 2026-04-27
fix(profiling): keep real proxy when only profiling is enabled

szegedi:
## Summary

When `DD_TRACE_ENABLED=false` (or its alias `DD_TRACING_ENABLED=0`), `packages/dd-trace/src/index.js` exports `NoopProxy`. `NoopProxy.init()` is a no-op, so the profiler never starts even though the user explicitly opted into profiling. This breaks the "profiling-only, no tracing" configuration.

There is already an allowlist (`shouldUseProxyWhenTracingDisabled`) for `DD_DYNAMIC_INSTRUMENTATION_ENABLED` and `DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED` that keeps the real proxy alive when tracing is off. This PR adds `DD_PROFILING_ENABLED` to that allowlist.

`DD_PROFILING_ENABLED` is declared with `"allowed": "false|true|auto|1|0"`, so the check recognizes `true`/`1`/`auto` as values that should keep the proxy alive (matching how `proxy.js` already handles those values to start the profiler or arm the SSI heuristic).

## Background

This was discovered while investigating a CI regression in [DataDog/prof-correctness](https://github.com/DataDog/prof-correctness/pull/115), where node profiling scenarios stopped producing pprof files after #6982 began resolving aliases in `getValueFromEnvSources`. The companion prof-correctness PR currently works around this by setting `DD_DYNAMIC_INSTRUMENTATION_ENABLED=true`; once this PR ships, that workaround can be removed.

## Test plan

- [x] Smoke test all combinations locally:
  - `DD_TRACE_ENABLED=false` alone → `NoopProxy`
  - `DD_TRACE_ENABLED=false` + `DD_PROFILING_ENABLED=true|1|auto` → real `Tracer`
  - `DD_TRACE_ENABLED=false` + `DD_PROFILING_ENABLED=false` → `NoopProxy`
  - `DD_TRACING_ENABLED=0` alone → `NoopProxy` (alias)
  - `DD_TRACING_ENABLED=0` + `DD_PROFILING_ENABLED=1` → real `Tracer`
- [x] `eslint packages/dd-trace/src/index.js` passes
- [ ] CI green

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -254,6 +254,7 @@
 /packages/dd-trace/test/histogram.spec.js @DataDog/lang-platform-js
 /packages/dd-trace/test/id.spec.js @DataDog/lang-platform-js
 /packages/dd-trace/test/iitm.spec.js @DataDog/lang-platform-js
+/packages/dd-trace/test/index.spec.js @DataDog/lang-platform-js
 /packages/dd-trace/test/log.spec.js @DataDog/lang-platform-js
 /packages/dd-trace/test/msgpack/ @DataDog/lang-platform-js
 /packages/dd-trace/test/noop.spec.js @DataDog/lang-platform-js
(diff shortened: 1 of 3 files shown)


BridgeAR (commented):
Shall we add a test to make sure this is working even with a refactoring? :)

BridgeAR:
I think this is now obsolete with the other PR reverting to the former behavior. Closing

PR #8117, opened 2026-04-27
remove empty after blocks in cypress tests

rochdev:
### What does this PR do?

Remove empty after blocks in Cypress tests.

### Motivation

Not clear why it was there to begin with.


BridgeAR (approved):

BridgeAR:
I would love a linter for these, while there is none for mocha hooks. We could use one for empty methods though. It might just report a lot of noop methods.

juan-fernandez (approved):

PR #8119, opened 2026-04-27
fix(llmobs, anthropic): do not mutate anthropic create parameters

### LINKED ISSUE #8105: [BUG]: LLM Obs anthropic plugin mutates caller's options.messages array
### Tracer Version(s)

5.98.0 (also reproducible on 6.0.0-pre dev tip — same code path)

### Node.js Version(s)

24.12.0

### Bug Report

The LLM Obs plugin for `@anthropic-ai/sdk` mutates the caller's `options.messages` array when tagging the input on a span. For one-shot `messages.create` callers this is invisible, but anything that reuses the array (the SDK's `BetaToolRunner`, hand-rolled loops, retries, multi-turn conversations) sends a `{role: 'system'}` element on the next request and gets a 400 from Anthropic:

```
BadRequestError: 400 messages: Unexpected role "system". The Messages API accepts a top-level `system` parameter, not "system" as an input message role.
```

#### Root cause

`packages/dd-trace/src/llmobs/plugins/anthropic.js`:

```js
#tagAnthropicInputMessages (span, options) {
  const { system, messages } = options       // messages aliases options.messages
  const inputMessages = []                   // already a fresh array for the tagged output

  if (system) {
    messages.unshift({ content: system, role: 'system' })   // mutates the caller's array
  }

  for (const message of messages) {
    // ...build inputMessages from each message
  }

  this._tagger.tagLLMIO(span, inputMessages)
}
```

The `inputMessages` array already exists for the tagged output. The fix is to push the system pseudo-message into `inputMessages` directly, never touching `options.messages`:

```diff
   if (system) {
-    messages.unshift({ content: system, role: 'system' })
+    inputMessages.push({ content: system, role: 'system' })
   }
```

The loop body would then need to start from index 0 of the original `messages` (which it already does) and continue pushing into `inputMessages` (which it already does).

#### Real-world impact

`BetaToolRunner` `structuredClone`s `params.messages` once at construction (good — protects the caller), but reuses that internal clone across loop iterations and passes it to `client.beta.messages.create` on every iteration. The plugin mutates the runner's internal clone each call, so iteration 2's wire payload starts with `{role:'system'}` and Anthropic rejects.

Same hazard for any caller that holds a reference to `messages` across requests: manual loops, retries, SDK `client.messages.create` in a multi-turn agent.

### Reproduction Code

```js
// repro.mjs — Node 20+, ESM
import 'dd-trace/register.js';

tracer.init({
  service: 'anthropic-mut-repro',
  llmobs: { enabled: true, mlApp: 'repro', agentlessEnabled: true },
});
tracer.use('anthropic');

const { default: Anthropic } = await import('@anthropic-ai/sdk');
const client = new Anthropic({ apiKey: 'sk-ant-fake' });

const messages = [{ role: 'user', content: 'hi' }];
console.log('BEFORE:', JSON.stringify(messages));

try {
  await client.messages.create({
    model: 'claude-3-5-haiku-latest',
    max_tokens: 32,
    system: 'You are terse.',
    messages,
  });
} catch {
  // ignore — fake key, request fails, but the LLM Obs tagger already mutated
}

console.log('AFTER: ', JSON.stringify(messages));
```

Output:

```
BEFORE: [{"role":"user","content":"hi"}]
AFTER:  [{"content":"You are terse.","role":"system"},{"role":"user","content":"hi"}]
```

The mutation happens on every `messages.create` call when `system` is set, regardless of whether the API call succeeds.

### Error Logs

The mutation itself is silent. Downstream symptom in real callers (e.g. `BetaToolRunner` iteration 2):

```
BadRequestError: 400 messages: Unexpected role "system". The Messages API accepts a top-level `system` parameter, not "system" as an input message role.
    at AnthropicBedrock.makeStatusError (.../@anthropic-ai/sdk/src/core/error.ts:75:14)
    at AnthropicBedrock.makeRequest (.../@anthropic-ai/sdk/src/client.ts:752:24)
    at async BetaToolRunner.[Symbol.asyncIterator] (.../@anthropic-ai/sdk/src/lib/tools/BetaToolRunner.ts:212:13)
    at async BetaToolRunner.runUntilDone (.../@anthropic-ai/sdk/src/lib/tools/BetaToolRunner.ts:389:24)
```

### Tracer Config

```js
tracer.init({
  service: 'anthropic-mut-repro',
  llmobs: { enabled: true, mlApp: 'repro', agentlessEnabled: true },
});
tracer.use('anthropic');
```

`@anthropic-ai/sdk@0.91.1`. Reproduces with both direct API and `@anthropic-ai/bedrock-sdk` (the bedrock-sdk inherits the same `Messages.prototype.create`).

### Operating System

Darwin 24.6.0

### Bundling

No Bundling

sabrenner:
### What does this PR do?

Ensures we do not modify the parameters of the anthropic call by removing `messages.unshift`, which prepended a system message in that branch. Adds a regression test.

### Motivation

Fixes https://github.com/DataDog/dd-trace-js/issues/8105


packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
```diff
@@ -107,6 +107,29 @@ describe('Plugin', () => {
...
       })
+
+      it('does not modify the `system` property', async () => {
+        const params = {
+          model: 'claude-haiku-4-5-20251001',
+          messages: [{ role: 'user', content: 'Hello, world!' }],
+          max_tokens: 100,
+          temperature: 0.5,
+          system: 'talk like a pirate',
+        }
```
BridgeAR:
```suggestion
        const params = Object.freeze({
          model: 'claude-haiku-4-5-20251001',
          messages: Object.freeze([Object.freeze({ role: 'user', content: 'Hello, world!' })]),
          max_tokens: 100,
          temperature: 0.5,
          system: 'talk like a pirate',
        })
```

packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
```diff
@@ -107,6 +107,29 @@ describe('Plugin', () => {
...
+        assert.deepEqual(params.messages, [{ role: 'user', content: 'Hello, world!' }])
+        assert.ok(response)
+
+        const { llmobsSpans } = await getEvents()
+        assert.equal(llmobsSpans.length, 1)
+
+        assert.deepEqual(llmobsSpans[0].meta.input.messages, [
+          { role: 'system', content: 'talk like a pirate' },
+          { role: 'user', content: 'Hello, world!' },
+        ])
```
BridgeAR, line 131:
```suggestion
        await getEvents()
```

BridgeAR (commented):
Can we just freeze the object? That way we know if it would actually be touched.

Right now we just check if the structure is kept.

If something is now tried to be edited in-between, it would throw an error.

packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
sabrenner: good call, i can do that!

packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
sabrenner: sorry, can you clarify the removal here? i wanted to verify we still tag the appropriate messages on the llmobs events

packages/dd-trace/src/llmobs/plugins/anthropic.js
```diff
@@ -144,7 +144,7 @@ class AnthropicLLMObsPlugin extends LLMObsPlugin {
     const inputMessages = []
 
     if (system) {
-      messages.unshift({ content: system, role: 'system' })
+      inputMessages.push({ content: system, role: 'system' })
```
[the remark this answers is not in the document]
sabrenner: this is a good point actually, let me push up a fix for this as well.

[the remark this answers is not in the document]
sabrenner: fixed & verified against system tests locally as well (they are running in CI but need to be re-enabled)

packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
```diff
@@ -107,6 +107,29 @@ describe('Plugin', () => {
           assertLLMObsSpan(apmSpans, llmobsSpans)
         })
       })
+
+      it('does not modify the `system` property', async () => {
+        const params = Object.freeze({
+          model: 'claude-haiku-4-5-20251001',
+          messages: Object.freeze([{ role: 'user', content: 'Hello, world!' }]),
```
BridgeAR:
```suggestion
          messages: Object.freeze([Object.freeze({ role: 'user', content: 'Hello, world!' })]),
```

packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
BridgeAR: It is fine to keep that around. It was more a: if we manipulate entries, we would cause an error, so it seemed an independent test case. I am good with keeping it.

packages/dd-trace/src/llmobs/plugins/anthropic/index.js
```diff
@@ -144,51 +145,15 @@ class AnthropicLLMObsPlugin extends LLMObsPlugin {
     const inputMessages = []
 
     if (system) {
-      messages.unshift({ content: system, role: 'system' })
+      const formattedSystemMessage = formatMessage('system', system)
+      inputMessages.push(...formattedSystemMessage)
```
BridgeAR: While it is normally not a good idea to manipulate objects from a top level, it would e more performant, to pass through the array to `formatMessage` to add the entries there. I would just change the name in that case.

The main benefit is also that this allows for more than 65k entries being pushed (I am not sure about the concrete limit anymore but that is roughly the arguments upper bound).

The same about the other spot.

sabrenner: mm yeah i thought about that but also thought it might break conventional style. although yeah it would create one less array per call - i can push up the change!

sabrenner: updated, let me know what you think 😄

BridgeAR (approved):

PR #8124, opened 2026-04-27
test: reduce agent.load() overhead and guard against latency regressions

wconti27:
## Summary

- Skip `proxyquire.noPreserveCache()` on the **first** `agent.load()` call: when modules aren't yet cached a plain `require('../..')` produces fresh instances without the proxyquire overhead. Subsequent calls still go through `proxyquire.noPreserveCache()` for isolation between test suites.
- Drop the `checkAgentStatus` socket timeout from 2000 ms → 500 ms. A live agent responds in < 10 ms; 500 ms is plenty to distinguish "no agent" from a slow response and halves the minimum latency of every `agent.load()` call.
- Drain the HTTP response body in `checkAgentStatus` (`response.resume()`) so the socket closes immediately instead of waiting for the inactivity timer.
- Add `agent.spec.js` with a timing assertion: `agent.load()` must complete in < 2500 ms (50 % of the default mocha timeout), catching any regression back to the 5–8 s proxyquire path.

Related: #8099

## Test plan

- [ ] `./node_modules/.bin/mocha packages/dd-trace/test/plugins/agent.spec.js` passes locally (160 ms)
- [ ] Plugin test suite passes (`PLUGINS=http npm run test:plugins`)

### DIFF
--- a/.github/workflows/apm-integrations.yml
+++ b/.github/workflows/apm-integrations.yml
@@ -20,6 +20,7 @@ concurrency:
 env:
   MOCHA_OPTIONS: ${{ github.ref == 'refs/heads/master' && '--retries 1' || '' }}
   LATEST_VERSION: ${{ inputs.latest-version }}
+  DD_REMOTE_CONFIGURATION_ENABLED: false
 
 # TODO: upstream jobs
 
--- a/packages/dd-trace/test/plugins/agent.js
+++ b/packages/dd-trace/test/plugins/agent.js
@@ -28,6 +28,7 @@ let plugins = []
 const testedPlugins = []
 let dsmStats = []
 let currentIntegrationName = null
+let loaded = false
 
 function isMatchingTrace (spans, spanResourceMatch) {
   if (!spanResourceMatch) {
@@ -229,9 +230,12 @@ function handleTraceRequest (req, res, sendToTestAgent) {
 function checkAgentStatus () {
   return new Promise((resolve) => {
     const agentUrl = process.env.DD_TRACE_AGENT_URL || 'http://127.0.0.1:9126'
-    const timeoutMs = 2000
+    const timeoutMs = 500
+    let done = false
 
     const request = http.request(`${agentUrl}/info`, { method: 'GET', timeout: timeoutMs }, response => {
+      response.resume() // drain body so the socket is released before the idle timer re-fires
+      done = true
       resolve(response.statusCode === 200)
     })
 
@@ -241,12 +245,13 @@ function checkAgentStatus () {
         `checkAgentStatus: Timed out after ${timeoutMs}ms trying to reach test agent at ${agentUrl}. ` +
         'Proceeding without test agent. If this happens frequently, investigate what is listening on that port.'
       )
+      done = true
       request.destroy()
       resolve(false)
     })
 
     request.on('error', (/** @type {NodeJS.ErrnoException} */ err) => {
-      if (err.code !== 'ECONNREFUSED') {
+      if (!done && err.code !== 'ECONNREFUSED') {
         // eslint-disable-next-line no-console
         console.warn(`checkAgentStatus: Unexpected error reaching test agent at ${agentUrl}`, err)
       }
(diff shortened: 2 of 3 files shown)


packages/dd-trace/test/plugins/agent.js
```diff
@@ -424,22 +426,27 @@ module.exports = {
...
+      })
+      const TracerProxy = proxyquire('../../src', {
+        './proxy': proxy,
+      })
+      tracer = proxyquire('../../', {
+        './src': TracerProxy,
+      })
+    } else {
+      tracer = require('../..')
+      loaded = true
```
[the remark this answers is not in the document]
wconti27: Fixed with a check on the cache.

.github/workflows/apm-integrations.yml
```diff
@@ -20,6 +20,7 @@ concurrency:
 env:
   MOCHA_OPTIONS: ${{ github.ref == 'refs/heads/master' && '--retries 1' || '' }}
   LATEST_VERSION: ${{ inputs.latest-version }}
+  DD_REMOTE_CONFIGURATION_ENABLED: false
```
BridgeAR, line 23: Should we really deactivate this globally? I think we normally do not use that anyway in the tests?

packages/dd-trace/test/plugins/agent.js
```diff
@@ -424,22 +435,27 @@ module.exports = {
...
-      './config': getConfigFresh,
-      './dogstatsd': dogstatsd,
-    })
-    const TracerProxy = proxyquire('../../src', {
-      './proxy': proxy,
-    })
-    tracer = proxyquire('../../', {
-      './src': TracerProxy,
-    })
+    if (loaded || require.cache[require.resolve('../..')]) {
```
BridgeAR, line 438: I believe we still have some legit failures. Sometimes we might have a weird test state and the cache could still be empty.

packages/dd-trace/test/plugins/agent.js
```diff
@@ -229,9 +236,12 @@ function handleTraceRequest (req, res, sendToTestAgent) {
 function checkAgentStatus () {
   return new Promise((resolve) => {
     const agentUrl = process.env.DD_TRACE_AGENT_URL || 'http://127.0.0.1:9126'
-    const timeoutMs = 2000
+    const timeoutMs = 500
```
BridgeAR, line 239: Suggestion: let's keep the checkAgentStatus alone, I am going to remove that in the other PR anyway.

packages/dd-trace/test/plugins/agent.js
```diff
@@ -14,6 +14,12 @@ const semifies = require('semifies')
 const { assertObjectContains } = require('../../../../integration-tests/helpers')
 const { storage } = require('../../../datadog-core')
 const ritm = require('../../src/ritm')
+
+// Pre-warm modules that are cold-loaded inside tracer.init() so they land in
+// require.cache before any mocha before-all hook runs (and its timeout ticks).
+try { require('../../src/dogstatsd') } catch (_) {}
+try { require('../../src/exporters/agent') } catch (_) {}
+try { require('../../src/appsec/iast/taint-tracking/rewriter') } catch (_) {}
```
BridgeAR, line 22: I understand the why, I think it should not matter all that much though. This will not be a lasting solution overall 😅

BridgeAR (commented):
The cache check is looking good! I believe we still have some related failures though.

rochdev (changes requested):
> Skip proxyquire.noPreserveCache() on the first agent.load() call: when modules aren't yet cached a plain require('../..') produces fresh instances without the proxyquire overhead. Subsequent calls still go through proxyquire.noPreserveCache() for isolation between test suites.

Proxyquire shouldn't add any latency, if anything it should reduce it since it replaces a module tree with a stub. So I'm not sure the added complexity is justified. Do you have any numbers?

> Drop the checkAgentStatus socket timeout from 2000 ms → 500 ms. A live agent responds in < 10 ms; 500 ms is plenty to distinguish "no agent" from a slow response and halves the minimum latency of every agent.load() call.

Could that cause issues if there is any hiccup on the machine? Why was it 2s to begin with?

> Drain the HTTP response body in checkAgentStatus (response.resume()) so the socket closes immediately instead of waiting for the inactivity timer.

That change seems fine, but is it needed? What is it fixing?

> Add agent.spec.js with a timing assertion: agent.load() must complete in < 2500 ms (50 % of the default mocha timeout), catching any regression back to the 5–8 s proxyquire path.

The additional test doesn't quite make sense. It depends on what features are enabled etc. This is not something we can really measure in that way. If anything we should be adding a benchmark instead.

PR #8130, opened 2026-04-28
warmup the nyc cache before running the tests

rochdev:
### What does this PR do?

Warmup the nyc cache before running the tests.

### Motivation

When running our tests, we defer calling `tracer.init()` until the first call to `agent.load()`. This means that the cost of instrumenting all of the project files for code coverage is deferred until that moment, which can be quite heavy without an existing nyc cache. This causes the first `beforeEach` to call `agent.load()` to sometimes take more than 5 seconds and ultimately timeout. By warming up the cache before running the tests, we avoid this issue entirely. Not all files are included in the warmup to keep it fast.

### Additional Notes

Before the change, running the Express test locally without any cache made `agent.load()` take 2.3s compared to 230ms with the cache. With the partial warmup, it now takes ~400ms instead. Not as good as a full warmup, but good enough and takes way less time to generate.


pabloerhard (commented):
LGTM, but wouldn’t this hide the issue that our init is now occasionally taking longer than 5s and therefore failing? Is this expected as the codebase grows?

BridgeAR (approved):
LGTM, this is a good idea and will reduce the main flakiness!

I wonder if it would make sense to move this inside of the workflows at a later point.

BridgeAR:
@pabloerhard the init just took longer due to the NYC instrumentation. Without that, it will be faster.

We can theoretically reduce the time budget from 5 seconds to a smaller number, if we make sure all our setups are fast enough.

PR #8131, opened 2026-04-28
use mocha parallel instead of a matrix for cypress and playwright

rochdev:
### What does this PR do?

Use Mocha parallel instead of a matrix for Cypress and Playwright.

### Motivation

Hopefully use less machines and parallelize on cores instead.


BridgeAR:
@rochdev should this stay open?

rochdev:
> @rochdev should this stay open?

This ended up not working because of how we do coverage, but now coverage changed so I'm not sure. I guess let's close it for now and we can try this again later.

PR #8132, opened 2026-04-28
PoC: Changes to add olly and caching to flaky test runs

tabgok:
## Summary
- `useSandbox()`'s `after()` hook crashed with `TypeError: Cannot read properties of undefined (reading 'remove')` whenever its `before()` failed, masking the real failure as a generic `exit code 7`.
- Same shape in `packages/datadog-plugin-aerospike/test/index.spec.js` (`releaseEventLoop` of undefined).
- Both teardown hooks now bail when the resource was never set up so the original error stays visible.

## Why this is a draft
This PR is the basis for investigating the APM Integrations flake cluster. The 7-day flakiness scan showed that ~25 of the 59 flaky APM Integrations jobs (plus several Serverless, Platform, and LLMObs jobs) all surface as the same masked `TypeError`. With the mask removed, retries should expose the **real** underlying causes (missing services, install timeouts, etc.) and let us triage them individually.

## Investigation plan
1. Merge once green.
2. Watch the next run of the apm-integrations workflow on master + retries on PRs.
3. Triage the now-visible underlying failures one-by-one in follow-up branches under `teague.bick/<test-slug>`.

## Test plan
- [ ] CI green on this branch
- [ ] Trigger a `before()` failure in any sandbox-using integration test locally and verify the error message is the original one (not the masked TypeError)
- [ ] Re-run flakiness report after merge to confirm the `exit code 7` cluster shrinks

### DIFF
--- a/integration-tests/helpers/index.js
+++ b/integration-tests/helpers/index.js
@@ -943,6 +943,7 @@ function useSandbox (...args) {
 
   after(function () {
     this.timeout(30_000)
+    if (!sandbox) return
     return sandbox.remove()
   })
 }
--- a/packages/datadog-plugin-aerospike/test/index.spec.js
+++ b/packages/datadog-plugin-aerospike/test/index.spec.js
@@ -51,6 +51,7 @@ describe('Plugin', () => {
         })
 
         after(() => {
+          if (!aerospike) return
           aerospike.releaseEventLoop()
         })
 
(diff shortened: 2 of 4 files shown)


BridgeAR (commented):
Note: these are nice to have changes. They only trigger in case the test itself already fails

PR #8135, opened 2026-04-28
fix: telemetry for test optimization auto instrument provider

BridgeAR:
Also add test cases for the worker environment variables to work as expected. This just increases coverage. The string is working as a simple existence check.


packages/dd-trace/src/config/supported-configurations.json
```diff
@@ -446,9 +446,10 @@
     ],
     "DD_CIVISIBILITY_AUTO_INSTRUMENTATION_PROVIDER": [
       {
-        "implementation": "B",
-        "type": "boolean",
-        "default": "false"
+        "implementation": "A",
```
juan-fernandez, line 449: question: is it possible to go back in the implementation?

packages/dd-trace/test/config/index.spec.js
```diff
@@ -3040,6 +3040,16 @@ describe('Config', () => {
...
       })
     })
+
+    it('should only accept allowed values for DD_CIVISIBILITY_AUTO_INSTRUMENTATION_PROVIDER', () => {
+      for (const provider of ['github', 'gitlab', 'circleci']) {
+        process.env.DD_CIVISIBILITY_AUTO_INSTRUMENTATION_PROVIDER = provider
+        assert.strictEqual(getConfig(options).DD_CIVISIBILITY_AUTO_INSTRUMENTATION_PROVIDER, provider)
+      }
+      process.env.DD_CIVISIBILITY_AUTO_INSTRUMENTATION_PROVIDER = 'jenkins'
+      assert.strictEqual(getConfig(options).DD_CIVISIBILITY_AUTO_INSTRUMENTATION_PROVIDER, undefined)
```
juan-fernandez: just to confirm: setting an unsupported name in this env var will _not_ crash the customer test process, right?

packages/dd-trace/src/config/supported-configurations.json
```diff
@@ -446,9 +446,10 @@
...
+        "type": "string",
+        "default": null,
+        "allowed": "github|gitlab|circleci"
```
juan-fernandez: could we allow any? If we add support for new CI providers, we'll need customers to bump `dd-trace` to get proper telemetry

juan-fernandez (commented):
LGTM. I'm not sure if we gain much from having an allowlist vs allowing any string and we do lose a bit (not much but not 0)

packages/dd-trace/src/config/supported-configurations.json
BridgeAR: Yes, that is totally fine. The implementation just defines a "behavior".

packages/dd-trace/test/config/index.spec.js
BridgeAR: Correct. It is just plainly ignored, if it does not contain a boolean value.

packages/dd-trace/src/config/supported-configurations.json
BridgeAR: Yes, that is totally fine. I just removed that again

juan-fernandez (dismissed):

juan-fernandez (approved):

PR #8140, opened 2026-04-28
fix: make DD_TRACING_ENABLED an alias of DD_APM_TRACING_ENABLED

BridgeAR:
It was accidentially made an alias of DD_TRACE_ENABLED, which acts differently.


szegedi (dismissed):

watson:
@codex review

packages/dd-trace/src/index.js
```diff
@@ -1,18 +1,15 @@
...
 
 const ddTraceDisabled = getValueFromEnvSources('DD_TRACE_ENABLED')
   ? isFalse(getValueFromEnvSources('DD_TRACE_ENABLED'))
   : String(getValueFromEnvSources('OTEL_TRACES_EXPORTER')).toLowerCase() === 'none'
-const shouldUseProxyWhenTracingDisabled =
-  isTrue(getValueFromEnvSources('DD_DYNAMIC_INSTRUMENTATION_ENABLED')) ||
-  isTrue(getValueFromEnvSources('DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED'))
 
-module.exports = (ddTraceDisabled && !shouldUseProxyWhenTracingDisabled) || inJestWorker
+module.exports = ddTraceDisabled || inJestWorker
```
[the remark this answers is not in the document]
watson: AI is right about the problem, but got the solution wrong: I think the real answer here is to update the tests to not expect debugger snapshots when `DD_TRACE_ENABLED` is false. Or if it makes more sense, simply remove those tests.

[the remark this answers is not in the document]
BridgeAR: Agreed, I removed the test

szegedi (approved):

integration-tests/debugger/tracing-integration.spec.js
```diff
@@ -67,18 +52,15 @@ describe('Dynamic Instrumentation', function () {
...
       env: { DD_TRACING_ENABLED: 'false' },
       dependencies: ['fastify'],
     })
 
     describe('input messages', function () {
-      it(
-        'should capture and send expected payload when a log line probe is triggered',
-        testBasicInputWithoutDD.bind(null, t)
-      )
+      it('should capture and send expected payload when a log line probe is triggered', testBasicInput.bind(null, t))
```
watson, line 63: I'm not sure we want to remove `testBasicInputWithoutDD`. I would assume that's still relevant here?

BridgeAR: In standalone APM mode (`DD_APM_TRACING_ENABLED=false` / `DD_TRACING_ENABLED=false`) the tracer is fully active locally, it just stops shipping spans the agent does not need.

The active span returned by `tracer.scope().active()?.context()` is still a real span with a `trace_id` and `span_id`, so the `dd` field IS populated in the debugger payload. That's why the new test correctly uses `testBasicInput` (which asserts the `dd` field is present and equal to the local span's ids) instead of `testBasicInputWithoutDD`.

The only scenario where the helper would be relevant is `DD_TRACE_ENABLED=false`, but after this PR that flag unconditionally short-circuits to the noop proxy, which means DI is never started in the test app and the breakpoint would never fire — so the helper had no remaining caller and I removed it.

I will open a follow-up that adds an explicit assertion that DI does not start when `DD_TRACE_ENABLED=false`, so the new behavior is locked down by a test.

watson: I thought that when `DD_TRACING_ENABLED=false` then there would be no trace/span created?

BridgeAR: The spans are still created, they are just not send out later. Span creation is only turned off with `DD_TRACE_ENABLED` being false.

PR #8143, opened 2026-04-28
chore(node) allow node 26

pabloerhard:
### What does this PR do?

Allow test to run with node 26


BridgeAR (approved):
LGTM. I think we should also change our CI jobs. I am actually uncertain how we will do this when we have v6 released. We pretty much need to use different test versions in our CI

PR #8144, opened 2026-04-28
test: fix max listener warning in test optimization tests

BridgeAR:
The warning would show up, while it was ignored. Now, we increase the limit for mocha so that this is not triggered anymore and does not confuse about what is really going on.

### DIFF
--- a/packages/dd-trace/test/setup/core.js
+++ b/packages/dd-trace/test/setup/core.js
@@ -65,7 +65,7 @@ temporaryWarningExceptions.add = (warning) => {
 }
 
 process.on('warning', (warning) => {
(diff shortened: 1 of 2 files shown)


pabloerhard (commented):
I'm confused by the objective of this PR. Is the objective to now throw when the warning includes Runner but with a higher threshold?

BridgeAR:
> I'm confused by the objective of this PR. Is the objective to now throw when the warning includes Runner but with a higher threshold?

In a way, yes. Right now, we always ignore these warnings, they are however reached and triggered and printed in the CI. So we get the warnings, they are just ignored. With this change, we do not receive the warnings anymore and if we breach the higher limit which is needed for test optimization to work properly, we get the actual failure.

pabloerhard (approved):

PR #8146, opened 2026-04-28
feat(llmobs, anthropic): add reasoning/extended thinking support for Anthropic 

sabrenner:
### What does this PR do?

Adds support for properly parsing reasoning and extended thinking blocks from Anthropic inputs and outputs.

### Motivation

Make sure out anthropic integration is up-to-date with latest features.
Fixes https://github.com/DataDog/dd-trace-js/issues/7890

### Additional Notes
About 250 LOC additions are cassette/fixture changes


BridgeAR (dismissed):
LGTM as far as I can tell by reading the code

packages/dd-trace/src/llmobs/plugins/anthropic/index.js
```diff
@@ -56,20 +58,29 @@ class AnthropicLLMObsPlugin extends LLMObsPlugin {
...
+              const { thinking } = delta
+              if (thinking) lastBlock.thinking += thinking
+            } else if (delta.type === 'signature_delta') {
+              // Signature is for internal verification only; skip it.
+            } else if (delta.type === 'input_json_delta') {
+              const partialJson = delta.partial_json
+              if (partialJson) lastBlock.input += partialJson
+            } else {
+              const { text } = delta
+              if (text) lastBlock.text += text
```
Kyle-Verhoog, line 75: **Default `else` branch is too permissive.**

The new switch on `delta.type` falls through to a default branch that always tries `lastBlock.text += delta.text`. If Anthropic introduces (or has already introduced — `citations_delta` exists in the API spec for documents) a new delta type that happens to carry a `text`-ish field, content will be silently appended to whatever the last block is — including a thinking block, which has no `text` property. You'd end up with `lastBlock.text === 'undefinedfoo...'`.

The pre-PR code had the same hazard accidentally, but the new explicit-switch shape makes it look intentional.

**Suggested action:** make the default branch explicit:

```js
} else if (delta.type === 'text_delta') {
  const { text } = delta
  if (text) lastBlock.text += text
}
```

Anything else falls off and is ignored. This matches the pattern in dd-trace-py.

packages/dd-trace/src/llmobs/plugins/anthropic/index.js
```diff
@@ -47,6 +47,8 @@ class AnthropicLLMObsPlugin extends LLMObsPlugin {
             const { type } = contentBlock
             if (type === 'text') {
               response.content.push({ type, text: contentBlock.text })
+            } else if (type === 'thinking') {
+              response.content.push({ type, thinking: contentBlock.thinking ?? '' })
```
Kyle-Verhoog, line 52: **Missing handling of `redacted_thinking` content blocks.**

Anthropic returns `redacted_thinking` blocks (with an opaque `data` field, no `thinking` field) when safety filters trigger on the model's reasoning. With this PR:

- **Non-streaming output** (`#tagAnthropicOutputMessages`): a `redacted_thinking` block has no `text`, no `thinking`, isn't `tool_use` — silently dropped. User gets no signal that the model produced (filtered) reasoning at all.
- **Streaming**: `content_block_start` for `redacted_thinking` doesn't push a block here, so any subsequent deltas would target the wrong block. (Today there are no deltas for this block type, but it's an implicit assumption.)

This mirrors a pre-existing gap in dd-trace-py, so not a regression — but since you're touching exactly this code, it's a low-cost moment to also handle it.

**Suggested action:** add `else if (type === 'redacted_thinking') { response.content.push({ type, thinking: '([REDACTED THINKING])' }) }` (or push nothing here but skip the JSON-stringify fallthrough in `util.js` — see related comment on `util.js`). Mirrors the `([IMAGE DETECTED])` pattern already used for image blocks.

packages/dd-trace/src/llmobs/plugins/anthropic/util.js
```diff
@@ -70,6 +71,8 @@ function appendMessage (messages, { role, content }) {
       messages.push({ content: block.text, role })
     } else if (block.type === 'image') {
       messages.push({ content: '([IMAGE DETECTED])', role })
+    } else if (block.type === 'thinking') {
```
Kyle-Verhoog, line 74: **Companion to the `redacted_thinking` comment on `index.js`.**

For input messages, a `redacted_thinking` block falls through to the `else` branch at the bottom of this loop and gets `JSON.stringify`-ed into the message content — which dumps the entire opaque encrypted blob into the input record. That's noisy and a small data-leak surface.

**Suggested action:** add `} else if (block.type === 'redacted_thinking') { messages.push({ content: '([REDACTED THINKING])', role: 'reasoning' }) }` here, paralleling the new thinking branch.

packages/dd-trace/src/llmobs/plugins/anthropic/util.js
```diff
@@ -3,6 +3,7 @@
 /**
  * @typedef {{type: 'text', text: string}} TextBlock
  * @typedef {{type: 'image'}} ImageBlock
+ * @typedef {{type: 'thinking', thinking: string, signature?: string}} ThinkingBlock
  * @typedef {{
  *  type: 'tool_use', text: string, name: string, id: string, input: string | Record<string, unknown>
```
Kyle-Verhoog, line 8: **Nit (pre-existing, unrelated to this PR):** `tool_use` blocks don't have a `text` field per the Anthropic spec — the `text: string` in this typedef is misleading. Easy to fix while you're touching this file. Feel free to ignore if you'd rather keep the diff focused.

packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
```diff
@@ -182,6 +182,239 @@ describe('Plugin', () => {
...
+        const { apmSpans, llmobsSpans } = await getEvents()
+        assertLlmObsSpanEvent(llmobsSpans[0], {
+          span: apmSpans[0],
+          spanKind: 'llm',
+          name: 'anthropic.request',
+          modelName: 'claude-haiku-4-5-20251001',
+          modelProvider: 'anthropic',
+          inputMessages: [{ role: 'user', content: 'What is the best selling book of all time?' }],
+          outputMessages: [
+            { role: 'reasoning', content: MOCK_STRING },
```
Kyle-Verhoog, line 218: **Optional: assert the signature isn't leaking into the captured reasoning content.**

The `signature` field on thinking blocks is opaque/large (~2KB+) and intentionally excluded from the LLMObs output. There's no test that pins this exclusion — a future refactor could accidentally include `block.signature` in the reasoning content.

With both signed thinking and tool-use cassettes available, you have everything needed to assert this. One extra line, e.g.:

```js
assert.ok(!llmobsSpans[0].meta.output.messages[0].content.includes('signature'))
```

or a regex check that the reasoning content doesn't look like base64-of-protobuf. Optional, but cheap insurance.

packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
```diff
@@ -182,6 +182,239 @@ describe('Plugin', () => {
...
+        const { apmSpans, llmobsSpans } = await getEvents()
+        assertLlmObsSpanEvent(llmobsSpans[0], {
+          span: apmSpans[0],
+          spanKind: 'llm',
+          name: 'anthropic.request',
+          modelName: 'claude-haiku-4-5-20251001',
+          modelProvider: 'anthropic',
+          inputMessages: [
+            { role: 'user', content: WEATHER_PROMPT },
+            { role: 'reasoning', content: thinkingBlock.thinking },
```
Kyle-Verhoog, line 379: **Nit: this is the only assertion in the new tests that pins on cassette text rather than `MOCK_STRING`.**

If the cassette is ever regenerated with a different prompt-response, this test breaks for no good reason. The role + position is the meaningful part. Suggest:

```js
{ role: 'reasoning', content: MOCK_STRING },
```

for consistency with the other thinking assertions in this `describe` block.

Kyle-Verhoog (commented):
Nice incremental change — overall low risk, tests are readable, and the streaming-handler rewrite is a quiet correctness win (pre-PR, a thinking block in a stream would have silently corrupted the prior block). Already approved by BridgeAR; my comments below are non-blocking suggestions.

---

### Cross-SDK consistency: missing system tests for thinking/reasoning

**Risk:** medium · **Category:** test coverage / consistency · **Confidence:** high

Per the team's cross-language policy, any new LLMObs SDK feature should have a parametric/integration-framework system test in `~/dd/system-tests/` so all SDKs (Python, Node, Java, Go) emit consistent shape. `dd-trace-py` already supports thinking → `role: 'reasoning'`. With this PR, JS does too. But `system-tests/tests/integration_frameworks/llm/anthropic/test_anthropic_llmobs.py` has no `thinking`/`reasoning` cases — there's no enforcement that future contributors keep the role/order/content consistent across SDKs.

**Suggested action:** Add a parametrized test case that submits a request with `thinking.type: 'enabled'` and asserts the resulting LLMObs span has a `reasoning`-role message before the `assistant` one. Generate the corresponding cassette under both `python/` and `node/` weblog apps. This can be a follow-up PR rather than blocking this one — and it's a recurring miss across recent SDK feature PRs, not unique to this change. Worth a tracking ticket in MLOB-* if one doesn't exist.

packages/dd-trace/src/llmobs/plugins/anthropic/index.js
sabrenner: valid but afaict, this follows the behavior merged in python: https://github.com/DataDog/dd-trace-py/blob/1be83497d7142d094b869c5c9f454cefa7155c00/ddtrace/contrib/internal/anthropic/_streaming.py#L189-L192.

so i guess we should follow up in dd-trace-py as well. but i can make the change here

packages/dd-trace/src/llmobs/plugins/anthropic/index.js
sabrenner: same as the above, looks like we don't handle this in dd-trace-py either (in this [file](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/contrib/internal/anthropic/_streaming.py), or the whole repo, grep for `redacted_thinking` returns nothing).

i think this one i will leave out of this PR and we can follow up in both dd-trace-py, here, and system tests if that sounds good 👍

packages/dd-trace/src/llmobs/plugins/anthropic/util.js
sabrenner: i can update, although we have been using it as we use it in python as well: https://github.com/DataDog/dd-trace-py/blob/1be83497d7142d094b869c5c9f454cefa7155c00/ddtrace/llmobs/_integrations/anthropic.py#L148

i think to be safe, i'll keep it.

packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
sabrenner: i think this is a non-issue, if the cassette changes, the `thinkingBlock` content will change dynamically as well, as we grab this from the parsed responses from the anthropic SDK. If cassette content changes, so will the anthropic SDK response, and so will the 
```js
const thinkingBlock = firstResponse.content.find(b => b.type === 'thinking')
```
line

packages/dd-trace/test/llmobs/plugins/anthropic/index.spec.js
sabrenner: good point, i can add a small check for this

packages/dd-trace/src/llmobs/plugins/anthropic/util.js
sabrenner: similar to the above, maybe we leave this change for a follow-up? since we don't do anything for redacted thinking in dd-trace-py either afaict

packages/dd-trace/src/llmobs/plugins/anthropic/index.js
```diff
@@ -56,20 +58,29 @@ class AnthropicLLMObsPlugin extends LLMObsPlugin {
...
+
+            if (delta.type === 'thinking_delta') {
+              const { thinking } = delta
+              if (thinking) lastBlock.thinking += thinking
+            } else if (delta.type === 'signature_delta') {
+              // Signature is for internal verification only; skip it.
+            } else if (delta.type === 'input_json_delta') {
+              const partialJson = delta.partial_json
+              if (partialJson) lastBlock.input += partialJson
+            } else if (delta.type === 'text_delta') {
```
Yun-Kim: looks like we will miss any new delta types that get introduced in the future, can we do a safe else default branch to stringify and add the text (i think we do this in python)

packages/dd-trace/src/llmobs/plugins/anthropic/index.js
Yun-Kim: yup agreed python is missing redacted_thinking support too i think we can leave out of this PR as well

Yun-Kim (dismissed):
Small suggestion on default handling but otherwise LGTM!

packages/dd-trace/src/llmobs/plugins/anthropic/index.js
sabrenner: yeah i can revert to the previous state!

Yun-Kim (dismissed):

Yun-Kim (approved):

PR #8147, opened 2026-04-28
test: do not hide mocha errors

BridgeAR:
If a after* hook fails after a failure in a before* hook or the test, the after hook error is going to be read first and tears down the test runner. This causes the actual error not to be visible anymore.

This surfaces these again by hooking into mocha internals and ignoring the after* errors, if it comes after a former error. That way we know what actually fails.

Downside is not knowing about teardown issues that are in addition to former issues, while that is in almost all cases not important.

Hiding also seemed better than logging both to let us focus on the main problem area.

### DIFF
--- a/integration-tests/helpers/index.js
+++ b/integration-tests/helpers/index.js
@@ -944,8 +944,7 @@ function useSandbox (...args) {
     this.timeout(30_000)
     const oldSandbox = sandbox
     sandbox = undefined
-    if (!oldSandbox) return
-    return oldSandbox.remove()
+    return oldSandbox?.remove()
   })
 }
 
--- a/integration-tests/helpers/index.js
+++ b/integration-tests/helpers/index.js
@@ -942,7 +942,10 @@ function useSandbox (...args) {
 
   after(function () {
     this.timeout(30_000)
-    return sandbox.remove()
+    const oldSandbox = sandbox
+    sandbox = undefined
+    if (!oldSandbox) return
+    return oldSandbox.remove()
   })
 }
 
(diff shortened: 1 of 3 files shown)


integration-tests/helpers/index.js
```diff
@@ -942,7 +942,10 @@ function useSandbox (...args) {
 
   after(function () {
     this.timeout(30_000)
-    return sandbox.remove()
+    const oldSandbox = sandbox
+    sandbox = undefined
+    if (!oldSandbox) return
+    return oldSandbox.remove()
```
szegedi:
```suggestion
    return oldSandbox?.remove()
```

IlyasShabi (dismissed):

packages/dd-trace/test/setup/mocha-hooks.js
```diff
@@ -0,0 +1,128 @@
...
+  for (const r of runnables) if (r?.isFailed?.()) return true
+  return false
+}
+
+/**
+ * @param {unknown} value
+ * @returns {value is PromiseLike<unknown>}
+ */
+function isThenable (value) {
+  return value != null && typeof value === 'object' && 'then' in value && typeof value.then === 'function'
```
szegedi: Does `'then' in value` add some value (pun not intended) here considering that `value.then` is evaluated immediately after? It feels like leaving it out would make no difference.

Or even just evaluating it as
```
return typeof value === 'object' && typeof value?.then === 'function'
```
would be enough.

szegedi (changes requested):

packages/dd-trace/test/setup/mocha-hooks.js
BridgeAR: In theory, it should be fine to use a short circuit directly. I just kept it to make sure mocha does not trigger on the lookup some internal flags.

I can check closer if we need it right now.

I definitely want to remove this hack again later. I guess either mocha is updating fast enough with fixes or we will at some point switch to the Node.js test runner anyway.

szegedi (approved):

PR #8150, opened 2026-04-28
chore: add sonar linter

BridgeAR:
This adds the sonar linter and activates a couple of lint rules to detect minor issues that other linter did not find.


eslint.config.mjs
```diff
@@ -454,6 +455,43 @@ export default [
       strict: 'error',
     },
   },
+  {
+    name: 'dd-trace/sonar',
+    // Apply SonarJS to both production and test code. Many SonarJS rules are test-oriented.
+    plugins: {
+      sonarjs: eslintPluginSonar,
+    },
+    rules: {
```
watson, line 464: Why don't we use `eslintPluginSonar.configs.recommended` and then just override the ones we don't want?

watson (dismissed):

eslint.config.mjs
BridgeAR: Sonar has a lot of rules that we do not want as recommended. A lot of them are for cases that do not apply to us. So the list would be very big in this case.

watson (approved):

PR #8161, opened 2026-04-29
fix(llmobs): bound Bedrock token-headers cache

BridgeAR:
The deserialize channel populates `pendingTokenHeaders` for every
Bedrock response, including non-LLM operations and embedding-model
invocations the plugin does not tag. The previous cleanup ran only
inside the LLM-tagging branch of `:request:complete:`, so any other
response leaked an entry keyed by AWS request id for the lifetime
of the process.

Two coordinated pieces close it:

1. `:request:complete:` now consumes the cache entry up front,
   before the operation/model checks, so non-tagged calls release
   their headers like tagged ones do.
2. `pendingTokenHeaders` is now a `TTLCache` (max 10_000, ttl 10
   minutes) instead of a `Map`, so any divergent bookkeeping clears
   itself instead of hand-rolling FIFO eviction.


packages/dd-trace/src/llmobs/plugins/bedrockruntime.js
```diff
@@ -14,7 +15,27 @@ const llmobsStore = storage('llmobs')
...
+ * @typedef {{
+ *   inputTokensFromHeaders?: number,
+ *   outputTokensFromHeaders?: number,
+ *   cacheReadTokensFromHeaders?: number,
+ *   cacheWriteTokensFromHeaders?: number,
+ * }} HeaderTokens
+ */
+
+/** @type {TTLCache} */
+const pendingTokenHeaders = new TTLCache({
```
sabrenner: do we need to use a ttl cache? can we not use a normal `Map` and manually delete as we do below?

packages/dd-trace/src/llmobs/plugins/bedrockruntime.js
```diff
@@ -24,33 +45,38 @@ class BedrockRuntimeLLMObsPlugin extends BaseLLMObsPlugin {
...
+
+      if (!ENABLED_OPERATIONS.has(operation)) return
+
       const { modelProvider, modelName } = parseModelId(request.params.modelId)
 
-      // avoids instrumenting non llm type
-      if (modelName.includes('embed')) {
-        return
-      }
+      // Embedding models do not return token metrics; nothing to tag.
```
sabrenner: nit: this comment change, while not false, does not accurately reflect the original intention of the check

packages/dd-trace/src/llmobs/plugins/bedrockruntime.js
```diff
@@ -110,21 +133,39 @@ class BedrockRuntimeLLMObsPlugin extends BaseLLMObsPlugin {
...
+  } = tokensFromHeaders ?? {}
 
   const inputTokens = usage.inputTokens || inputTokensFromHeaders || 0
   const outputTokens = usage.outputTokens || outputTokensFromHeaders || 0
   const cacheReadTokens = usage.cacheReadTokens || cacheReadTokensFromHeaders || 0
   const cacheWriteTokens = usage.cacheWriteTokens || cacheWriteTokensFromHeaders || 0
 
-  // adjust for the fact that bedrock input tokens only count non-cached tokens
+  // Bedrock input tokens only count non-cached tokens; normalize so the total
+  // reflects everything we sent to the model.
```
sabrenner: small request that comments are not updated unless the corresponding code is touched or significant logic changes require it to be changed, as to preserve original comment intention 😄 (i believe this is an LLM artifact, please correct me if I'm wrong, and if you agree with my assessment maybe we can add a line in our AGENTS.md to reflect that?)

packages/dd-trace/test/llmobs/plugins/aws-sdk/bedrockruntime.token-cache.spec.js
sabrenner, line 1: i would maybe argue instead of this test we could add some benchmark since this was a memory leak issue, and assert the fix that way. the actual behavior of grabbing the tokens themselves is still asserted in the actual LLMObs plugins tests, and i think it doesn't make a ton of sense given our current testing infrastructure to manually import & initialize a plugin to assert the change here is being made properly.

but, i'm not too set on it either way, lmk what you think!

sabrenner (commented):
just a couple thoughts & request on not changing the contents of comments if not needed!

packages/dd-trace/src/llmobs/plugins/bedrockruntime.js
BridgeAR: The TTLCache is not necessary, it was just the safest way to keep the data around as long as possible, if ever needed.

packages/dd-trace/src/llmobs/plugins/bedrockruntime.js
BridgeAR: Shall I remove it? Or what should we use instead? :)

sabrenner: i think just reverting the comment is ok! returning early is still fine and the current comment is still true i think it just doesnt capture what the original comment was saying, that's all. definitely not a blocker or anything haha

packages/dd-trace/src/llmobs/plugins/bedrockruntime.js
sabrenner: cool, maybe we can just use a `Map` if that sounds good to you, ideally so we just don't overcomplicate it if not needed & we're going to manually delete anyhow. I can't think of a case where the response headers are ever not deserialized

packages/dd-trace/test/llmobs/plugins/aws-sdk/bedrockruntime.token-cache.spec.js
BridgeAR: Having a benchmark for this kind of thing is tricky to get stable. We should set up different scenarios to detect memory leaks, but for the current case, I would rather use the test as is.

sabrenner (approved):

